PR #730, opened 2015-10-06
Added Matrix.RotationAxisToRef

jbousquie:
New Matrix.RotationAxisToRef() method added.

This will be useful for saving memory and garbage collection on axis-based mesh morphing.


src/Math/babylon.math.ts
```diff
@@ -2484,12 +2484,17 @@
         }
 
         public static RotationAxis(axis: Vector3, angle: number): Matrix {
+            var result = new Matrix;
```
deltakosh: Should be Matrix()

PR #735, opened 2015-10-07
.OBJ export of Mesh

BitOfGold:


src/Tools/babylon.sceneSerializer.ts
```diff
@@ -951,6 +951,66 @@
...
+            var output = [];
+            output.push("newmtl mat1");
+            output.push("  Ns 10.0000");
+            output.push("  Ni 1.5000");
+            output.push("  d 1.0000");
+            output.push("  Tr 0.0000");
+            output.push("  Tf 1.0000 1.0000 1.0000");
+            output.push("  illum 2");
+            output.push("  Ka 0.0000 0.0000 0.0000");
+            output.push("  Kd 1.0000 1.0000 1.0000");
```
RaananW: Why not actually use the real values from the material? this will export the same material for all meshes.

src/Tools/babylon.tools.ts
```diff
@@ -341,6 +341,22 @@
             }
         }
 
+        //returns a downloadable url to a file content.
+        public static FileAsURL(content:string): string {
+            var fileBlob;
+            try {
+                fileBlob = new Blob([content]);
+            } catch(e) {
+                var blobBuilder = window.BlobBuilder || window.MozBlobBuilder || window.WebKitBlobBuilder;
```
RaananW: BlobBuilder is obsolete and should not be used, https://developer.mozilla.org/en-US/docs/Web/API/BlobBuilder
Since this is a WebGL framework, only the newest browsers are supported, and they all support blob - http://caniuse.com/#search=blob

src/babylon.mixins.ts
```diff
@@ -18,12 +18,21 @@ interface Window {
     SIMD: any;
     AudioContext: AudioContext;
     webkitAudioContext: AudioContext;
+    BlobBuilder: HTMLBLOBBUILDER;
```
RaananW: See the BlobBuilder comment, the entire changes in this file are not needed.

BitOfGold:
You're right, making the changes :)

BitOfGold:
Tested the output and Imported the .OBJ with textured material in .MTL and a .jpg to Blender, looks OK!

deltakosh:
I'll do some updates and keep you posted:)

deltakosh:
So I moved your code to the serializers folder. can you update the [readmd](https://github.com/BabylonJS/Babylon.js/blob/master/serializers/OBJ/readme.md) with instructions on how to use it?

PR #738, opened 2015-10-08
Possible wrong calculation of camera velocity

RaananW:
The velocity, which is the global cameraDirection variable was constantly changed if gravity was enabled. Gravity was added to it over and over again.


deltakosh:
Well done!

PR #744, opened 2015-10-10
Bug fix for material cloning problem

RaananW:


deltakosh:
Ouch! How did I miss that ? Oo

PR #751, opened 2015-10-14
Solid Particle System

jbousquie:
Port of the BJSX Solid Particle System into the BJS core.

It is simplified in term of shapes : we can now only add a classic BJS mesh (no more internal shapes :  _addTriangles_, _addQuads_,_ addTetrahedrons_, etc).
So the usage is : 
```javascript
var mesh = BABYLON.Mesh.CreateXXX();
var sps = new BABYLON.SolidParticleSystem("sps", scene);
sps.addShape(mesh, 50);   // adds 50 particles shaped like the mesh
var spsMesh = sps.buildMesh();

mesh.dispose(); // if not needed somewhere else
```
The particle _uvs_ property is now a _Vector4_ like in the _faceUV_ properties of some meshes and no longer an array(4). Documentation coming soon.

It still needs some tests since the port to TS. It's just the first PR.
This will be announced on the forum only when all the tests will achieved.

### DIFF
--- a/src/Particles/babylon.solidParticleSystem.ts
+++ b/src/Particles/babylon.solidParticleSystem.ts
@@ -1,0 +1,6 @@
+module BABYLON {
+
+    export class SolidParticleSystem implements IDisposable {
+
+    }
+}
--- a/src/Particles/babylon.solidParticleSystem.ts
+++ b/src/Particles/babylon.solidParticleSystem.ts
@@ -1,6 +1,427 @@
 module BABYLON {
 
     export class SolidParticleSystem implements IDisposable {
+        // public members  
+        public particles = [];
+        public nbParticles:number = 0;
+        public billboard: boolean = false;
+        public counter: number = 0;
+        public name: string;
+        public mesh: Mesh;
+        
+        // private members
+        private _scene: Scene;
+        private _positions = [];
+        private _indices = [];
+        private _normals = [];
+        private _colors = [];
+        private _uvs = [];
+        private _index = 0;  // indices index
+        private _shapeCounter = 0;
+        private _useParticleColor = true;
+        private _useParticleTexture = true;
+        private _useParticleRotation = true;
+        private _useParticleVertex = false;
+        private _cam_axisZ = Vector3.Zero();
+        private _cam_axisY = Vector3.Zero();
+        private _cam_axisX = Vector3.Zero();
+        private _axisX = Axis.X;
+        private _axisY = Axis.Y;
+        private _axisZ = Axis.Z;
+        private _camera: Camera;
+        private _particle: SolidParticle;
+        private _fakeCamPos = Vector3.Zero();
+        private _rotMatrix = new Matrix();
+        private _invertedMatrix = new Matrix();
+        private _rotated = Vector3.Zero();
+        private _quaternion = new Quaternion();
+        private _vertex = Vector3.Zero();
+        private _yaw = 0.0;
+        private _pitch = 0.0;
+        private _roll = 0.0;
+        private _halfroll = 0.0;
+        private _halfpitch = 0.0;
+        private _halfyaw = 0.0;
+        private _sinRoll = 0.0;
+        private _cosRoll = 0.0;
+        private _sinPitch = 0.0;
+        private _cosPitch = 0.0;
+        private _sinYaw = 0.0;
+        private _cosYaw = 0.0;
 
+        constructor(name: string, scene: Scene) {
+            this.name = name;
+            this._scene = scene;
+            this._camera = scene.activeCamera;
+        }
+
+        // build the SPS mesh : returns the mesh
+        public buildMesh(): Mesh {
+            if (this.nbParticles == 0) {
+                //this.addTriangles(1, 1);
+            }
+            VertexData.ComputeNormals(this._positions, this._indices, this._normals);
+            var vertexData = new VertexData();
+            vertexData.positions = this._positions;
+            vertexData.indices = this._indices;
+            vertexData.normals = this._normals;
+            if (this._uvs) {
+                vertexData.uvs = this._uvs;
+            }
+            if (this._colors) {
+                vertexData.colors = this._colors;
+            }
+            var mesh = new Mesh(name, this._scene);
+            vertexData.applyToMesh(mesh, true);
+            this.mesh = mesh;
+            return mesh;
+        }
+
+
+        // adds a new particle object in the particles array
+        private _addParticle(p: number, idxpos: number, shape: Vector3[], shapeUV: number[], shapeId: number): void {
+            this.particles.push(new SolidParticle(p, idxpos, shape, shapeUV, shapeId));
+        }
+
+        // add solid particles from a shape model in the particles array
+        public addShape(mesh: Mesh, nb: number): number {
+            var meshPos = mesh.getVerticesData(VertexBuffer.PositionKind);
+            var meshInd = mesh.getIndices();
+            var meshUV = mesh.getVerticesData(VertexBuffer.UVKind);
+            var meshCol = mesh.getVerticesData(VertexBuffer.ColorKind);
+            // shape and shapeUV
+            var posToShape = (positions) => {
+                var shape = [];
+                for (var i = 0; i < positions.length; i += 3) {
+                    shape.push(new BABYLON.Vector3(positions[i], positions[i + 1], positions[i + 2]));
+                }
+                return shape;
+            };
+            var uvsToShapeUV = (uvs) => {
+                var shapeUV = [];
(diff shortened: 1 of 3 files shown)


src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+            this._sinYaw = Math.sin(this._halfyaw);
+            this._cosYaw = Math.cos(this._halfyaw);
+            this._quaternion.x = (this._cosYaw * this._sinPitch * this._cosRoll) + (this._sinYaw * this._cosPitch * this._sinRoll);
+            this._quaternion.y = (this._sinYaw * this._cosPitch * this._cosRoll) - (this._cosYaw * this._sinPitch * this._sinRoll);
+            this._quaternion.z = (this._cosYaw * this._cosPitch * this._sinRoll) - (this._sinYaw * this._sinPitch * this._cosRoll);
+            this._quaternion.w = (this._cosYaw * this._cosPitch * this._cosRoll) + (this._sinYaw * this._sinPitch * this._sinRoll);
+        }
+
+        // internal implemenation of BJS toRotationMatric() with memory reuse
+        private _quaternionToRotationMatrix(): void {
```
deltakosh: Why not in math.ts?

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+            this._rotMatrix.m[12] = 0;
+            this._rotMatrix.m[13] = 0;
+            this._rotMatrix.m[14] = 0;
+            this._rotMatrix.m[15] = 1.0
+        }
+
+        // dispose the SPS
+        public dispose(): void {
+            this.mesh.dispose();
+            this.mesh = null;
```
deltakosh: no need to clear all variables

jbousquie: I thought it was cleaner

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+            this._sinRoll = null;
+            this._cosRoll = null;
+            this._sinPitch = null;
+            this._cosPitch = null;
+            this._sinYaw = null;
+            this._cosYaw = null;   
+        }
+
+        // Optimizers
+        public enableParticleRotation(): void {
```
deltakosh: please use property instead (getter / setter)

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+        public particles = [];
+        public nbParticles:number = 0;
+        public billboard: boolean = false;
+        public counter: number = 0;
+        public name: string;
+        public mesh: Mesh;
+        
+        // private members
+        private _scene: Scene;
+        private _positions = [];
```
deltakosh: number[] ?

src/Particles/babylon.solidParticleSystem.ts
jbousquie: because of the garbage collection : these reimplementation just 
allocates/reallocates only 8 variables
these 8 variables belong to the SPS object
as we iterate over thousands of objects, using the core math functions, 
which create new temp variable each call, will crumble the FPS
this has been over-tested : there's a huge gain to reimplement these 
dedicated methods here

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        // add solid particles from a shape model in the particles array
+        public addShape(mesh: Mesh, nb: number): number {
+            var meshPos = mesh.getVerticesData(VertexBuffer.PositionKind);
+            var meshInd = mesh.getIndices();
+            var meshUV = mesh.getVerticesData(VertexBuffer.UVKind);
+            var meshCol = mesh.getVerticesData(VertexBuffer.ColorKind);
+            // shape and shapeUV
+            var posToShape = (positions) => {
```
deltakosh: You should not create this function each time. Just extract it and mark it as private

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+                var shape = [];
+                for (var i = 0; i < positions.length; i += 3) {
+                    shape.push(new BABYLON.Vector3(positions[i], positions[i + 1], positions[i + 2]));
+                }
+                return shape;
+            };
+            var uvsToShapeUV = (uvs) => {
```
deltakosh: idem

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,427 @@
...
+                var shapeUV = [];
+                if (uvs) {
+                        shapeUV.push(uvs.x, uvs.y, uvs.z, uvs.w);
+                }
+                return shapeUV;
+            };
+            var shape = posToShape(meshPos);
+            var shapeUV = uvsToShapeUV(meshUV);
+            // builder
+            var meshBuilder = (p, shape, positions, meshInd, indices, meshUV, uvs, meshCol, colors) => {
```
deltakosh: idem :)

src/Particles/babylon.solidParticleSystem.ts
jbousquie: oops
I fix it !

I coded it before I created the SolidParticle class

deltakosh: no need for excuse :) I definitely appreciate your work

src/Particles/babylon.solidParticleSystem.ts
deltakosh: Understood!

src/Particles/babylon.solidParticleSystem.ts
jbousquie: ok, this will be done

src/Particles/babylon.solidParticleSystem.ts
jbousquie: will be done also, ok

src/Particles/babylon.solidParticleSystem.ts
jbousquie: so just mesh.dispose() in this case ?

src/Particles/babylon.solidParticleSystem.ts
jbousquie: do you have an example ?

src/Particles/babylon.solidParticleSystem.ts
deltakosh: Yep

src/Particles/babylon.solidParticleSystem.ts
deltakosh: https://github.com/BabylonJS/Babylon.js/blob/332bb3c4bc3a66ee80468e18165bc415ca348993/src/Mesh/babylon.abstractMesh.ts#L10

jbousquie: mmh... I don't really get what I should do
do you prefer  : 

```javascript
public static set ParticleRotation(rot) {
  this._useParticleRotation = rot;
}
```
instead of enable/disable function ?

jbousquie: ok, got it

deltakosh: yep a get method and a set method so users will be able to do xxx.particleRotation = ...

jbousquie: everything done
and compiling :-)

and working (just still a little trick on uvs)

deltakosh:
Thanks dude!! excellent job!

deltakosh:
can you also update this file: http://doc.babylonjs.com/generals/Creating_the_Mini-fied_Version

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -0,0 +1,393 @@
...
+
+            if (this._useParticleColor) {
+                this.mesh.updateVerticesData(VertexBuffer.ColorKind, this._colors, false, false);
+            }
+            if (this._useParticleTexture) {
+                this.mesh.updateVerticesData(VertexBuffer.UVKind, this._uvs, false, false);
+            }
+            this.mesh.updateVerticesData(VertexBuffer.PositionKind, this._positions, false, false);
+            if (!this.mesh.areNormalsFrozen) {
+                var indices = this.mesh.getIndices();
```
deltakosh: indices is not used here

jbousquie: I'll check all this tomorrow

PR #755, opened 2015-10-15
Cannon.js Update

RaananW:
Now supporting the latest Cannon.JS (0.6.2).
Plugin was partly refactored, coordinate-system was adjusted to fit Babylon.js
Tested with the physics example in github and a private physics-controlled scene.


deltakosh:
So we also need to update the cannon.js file right?

RaananW:
Ah, yep :-) I'll update the file soon. 

This is just the start, I wanted to change the way the plugins store the data and add collision events for physics objects.

deltakosh:
Cool:)

PR #761, opened 2015-10-18
Physics api changes and Oimo fixes

RaananW:
Two new functions were added to the physics engine (for low-level manipulations):
1) getWorldObject - will give the developer the low-level world object of the physics engine
2) getPhysicsBodyOfMesh - will return the body object of the specific mesh (or null if none exists).

Oimo now uses Quaternion rotation only, and initialized rotations correctly. It is also partly restructured.

Small bug fix for Cannon.js plugin that will appear if the mesh's center of mass is not in the middle.

Scene's enable physics function - gravity is optional. It will be set to the default if the function will be executed with no parameters at all.

All tested with the Gravity scene from the samples.


RaananW:
Wait with merging, further changes coming.

RaananW:
All should be fine now.

PR #763, opened 2015-10-19
Fix.sps

jbousquie:
* added type declarations
* changed _setParticleXXX_ property names to _computeParticleXXX_ property names, confusing since the getter/setter
* fix : particles scaled first, then only rotated

### DIFF
--- a/src/Particles/babylon.solidParticleSystem.ts
+++ b/src/Particles/babylon.solidParticleSystem.ts
@@ -18,10 +18,10 @@ module BABYLON {
         private _uvs: number[] = new Array<number>();
         private _index: number = 0;  // indices index
         private _shapeCounter: number = 0;
-        private _setParticleColor: boolean = true;
-        private _setParticleTexture: boolean = true;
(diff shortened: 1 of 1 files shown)


src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -2,53 +2,53 @@ module BABYLON {
 
     export class SolidParticleSystem implements IDisposable {
         // public members  
-        public particles = new Array<SolidParticle>();
-        public nbParticles = 0;
-        public billboard = false;
-        public counter = 0;
+        public particles: SolidParticle[] = new Array<SolidParticle>();
```
deltakosh: No need to precise the type when you are initializing the variable

jbousquie: I did it because I plan to try a test with asm.js later and having the types will be easier to convert the numerical ones to the right low level types

deltakosh: Oki doki  👌

PR #764, opened 2015-10-20
Feature.sps

jbousquie:
* Added the parameter _updatable_ to the method _buildMesh()_ : this allows to create a non-updatable mesh, so very performant (ex : many immutable objects)

* Added Float32Array, once the mesh is built, to improve perfs on calls to setVerticesData/updateVerticesData


src/Mesh/babylon.meshBuilder.js
```diff
@@ -330,6 +330,7 @@ var BABYLON;
             var updatable = options.updatable;
             var sideOrientation = options.sideOrientation || BABYLON.Mesh.DEFAULTSIDE;
             var instance = options.instance;
+            options.arc = (options.arc <= 0) ? 1.0 : options.arc || 1.0;
```
deltakosh: Don't you want to publish the .ts instead?

jbousquie: I don't know why this file is here, I didn't touch it... I guess it came with my pull upstream
grrr, again git and my fat fingers :-(

skip this file please
I hope someone else (RaananW) fixed the arc default value from these last days in the TS

PR #766, opened 2015-10-21
Feature.sps

jbousquie:
* Added _vertexFunction_ and _positionFunction_ properties : you can now modify your SPS mesh at creation time for ever and set it as non _updatable_ if you don't need to update it afterwards
* Some little fixes


RaananW:
yei :-)

PR #770, opened 2015-10-21
Optimization default submesh creation

RaananW:
Minimum and maximum can be reused from the mesh's bounding info object,
as it is the same. This removes one more call to
Tools.ExtractMinAndMaxIndexed , a very expensive calculation.


deltakosh:
Unfortunately I do not think so:( 
The minimum and maximum of the submesh can be different as it works on a subset of the overall mesh

deltakosh:
Excellent idea!

PR #775, opened 2015-10-23
OBJLoader

Jaskar:
Correct the loading, taking into account the UVs.
Set a new parameter to specify to take it into account or not.
Update the readme

Asked by @Temechon , following this topic http://www.html5gamedevs.com/topic/18006-babylonjs-very-slow/

### DIFF
--- a/loaders/OBJ/README.md
+++ b/loaders/OBJ/README.md
@@ -29,6 +29,20 @@ BABYLON.SceneLoader.ImportMesh("batmanface", "", "batman.obj", scene, function (
 * A Babylon.Mesh will be created for each object/group
 * The obj model should be exported with -Z axis forward, and Y axis upward to be compatible with Babylon.js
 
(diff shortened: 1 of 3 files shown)


Temechon:
And fix the texture bug here : http://www.html5gamedevs.com/topic/17186-incorrect-obj-loader/?hl=loader

RaananW:
Looking good! How much does it impact the performance, when the flag is true?

Jaskar:
With an object like the desk on this topic : http://www.html5gamedevs.com/topic/18006-babylonjs-very-slow/ , 5000ms without, 14000ms with. It's the reason why I add a flag.
With this one : http://i.imgur.com/IkGNZA9.png
300ms without, 400ms with.

RaananW:
na na na na Batman! :-)
Cool. Thanks for the fix!

Temechon:
Go Batman ! :+1:

PR #785, opened 2015-10-27
Surface heightmap

RaananW:
Any flat (and square) surface (ribbons for example) can now be a height map.
Fixed a bug with updateBodyPosition and the calculation of transformation of the physics body.


RaananW:
Demo coming the minute it's possible :-)

deltakosh:
And it's on the PG as well :)

RaananW:
awesome. playground created! :fireworks:

PR #786, opened 2015-10-28
Adding water and fire materials to the materialsLibrary

julien-moreau:
Adding water and fire materials to the materialsLibrary

- Added fire and water textures
- Added a skybox to the controls to test reflection with water and others


RaananW:
awesome

PR #791, opened 2015-10-29
Materials library

julien-moreau:
Materials library:
- Improved water material, the sourceMesh parameter is not needed
- added readme


materialsLibrary/materials/water/readme.md
```diff
@@ -0,0 +1,41 @@
+# Water material
```
deltakosh: You should add links to demo on PG

materialsLibrary/test/index.html
```diff
@@ -144,7 +144,7 @@
 				simple.diffuseTexture.uScale = 5;
 				simple.diffuseTexture.vScale = 5;
 				
-				var water = new BABYLON.WaterMaterial("water", scene, currentMesh);
+				var water = new BABYLON.WaterMaterial("water", scene);
 				water.backFaceCulling = false;
 				water.enableRenderTargets(false);
```
deltakosh: Why?

julien-moreau: The water material no longer needs a source mesh since we can have the mesh reference in the isReady function

deltakosh: my why is about the enableRenderTargets. Why this function?

materialsLibrary/materials/water/readme.md
julien-moreau: In the PG I still need to reference the source mesh in the constructor :(

materialsLibrary/test/index.html
julien-moreau: Oh, the functon definition is:
    public enableRenderTargets(enable: boolean): void;
true to enable and false to disable.
It is enabled by default

materialsLibrary/materials/water/readme.md
deltakosh: Should be good now

materialsLibrary/test/index.html
deltakosh: Ok understood!

PR #794, opened 2015-11-02
Allow CORS on HTML5 audio element

rbozan:
This allows the newly added HTML 5 audio element (streaming: true) to access audio files from outside of the local server.


deltakosh:
hello you have yo submit PR for .ts files:) cc@davrous

davrous:
I've done the merging manually directly in the TypeScript file.

PR #799, opened 2015-11-02
Variable Bone Influencers

Palmer-JC:
Fully implemented on BJS side.  Limited testing done, but this is alpha.
Not tested:
- Any shader other than default
- CPU fallback (added message though)
- FileLoader.  Probably better way to expand, but not standard TS not
understood.


deltakosh:
I fixed some issues. We should be good now

PR #815, opened 2015-11-11
enable POV at node level

Palmer-JC:
Allow TargetCamera sub-classes, that have no target, as well as Spolight & DirectionalLight to be positioned & "rotated" in amountRight-amountUp-mountForward & flipBack-twirlClockwise-tiltRight terms.

When integrated into the POV replacement extension, QueuedInterpolation, a scene beforerender will be used instead of a mesh beforerender.

### DIFF
--- a/src/Cameras/babylon.camera.ts
+++ b/src/Cameras/babylon.camera.ts
@@ -152,8 +152,9 @@
 
         private _globalPosition = Vector3.Zero();
 
-        constructor(name: string, public position: Vector3, scene: Scene) {
+        constructor(name: string, position: Vector3, scene: Scene) {
             super(name, scene);
+            this.position = position;
 
             scene.addCamera(this);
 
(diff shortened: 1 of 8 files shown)


src/Cameras/babylon.camera.ts
```diff
@@ -152,8 +152,9 @@
 
         private _globalPosition = Vector3.Zero();
 
-        constructor(name: string, public position: Vector3, scene: Scene) {
+        constructor(name: string, position: Vector3, scene: Scene) {
```
deltakosh: why this change? this is the same behavior

src/Lights/babylon.directionalLight.ts
```diff
@@ -16,9 +15,10 @@
         private _orthoTop = Number.MIN_VALUE;
         private _orthoBottom = Number.MAX_VALUE;
 
-        constructor(name: string, public direction: Vector3, scene: Scene) {
+        constructor(name: string, direction: Vector3, scene: Scene) {
```
deltakosh: Same here? Why this change?

deltakosh:
I have to disagree with this changes. I do not want to pollute top level classes with properties required by some children only

Palmer-JC:
I will go back to yesterday repository, EOD. Implement this with code all in extension.

PR #817, opened 2015-11-12
Disposing the skeleton when disposing the mesh

RaananW:
I am not sure if skeletons are reusable, but if not - it makes sense to actually dispose them when disposing the mesh to prevent from disposing two objects one after the other.


deltakosh:
Not a good idea. Skeleton can be shared between meshes

PR #819, opened 2015-11-13
Fix.compute normal

jbousquie:
* no more JS objects, neither per property value access
* reduced the number of temporary variables
* only scalar values and array access

This optimization is a consequence of my profiling tests when coding the SPS : huge mesh to be updated each frame after many vertex recomputations.


deltakosh:
Excellent update!

jbousquie:
tinkiou

PR #820, opened 2015-11-13
Added terrain material with bump mapping support to materials library

julien-moreau:
- Based on a mixmap (RGB)
- Support of bump mapping
- Documentation is coming


deltakosh:
This is excellent!

deltakosh:
Doc is required as you mentioned :)

julien-moreau:
I'd like to add some textures to the playground and create a better Mixmap to create an awesome playground example in the documentation

deltakosh:
Feel free to send me those textures

PR #821, opened 2015-11-13
Feature.cylinder.has rings

jbousquie:
Added the boolean parameter _hasRings_ to `MeshBuilder.CreateCylinder()`  (default = false).  
This parameter duplicates the vertices in the geometry when there is a _subdivision_ value greater than 1. Each subdivision is thus independent in term of vertices and is then call a "ring".  

This will allow to set different colors and textures per ring. (next upcoming PR, soon)


jbousquie:
actually this commit, called "Fixing cache issue for computeWorldMatrix" https://github.com/BabylonJS/Babylon.js/commit/8971e61102d97d5aee13455f633a25d73069d4f3
kept everything right for CreateCylinder (hasRings) in the file VertexData.ts but it removed the former right ComputeNormals() commit (optimization)

Then this commit, , called "Cleaning code for PRB"  https://github.com/BabylonJS/Babylon.js/commit/534f87a3f70c5264310185aa4d34db6801f06ae8 removed the Cylinder hasRings feature in the file VertexData.ts (and doesn't set back the right ComputeNormals() optimized algo either)

RaananW:
Should this be reverted then?

jbousquie:
I don't know how this works, when many things have been commited after ...

RaananW:
you might have mixed a few branches and commits in your local git. I am not too sure why every PR of yours starts with "commit before merge". Try finding locally if you commited soemthing with a branch that was not updated.
I am soon checking here how it can be restored.

RaananW:
Sorry, I stand correct - as you said in the forum, your changes were overwritten :-) I am remerging them now.

jbousquie:
No I didn't everything was clear here. Check my PR merges please, they are good.

The initial "commit before merge" is because when I merge upstream/master, it forces me to commit the last modified JS file I generated locally (here : SolidParticleSystem.**js** which has nothing to do with VertexData.ts), probably because the linux typescript compilation is slightly different from the windows one.

RaananW:
I get those too... you can simply discard those changes instead of commiting them. would be better for the commit-list. Discard after sending the PR to mvoe from one branch to the other. Yep,

jbousquie:
ok
it forces me either to commit either to stash. 
Should I stash instead ?

RaananW:
Yep, stash would work just find.
This is a nice two-liner: https://stackoverflow.com/questions/52704/how-do-you-discard-unstaged-changes-in-git

jbousquie:
nice !!!

PR #822, opened 2015-11-15
Update UV mapping on icosphere

regis-c:
- better use of texture surface compare to previous map
  Use 2 islands of triangle to cover about 61 % of the square bitmap
- create alias vertex (case of same 3D position with different UV)


RaananW:
Oh..
the conflics are my fault. Jerome siw roking on create cylinder and I have forced updated his changes after they were deleted.
Could you try merging the chanegs and fix the conflits?
Sorry again...

RaananW:
I merged the PR directly into the master. Please update the readme with the change

PR #825, opened 2015-11-18
Events can now be linked to animations

Temechon:
.gitignore updated for jetbrains editors. The documentation will also be updated soon.


src/Animations/babylon.animation.js
```diff
@@ -64,7 +78,25 @@ var BABYLON;
...
+        };
+        /**
+         * Remove all events found at the given frame
+         * @param frame
+         */
+        Animation.prototype.removeEvents = function (frame) {
+            for (var index = 0; index < this._events.length; index++) {
+                if (this._events[index].frame === frame) {
+                    this._events.splice(index, 1);
+                    return;
```
RaananW: This will stop at the first event it finds. is there a possibility that the frame is shared between events?

src/Animations/babylon.animation.ts
```diff
@@ -79,7 +90,27 @@
...
+
+        /**
+         * Remove all events found at the given frame
+         * @param frame
+         */
+        public removeEvents(frame:number) : void {
+            for (var index = 0; index < this._events.length; index++) {
+                if (this._events[index].frame === frame) {
+                    this._events.splice(index, 1);
+                    return;
```
RaananW: This will stop at the first event it finds. is there a possibility that the frame is shared between events?

Temechon:
You are right, this "return" should not be here, because an animation can have several events linked to the same frame.

RaananW:
Oh, you could have simply pushed a change :-)

PR #826, opened 2015-11-18
Events can now be added to animations.

Temechon:


src/Animations/babylon.animation.ts
```diff
@@ -79,7 +90,26 @@
...
+        }
+
+        /**
+         * Remove all events found at the given frame
+         * @param frame
+         */
+        public removeEvents(frame:number) : void {
+            for (var index = 0; index < this._events.length; index++) {
+                if (this._events[index].frame === frame) {
+                    this._events.splice(index, 1);
```
RaananW: Sorry, should have noticed before - this will skip the element after the splice. You will need to deduct 1 from the index (or use splice(index--,1)), otherwise some elements won't be iterated...

Leave this open, just push a change to this branch and it will appear here.

src/Animations/babylon.animation.ts
```diff
@@ -421,6 +451,18 @@
             var currentFrame = returnValue ? from + ratio % range : to;
             var currentValue = this._interpolate(currentFrame, repeatCount, this.loopMode, offsetValue, highLimitValue);
 
+            // Check events
+            for (var index = 0; index < this._events.length; index++) {
+                if (this._events[index].frame >= currentFrame) {
+                    var event = this._events[index];
+                    // If event should be done only once, remove it.
+                    if (event.onlyOnce) {
+                        this._events.splice(index, 1);
```
RaananW: Same here (index--)

Temechon:
No problem at all, thank you for your review :)

PR #827, opened 2015-11-18
added per ring + per quad section if enclosed to VertexData.CreateCylinder

jbousquie:
```javascript
MeshBuilder.CreateCylinder('c', {subdivisions: 3, hasRings: true, arc: 0.6, enclose: true, faceColors: colors}, scene);
```
If colors is an array of `Color4`, we can now define a color per : 
* subdivision if `hasRings` if set to true
* even per section quad if the cylinder is sliced with an `arc`

_faceColors_ is an array of `Color4` per cylinder surface.
first surface : bottom cap
last surface : top cap

Each other surface if a tubular ring surface (so as many as subdivisions if `hasRings` is true, one else), so one color4 element in the _faceColor_ array. 
If the cylinder is sliced with `arc`, whatever the value of `arc` to keep consistent, there are 2 more quad surfaces per ring, so 3 color4 elements in this case (1 tubular surface + 2 quads).

### DIFF
--- a/src/Mesh/babylon.mesh.vertexData.ts
+++ b/src/Mesh/babylon.mesh.vertexData.ts
@@ -857,12 +857,19 @@
             var j: number;
             var r: number;
             var ringIdx: number = 1;
+            var c: number = 1;
 
             for (i = 0; i <= subdivisions; i++) {
                 h = i / subdivisions;
                 radius = (h * (diameterTop - diameterBottom) + diameterBottom) / 2;
                 ringIdx = (hasRings && i !== 0 && i !== subdivisions) ? 2 : 1;
                 for (r = 0; r < ringIdx; r++) {
+                    if (hasRings) {
+                        c += r;
+                    }
+                    if (enclose) {
+                        c += 2 * r;
+                    }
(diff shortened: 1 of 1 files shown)


src/Mesh/babylon.mesh.vertexData.ts
```diff
@@ -857,12 +857,19 @@
...
 
             for (i = 0; i <= subdivisions; i++) {
                 h = i / subdivisions;
                 radius = (h * (diameterTop - diameterBottom) + diameterBottom) / 2;
                 ringIdx = (hasRings && i !== 0 && i !== subdivisions) ? 2 : 1;
                 for (r = 0; r < ringIdx; r++) {
+                    if (hasRings) {
+                        c += r;
+                    }
+                    if (enclose) {
```
RaananW: if? or else if?

jbousquie: "if" : I confirm
those are two different tests to increment the index c

jbousquie: both can be true in the same time, and we need then to +r and +2*r

PR #836, opened 2015-11-19
Physics serialization

RaananW:
The Physics engine can now be defined in the babylon file (optional).
if the physics engine is defined, one can select between the two plugins that BJS has and set the gravity individually.
If it is not defined, but bodies have physics enabled, default values will be used.

### DIFF
--- a/src/Physics/babylon.physicsEngine.ts
+++ b/src/Physics/babylon.physicsEngine.ts
@@ -1,5 +1,6 @@
 ﻿module BABYLON {
     export interface IPhysicsEnginePlugin {
+		name: string;
         initialize(iterations?: number);
         setGravity(gravity: Vector3): void;
         runOneStep(delta: number): void;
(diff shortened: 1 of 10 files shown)


src/Physics/babylon.physicsEngine.ts
```diff
@@ -54,6 +56,10 @@
             this.gravity = gravity || new Vector3(0, -9.807, 0);
             this._currentPlugin.setGravity(this.gravity);
         }
+		
+		public _getGravity() : Vector3 {
```
Temechon: Why put an underscore here as the method is public ?

RaananW: the rest of the functions (like setGravity) also have an underscrore, i assumed it is due to the direct interaction with the plugin. Can be taken out actually (for all other functions as well)

deltakosh: it is internal actually (public _)

PR #838, opened 2015-11-21
Added a button to log the camera object in the debug layer

Temechon:


RaananW:
that would be helpful :)

PR #842, opened 2015-11-24
Pasring functions

RaananW:
parsing functions.
Also added Int32Array as type for indices all over.


RaananW:
Gulp compiles correctly, I need a 3rd (and 4th) eye to check that everything is ok.

PR #847, opened 2015-11-29
Update babylon.babylonFileLoader.ts

dad72:
Optimizing loops. This prevents each loop has to recalculate the length of the array/objects


dad72:
OOps, I made a small error on line 2. I cancel the PR and I remade into other

PR #858, opened 2015-12-04
ParticleSystem serialize and parse

wyklion:
1. add '.name' to serialization.
2. support object emitter and position emitter.


src/Particles/babylon.particleSystem.ts
```diff
@@ -449,7 +449,12 @@
         public serialize(): any {
             var serializationObject: any = {};
 
-            serializationObject.emitterId = this.emitter.id;
+            serializationObject.name = this.name;
+            if (this.emittor.position) {
```
Temechon: It should be emitter and not emittor ?

wyklion: haha... u're good!

PR #859, opened 2015-12-04
Update babylon.particleSystem.js

wyklion:


Temechon:
Hi, 

No need to commit the js file, it will be overwritten by the next compilation.

PR #860, opened 2015-12-04
Master

punkoffice:
Unity prefab instances will now be exported as Babylon.js mesh instances.  The Babylon.js mesh will be the Unity prefab.  Details from the first instance (position, rotation, scale) are copied into the Babylon.js mesh then that first instance is removed from the list of instances


deltakosh:
Woot! Live it! I'll check it this week end.
Thank you for this great contrib

Do not forget to update the what's new for 2.3:)

punkoffice:
Awesome!  This isn't very flexible though.  You can't change materials or components, but its still a useful feature.

deltakosh:
Yes instances are performant at the cost of flexibility

punkoffice:
so is this considered a "major update" or just an "update"?  I'm in the process of updating the What's New file now

PR #863, opened 2015-12-07
particle create file and load from file.

wyklion:


dad72:
Hi,
Changes made in the JS files are not necessary, the JS are generated from the TS file.
In PR 'particle creat file and load file', you modified more code in the JS file to compare TS file. I think in this case, the TS file will overwrite your changes the JS file

RaananW:
Hi,

Thanks for the PR!
We are trying to keep the loading to a single major function (the scene load and append functions). Those functions are already able to do that. Starting to export different file types for different elements of a scene won't do any good to the babylon file structure.

However, you have some extra properties that are missing in the serialize and loader function, so it would be great having them! If you could create a new PR with those changes only, it would be wonderful.

As dad72 previously said, please make changed to .ts files only. The JS files are being compiled using gulp and are automatically merged once we run a clean build.

Waiting for your amended PR! I am closing this one.
Thanks again!

PR #874, opened 2015-12-21
audio negative startOffset bug

wyklion:
1. play(10)
3. play() error...


wyklion:
let me think more. fix it later..

PR #877, opened 2015-12-21
MeshImpostor improvement

RaananW:
Using a new Cannon.js feature, MeshImpostor now uses a triangular mesh
implementation.
MeshImpostor can only collide against Spheres at the moment. This is a
cannon.js constraint. But better than nothing.


RaananW:
(tmp) demo - http://raananweber.com/physics/

PR #881, opened 2015-12-27
- MapPanCamera implemented

mstdokumaci:
This camera pans only on XoZ plane and looks at a game map with a fixed beta angle.

### DIFF
--- a/Tools/Gulp/config.json
+++ b/Tools/Gulp/config.json
@@ -43,6 +43,7 @@
       "../../src/Cameras/babylon.followCamera.js",
       "../../src/Cameras/babylon.touchCamera.js",
       "../../src/Cameras/babylon.arcRotateCamera.js",
+      "../../src/Cameras/babylon.mapPanCamera.js",
       "../../src/Rendering/babylon.renderingManager.js",
       "../../src/Rendering/babylon.renderingGroup.js",
       "../../src/babylon.scene.js",
--- a/src/Cameras/babylon.mapPanCamera.js
+++ b/src/Cameras/babylon.mapPanCamera.js
@@ -1,0 +1,330 @@
+var __extends = (this && this.__extends) || function (d, b) {
+    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
+    function __() { this.constructor = d; }
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
+};
+var BABYLON;
+(function (BABYLON) {
+    var eventPrefix = BABYLON.Tools.GetPointerPrefix();
+    var MapPanCamera = (function (_super) {
+        __extends(MapPanCamera, _super);
+        function MapPanCamera(name, beta, radius, target, scene) {
+            _super.call(this, name, BABYLON.Vector3.Zero(), scene);
+            this.beta = beta;
+            this.radius = radius;
+            this.target = target;
+            this.inertialRadiusOffset = 0;
+            this.lowerRadiusLimit = null;
+            this.upperRadiusLimit = null;
+            this.wheelPrecision = 3.0;
+            this.pinchPrecision = 2.0;
+            this.panningSensibility = 50.0;
+            this.inertialPanningX = 0;
+            this.inertialPanningZ = 0;
+            this.keysUp = [38];
+            this.keysDown = [40];
+            this.keysLeft = [37];
+            this.keysRight = [39];
+            this.zoomOnFactor = 1;
+            this._keys = [];
+            this._viewMatrix = new BABYLON.Matrix();
+            if (!this.target) {
+                this.target = BABYLON.Vector3.Zero();
+            }
+            this.getViewMatrix();
+        }
+        MapPanCamera.prototype._getTargetPosition = function () {
+            return this.target.position || this.target;
+        };
+        // Cache
+        MapPanCamera.prototype._initCache = function () {
+            _super.prototype._initCache.call(this);
+            this._cache.target = new BABYLON.Vector3(Number.MAX_VALUE, Number.MAX_VALUE, Number.MAX_VALUE);
+            this._cache.radius = undefined;
+        };
+        MapPanCamera.prototype._updateCache = function (ignoreParentClass) {
+            if (!ignoreParentClass) {
+                _super.prototype._updateCache.call(this);
+            }
+            this._cache.target.copyFrom(this._getTargetPosition());
+            this._cache.radius = this.radius;
+        };
+        // Synchronized
+        MapPanCamera.prototype._isSynchronizedViewMatrix = function () {
+            if (!_super.prototype._isSynchronizedViewMatrix.call(this))
+                return false;
+            return this._cache.target.equals(this._getTargetPosition())
+                && this._cache.radius === this.radius;
+        };
+        // Methods
+        MapPanCamera.prototype.attachControl = function (element, noPreventDefault) {
+            var _this = this;
+            var cacheSoloPointer; // cache pointer object for better perf on camera rotation
+            var previousPinchDistance = 0;
+            var pointers = new BABYLON.SmartCollection();
+            if (this._attachedElement) {
+                return;
+            }
+            this._attachedElement = element;
+            var engine = this.getEngine();
+            if (this._onPointerDown === undefined) {
+                this._onPointerDown = function (evt) {
+                    // manage pointers
+                    pointers.add(evt.pointerId, { x: evt.clientX, y: evt.clientY, type: evt.pointerType });
+                    cacheSoloPointer = pointers.item(evt.pointerId);
+                    if (!noPreventDefault) {
+                        evt.preventDefault();
+                    }
+                };
+                this._onPointerUp = function (evt) {
+                    cacheSoloPointer = null;
+                    previousPinchDistance = 0;
+                    //would be better to use pointers.remove(evt.pointerId) for multitouch gestures,
+                    //but emptying completly pointers collection is required to fix a bug on iPhone :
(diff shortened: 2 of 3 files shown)


RaananW:
Hi, thanks for the contribution!

Just a few questions - 
1) what is the difference between this and the Ctrl-Panning of the arc rotate camera?
2) This looks like an exact copy of the ArcRotateCamera, apart from small changes - why not extend the ArcRotateCamera instead of copying everything?

mstdokumaci:
1a) There is no option for making ctrl-panning work as default (without ctrl) in ArcRotateCamera because that behaviour is hard-coded.
1b) ArcRotateCamera is panning XoY plane, that's not what you need when user wants to pan a game map. You need panning XoZ plane, like a player of the famous mobile game Clash of Clans, moving around own village.
2a) Because this camera is much more simple than ArcRotateCam, I would need to override many functions just to remove them and it sounds weird for extending. It's not actually extending it, mostly shrinking :)
2b) There were conflicts to extend ArcRotateCam. It was not possible to extend and change the functionality, because of some private variables on the way. And I didn't want to touch ArcRotateCam code at all.

mstdokumaci:
Here is a demonstration:
https://plnkr.co/edit/rUY0w1J4PoksR7Vurohu

deltakosh:
Does it support collisions like arcrotate?

mstdokumaci:
No it doesn't support collisions. I thought it was not necessary because he camera is overlooking the map already. I don't think it's hard to implement if necessary though.

deltakosh:
I think this is great work but I'm questioning myself is this is a feature we want in the core or in the extensions

RaananW:
I think extending the Arc rotate camera to somehow support this feature will be the best. It feels like "already implemented", but I do agree that XZ makes more sense than XY for panning.

mstdokumaci:
Providing an XoY panning is actually easier because mouse and touch actions also has X and Y values on 2 dimension. When you apply these x & y to XoY plane of meshes, it perfectly fits. Things move right-left or up-down when you move your mouse right-left or up-down.

XoZ panning on the other hand, has this difficulty. If you transform mouse event 2d coordinates using camera transform matrix, a y value also appears and the camera moves like zooming in and out. Panning I implemented for MapPanCamera moves the camera above the scene while protecting both of it's overlook angle and distance radius then moves the target point of camera in XoZ plane only. That's why I removed alpha angle and fixed it to 0 degrees. If this thing will be applied to ArcRotateCam, calculation of camera transform of panning direction should be altered to convert event.x & event.y into XoZ values only.

RaananW:
I am totally not against this, i really like it! I just think that having a specific camera that does ONLY that is a bit redundant. 
This camera, in its core, is a (limited) Arc-Rotate camera that pans above a specific plane. Taking your great implementation and using is as a general implementation in the Arc Rotate Camera might be more helpful. Because then the users can decide what they are using, instead of moving from one camera to the other. 
And, as DK said - the extensions library (or BabylonJSX) is a perfect place to try this and see how it fits in the framework.

mstdokumaci:
Maybe this is all because of keeping control logic in camera class. Maybe there should be separate control functions attachable to any html element & camera separately. So developers choose to attach any of these controls to any key, mouse or touch event coming from whichever html element they want, and updating position of any camera, maybe even any mesh. This will abstract action controls from cameras. Then we won't need to implement TouchCamera or PansInThisDirectionCamera, RotatesInThisAxisCamera.

deltakosh:
The main reason for this behavior is simplicity. I do not want users to connect tons of features to get the camera they want. It should be provided out of the box for the sake of simplicity.

deltakosh:
Under the hood I'm not opposed to decouple things if this remains simple to use.

vousk:
This could be done with base ArcRotate simply by freezing alpha beta and radius (with limits), and then moving the target alongside the mouse or finger movement, no ?

vousk:
I also ear mstdokumaci arguments, would be very handy and easy to have a unique camera with behavior modules we can attach to it (free flying module, orbital movement module, pointer event module, panning module, targetfollower module, etc...). But that would imply big thinking and vast change of the core cameras managment (not even sure such a camera is possible)...

mstdokumaci:
> This could be done with base ArcRotate simply by freezing alpha beta and radius (with limits), and then moving the target alongside the mouse or finger movement, no ?
Yes, plus changing the panning plane from XoY to XoZ and  that's what you can't do by just extending ArcRotate camera.

mstdokumaci:
I tried doing same thing with PlayCanvas:
https://plnkr.co/edit/H28v8oBoq4SP38aJWXMT

I didn't need to fork the library or send a pull request. Because camera behaviours were completely scriptable. I feel that, this way is taking me to a result easier as a developer. Just consider.

deltakosh:
This is great but a bit complicated for beginners. It is hard to satisfy everyone:(

deltakosh:
But tbh I think we should rethink your first assertion: I think it is easy to just inherit from ArcRotateCamera and just change the behavior you want with an overloaded function. (Unless I missed something which happens often:))

mstdokumaci:
I tried extending ArcRotateCamera first but failed due to some private variables. Anyone interested can try his/her chance :)

deltakosh:
Perhaps we can work together to expose some variables?

RaananW:
I'll try extending the ARcRotate. Will update you guys here.

BTW, about PlayCanvas - you can obviously work like that with Babylon.js - A camera is just an object holding functions. Your code will work, if I copy it to my local project and use it. The difference here is that Babylon offers predefined cameras to the users, so they wouldn't have to do that themselves. We are, however, trying to minimize the amount of such cameras, to keep the confusion level at minimum :smile:

RaananW:
ok!
So, with adding a new panning-axis variable to the ArcRotate camera, I have achieved this:
http://www.babylonjs-playground.com/#14KLYA
You can see a few examples you can play with. The current one is XZ panning. Use the right mouse click to pan.

vousk:
Nicely done Raanan !! I was pretty sure this would be possible directly inside the ArcRotate. What I like here is we keep the ability to orbit the camera.

deltakosh:
This could be added as an official parameter to ArcRotateCamera

mstdokumaci:
Thanks for the code! is it also possible to make ctrl for panning behaviour optional? I mean with a flag, panning becomes default behaviour without ctrl. And I guess all these requires editing ArcRotateCam code and releasing a new version. Am I right?

RaananW:
I'll add it later today. 
About the default ctrl behaviour - the flag in attach control is working, but you can simply change the default value to false. This way, the only way to get ctrl is to pass true as the third variable.

vousk:
Raanan I've already added (some month ago) the ability to deactivate ArcRotateCamera panning mechanism (by setting panningSensibility to 0), if we want to add a flag for managing panning we can play with this param.

RaananW:
I think that @mstdokumaci  (and correct me if I am wrong) wanted to cancel the ctrl-addition to the panning and use right-click only. This is possible with the attachControl function. It has to be done there due to the contextmenu event that needs to be registered if ctrl is not being used.

RaananW:
https://github.com/BabylonJS/Babylon.js/pull/913

vousk:
Ah ok. But the attachControl third flag is already there to choose between CTRL+left click or only rigth click... Maybe @mstdokumaci wants to pan with left click (without CTRL) ?

mstdokumaci:
> Ah ok. Since the attachControl third flag is already there to choose between CTRL+left click or only rigth click, maybe @mstdokumaci wants to pan with left click (without CTRL) ?

Yes, for creating a mobile game that was my intention. Panning as default behaviour with no right click or ctrl. Because player doesn't need to rotate the map, just needs to fly around it. But I couldn't find a way to extend ArcRotateCam without editing it's code in a TypeScript way because some of the variables were private. And I didn't want to edit ArcRotateCam code because it might break someone's code who already uses it.

vousk:
Ok. I have an idea on how to answer this problem (it will also solve the "I-find-it-not-so-nice" ctrlpanning flag in attachControl signature). I won't be on my computer for two days but I'll then send a proposal PR.

caracal7:
But how to pan by left mouse button without CTRL? Anybody solve this issue?

deltakosh:
There is a parameter in the arcrotatecamera.attachControl to enable this behavior 
http://doc.babylonjs.com/classes/3.1/arcrotatecamera#attachcontrol-element-nopreventdefault-usectrlforpanning-panningmousebutton-rarr-void

artfiedler:
Would be great if I could find a working example of getting this to work. Happen to have one? I'm looking to stop the rotating of the camera and just pan XY at a 60 deg angle or so with zoom. Without ctrl. Would be great to add it to these comments as bing.com brought me here.

deltakosh:
When calling camera.attachControl, set the ctrlForPanning parameter to false:
https://doc.babylonjs.com/api/classes/babylon.arcrotatecamera#attachcontrol

Stop rotating: set angularSensibilityX (and Y) to 0

artfiedler:
Ok, I got the ctrl for panning part already, not sure what numbers it expects for buttons so I just left that off, and thanks the angularSensibility tip set that, now when I click the mouse button and drag, the entire scene disappears. I tried setting panningSensibility to a really low number just to see if it was panning too fast or fair or something

artfiedler:
https://www.babylonjs-playground.com/#AVK7Y3

artfiedler:
angularSensibilityX & Y = 0 makes it spin so fast it stops rendering that's why the scene disappears. I figured out how to make it pan like a map https://www.babylonjs-playground.com/#9505R9 only problem now is have to figure out the panningSensibility so its 1:1 ratio from where you click and drag and release... I suppose I can turn off the rotation by preventing default

By the way button 0: left click, 1: middle click, 2: right click

PR #887, opened 2016-01-04
fix SPS mesh name that wasn't correctly set

jbousquie:
fix the SPS underlying mesh name that wasn't correctly set


RaananW:
I wonder why the build passed until now...

PR #893, opened 2016-01-06
new ActionManager triggers : OnPickDownTrigger, OnLongPressTrigger

vousk:
modified ActionManager.OnPickTrigger to be raised on real single click/tap (immediate action, not while dragging/swiping), added ActionManager.OnPickDownTrigger to replace previous pick trigger (raised even while moving), and added ActionManager.OnLongPressTrigger to detect long (>500ms) not moving click/press

### DIFF
--- a/src/Actions/babylon.actionManager.ts
+++ b/src/Actions/babylon.actionManager.ts
@@ -57,14 +57,16 @@
         private static _OnLeftPickTrigger = 2;
         private static _OnRightPickTrigger = 3;
         private static _OnCenterPickTrigger = 4;
-        private static _OnPointerOverTrigger = 5;
-        private static _OnPointerOutTrigger = 6;
-        private static _OnEveryFrameTrigger = 7;
-        private static _OnIntersectionEnterTrigger = 8;
-        private static _OnIntersectionExitTrigger = 9;
-        private static _OnKeyDownTrigger = 10;
-        private static _OnKeyUpTrigger = 11;
-        private static _OnPickUpTrigger = 12;
+        private static _OnPickDownTrigger = 5;
+        private static _OnPickUpTrigger = 6;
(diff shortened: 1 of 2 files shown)


src/Actions/babylon.actionManager.ts
```diff
@@ -57,14 +57,16 @@
...
+        private static _OnPickDownTrigger = 5;
+        private static _OnPickUpTrigger = 6;
+        private static _OnLongPressTrigger = 7;
+        private static _OnPointerOverTrigger = 8;
+        private static _OnPointerOutTrigger = 9;
+        private static _OnEveryFrameTrigger = 10;
+        private static _OnIntersectionEnterTrigger = 11;
+        private static _OnIntersectionExitTrigger = 12;
+        private static _OnKeyDownTrigger = 13;
+        private static _OnKeyUpTrigger = 14;
```
deltakosh: For compat reason,please keep the original values unchanged

vousk: I changed those to simplify reading and optimise code (only a very little bit, I know) in hasPickTriggers and hasPointerTriggers methods. Those private statics are not meant to be modified by user and I've checked the whole src for internal impact of changing the values and there should be no problem, that's why I decided to reorder them. 
If you still prefer original values and new one added incrementaly, I'll do that, that will imply just some additionnal test cases in the hasXXXTriggers methods.

vousk: Knowing that I double check the internal impact of changing the order (it's ok) and those private static are not meant to be changed by final user, what do you prefer : slightly optimised current version or previous order with new triggers roughly added at the end ?

deltakosh: No problem on my side (I checked the actions Builder and this is fine)

PR #899, opened 2016-01-07
fix cap start error on arc

jbousquie:
fixed a math error on cap start when some arc value was given to the lathe


deltakosh:
There is a conflict :(

jbousquie:
I don't get what ...
I'll check tomorrow or I'll redo the PR from scratch

PR #920, opened 2016-01-20
Materials Library: improved fur to an optional high level

julien-moreau:
Materials Library: improved fur to an optional high level
Updated documentation


deltakosh:
Agree about the doc :)

julien-moreau:
Still don't know why it is not a reflex in my mind.
Can you tell me when you added the FurMaterial in the Playground? Thanks ! :)

PR #925, opened 2016-01-26
audio negative startOffset bug 2nd

wyklion:
1. play(10)
3. play() error...

i made a music game demo:
http://42.62.1.80/MusicGame/MusicGame/?id=13
id can be 1 ~ 13.

so believe me, it's really a bug. have a try.


RaananW:
I'll wait for davrous to validate it. would it be possible in the meantime to rebase all of those commits to one?

wyklion:
i tried and failed...
i'll try again...

PR #927, opened 2016-01-29
Update babylon.sound.ts

wyklion:
1. play(10)
3. play() error...
i made a music game demo:
http://42.62.1.80/MusicGame/MusicGame/?id=13
id can be 1 ~ 13.
so believe me, it's really a bug. have a try.

rebase failed, i delete and fork new one to modify.


davrous:
Ok makes sense now. I haven't imagined this scenario. I haven't reproduced the problem using your link but by creating a small repro code. Can you please use our playground next time? Thanks!

PR #929, opened 2016-01-30
Pointer down up

j-o-d-o:
I added a OnPointerUpDown click event to the scene. It saves the Name of the Mesh the user clicked on when the OnPointerDown event is raised. And checks if the Name of the Mesh is still the same on the OnPointerUp event. If yes => fire OnPointerDownUp.

so you can do something like this:
`     
scene.onPointerDownUp = function(evt, pickResult){
	pickResult.pickedMesh.material.diffuseColor = new BABYLON.Color3(0.3, 0.9, 0.3);
	pickResult.pickedMesh.active = true;
}
`

In a scene where the user can move around with the camera and interact with diffrent mesh by clicking on it, it helps that the user doesnt accently raise a click event on a mesh when he actually wants to just rotate the camera. 

Something General:
As you can see at my github account, I am not expirienced with github (or version controll over all). I hope thats how its done.
- fork babylon.js to my github account
- create a branch in of the forked repo where I can add the new feature or do some bug fixing
- when done I test it
- if it works i create a pull request

I only changed the files typescript and javascript file in the source (I created the javascript file with gulp typescript-compile). Is that enough or do I have to also change the files in the dist or the definitely typed file?

### DIFF
--- a/src/babylon.scene.js
+++ b/src/babylon.scene.js
@@ -137,6 +137,7 @@ var BABYLON;
             this._transformMatrix = BABYLON.Matrix.Zero();
             this._edgesRenderers = new BABYLON.SmartArray(16);
             this._uniqueIdCounter = 0;
+            this._pickedMeshName = null;
             this._engine = engine;
             engine.scenes.push(this);
             this._renderingManager = new BABYLON.RenderingManager(this);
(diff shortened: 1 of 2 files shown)


deltakosh:
Hello!
the push is perfect :)
My concern is more about the need for this event. Why onPonterUp is not enough?

deltakosh:
Ok gotcha. I'm just stupid. Excellent PR. Thank you

deltakosh:
Can you please just update the doc and write a small article around up/down et how they work regarding DownUp ?

j-o-d-o:
Cool. Will do tomorrow!

RaananW:
Hi, 
I slap find this to be very helpful! Just thinking about the name. Wouldn't onPointerPick be better? This down up can be confusing.

j-o-d-o:
Yep, you are right, this makes it more clear what the event actually does. I always struggle naming things ;). Should I open a new pull request with the changed name?

PR #939, opened 2016-02-02
fix : returned shapeId by addShape()

jbousquie:
the returned shapeId by addShape() was wrong.

### DIFF
--- a/src/Particles/babylon.solidParticleSystem.ts
+++ b/src/Particles/babylon.solidParticleSystem.ts
@@ -286,7 +286,7 @@ module BABYLON {
             }
             this.nbParticles += nb;
             this._shapeCounter++;
-            return this._shapeCounter;
+            return this._shapeCounter - 1;
         }
 
         // rebuilds a particle back to its just built status : if needed, recomputes the custom positions and vertices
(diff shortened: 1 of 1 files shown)


src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -286,7 +286,7 @@ module BABYLON {
             }
             this.nbParticles += nb;
             this._shapeCounter++;
-            return this._shapeCounter;
+            return this._shapeCounter - 1;
```
RaananW: `return this,_shapeCounter++`? :smile:

jbousquie:
No, we must store shapeCounter incremented, but we must return the value before incrementation

RaananW:
that's exactly what this line will do (put a return in line 288, remove the rest). return the original value, and increment right after. But it is not too readable. The way you did it is clearer.

jbousquie:
ok, I thought it would increment first, then return the computed value

Temechon:
@jbousquie : 
- increment first, then return : ++this._shapeCounter
- return, then increment: this._shapeCounter++

PR #945, opened 2016-02-05
SPS : Added the public method digest()

jbousquie:
New SPS feature : `digest(mesh, {facetNb?: int, number?: int})`
This populates the SPS with _number_ particles (or _total_mesh_facets / facetNb_ ).
These particles have then the same same geometry and position than the mesh original facets.


deltakosh:
Don't forget to update the what's new :)

jbousquie:
Damned, you're right
done.

2016-02-05 19:28 GMT+01:00 David Catuhe <notifications@github.com>:

PR #950, opened 2016-02-08
Updating gulp

RaananW:
Now using an npm plugin to convert files to inline strings.


deltakosh:
Cool! any difference in the generated code?

RaananW:
Ah :) I wanted to see if the build is successful :)
no change whatsoever, simply less files in the gulp directory, and one npm package published.

PR #953, opened 2016-02-09
Fix bug when disposing the water material 

Temechon:
(Cannot read property '_framebuffer' of undefined), because customrendertarget were not deleted from the array.
It's relative to [this post](http://www.html5gamedevs.com/topic/20408-how-to-dispose-watermaterial-with-renderlist/)


materialsLibrary/materials/water/babylon.waterMaterial.ts
```diff
@@ -695,6 +695,17 @@ module BABYLON {
             if (this.bumpTexture) {
                 this.bumpTexture.dispose();
             }
+
+            var index = this.getScene().customRenderTargets.indexOf(this._refractionRTT);
+            if (index != -1){
+                this.getScene().customRenderTargets.splice(index, 1);
+            }
+            index = -1;
+            index = this.getScene().customRenderTargets.indexOf(this._reflectionRTT);
```
RaananW: why running it twice?

julien-moreau: Oh, I forgot this !
Twice because there are 2 render target textures: reflection and refraction

julien-moreau: Once the commit is merged, I'll add the renderlist to the serialize and parse methods

Temechon: Yes twice because of refraction and reflection.

RaananW: woops, those words are so close to one another

PR #959, opened 2016-02-11
Scene and Particle Systems: support of animations (serialize and parse)

julien-moreau:
- Serialize and Parse scene animations
- Serialize and Parse particle systems animations
- Added static function "ClearCache" in SceneSerializer, in case you call SerializeMesh or SerializeScene multiple times

### DIFF
--- a/src/Tools/babylon.sceneSerializer.ts
+++ b/src/Tools/babylon.sceneSerializer.ts
@@ -207,6 +207,10 @@
     }
 
     export class SceneSerializer {
+        public static ClearCache(): void {
+            serializedGeometries = [];
+        }
+
         public static Serialize(scene: Scene): any {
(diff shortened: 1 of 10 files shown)


src/Tools/babylon.sceneSerializer.ts
```diff
@@ -255,6 +259,14 @@
...
             }
 
+            // Animations
+            serializationObject.animations = [];
+            var animation: Animation;
+            for (index = 0; index < scene.animations.length; index++) {
+                animation = scene.animations[index];
+                serializationObject.animations.push(animation.serialize());
+            }
+
```
deltakosh: Why implementing IAnimatable If you don';t use AppendSerializedAnimations

PR #969, opened 2016-02-16
Fix.ground.get height at coordinates.optimization

jbousquie:
Sorry, no idea why all these files came with in this PR ...

This fix does two things in the Ground class 
* the private method `_computeHeightQuads()` now never reallocates any memory since the _heightQuads_ array is now created and populated once for all by `_initHeightQuads()`
* a new public method is provided : `updateCoordinateHeights()`. The user now can force, on demand,  the _heightQuads_ array recomputation if, for instance, he morphes his ground within the render loop. (PG forthcoming)

In short, this gives to the ground object now the possibility to do natively such a thing (was done with a ribbon and a dedicated function) at better performance than the example : http://www.babylonjs-playground.com/#1QC4YQ#19


RaananW:
Hi, 
way to many files that shouldn't be here. Can you clean it up a bit? Maybe in a branch from the current master?

PR #983, opened 2016-02-22
Physicsengine

RaananW:
The new physics engine architecture!
Is backwards compatible, so there is no need to change existing code.
Some functions are deprecated.
Documentation is coming soon.


jbousquie:
WWAAAAAOOOOWWwwww

2016-02-22 21:26 GMT+01:00 Raanan Weber <notifications@github.com>:

Tools/Gulp/config.json
```diff
@@ -83,6 +83,8 @@
       "../../src/PostProcess/babylon.postProcess.js",
       "../../src/PostProcess/babylon.postProcessManager.js",
       "../../src/PostProcess/babylon.passPostProcess.js",
+      "../../src/Physics/babylon.physicsJoint.js",
+      "../../src/Physics/babylon.physicsImpostor.js",
```
deltakosh: don't forget to update the file list on the documentation :)

dad72:
Hi Raanan,
Why Mesh.prototype.**getChildren** has been deleted ? another function replaces it? I use this function to reach all children of objects.

RaananW:
it is not deleted, just moved to an upper class. In this commit it is in AbstractMesh, and in a future commit it will be in the Node class, but it should be transparent to you, as the mesh and abstract mesh are both extentions of the Node class.

jbousquie:
maybe is this the answer ?
http://www.html5gamedevs.com/topic/20650-questions-about-meshgetchildren/?do=findComment&comment=117827

dad72:
Ok, I did not have all followed. thanks guys

PR #988, opened 2016-02-23
First commit of the NormalHeightMap command line tool.

nockawa:
With it you can:
 - pack two separated normal and height map files into a single one
 - invert the R &| G components of a normal map file to change the normal orientation.


jbousquie:
.exe ???

there are Mac and Linux users here also ...

deltakosh:
is there a way to execute it on Linux/Mac using .Net Core?

nockawa:
The tool uses this library http://imageprocessor.org/ which target only .net40 and .net45.
So I'm not sure for .net core... It would require way more work than I already did to achieve this I think...
Can you elaborate why you would like ideally?

jbousquie:
What I meant is more :
We do things for the web, we use OS agnostic languages like JS or TS.

So why not keep on providing tools in this ecosystem (gulp, node, js)
rather than going to .NET, JVM or any other languages ?

Just a question ...

2016-02-23 17:51 GMT+01:00 nockawa <notifications@github.com>:

nockawa:
Well, I'm pretty new here, so when I looked at the Tools directory I saw there were already few .net projects and I figured I could contribute with the small one I just wrote yesterday and it was ok.

Now the real question behind this (imho) is: do we have to be platform agnostic for the production pipeline?

I think this question is legitimate and depending on the answer some guidelines could be setup to give the devs clear directions of how things are supposed to be made.

@jbousquie : this tool should have been a node.js  module, ideally for you? Considering it's supposed to be used by the graphists guys.

deltakosh:
Interesting debate :) We should as much as possible be platform agnostic but it also depends on everyone competencies and timing.

RaananW:
my question would be (even after merging :) ) if it doesn't belong in the extensions repository

deltakosh:
Good question. We have tools about .Babylon file for instance here. Perhaps you're right actually
@nockawa : do you mind moving your tool to /extensions here:https://github.com/BabylonJS/Extensions?

nockawa:
@deltakosh not at all... :)
I'll do it asap, won't be long if Git is friendly with me...so it may takes some time...

nockawa:
Ok, I've made the PR on the Extensions project. Who should take care of deleting the tool from this repo?

RaananW:
:+1: 
I'm on both

nockawa:
Thank you !

nockawa:
Thank you !

PR #989, opened 2016-02-23
Travis - gulp fail on TS errors

RaananW:
When building on travis - If TypeScript fails for some reason the PR will be marked as failed.
Until now errors on gulp were ignored.


RaananW:
And this PR is a wonderful example for that. There is some kind of a problem with the geometry class.

RaananW:
Fixing the issues. The errors are due to a new TS version (1.8.2). working on a solution.

PR #991, opened 2016-02-23
Added Parallax Mapping in StandardMaterial

nockawa:
Two modes: Parallax Mapping and Parallax Occlusion Mapping.
This feature lies on the top of the Bump's one, requires a Normal Map with the Alpha Channel contain Heightmap information.

A little demo of the effect: https://youtu.be/SEFCZpSjWF8

### DIFF
--- a/src/Materials/babylon.standardMaterial.ts
+++ b/src/Materials/babylon.standardMaterial.ts
@@ -792,7 +792,7 @@
                     if (this.bumpTexture && scene.getEngine().getCaps().standardDerivatives && StandardMaterial.BumpTextureEnabled) {
                         this._effect.setTexture("bumpSampler", this.bumpTexture);
 
-                        this._effect.setFloat2("vBumpInfos", this.bumpTexture.coordinatesIndex, 1.0 / this.bumpTexture.level);
+                        this._effect.setFloat3("vBumpInfos", this.bumpTexture.coordinatesIndex, 1.0 / this.bumpTexture.level, this.parallaxScaleBias);
                         this._effect.setMatrix("bumpMatrix", this.bumpTexture.getTextureMatrix());
 
                         if (this.useParallax) {
--- a/src/Materials/babylon.standardMaterial.ts
+++ b/src/Materials/babylon.standardMaterial.ts
@@ -8,6 +8,8 @@
         public EMISSIVE = false;
         public SPECULAR = false;
         public BUMP = false;
+        public PARALLAX = false;
+        public PARALLAXOCCLUSION = false;
         public SPECULAROVERALPHA = false;
         public CLIPPLANE = false;
         public ALPHATEST = false;
@@ -153,6 +155,15 @@
 
         @serialize()
         public disableLighting = false;
+
+        @serialize()
+        public useParallax = false;
+
+        @serialize()
+        public useParallaxOcclusion = false;
+
+        @serialize()
+        public parallaxScaleBias = 0.05;
 
         @serialize()
         public roughness = 0;
@@ -398,6 +409,13 @@
                     } else {
                         needUVs = true;
                         this._defines.BUMP = true;
+
+                        if (this.useParallax) {
+                            this._defines.PARALLAX = true;
+                            if (this.useParallaxOcclusion) {
+                                this._defines.PARALLAXOCCLUSION = true;
+                            }
+                        }
                     }
                 }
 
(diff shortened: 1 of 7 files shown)


sebavan:
Hello,

This look nice but it probably breaks other materials using the bump fragment due to the signature change. Could you look into both src/materials and the material library if other materials require fixing?

Also, comments in shader are not removed during minification and could quickly increase the size of the package. Could you reduce the size of the comment in the shader? (agree we should try to work it out).

Thank you,

nockawa:
Oops sorry, I wasn't aware that other shaders could rely on the perturbNormal function.
I just did a search and that's the case only for pbr.fragment.fx in the shader library.
Do you want me to fix it?

Ok for the comments, I was wondering why it was so commentless (but well...it's not that unusual anyway)

sebavan:
Yep, that would be nice if you can fix it.

Agree fro the comments, would be nice to had something during minification.

nockawa:
Btw, it would really be nice if comments could be stripped away during minification...
This should be not too complicate to implement at least for the // comment form. We don't use the /* */ a lot anyway...

sebavan:
Yep, feel free to create a PR on the gulp processes for shader processing to strip out comments (both tools/gulp and material library) :-)

nockawa:
I said "gulp specialist", and I'm quite the opposite of that statement! :)

src/Materials/babylon.standardMaterial.ts
```diff
@@ -622,7 +648,7 @@
                         "vLightData2", "vLightDiffuse2", "vLightSpecular2", "vLightDirection2", "vLightGround2", "lightMatrix2",
                         "vLightData3", "vLightDiffuse3", "vLightSpecular3", "vLightDirection3", "vLightGround3", "lightMatrix3",
                         "vFogInfos", "vFogColor", "pointSize",
-                        "vDiffuseInfos", "vAmbientInfos", "vOpacityInfos", "vReflectionInfos", "vEmissiveInfos", "vSpecularInfos", "vBumpInfos", "vLightmapInfos", "vRefractionInfos",
+                        "vDiffuseInfos", "vAmbientInfos", "vOpacityInfos", "vReflectionInfos", "vEmissiveInfos", "vSpecularInfos", "vBumpInfos", "vParallaxScaleBias", "vLightmapInfos", "vRefractionInfos",
```
deltakosh: You can use a component of vBumpInfos I guess to save an uniform
(And do not forget to credit yourself in the 2.4 what's new :))

nockawa: ok, I make the change

nockawa:
@deltakosh @sebavan did you guys made recent chandes in the lighting?
Because with bump on (no parallax), the back faces are now sometime lit, looks like a neg z vector check is not missing.
Check this img

deltakosh:
I don't think we change something recently

sebavan:
Not on the standard material, I do not think so.

nockawa:
Men, tell me it's your fault and not mine... pleaaaase.. :'(
I had weird things during the coding of the shader, but not this one, I guess it's a merge issue...
Ok, gotta look at everything...again.

deltakosh:
Lol but think about the glory when it will be done:)

nockawa:
Ok, I'm pretty sure it's not due to what I did. For me the lighting does not clamp on neg vector, looks like it's only on the specular (but well, I can't tell because it appears very strong in my test).
I commit the changes you guys requested and tomorrow I'll take a look at the lighting.
It's not something that can be spotted when you don't use heavy normal mapping, so maybe it's there since a while... dunno. but I didn't touched a damn thing on the normal mapping formula...

deltakosh:
Did you see it working at some point in time? this can come from the rework I did for includes

src/Materials/babylon.standardMaterial.ts
```diff
@@ -766,8 +792,12 @@
                     if (this.bumpTexture && scene.getEngine().getCaps().standardDerivatives && StandardMaterial.BumpTextureEnabled) {
                         this._effect.setTexture("bumpSampler", this.bumpTexture);
 
-                        this._effect.setFloat2("vBumpInfos", this.bumpTexture.coordinatesIndex, 1.0 / this.bumpTexture.level);
+                        this._effect.setFloat3("vBumpInfos", this.bumpTexture.coordinatesIndex, 1.0 / this.bumpTexture.level, this.parallaxScaleBias);
                         this._effect.setMatrix("bumpMatrix", this.bumpTexture.getTextureMatrix());
+
+                        if (this.useParallax) {
+                            this._effect.setFloat("vParallaxScaleBias", this.parallaxScaleBias);
```
deltakosh: Why keeping this then? :)

nockawa: it's a fucking mess!!! all the commits I didn't want end up in the master and the only one I wanted is not!!
The code you see is what were before I made the change!
My dev branch is ok, after I tried to merge just one commit to the master and now it's a mess!

nockawa:
No, it was not working before the include refactoring, I just checked for this. I though the issue appeared after the merge but in fact it were already there before. I'll take a look at the lighting tomorrow. If I can get out the git mess..

deltakosh:
Do not forget to update the readme :)

PR #995, opened 2016-02-24
Stripping comments and unneeded chars from shaders

RaananW:


RaananW:
wait with this one, still need to test fully

PR #1000, opened 2016-02-26
Physics Bug Fixing

RaananW:
Delta rotation should be conjugated, otherwise the body will keep on
constantly flipping (if deltaRotation is set).
cannonJS collision of joined objects improved.
Added Conjugate function to the quaternion class.

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -1592,6 +1592,23 @@
 
             return this;
         }
+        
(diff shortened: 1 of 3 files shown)


nockawa:
#1000 milestone reached! yey champagne!

deltakosh:
That's a lot!

RaananW:
1000 ftw!
Should have commited more than a bug fix :smile:

rsarieddine:
That's pretty awesome!

PR #1008, opened 2016-03-02
Skeletal Animation Blending...   smooth transitions

lemontmoon:
simply play skeletal animation as always for character.......
scene.beginAnimation(target,from,to,loop,speedRatio,onAnimationEnd(),animatable, transitionSpeed)

transition speed is from range 0 to 1....... 
by default is 1 -- > no transitions
try for example 0.1 or less to see smooth transition blending between two poses


RaananW:
Hi,
Thanks for the commit!
It seems like there are conflicts on some of the files. I also see that you committed many files that are not related to your commit.
Could you clean up the code, check for conflicts, and commit the changes only? This will help us validate your PR. 

Thanks!

lemontmoon:
YEs,,, this was my first commit,,,, and first day Using VScode .

I dont know how to commit only files that has only changed,,, because I am using CTRL + SHIF +B 
 for building,,
maybe I am doing something wrong,,,,,  if U know what I am doing wrong,,, please let me to know
On my side there are no Warnings errors while compiling

RaananW:
Hi!

So, I think the main problem are all of the unneeded files you added to the commit.

The simplest solution is a kind of a "restart". Clone the master again, make changes and commit those (before using gulp, which is what made all of those extra files appear as "changed").
After you commit the changed files, run gulp and see that it compiles. If it does, create a PR (without commiting all of the files again, as the changes are already there).
This will create a clean PR with only the files that you actually changed.
Make sure you have the newest master branch before starting, this will also avoid the conflicts, if any.

If you need any help, PM me on the forum!

Let's make it happen :)

I am closing this PR for now. Waiting for your new PR :smile:

lemontmoon:
I have commited new PR just 5 minutes ago,,,now its ok,,

PR #1009, opened 2016-03-02
ANIMATION transition BLENDING("CROSSFADES") 

lemontmoon:
dynamic crossfade between…… two animations using function   linearFADE in math.ts

I am new to GIT,, and CONTRIBUTION,,,, som please provide me some info what files i have really to commit..... and gimme some rules 

also neeed to know what is it means if ("All checks have failed")


RaananW:
Hi,

Wonderful, this looks much better.
The "all checks failed" means that Travis, our CI-Tool, has failed building your pull-request.

The reason is probably the Babylon.js file you have commited, which adds a submodule to the project. This file should certainly not be a part of the commit :)
Could you please remove it and commit the change? Travis will try building again and will probably have no problems running the gulp task.

Babylon.js
```diff
@@ -0,0 +1 @@
+Subproject commit 8aaa16d76eac73e2e7ec535ae6319c119e5bcb5d
```
deltakosh: What is this? :)

RaananW: this needs to be removed :)

src/Math/babylon.math.ts
```diff
@@ -3731,6 +3731,51 @@
             Matrix.Zero(), Matrix.Zero(),
             Matrix.Zero(), Matrix.Zero()];                      // 6 temp Matrices at once should be enough
     }
+    
+    export class Lerp {
```
deltakosh: I'm not sure we need a class for this :)

src/babylon.scene.ts
```diff
@@ -830,29 +830,46 @@
...
                 var animatables = target.getAnimatables();
                 for (var index = 0; index < animatables.length; index++) {
-                    this.beginAnimation(animatables[index], from, to, loop, speedRatio, onAnimationEnd, animatable);
+                    this.beginAnimation(animatables[index], from, to, loop, speedRatio, onAnimationEnd, animatable,transitionSpeed);
                 }
             }
 
             animatable.reset();
 
+              var maxAnimations = 2;
```
deltakosh: Why 2?

lemontmoon:
problem is that I didnt commited babylon.js,,

looking for whats wrong

src/babylon.scene.ts
```diff
@@ -875,7 +892,43 @@
...
+                    AT.push(this._activeAnimatables[index]);
+                    
+                }
+            }
+            if(AT.length) { AT.reverse() };
+            return AT;
+           
+        };
+        //////////we need this for now to get last current running animatable on target
+        public GetCurrentAnimatableByTarget(target: any): Animatable {
```
deltakosh: This function seems unused?

src/babylon.scene.ts
```diff
@@ -830,29 +830,46 @@
...
+     
+          //  this.stopAnimation(target);
+            var transitionFunction = null;
+            if(transitionSpeed < 1)  {
+                                       transitionFunction = new BABYLON.LinearFADE(transitionSpeed,this._engine);
+                                     }
+              
             if (!animatable) {
-                animatable = new Animatable(this, target, from, to, loop, speedRatio, onAnimationEnd);
+                animatable = new Animatable(this, target, from, to, loop, speedRatio, onAnimationEnd,animatable,transitionFunction);
```
deltakosh: the animatable parameter (8# parameter) seems really suspicious as animatable must be null to enter this condition

src/Math/babylon.math.ts
```diff
@@ -3731,6 +3731,51 @@
             Matrix.Zero(), Matrix.Zero(),
             Matrix.Zero(), Matrix.Zero()];                      // 6 temp Matrices at once should be enough
     }
+    
+    export class Lerp {
+        constructor(from: number, to: number, factor: number) {
```
deltakosh: No class needed here. Just returning a function will be fine

deltakosh:
Just a global question. Why don't you use regular animations blending? It works well with matrices as well

src/babylon.scene.ts
lemontmoon: for now we want to have max
t 2 animatable per target,,,, i can remove this variable if u want

src/Animations/babylon.animation.ts
```diff
@@ -314,6 +315,12 @@
                             switch (loopMode) {
                                 case Animation.ANIMATIONLOOPMODE_CYCLE:
                                 case Animation.ANIMATIONLOOPMODE_CONSTANT:
+                                   if(this.transitionFunction)
+                                    {
+                                        var transitionWeight = this.transitionFunction.fadeIn();
+                                        return this.matrixInterpolateFunction(this._target._matrix, startValue, transitionWeight); ////ease from previousAnimation to CurrentFrame with fraction
```
deltakosh: I think that this is not required (I may be wrong) but this will be automatically done if you use animatable.enableBlending(0.1).

Does it make sense?

src/babylon.scene.ts
lemontmoon: yes it is suspicous,,,, but it was there before I started coding this,,,, so can I remove it?

src/Animations/babylon.animation.ts
deltakosh: See here:
https://github.com/lemontmoon/Babylon.js/blob/master/src/Animations/babylon.animation.ts#L359

src/babylon.scene.ts
deltakosh: He was not as you can see on the diff:
Before=> animatable = new Animatable(this, target, from, to, loop, speedRatio, onAnimationEnd);
After=>animatable = new Animatable(this, target, from, to, loop, speedRatio, onAnimationEnd,**animatable**,transitionFunction);

PR #1017, opened 2016-03-04
Removing Tools from Math completely 

RaananW:
Math has now no more Tools references.
We should consider throwing errors instead of Tools.Error (but then exceptions will have to be handeled)


RaananW:
Found some other references, wait with that for a minute.

deltakosh:
oups!

RaananW:
it's ok,
nothing is broken.

RaananW:
Will create another PR after making the final changes.

PR #1018, opened 2016-03-05
Added an optional predicate on Node.getDescendants, Node.getChildren to filter out Node based on a callback execution.

nockawa:
Ok guys, I'm still not "comfortable" with Pull Request, I really don't understand why I can't Pull Request only ONE commit which is the one of the dev I'm proposing.
On my side the PR lists 9 commits, the last being the only interesting one.
Please, do tell me if I would have been able to do better or if it's simply the way things work...

On a code related matter:
 - I also added the predicated on Node.getChildren even though it's deprecated, I figure it can't harm and many people still rely in this one.
 - Looks like to me that Node._getDescendants should be private instead of public, but I didn't touch anything, it's not my call...


deltakosh:
There are too many files not related to this commit (vs files, assets, .sln..)

nockawa:
I know ! Git is a f**kin nightmare to me! 
Still can't figure out a clear process to not push my own evol that I don't want to share... ;'(
And it's not a lack of trying...

deltakosh:
I share your pain...First you should not add all vs files. To do so just update your .gitignore file

nockawa:
Well, no... All these files I wanted them to be sourced control, but some of them I wanted them only to be in my own dev branch (.sln, .csproj, assets), WITHOUT merging them back to master.

But each time I make a contribution, I want to do it in my Dev branch (because it's where I can dev with the .sln/.csproj), then I commit only the contribution.

It's the merge phase from my dev to my master that I mess up every single time! Instead of cherry picking and merge the only commit you care about, everything missing from the dev in the master is merged...Then the PR looks like this!

nockawa:
So you guys can confirm that now you only see 2 files changed instead of a bunch of crap?
The dev I wanted to PR is no longer here, because I revert everything.
But the question is: does it looks like a "clean" situation now?

RaananW:
Are these the right files? I see a zip file and a fragment shader

nockawa:
Nono, these are not, but just before I revert the changes there were 40 files...
So it's in better shape, but still not the expected files

nockawa:
By the way, I don't understand while these two files are different, I'm supposed to be on synced with the bjs master branch, but if these two files are showing up in the PR it means they are different, right ?
I just don't understand why...

nockawa:
YEY here we are! 14 commits to get two files modified!
I did it...not proud though!

src/babylon.node.ts
```diff
@@ -176,10 +176,17 @@
...
+         * @param {BABYLON.Node[]} list the input array of nodes to evaluate
+         * @param {BABYLON.Node[]} results the result array containing the nodes matching the given criterias
+         * @param {boolean} directDecendantsOnly if true only direct descendants of 'this' will be considered, if false direct and also indirect (children of children, an so on in a recursive manner) descendants of 'this' will be considered.
+         * @param predicate: an optional predicate that will be called on every evaluated children, the predicate must return true for a given child to be part of the result, otherwise it will be ignored.
+         */
+        public _getDescendants(list: Node[], results: Node[], directDecendantsOnly: boolean = false, predicate?: (node: Node) => boolean): void {
             for (var index = 0; index < list.length; index++) {
                 var item = list[index];
-                if ((directDecendantsOnly && item.parent === this) || (!directDecendantsOnly && item.isDescendantOf(this))) {
+                if (((directDecendantsOnly && item.parent === this) || (!directDecendantsOnly && item.isDescendantOf(this))) && (predicate==null || predicate(item))) {
```
deltakosh: please use === instead of ==

deltakosh:
Nice! I can merge :)

src/babylon.node.ts
nockawa: what could be the difference against null to use == instead of === ?

deltakosh: Just for consistency

RaananW: Actually in this case !predicate would be better. === null will be false for undefined.

deltakosh: Correct if undefined if a potential value

RaananW: fixed here - https://github.com/BabylonJS/Babylon.js/pull/1023

PR #1023, opened 2016-03-07
Update babylon.node.ts

RaananW:
Fixing the === null reference


RaananW:
http://www.html5gamedevs.com/topic/21008-getchildren-broken-in-babylon-playground/
This is caused by the "=== null"

nockawa:
===null was returning false when undefined was passed, which caused the bug?

RaananW:
yes, since the variable was not null but undefined. So !predicate (or (predicate === void 0)) would be the right way to go.

nockawa:
good to know thank  you. the undefined concept is really interesting, but I'm still not used to it.

PR #1024, opened 2016-03-07
Assorted toString() methods

Palmer-JC:
Will commit Loading classes which use these when ready.
Some of these can be more filled in, but just getting
starter versions.


src/Lights/babylon.spotLight.ts
```diff
@@ -24,6 +24,7 @@
             this.direction = direction;
             this.angle = angle;
             this.exponent = exponent;
+            this._type = 2;
```
deltakosh: There is already the getTypeID function

Palmer-JC:
Ok, will go back to pre-commit yesterday file system.  I was looking for a member property not a method with hardcoded return

PR #1025, opened 2016-03-08
Grid Material

sebavan:


materialsLibrary/materials/grid/babylon.gridmaterial.ts
```diff
@@ -0,0 +1,194 @@
...
+        public dispose(forceDisposeEffect?: boolean): void {
+            super.dispose(forceDisposeEffect);
+        }
+        
+        public clone(name: string): GridMaterial {
+            return SerializationHelper.Clone(() => new GridMaterial(name, this.getScene()), this);
+        }
+
+        public serialize(): any {
+            return SerializationHelper.Serialize(this);
```
deltakosh: you need to export class name:https://github.com/BabylonJS/Babylon.js/blob/master/materialsLibrary/materials/normal/babylon.normalMaterial.ts#L352

PR #1033, opened 2016-03-09
Update babylon.node.ts

RaananW:
Revising the getChildren (again?), for some reason this was rewritten.


RaananW:
I'll document those functions later this evening

PR #1034, opened 2016-03-09
API doc : some static functions inline doc added

jbousquie:
Will be continued some other day in some other branch


deltakosh:
Woot!

jbousquie:
f%*$@!ng boring...

and it's not complete, pffffff

PR #1038, opened 2016-03-11
API doc : some more function doc added

jbousquie:
... painful
my day contribution to the Cause


deltakosh:
And the Cause thanks you :)

PR #1041, opened 2016-03-13
Update babylon.scene.ts

dad72:
One can choose the skeleton where you want to search the bone by name or by id

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -1146,35 +1146,36 @@
                     return this.cameras[index];
                 }
             }
-
             return null;
         }
 
         /**
          * get a bone using its id
          * @param {string} the bone's id
-         * @param {Skeleton} the skeleton
          * @return {BABYLON.Bone|null} the bone or null if not found
          */
-        public getBoneByID(id: string, skeleton: Skeleton): Bone {
-        	for (var boneIndex = 0, cache = skeleton.bones.length; boneIndex < cache; boneIndex++) {
+        public getBoneByID(id: string): Bone {
+            for (var skeletonIndex = 0; skeletonIndex < this.skeletons.length; skeletonIndex++) {
+                var skeleton = this.skeletons[skeletonIndex];
+                for (var boneIndex = 0; boneIndex < skeleton.bones.length; boneIndex++) {
                     if (skeleton.bones[boneIndex].id === id) {
                         return skeleton.bones[boneIndex];
                     }
                 }
-                return null;
+            }
+            return null;
         }
 
         /**
         * get a bone using its id
         * @param {string} the bone's name
         * @param {Skeleton} the skeleton 
-        * @return {BABYLON.Bone|null} the bone or null if not found
+        * @return {number} the indice of bone on
         */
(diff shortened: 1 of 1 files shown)


RaananW:
Hi David,

I am trying to understand what the difference is between the implemented method and the one you PRed here.
There are also Travis errors to fix (no need for a break after the return method), but would be great to know first what you actually did :smile:

dad72:
I add the ability to choose the skeleton to research of a bone.
before, it was looking at all the available skeleton on scene.
Now we can choose a skeleton among those on scene

dad72:
For the 2 first error, I corrected.

I think to find a bone by name or id, several character on stage for example has the same bone name. Seeking a bone of a skeleton, you get the skeleton that we want.

src/babylon.scene.ts
```diff
@@ -1153,37 +1153,31 @@
...
         * get a bone using its id
         * @param {string} the bone's name
+        * @param {Skeleton} the skeleton 
         * @return {BABYLON.Bone|null} the bone or null if not found
         */
-        public getBoneByName(name: string): Bone {
-            for (var skeletonIndex = 0; skeletonIndex < this.skeletons.length; skeletonIndex++) {
-                var skeleton = this.skeletons[skeletonIndex];
-                for (var boneIndex = 0; boneIndex < skeleton.bones.length; boneIndex++) {
+        public getBoneByName(name: string, skeleton: Skeleton): Bone {
```
RaananW: This is a breaking change. I can't say if this feature is needed or not (I guess it is needed if you have two bones with the same name in different skeletons), but the skeleton variable should be optional.

src/babylon.scene.ts
```diff
@@ -1426,10 +1424,13 @@
...
                 return camera;
             }
-
-            var bone = this.getBoneByName(name);
-
-            return bone;
+            var bone = null;
+            for (var skeletonIndex = 0, cache = this.skeletons.length; skeletonIndex < cache; skeletonIndex++) {
+                var skeleton = this.skeletons[skeletonIndex];
+		bone = this.getBoneByName(name, skeleton);
```
RaananW: If a bone was found return it, otherwise it will traverse the entire skeleton index for no reason (performance reduction)

dad72:
By imagining a scene you have more character. you can attach an object on the hand of one of the characters. seeking the bone by the name 'right hand', my method can only select the skeleton on which you want to attach the object and not on the first skeleton found on the scene.

So I think we should keep functions as before and create new ones with a different name.
Ok, I'll do that.

dad72:
I handed over the functions as before and have added another that better suits what I wanted to do.
This allows you to:

```
var hant_left = scene.searchIndexOnSkeletonByName(skeleton, "hant_left ");
sword.attachToBone(skeleton.bones[hant_left], character);
```

deltakosh:
This function should be on Skeleton class

dad72:
I thought that by putting it on scene. we needed a skeleton on scene so the bone.

But yes it could also be on skeleton. I do like that.

dad72:
I close this pull

PR #1046, opened 2016-03-15
Force Streaming Video on Load

wpdildine:


RaananW:
Hi,

thanks for the commit!
I can't seem to see the difference between the old implementation and this one, apart from the fact that the event is registered only when urls are given.
There is also a lot of repeated code, and the URLs are actually processed twice.
Care to share a bit more information on the need and uses of this?

wpdildine:
From my testing HLS and Mpeg-Dash requires a full play through before the canplaythrough event will trigger.  I would suggest removing that event or finding some other even trigger.  Otherwise, you could end up waiting 1 minute or longer for your textures to appear.

It could be rewritten to something like:

this.video.endsWith(".m3u8") || this.video.endsWith(".mpd") ? true : false;

deltakosh:
No need for testing on the extension. Can you please just factorize the repeated code!

wpdildine:
Sorry not pro at git  :/

deltakosh:
Sorry but I cannot valide a pr with 172 files:)

PR #1052, opened 2016-03-19
Physics updatable objects

RaananW:
It is now possible to assign a physics impostor to any object with a position and a quaternion. Including solid particles :)


jbousquie:
waaaaoowwww
I can't wait to see a PG with solid particles under real physics engine
management

2016-03-19 22:39 GMT+01:00 Raanan Weber <notifications@github.com>:

PR #1064, opened 2016-03-24
Ray.intersectSegment method added

nockawa:


src/Culling/babylon.ray.ts
```diff
@@ -192,6 +192,91 @@
...
+
+            // get the difference of the two closest points
+            var dP = w.add(u.multiplyByFloats(sc, sc, sc)).subtract(v.multiplyByFloats(tc, tc, tc));  // = S1(sc) - S2(tc)
+
+            var isIntersected = dP.lengthSquared() < (threshold * threshold);   // return intersection result
+
+            if (isIntersected) {
+                return tc;
+            }
+            return undefined;
```
RaananW: maybe returning -1 will be better?

nockawa:
The current commit has an incomplete version of the method, I don't check if the intersection is behind or further the ray.
For the return data, I think returning null in case of no intersection is good. I don't think we should use intersectionInfo considering the current state of the class.

nockawa:
Am i supposed to update the what's new for such small changes?

deltakosh:
yes please..all changes should be documented

nockawa:
I'll do it altogether the next pr then

PR #1069, opened 2016-03-27
Update babylon.meshSimplification.js

asafge:
Hey,
I noticed a small bug in the porting of this code from C++. It's mentioned and fixed in this blog post - http://voxels.blogspot.co.il/2014/05/quadric-mesh-simplification-with-source.html?_sm_au_=iVVSRHnRQ0SW0JFT.

In case a mesh is not two-manifold, it's possible for both vertices of an edge to be "borders", and this is an edge that should not be collapsed.
Testing for !== instead of || is ignoring this case and so naked edges are collapsed when they shouldn't.

### DIFF
--- a/src/Mesh/babylon.meshSimplification.js
+++ b/src/Mesh/babylon.meshSimplification.js
@@ -250,7 +250,7 @@ var BABYLON;
                                 var deleted1 = [];
                                 var v0 = t.vertices[j];
                                 var v1 = t.vertices[(j + 1) % 3];
-                                if (v0.isBorder !== v1.isBorder)
+                                if (v0.isBorder || v1.isBorder)
                                     continue;
                                 var p = BABYLON.Vector3.Zero();
                                 var n = BABYLON.Vector3.Zero();
--- a/src/Mesh/babylon.meshSimplification.ts
+++ b/src/Mesh/babylon.meshSimplification.ts
@@ -320,7 +320,7 @@
                                 var v0 = t.vertices[j];
                                 var v1 = t.vertices[(j + 1) % 3];
 
-                                if (v0.isBorder !== v1.isBorder) continue;
+                                if (v0.isBorder || v1.isBorder) continue;
 
                                 var p = Vector3.Zero();
                                 var n = Vector3.Zero();


RaananW:
Hi, 
Thanks for the commit. I remember changing it when I implemented this, I'm just not to sure why. Did you check performance changes? Or overall results?

asafge:
Hey,
Thanks for responding.

About overall results - I'm still testing this to make sure that it works as I expect, but here are the four possible scenarios, both before after the change:

Case | v0.isBorder | v1.isBorder | not eq. (before change) | or (after change) | Comments
------- | --------------- | --------------- | -------------------------------- | ---------------------- | --------
1       | false | false | false | false | No effect, since the condition's result is the same for both cases. Both vertices are not on a border, so this edge can be collapsed.
2       | false | true  | true   | true  | No effect, for the same reason. In this case one of the edge's vertices is on a border and the other one is not. Collapsing the edge might result in some deformation of the mesh in that area, so we shouldn't collapse it.
3       | true  | false | true   | true  | Same as previous case.
4       | true  | true  | false  | true  | **This is the case that we're fixing**. If both vertices are borders, this is border edge that defines appearance of the mesh in this area, and we don't want to collapse it, the same way we don't want to collapse edges in cases 2 & 3.

About performance, I don't think it makes any difference, since we're just changing a not-equals condition with an or condition, and further restricting the edges that are allowed to be collapsed. Anyway this is really only relevant to non-manifold meshes, which I'm assuming is least common scenario.

RaananW:
Hi Asaf,
I will test it when I get the chance (holiday season in europe, we are never home :smile: ). Will trust you if you give me the ok in regards to performance and the overall result.
Again, I really remember doing that, but I don't remember why I changed it. But what you are saying does make sense.

asafge:
Looks OK in my tests. Enjoy your holiday!

RaananW:
Hi Asaf, 
Looking good! I don't see any performance issue or any simplification errors.
Do you mind changing the typescript file? the .js files are only here for reference.

asafge:
OK just edited the ts file as well.

RaananW:
Toda!

PR #1072, opened 2016-03-30
Introduced a Scene.onPointerObservable property 

nockawa:
Introduced a Scene.onPointerObservable property that allow observers to listen to pointer input events as well as mouse wheel. You should rely on it when listening to PointerUp,Down,Move,Wheel,Pick events.

ArcRotate/FreeCamera are now relying on this Observable.
Observable.hasObservers() method added.
ArcRotate/FreeCamera are now manipulating beyond the rendering canvas (if the mouse cursor or touch goes outside of it).


nockawa:
lol, I didn't update the what's new to reflect this PR, i'll do it for the next contrib! :)

PR #1075, opened 2016-03-31
Mouse Wheel sanity check

RaananW:
This will check that the event sent to the mouse wheel input manager is a pointer wheel event.


deltakosh:
I'm merging but we won't need it soon:)

PR #1086, opened 2016-04-05
Few math stuffs additions/fixes

nockawa:
Fix Ray.intersectPlane method to return the correct distance.
Added Vector4.toVector3() method.
Added Matrix.getTranslation() and get/setRow() methods.

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -1455,6 +1455,10 @@
             return this;
         }
 
+        public toVector3(): Vector3 {
+            return new Vector3(this.x, this.y, this.z);
+        }
+
         public clone(): Vector4 {
             return new Vector4(this.x, this.y, this.z, this.w);
         }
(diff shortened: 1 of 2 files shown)


src/Math/babylon.math.ts
```diff
@@ -2227,6 +2235,27 @@
...
+            if (index < 0 || index > 3) {
+                return false;
+            }
+
+            var i = index * 4;
+            this.m[i + 0] = row.x;
+            this.m[i + 1] = row.y;
+            this.m[i + 2] = row.z;
+            this.m[i + 3] = row.w;
+        }
```
deltakosh: I suggest return this to keep a fluent API

nockawa: Well, I'm not a fan of fluent API for primitive like Matrix, the Invert() method is a prime example of 90% of the devs thinking it returns a new object but it doesn't.
Now for the setRow I think we can return this because the method doesn't change the whole Matrix object, so it may be easier to understand that it's 'this' that is returned.

PR #1088, opened 2016-04-07
Adding exception object to the list of callback params

RaananW:
Optional., not breaking.


dad72:
Raanan Hi, this is something else, but we can no longer post to the forum from 2 hours around. Is that too?
I post here because here we can not communicate on the forum, confirmation of that?

RaananW:
Something is wrong. This is missing an important line of code... I'll check and commit again.

deltakosh:
oh ok I though it was on purpose as a first step :) (only changing the signature first)

PR #1089, opened 2016-04-07
Sending the exception.

RaananW:
Was missing in the last commit for some reason.


RaananW:
Already committed by dk

PR #1105, opened 2016-04-23
StringDictionary + misc

nockawa:
- Added the StringDictionary<T> class
 - Removed the SmartCollection class and replace its usage by StringDictionary
 - Added Tools.getClassName() method to retrieve the name of a primitive type or a user defined class based instance.


deltakosh:
Hello I cannot merge due to conflicts (and some comments below :))

src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
```diff
@@ -70,7 +72,9 @@ module BABYLON {
                     //but emptying completly pointers collection is required to fix a bug on iPhone : 
                     //when changing orientation while pinching camera, one pointer stay pressed forever if we don't release all pointers  
                     //will be ok to put back pointers.remove(evt.pointerId); when iPhone bug corrected
-                    pointers.empty();
+                    pointers.clear();
+                    pointersOrder = new Array<number>();
```
deltakosh: We cannot afford having a new array built on each frame (or input). This will stress the GC. You should use a SmartArray

src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
```diff
@@ -135,7 +143,9 @@ module BABYLON {
 
             this._onLostFocus = () => {
                 //this._keys = [];
-                pointers.empty();
+                pointers.clear();
+                pointersOrder = new Array<number>();
```
deltakosh: idem.

src/Tools/babylon.stringDictionary.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        /**
+         * Get a value from its key or add it if it doesn't exist.
+         * This method will ensure you that a given key/data will be present in the dictionary.
+         * @param key the given key to get the matchin value from
+         * @param factory the factory that will create the value if the key is not present in the dictionary.
+         * The factory will only be invoked if there's no data for the given key.
+         * @return the value corresponding to the key.
+         */
+        public getOrAdd(key: string, factory: (key: string) => T): T {
```
deltakosh: Question: should we rename it to getOrAddWithFactory and create a getOrAdd(key, value) which seems more straightforward to me

src/Tools/babylon.tools.ts
```diff
@@ -698,6 +698,12 @@
             return false;
         }
 
+        public static getClassName(obj): string {
```
deltakosh: big warning here. We used to have this function. But it won't work with minified version where code is scrambled and where constructor and names can be changed

src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
```diff
@@ -70,7 +72,9 @@ module BABYLON {
                     //but emptying completly pointers collection is required to fix a bug on iPhone : 
                     //when changing orientation while pinching camera, one pointer stay pressed forever if we don't release all pointers  
                     //will be ok to put back pointers.remove(evt.pointerId); when iPhone bug corrected
-                    pointers.empty();
+                    pointers.clear();
```
deltakosh: I must be stupid but I do not see a clear function on StringDictionary

src/Tools/babylon.stringDictionary.ts
```diff
@@ -0,0 +1,125 @@
...
+                return true;
+            }
+            return false;
+        }
+
+        /**
+         * Clear the whole content of the dictionary
+         */
+        public clear() {
+            this._data = {}
```
deltakosh: This is a regression regarding SmartCollection. The main goal of SmartCollection was to avoid any GC. Here you create a new array on every clear.

src/Tools/babylon.stringDictionary.ts
nockawa: sure no problem

src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
nockawa: I'm fixing the whole implem, but someone will have to check if the pinch still work (I don't have touch screen with me)

src/Tools/babylon.tools.ts
nockawa: man, if that's still the case I'm in big trouble. I have to test my code on a minified version then and I will keep you updated

src/Tools/babylon.stringDictionary.ts
nockawa: as discussed the purpose of StringDictionary is more a long running object. that's why I've changed the implem in ArcRotate

nockawa:
By the way, I don't understand why there're conflict? Am I able to see them or only you can deal with it?

nockawa:
Ok, I made some changes, could someone please test if touch pinching is still working on ArcRotate? thanks.

deltakosh:
I cannot. There is a conflict. I cannot merge. And btw you could be able to test on your phone

src/Tools/babylon.tools.ts
julien-moreau: This is the reason why the Actions Builder uses the babylon.max.js file instead of the minified version :(

nockawa: Well, I don't care about the class' Name if it changes due to minification, I mean I won't do a getClassName() == MyClass. I just want to built a tree of the class hierarchy and add metadata on them. So typically I call this method to store the name of the class and then later to find it from an instance.
So maybe it could work in my case, right?

nockawa:
Ok, I've updated my repo from upstream locally in a separeted clone, looks like it's working now.
Sorry for the hassle..

PR #1107, opened 2016-04-24
RectPackingMap class + misc

nockawa:


deltakosh:
Don't forget to update doc as well (list of files)

PR #1108, opened 2016-04-25
DynamicFloatArray class

nockawa:
The purpose of this class is to store float32 based elements of a given size (defined by the stride argument) in a dynamic fashion, that is, you can add/free elements. You can then access to a defragmented/packed version of the underlying Float32Array by calling the pack() method.

The intent is to maintain through time data that will be bound to a WebGlBuffer with the ability to change add/remove elements.

It was first built to effiently maintain the WebGlBuffer that contain instancing based data.

Allocating an Element will return a instance of DynamicFloatArrayEntry which contains the offset into the Float32Array of where the element starts, you are then responsible to copy your data using this offset.

Beware, calling pack() may change the offset of some Entries because this method will defrag the Float32Array to replace empty elements by moving allocated ones at their location.

This method will retrun an ArrayBufferView on the existing Float32Array that describes the occupied elements. Use this View to update the WebGLBuffer and NOT the "buffer" field of the class. The pack() method won't shrink/reallocate the buffer to keep it GC friendly, all the empty space will be put at the end of the buffer, the method just ensure there're no "free holes".

### DIFF
--- a/Tools/Gulp/config.json
+++ b/Tools/Gulp/config.json
@@ -23,6 +23,7 @@
       "../../src/Tools/babylon.smartArray.js",
       "../../src/Tools/babylon.stringDictionary.js",
 	  "../../src/Tools/babylon.rectPackingMap.js",
+	  "../../src/Tools/babylon.dynamicFloatArray.js",
       "../../src/Tools/babylon.tools.js",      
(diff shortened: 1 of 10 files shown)


deltakosh:
Don't forget the doc (list of files :))

nockawa:
done

PR #1110, opened 2016-04-26
Fixed Vector2.Transform and DynamicFloatArray.pack

nockawa:


nockawa:
don't  merge yet...i've just spotted one more mistake...

nockawa:
Ok, I believe it's ok now. I'm not fan of large class names, but well, I think the terminology has to be good to be understandable. 
So now we have Element, the Array has a Total, Free, Used Element Count. When we allocate an Element it return an ElementInfo (because it's not the element itself).
I hope it make things clearer this way

deltakosh:
Should I mention config.json and documentation?

deltakosh:
I'll merge it but please test you code with the minified version next time. This one is broken because DynaicFloattArrayElementInfo is not a the right place

nockawa:
Sorry, I don't understand, what is not at the right place and why? The DynamicFloatArrayElementInfo class should be defined before DynamicFloatArray ?
If so why does it work on my machine and not in the minified version?

deltakosh:
Call me stupid I merged two things simultaneously and I mixed the error :)

deltakosh:
Sorry!

nockawa:
no that's ok, in fact I'm far from being confident with what I do in JS :)

deltakosh:
:D

PR #1112, opened 2016-04-28
Update DynamicText options to object descriptor

mwpowellhtx:
With appropriate type checks. Also added *textAlign* to the descriptor. No need to hold a reference, after all, it is ctor only. I'll have a look at any docs following merge.

### DIFF
--- a/src/Materials/Textures/babylon.dynamicTexture.ts
+++ b/src/Materials/Textures/babylon.dynamicTexture.ts
@@ -14,16 +14,17 @@
 
             this._generateMipMaps = generateMipMaps;
 
-            if (options.getContext) {
+            /* there are probably clearer ways to verify all of this */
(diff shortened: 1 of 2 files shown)


deltakosh:
Sorry I may have not been clear enough on the forum..This is not something I want. The textAlign feature should be on the drawText function (And not global for the texture) because you may not want to ALWAYS write aligned texts in your DynamicTexture.
Furthermore, we try to never break backward compatibility.
And you're commit failed on travis for some reason.

mwpowellhtx:
I appreciate the clarification. I'll work on that.

mwpowellhtx:
It's done. Let me know how that works. Also, to clarify, "context" is the
"canvas", or object/mesh containing the drawn text, correct? So, when we
talk about "always aligning" text one way or another, that's local to the
container, no?

PR #1113, opened 2016-04-28
added second drawText function with textAlign argument

mwpowellhtx:
Keep me posted how this works. Thank you...

### DIFF
--- a/src/Materials/Textures/babylon.dynamicTexture.ts
+++ b/src/Materials/Textures/babylon.dynamicTexture.ts
@@ -86,6 +86,11 @@
             }
         }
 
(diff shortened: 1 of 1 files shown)


deltakosh:
actually this could be added directly on drawText itself as an optional parameter.

mwpowellhtx:
That wouldn't be a breaking change for some?

deltakosh:
nope if we just add a if (textAlign){}

mwpowellhtx:
We aren't communicating. Yes, I understand we can test for whether the argument exists.

Breaking changes are in terms of the caller, not the implementation.

RaananW:
The caller won't notice the function had changed of it's optional. If he checks the signature he will understand he can set it, but it's not a must. So an extra optional parameter is not a breaking change.

PR #1114, opened 2016-04-28
Added drawText optional textAlign argument

mwpowellhtx:


nockawa:
hi @mwpowellhtx, for your information the next version of Babylon.JS will feature a text rendering manager, you won't have to draw in texture to render text anymore. and I think you will have the features you're looking for.
this feature is part of an incoming 2D Engine (to render primitives, sprites, text), just wait few more days and you'll see the first version in the 2.4 beta.

mwpowellhtx:
@nockawa Well, this is certainly pleasant news just besides. Hopefully it will be easier to "envision", if you will, text rendered in a plane. This would be a huge step forward for what I'm doing, because right now the relationship between coordinate systems doesn't behave quite right.

nockawa:
you will be able to render text in screen space OR world space, so in a plane that has a position/rotation/scale in the world scene. it will be the same of any other kind of primitive.
basically you create a Canvas2D, specify if it lies in screen space or world space, then you insert primitive in this canvas.

mwpowellhtx:
@nockawa To be clear, this is (basically) what we do now with "DyanmicTexture", although perhaps in "name only", so to speak. How will that be different ensuring X and Y center is truly center of the mesh?

nockawa:
everything has an "origin" which specify where the point of reference is (typically top/left or middle). So alignment will be made regarding the origin you set.

mwpowellhtx:
@nockawa Yes, I understand that. But text being rendered is not that way.

nockawa:
well I know pretty much how text rendering should be done, now the feature will be limited by what we have in JS/HTML. And currently you can't access baseline, ascent, descent information of a particular font. I have access to the width of a text/char, I can detect the max height of a font, but that's all.
I'm not the only one.
[Look at this for instance](https://videlais.com/2014/03/16/the-many-and-varied-problems-with-measuring-font-height-for-html5-canvas/)

mwpowellhtx:
@nockawa I do not expect it should make my coffee, too. However, given those constraints, what are the expectations around centering?

nockawa:
a text primitive will have its size (width/height) automatically computed, if you set its origin to 0,0 the text will be vertically aligned UP and horiz aligned LEFT. If you set the origin to 0.5,0.5, the text will be vertically aligned CENTER and horiz aligned CENTER. And so on...
This text will be positioned in a canvas which will have the same feature, I let you think about the possibilities. I hope this is clear enough now. thank you for your kind patience and time.

mwpowellhtx:
@nockawa Okay, this is pretty much what I do now, with a couple of reasonably informed guesses. But that's all they are, is guesses. What I'm getting at is to respond to more precise revelations (calculations, or what have you) of width, height, etc, as the blog suggests. So I'm not sure what this will add that I am not already doing, though in a roundabout manner.

Anyway, we've beaten this one up pretty well. I'm confident we understand one another.

nockawa:
what you're doing currently is writing text in a dynamic texture, you render text in a HTML canvas mapped to a WebGLTexture. Each time you change the text you lock/render/unlock everything, which is damn slow. About 2000 times slower compared to the feature I developed where I create a texture to store the many used character of a given font, then to render text I render 2 triangles per character, with the texture UV to pick the appropriate character in the map. You can have hundred of char displayed on screen, changing every frame, it won't change a thing concerning the rendering perf and it won't cost you very big texture to hold the whole content.
to sum up: you create an picture containing text. I render text the same way it's done in many 2D/3D engines and hardware accelerated applications/browsers.

mwpowellhtx:
@nockawa Not arguing the speed issue. It's a fair point, but in my use case, it is a non sequitur; I don't care. In my use case, once it's rendered, I leave it alone, so it is not recalculating, unless an input parameter changes, then I recalculate the entire scene, not just the text. It's not instantaneous, but the performance is acceptable. AFAIK, the underlying primitives calculating text, width, etc, are still the same, whether the current implementation runs, or a drop in replacement "manager".

All that to say, I'm looking forward to seeing what the manager adds to the value of the equation. But I haven't heard anything thus far that changes my mind about the underlying inaccuracies, per se.

To be fair, this has been a challenge since time immemorial, for as long as text has been rendered to a graphics architecture.

mwpowellhtx:
Thanks @deltakosh

PR #1121, opened 2016-05-05
Enabled other post processes to be used when also using a 3D Rig

Palmer-JC:


deltakosh:
Hey
I saw that you removed the postProcessesTakenIndices stuff. Does it still work for postprocess pipelines like VolumetricLightScattering?

PR #1130, opened 2016-05-09
interleaved buffers+custom instancing

benaadams:
Add interleaved buffers and custom attributed for instancing


nockawa:
Hello @benaadams can you elaborate more on your intents concerning the interleaved buffers? Why do you need them for?
Concerning custom instancing, I've made a contribution that just got merged few minutes ago which I believe will support it. You can check the methods updateAndBindInstancesBuffer and unBindInstancesBuffer from the Engine class and give me your feedback, thanks!

benaadams:
@nockawa interleaved for 4 reasons 

1. Best practice on [locality of data](https://developer.apple.com/library/ios/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/TechniquesforWorkingwithVertexData/TechniquesforWorkingwithVertexData.html) for mobile chipsets 
iOS Use Interleaved Vertex Data
2. Using a single bindBuffer and single bufferData/bufferSubData call rather than 3 of each to update position+uv+normal
3. Efficent binary object loading. Can use data format: `indexCount`|`indexUint16Array`|`vertexCount`|`vertexFloat32Array` and hoist it straight to GPU memory from the wire with minimal parsing with `vertexFloat32Array` going straight into an interleaved buffer.
4. Instancing updates. We use 6 x 4 float buffers for our instanced objects; again its better to only call `bufferSubData` once rather than 6 times. We also only update "dirty" windows of this data rather than the whole array, but we do update all 6 at the same time, but that is an extension of updating once rather than 6 times.

benaadams:
Also something else we do is share the instancing buffers between models; so we may have a ship, a shield, a weapon, a target readout etc all in the same place as determined by the 6x4 data - further reducing the upload/sync.

nockawa:
@benaadams ok, I was wondering if there was any other reason that performance wise. It all makes sense. 

Concerning your 4., I've developed a class called DynamicFloatArray which is now part of babylon.js. You can check it, basically you define a stride (6 in your case to store elements of 6 float), then you allocate many elements in a big preallocated floatarray buffer. 

The nice part is you can also free some of those elements and then call a pack() method which defrags everything and returns you a subArray of all the data, for you to send to update the instanced Array (for instance). 

Which is useful if you need to cull/hide some instanced objects, then don't draw their primitive by the GPU.

benaadams:
@nockawa it would be stride 24 (6 x 4) - will check it out

nockawa:
Well, as the class is "Dynamic**Float**Array", the stride is in float size, not in byte. (it relies on a Float32Array btw)

benaadams:
@nockawa shouldn't `DynamicFloatArrayElementInfo` just be a number? Rather than `new`ing a `DynamicFloatArrayElementInfo` on each use?

nockawa:
Nope you need an actual object (which is allocated in an array by the way) because the "number" is the offset of your element in the float32array and it can change when you call pack().
The pack() method will move allocated elements to fill "empty" holes, so the offset to access their data will change. (and I copy the data which is moved)

benaadams:
Interesting :)

nockawa:
Yep, I'm pretty happy about this class, the pack() method is most of the time in O(1), worst case is close to O(k) is your browser use merge sort for the sort method. Anyway: it's fast.
My only regret so far is it stores only float32 so you can't have other data type in your instanced array. But I don't think JavaScript will allow me to create something efficient anyway, it's not C++...

I've developed a whole system (part of the new Canvas2D feature of babylon.js) that create/setup/update Instanced Array generically based on a mapping structure with properties decorated that write through the DynamicFloatArray buffer. I also do the Effect's attributes, uniforms, defines wiring generically. Tell me if you want to know more, I'd be happy to share pointers. You can start [the overview](http://doc.babylonjs.com/overviews/Using_The_Canvas2D), but it's still WIP.

benaadams:
> Concerning custom instancing, I've made a contribution that just got merged few minutes ago which I believe will support it. You can check the methods updateAndBindInstancesBuffer and unBindInstancesBuffer from the Engine class and give me your feedback, thanks!

@nockawa it isn't very integrated and needs a new type with a custom override of render to use?

nockawa:
The updateBind/unBind now accept an array of InstancingAttributeInfo objects, which contains everything to setup the attributes properly (at the exception of the repeating count, which is set to one).
So basically you keep a track of this array and submit it to bind/unbind.
I'm sorry but I don't get your point concerning the custom override of render. You can directly call engine.draw(), specifying an instanceCount and things will work as expected

benaadams:
But using `engine.draw()` you can't call `scene.render()`

nockawa:
sorry, I don't get your point. all I did was making these two methods generic regarding the WebGL extensions. it's low level stuff.

deltakosh:
Hello folks! Interesting discussion :) @benaadams : did you find what you were looking for?

benaadams:
@deltakosh refactoring on top... You actually use interleaved buffers all over the place, so standardising.

deltakosh:
Ok sounds good:)

benaadams:
@deltakosh still wip, but showing progress

deltakosh:
No rush:)

deltakosh:
we plan to ship 2.4 next week. So based on your timing this may be merged in 2.5

benaadams:
Code updated, but checking it doesn't break stuff and works for what we want.

Is there a recommended way to check compat; or just go through every sample?

deltakosh:
No official recommended way. Checking samples is great:)

deltakosh:
And I can do it for you if you wish

src/Mesh/babylon.mesh.ts
```diff
@@ -900,51 +920,55 @@
...
                     offset += 16;
                     instancesCount++;
                 }
             }
 
-            var offsetLocation0 = effect.getAttributeLocationByName("world0");
-            var offsetLocation1 = effect.getAttributeLocationByName("world1");
-            var offsetLocation2 = effect.getAttributeLocationByName("world2");
-            var offsetLocation3 = effect.getAttributeLocationByName("world3");
+            if (!instancesBuffer || currentInstancesBufferSize != this._instancesBufferSize) {
```
nockawa: I understand that checking the size is enough for now, but it might be dangerous in the future, or definitely something worth remembering...

nockawa:
this is a great addition @benaadams, congratulations and thank you!
I'll take a closer look at the code tomorrow, I'm dead tired now. time to bed.

benaadams:
@deltakosh I'll make sure it seems to be working then let you know; will also get me familiar with the build system

@nockawa thank you :)

deltakosh:
You have to update config.json file to add new added files (in tools/gulp)

And most of all: Credit yourself in what's new.md (in dist/preview release)

deltakosh:
I would also recommend adding a ref to the breaking change: bindBuffers to bindBuffersDirectly (we have a small section for that at the end of the what's new)

src/Mesh/babylon.geometry.ts
```diff
@@ -119,6 +119,37 @@
...
 
+        public setVerticesBuffer(buffer: IVertexBuffer): void {
+            var kind = buffer.getKind();
+            if (this._vertexBuffers[kind]) {
+                this._vertexBuffers[kind].dispose();
+            }
+
+            this._vertexBuffers[kind] = buffer;
+
+            if (kind === VertexBuffer.PositionKind) {
```
deltakosh: I think we can factorize code here with setVerticesData

src/Mesh/babylon.mesh.ts
deltakosh: You mean because same size but different data, right?

src/Mesh/babylon.interleavedBuffer.ts
```diff
@@ -0,0 +1,115 @@
...
+
+            if (this._updatable) { // update buffer
+                this._engine.updateDynamicVertexBuffer(this._buffer, data, offset);
+                this._data = null;
+            }
+        }
+
+
+        public createReference(vertexBuffer: InterleavedVertexBuffer) {
+            this._references.push(vertexBuffer);
```
deltakosh: What's the purpose of references?

src/Mesh/babylon.vertexBuffer.ts
```diff
@@ -1,14 +1,28 @@
...
+        getOffset(): number;
+        getSize(): number;
+        getIsInstanced(): boolean;
+        isUpdatable(): boolean;
+        create(data?: number[] | Float32Array): void;
+        update(data: number[] | Float32Array): void;
+        updateDirectly(data: Float32Array, offset: number): void;
+    }
+
+    export class VertexBuffer implements IVertexBuffer {
```
deltakosh: I'm wondering if we need this class. I think we could rename InterleavedBuffer to VertexBuffer and get rid of this one.
Thoughts?

Stride parameter can be set to optional and deduced using the same code as here

benaadams: could work :)

benaadams:
Definitely some bugs, working though them.

src/Mesh/babylon.vertexBuffer.ts
benaadams: Not sure as `InterleavedBuffer` doesn't have a kind or implement a lot of the interface; its the `InterleavedVertexBuffer`s that map onto it that have the `kind`s and have `offset` and `size`

src/Mesh/babylon.mesh.ts
benaadams: The data change should get picked up by the other half of the if?

src/Mesh/babylon.interleavedBuffer.ts
benaadams: Removed

src/Mesh/babylon.geometry.ts
benaadams: Refactored

benaadams:
K, looking better - haven't done a wide test, or tried how it works with what we want to do - but the current instancing seems to continue working with the new approach

src/Mesh/babylon.vertexBuffer.ts
deltakosh: we can at least rename interleavedbuffer to buffer and reuse it inside vertexbuffer. my thought is that we can definitely reuse code here:)

src/Mesh/babylon.mesh.ts
deltakosh: i think so:)

src/Mesh/babylon.vertexBuffer.ts
benaadams: ahh, this could work...

benaadams: Refactored

benaadams:
Refactored, works with what we are trying to do, demos seem to work (haven't tested extensively), updated what's new.

Could you review?

deltakosh:
Working on it

benaadams:
Rebased for what's new merge clash

benaadams:
I think I broke particles... looking into it

deltakosh:
Ok thanks:)

benaadams:
Fixed particles

deltakosh:
Merging!
Thanks a lot Ben this is awesome

deltakosh:
Still getting errors with particles..Working on it

deltakosh:
Also found issue with instancing + shadows

benaadams:
What kind of errors? They look the same to me

particles

deltakosh:
I tested with this one: http://localhost:29919/Demos/Starfield/

deltakosh:
and this one for shadows: http://www.babylonjs.com/Demos/InstancedBones/

deltakosh:
ok..shadows are not related to your updates :D

benaadams:
K, can create starfield issue - working on it

deltakosh:
Ok I'm focusing on isntanced + shadows then

deltakosh:
Ok shadows fixed..

deltakosh:
The filler seems not required for particles

deltakosh:
Question: The _currentBoundBuffer should be indexed by target right? Else we lost many cache opportunities

deltakosh:
Found the bug :)
In particleSystem.ts: line 127: this._vertexData = new Float32Array(capacity * 11 * **4**);

deltakosh:
Ok merging the fix! we are good to go! Thank you so much for this huge improvement

benaadams:
> Found the bug :)
In particleSystem.ts: line 127: this._vertexData = new Float32Array(capacity * 11 * 4);

Of course, 4 verteices per particle

benaadams:
> Question: The _currentBoundBuffer should be indexed by target right? Else we lost many cache opportunities

Yes you are right, will make a PR

benaadams:
Ah, you already did it :+1:

deltakosh:
LOl yeah..I was already pushing an update:)

jbousquie:
This will be an incredible underground improvement !
We always under estimate the people who dig the sewer or the building foundations.

Congratulations, guys :-)

note : this should be written in bold in the "what's new" file imho

nockawa:
@benaadams I merge my latest code with your changes and Rectangle2D doesn't work anymore. It works when I don't use the Instanced Array rendering path, but when it's on the rendering is totally f**ked up...

benaadams:
@nockawa do you have a sample I can look at?

nockawa:
```            let canvas = Canvas2D.CreateScreenSpace(this.scene, { id: "ScreenCanvas", cachingStrategy: Canvas2D.CACHESTRATEGY_DONTCACHE });
            var rect = Rectangle2D.Create(canvas, {id:"button", x:100, y:100, width:100, height:40, fill:Canvas2D.GetSolidColorBrushFromHex("#40C040FF")});
            rect.roundRadius = 10;
```

nockawa:
but you have to get the sources and compile them from my repo in the engine2d branch. thanks!

benaadams:
@nockawa so are you saying your merge breaks? What if you rebase on top rather than merge as your commit stream is a little hard to follow.

When I run on the current
```ts
window.addEventListener('DOMContentLoaded', () => {
    // get the canvas DOM element
    var canvas = <HTMLCanvasElement>document.getElementById('renderCanvas');
    
    var engine = new BABYLON.Engine(canvas, true);

    let canvas2d = BABYLON.Canvas2D.CreateScreenSpace(scene, { id: "renderCanvas", cachingStrategy: BABYLON.Canvas2D.CACHESTRATEGY_DONTCACHE });
    var rect = BABYLON.Rectangle2D.Create(canvas2d, { id: "button", x: 100, y: 100, width: 100, height: 40, fill: BABYLON.Canvas2D.GetSolidColorBrushFromHex("#40C040FF") });
    rect.roundRadius = 10;
    // create a FreeCamera, and set its position to (x:0, y:5, z:-10)
    var camera = new BABYLON.UniversalCamera('camera1', new BABYLON.Vector3(0, 5, -10), scene);

    // target the camera to scene origin

    // attach the camera to the canvas
    camera.attachControl(canvas, false);
    scene.render();
});
```
The output I get looks like
image

nockawa:
Yep, this output is ok, but it's not the same with the version I have locally.
Sorry but I don't know much about git, what exactly would you like me to do?

benaadams:
Looking, this is the set of changes https://github.com/benaadams/Babylon.js/pull/1

benaadams:
@nockawa I get the same output with those changes https://github.com/benaadams/Babylon.js/tree/nockawa?

benaadams:
After doing a build with 
`gulp typescript-compile`
`gulp`

nockawa:
I have to say, I really don't understand what's going on....

nockawa:
So to recap, you created a branch on your repo, you merge the changes from mine, recompile and execute the sample I gave and it works on your machine, that's correct?

benaadams:
Yes, I pulled all the changes from the current BabylonJS/master, made a branch, then merged your branch into it and it all seemed ok

nockawa:
I just cloned your branch locally and run, and it works! wtf!
I think I have to diff your branch and mine...

benaadams:
Re-branch from latest main, then do a PR from your branch into it, and work from that? Is what I did

nockawa:
I just PR from your branch to mine, rebuild and it's failing again. I'm going to do what you just said.

nockawa:
did that, still doesn't work...attempting something else, I clone my git repo into another local location...
I may have something wrong locally...

PR #1142, opened 2016-05-18
Device Orientation Cameras

RaananW:
Fixing the device-orientation cameras (both VR and non-VR one).
Up-Vector is now recalculated using the quaternion.

Target camera now has a quaternion that acts the same as a mesh's rotationQuaternion - if it is defined it will be used, and if not the Vector3 rotation will be used.

With great thanks to Neo-Pangea.


deltakosh:
Why deleting Autodesk dll :) ?

RaananW:
Now that is interesting :)
It came from the merge, Let me look into that.

RaananW:
The DLL is being ignored for some reason, had to force-add it. I can't find the gitignore line that does it, but it should be fine now.

Temechon:
Super mega cool :)

NeoHoffa:
You guys are pretty swell!!!

PR #1144, opened 2016-05-19
 Canvas2D: Animation and ActionManager support + bugs fixes

nockawa:
ActionEvent: adding a CreateNewFromPrimitive (which takes a prim as 'any' to allow building Core)

SmartPropertyPrim:
 - Adding the animation proeprty to support animations.
 - Adding Inline documentation

Prim2DBase:
 - Adding actionManager get property, supporting every trigger except KeyDown/Up and IntersectionEnter/Exit

Text2D: improved rendering quality and fix Origin computation

FontTexture: Now use NEAREST sampling mode to achieve a better rendering quality

### DIFF
--- a/src/Animations/babylon.animation.ts
+++ b/src/Animations/babylon.animation.ts
@@ -124,6 +124,8 @@
                 dataType = Animation.ANIMATIONTYPE_VECTOR2;
             } else if (from instanceof Color3) {
                 dataType = Animation.ANIMATIONTYPE_COLOR3;
+            } else if (from instanceof Size) {
+                dataType = Animation.ANIMATIONTYPE_SIZE;
             }
 
             if (dataType == undefined) {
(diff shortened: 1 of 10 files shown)


src/Animations/babylon.animation.ts
```diff
@@ -453,15 +468,16 @@
                 destination = this._target;
             }
 
+            if (!this._originalBlendValue) {
```
deltakosh: why did you move this outside the loop?

src/Animations/babylon.animation.ts
```diff
@@ -482,7 +498,7 @@
             }
 
             if (this._target.markAsDirty) {
-                this._target.markAsDirty(this.targetProperty);
+                this._target.markAsDirty(this.targetProperty, this._originalBlendValue);
```
deltakosh: Is this mandatory? actions are not providing the 2nd parameter for now

nockawa: I figured that in JS it's no big deal to call with a second arg, right?

deltakosh: My point is: actions will NOT provide the 2nd parameter.

PR #1147, opened 2016-05-20
Added SPS per particle visibility + fix config

jbousquie:
Added SPS per particle visibility : now each particle has a new property _isVisible_ (default true)
This property actually temporary scales the particle to zero and sets it to the camera position in the VBO so it's not visible, either pickable and doesn't cast any shadow. This doesn't change the current particle property values like position, rotation, scale that can be still computed and updated by the user as user, because they all keep only logical properties and not VBO values.

Added a fix (again) in the config.json file : case was wrong in file names


jbousquie:
oopss sorry, it seems I missed a commit
fixing it right now

PR #1154, opened 2016-05-24
Improved typing for PostProcess constructor

Kesshi:
This PR improves the typing for "ratio" parameter of the PostProcess constructor.
I also modified most of the derived classes because they defined the ration as "number" only and not "number | { width: number, height: number }".


deltakosh:
This seems wrong to me. Ratio should be number | PostProcessRatio

Kesshi:
Ok i can change that if you want but PostProcessRatio should be renamed then in my opinion because its not a "ratio" anymore. Maybe PostProcessSize? Or a general PostProcessOptions for future additions?

deltakosh:
options is cool

Kesshi:
done

PR #1157, opened 2016-05-25
Webvr fixes

RaananW:
Tested on the oculus rift DK2, WebVR is now supported (both orientation and correctly displaying). 
A new requestVRFullscreen is available, this will send the entire canvas to the VR device chosen by the camera.


deltakosh:
You should update readme as well :)

NeoHoffa:
WOOHOO!!!!!!!!!!!!!!!!!!!!!!!

RaananW:
Just noticed something might be missing in certain cases.
Fixing!

PR #1159, opened 2016-05-26
Blender Exporter 4.5.1

Palmer-JC:
Ensure every single mesh, light, armature, & camera are all exported
based on the same frame.  Mesh & armature were being define out of sync
depending on whether there was a frame 0 or not.


Palmer-JC:
forgot lights

PR #1180, opened 2016-06-07
SIMD fixes + asm.js

benaadams:
Issue with minifed

### DIFF
--- a/Tools/Gulp/package.json
+++ b/Tools/Gulp/package.json
@@ -8,7 +8,7 @@
   "license": "(Apache-2.0)",
   "devDependencies": {
     "gulp": "^3.8.11",
-    "gulp-uglify": "~1.2.0",
+    "gulp-uglify": "~1.5.3",
     "gulp-sourcemaps": "~1.5.2",
     "typescript": "^1.7.5",
     "gulp-typescript": "~2.13.0",
--- a/src/Math/babylon.math.SIMD.ts
+++ b/src/Math/babylon.math.SIMD.ts
@@ -1,255 +1,595 @@
 ﻿module BABYLON {
+    interface SIMDFuncs {
+        TransformCoordinatesSIMD(x: number, y: number, z: number): void;
+        multiplyToArraySIMD(): void;
+        multiplyResultToArraySIMD(): void;
+        invertToRefSIMD(): void;
+        LookAtLHToRefSIMD(): void;
+    }
 
-    declare var SIMD;
+    class SIMDMath {
+        public static SIMDHeap: Float32Array;
+        public static SIMDFuncs: SIMDFuncs;
 
-    export class SIMDVector3 {
-        public static TransformCoordinatesToRefSIMD(vector: Vector3, transformation: Matrix, result: Vector3): void {
-            var v = SIMD.float32x4.loadXYZ((<any>vector)._data, 0);
-            var m0 = SIMD.float32x4.load(transformation.m, 0);
-            var m1 = SIMD.float32x4.load(transformation.m, 4);
-            var m2 = SIMD.float32x4.load(transformation.m, 8);
-            var m3 = SIMD.float32x4.load(transformation.m, 12);
-            var r = SIMD.float32x4.add(SIMD.float32x4.add(SIMD.float32x4.mul(SIMD.float32x4.swizzle(v, 0, 0, 0, 0), m0), SIMD.float32x4.mul(SIMD.float32x4.swizzle(v, 1, 1, 1, 1), m1)), SIMD.float32x4.add(SIMD.float32x4.mul(SIMD.float32x4.swizzle(v, 2, 2, 2, 2), m2), m3));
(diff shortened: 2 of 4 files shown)


benaadams:
Disabled multiplyToArraySIMD function as that is one causing issue atm

RaananW:
If you want to revert the vector3 changes you will have to revert them all. You have create an infinite loop : `get x() { return this.x }` will constantly call the getter.

benaadams:
Whoops, fixed

deltakosh:
Regarding perf, we must keep the vector3 as is and instead provide a replacment class when SIMD is active. Accessing filed values through properties is kind of slow for an object like vector3 (which is used everywhere)

nockawa:
a little jsfiddle to proove that we did consider it seriously: https://jsfiddle.net/v6nkbo6b/4/
prop access is 10 times slower...(it depends on the browser, etc, but it's really a big hit)

benaadams:
but then you take the hit by using simd; direct on the Float32Array is petty bad in that jsfiddle, might be better to load directly into the SIMD types from the x,y,z - will look into it...

deltakosh:
Agree. In this case we need a Vector3SIMD which will override the Vector3

benaadams:
@deltakosh question was there a big diff between SIMD vs non-SIMD previously? Not seeing it - might be my gpu hitting limits early though

deltakosh:
No it was definitely not obvious

benaadams:
ah, k might be worth continuing then, SIMDing more things

deltakosh:
yep but we need to stay away from updating pure js code by providing classes to override existing code

benaadams:
Looking much better, still investingating

benaadams:
I'm hitting other perf issues before the SIMD, so i'm going to knock those off along the way

benaadams:
Sorry if this is going to be a noisy PR, just keeping track of the changes

Reduce Observable.notifyObservers allocations
image

deltakosh:
No worry:) regarding observables: If creating them is a problem, we can think about adding a getter to create them on demand

benaadams:
@deltakosh is `EventState` creation, resolved it in last commit https://github.com/BabylonJS/Babylon.js/pull/1180/commits/caa220a55ffb6d5124cb1b294c4c1f62de6f75a1

deltakosh:
Gotcha

benaadams:
Next big ones seem to be `MaterialDefines`

image

benaadams:
define caching isn't working well on frame, and very much not cross frame

deltakosh:
What do you mean by "on frame"?

benaadams:
The simd demo has 50 meshes that all the defines are the same so 3 of the top 3 cpu methods shouldn't be dealing with defines as they are all the same in the same `requestAnimationFrame`; and the same between frames but there may be other objects mixing with that; but could be cached cross-mesh for example.

There is a lot of identical work going on

deltakosh:
Agree. The defines should be cached and reuse as there is only one single material. Something is wrong here

deltakosh:
This is because of the mirror :( It forces meshes to be drawn with a clipplane and then it invalidates the cached define

deltakosh:
For this specific scene you can add "newOne.material.freeze();" to the addDancer function
This will force the material to remain the same between frame (There is no need for clipplane here as no object go through the mirror)

deltakosh:
I updated the scene online with this small update

benaadams:
@deltakosh not seeing this change; is it in Samples?

deltakosh:
Only online so far: www.babylonjs.com/Demos/SIMD/

benaadams:
Hmm different to src, e.g. I can't take the index.html code and apply it locally as material and freeze don't exist...

deltakosh:
Not sure to follow you. Which code are you using?

deltakosh:
https://github.com/BabylonJS/Samples/blob/master/Demos/SIMD/index.html

benaadams:
My fault, had broken something elsewhere

benaadams:
Moved the non-SIMD changes to #1185

arunetm:
@benaadam, We are evaluating some new simdjs optimizations (type specialization for non-asm code) in Msft edge browser. It will be great if we can use the version of simd.js demo based on the changes you have in this PR. The plan is to use the demo on a test version of edge with the new simd optimizations enabled and we might to see more speedup when using the non-asm simd library. 

Is there any copy of the simdjs demo form babylonjs using your changes? 
We tried doing an asmjs version of the simd math library in Babylonjs, but creating and initializing an asmjs module every time when the library is used for computation turned out to be a major bottleneck. Are you able to see any speedup for either versions?

benaadams:
@arunetm the main bottleneck I'm seeing with this simd asmjs version in the is in the copy step to and from the asmjs buffer - the simd portions of the functions themselves are very fast. However I was experiencing other bottlenecks with the non-asmjs versions; which I have now addressed and they have been merged to master.

The asm+simd did seem faster than the pure simd version; however with some of the changes it might be better now and obv if you are looking to improve the pure simdjs code its probably worth pursuing further. I'll remake a pure simdjs PR for you to look at.

PR #1181, opened 2016-06-07
SIMD fixes

benaadams:
Just fixes for SIMD with asm.js removed

See also: https://github.com/BabylonJS/Babylon.js/pull/1180

### DIFF
--- a/Tools/Gulp/package.json
+++ b/Tools/Gulp/package.json
@@ -8,7 +8,7 @@
   "license": "(Apache-2.0)",
   "devDependencies": {
     "gulp": "^3.8.11",
-    "gulp-uglify": "~1.2.0",
+    "gulp-uglify": "~1.5.3",
     "gulp-sourcemaps": "~1.5.2",
     "typescript": "^1.7.5",
     "gulp-typescript": "~2.13.0",
--- a/src/Math/babylon.math.SIMD.ts
+++ b/src/Math/babylon.math.SIMD.ts
@@ -1,255 +1,236 @@
 ﻿module BABYLON {
 
-    declare var SIMD;
+    declare var SIMD: SIMD;
+    var f4 = SIMD.Float32x4;
+    var load = f4.load;
+    var add = f4.add;
+    var sub = f4.sub;
+    var mul = f4.mul;
+    var div = f4.div;
+    var swizzle = f4.swizzle;
+    var splat = f4.splat;
+    var store = f4.store;
+    var shuffle = f4.shuffle;
+    var extract = f4.extractLane;
+    var replaceLane = f4.replaceLane;
+    var neg = f4.neg;
+    var reciprocalApproximation = f4.reciprocalApproximation;
+    var reciprocalSqrtApproximation = f4.reciprocalSqrtApproximation;
 
-    export class SIMDVector3 {
+    class SIMDVector3 {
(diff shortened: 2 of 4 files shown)


RaananW:
Hi Ben,

1. what's the difference between the last PR and this one?
2. The SIMD file is a d.ts file. Let me think where it fits the best.
3. Was the change of Vector3 needed?

benaadams:
@RaananW 

1. Last one the SIMD functions are written to be asm.js AST pre-compiled with the parameters going to the functions via a shared heap (between asm.js and js). I wasn't convinced it added any perf, so I tried this one where just fixed the current implementation - not convinced by its performance either :see_no_evil: (using the SIMD demo to test)

2. Sure

3. The previous SIMD implementation when activated mutated each Vector object to add the `Float32Array` `_data` backing field in the `x` setter and indirected `x`,`y`,`z` on the prototype to via property getters and setters, so I changed it to happen in the constructor for a monomorphic type cache rather than having a polymorphic type cache on all the functions taking `Vector3`s (e.g. https://blogs.msdn.microsoft.com/ie/2014/10/09/announcing-key-advances-to-javascript-performance-in-windows-10-technical-preview/). I've reverted most of the changes to go back to using `x`,`y`,`z` however.

benaadams:
e.g. I'm not sure if  #1180 or #1181 is better as I don't seem to getting a great benefit from either in Firefox Nightly or Edge (with experiment js on)

RaananW:
@benaadams - I totally forget we are currently in code freeze, bug fixing only. This will be a great feature for 2.5, but we will need to test all together and see how much benefit it brings. 
I am commenting here, but it is, of course, referring to https://github.com/BabylonJS/Babylon.js/pull/1180 as well.

deltakosh:
Which on (between this one an #1180) do you want to keep?

benaadams:
Closing this as making good progress on other

benaadams:
@arunetm this is the pure simdjs change, however it currently performs worse that the asmjs with the double copy step

image

arunetm:
Thanks. Edge currently wont optimize the non-asm version of SIMD, and this explains the degraded performance here. The new typespecialization optimization for simd should address this case and we plan to use this update to evaluate the same. Thank you for the PR.

arunetm:
@benaadams. The time spent in "SIMDMatrix.prototype.multilyToArraySIMD" drops from 60% to 14%, in Edge with new simd optimizations running the simd enabled version of the demo. I could see that the "SIMDMatrix.prototype.multilyToArraySIMD" shows speedups compared to the non-simd counterpart, but this didnt translate to an increase in number of dancers in the actual demo. The demo is spending most of the time outside the simd math library and the libraries are not stressed enough to see the benefits of simd speedup. 
Did you notice anything similar in your trials? Are you seeing any increase in dancers with asmjs changes? 

edge-simd

benaadams:
Do you have the call stacks for the other areas? (or better the `.diagsession` file) can improve them.

The asm.js approach with the copy-to and copy-from steps is too slow for the general js case; whereas if the simd is recognised in general js then the entire maths functions can be switched when SIMD is enabled.

deltakosh:
Hey Ben,
I plan to close your 2 PRs unless you want to merge one of them?

benaadams:
I think this one is better as it does fix the SIMD not working.

While the other one seems to be more SIMD accelerated; too much speed is lost copying the data back and forth between asm and regular.

PR #1183, opened 2016-06-08
Don't force DynamicTexture height/width to power of two

dahlbyk:
I found this behavior of `DynamicTexture`'s constructor to surprising/frustrating. I think the better approach is to respect the given `height`/`width` by default, with an option to force an exponent of two. If you don't agree with that change in default behavior, I structured the commits such that https://github.com/BabylonJS/Babylon.js/commit/4544934ff785bf18f9ebd1947681cc7603ddba78 can be reverted to remove the breaking change but keep the `forceExponentOfTwo` option.

First PR here, so eager for any feedback. I've tried to update What's New, as well.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -73,6 +73,7 @@
     - Fixed bug with OBJ Loader - All meshes were concatenated with the previous one ([Temechon](https://github.com/Temechon))
     - Fixed the device orientation cameras (both VR and non-VR cameras)  ([RaananW](https://github.com/RaananW))
     - Fixed the WebVR implementation  ([RaananW](https://github.com/RaananW))
+    - `DynamicTexture.clone()` now preserves height in addition to width  ([dahlbyk](https://github.com/dahlbyk))
   - **Breaking changes**
     - `VertexData.CreateLines()` removed as `MeshBuilder.CreateLines()` now calls `MeshBuilder.CreateLineSystem()`
     - `scene.onNewXXXAdded` and `scene.onXXXRemoved` callbacks were removed and replaced by `scene.onNewXXXAddedObservable` and `scene.onXXXRemovedObservable`
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -30,7 +30,7 @@
     - Enabled other post processes to be used when also using a 3D Rig ([jcpalmer](https://github.com/Palmer-JC))
     - Got Skeleton.copyAminationRange scaling better for different bone lengths ([jcpalmer](https://github.com/Palmer-JC))
     - Added skeleton.getBoneIndexByName(boneName: string) ([dad72](https://github.com/dad72))
-    - Added node._children to track children hierarchy ([deltakosh](https://github.com/deltakosh))
+    - Added `node._children` to track children hierarchy ([deltakosh](https://github.com/deltakosh))
     - Added Camera.ForceAttachControlToAlwaysPreventDefault to help embedding Babylon.js in iFrames ([deltakosh](https://github.com/deltakosh))
     - Support for Layer.alphaTest ([deltakosh](https://github.com/deltakosh))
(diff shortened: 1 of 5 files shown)


dahlbyk:
Nevermind, this isn't working as expected.

dahlbyk:
So maybe not a total loss? I now understand the power-of-two requirement:

> [.CommandBufferContext]GL ERROR :GL_INVALID_OPERATION : glGenerateMipmap: Can not generate mips
> [.CommandBufferContext]RENDER WARNING: texture bound to texture unit 0 is not renderable. It maybe non-power-of-2 and have incompatible texture filtering.

If `generateMipMaps` is specified as `false` this seems to work as expected. Not sure my proposed `forceExponentOfTwo` option actually provides much value in this context - perhaps simply using `generateMipMaps` as `forceExponentOfTwo` is sufficient to address my concern? Would be good to document the behavior/constraint regardless.

RaananW:
Hi,
First, thanks for the commit!

As this is not a bug, we will have to wait with it to 2.5 as we are currently in code freeze . As I see you already understood that the exp of 2 is important :smile:

About changing default behavior - we try to avoid this as much as possible, for backwards-compatibility. 
The PR needs to be tested fully, as I see you also called it "WIP". Wanna close it and reopen when you are sure everything is working?

Thanks!

dahlbyk:
> As this is not a bug, we will have to wait with it to 2.5 as we are currently in code freeze.

Unless there is some constraint that a `DynamicTexture` with MIP maps must also be square, I'd call https://github.com/BabylonJS/Babylon.js/pull/1183/commits/2f5bc8d46ced2eda3ab0924f2cc1992af0d78c75 a bug, but it's been around for years so it's certainly not urgent. (FWIW, my use case is also square, I just noticed the edge case while I was looking at height/width behavior.)

Feel free to cherry-pick https://github.com/BabylonJS/Babylon.js/pull/1183/commits/7c1c5bc57d826144042ff506fe5ce7b1a35e668b, though. :grinning: 

> As I see you already understood that the exp of 2 is important :smile:

Indeed I have! If you agree that there's value in supporting `options.forceExponentOfTwo = false` when MIP maps are not required, would you prefer the problematic combination of `options.forceExponentOfTwo = false` and `generateMipMaps = true` to:

1. Silently force exponent anyway, so MIP maps work
2. Respect `forceExponentOfTwo` and let users encounter the same error I found
3. Throw an error that precisely describes the configuration mismatch (if height/width aren't powers of 2)

I lean toward (2) simply because the implementation is trivial and the WebGL error is pretty self-explanatory.

> About changing default behavior - we try to avoid this as much as possible, for backwards-compatibility.

Understood. I'm glad to revert the breaking portion of the change in favor of the option to opt into my expected behavior.

> The PR needs to be tested fully, as I see you also called it "WIP". Wanna close it and reopen when you are sure everything is working?

Other than the question above, I'm pretty confident in the change now—at least, it's working fine for my limited use case. I will close and reopen if you prefer, otherwise I can revert the breaking change and rebase this PR on v2.4 once available.

deltakosh:
The clone issue is certainly a bug.
Regarding powof2, here is what I suggest: if samplingmode =nearest and generateMips = false, then we keep the original values. This way no need for breaking changes or additional parameter

dahlbyk:
> Regarding powof2, here is what I suggest: if samplingmode =nearest and generateMips = false, then we keep the original values. This way no need for breaking changes or additional parameter

This is working well for my project, and this branch has been updated accordingly. I left the What's New note under Breaking Changes because _technically_ someone using `false, Texture.NEAREST_SAMPLINGMODE` will experience a change in actual height/width for non-powers-of-2.

Thanks, guys!

deltakosh:
Actually I wanted to mention clamping and not filtering ;) I'll merge your PR and do the appropiate update:)

dahlbyk:
:cool: As a WebGL newbie I'm rather curious what the constraints are on `DynamicTexture` with respect to the dimensions, sampling, clamping, etc.

PR #1185, opened 2016-06-12
Reduce allocations and GPU state changes

benaadams:
Non SIMD/asm changes from #1180 

With the tweak to the sample https://github.com/BabylonJS/Samples/commit/b21a56c51044db967bad7544b7b246f939f3a7c0 and these changes I gain about 20 dancers in the non-SIMD path

Allocations (reduced to just rAF callback allocation per frame)
* Prevent Observable.notifyObservers allocation
* Prevent per mesh per render closure allocation
* Avoid inadvertent per geometry per render alloc
* Avoid inadvertent per mesh per render alloc
* Avoid Collider root calc allocations
* Reduce mesh rotation allocs
* Allocate less in vector project and unproject

Decrease GPU state changes and redundant calls
* Improved uniform caching
* Cache program
* Cache vertexAttribPointer
* Reduce buffer binds	
* Only bind unbound framebuffers
* Cache texture changes better

### DIFF
--- a/src/Materials/babylon.effect.ts
+++ b/src/Materials/babylon.effect.ts
@@ -76,7 +76,7 @@
         private _indexParameters: any;
 
         private _program: WebGLProgram;
-        private _valueCache = [];
+        private _valueCache: { [key: string]: any } = {};
 
         constructor(baseName: any, attributesNames: string[], uniformsNames: string[], samplers: string[], engine, defines?: string, fallbacks?: EffectFallbacks, onCompiled?: (effect: Effect) => void, onError?: (effect: Effect, errors: string) => void, indexParameters?: any) {
             this._engine = engine;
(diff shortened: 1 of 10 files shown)


benaadams:
@deltakosh assume won't go in to current due to freeze? Will keep adding items if not

src/Materials/babylon.effect.ts
```diff
@@ -391,103 +391,164 @@
...
+            var changed = false;
+            if (cache[0] !== x) {
+                cache[0] = x;
+                changed = true;
+            }
+            if (cache[1] !== y) {
+                cache[1] = y;
+                changed = true;
+            }
+            if (cache[2] !== y) {
```
deltakosh: should be z

src/Materials/babylon.effect.ts
```diff
@@ -391,103 +391,164 @@
...
```
deltakosh: same here :)

deltakosh:
Correct. I'll merge it in 2.5

src/babylon.engine.ts
```diff
@@ -976,17 +987,45 @@
...
+                if (pointer.buffer !== buffer) { pointer.buffer = buffer; changed = true; }
+                if (pointer.size !== size) { pointer.size = size; changed = true; }
+                if (pointer.type !== type) { pointer.type = type; changed = true; }
+                if (pointer.normalized !== normalized) { pointer.normalized = normalized; changed = true; }
+                if (pointer.stride !== stride) { pointer.stride = stride; changed = true; }
+                if (pointer.offset !== offset) { pointer.offset = offset; changed = true; }
+            }
+
+            if (changed) {
+                if (this._currentBoundBuffer[this._gl.ARRAY_BUFFER] !== buffer) {
```
deltakosh: why not keeping bindBuffer function here? It is the same code as in bibdArrayBuffer

src/babylon.engine.ts
```diff
@@ -976,17 +987,45 @@
...
+                    this._gl.bindBuffer(this._gl.ARRAY_BUFFER, buffer);
+                    this._currentBoundBuffer[this._gl.ARRAY_BUFFER] = buffer;
+                }
+                this._gl.vertexAttribPointer(indx, size, type, normalized, stride, offset);
+            }
+        }
+
+        private bindIndexBuffer(buffer: WebGLBuffer): void {
+            if (this._currentBoundBuffer[this._gl.ELEMENT_ARRAY_BUFFER] !== buffer) {
```
deltakosh: same here: we should centralize this code into bindBuffer

src/Materials/babylon.effect.ts
benaadams: fixed

src/Materials/babylon.effect.ts
benaadams: fixed

src/babylon.engine.ts
```diff
@@ -1271,7 +1307,11 @@
             this._vertexAttribArraysEnabled = this._vertexAttribArraysEnabled || [];
 
             // Use program
-            this._gl.useProgram(effect.getProgram());
+            var program = effect.getProgram();
+            if (this._currentProgram !== program) {
```
deltakosh: I wanted to do this since ages and I always forgot about it! Thanks :)

src/babylon.engine.ts
```diff
@@ -2208,7 +2248,11 @@
         }
 
         public bindSamplers(effect: Effect): void {
-            this._gl.useProgram(effect.getProgram());
+            var program = effect.getProgram();
+            if (this._currentProgram !== program) {
```
deltakosh: same here we should centralize this into a function setProgram

deltakosh:
Awesome PR! Thank you very much for it!

src/babylon.engine.ts
benaadams: Done

src/babylon.engine.ts
benaadams: Done

src/babylon.engine.ts
benaadams: Done

deltakosh:
Cool! I'll merge it as soon as 2.4 is out

jbousquie:
can't wait for it
huge work :-)

deltakosh:
Ok I'm gonna merge it soon. Can you just update the preview what's new

benaadams:
Updated what's new

PR #1210, opened 2016-06-17
Null Quaternion fix

benaadams:
Fix to issue pointed out in https://github.com/BabylonJS/Babylon.js/commit/bddbd42593058c414b8ea80722d067201266a602#commitcomment-17908278

### DIFF
--- a/src/Mesh/babylon.abstractMesh.ts
+++ b/src/Mesh/babylon.abstractMesh.ts
@@ -763,7 +763,7 @@
             var yaw = -Math.atan2(dv.z, dv.x) - Math.PI / 2;
             var len = Math.sqrt(dv.x * dv.x + dv.z * dv.z);
             var pitch = Math.atan2(dv.y, len);
-            Quaternion.RotationYawPitchRollToRef(yaw + yawCor, pitch + pitchCor, rollCor, this.rotationQuaternion);
+            this.rotationQuaternion = Quaternion.RotationYawPitchRollToRef(yaw + yawCor, pitch + pitchCor, rollCor, this.rotationQuaternion || new Quaternion());
         }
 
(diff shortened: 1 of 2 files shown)


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -763,7 +763,7 @@
             var yaw = -Math.atan2(dv.z, dv.x) - Math.PI / 2;
             var len = Math.sqrt(dv.x * dv.x + dv.z * dv.z);
             var pitch = Math.atan2(dv.y, len);
-            Quaternion.RotationYawPitchRollToRef(yaw + yawCor, pitch + pitchCor, rollCor, this.rotationQuaternion);
+            this.rotationQuaternion = Quaternion.RotationYawPitchRollToRef(yaw + yawCor, pitch + pitchCor, rollCor, this.rotationQuaternion || new Quaternion());
```
RaananW: Quaternion.RotationYawPitchRollToRef doesn't return the quaternion. It only copies to the reference quaternion.
better would be to set the quaternion of the parent, if not set before. 
for example:
```javascript
this.rotationQuaternion = this.rotationQuaternion || new Quaternion() 
```

benaadams: hmm it does return it; maybe behaviour is wrong tho?

RaananW: Hey... that's totally new :) not even updated on my local repository yet.
This should work then, apologies.

RaananW: Having said that - I do think that ToRef functions should not return anything. Might be a bit confusing. But it was merged, so it was also reviewed. 
I think the best solution would be to create the rotationQuaternion beforehand.

benaadams: Fixed

PR #1211, opened 2016-06-17
Update what's new.md

jbousquie:


nockawa:
did you get the latest version of the file before doing your PR?

RaananW:
@nockawa  - before merging, why do you ask that? is there something missing? seems like nothing was deleted.

nockawa:
I believe the text was added in the bug fix section of the Canvas2D and not of the "main bug fix" section.

PR #1214, opened 2016-06-20
Tools and rendering improvement with PerfCounter

nockawa:
Tools:
 - PerfCount class added. records min, max, current, average, last second average of a monitored counter

Scene:
 - xxxDuration are now using PerfCounter, one property is added for each to access the underlying PerfCounter
 - getxXXDuration() has still the same behavior as before.


nockawa:
@deltakosh this is ready for review and merge. I wait for your comments/changes if required.

PR #1228, opened 2016-06-23
Added 'panningMouseButton' parameter in ArcRotateCamera class

fcicconeto:
I would like to use the middle mouse button to drag the camera (pan feature).

So I added an optional parameter named 'panningMouseButton' to the 'attachControl' method in 'ArcRotateCamera' class.


deltakosh:
Thanks!

PR #1229, opened 2016-06-24
fixed missing instance parameter in Mesh.CreateDashedLines

jbousquie:
everything is in the title


jbousquie:
seems to have not been merged

PR #1251, opened 2016-07-08
Bbox update for ribbon based shape updates

jbousquie:
The Bbox of a ribbon based shape (ribbon, tubes, extrusions, etc) is now updated when the shape is dynamically morphed.
Note : for performance reasons, the min and max computations are done in the same loop than the position computations. So no call to refreshBoundingInfo and no extra iteration.


RaananW:
Hi Jerome,
Quick question - why are there so many file changes?

jbousquie:
WTF... I just commited **one** ts file in VS Code (new version) and those compiled js files came along with... I did not even noticed them, tsssss

jbousquie:
do I need to redo everything ?

RaananW:
It seems like js file were added to the texture library. Need to check if their names are correct. On mobile, give me an hour.

jbousquie:
well, I modified just MeshBuilder, so the rest comes from the repo, is compiled here and goes back (not intentionally) to the repo. Nothing is added or removed by me (I hope)

jbousquie:
not sure to understand
let me recap :
I close my branch
I open new one
I modify the ts only,
I commit first
then I compile and test
If ok, then PR

right ?

RaananW:
Sounds like a great plan! 
Thanks. I'll see how to solve this case problem with those files. A git rename will probably do the trick.

RaananW:
Ohh, and of course, don't commit again after building.

PR #1261, opened 2016-07-10
Fix WebGl Inspector Main Tab

sebavan:


src/babylon.engine.ts
```diff
@@ -2739,6 +2734,9 @@
 
             // extension.
             var ext = gl.getExtension(extension);
+            if (ext === null) {
```
RaananW: Just wondering - why not if(!ext) ?

PR #1262, opened 2016-07-14
SPS billboard doesn't use ComputeNormals() any longer

jbousquie:
SPS billboard now correctly rotates the normals in the same time than the particle positions.
This means the billboard mode doesn't use `ComputeNormals()` any longer and thus doesn't need to re-iterate over the huge number of vertices/indices/normals.

In other terms, the SPS billboard mode is now as fast as the standard SPS mode.
This makes the SPS a good option for planar billboarded particles what would need the light reflection compared to the BJS standard particle system.


jbousquie:
Just added a modification in the way the billboard is computed : it now takes in account the camera direction of looking and not only the cam-mesh axis.
This will be a better expected result with FreeCameras when they don't look at the SPS but some particles enter their frustum.

PR #1267, opened 2016-07-22
Panning control fix

haxiomic:
Fixed issue where panning could get stuck on if ctrl was released when pointer outside window


haxiomic:
Same issue with mouse up/down

PR #1269, opened 2016-07-25
Ground getHeightAtCoordinates debug

PierreHei:
Vector3.CrossToRef function was modifying Tmp.Vector3[0]'s value to store its result. But GroundMesh.prototype._computeHeightQuads was both using Tmp.Vector3[0]'s value, and calling Vector3.CrossToRef, resulting in an undesired value in Tmp.Vector3[0] and a bug.

### DIFF
--- a/src/Math/babylon.math.js
+++ b/src/Math/babylon.math.js
@@ -899,10 +899,10 @@ var BABYLON;
             return result;
         };
         Vector3.CrossToRef = function (left, right, result) {
-            Tmp.Vector3[0].x = left.y * right.z - left.z * right.y;
-            Tmp.Vector3[0].y = left.z * right.x - left.x * right.z;
-            Tmp.Vector3[0].z = left.x * right.y - left.y * right.x;
-            result.copyFrom(Tmp.Vector3[0]);
+            var x = left.y * right.z - left.z * right.y;
+            var y = left.z * right.x - left.x * right.z;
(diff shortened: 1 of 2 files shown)


RaananW:
nice catch :)

PierreHei:
Thanks !

jbousquie:
I don't get where computeHeightQuads is still using a conflicting 
Tmp.Vector3[0] ?

yep, they both used once the same tmp vector, but not a conflicting way 
imho (i fixed this months ago)

PierreHei:
In computeHeightQuads, v1 is Tmp.Vector3[0], then crossToRef modified Tmp.Vector3[0] to calculate its result, modifying v1 by the way, but v1 is still used after call to crossToRef !

jbousquie:
nope

this was fixed long ago : 
https://github.com/BabylonJS/Babylon.js/blob/master/src/Mesh/babylon.groundMesh.ts#L139

v1 is Tmp.Vector3[3], not [0]

are you sure that you're using the latest source ?

PierreHei:
Clearly you're right, I saw this on 2.4 release and only modified Vector3.crossToRef source to fix, thus I didn't saw there already was a fix in Mesh.babylon.groundMesh. Sorry for losing your time !

jbousquie:
no worry :-)

PR #1271, opened 2016-07-25
resetToCurrentRotation to device orientation cameras

RaananW:
The idea is to allow a developer of device-oriented scenes to reset the rotation in one direction (defaults to the rotation around the y axis).
Good when you don't want the user to search for the interesting parts of the scene after initializing the scene.

Unneeded files were also removed.


Tools/Gulp/config.json
```diff
@@ -48,7 +48,6 @@
       "../../src/Cameras/Inputs/babylon.freecamera.input.keyboard.js",
       "../../src/Cameras/Inputs/babylon.freecamera.input.touch.js",
       "../../src/Cameras/Inputs/babylon.freecamera.input.deviceorientation.js",
-      "../../src/Cameras/Inputs/babylon.freecamera.input.vrdeviceorientation.js",
```
deltakosh: No need because you are only using babylon.freecamera.input.deviceorientation?

src/Cameras/babylon.deviceOrientationCamera.ts
```diff
@@ -1,22 +1,10 @@
 module BABYLON {
     // We're mainly based on the logic defined into the FreeCamera code
     export class DeviceOrientationCamera extends FreeCamera {
-        //-- Begin properties for backward compatibility for inputs
```
deltakosh: this may cause backward compat issues no?

RaananW: The only thing it might cause is missing functions in older VR DO Cameras, but shouldn't really be a problem. The constructor is the same, everything else as well.

Tools/Gulp/config.json
RaananW: exactly. They use the same function, so it was redundant. It is already like that in 2.4, I left it just in case there would be problems. but there weren't :smile:

src/Cameras/babylon.deviceOrientationCamera.ts
RaananW: Ah! Sory, I was commenting on something different.
The function exist in FreeCamera, and only return values in case mouse / keyboard is attached. the setters are only set when there is a mouse/keyboard. So I think this is actually better this way, because this way you could have keyboard movement and device orientation in a nicer way.

deltakosh: perfect

PR #1272, opened 2016-07-25
Same initial quaternion rotation, but for WebVR

RaananW:
Different implementation of the same concept.


deltakosh:
What are these files?

RaananW:
holy ****.
Give me a sec. what the heck is that....

PR #1291, opened 2016-08-03
Fixed some bugs in LoadFile

haxiomic:
Two bugs fixed:

**HTTP errors being considered as a successful response**
Success was conditional on HTTP code 200 *OR* `ValidateXHRData()` returning true. When dataType is 1 (text) ValidateXHRData returns true if the response is a non-empty string. Often when the request fails the response text will be set to some HTTP status error message (like "404 Not found"), causing ValidateXHRData to return true and LoadFile to execute the callback.

I've removed the ValidateXHRData check from LoadFile because more generally we shouldn't want LoadFile to validate the data at all, just load it - then the user of LoadFile can validate the data based on their own criteria (e,g. whether or not it's a valid TGA)

I've also allowed it to accept all HTTP success codes (2XX)

**Callback being called twice in rare cases**
onreadystatechanged may be called multiple times with the value 4 (XMLHttpRequest.DONE), https://bugs.chromium.org/p/chromium/issues/detail?id=162837  nullifying it after it occurs once prevents this

### DIFF
--- a/src/Tools/babylon.tools.ts
+++ b/src/Tools/babylon.tools.ts
@@ -394,7 +394,7 @@
                 request.onprogress = progressCallBack;
 
                 request.onreadystatechange = () => {
-                    if (request.readyState === XMLHttpRequest.DONE) {
+                    if (request.readyState === 4) {
                         request.onreadystatechange = null;//some browsers have issues where onreadystatechange can be called multiple times with the same value
 
                         if (request.status >= 200 && request.status < 300) {
--- a/src/Tools/babylon.tools.ts
+++ b/src/Tools/babylon.tools.ts
@@ -397,7 +397,7 @@
                     if (request.readyState === 4) {
                         request.onreadystatechange = null;//some browsers have issues where onreadystatechange can be called multiple times with the same value
 
-                        if (request.status >= 200 && request.status < 300) {
+                        if (request.status >= 200 && request.status < 300 || (navigator.isCocoonJS && (request.status === 0))) {
                             callback(!useArrayBuffer ? request.responseText : request.response);
                         } else { // Failed
                             if (onError) {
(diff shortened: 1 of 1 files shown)


src/Tools/babylon.tools.ts
```diff
@@ -395,7 +395,9 @@
 
                 request.onreadystatechange = () => {
                     if (request.readyState === 4) {
-                        if (request.status === 200 || Tools.ValidateXHRData(request, !useArrayBuffer ? 1 : 6)) {
+                        request.onreadystatechange = null;//some browsers have issues where onreadystatechange can be called multiple times with the same value
+
+                        if (request.status >= 200 && request.status < 300) {
```
deltakosh: unfortunately you still need Tools.ValidateXHRData as specific browsers or envrionment (like Cocoon) return a status ==0

haxiomic: Would it not be better to handle the special case for Cocoon (ie `|| (navigator.isCocoonJS && (request.status === 0))`)? At present it's broken in all browsers

haxiomic:
ValidateXHRData will return true if you pass it a string, passing it the response of the http request means it'll almost always return true, irrespective of if whether or not that HTTP request succeeded of failed. So when a http request fails, it's not uncommon the responseText will be something like "404 Not found".

This was causing issues for me when the Babylon was using LoadFile to load shaders, if the file didn't exist it'd load up the shader glsl code as "404 Not found". It's ok for the file not to be found, as long as the error callback is called so it can be handled correctly.

The change should not affect other uses ValidateXHRData. In LoadFile, ValidateXHRData only uses cases 1 and 6, case 6 doesn't exist in ValidateXHRData anyway so it uses ValidateXHRData for case 1. Case 1 is the broken string case I explained earlier.

deltakosh:
Gotcha!

PR #1300, opened 2016-08-08
Fixed issues with vertex attribute arrays

haxiomic:
This change shouldn't affect the existing behavior of Bablyon other than to fix edge cases that were unaccounted for

**The TL;DR of this PR is:**
This makes Babylon enable and disable vertex attribute arrays as they're used rather than only when an effect is enabled. Before this change, when rendering a mesh/set of vertex buffers, if the vertex buffers don't match with the effect attributes (eg, one isn't supplied) then the vertex data it likely to be incorrect and WebGL errors may occur.

--------

### Background
In WebGL, vertex attributes are global state and can either be a constant value or a pointer to an array, the state could be represented with a table like this:

```
Attribute Slot | isArrayMode | Constant Value (vec4) | Pointer to Array
-------------------------------------------------------------------
0              | true        | 0, 0, 0, 1            | [ArrayBuffer]
1              | false       | 0, 0, 0, 1            | [ArrayBuffer]
2              | false       | 0, 0, 0, 1            | unset
3              | false       | 0, 0, 0, 1            | unset
 ...
```

Is when isArrayMode is `true`, the vertex shader uses the pointer to the array for the attribute, otherwise it just uses the constant value. (The constant values are stored as four components, but if the attribute type uses less than four, like in a vec2, then a subset of the values will be used).

`enableVertexAttribArrray(slot)` and `disableVertexAttribArray(slot)` toggles isArrayMode `true` and `false`

The `vertexAttrib[1,2,3,4]f(slot, value)` and `vertexAttrib[1,2,3,4]fv(slot, values)` functions let you change the constant values, but the default for each slot is `0, 0, 0, 1`.

`vertexAttribPointer(slot, ...)` lets you specify the pointer-to-array for a particular slot (along with packing information)

### The affect of this PR
When rendering a mesh, if it's missing a particular attribute, (like UVs), executing `disableVertexAttribArray` on the attribute will set it so that is uses the constant value for each vertex instead of an array. Prior to this PR, Babylon was not enabling/disabling as required, so when a mesh was missing a particular attribute it'd just use whatever was previously set for that slot. If the slot was last used in array mode, and the currently set array is shorter than the number of vertices, an `INVALID_OPERATION` is likely to occur because there will be out of range access on the array (which was the issue we had), this meant the mesh would not be rendered! If an error doesn't occur, the junk-data may unexpectedly affect rendering.

This issue hasn't previously shown up because the built-in materials in Babylon tend to use defines to disable attributes in the shader (so a new effect is used for each set of attributes), however in our project we're using custom materials that don't do this (for good reasons), and so vertex array modes need to be managed correctly.

### Further work
- It may be worth adding an interface to effects to let them specify constant values in case the default `0, 0, 0, 1` doesn't suit them
- `MAX_VERTEX_ATTRIBS` is not handled gracefully (irrespective of this PR). WebGL stats has pretty much all devices on 15 attributes and is rare in practice to exceed this so there's no pressing need for this but it's a nice-to-have

### DIFF
--- a/src/babylon.engine.ts
+++ b/src/babylon.engine.ts
@@ -166,6 +166,7 @@
         public maxTextureSize: number;
         public maxCubemapTextureSize: number;
         public maxRenderTextureSize: number;
+        public maxVertexAttribs: number;
         public standardDerivatives: boolean;
         public s3tc: WEBGL_compressed_texture_s3tc;
         public textureFloat: boolean;
@@ -362,8 +363,7 @@
         private _currentEffect: Effect;
         private _currentProgram: WebGLProgram;
         private _compiledEffects = {};
-        private _vertexAttribArraysEnabled: boolean[];
-        private _vertexAttribArraysToUse: boolean[];
+        private _vertexAttribArraysEnabled: boolean[] = [];
         private _cachedViewport: Viewport;
         private _cachedVertexBuffers: any;
         private _cachedIndexBuffer: WebGLBuffer;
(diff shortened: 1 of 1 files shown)


jbousquie:
I like such teaching (for me) PR :-D

PR #1302, opened 2016-08-11
tga/dds file load arguments bug.

wyklion:
tga/dds file load arguments bug


wyklion:
tga/dds file load arguments bug

PR #1309, opened 2016-08-17
Fixing materials library (shadows vertex)

julien-moreau:


Masterchoc:
Hi,

it look like the problem just appear again, i gave further description here : 
http://www.html5gamedevs.com/topic/25420-terrain-material-shadows/

PR #1317, opened 2016-08-20
Webvr new api

RaananW:
The engine is now responsible to search for new devices, if WebVR is
enabled in the browser.
If WebVR is enabled (and a device was chosen (attachControl), the
fullscreen and request animation frame will be triggered by the
vrDisplay device.
Users with no WebVR enabled can still use the API in the same way.


RaananW:
Wait with this, I have 2 more improvements. :smile:

deltakosh:
Please udpate what's new then :)

PR #1322, opened 2016-08-22
Solving a problem when initializing two textures using the same WebGL texture

RaananW:
if the webgl teture is not yet ready, but exists in the cache, the
onLoad function  will be called when the texture is not yet ready.

The signature of an internal function in the engine was changed, but it cannot be externally used. It is fully backwards compatible.

### DIFF
--- a/src/babylon.engine.ts
+++ b/src/babylon.engine.ts
@@ -58,7 +58,7 @@
     }
 
     var prepareWebGLTexture = (texture: WebGLTexture, gl: WebGLRenderingContext, scene: Scene, width: number, height: number, invertY: boolean, noMipmap: boolean, isCompressed: boolean,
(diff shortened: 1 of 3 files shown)


src/babylon.engine.ts
```diff
@@ -88,9 +88,9 @@
         engine.resetTextureCache();
         scene._removePendingData(texture);
 
-        if (onLoad) {
-            onLoad();
-        }
+        texture.onLoadedCallbacks.forEach(function (callback) {
+            callback();
```
nockawa: Should it be a good place to splice the onLoadedCallbacks array to clean things up?

RaananW: More than possible. I can simply clear the array once it's done.

PR #1336, opened 2016-09-04
Add package.json for npm distribution (+typings)

enricostara:


enricostara:
Add also 'typings' attribute in order to import into typescript project as npm module

RaananW:
Hi,

Thanks! But we already have one. In tools/gulp. 
If you want to change it (if you feel a change is needed), please submit a PR. 
Ooh, and there are 89 contributes to the framework, not just one.

enricostara:
The tools/gulp package.json seems to be useful only for gulp dependencies and not to correctly identify  a first class node/npm project.

The project should have a general package.json  for node/npm configuration in the root folder (not mandatory), as already done for bower, in order to accomplish a correct npm publication in the "node world" (and also set correctly the 'typings' attribute for typescript).

btw, where can I submit a "PR"?

thanks

RaananW:
We do have an NPM package, https://www.npmjs.com/package/babylonjs
Which works rather well. I do agree that it might need modifications, but this would be the place to look.
I understand that if you need to use the latest build with npm  it might present a problem. Let's see how we solve this. But having two package.json files in one project is also wrong.

And PR is a pull request, which you already submitted.

enricostara:
I saw the npm package in https://www.npmjs.com/package/babylonjs, but I didn't find the package.json file in the project that produces it! The regular position for a such file should be in the root project folder... I tried to use this npm dependency but the "typings" attribute is missing, it's a shame..
Therefore, what I already pulled as request is a package.json that produces an npm publication like the https://www.npmjs.com/package/babylonjs, + typings attribute

PR #1338, opened 2016-09-04
waterMaterial to support .useLogarithmicDepth + more physically realistic

BitOfGold:


julien-moreau:
Hey, thanks for your contribution !
I have one question, why do you apply the color blend factor only on the refractive color ? According to this documentation https://doc.babylonjs.com/extensions/Water, the color blend factor is used to simulate the deep water (mixed with the water color). Why don't apply the water color also on reflective color ?

BitOfGold:
The original color blend factor was directly blended the water color to the combined reflection+refraction.
this meant it was blending from a totally transparent water to a solid color flat surface. and after that, lighted this flat-color surface with all the lights. It works only if you use a transparent water, and no colored lighting. (so, it does not work if it is not zero :) )

`baseColor = colorBlendFactor * waterColor + (1.0 - colorBlendFactor) * combinedColor;`

**water color** (color of the water medium) and lights (the refracted, and scattered lights underwater) only affect the refracted (underwater) part of the light. (except multiple reflections like an thin oil surface). The reflected light is just a reflection.

The other thing is the wave normals and **fresnelTerm**. The original shader calculated it as linear (like total reflection if you look perpendicular, total reflection if you look parallel, which is infinitely far from the viewer). If you perturbate this term with the wave normal too (and not just perturbate the reflected and refracted textures), improves the look significantly, waves start to look like waves. reflected light (like a sunset) starts to make dancing waves!

`vec3 normalW = normalize(vNormalW + vec3(perturbation.x*3.0,perturbation.y*3.0,0.0));`


Normals are also perturbated with a little octave of the same bump map, a lot of shaders I watched on _shadertoy_ uses this. Without normal (or height) perturbation of different octaves, the movement looks like a _conveyor belt_.

```
/smaller bumps superimposed (better moving waves, no "conveyor belt" look):
	baseColor = (texture2D(normalSampler, vNormalUV) + texture2D(normalSampler,vec2(-vNormalUV.y*0.33,vNormalUV.x*0.33)))/2.0;
```


The other thing is the fresnel term calculation.

`float fresnelTerm = max(dot(eyeVector, upVector), 0.0);`

Of course it works at some level :), but much of the reason a water rendering looks water, is the fresnel equation, which is not linear.

[](https://en.wikipedia.org/wiki/Reflectance#/media/File:Water_reflectivity.jpg)

It can be approximated with a simple exponential equation like this:
`float fresnelTerm = min(0.95, pow(max(dot(eyeVector, upVector), 0.0),1.5));`

[](http://www.scratchapixel.com/old/lessons/3d-basic-lessons/lesson-15-introduction-to-shading-and-radiometry/materials/)

but you can find many different versions. If you look at the difference between totally linear and somewhat exponential, you notice it all at once: with nonlinear, water that is far away, starts to make those outstanding waves, almost like noise. It looks like moving water now! Too much linearity makes it too noisy, of course it is subjective. All are jut approximations.

The only thing left which was too much to understand for me, is the shallow water calculation.
The water color works like a fog underwater, shallow water is more transparent, deep water is more colored.

I think this looks way better than the original.

Yay, sorry for too much text. :)

BitOfGold:
You gave me an idea and now I understand what that documentation writes about deep water.
To simulate deep water, at far distances you have to modify reflective color too (like the sea is darker at far distance than the sky)
another color and another colorblend property solved this problem, and also solved my shallow water problem:

now I can make the water greenish brown at a near distance and darker blue at a far distance.
It looks even better now :+1: 

take a look at
[before after picures](https://www.dropbox.com/sh/4t4r5d0l69v8jsq/AAC30ZOsiq_4_eBhjdk0w09ma?dl=0)

julien-moreau:
The pictures look really great !
I'll try and let you know if we are able to merge and avoid breaking changes (visual breaking changes also count) :)

BitOfGold:
if you think it is mandatory to keep the visuals from any change, even if it looks better (I understand that it is subjective, but what makes it a breaking change, if it brakes the visuals (not working) or makes any subtle visual change? ), in that case I have to make the changes optional.

Like something this:
waterMaterial.fresnelMode = FRESNEL_DEFAULT
(.waterColor affect both reflected and refracted light)

waterMaterial.fresnelMode = FRESNEL_SEPARATE
.waterColor and .colorBlendFactor affects only refracted (underwater) light
.waterColor2 and .colorBlendFactor2 affects only reflected (mirrored) light

waterMaterial.bumpSuperimpose = false by default
if true, smaller waves are superimposed on the bump map, to look less solid.

waterMaterial.bumpAffectsReflection = false by default
if true, bump map also perturbes the direction of 

the specular, more concentrated glare can be achieved with supporting.sepcularPower, like on the standard material. This would be default to 1 on waterMaterial.

This way, if you do not set any of the new parameters, the look should stay the same as before.
I can do it in some days if I find an hour. I think it should be done. :)

BitOfGold:
I don't really know why changes from my other PR (lightmapsexcluded) are here. . :-O

deltakosh:
This is weird.

BitOfGold:
some new changes, but I think I have to close this PR and start from new (new fork).
I cannot revert theese weird changes, I'm sure it is because I'm novice in github.
I can diff the changes again.

To show you the progress, here is a little demo:
https://www.bitofgold.com/ocean/

You can check off the renderMode, then it reverts back to the default settings.

julien-moreau:
@BitOfGold your demo is awesome :)
In your future pull request, if the default values allow to keep the default behavior of the material, everything will be ok. We can see your awesome additions as advanced features (explained in the doc in future) !
Once you get clean pull request (only water material) we'll be able to merge it sure :)
Thanks again !

PR #1339, opened 2016-09-05
added subdivisionsX and subdivisionsY options to groundMesh

abow:
http://www.html5gamedevs.com/topic/24962-working-on-subdivisionz-on-groundmesh/

### DIFF
--- a/src/Mesh/babylon.groundMesh.ts
+++ b/src/Mesh/babylon.groundMesh.ts
@@ -5,6 +5,8 @@
         private _worldInverse = new Matrix();
         private _heightQuads: { slope: Vector2; facet1: Vector4; facet2: Vector4 }[];
         public _subdivisions: number;
+        public _subdivisionsX: number;
+        public _subdivisionsY: number;
         public _width: number;
         public _height: number;
         public _minX: number;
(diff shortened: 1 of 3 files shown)


src/Mesh/babylon.groundMesh.ts
```diff
@@ -5,6 +5,8 @@
         private _worldInverse = new Matrix();
         private _heightQuads: { slope: Vector2; facet1: Vector4; facet2: Vector4 }[];
         public _subdivisions: number;
```
RaananW: isn't subdivisions redundant? It is actually not needed anymore (here, I am not talking about the meshBuilder).

src/Mesh/babylon.groundMesh.ts
```diff
@@ -20,6 +22,14 @@
             return this._subdivisions;
```
RaananW: If subdivionX and Y are defined, this will return the wrong result (Same as my last comment)

src/Mesh/babylon.groundMesh.ts
abow: It does seem to be redundant.  I didn't dare to remove it, since I thought it could break something (public var).

RaananW: The public getter can stay, the private can be removed. The public will anyhow return a wrong value, if x != y. Return math.min(x,y), this will be the safest.

PR #1340, opened 2016-09-05
Add module definition in babylon.2.4.d.ts

enricostara:
Add module definition in babylon.2.4.d.ts in order correctly import 'babylonjs' as module:

`import * as BABYLON from "babylonjs" `

Remove "contributors" from package.json

### DIFF
--- a/dist/babylon.2.4.d.ts
+++ b/dist/babylon.2.4.d.ts
@@ -13715,3 +13715,7 @@ declare module BABYLON {
         dispose(): void;
     }
 }
+
+declare module "babylonjs" {
+    export = BABYLON;
+}
--- a/package.json
+++ b/package.json
@@ -1,0 +1,53 @@
+{
+  "author": {
+    "name": "David CATUHE"
+  },
+  "bugs": {
+    "url": "https://github.com/BabylonJS/Babylon.js/issues"
+  },
+  "contributors": [
+    {
+      "name": "Enrico Stara"
+    }
+  ],
+  "description": "Babylon.js is a JavaScript 3D engine based on webgl (+typings)",
+  "directories": {},
+  "engines": {
+    "node": "*"
+  },
+  "files": [
+    "dist/babylon.2.4.d.ts",
+    "dist/babylon.2.4.js",
+    "dist/babylon.2.4.max.js",
+    "dist/babylon.2.4.noworker.js",
+    "dist/babylon.2.4.core.js",
+    "dist/Oimo.js",
+    "package.json"
+  ],
+  "homepage": "https://github.com/BabylonJS/Babylon.js#readme",
+  "keywords": [
+    "3D",
+    "javascript",
+    "html5",
+    "webgl",
+    "typescript"
+  ],
+  "license": "Apache-2.0",
+  "main": "dist/babylon.2.4.js",
+  "maintainers": [
+    {
+      "name": "babylonjs",
+      "email": "contact@babylonjs.com"
+    }
+  ],
+  "name": "babylonjs",
+  "optionalDependencies": {},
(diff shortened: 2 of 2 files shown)


RaananW:
Hi enrico,

I think we had this conversation already :)
What bothered me wasn't the contributors, this was just an anecdote. The problem was, that we already have an NPM package and a package.json. I wouldn't want to have a 2nd one. You can change the existing one.

One more note - the .d.ts file is being generated automatically, making changes there won't really help, since it will eventually be overwritten at the next build (if 2.4 will ever be built again).

I am not entirely sure what the problem with the location of package.json, especially if you can use both npm and bower to load babylon as a dependency. 
Want to elaborate a bit more?

enricostara:
Your NPM package is almost OK, only the "typings" attribute is missing in package.json provided with npm installation:
https://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html

The problem is that I don't find the "package.json" in the https://github.com/BabylonJS/Babylon.js that is used to publish the npm project.. please, let me know the path (for sure not one in tools/gulp), and I'll happy to modify it :)

I know that xxx.d.ts is generated, but the few added rows (at the end of the file) are mandatory to correctly import the project as module:
https://www.typescriptlang.org/docs/handbook/writing-declaration-files.html

RaananW:
Why is it not the one in tools/gulp? This is the right one to change. 

regarding the module loading - babylon declares the BABYLON module, which can be imported. Since you are being very consistent about it I can only assume I am not seeing something, but I am not sure what it is :smile:

enricostara:
Hi Raanan,

the Tools/Gulp/package.json file seems added by contributors only to provide a suitable "automation-tool" to compile typescript. 

What the project now  needs is a "nodejs" identity that can be only provided by a package.json in the root folder, as already done on the "bower" side (for this reason you can find a bower.json in the root foder and not in a Tools/ subfolder). 

A totally  new "package.json" file in the root folder could provide all the information to indentify BabylonJS in the node/npm world:
https://docs.npmjs.com/files/package.json

All the files needed to distribute the project with npm will be relative to package.json (without ../../) and all the files usually in the root folder (like README, LICENSE, etc)  are always included and expected as siblings:
https://docs.npmjs.com/files/package.json#files

deltakosh:
Hey @enricostara 
Until now I was managing npm on my own.
I can publish the package.json here if you want (in a tools/npm folder)

enricostara:
Hi @deltakosh 

better  there than nothing :)

enricostara:
Hi @deltakosh 

first I tried to launch:

`npm pack Tools/Npm`

and it's not working... 

then I  modified the Tools/Npm/package.json relative file paths  and names (in "files" attribute) to point dist/ files one by one, but still didn't work for me.. 
Seems that, for example,  "../../dist/babylon.2.4.max.js" is not working using npm pack command (and I suppose the same for npm publish)

Finally, if you try a test moving the dist/ folder into Tools/Npm and so removing "../../" from paths then it works!.. Seems that "../../" was a problem here and I cannot figure out the reason

enricostara:
I found this:
https://docs.npmjs.com/files/folders#more-information

> Starting at the $PWD, npm will walk up the folder tree checking for a folder that contains either a package.json file, or a node_modules folder. If such a thing is found, then that is treated as the effective "current directory" for the purpose of running **npm commands**. (This behavior is inspired by and similar to git's .git-folder seeking logic when running git commands in a working dir.)

could be the reason why  relatives path that include package.json's  parent folders are ignored..

deltakosh:
Don't spend too much time on it:)
I copied the file here if you want to update it but when it will come to update npm I will get it and use it in my own folder with all files side by side

PR #1343, opened 2016-09-05
Combining baked and dynamic shadows nicely (lightmap)

BitOfGold:
https://www.bitofgold.com/lightmaptest/
A little demonstration to show what is wrong with the previous ligthmap calculation - no ambient lighting, no other light, no specular highlights.

Does not break the previous lightmap calculation.

material.lightmapTexture needs to be set to activate light map.

light.lightmapExcluded
if true, gets de diffuse lighting from the light map. adds specular highlights mixed widh baked and dynamic shadows.

if false, lightmap calculation is not affected. ("Old" lightmap, material.useLightmapAsShadowmap works as before (simply multiplying color.rgb after composition). Does not look right.)

light.lightmapNoSpecular
if true, does not even calculate the light functions, only the baked lightmap + dynamic shadow. (no specular then from this light, only the baked diffuse)


BitOfGold:
I commited to the wrong branch I think, mixed this with my previus pull request.
It's OK now.

deltakosh:
This is really cool
I would not use a double boolean as if you do not set ligtmapExcluded then noSpecular as no effect.

I would prefer some kind of mode enum:
* None
* LightmapExcluded
* LightmapExcludedNoSpecular

BitOfGold:
Thanks!
I'm on it.

BitOfGold:
Oh, working on two pull requests on two branches is too much for me...
Now **both looks fine**, but i do not now how to update these branches with the updates that correct the following error:

`babylon.js:23 Uncaught ReferenceError: SIMD is not defined`
It was an error that is now corrected. My 'master' branch is ok

I'm not an experienced github user, and I pull my hair out. If I update the 'lightmapsexcluded' branch from the 'master', why that error stays, I cannot comprehend... I see no other option than deleting my fork and starting over.
Sorry!

I hope theese 2 PG-s get in soon.

deltakosh:
No rush:)

src/Materials/babylon.materialHelper.ts
```diff
@@ -103,6 +104,20 @@
                     }
                 }
 
+                if (light.lightmapMode != 0 ) {
```
deltakosh: to ensure people know which value to use, I suggest also creating named values like here:
https://github.com/BabylonJS/Babylon.js/blob/master/src/babylon.engine.ts#L199

deltakosh: And this is where the naming will be important to help users understand each mode

deltakosh:
Good job!

nockawa:
great job! it deserves credits in the what's new, right?

BitOfGold:
I refreshed my little demonstration app:
[https://www.bitofgold.com/lightmaptest/](https://www.bitofgold.com/lightmaptest/)

deltakosh:
Excellent

GuDuJian-J-Zhang (commented):
have a question about this PR:
why lightmode is a pre light property
since we can specify only one lightmap:

// Create lightmap texture on a material
    var lightmap = new BABYLON.Texture("textures/candleopacity.png", scene);
    var groundMaterial = new BABYLON.StandardMaterial("groundMaterial", scene);
    groundMaterial.lightmapTexture = lightmap;

light.lightmapMode = BABYLON.Light.LIGHTMAP_SPECULAR;

BitOfGold:
It is per-light, because the property decides the behaviour of the light, not the material:

- default (calculates the light normally)
- specular (only calculates the specular lighting, this light is pre-baked with materials having a light map)
- shadows only (no light calculation on materials with a light map)

GuDuJian-J-Zhang:
> It is per-light, because the property decides the behaviour of the light, not the material:
> 
> * default (calculates the light normally)
> * specular (only calculates the specular lighting, this light is pre-baked with materials having a light map)
> * shadows only (no light calculation on materials with a light map)

thanks for your response. but there are two more questions if there are multiple lights in a scene.

1. how to specify multiple lightmaps for a material ?

2. how can i decide which light will be influenced by which lightmap?

PR #1344, opened 2016-09-06
Fixing the vr arc rotate camera (which doesn't have quaternions)

RaananW:


RaananW:
Note for the next version - move this function as a protected element in each VR camera, this way we avoid those extra "ifs".

PR #1348, opened 2016-09-13
waterMaterial supports logarithmic depth + separate fresnel color blending

BitOfGold:
- logarithmic depth support
- correcting uwrap/vwrap bug in rendertargets, causing weird artifacts
at the sides of the screen
- superinposing smaller bump waves for more dynamic waves
- separate refraction/reflection color blending with physically correct
(nonlinear) fresnel calculation


julien-moreau:
Hey, perfect ! Just tested and updated the materials library with your additions
The result is now awesome :)
Thanks for the contribution

PR #1354, opened 2016-09-17
lensFlareSystem .viewportBorder (lensflare from outside of the screen)

BitOfGold:
default = 0, no change in default behaviour.
if > 0, lensflares start to intensify when they are slightly outside of
the viewport (as they should I think, most real lens flares are visible
when the light is not yet in the view)


BitOfGold:
[Lensflare screenshots](https://www.dropbox.com/sh/uebtlpbd9h6do3g/AAA2UVTu6xrXxNW6YuJbfcaga)

deltakosh:
Brilliant

PR #1358, opened 2016-09-19
Add support for loading binary files using array buffers

garyhsums:


src/Loading/babylon.sceneLoader.ts
```diff
@@ -1,16 +1,27 @@
 ﻿module BABYLON {
+    export interface ISceneLoaderPluginExtensions {
+        [extension: string]: {
```
nockawa: Can you please explain me what are doing the lines 3, 4 and 5? I really can't figure this out! :)

julien-moreau: For example, glTF can have two extensions (.gltf and .glb) where .gltf is a simple JSON file, and glb a binary version of the JSON file containing binary data such as the vertex/indices buffers etc.

nockawa: it's more the syntax I don't get, what is a `[extension: string]: { isBinary: boolean; };` ?

garyhsums: Yes, each extension can now specify an isBinary flag. If false, it will load the file into a string as it did before. If true, it will load the file into an array buffer for easier binary parsing.

garyhsums: That just the definition of a hash table. It means the key is a string and the value is another hash table with a isBinary property.

garyhsums: http://stackoverflow.com/questions/21070401/how-does-the-hash-variable-syntax-work-in-typescript

nockawa: Ok, now I finally understand thanks to you SO link! thank you, I've learned something today! :)

src/Loading/babylon.sceneLoader.ts
```diff
@@ -71,34 +82,40 @@
             }
 
             var extension = sceneFilename.substring(dotPosition, queryStringPosition).toLowerCase();
+            return SceneLoader._registeredExtensions[extension];
```
deltakosh: what if the extension is unknown? We should return the first plugin in this case

garyhsums: okay, yeah, that was what it was doing before. if the extension is unknown, return the Babylon loader.

deltakosh:
206 files changed:). I think something went wrong

deltakosh:
Sounds good!

garyhsums:
Thanks!

PR #1366, opened 2016-09-24
sprite2D.scaleToSize(size)

albertbuchard:
Sets the scale of the sprite using a BABYLON.Size(w,h).
Keeps proportion by taking the maximum of the two scale for x and y.

If size is not set it waits for the texture to load.

### DIFF
--- a/src/Canvas2d/babylon.sprite2d.ts
+++ b/src/Canvas2d/babylon.sprite2d.ts
@@ -229,6 +229,31 @@
         }
 
         /**
+         * Sets the scale of the sprite using a BABYLON.Size(w,h).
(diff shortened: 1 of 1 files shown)


nockawa:
Tell me if I'm correct: this is a helper function which compute the scale to apply to the Sprite2d to get a given size in the Cavnas, independently of the size of the assigned Texture, right?

albertbuchard:
Yes, but if no size is set it takes the size of the texture as the base size. Like the sprite2D constructor.

nockawa:
Merged, thanks for the contribution ! it's the very first Pull Request for Canvas2D! :)

albertbuchard:
Ahah, I'm glad I could help a little ! :)

PR #1367, opened 2016-09-26
Run addModuleExports for the generated d.ts file.

rehmsen:
This apparently used to work, but no longer did. Context here:
http://www.html5gamedevs.com/topic/25425-using-babylonjs-npm-package-from-typescript-with-browserify/#comment-145494

### DIFF
--- a/Tools/Gulp/gulpfile.js
+++ b/Tools/Gulp/gulpfile.js
@@ -98,6 +98,7 @@ gulp.task('typescript-compile', function () {
     return merge2([
         tsResult.dts
             .pipe(concat(config.build.declarationFilename))
+            .pipe(addModuleExports("BABYLON"))
             .pipe(gulp.dest(config.build.outputDirectory)),
         tsResult.js
             .pipe(gulp.dest(config.build.srcOutputDirectory))


deltakosh:
We have it here:
https://github.com/BabylonJS/Babylon.js/blob/master/Tools/Gulp/gulpfile.js#L152
https://github.com/BabylonJS/Babylon.js/blob/master/Tools/Gulp/gulpfile.js#L172
https://github.com/BabylonJS/Babylon.js/blob/master/Tools/Gulp/gulpfile.js#L133

rehmsen:
I see that - but that does not seem to solve the problem - try it. I am not sure what the problem exactly is, maybe you are in a better position to answer that - but I have tried to run it and it was not working. As you see in the linked thread, this results in the package on npm not having the module.exports.

deltakosh:
For me it is working
Just have a look at all .js file generated here:
https://github.com/BabylonJS/Babylon.js/tree/master/dist/preview%20release

They all have the module exports

So this must be something else

rehmsen:
You are right, this code goes into the JS files. But that does not solve my problem, which is that I cannot require BABYLON from Typescript, because the d.ts file is not in module format. This PR is of course the wrong fix, so I am closing it and see if I cannot send you one to fix the d.ts.

PR #1368, opened 2016-09-27
Highlight Layer First step

sebavan:
First step on a glowy outline for meshes.

### DIFF
--- a/src/Layer/babylon.highlightlayer.js
+++ b/src/Layer/babylon.highlightlayer.js
@@ -1,0 +1,480 @@
+var __extends = (this && this.__extends) || function (d, b) {
+    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
+    function __() { this.constructor = d; }
+    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
+};
+var BABYLON;
+(function (BABYLON) {
+    /**
+     * Special Glow Blur post process only blurring the alpha channel
+     * It enforces keeping the most luminous color in the color channel.
+     */
+    var GlowBlurPostProcess = (function (_super) {
+        __extends(GlowBlurPostProcess, _super);
+        function GlowBlurPostProcess(name, direction, blurWidth, options, camera, samplingMode, engine, reusable) {
+            var _this = this;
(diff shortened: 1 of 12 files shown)


src/Mesh/babylon.meshBuilder.ts
```diff
@@ -1,6 +1,10 @@
 ﻿module BABYLON {
     export class MeshBuilder {
         private static updateSideOrientationForRightHandedSystem(orientation: number, scene: Scene): number {
+            if (orientation == Mesh.DOUBLESIDE) {
```
deltakosh: Thanks ;)

src/Layer/babylon.highlightlayer.js
```diff
@@ -0,0 +1,480 @@
...
+            this._blurTexture = new BABYLON.RenderTargetTexture("HighlightLayerBlurRTT", {
+                width: blurTextureWidth,
+                height: blurTextureHeight
+            }, this._scene, false, true, BABYLON.Engine.TEXTURETYPE_UNSIGNED_INT);
+            this._blurTexture.wrapU = BABYLON.Texture.CLAMP_ADDRESSMODE;
+            this._blurTexture.wrapV = BABYLON.Texture.CLAMP_ADDRESSMODE;
+            this._blurTexture.anisotropicFilteringLevel = 16;
+            this._blurTexture.updateSamplingMode(BABYLON.Texture.TRILINEAR_SAMPLINGMODE);
+            this._blurTexture.renderParticles = false;
+            this._downSamplePostprocess = new BABYLON.PassPostProcess("HighlightLayerPPP", this._options.blurTextureSizeRatio, null, BABYLON.Texture.BILINEAR_SAMPLINGMODE, this._scene.getEngine());
```
deltakosh: Question: Do you think you could have leveraged the PostProcess renderPipeline engine here?
https://github.com/BabylonJS/Babylon.js/blob/master/src/PostProcess/babylon.lensRenderingPipeline.ts

sebavan: Would have been way harder cause the postprocess pipeline is on the scene not the render target and the effect is more like a preprocess than a postprocess. The main scene is not renderered to a texture, the effect is. not sure I make much more sense but for the TL;DR, not really ;-)

PR #1372, opened 2016-09-28
Make d.ts file a module so that it can be imported

rehmsen:
Add "export = BABYLON;" to the end of d.ts file so that it can be imp…orted with an import statement.

I tried to use the same style as gulp-addModuleExports.js for consistency.
I slightly simplified the String/Buffer concatenation in gulp-addModuleExports.js. It still seems pretty complicated to use through2 here (as compared to other options like gulp-insert), but maybe there is some reason for it.


Tools/Gulp/gulp-addModuleExports.js
```diff
@@ -18,11 +18,11 @@ module.exports = function (varName) {
         '};\n';
 
         var decorateAddition =
-        'var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {\n' +
```
rehmsen: Sorry for this - my editor removed the windows carriage return - do you want me to revert that change? It seemed to be inconsistently used in the repo.

deltakosh: Please :)

rehmsen: Done.

deltakosh:
Ok last point (I should have not merged acutally). Not sure that export = BABYLON is the right syntax here

deltakosh:
To "import" it in your code why not using this:
/// <reference path="../babylon.d.ts" />

rehmsen:
If I do that, I have to load the JavaScript via a script tag. If it is
additionally a module, I can load both js and ts with an import statement.

Obviously both works, but the Typescript handbook recommends modules over
namespaces, particularly on Node (which npm belongs to):

"It is also worth noting that, for Node.js applications, modules are the
default and the recommended approach to structure your code.

Starting with ECMAScript 2015, modules are native part of the language, and
should be supported by all compliant engine implementations. Thus, for new
projects modules would be the recommended code organization "
https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html

I believe the export statement does no harm. I think you can still load it
using a reference tag. That is why lots of other libraries like Angular
also provide it.

deltakosh:
My concern is more about the "export = " string
Can you confirm with a doc or a blog post that this is the right one?

rehmsen:
Thanks for being diligent. I did some more testing and also looked at how other libraries like Angular handle the situation. Apparently I am indeed still missing something: `export as namespace BABYLON`

Here is documentation that I think fits well: We want to be able to use BabylonJS both with module loaders and without:
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions

I have tried that and it indeed allows us to load the module in these two ways:

    import * as BABYLON from 'babylonjs/babylon';

and
 
    /// <reference path="babylonjs/babylon.d.ts" />

I would prefer if it worked just with `babylonjs` instead of `babylonjs/babylon`, but I am not quite sure how to do that except by renaming the output of the Typescript compilation to index.js and index.d.ts, and its not a huge deal.

I will see if I can still edit the branch.

PR #1373, opened 2016-09-28
TypeScript 2.0 compilation compliance

nockawa:


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -374,7 +374,7 @@
                 this.rotation = Vector3.Zero();
             }
             var rotationQuaternion: Quaternion;
-            if (!space || space === Space.LOCAL) {
+            if (!space || (space as any) === Space.LOCAL) {
```
deltakosh: don't understand why we need this

src/Tools/babylon.database.ts
```diff
@@ -11,7 +11,7 @@ module BABYLON {
         private isSupported: boolean;
 
         // Handling various flavors of prefixed version of IndexedDB
-        private idbFactory = <IDBFactory> (window.indexedDB || window.mozIndexedDB || window.webkitIndexedDB || window.msIndexedDB);
+        private idbFactory = <IDBFactory> (window.indexedDB || window.mozIndexedDB || window.webkitIndexedDB || (window as any).msIndexedDB);
```
deltakosh: same here

src/Mesh/babylon.abstractMesh.ts
nockawa: honestly, I don't either, I've search the compilation error on google several minutes, but it just doesn't want to compile with the === operator. and there's nothing on google about it.
I also tried a const enum, it doesn't change...

src/Tools/babylon.database.ts
nockawa: the field is not recognized as part of the Window type....

deltakosh: we have the mixin.s for that

deltakosh:
The check is failing :(

src/Tools/babylon.database.ts
```diff
@@ -530,7 +530,7 @@ module BABYLON {
                                 try {
                                     //backwards compatibility with ts 1.0, srcElement doesn't have an "error" according to ts 1.3
                                     if (event.srcElement['error'] && event.srcElement['error'].name === "QuotaExceededError") {
-                                        this.hasReachedQuota = true;
+                                        (this as any).hasReachedQuota = true;
```
deltakosh: same here? why this?

src/Physics/Plugins/babylon.cannonJSPlugin.ts
```diff
@@ -116,7 +116,7 @@
                                 this.removePhysicsBody(childImpostor);
                                 childImpostor.physicsBody = null;
                             }
-                            childImpostor.parent = mainImpostor;
+                            (childImpostor as any).parent = mainImpostor;
```
deltakosh: why?

src/Mesh/babylon.abstractMesh.ts
deltakosh: why not "===" ?

src/Tools/babylon.database.ts
nockawa: if I replace the "function" by a lambda, it does compile.

src/Physics/Plugins/babylon.cannonJSPlugin.ts
nockawa: well, PhysicImpostor.parent doesn't have a setter! so unless the property setter is added at runtime, it's a bug! :)

deltakosh: This is not the way to fix it then :) I don;t want to do a fast fix
Ping @RaananW

src/Tools/babylon.database.ts
deltakosh: Ping @davrous

src/Physics/Plugins/babylon.cannonJSPlugin.ts
RaananW: let me check :smile:

src/Tools/babylon.database.ts
nockawa: it compile with a lambda, otherwise we can keep a function and declare "var self = this" outside the function and use "self" instead of this and it will also compile

deltakosh: There is something wrong here. We need to wait for @davrous comments

src/Canvas2d/babylon.canvas2d.ts
```diff
@@ -1688,7 +1688,15 @@
                 this._worldSpaceNode = settings.customWorldSpaceNode;
                 this.applyCachedTexture(null, null);
             }
+
+            this.propertyChanged.add((e, st) => {
```
deltakosh: this seems not related to TS..do you mind doing another PR for this?

nockawa: well, that was the original intend of the branch....which slightly derived.
if I do a separated PR it will be on uncompiled code (since i've got TS 2.0 now). but it's a copy/paste... ok...let's try! :)

deltakosh: This is just because I would like to merge this now but I cannot yet merge ts2 stuff

src/Tools/babylon.database.ts
davrous: They have probably removed msIndexedDB from their d.ts, we have to add it to our mixin indeed rather than casting it as any.

nockawa: yep, that's what I did

PR #1380, opened 2016-09-29
Support loading with a module loader.

rehmsen:
This is a new attempt of the failed and rolled back #1372.

Babylon.js currently uses namespaces and its types can be pulled in using

    /// <reference path="babylonjs/babylon.d.ts" />

It's generated JavaScript can be loaded using either a script tag, or using a module loader from JavaScript. However, what was not yet supported is loading types and implementation using a module loader from Typescript, because babylon.d.ts was not a module.

This commit adds support for that, following the example here:
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions

With it, it is additionally possible to load BabylonJS like this from Typescript:

    import * as BABYLON from 'babylonjs/babylon';

### DIFF
--- a/Tools/Gulp/gulp-addDtsExport.js
+++ b/Tools/Gulp/gulp-addDtsExport.js
@@ -1,0 +1,31 @@
+var gutil = require('gulp-util');
+var through = require('through2');
+
+module.exports = function (varName) {
+    return through.obj(function (file, enc, cb) {
+
+        var moduleExportsAddition =
+            '\nexport as namespace ' + varName + ';\n' +
+            '\nexport = ' + varName + ';\n';
+
+        if (file.isNull()) {
+            cb(null, file);
+            return;
+        }
+
+        if (file.isStream()) {
+            //streams not supported, no need for now.
+            return;
+        }
+
+        try {
+            file.contents = new Buffer(String(file.contents) + moduleExportsAddition);
+            this.push(file);
+
+        } catch (err) {
+            this.emit('error', new gutil.PluginError('gulp-add-module-exports', err, { fileName: file.path }));
+        }
+        cb();
+    });
+};
+
--- a/Tools/Gulp/gulpfile.js
+++ b/Tools/Gulp/gulpfile.js
@@ -3,6 +3,7 @@ var uglify = require("gulp-uglify");
 var typescript = require("gulp-typescript");
 var sourcemaps = require("gulp-sourcemaps");
 var srcToVariable = require("gulp-content-to-variable");
+var addDtsExport = require("./gulp-addDtsExport");
 var addModuleExports = require("./gulp-addModuleExports");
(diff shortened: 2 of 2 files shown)


deltakosh:
I still do not get why "export = BABYLON"

rehmsen:
I am not surprised. Typescripts modules and namespaces are actually pretty confusing I think. Maybe I did a bad job explaining. I will try again.

Typescript has [two ways of organizing code](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html), namespaces and modules. Namespaces used to be called internal modules, and that is what BabylonJS uses. Modules (used to be called external modules) have the advantage that they can be loaded using a module loader, and that they are equivalent to ES2015's modules. 

A file becomes a module when it has a top-level import or export statement (this is the same between ES2015 and Typescript). If you write 

    namespace BABYLON { // or the deprecated `module BABYLON` that you use
       export class SomeClass { ... } // this export is not top-level, so the file is not a module 
    }

your file is not a module. It is a file using namespaces. So one cannot use a module loader to load it. Adding `export = BABYLON;` outside of the namespace is a top-level export, so it makes it a module. (writing export namespace BABYLON { ... }might also work, I am not sure). It is documented [here](https://www.typescriptlang.org/docs/handbook/modules.html) in the section "Default exports". Note that according to my testing and contrary to what is described in the handbook (which is often slightly outdated), one can import modules created by `export =` not only with `import ... = require(...)` but also with the more standard `import ... from ...`.

So the problem with just `export = BABYLON;` was that it makes the file a module, and that now people that do not want to use a module loader no longer could reference the d.ts using a triple-slash references. This is where `export as namespace BABYLON` comes in. According to [this documentation](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions), this makes it possible to use it in global mode, despite it being a module.

Just from the documentation by itself, I think this combination still seems weird. So I checked out how other libraries are doing this. Here are the [types for AngularJS](https://registry.npmjs.org/@types/angular/-/angular-1.5.16.tgz) - I could only find the tar ball, no Github repo. They do the same.

Let me know what is still unclear. I am by no means an expert of this, but I think I am starting to understand, and hopefully we can come to a common understanding together.

rehmsen:
I should add that, the reason we go down the "adding `export = BABYLON;` in the d.ts build process"-route is that this is the equivalent to the "adding module.exports = BABYLON;` in the js build process"-route. The probably cleaner alternative would have been to add the appropriate code in the *.ts files, pass the right `--module` parameter during the compile and have Typescript compiler generated the appropriate code for both .js and .d.ts. 

That seemed the more intrusive change at this point though, so I did not even try that. But we can experiment with that also.

deltakosh:
Ok makes sense!

PR #1382, opened 2016-09-30
changing subdivisionsY to subdivisionsZ

abow:
When I first added subdivisionsY to the groundMesh I was looking at how you would subdivide a grid/plane in Blender.  It makes more sense to call it subdivisionsZ though, because the z axis in BJS is where the y axis is in Blender.

### DIFF
--- a/src/Mesh/babylon.groundMesh.ts
+++ b/src/Mesh/babylon.groundMesh.ts
@@ -6,7 +6,7 @@
         private _heightQuads: { slope: Vector2; facet1: Vector4; facet2: Vector4 }[];
         
         public _subdivisionsX: number;
-        public _subdivisionsY: number;
+        public _subdivisionsZ: number;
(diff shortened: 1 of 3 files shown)


RaananW:
Hi, thanks for the commit.

First, changing this is not backwards compatible, which will pose a problem. Second - x and y were chosen because a ground mesh is eventually a 2D object. It's not because of blender's axis system.

abow:
This was a recent feature, that only appears in preview, so I didn't think that backward compatibility would be an issue.  If subdivisionY makes sense to you, then it is good with me.

RaananW:
sorry! you are right about the backwards compatibility. 
I don't want to be the judge on Y vs Z, I would personally leave it Y. Let's wait for one more opinion :smile:

deltakosh:
Agree with Raanan:)

PR #1393, opened 2016-10-10
New property in Scene that contains all instances of canvas2d

Temechon:


src/babylon.scene.ts
```diff
@@ -339,6 +339,9 @@
         */
         public meshes = new Array<AbstractMesh>();
 
+        // Contain all canvas2D used in the scene
+        public canvas2ds = new Array<Canvas2D>();
+
```
nockawa: Please, no Canvas2D stuff in Scene.ts.
Instead you should rely on a public get Canvas(): Array<Canvas2D> in Canvas2D class with a private static class behind.

deltakosh: Yep Canvas2d is not in core so cannot have a strong reference,

dist/preview release/what's new.md
```diff
@@ -26,6 +26,7 @@
  - Text2D super sampling to enhance quality in World Space Canvas
  - World Space Canvas is now rendering in an adaptive way for its resolution to fit the on screen projected one to achieve a good rendering quality
  - Transparent Primitives are now drawn with Instanced Array when supported
+ - New property in Scene (canvas2ds) that contains all instances of canvas2d in this scene [Temechon](https://github.com/Temechon)
```
nockawa: not anymore! ;)

nockawa: Update the what's new and it's good to go.

Temechon: Already done !

PR #1396, opened 2016-10-10
Patch array iteration

vousk:
Chanded some array iterations for compatibility with web environnement where array prototype has been changed
cf. http://www.html5gamedevs.com/topic/25135-bad-array-iteration-habits-in-bjs/


deltakosh:
Cool! Thanks a lot

PR #1397, opened 2016-10-10
fix lost param in _ExtrudeShapeGeneric

vousk:
Fixed a forgotten param in the shape extrusion process


deltakosh:
you're on fire :)

vousk:
hi hi, and it's not finished, we'll probably have other array iteration to patch, I'll fix when I notice them

PR #1400, opened 2016-10-12
Gui lib: first commit still very early stage

nockawa:


nockawa:
The build failed because I used TypeScript 2.0 only feature (abstract declaration on property). I initially didn't want to commit that early and I forgot about it... I will remove the abstract for now.

PR #1407, opened 2016-10-13
Blender Exporter 5.0.6

Palmer-JC:
- Fixed NoAppliedTransforms check when mesh vertices > 65k


Palmer-JC:
Not sure why that travis thing failed,  my python code should not matter.

PR #1430, opened 2016-10-18
Fixing Vector4.toString() formatting

nockawa:
My biggest contribution so far! ;)


deltakosh:
Imrpessive :)

nockawa:
I know, I know, I was about to use string interpolation instead, but I revised my judgement...
On a more serious note, is this method called in the act of serialization?

deltakosh:
Nope..only for debugging purpose

PR #1432, opened 2016-10-19
Fix HighlightLayer crash when working with an InstancedMesh.

Swiftwork:
Added multiple checks to HighlightLayer to allow InstancedMesh:es. This by referring to their source rather than themselves when adding or removing.

### DIFF
--- a/canvas2D/gulp-addModuleExports.js
+++ b/canvas2D/gulp-addModuleExports.js
@@ -25,6 +25,11 @@ module.exports = function (varName) {
             'return c > 3 && r && Object.defineProperty(target, key, r), r;\n' +
         '};\n';
 
+        var dependencyAddition =
+        'if (typeof BABYLON === "undefined") {\n' +
+            'throw "BabylonJS is a required dependency, please include it first!"\n' +
+        '};\n';
+
         if (file.isNull()) {
             cb(null, file);
             return;
@@ -36,7 +41,7 @@ module.exports = function (varName) {
         }
 
         try {
-            file.contents = new Buffer(decorateAddition.concat(new Buffer(extendsAddition.concat(String(file.contents)).concat(moduleExportsAddition))));
+            file.contents = new Buffer(decorateAddition.concat(new Buffer(extendsAddition.concat(dependencyAddition).concat(String(file.contents)).concat(moduleExportsAddition))));
             this.push(file);
 
         } catch (err) {
--- a/src/Layer/babylon.highlightlayer.ts
+++ b/src/Layer/babylon.highlightlayer.ts
@@ -630,19 +630,22 @@
          * Add a mesh in the exclusion list to prevent it to impact or being impacted by the highlight layer.
          * @param mesh The mesh to exclude from the highlight layer
          */
-        public addExcludedMesh(mesh: Mesh) {
-            if (mesh instanceof BABYLON.InstancedMesh) {
-              mesh = (<BABYLON.InstancedMesh> mesh).sourceMesh;
+        public addExcludedMesh(mesh: AbstractMesh) {
+            var sourceMesh: Mesh;
+            if (mesh instanceof InstancedMesh) {
+              sourceMesh = (<InstancedMesh> mesh).sourceMesh;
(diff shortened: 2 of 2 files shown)


deltakosh:
Ping @sebavan for validation

sebavan:
All Good on the highlight layer parts, THANK YOU.

deltakosh:
Before I can merge please fix the issue raised by Travis CI

sebavan:
The code now only supports Instances and meshes, no more abastract coud you change it to : 

 if (mesh instanceof InstancedMesh) {
     mesh = (<InstancedMesh> mesh).sourceMesh;
 }

?

deltakosh:
Please test it a bit more deeply as well :)

Swiftwork:
My apologies, it seems as if turned into a bigger issue than first intended.

@sebavan The issue is that an InstancedMesh cannot be type-casted from a Mesh due to them being completely separate classes. This leaves three available options for the methods "mesh" argument:

1. Allow the it to be of either type **Mesh | InstancedMesh**.
2. Use a common ancestor such as **AbstractMesh** (which currently only inherited by Mesh and InstancedMesh).
3. Test for type later and use a wildcard **any** type.

I believe I have seen option number 2 being used previously and therefore assumed that would be the preferred option. Regardless I believe the instanceof checking would look largely similar which ever option is chosen and I am happy to change it for another if you want.

@deltakosh Sure I will try to figure out how to break it ^^. After all the only change I have made to the code is that it refers to an InstancedMesh source rather than itself. I also checked to see if duplicate referrals to the same mesh would cause an issue, but it seemed to be handled fine. If you can come up with a potential issue I could try out I would be happy to :)

Regards
Swiftwork

deltakosh:
No worry, this is fine 👍 
By the way I'm for option 2

Swiftwork:
Any updates on this? Are we waiting for confirmation or validation from some individual? :)

deltakosh:
I'mwauting for your PR update:)

Swiftwork:
Oh sorry, there seems to be some miscommunication. I thought we settled the matter earlier when you said "No worry, this is fine 👍". I have no more commits to add and the latest commit uses the second method posted earlier which everyone seemed content with. I have yet to find any issues, so with your approval and @sebavan I believe it is ready. Travis CI shows no errors now so go ahead and merge when you two OK it! :)

deltakosh:
Oh! Ok:)
@sebavan good to merge?

canvas2D/gulp-addModuleExports.js
```diff
@@ -18,11 +18,16 @@ module.exports = function (varName) {
...
-            'else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;\n' +
-            'return c > 3 && r && Object.defineProperty(target, key, r), r;\n' +
+        'var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {\n' +
+            'var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;\n' +
+            'if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);\n' +
+            'else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;\n' +
+            'return c > 3 && r && Object.defineProperty(target, key, r), r;\n' +
+        '};\n';
+
+        var dependencyAddition =
```
deltakosh: Why this? Seems weird as it will be merge in the babylonjs file

sebavan:
Yep I am good :-) sorry for the delay

canvas2D/gulp-addModuleExports.js
deltakosh: Sorry I thought it was the file in bjs gulps folder. Everything's c'est for me

PR #1433, opened 2016-10-19
Adding Scene.onRenderingGroupObservable

nockawa:
Also removed the /// <reference tags in some files


deltakosh:
Like it! really cool!

PR #1434, opened 2016-10-19
Implicitly create renderingGroup when Observer are registered

nockawa:


src/Rendering/babylon.renderingManager.ts
```diff
@@ -170,7 +170,16 @@
...
+
+                    if (observable) {
+                        let renderingGroupMask = Math.pow(2, index);
+                        info.renderStage = RenderingGroupInfo.STAGE_PRECLEAR;
+                        info.renderingGroupId = index;
+                        observable.notifyObservers(info, renderingGroupMask);
+
+                        info.renderStage = RenderingGroupInfo.STAGE_POSTTRANSPARENT;
+                        observable.notifyObservers(info, renderingGroupMask);
+                    }                }
```
deltakosh: need a line feed

nockawa: yey, my commit stat will raise like hell!

nockawa: done

PR #1440, opened 2016-10-20
dispose renderGroupObserver

wyklion:
dispose renderGroupObserver when dispose canvas2d


canvas2D/src/Engine/babylon.canvas2d.ts
```diff
@@ -1219,6 +1224,7 @@
         private _cachingStrategy: number;
         private _hierarchyLevelMaxSiblingCount: number;
         private _groupCacheMaps: StringDictionary<MapTexture[]>;
+        private _renderingGroupObserver: Observer<Scene>;
```
nockawa: you've done half of the job! :) if the Canvas2D is not screenSpace, this member won't be initialized!
so I suggest you initialize this member in the constructor before the conditional path that sets it.

PR #1453, opened 2016-10-21
Fixing sky material

julien-moreau:
where "pow(skyColor,vec3(1.0/(1.2+(1.2*sunfade)));" may generate wrong values. Just keep retColor = skyColor instead


materialsLibrary/materials/sky/sky.fragment.fx
```diff
@@ -151,18 +153,15 @@ void main(void) {
...
-	// Lighting
-	vec3 diffuseBase = vec3(1.0, 1.0, 1.0);
-
 #ifdef VERTEXALPHA
 	alpha *= vColor.a;
 #endif
 
 	// Composition
-	vec4 color = vec4(baseColor.rgb, alpha);
+	vec4 color = vec4(retColor.rgb, alpha);
```
deltakosh: you should clamp here just in case

julien-moreau:
Just committed the clamp

PR #1454, opened 2016-10-22
ShaderMaterial Base64 Encoded Vertex And Fragment Programs

MackeyK24:
Added support for ShaderMaterial Base64 Encoded Vertex And Fragment
Programs:
{
   vertextElement: “base64:xxxxxxxx”,
   fragmentElement: “base64:xxxxxxxx”
}

### DIFF
--- a/src/Materials/babylon.effect.js
+++ b/src/Materials/babylon.effect.js
@@ -133,6 +133,12 @@ var BABYLON;
                 callback(vertexCode);
                 return;
             }
+            // Base64 encoded ?
+            if (vertex.startsWith("base64:")) {
+            	var vertexBinary = window.atob(vertex.substr(7));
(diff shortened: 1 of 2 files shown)


nockawa:
@deltakosh @MackeyK24 in the method _loadVertexShader you added the following code:
```javascript
            // Base64 encoded ?
            if (vertex.startsWith("base64:")) {
            	var vertexBinary = window.atob(vertex.substr(7));
            	callback(vertexBinary);
            	return;
            }```

Unfortunately the "startWidth" function is not available for the version of TS/JS I compile... so it crashes.

deltakosh:
You're definitely right. Reverting

nockawa:
indexOf could be used instead

PR #1456, opened 2016-10-24
add set id method to prim2dBase

wyklion:


canvas2D/src/Engine/babylon.prim2dBase.ts
```diff
@@ -1661,6 +1661,10 @@
         public get id(): string {
             return this._id;
         }
+        
+        public set id(value: string) {
```
nockawa: Sorry, forgot to tell you that the proper code you be like this:
```        public set id(value: string) {
            if (this._id === value) {
                return;
            }

            let oldValud = this._id;
            this.onPropertyChanged("id", oldValud, this._id);
        }
```
it's in order to throw a propertyChangedObservable when the id value changes

PR #1469, opened 2016-10-26
added a method to force camera's projection matrix

vousk:
In some cases it can be usefull to be able to define and freeze the camera's projection matrix.

### DIFF
--- a/src/Cameras/babylon.camera.ts
+++ b/src/Cameras/babylon.camera.ts
@@ -414,9 +414,11 @@
             return this._computedViewMatrix;
         }
 
-        public setForcedProjectionMatrix(array: number[]): void {
+        public freezeProjectionMatrix(array?: number[]): void {
             this._doNotComputeProjectionMatrix = true;
-            BABYLON.Matrix.FromArrayToRef(array, 0, this._projectionMatrix);
+            if (array) {
+                BABYLON.Matrix.FromArrayToRef(array, 0, this._projectionMatrix);
+            }
         };
         
         public getProjectionMatrix(force?: boolean): Matrix {
--- a/src/Cameras/babylon.camera.ts
+++ b/src/Cameras/babylon.camera.ts
@@ -414,10 +414,10 @@
             return this._computedViewMatrix;
         }
 
-        public freezeProjectionMatrix(array?: number[]): void {
+        public freezeProjectionMatrix(projection?: Matrix): void {
             this._doNotComputeProjectionMatrix = true;
-            if (array) {
-                BABYLON.Matrix.FromArrayToRef(array, 0, this._projectionMatrix);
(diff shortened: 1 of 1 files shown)


src/Cameras/babylon.camera.ts
```diff
@@ -413,8 +414,13 @@
             return this._computedViewMatrix;
         }
 
+        public setForcedProjectionMatrix(array: number[]): void {
```
deltakosh: can you call it freezeProjectionMatrix(...) ?

deltakosh: I was thinking about option 2 ;)

vousk: Done. But I did it directly in github (first time I noticed we can edit files here ;-) ), so I didn't tried to compile TS file (but it should pass, the change is so tiny I couldn't barely have made a typo)

src/Cameras/babylon.camera.ts
```diff
@@ -413,8 +414,15 @@
             return this._computedViewMatrix;
         }
 
+        public freezeProjectionMatrix(array?: number[]): void {
```
deltakosh: Why an array? Please use a Matrix here :)

vousk: Is it better now ? ;)

src/Cameras/babylon.camera.ts
```diff
@@ -413,8 +414,15 @@
...
+        public freezeProjectionMatrix(projection?: Matrix): void {
+            this._doNotComputeProjectionMatrix = true;
+            if (projection !== undefined) {
+                this._projectionMatrix = projection;
+            }
+        };
+        
         public getProjectionMatrix(force?: boolean): Matrix {
-            if (!force && this._isSynchronizedProjectionMatrix()) {
+            if ((!force && this._isSynchronizedProjectionMatrix()) || this._doNotComputeProjectionMatrix) {
```
deltakosh: One last and we are good
In order to avoid the code in isSynchonized we should move the freeze test right after the force

deltakosh: And we also need a unfreeze function :)
(I promise to merge after that :))

deltakosh: Ok I'll do it :)

vousk: Oooh, so sorry... I could have done it... 
C'était grand w-e férié + pont là, j'étais en déplacement familial, sans ordi ni internet pour souffler un peu... :-p

vousk: I notice you haven't done it yet, I'll send a PR

deltakosh: I did it

PR #1478, opened 2016-10-29
fixed bug with Bone.rotate, Bone.Scale (initialized private vars) -ad…

abow:
…ded Bone.setYawPitchRoll -added Bone.getAbsolutePosition

http://www.babylonjs-playground.com/#I6RJJ#13
http://www.babylonjs-playground.com/#11BH6Z#168


abow:
Sorry don't merge yet.  I need to make one more commit.

abow:
I think I'm all set.  Thanks

abow:
I think I to a little overzealous.  This seems to have broken rotation in World Space.  :(

src/Bones/babylon.bone.ts
```diff
@@ -329,5 +355,51 @@
...
+                tmat = this.getAbsoluteTransform();
+            }
+
+            result.x = tmat.m[12];
+            result.y = tmat.m[13];
+            result.z = tmat.m[14];
+
+        }
+
+        public static computeAbsoluteTransforms (bone:BABYLON.Bone): void {
```
deltakosh: Why static?

abow: because it's a recursive function and seemed to make more sense to me.

abow: Now that you mentioned it and I looked at it some more, I think it would be better as a public method.

abow:
It looks like rotating in world space is working for Bone.rotate, but not for Bone.setYawPitchRoll.

deltakosh:
Ok I'll wait to merge (thanks for the static stuff btw)

abow:
Darn, I committed the js file.

abow:
I discovered a bug that I will be committing a fix for in the next few minutes.

abow:
I fixed that last bug.

http://www.babylonjs-playground.com/#I6RJJ#22

abow:
Can you suggest a way to get the worldMatrix of the mesh from a bone so that we don't have to pass rotate and setYawPitchRoll the mesh when working in world space?

deltakosh:
Bones are not linked to a specific mesh (as they can be applied to several meshes)

PR #1479, opened 2016-10-30
FontTexture and Canvas2D's Text2D: Signed Distance Field

nockawa:
FontTexture: now support the creation of a Font Map built using the Signed Distance Field algo (check this out: https://youtu.be/CGZRHJvJYIg)

Text2D has now a fontSignedDistanceField settings to use this mode


nockawa:
[Signed Distance Field Font support](https://trello.com/c/5I9NvweO/80-signed-distance-field-font-support)

PR #1480, opened 2016-10-31
Fixed default epsilon value, check for cellSize

pbalmasov:
CellSize height was not validate, or if cellSize was not defined

### DIFF
--- a/src/Sprites/babylon.spriteManager.ts
+++ b/src/Sprites/babylon.spriteManager.ts
@@ -52,7 +52,7 @@
             if (cellSize.width && cellSize.height) {
                 this.cellWidth = cellSize.width;
                 this.cellHeight = cellSize.height;
-            } else if(cellSize === undefined) {
+            } else if(cellSize !== undefined) {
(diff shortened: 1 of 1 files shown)


src/Sprites/babylon.spriteManager.ts
```diff
@@ -43,20 +43,20 @@
...
             this._spriteTexture.wrapV = Texture.CLAMP_ADDRESSMODE;
 
-            this._epsilon = epsilon === undefined ? 0.01 : epsilon;
-
-            if (cellSize.width) {
+            if (cellSize.width && cellSize.height) {
                 this.cellWidth = cellSize.width;
                 this.cellHeight = cellSize.height;
-            } else {
+            } else if(cellSize === undefined) {
```
deltakosh: are you sure?
if cellSize is undefined then you set cellWith to udnefined?

pbalmasov: if cellSize is undefined then return this function

PR #1484, opened 2016-11-01
Try Fix HDR Cube Texture Deserialization

sebavan:


MackeyK24:
I tried this and is still tries to load the _px.jpg and _nx.jpg

PR #1492, opened 2016-11-04
Bone.setScale now works while animation is running

abow:


deltakosh:
You rock! Sorry I'm not really available these days

PR #1493, opened 2016-11-04
Added 'buttons[]' for the Arc and Free cameras to customize which pointer buttons rotate them

Poolminer:
In commit 1/2 they were named 'pointerButtons', but since accessing input properties directly from the camera instead of from the input class turned out to only be kept for backward compatibility, I renamed them to 'buttons'.

What do you think?

### DIFF
--- a/src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
+++ b/src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
@@ -42,7 +42,7 @@ module BABYLON {
             this._pointerInput = (p, s) => {
                 var evt = <PointerEvent>p.event;
 
-                if (this.buttons.indexOf(evt.button) === -1) {
+                if (p.type !== PointerEventTypes.POINTERMOVE && this.buttons.indexOf(evt.button) === -1) {
                     return;
                 }
 
--- a/src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
+++ b/src/Cameras/Inputs/babylon.arcrotatecamera.input.pointers.ts
@@ -5,7 +5,7 @@ module BABYLON {
         camera: ArcRotateCamera;
 
         @serialize()
-        public pointerButtons = [0, 1, 2];
+        public buttons = [0, 1, 2];
 
         @serialize()
         public angularSensibilityX = 1000.0;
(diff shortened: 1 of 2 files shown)


deltakosh:
Sounds good

deltakosh:
Actually did it work with touch?
I don't think so and for arcrotate it must.

deltakosh:
After checking main code touch must be supported also on freecamera

So I guess you need to add a test to see if the pointer is for mouse before testing the button

Poolminer:
I just checked the W3C Recommendation 24 February 2015
https://www.w3.org/TR/pointerevents/#button-states

As well as the W3C Editor's Draft 23 September 2016
https://w3c.github.io/pointerevents/#the-button-property

And I tested the following desktop browsers:
Firefox, Chrome, Edge, Internet Explorer, Opera

As well as the following mobile browsers:
Chrome, Firefox

And these are my findings:

The W3C documents both have '0' for 'Touch Contact'
This works in Babylon on both mobile Firefox and Chrome
The 2015 document has '-1' for 'Mouse move with no buttons pressed'
The 2016 document has '-1' for 'Neither buttons nor touch/pen contact changed since last event'
Of the desktop browsers, only IE and Edge use '-1', other browsers use '0'
My second commit would break IE/Edge, because '-1' is not in the list, so moves would not be registered (tested it)
My second commit would break other browsers if '0' wasn't in the list, because '0' is used for moves (tested it)

So I think instead of checking if the pointer type is a mouse (which doesn't seem to be needed, and would still break browsers), it should instead check if the event is a 'move' type (PointerEventTypes.POINTERMOVE), and only do the list check if it is NOT. That way, it would be the same as putting the list check in both the 'up' and 'down' ifs.

So
`                    if(this.buttons.indexOf(evt.button) === -1){
                        return;
                    }`

would become

`                    if(p.type !== BABYLON.PointerEventTypes.POINTERMOVE && this.buttons.indexOf(evt.button) === -1){
                        return;
                    }`

I tested it on both the desktop and mobile browsers, and it works (IT'S NOT IN THIS PULL REQUEST). So, what do you think?

deltakosh:
Sounds like a good plan

deltakosh:
So you confirm that you tested touch on majors browsers correct?

Poolminer:
That's correct

deltakosh:
thanks!

PR #1494, opened 2016-11-04
More bone work

abow:
-added Bone.getDirection, Bone.getDirectionToRef, Bone.setRotationMatrix, Matrix.FromXYZAxesToRef
-removed unnecessary references to BABYLON

http://www.babylonjs-playground.com/#I6RJJ#50

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -3116,6 +3116,35 @@
             result.m[14] = temp3 * plane.d;
             result.m[15] = 1.0;
         }
+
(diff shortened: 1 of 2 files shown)


src/Math/babylon.math.ts
```diff
@@ -3116,6 +3116,35 @@
             result.m[14] = temp3 * plane.d;
             result.m[15] = 1.0;
         }
+
+        public static FromXYZAxesToRef(xaxis: Vector3, yaxis: Vector3, zaxis: Vector3, mat: Matrix) {
```
deltakosh: Should it be Axis and not axes?

deltakosh: As a French guy I won't correct you here:)

PR #1496, opened 2016-11-09
Improve error if canvas is null or undefined

nycdotnet:
Greetings.  With the current version of Babylon.js, if the user passes in a null or undefined canvas, the error is "WebGL is not supported".  This is misleading.  This PR includes a better error.

I ran Gulp and it seems fine.  There was a bunch of other stuff that changed when I built, but I didn't include them - I am assuming this is OK and someone will do a full build before a release, right?

### DIFF
--- a/src/babylon.engine.js
+++ b/src/babylon.engine.js
@@ -214,6 +214,9 @@ var BABYLON;
             //    // Do nothing
             //}
             if (!this._gl) {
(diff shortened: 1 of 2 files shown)


deltakosh:
Really cool. This is perfect (no need to provide js either actually)

nycdotnet:
OK thanks.

PR #1497, opened 2016-11-09
Add Scene Metadata Support

MackeyK24:
Add scene level metadata support

### DIFF
--- a/src/Loading/Plugins/babylon.babylonFileLoader.ts
+++ b/src/Loading/Plugins/babylon.babylonFileLoader.ts
@@ -254,6 +254,11 @@
                     scene.enablePhysics(physicsGravity, physicsPlugin);
                 }
                 
+                // Metadata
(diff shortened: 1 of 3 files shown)


deltakosh:
can you also update the sceneSerializer?

MackeyK24:
Ok... I added:
// Metadata
            if (scene.metadata) {
                serializationObject.metadata = scene.metadata;
            }

to sceneSerializer... But i never did this for node.metadata.

Because node.metadata has @serialize, does it also need to be handled in sceneSerialize or is that automatic... What about mesh.metedata... I had to manually "Parse" mesh.metadata... So do i need to do anything with node.metadata in sceneSerilizer ?

deltakosh:
Correct
We need to do it for the mesh serialization (only) 
Other nodes are already handled

MackeyK24:
Done.

PR #1501, opened 2016-11-10
-added Bone.getRotation and Bone.getRotationToRef  -fixed Bone.getDir…

abow:
…ection (wrong values if mesh has moved)

### DIFF
--- a/src/Bones/babylon.bone.ts
+++ b/src/Bones/babylon.bone.ts
@@ -511,17 +511,17 @@
 
         }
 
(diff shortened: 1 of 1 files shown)


deltakosh:
CI is returning an error

abow:
Darn.  I guess I can't add another Mesh parameter to Bone.getDirection due to Node.getDirection.

abow:
I'm going to need to make it optional and add it as the last parameter.

PR #1502, opened 2016-11-11
-fix for Bone.getDirection when mesh is scaled -fix for scaling a bon…

abow:
…e before a bone is rotated and while animation is playing

### DIFF
--- a/src/Bones/babylon.bone.ts
+++ b/src/Bones/babylon.bone.ts
@@ -473,8 +473,8 @@
                 var poseMatrix = this._skeleton.getPoseMatrix();
 
                 if(poseMatrix){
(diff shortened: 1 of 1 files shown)


abow:
I might have a better fix for the second issue.  It looks like it only happens with Bone.setRotationWithMatrix.  I'm going to do some more testing.

deltakosh:
sounds good

abow:
I was able to replicate the issue with the other rotation functions.  The current commit is the only way I could fix the issue.

abow:
I think I figured out a better fix.  I'll commit it tonight.

deltakosh:
lol ok!

PR #1511, opened 2016-11-15
Cleanup mouse move management

noalak:
Previous update on mouse move management was a bit ugly.
PreviousPoint storage seems to have no use now.


deltakosh:
Actually I merged it too fast. This does not work.
I've just tested it on Chrome and Edge and it does not work on Chrome

PR #1515, opened 2016-11-17
More bone work

abow:
-changed some function signatures to make things more consistent within the Bone class
-fixed Bone.translate in world space

### DIFF
--- a/src/Bones/babylon.bone.ts
+++ b/src/Bones/babylon.bone.ts
@@ -184,9 +184,8 @@
             }else{
 
                 this._skeleton.computeAbsoluteTransforms();
-
                 var tmat = Tmp.Matrix[0];
-                var vec = Tmp.Vector3[0];
+                var tvec = Tmp.Vector3[0];
 
                 if (mesh) {
                     tmat.copyFrom(this._parent.getAbsoluteTransform());
@@ -195,12 +194,16 @@
                     tmat.copyFrom(this._parent.getAbsoluteTransform());
                 }
 
-                tmat.invert();
-                Vector3.TransformCoordinatesToRef(vec, tmat, vec);
+                tmat.m[12] = 0;
+                tmat.m[13] = 0;
+                tmat.m[14] = 0;
 
-                lm.m[12] += vec.x;
-                lm.m[13] += vec.y;
-                lm.m[14] += vec.z;
+                tmat.invert();
+                Vector3.TransformCoordinatesToRef(vec, tmat, tvec);
+                
+                lm.m[12] += tvec.x;
+                lm.m[13] += tvec.y;
+                lm.m[14] += tvec.z;
 
             }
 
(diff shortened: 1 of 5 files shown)


src/babylon.node.ts
```diff
@@ -346,12 +346,12 @@
...
             var result = BABYLON.Vector3.Zero();
 
-            this.getDirectionToRef(localAxis, result);
+            this.getDirectionToRef(result, localAxis);
             
             return result;
         }
 
-        public getDirectionToRef(localAxis:BABYLON.Vector3, result:BABYLON.Vector3): void {
+        public getDirectionToRef(result:BABYLON.Vector3, localAxis:BABYLON.Vector3): void {
```
deltakosh: Why inverting? Ref parameter is always the last in math.ts for instance

abow: Optional params have to go on the end and there are several functions in the Bone class that have optional Space and Mesh.

To keep the Bone class consistent I figured it would be a good to do the same with Bone.getDirectionToRef.  I also didn't like how the ref was forced to the middle due to extending Node.
Bone.getDirectionToRef(localAxis, result, mesh);

deltakosh: But here the localAxis is not optional

src/Bones/babylon.bone.ts
```diff
@@ -551,26 +612,41 @@
...
-            this.getRotationToRef(mesh, result);
+            this.getRotationToRef(result, space, mesh);
 
             return result;
 
         }
 
-        public getRotationToRef(mesh: AbstractMesh, result: Quaternion): void{
+        public getRotationToRef( result: Quaternion, space = Space.LOCAL, mesh?: AbstractMesh): void{
+
```
deltakosh: same here. I'm not convinced we should change the order. 
I would prefer using overloads instead:
public getRotationToRef(result: Quaternion):
public getRotationToRef(space ,result: Quaternion):
public getRotationToRef(space = Space.LOCAL, mesh: AbstractMesh = null, result: Quaternion):

deltakosh: with overloads you can have:
getRotationToRef(result);
getPositionToRef(result);
getDirectionToRef(localAxis, result);

abow: I tried this with Bone.getDirectionToRef:

getDirectionToRef(localAxis, mesh, result);

Typescript wouldn't allow it though due to Node.getDirectionToRef(localAxis, result).

deltakosh: correct, we need to rename the bone version

abow: Then I won't feel like I need to move the ref to the first param.

abow: I just reviewed the class again.  I'm not sure I like that idea.  :)

deltakosh: Lol ok

abow:
This adds:
translate in world space
getRotation in local space
getPosition in world space
setPosition in world space
getAbsolutePosition and setAbsolutePosition now call getPosition, setPosition

PR #1517, opened 2016-11-19
added Bone.setRotation and Matrix.FromQuaternionToRef

abow:


abow:
I just added SpotLight.getRotation that returns a Vector3.  To be consistent, I'm thinking I should probably change setRotation and getRotation in this class to setRotationQuaternion and getRotationQuaternion.

deltakosh:
Completely agree

PR #1522, opened 2016-11-20
Added more null checks for Mesh/Sprite actionManager

Poolminer:
Currently, whether a Mesh/Sprite has an _actionManager_ is checked once, and then _actionManager.processTrigger_ may be called multiple times. This may cause an error if the trigger is an _ExecuteCodeAction_, because the custom code can cause the _actionManager_ to be deleted, e.g. when the Mesh/Sprite is _.dispose()_ d. In my first commit, I've added the necessary checks. In my second commit, I've reformatted the whole code (my first commit wasn't formatted properly).

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -157,8 +157,8 @@
         public hoverCursor = "pointer";
 
         // Metadata
-        public metadata:any = null;
-        
+        public metadata: any = null;
+
         // Events
 
         /**
@@ -822,7 +822,7 @@
                     this.setPointerOverMesh(pickResult.pickedMesh);
 
                     if (this._pointerOverMesh.actionManager && this._pointerOverMesh.actionManager.hasPointerTriggers) {
-                        if(this._pointerOverMesh.actionManager.hoverCursor){
+                        if (this._pointerOverMesh.actionManager.hoverCursor) {
                             canvas.style.cursor = this._pointerOverMesh.actionManager.hoverCursor;
                         } else {
                             canvas.style.cursor = this.hoverCursor;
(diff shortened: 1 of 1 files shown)


src/babylon.scene.ts
```diff
@@ -906,10 +906,12 @@
                                     pickResult.pickedMesh.actionManager.processTrigger(ActionManager.OnRightPickTrigger, ActionEvent.CreateNew(pickResult.pickedMesh, evt));
                                     break;
                             }
-                            pickResult.pickedMesh.actionManager.processTrigger(ActionManager.OnPickDownTrigger, ActionEvent.CreateNew(pickResult.pickedMesh, evt));
+                            if (pickResult.pickedMesh.actionManager) {
```
deltakosh: why this? (Check line 894, the test is already done)

src/babylon.scene.ts
```diff
@@ -1003,13 +1007,14 @@
                     }
                     if (pickResult.pickedMesh.actionManager) {
                         pickResult.pickedMesh.actionManager.processTrigger(ActionManager.OnPickUpTrigger, ActionEvent.CreateNew(pickResult.pickedMesh, evt));
-
-                        if (Math.abs(this._startingPointerPosition.x - this._pointerX) < ActionManager.DragMovementThreshold && Math.abs(this._startingPointerPosition.y - this._pointerY) < ActionManager.DragMovementThreshold) {
-                            pickResult.pickedMesh.actionManager.processTrigger(ActionManager.OnPickTrigger, ActionEvent.CreateNew(pickResult.pickedMesh, evt));
+                        if (pickResult.pickedMesh.actionManager) {
```
deltakosh: already tested in line 1008

deltakosh: Woot. Interesting :) This is clearly a rare but exisiting case :)
Thank you for finding it

PR #1523, opened 2016-11-21
check that to and from keys are not the same

boyofgreen:
adding in check to make sure to and from are not the same key, if so they error when used to set range as it will be 0 and then get an error when dividing by it later.

I noticed when I was animating small segments of time (a few key frames) the execution time would cause the "to" and "from" to be the same value.  This would then cause an error later due to it setting the range to 0, and trying to divide by it.  Let me know if there is a better way of doing this.


deltakosh:
Sounds like a great fix

PR #1542, opened 2016-11-30
added Ray.intersectsMesh and Ray.render

abow:
http://www.babylonjs-playground.com/#ZHDBJ#10


abow:
Maybe it should only render for 1-2 seconds?  The user could change the render time through an optional param.

abow:
or maybe just a show and hide function?

deltakosh:
Yep I'm more for a show/hide version

abow:
http://www.babylonjs-playground.com/#ZHDBJ#13

PR #1550, opened 2016-12-01
Compressed Textures Framework

Palmer-JC:
For using compressed textures formats based on what the hw / browser
supports, and those you said you have.


src/babylon.engine.ts
```diff
@@ -665,6 +698,9 @@
             this._badOS = regexp.test(navigator.userAgent);
 
             Tools.Log("Babylon.js engine (v" + Engine.Version + ") launched");
+            if (this._texturesSupported.length > 0) {
+                Tools.Log("Compressed textures formats HW / browser supports: " + this._texturesSupported);
```
deltakosh: Do you mind using Warn instead of Log?

Palmer-JC:
No, think it is important to definitely automatically log this.  I found differences even on the same machine between Chrome & Firefox.  If things in-inexplicably look different on the same machine, this would give time saving guidance. 

Do you wish me to do the change?

deltakosh:
Unfortunately yes. Most of the users of the framework does not want automatic logging. They want to opt in for it

Palmer-JC:
Did not know warn did not automatically come out, only in red.  Still, do you edit PR, or I commit again?

deltakosh:
I'll edit :)

deltakosh:
Actually the logging level was only for the loader. I'll remove it for now. We can add it in the debug layer for instance

Palmer-JC:
2 things:
first, now that I am building a test scene, am getting an error.  Pretty sure I need to add a line changing the extension like here:
                if (this._textureFormatInUse && !fromData && !scene.database) {
                    **extension = this._textureFormatInUse;**
                    url = url.substring(0, lastDot) + this._textureFormatInUse;
                }

unfortunately, right before I committed, I did a fetch / merge.  My Gulp is now broken.   

/js-dev/Babylon.js/Tools/Gulp/gulpfile.js:198
                    sourceRoot: (filePath) => {
                                           ^^
SyntaxError: Unexpected token =>
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Liftoff.handleArguments (/usr/local/lib/node_modules/gulp/bin/gulp.js:116:3)
    at Liftoff.<anonymous> (/usr/local/lib/node_modules/gulp/node_modules/liftoff/index.js:192:16)
    at module.exports (/usr/local/lib/node_modules/gulp/node_modules/liftoff/node_modules/flagged-respawn/index.js:17:3)

Kind of dead in the water.  Did a Gulp fine just minutes before.  Any ideas?

deltakosh:
can you make sure to run a npm install in the tools/gulp folder ?
We reorganized things a bit

sebavan:
It seems to not support the arrow for lambda I wonder what version of node you are using, if it still not work after upgrading I ll revert it to a normal function notation.

Palmer-JC:
You are going to need to make that one line change.  I no longer have a development worthy device.  Will not be doing pvr/etc implementation anytime soon.  .dds should should be better than 2.4 though.

deltakosh:
I will

PR #1551, opened 2016-12-01
Fixing mesh and vertex data serialization

julien-moreau:
Do you think we should also serialize overlay values ?
Here I serialized overlayAlpha


deltakosh:
Which overlay values? (thanks btw :))

julien-moreau:
:)
I mean mesh.overlayColor and mesh.renderOverlay

deltakosh:
sure please go ahead

PR #1552, opened 2016-12-01
Now serializing also overlays in SceneSerializer

julien-moreau:


deltakosh:
is the loader taking care of them?

julien-moreau:
Good catch !
I completely forget to check, commit is comming

julien-moreau:
I tested and it's fine now :)

PR #1558, opened 2016-12-04
Added slerp to BoneIKController

abow:


abow:
http://www.babylonjs-playground.com/#1EVNNB#24

PR #1561, opened 2016-12-06
Change handjs link to https://github.com/deltakosh/handjs repo

winksaville:
The http://handjs.codeplex.com indicates the project has moved to github
so updating the link.

Note: that https://github.com/deltakosh/handjs is discontinued and
suggests using https://github.com/jquery/PEP until native support is
pervasive.

The PE W3C spec https://www.w3.org/TR/pointerevents/ was approved as a
recommendation 24 Feburary 2015. There are currently are three native
implementations IE v11, Edge v14 and Chrome v55 according to
https://www.w3.org/TR/pointerevents/.


winksaville:
What is the time frame for moving Babylon move to jquery/PEP or PE W3C, if ever?

deltakosh:
There is no timeframe. You can keep using handjs

winksaville:
K, txs

PR #1600, opened 2016-12-19
Consolidate render target options into object and fix related bugs

haxiomic:
- Fixed bug where resizing would cause critical render target options to be lost
- Fixed bug where cloning would cause critical render target options to be lost
- Corrected types in MirrorTexture parameters
- Added additional render target parameters to MirrorTexture

### DIFF
--- a/src/Materials/Textures/babylon.renderTargetTexture.ts
+++ b/src/Materials/Textures/babylon.renderTargetTexture.ts
@@ -20,7 +20,7 @@
         * Use this predicate to dynamically define the list of mesh you want to render.
         * If set, the renderList property will be overwritten.
         */
-        public renderListPredicate: (AbstractMesh: AbstractMesh) => boolean;
+        public renderListPredicate: (AbstractMesh) => boolean;
 
         /**
         * Use this list to define the list of mesh you want to render.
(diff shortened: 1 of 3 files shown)


src/Materials/Textures/babylon.renderTargetTexture.ts
```diff
@@ -20,7 +20,7 @@
         * Use this predicate to dynamically define the list of mesh you want to render.
         * If set, the renderList property will be overwritten.
         */
-        public renderListPredicate: (AbstractMesh: AbstractMesh) => boolean;
+        public renderListPredicate: (AbstractMesh) => boolean;
```
deltakosh: why removing the type?

haxiomic: It doesn't remove the type, just the parameter name since the name is redundant
ie
```typescript
let x: (number) => void;
is the same as 
let x: (param:number) => void;
```

nevertheless can restore if prefered

src/Materials/Textures/babylon.renderTargetTexture.ts
```diff
@@ -99,9 +99,15 @@
         private _currentRefreshId = -1;
         private _refreshRate = 1;
         private _textureMatrix: Matrix;
-
-        constructor(name: string, size: any, scene: Scene, generateMipMaps?: boolean, doNotChangeAspectRatio: boolean = true, type: number = Engine.TEXTURETYPE_UNSIGNED_INT, public isCube = false, samplingMode = Texture.TRILINEAR_SAMPLINGMODE, 
-            generateDepthBuffer = true, generateStencilBuffer = false) {
+        protected _renderTargetOptions: {
```
deltakosh: Wondering if this could be a public type actually

src/Materials/Textures/babylon.renderTargetTexture.ts
haxiomic: This was a mistake, thanks for spotting, removed change

deltakosh:
Really good PR. Thanks a lot.
I think the RenderTargetOptions should be made public and move at the beginning of the file

PR #1601, opened 2016-12-19
Fix bug In CreateLathe constructor when assigning var arc

winksaville:
Added another ternary expression to test if options.arc exists before
using testing if options.arc <= 0 || options.arc > 1.

It might be worth considering using the Typescript compiler option
--strictNullChecks which can catch this type of bug.


winksaville:
I should have probably added a test but wasn't sure how, is there an example test for things like checking parameters? (I did a test on a little test app, but I'd like to learn the best practices for Babylon)

deltakosh:
We don't...we have an embryo here though: https://github.com/BabylonJS/Babylon.js/tree/master/tests
(only for canvas2d so far but feel free to add more n a /core/ folder ;))

btw thanks a lot for your contributions!

winksaville:
You're welcome, I'll take a look at the embryo later, cheers.

PR #1610, opened 2016-12-20
Engine texture fixes

haxiomic:
- Added missing gl.generateMipmaps to Engine createRawTexture
- Removed gl.generateMipmaps from createRenderTargetTexture since it's generating mipmaps of blank texture (and mipmaps are generated for the framebuffer color attachment elsewhere)

### DIFF
--- a/src/babylon.engine.ts
+++ b/src/babylon.engine.ts
@@ -2349,6 +2349,10 @@
             }
             gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 0);
 
+            if (generateMipMaps) {
+                this._gl.generateMipmap(this._gl.TEXTURE_2D);
+            }
+
             // Unbind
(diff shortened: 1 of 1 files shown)


src/babylon.engine.ts
```diff
@@ -2344,10 +2349,6 @@
             }
             gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 0);
 
-            if (generateMipMaps) {
```
deltakosh: This may be required to get a complete texture. Because this could be bind as a texture before a render is done

haxiomic: You're right, good spot, thanks! Will undo this change

nockawa:
So the mipmap behavior was unchanged for createRenderTargetTexture? Because I do use it on C2D for cached Groups.

deltakosh:
nope did not change. only for rawtexture

PR #1615, opened 2016-12-21
Development

sebavan:


src/babylon.scene.ts
```diff
@@ -1983,6 +1983,8 @@
             this._boundingBoxRenderer.reset();
             this._edgesRenderers.reset();
 
+            this.activeCamera.computeFrustumPlanes(this._frustumPlanes);
```
deltakosh: I'm a little bit concern by this. I do not like having a call just for a specific camera. (and it is weird as other cameras also have frustum planes)

sebavanmicrosoft: Yep that is not the version I intented to push at all... thanks for seeing this... I ll make another PR later tonight I lost my local final changes :-) the file is not having log in case holographic apis are not using and so on...

PR #1618, opened 2016-12-21
Fix a scene with many multi material starting with the same material

chloridrik:
solve this problem : 
http://www.html5gamedevs.com/topic/27204-unity-exporter-bugs-lightmap-mesh-and-textures/?do=findComment&comment=156059


deltakosh:
Ping @MackeyK24 for validation before merge

MackeyK24:
This looks good delta... A good job finding that bug. 

@deltakosh ... Please commit :)

PR #1621, opened 2016-12-21
Disable unused attributes when binding effects

haxiomic:
In Engine.enableEffect, disable attribute arrays not used by the effect.

When a buffer bound to an attribute array is disposed, the attribute array must be changed to a valid buffer or set to constant for the next draw call, even if the attribute is not used by a shader. Failing this will produce WebGL errors


haxiomic:
Closing: looks like this was resolved in 3474b9378272800c9cbd681f1df643785ad87bff

PR #1622, opened 2016-12-21
Use OpenGL projection matricies over Direct3D for better depth precision

haxiomic:
**Needs discussion before merge**

**TL;DR**: Small adjustment to projection matrices = a little less z-fighting

[WebGL's normalized device coordinates span the unit cube](http://www.songho.ca/opengl/gl_projectionmatrix.html), Direct3D normalized device coordinates are the same execpt z spans from 0 to 1 instead of -1 to 1.

Babylon uses projection matrices suited for Direct3D, consequently if a vertex is at the near plane a value of 0.5 will be written to the depth buffer and if a vertex is at the far plane 1.0 will be written to the depth buffer, so we're missing out on the full bit-depth available to us.

The fix (in this PR) is to adjust the projection matrices so when z = near plane, -1.0 is written and when z = far plane, 1.0 is written.

I noticed `PerspectiveFovWebVRToRef` has this adjustment but it's commented out.

In most cases this change will have no impact (other than increasing depth precision), but some code may rely on the old values (shadow techniques like VSM may look different for example) so we need input from people who know more about where this change may impact

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -2934,15 +2934,21 @@
         }
 
         public static OrthoLHToRef(width: number, height: number, znear: number, zfar: number, result: Matrix): void {
-            var hw = 2.0 / width;
-            var hh = 2.0 / height;
-            var id = 1.0 / (zfar - znear);
-            var nid = znear / (znear - zfar);
+            let n = znear;
+            let f = zfar;
 
-            Matrix.FromValuesToRef(hw, 0, 0, 0,
-                0, hh, 0, 0,
-                0, 0, id, 0,
-                0, 0, nid, 1, result);
+            let a = 2.0 / width;
+            let b = 2.0 / height;
+            let c = 2.0 / (f - n);
+            let d = -(f + n)/(f - n);
+
+            BABYLON.Matrix.FromValuesToRef(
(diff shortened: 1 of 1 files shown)


deltakosh:
Excellent! It was on my todo list since ages
I'll proceed to merge and do fixes as required

deltakosh:
Please credits yourself in the what's new for 2.6 :)

wyklion:
i see something wrong here.... 
code changing made my hero model display alpha wrong...

deltakosh:
alpha? This is not related to shaders or colors. Can you provide a repro on PG?

wyklion:
no... too hard to make it. i try in my project. my rotateCamera's radius = 1. the model is very closed to screen. and some skin looks like disppear or look otherside. and when i set radius = 3. whole model skin can be seen. so, it's a z calculate problem.

wyklion:
maybe it's right. just screen cut model.

haxiomic:
It'll be the near plane frustum culling - with the old matricies if a vertex is as at the near plane of the camera frustum, it's transformed into clip space as half way between the near at the far planes, so it can still get closer to the camera before being culled.

Try reducing your current near plane by a factor of 2

PR #1638, opened 2017-01-02
Master

MackeyK24:
Passing ‘null’ to registerOnPyhsicsCollide will allow to register
collision events for any/all imposters. Can filter collisions in your
collision event handler.


RaananW:
Hey,

The idea is great! 
this will, however, not quite work, as this function, thou public, is called by the physics engine itself, and is using pre-registered functions added using the registerOnPhysicsCollide function - https://github.com/MackeyK24/Babylon.js/blob/1aa728e76eb30f517ea70b806f7965dfde31ad4c/src/Physics/babylon.physicsImpostor.ts#L278
This function will have to be changed in order to get it to work correctly. Otherwise, this list will always exist (it will be an array, containing "null" if you provide null).

My quickest suggestion - an empty array means "all", and then you can inspect the array's length (instead of null checking). This has, of course, performance implications that the developer will have to face. That's why I tried avoiding "global" callbacks for collisions.

MackeyK24:
As Far As For Performance... I actually do filter based off metadata tags... Should not be MUCH less performant than if you passed a set of imposters to collide against: This allows me to assign game object tags as use as collision tags... So i can say tag ALL enemy ships with 'Enemy' then in my controller...   
mesh.setCollisionTags(["Enemy", "Asteroids"]);
mesh.onCollisionEnter =  (collider) => { collider.destroy()  }

WHICH INTERNALLY DOES A
this._mesh.physicsImpostor.registerOnPhysicsCollide([], this.processPhysicsCollisions)

Process Collision Events:
private processPhysicsCollisions(collider: PhysicsImpostor, collidedAgainst:PhysicsImpostor) : void {
            if (collider.object != null && collidedAgainst.object != null) {
                var colliderAny:any = collider.object;
                var collidedAgainstAny:any = collidedAgainst.object;
                if (colliderAny != null && colliderAny.metadata != null && colliderAny.metadata.collisionTags != null && collidedAgainstAny != null && collidedAgainstAny.metadata != null && collidedAgainstAny.metadata.tagName != null) {
                    var tag:string = collidedAgainstAny.metadata.tagName as string;
                    var tags:string[] = colliderAny.metadata.collisionTags as string[];
                    if (tags != null && tags.indexOf(tag) !== -1) {
                        console.log("===> OnPhysicsCollision: " + colliderAny.name);
                    }
                }
            }

NOTE: ALSO WORKS WITH DYNAMICALLY SPAWNED PREFABS.

PR #1641, opened 2017-01-04
Legacy Collision Detection Event Support

MackeyK24:
Allow for basic collision detection support without specific imposters.
Fixes otherImporters.length === 0 issue.

### DIFF
--- a/src/Physics/babylon.physicsImpostor.ts
+++ b/src/Physics/babylon.physicsImpostor.ts
@@ -337,13 +337,13 @@ module BABYLON {
 
         //event and body object due to cannon's event-based architecture.
         public onCollide = (e: { body: any }) => {
+            if (!this._onPhysicsCollideCallbacks.length && !this.onCollideEvent) return;
             var otherImpostor = this._physicsEngine.getImpostorWithPhysicsBody(e.body);
             if (otherImpostor) {
                 // Legacy collision detection event support
                 if (this.onCollideEvent) {
                     this.onCollideEvent(this, otherImpostor);
                 }
-                if (!this._onPhysicsCollideCallbacks.length) return;
                 this._onPhysicsCollideCallbacks.filter((obj) => {
                     return obj.otherImpostors.indexOf(otherImpostor) !== -1
                 }).forEach((obj) => {
--- a/src/Physics/babylon.physicsImpostor.ts
+++ b/src/Physics/babylon.physicsImpostor.ts
@@ -330,13 +330,22 @@ module BABYLON {
             }
         }
 
+        /**
+         * Legacy collision detection event support
+         */
+        public onCollideEvent: (collider:BABYLON.PhysicsImpostor, collidedWith:BABYLON.PhysicsImpostor) => void = null;
+
         //event and body object due to cannon's event-based architecture.
         public onCollide = (e: { body: any }) => {
-            if (!this._onPhysicsCollideCallbacks.length) return;
             var otherImpostor = this._physicsEngine.getImpostorWithPhysicsBody(e.body);
             if (otherImpostor) {
+                // Legacy collision detection event support
+                if (this.onCollideEvent) {
+                    this.onCollideEvent(this, otherImpostor);
+                }
+                if (!this._onPhysicsCollideCallbacks.length) return;
                 this._onPhysicsCollideCallbacks.filter((obj) => {
-                    return (obj.otherImpostors.length === 0 || obj.otherImpostors.indexOf(otherImpostor) !== -1)
+                    return obj.otherImpostors.indexOf(otherImpostor) !== -1
                 }).forEach((obj) => {
                     obj.callback(this, otherImpostor);
                 })
--- a/Exporters/3ds Max/BabylonExport.Entities/BabylonParticleSystem.cs
+++ b/Exporters/3ds Max/BabylonExport.Entities/BabylonParticleSystem.cs
@@ -12,6 +12,9 @@ namespace BabylonExport.Entities
         public bool preventAutoStart { get; set; }
 
         [DataMember]
+        public bool autoAnimate { get; set; }
+
+        [DataMember]
         public string emitterId { get; set; }
 
         [DataMember]
@@ -39,16 +42,13 @@ namespace BabylonExport.Entities
         public float[] colorDead { get; set; }
 
         [DataMember]
-        public float deadAlpha { get; set; }
-
-        [DataMember]
         public float emitRate { get; set; }
 
         [DataMember]
         public float updateSpeed { get; set; }
 
         [DataMember]
-        public int targetStopFrame { get; set; }
+        public float targetStopDuration { get; set; }
 
         [DataMember]
         public float minEmitPower { get; set; }
@@ -88,5 +88,19 @@ namespace BabylonExport.Entities
 
         [DataMember]
         public bool linkToEmitter { get; set; }
+
+        [DataMember]
+        public object[] animations { get; set; }
+
+        [DataMember]
+        public object customShader { get; set; }
+
+        public BabylonParticleSystem()
+        {
+            this.preventAutoStart = true;
+            this.autoAnimate = false;
+            this.animations = null;
+            this.customShader = null;
+        }
     }
 }
--- a/src/Particles/babylon.particleSystem.ts
+++ b/src/Particles/babylon.particleSystem.ts
@@ -41,6 +41,8 @@
 
         public layerMask: number = 0x0FFFFFFF;
 
+        public customShader: any = null;
+
         /**
         * An event triggered when the system is disposed.
         * @type {BABYLON.Observable}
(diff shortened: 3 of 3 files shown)


src/Physics/babylon.physicsImpostor.ts
```diff
@@ -330,13 +330,22 @@ module BABYLON {
...
         }
 
+        /**
+         * Legacy collision detection event support
+         */
+        public onCollideEvent: (collider:BABYLON.PhysicsImpostor, collidedWith:BABYLON.PhysicsImpostor) => void = null;
+
         //event and body object due to cannon's event-based architecture.
         public onCollide = (e: { body: any }) => {
-            if (!this._onPhysicsCollideCallbacks.length) return;
```
RaananW: Not sure I like this. There is a reason this is here. There was a serious drop in performance when it wasn't here. I noticed you are checking it further down, but it is already too late there.

You can check for both the length and the existence of the function before loading the other impostor.

MackeyK24: if (!this._onPhysicsCollideCallbacks.length) return; 

IS STILL THERE... Just moved AFTER the new onCollideEvent check... This will allow me to handle the collision check my (SHOULD I CHOOSE BY ASSIGNING onCollideEvent)
My Check in onCollideEvent is this:

`        private processPhysicsCollisions(collider: PhysicsImpostor, collidedAgainst:PhysicsImpostor) : void {
            if (collider.object != null && collidedAgainst.object != null) {
                var colliderAny:any = collider.object;
                if (colliderAny != null && colliderAny.metadata != null && colliderAny.metadata.collisionTags != null && colliderAny.metadata.collisionEvent != null) {
                    var collidedAgainstAny:any = collidedAgainst.object;
                    if ( collidedAgainstAny != null && collidedAgainstAny.metadata != null && collidedAgainstAny.metadata.tagName != null) {
                        if (colliderAny.metadata.collisionTags.length === 0 || colliderAny.metadata.collisionTags.indexOf(collidedAgainstAny.metadata.tagName) !== -1) {
                            colliderAny.metadata.collisionEvent(collidedAgainstAny);
                        }
                    }
                }
            }
        }
`

So you can see i do a form a checking that is almost like the original check for other imposter in a list... Instead i check if the otherImposter has a metadata tag that is in a list. This allows collision detection by the game object tag. Really helpful for dynamic clones created (like shots, or asteroid waves) that have a tag. Just need to detect when two rigid bodies collide with each other.

But to clarify ... Look at the WHOLE function (not just the minus in commit):

`        /**
         * Legacy collision detection event support
         */
        public onCollideEvent: (collider:BABYLON.PhysicsImpostor, collidedWith:BABYLON.PhysicsImpostor) => void = null;

        //event and body object due to cannon's event-based architecture.
        public onCollide = (e: { body: any }) => {
            var otherImpostor = this._physicsEngine.getImpostorWithPhysicsBody(e.body);
            if (otherImpostor) {
                // Legacy collision detection event support
                if (this.onCollideEvent) {
                    this.onCollideEvent(this, otherImpostor);
                }
                if (!this._onPhysicsCollideCallbacks.length) return;
                this._onPhysicsCollideCallbacks.filter((obj) => {
                    return obj.otherImpostors.indexOf(otherImpostor) !== -1
                }).forEach((obj) => {
                    obj.callback(this, otherImpostor);
                })
            }
        }
`

It basically give me a chance to handle the collision event AND STILL support the normal callback registration where i while have a list of imposters to give the registerOnPhyscisCollide...

Anyways... I really need a clean way of "Hooking" in the native collision detection and support all (and future physics engine)... Because this is the "BEST PLACE" to handle it.

MackeyK24: Note: Works beautiful on my end WITH NO PERFORMANCE HIT...

deltakosh: I see no problem on my side but I leave the final decision to @RaananW

RaananW: My dear, I can read code :-)
There is no reason to initialize the other impostor, if no function is defined. If you ask me, it's redundant. Even if you don't ask me, it's redundant.  
I'm actually not really sure why we need this function at all, but I understand why it will be used so it's fine.

I would prefer you well check before. This function is called in each and every collision. This can happen many times in each frame, and the extra work should be avoided. I read that you see no performance impact, but if you really want I'm prove you that there is one. Trust me please, because this line of code was added for a reason.

RaananW: And, I'm not sure what the particle system has to do with the pR? Was it a mistake?

abow: I've got to agree with RaananW.  Imagine if you have a hundreds of imposters and this gets called for every collision:

https://github.com/BabylonJS/Babylon.js/blob/dea49de07ac07a2bf7896aaa0e13c26e0f674c9e/src/Physics/babylon.physicsEngine.ts#L166

MackeyK24: But i do the checking in the actual function that is assigned onCollideEvent...

Even without my code change, in the original code... if you pass an imposter, it would check (several times per frame) if the imposter is in the list of callbacks... foreach one returned it would invoke the callback... I am doing the same exact thing in the onCollideEvent (except i am checkk for tags in a list on the metadata)... The ONLY difference with my code... it actually THUNKS into another function (whatever is assign onCollideEvent) to do my checking... So i ignore everything that is not tagged... If is only mesh (imposter) 1 tagged then it will only call is handler function... WHICH IS THE SAME THING... Now the if (!this._onPhysicsCollideCallbacks.length) return; get called right after that... if no other callbacks registered it will bail out... The mod code just gives me a chance to handle the RAW collision checking MYSELF... Should i choose to... I am just as carefule in my processCollisionEventHandler:


Here is the code that actually executes for onCollideEvent:

`        private processPhysicsCollisions(collider: PhysicsImpostor, collidedAgainst:PhysicsImpostor) : void {
            if (collider.object != null && collidedAgainst.object != null) {
                var colliderAny:any = collider.object;
                if (colliderAny != null && colliderAny.metadata != null && colliderAny.metadata.collisionTags != null && colliderAny.metadata.collisionEvent != null) {
                    var collidedAgainstAny:any = collidedAgainst.object;
                    if ( collidedAgainstAny != null && collidedAgainstAny.metadata != null && collidedAgainstAny.metadata.tagName != null) {
                        if (colliderAny.metadata.collisionTags.length === 0 || colliderAny.metadata.collisionTags.indexOf(collidedAgainstAny.metadata.tagName) !== -1) {
                            colliderAny.metadata.collisionEvent(collidedAgainstAny);
                        }
                    }
                }
            }
        }
` 
I just dont see the problem... But thats kool... its your code... I would NOT want folks just mucking around with my shit either ... I will figure out something else for my toolkit :)

As far as the particle System... I think i just changed different files to fix different issue... It would not allow me to cerate a seperate pull request... it just kept putting my changes onto the current open pull request... I dont know what hell i am doing most of the time with the github syncing... Most if the time i WAIT till one PR goes all the way thru then make a new one for next issue (i dont know why i didnt wait til you made a decision before adding the particle system changes... SORRY )

Do i have to take out all 3 of these commits then???

MackeyK24:
One last shot... Please take a look at this video of how i am using the onColideEvent event in the toolkit... Maybe it will change your mind :)

http://mackey.cloud/files/videos/u3dimposter.mp4

src/Physics/babylon.physicsImpostor.ts
deltakosh: No worry for the particle code. We prefer having just one feature per PR just in case we need to revert back.
Regarding your code, can you please change to this (We clearly need to be utterly cautious when dealing with performance intensive code):
`
                public onCollide = (e: { body: any }) => {

                if (!this._onPhysicsCollideCallbacks.length && !this.onCollideEvent) {
                   return;
                }

            var otherImpostor = this._physicsEngine.getImpostorWithPhysicsBody(e.body);
            if (otherImpostor) {
                // Legacy collision detection event support
                if (this.onCollideEvent) {
                    this.onCollideEvent(this, otherImpostor);
                }
                this._onPhysicsCollideCallbacks.filter((obj) => {
                    return obj.otherImpostors.indexOf(otherImpostor) !== -1
                }).forEach((obj) => {
                    obj.callback(this, otherImpostor);
                })
            }
        }`

MackeyK24:
SHIT.... @RaananW @Deltakosh @adam ... I AM A TOTAL DUMB ASS (AGAIN)... I AM SOOOOOO SORRY... BEING SOOOO THICK HEADED... I DIDNT REALLY CATCH WHAT YOU WERE SAYING....

Basically you just wanted my dumb ass to check before the 'getImposterWithPhysicsBody' if there are NO CALLBACK AND NO onCollideEvent handler... DONE MOVE DOWN AS I DID... JUST FIX UP TO INCLUDE the !onCollideEvent check...

I feel so stupid.... sorry for all the HU-HAH !!!

deltakosh:
Lol no worry:) we all have our good and bad days.. We work as a team, everything else is secondary :D

MackeyK24:
Hey @deltakosh ... Did you check out the Custom Effect on Particle System... Now we can serialze a customShader object (base on ShaderMaterial options object) to the buuild-in customEffect to support shaders on Particle Systems (serialized and cloned). That allow me to add a Custom Effect Shader option to the babylonJS toolkit... Now right from toolkit... you can have full control of particle system, and attached shaders and ANY controlling script attached. Even supports full prefab support and dynamic runtime instantiation ... Mixed with the new Babylon Prefab Layer (In upcoming Version 1.0.1)... is very sweet :)

deltakosh:
I saw that and I liked IT 💃

PR #1651, opened 2017-01-08
corrected applyForce() in CannonJSPlugin and made OimoJSPlugin implement IPhysicsEnginePlugin

ali-jawad:
Also deleted obsolete commented out code or unnecessary and wrong comments.

### DIFF
--- a/src/Physics/babylon.physicsEngine.ts
+++ b/src/Physics/babylon.physicsEngine.ts
@@ -195,7 +195,7 @@
         sleepBody(impostor: PhysicsImpostor);
         wakeUpBody(impostor: PhysicsImpostor);
         //Joint Update
-        updateDistanceJoint(joint: DistanceJoint, maxDistance:number, minDistance?: number);
+        updateDistanceJoint(joint: PhysicsJoint, maxDistance:number, minDistance?: number);
         setMotor(joint: IMotorEnabledJoint, speed: number, maxForce?: number, motorIndex?: number);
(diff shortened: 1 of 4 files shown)


deltakosh:
ping @raanan for validation

deltakosh:
@raananW

src/Physics/babylon.physicsEngine.ts
```diff
@@ -195,7 +195,7 @@
         sleepBody(impostor: PhysicsImpostor);
         wakeUpBody(impostor: PhysicsImpostor);
         //Joint Update
-        updateDistanceJoint(joint: DistanceJoint, maxDistance:number, minDistance?: number);
+        updateDistanceJoint(joint: PhysicsJoint, maxDistance:number, minDistance?: number);
```
RaananW: Why not leave it DistanceJoint ?

RaananW:
I am actually merging that, I see it compiled and actually does nothing on the BJS Joint level (only on the physics engine level, which is "any" anyhow).

Thanks for the change!

src/Physics/babylon.physicsEngine.ts
ali-jawad: Thanks @RaananW for the merge.

Just to answer you. I did this because it is the best common class to be used by both Cannon and Oimo plugins. Cannon plugin wants to use DistanceJoint in its updateDistanceJoint method parameter while Oimo wants to use IMotorEnabledJoint. PhysicsJoint class is the middle ground as it is the base class of DistanceJoint and the only implementer class of IMotorEnabledJoint interface.

I hope this clarifies things up :)

PR #1657, opened 2017-01-10
Update Preview Builds + bugs fix + minor changes

nockawa:
PreviewBuilds of all lib is updated

Fix bug on Engine.createTexture when no onLoad callback where passed.

Web.config now support .fnt file loading

Fix issue in config.json (.ts instead of .js for babylon.fontTexture)


nockawa:
Well, everything built successfully locally with a gulp typescript-all
I'm not sure the CI was ok on this one but I don't know how to request a new built again.

PR #1660, opened 2017-01-11
added AbstractMesh addChild, removeChild, setParent with option to ke…

abow:
…ep world position and rotation


abow:
http://www.babylonjs-playground.com/#28IXSE#22

PR #1663, opened 2017-01-11
Fully Implement Particle System Auto Start

MackeyK24:
Fix particle system ‘preventAutoStart’ for all instances, including
clones.


deltakosh:
good idea

PR #1664, opened 2017-01-11
new feature : facet data

jbousquie:
Brand new feature : **Mesh Facet Data**  
Examples :
```javascript
mesh.updateFacetData();  // enables and computes the facet data for this mesh
                                          // this can be called on demand, if the mesh is morphed for instance
                                          // even in the render loop


// get the index of the closest facet to (x, y, z) in the world
var i = mesh.getClosestFacetAtCoordinates(x, y, z);  

var position = mesh.getFacetPosition(i);  // returns the i-th facet world position
var normal = mesh.getFacetNormal(i);     // returns the i-th facet world normal
// then do stuff with these values, i.e your own physics...

// you can even get the closest point to (x, y,z) !
var proj = BABYLON.Vector3.Zero();
i = mesh.getClosestFacetAtCoordinates(x, y, z, proj); // proj is (x, y, z) projected on the closest facet

```
Note : all the facet data computation are done by _ComputeNormals()_ in the same loop, so it's really fast and doesn't impact the usual normal computation.

### DIFF
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -1792,7 +1792,7 @@
             });
         }
 
-        // facet data
+        // Facet data
         /** 
          *  Initialize the facet data arrays : facetNormals, facetPositions and facetPartitioning
          */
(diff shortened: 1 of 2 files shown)


siddjain-adobe-1:
Hello, I am currently stuck with a problem need help in this regard.
Problem: A mesh contains 3 face (6 triangles) Face 1 right end is joined to face 2 left end and same face 2 right end is joined to face3 left. These faces has zero depth. I can rotate each face individually. When I rotate a face by 180 degree then I see z-fighting problem because two face intersect each other. For this I would like to implement something by which I can manually provide facet shorting order to remove z-fighting.

jbousquie:
Could you please make a repo in the PG and set the question in the forum before we know if it's a bug or an usage issue ?

PR #1677, opened 2017-01-16
Master

MackeyK24:
Support Setting Default VRCameraMatrics for serialized cameras that by
default use VRCameraMatrics.GetDefault()


MackeyK24:
Sorry About Type-O ... I meant VRCameraMetrics :)

RaananW:
Just referencing your last comment - It is hard to tell if anything changed in the VRCameraMetrics. Was there any change, or is it just formatting?

MackeyK24:
Added a static funtion:

public static SetDefault(defaults:VRCameraMetrics): void {
 +            VRCameraMetrics.defaultCameraMetrics = defaults;
 +        }

And In the GetDefaults It wil Check if Satatic Defaults have been set, if so it will use that as the VRCameraMetrics.GetDefault() return value.

Note: This allows us to set what will be the VR Camera Default values that will use during Camera construction... Without having to RE-RIG the camera to change a lens distance value.

BTW... I want to add Camera Viewport default values here in the default so and use those in the set rig function so WE DONT have to RE-REG then camera_rig[0].viewport = new Viewport() because they will come fro the VRCameraMetrics.GetDefaults.

Would be way cleaner than having to do something like this to change the viewports and lens distance (i really dont like the re-rigging part):

`
// apply the metrics
camera.setCameraRigMode(BABYLON.Camera.RIG_MODE_VR, { vrCameraMetrics: metrics, interpupillaryDistance: 0.01, interaxialDistance: 0.01 });

// widen the nose-bridge
camera._rigCameras[0].viewport = new BABYLON.Viewport(0, 0, 0.45, 1.0);
camera._rigCameras[1].viewport = new BABYLON.Viewport(0.55, 0, 0.45, 1.0);
`

RaananW:
The VR camera accepts metrics as a variable. What's wrong with generating the defaults, changing its values, and providing it in the constructor?

MackeyK24:
OK... i wil find another way to handle that... I didnt see a public metrics variable on VR camera... But i will look again... 

BTW... How do i cancel the PR

MackeyK24:
Shit i dont no how to cancel this... sorry... Can you please canel this from your side if possible

RaananW:
All good, it is closed now. 
Your problem is probably that you use your master for pull requests, do you will always see it and commit to this closed pr.

PR #1678, opened 2017-01-16
partitioning auto-adjustment + fix division by zero

jbousquie:
* Internal mesh partitioning improvement : 
Now the number of subdivisions is auto-adjusted per axis. 
Example : if the mesh is sized 150 x 3  x 200 and if we want 20 subdivisions, the z axis will be divided in 20 subdivisions, the x axis will be divided in only 15 (instead of 20) and the y axis in only 1.

* fixed some possible divisions by zero : if the mesh bbox is sized of zero on an axis (in case of a dynamic morphing), it's considered as sized of Epsilon instead.

### DIFF
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -1915,10 +1915,24 @@
          */
         public getFacetsAtLocalCoordinates(x: number, y: number, z: number): number[] {
             var bInfo = this.getBoundingInfo();
-            var subRatio = this._partitioningSubdivisions * this._partitioningBBoxRatio;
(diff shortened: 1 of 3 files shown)


jbousquie:
Wait please, a code optimization is coming

jbousquie:
done
* partitioning subdivisions per axis are computed only once on _updateFacetData()_ and stored for further use instead of being recomputed on each call to _getFacetsAtLocalCoordinates()_
* better way to prepare and store the parameters for _ComputeNormals()_ (no re-alloc, no recomput)
* _mesh.dispose()_ now also disposes the mesh facet data if any

PR #1679, opened 2017-01-16
created Ray.attachToMesh, Ray.detachFromMesh, Ray.dispose, and fixed …

abow:
…issue with Ray.show when ray doesn't have a length

http://www.babylonjs-playground.com/#ZHDBJ#21

### DIFF
--- a/Tools/Gulp/config.json
+++ b/Tools/Gulp/config.json
@@ -34,6 +34,7 @@
       "../../src/Culling/babylon.boundingBox.js",
       "../../src/Culling/babylon.boundingInfo.js",
       "../../src/Culling/babylon.ray.js",
+      "../../src/Culling/babylon.rayHelper.js",
       "../../src/Mesh/babylon.abstractMesh.js",
       "../../src/Lights/babylon.light.js",
       "../../src/Lights/babylon.pointLight.js",
--- a/src/Culling/babylon.ray.ts
+++ b/src/Culling/babylon.ray.ts
@@ -393,6 +393,8 @@
                 this._attachedToMesh.getScene().registerBeforeRender(this._updateToMeshFunction);
             }
 
+            this._updateToMesh();
+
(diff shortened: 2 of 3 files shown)


abow:
http://www.babylonjs-playground.com/#ZHDBJ#22

RaananW:
Just an opinion, but:

Ray is a mathematical concept (at least until now), that was used for culling (hence the place where it is stored). I don't think a render function, dispose, mesh attachment, belongs in the Ray class (maybe a new RayMesh class in the mesh directory).  I haven't followed all changes, so I am not sure when the ray received the render function, but before extending this stand-alone class (that had no dependencies apart from math until now) I think we should consider if it is the right place.

abow:
How about I move the show functionality to a RayViewer class and the attachToMesh to Mesh.attachRay?

abow:
or instead of AbstractMesh.attachRay, a RayHelper class with attachRayToMesh function?

RaananW:
I think the AbstractMesh class is large enough :)

Why not add an attachToMesh function to the ray(viewer) class? Keep the dependencies (of AbstractMesh) in control?

I like the idea of a RayViewer, I think it is better than adding it to the ray class. For me it is the same difference of Plane (Mesh) and Plane (Math). 
But I would wait for other to comment, maybe there is a much better idea I am missing :smile:

abow:
Did you see my comment about RayHelper?  I'm liking that idea.

RaananW:
Yeah, I like the idea, I just think that a general user won't find this helper class, so they will never know it exists. 
Or, did you mean there is no RayViewer, there is only a RayHelper class? then it is all the same. Just a different name.

abow:
> then it is all the same. Just a different name.

Right

abow:
http://www.babylonjs-playground.com/#ZHDBJ#26

PR #1680, opened 2017-01-16
fix possible division by zero

jbousquie:
Well, the branch title is not pertinent : 
* fix : possible division by zero in the ribbon uvs computation
* simplify extrusion cap computation


RaananW:
I just tried dividing by zero, and it totally failed. 😉

PR #1690, opened 2017-01-19
fixed issue with mesh not staying attached to bone when animation cha…

abow:
…nges

http://www.babylonjs-playground.com/#OCCLB#40


deltakosh:
Thanks mate!

PR #1692, opened 2017-01-19
fixed all animations in scene stopping as a result of last fix

abow:
http://www.babylonjs-playground.com/#OCCLB#43
http://www.babylonjs-playground.com/#OCCLB#42

### DIFF
--- a/src/Animations/babylon.animatable.ts
+++ b/src/Animations/babylon.animatable.ts
@@ -99,26 +99,29 @@
         }
 
         public stop(animationName?: string): void {
-            var idx = this._scene._activeAnimatables.indexOf(this);
-
-            if (idx > -1) {
-
-                if(animationName){
+            
+            if (animationName) {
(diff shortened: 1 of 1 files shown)


src/Animations/babylon.animatable.ts
```diff
@@ -99,44 +99,51 @@
...
-
-            if (idx > -1) {
-
-                if(animationName){
+            
+            if (animationName) {
+                
+                var idx = this._scene._activeAnimatables.indexOf(this);
+                
+                if (idx) {
```
RaananW: idx can still be 0 and valid, no? shouldn't it be -1 like the deleted line?

abow: You might be correct.

This fix was a quick fix to get the working code from 2.4 back into 2.6.

The animationName option that was added to 2.5 caused the bug.

PR #1696, opened 2017-01-20
Deprecated Ray.show(), Ray.hide() and added Ray.intersectsMeshes()

abow:
http://www.babylonjs-playground.com/#ZHDBJ#31

### DIFF
--- a/src/Culling/babylon.ray.ts
+++ b/src/Culling/babylon.ray.ts
@@ -216,7 +216,46 @@
 
         }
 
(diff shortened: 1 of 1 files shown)


src/Culling/babylon.ray.ts
```diff
@@ -216,8 +216,47 @@
...
+
+        }
+
+        /**
+         *  @Deprecated. Use new RayHelper.show() instead.
+         * */
         public show(scene:Scene, color:Color3): void{
 
+            console.warn('Ray.show() has been deprecated.  Use new RayHelper.show() instead.');
+
```
deltakosh: is it possible to then use RayHelper internally as well to avoid code duplication?

PR #1697, opened 2017-01-20
Master

MackeyK24:
Babylon entities MUST include the ‘metallic’ and ‘roughness’ properties
in the output json. Null checks allow for presence of fields set null
to default No Metallic Workflow. Unless explicitly set to value using
babylon entities.


src/Materials/babylon.pbrMaterial.ts
```diff
@@ -938,7 +938,7 @@
                 this._defines.RADIANCEOVERALPHA = true;
             }
 
-            if (this.metallic !== undefined || this.roughness !== undefined) {
+            if ((this.metallic !== undefined && this.metallic != null) || (this.roughness !== undefined && this.roughness != null)) {
```
deltakosh: just use !== (double equal) and we are good to go

PR #1702, opened 2017-01-23
Added textAlign vertical, horizontal and wordWrap to Text2D

abow:
http://babylonjs-playground.com/#2AVSFH#172

### DIFF
--- a/canvas2D/src/Engine/babylon.text2d.ts
+++ b/canvas2D/src/Engine/babylon.text2d.ts
@@ -530,7 +530,7 @@
                 let lineLengths = [];
                 let charWidths = [];
                 let charsPerLine = [];
-                let charCount = 0;
+                let numCharsCurrenLine = 0;
                 let contentAreaWidth = this.contentArea.width;
                 let contentAreaHeight = this.contentArea.height;
                 let numCharsCurrentWord = 0;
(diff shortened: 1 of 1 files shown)


abow:
http://babylonjs-playground.com/#2AVSFH#174

abow:
http://babylonjs-playground.com/#2AVSFH#177

canvas2D/src/Engine/babylon.text2d.ts
```diff
@@ -132,6 +132,14 @@
         }
     }
 
+    export enum TextAlign {
```
nockawa: We are not allowed to use enum or const enum, you have to realy on a static readonly number. right @deltakosh (because I have a doubt concerning const enum) ?

canvas2D/src/Engine/babylon.text2d.ts
```diff
@@ -404,6 +418,18 @@
             this._textSize           = null;
             this.text                = text;
             this.size                = (settings.size==null) ? null : settings.size;
+            
+            if(settings.textAlignH != null){
+                this.textAlignH = settings.textAlignH;
+            }
+
+            if(settings.textAlignV != null){
```
nockawa: textAlignV/H and this._wordWrap shoudl always be initialized in the object to keep the object's layout consistent with other instances.

PR #1704, opened 2017-01-24
added support for Uint32Array and Uint16Array based indices

Kesshi:
As discussed here: http://www.html5gamedevs.com/topic/27815-why-int32array-instead-of-uint32array-for-mesh-indices/

I added Uint32Array and Uint16Array as new type for mesh indices. For best performance Uint16Array shoulb be used for small meshes and Uint32Array only for big meshes.
I kept the Int32Array because of backwards compatibility and Geometry.ImportGeometry() is also based on it.

### DIFF
--- a/src/Collisions/babylon.collider.ts
+++ b/src/Collisions/babylon.collider.ts
@@ -319,7 +319,7 @@
             }
         }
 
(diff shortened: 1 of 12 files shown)


deltakosh:
Really good! I'll take sometimes tomorrow to check, test and merge

thanks a lot

deltakosh:
Sorry did not find time yet but I won't forget I promise

PR #1710, opened 2017-01-25
Support https

RaananW:


RaananW:
Using relative protocol (//) instead of forcing http(or https) is, in general, a nice idea :smile:

PR #1717, opened 2017-01-28
Canvas2D: Primitive Collision Manager + WireFrame2D Prim Type + LayoutEngine data

nockawa:


wyklion:
Oh. It's not the only place effect designSize.

nockawa:
> Oh. It's not the only place effect designSize.
Sorry, I don't understand this, can you explain please?

PR #1741, opened 2017-02-04
fix for text2d letter overlapping when text is set

abow:
fixes this bug:
http://babylonjs-playground.com/#2AVSFH#208


nockawa:
[text2D overlapping sometimes occurs when text is set](https://trello.com/c/cfqZ4uCN/94-text2d-overlapping-sometimes-occurs-when-text-is-set)

PR #1742, opened 2017-02-04
fixed auto resize issue for text2D

abow:


abow:
It looks like this is still an issue:

http://babylonjs-playground.com/#2AVSFH#202

PR #1743, opened 2017-02-04
Add mesh.collisionFilter

dad72:
Used to exclude an object from collisions by using a filter.
Thanks to aWeirdo : http://www.html5gamedevs.com/topic/28193-exclude-mesh-for-collision/?do=findComment&comment=162108

exemple PG here:
http://www.babylonjs-playground.com/#20PQBI#1


dad72:
I cancel this PR due to the remaining bug

PR #1745, opened 2017-02-04
collisionMask, allows filtering of which meshes should collide.

aWeirdo:
@ http://www.html5gamedevs.com/topic/28193-exclude-mesh-for-collision/
@ http://www.babylonjs-playground.com/#20PQBI#17

Allows the option to add collisionMask to abstractMeshes & freeCameras

Example & in-depth explaination can be found in the PG linked above.

To-Do's; 
Integrate into webworker..

### DIFF
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -103,10 +103,10 @@
             this.cameraRotation = new Vector2(0, 0);
         }
 	
-	 // Collision
+	 // Collisions
 	private _collisionMask = -1;
 	public set setCollisionMask(mask: number) {
-	    this._collisionMask = !isNaN(mask) ? mask : -1;
+	    this._collisionMask = mask && !isNaN(mask) ? mask : -1;
 	}
 	 
         public _collideWithWorld(velocity: Vector3): void {
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -8,7 +8,7 @@
 
         @serialize()
         public applyGravity = false;
-
+                
         public inputs : FreeCameraInputsManager;
         
         //-- begin properties for backward compatibility for inputs
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -125,7 +125,7 @@
 
             globalPosition.subtractFromFloatsToRef(0, this.ellipsoid.y, 0, this._oldPosition);
             this._collider.radius = this.ellipsoid;
-	    this._collider._collisionMask = !isNaN(this._collisionMask) ? this._collisionMask : -1;
+	    this._collider._collisionMask = (!isNaN(this._collisionMask) ? this._collisionMask : -1);
 		
             //no need for clone, as long as gravity is not on.
             var actualVelocity = velocity;
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -115,7 +115,7 @@
             globalPosition.subtractFromFloatsToRef(0, this.ellipsoid.y, 0, this._oldPosition);
             this._collider.radius = this.ellipsoid;
 
-            if(!this._collider.collisionFilter) this._collider.collisionFilter = this.collisionFilter ? this.collisionFilter : 1;
+	    this._collider._collisionMask = this._collisionMask ? this._collisionMask : -1;
 		
             //no need for clone, as long as gravity is not on.
             var actualVelocity = velocity;
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -8,7 +8,7 @@
 
         @serialize()
         public applyGravity = false;
-                
+	    
         public inputs : FreeCameraInputsManager;
         
         //-- begin properties for backward compatibility for inputs
@@ -102,7 +102,13 @@
             this.cameraDirection = new Vector3(0, 0, 0);
             this.cameraRotation = new Vector2(0, 0);
         }
-
+	
+	 // CollisionMask
+	public _collisionMask = -1;
+	public setCollisionMask(mask: number): void {
+	    this._collisionMask = !isNaN(mask) ? mask : -1;
+	}
+	 
         public _collideWithWorld(velocity: Vector3): void {
             var globalPosition: Vector3;
 
@@ -114,7 +120,6 @@
 
             globalPosition.subtractFromFloatsToRef(0, this.ellipsoid.y, 0, this._oldPosition);
             this._collider.radius = this.ellipsoid;
-
 	    this._collider._collisionMask = this._collisionMask ? this._collisionMask : -1;
 		
             //no need for clone, as long as gravity is not on.
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -102,7 +102,7 @@
             this.cameraDirection = new Vector3(0, 0, 0);
             this.cameraRotation = new Vector2(0, 0);
         }
-	
+
 	 // Collisions
 	private _collisionMask = -1;
 	 
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -106,7 +106,7 @@
 	 // Collisions
 	private _collisionMask = -1;
 	 
-	public get collisionMask() {
+	public get collisionMask(): number {
 	    return this._collisionMask;
 	}
 	 
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -125,7 +125,7 @@
 
             globalPosition.subtractFromFloatsToRef(0, this.ellipsoid.y, 0, this._oldPosition);
             this._collider.radius = this.ellipsoid;
-	    this._collider._collisionMask = this._collisionMask ? this._collisionMask : -1;
+	    this._collider._collisionMask = !isNaN(this._collisionMask) ? this._collisionMask : -1;
 		
             //no need for clone, as long as gravity is not on.
             var actualVelocity = velocity;
--- a/src/Cameras/babylon.freeCamera.ts
+++ b/src/Cameras/babylon.freeCamera.ts
@@ -8,7 +8,7 @@
 
         @serialize()
         public applyGravity = false;
-	    
+
         public inputs : FreeCameraInputsManager;
         
         //-- begin properties for backward compatibility for inputs
@@ -103,9 +103,9 @@
             this.cameraRotation = new Vector2(0, 0);
         }
 	
-	 // CollisionMask
-	public _collisionMask = -1;
-	public setCollisionMask(mask: number): void {
+	 // Collision
+	private _collisionMask = -1;
+	public set setCollisionMask(mask: number) {
 	    this._collisionMask = !isNaN(mask) ? mask : -1;
 	}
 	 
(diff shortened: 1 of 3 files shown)


abow:
You should probably make this work the way that it works in Cannon and Oimo.

https://github.com/schteppe/cannon.js/blob/master/demos/collisionFilter.html

dad72:
This PR is intended for collision without physics. With moveWithCollisions for example. I find this approach to CollisionFilter very useful as it has.

deltakosh:
Hello instead of using >, could you please use mask like in layerMask? I would like to keep aligned with other parts of the framework

aWeirdo:
Hi Deltakosh, i'm not sure how it would work with layerMask, But i'll took into it when i have some time.
the point of the filter was to allow groups of meshes to collide with other groups of meshes but not collide with each other, e.g. if you want player to collide with walls but not with NPC's, while still needing collision enabled for NPC's vs walls

deltakosh:
Yep so in this case you can make something like that:
`
player.collisionGroup = "group1";
player.collisionMask = "group2";

npc.collisionGroup = "group3";
npc.collisionMask = "group2";

wall.collisionGroup = "group2";
`

and the rule is: collide if this.collisionMask == target.collectionGroup

aWeirdo:
Ah, alright i'll probably not have time for it until the weekend, but will update when it's done, Would you prefer pre-set groups, something like mesh.setCollisionGroup ( BABYLON.COLLISION_MASK_ONE ); 

or just plain numbers? mesh.setCollisionGroup( 1 );

deltakosh:
numbers are fine

deltakosh:
And as @abow just told me: you can use bitwise comparison for mask:
if ((this.mask & target.group) !== 0) {
// do collision
}

abow:
http://www.babylonjs-playground.com/#1VOXGX#1

aWeirdo:
I updated the PG to use mask's as requested, will update the PR in a bit.

http://www.babylonjs-playground.com/#20PQBI#9

aWeirdo:
Okay @deltakosh 
It should be ready ready now,
 please review the changes and check for any errors, typescript isn't my strongest suit.

I wanted to add a some default groups, BABYLON.Collision_GROUP_x , feels like there needs to be some defaults to avoid too many misunderstandings, but i wasn't sure where to add them without having to create an entirely new file which seemed a bit much just for that.

src/Cameras/babylon.freeCamera.ts
```diff
@@ -102,7 +102,13 @@
             this.cameraDirection = new Vector3(0, 0, 0);
             this.cameraRotation = new Vector2(0, 0);
         }
-
+	
+	 // Collisions
+	private _collisionMask = -1;
+	public set setCollisionMask(mask: number) {
```
deltakosh: You can use a property here as one may want to get the collision mask

src/Cameras/babylon.freeCamera.ts
```diff
@@ -102,7 +102,13 @@
             this.cameraDirection = new Vector3(0, 0, 0);
             this.cameraRotation = new Vector2(0, 0);
         }
-
+	
+	 // Collisions
+	private _collisionMask = -1;
+	public set setCollisionMask(mask: number) {
+	    this._collisionMask = mask && !isNaN(mask) ? mask : -1;
```
deltakosh: not sure of this line
why not just this._collisionMask = !isNaN(mask) ? mask : -1;

src/Collisions/babylon.collisionCoordinator.ts
```diff
@@ -390,7 +392,8 @@ module BABYLON {
             // Check all meshes
             for (var index = 0; index < this._scene.meshes.length; index++) {
                 var mesh = this._scene.meshes[index];
-                if (mesh.isEnabled() && mesh.checkCollisions && mesh.subMeshes && mesh !== excludedMesh) {
+                if(!mesh._collisionGroup) mesh._collisionGroup = -1;
```
deltakosh: warning collisionGroup could be 0 and then the "if" won't work

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -164,12 +164,22 @@
...
+        private _collisionGroup = -1;
+        private _collisionMask = -1;
         public ellipsoid = new Vector3(0.5, 1, 0.5);
         public ellipsoidOffset = new Vector3(0, 0, 0);
         private _collider = new Collider();
         private _oldPositionForCollisions = new Vector3(0, 0, 0);
         private _diffPositionForCollisions = new Vector3(0, 0, 0);
         private _newPositionForCollisions = new Vector3(0, 0, 0);
+        
+        public set setCollisionGroup(mask: number) {
```
deltakosh: same here..need properties

src/Collisions/babylon.collisionCoordinator.ts
```diff
@@ -354,6 +354,8 @@ module BABYLON {
             collider.retry = 0;
             collider.initialVelocity = this._scaledVelocity;
             collider.initialPosition = this._scaledPosition;
+            collider._collisionMask = excludedMesh && excludedMesh._collisionMask ? excludedMesh._collisionMask : collider._collisionMask;
```
deltakosh: no need to add this on the collider. The excluded mesh is sent to the _collideWithWorld

src/Collisions/babylon.collisionCoordinator.ts
deltakosh: Here we can have a local variable named collisionMask = (excludedMesh  ? excludedMesh._collisionMask : -1)

deltakosh:
For named group I'm not sure there is aneed. It could be done by the developer. No need to create GRoup1 just to set it to 1

src/Cameras/babylon.freeCamera.ts
aWeirdo: hah, sorry, i'm tired, i guess i was thinking it was to stop protential errors if someone was to setCollisionMask( emptyVariable ); but ofc isNaN would return true in that case..

src/Collisions/babylon.collisionCoordinator.ts
aWeirdo: that would break the freeCamera through, i added the collisionMask of the freeCamera to the collider, since it doesn't seem like the camera is passed further into the collision code, i'll remove the check completely and it should be okay.

deltakosh: "that would break the freeCamera through"
why?

aWeirdo: i'm assuming the excludedMesh is null when it is a freeCamera moving, so it's mask would be overridden to -1 and collide with everything

deltakosh: ok gotcha!

aWeirdo:
@deltakosh 
Alright, i believe i've made the changes requested + some,
I removed collisionGroup completely, i was just getting confused by it, and since we're using masks, not really needed either.

I hope i did the functions/property things right..

http://www.babylonjs-playground.com/#20PQBI#14

Masks;
 * -1 ; Always collides.
 *  0 ; Never collides.
 *  1 ; Collides with mask 1;
 *  2 ; Collides with mask 2;
 *  3 ; Collides with masks 1, 2 & 3;
 *  4 ; Collides with mask 4;
 *  5 ; Collides with masks; 1, 2, 3 & 5;  ( (1 + 2) = 3) + 2 = 5.
etc.

src/Cameras/babylon.freeCamera.ts
```diff
@@ -114,7 +125,8 @@
 
             globalPosition.subtractFromFloatsToRef(0, this.ellipsoid.y, 0, this._oldPosition);
             this._collider.radius = this.ellipsoid;
-
+	    this._collider._collisionMask = this._collisionMask ? this._collisionMask : -1;
```
deltakosh: we should compare to undefined here because mask can be set to 0 and this is valid

deltakosh:
Thanks a lot!

aWeirdo:
awesome :-)

wongwill86:
I'm getting semantic errors when i run `gulp typescript`. after this merge. is there something i'm doing wrong?
```
~/src/baby-2/Tools/Gulp$ git reset --hard
HEAD is now at a90c5cb Merge pull request #1745 from aWeirdo/patch-4
~/src/baby-2/Tools/Gulp$ gulp typescript
[14:01:30] Using gulpfile ~/src/baby-2/Tools/Gulp/gulpfile.js
[14:01:30] Starting 'typescript'...
[14:01:30] Starting 'typescript-compile'...
../../src/Cameras/babylon.freeCamera.ts(128,21): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(389,22): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(389,56): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(389,83): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(391,62): error TS2341: Property '_collisionMask' is private and only accessible within class 'AbstractMesh'.
../../src/Collisions/babylon.collisionCoordinator.ts(391,88): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(398,132): error TS2341: Property '_collisionMask' is private and only accessible within class 'AbstractMesh'.
../../src/Mesh/babylon.abstractMesh.ts(177,17): error TS2300: Duplicate identifier '_collisionMask'.
[14:01:40] TypeScript: 8 semantic errors
[14:01:40] TypeScript: emit succeeded (with errors)
```

deltakosh:
I fixed it, it should be good with latest version

aWeirdo:
ahhhhh me and typescript are not good friends, sorry.

wongwill86:
thanks for looking into this! not trying to make anyone feel bad here :)

but with the latest :(
```
~/src/baby-3/Tools/Gulp$ git reset --hard origin/master
HEAD is now at 5a49382 Restore missing updates
~/src/baby-3/Tools/Gulp$ gulp typescript
[14:27:16] Using gulpfile ~/src/baby-3/Tools/Gulp/gulpfile.js
[14:27:16] Starting 'typescript'...
[14:27:16] Starting 'typescript-compile'...
../../src/Cameras/babylon.freeCamera.ts(128,21): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(389,22): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(389,56): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(389,83): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(391,62): error TS2341: Property '_collisionMask' is private and only accessible within class 'AbstractMesh'.
../../src/Collisions/babylon.collisionCoordinator.ts(391,88): error TS2339: Property '_collisionMask' does not exist on type 'Collider'.
../../src/Collisions/babylon.collisionCoordinator.ts(398,132): error TS2341: Property '_collisionMask' is private and only accessible within class 'AbstractMesh'.
[14:27:25] TypeScript: 7 semantic errors
[14:27:25] TypeScript: emit succeeded (with errors)
```

deltakosh:
Ok my bad, the fixes were still on my computer :) should be good now

wongwill86:
awesome! thank you!

PR #1747, opened 2017-02-07
Canvas2D: Fix contentArea, actualSize update when changing the width/height/size of a prim

nockawa:


nockawa:
[Check width/height change of a Prim triggers its contentArea recomputation](https://trello.com/c/SW10wL4I/92-check-width-height-change-of-a-prim-triggers-its-contentarea-recomputation)

PR #1751, opened 2017-02-07
Canvas2D: Shape2D fill and border change bug

nockawa:
[Trello](https://trello.com/c/QqIn8U67)


nockawa:
[Shape2D change fill brushes doesn't work always](https://trello.com/c/QqIn8U67/97-shape2d-change-fill-brushes-doesn-t-work-always)

PR #1753, opened 2017-02-07
fixed a couple of GridPanelLayout issues

abow:


nockawa:
Can you please check if the GridPanel is still working as expected with the latest dev I did? thanks

PR #1754, opened 2017-02-08
check animations all stop bug.

wyklion:


deltakosh:
can you provide a PG which higlights this issue? I see no problem in the current code

src/Animations/babylon.animatable.ts
```diff
@@ -119,7 +119,7 @@
                         numberOfAnimationsStopped++;
                     }
 
-                    if (animations.length == numberOfAnimationsStopped) {
```
wyklion: it has no means...
animations length x,  if numberOfAnimationsStopped = y, see (animations.splice(index, 1);) then animations.length = x-y;
the check animations.length == numberOfAnimationsStopped means (x-y == y).

wyklion:
(animations.length == numberOfAnimationsStopped) it has no means...
animations length x, if numberOfAnimationsStopped = y, see (animations.splice(index, 1);) then animations.length = x-y;
the check animations.length == numberOfAnimationsStopped means (x-y == y).

src/Animations/babylon.animatable.ts
wyklion: the variable numberOfAnimationsStopped is useless..

deltakosh: ok make sense :)

deltakosh: Then numberOfAnimationsStopped has to be removed

deltakosh:
Thanks!

PR #1764, opened 2017-02-12
Separate pipe for npm module d.ts file

wongwill86:
This is a continuation of https://github.com/BabylonJS/Babylon.js/pull/1380

https://github.com/Microsoft/TypeScript/issues/8335#issuecomment-215194561 also seems to corroborate with this method for node modules.

This commit gets around the compile errors by creating a separate babylon.module.d.ts file that should be included in the npm package. I made a change in the Tools/Npm/package.json but I don't think it'll be sufficient since it doesn't seem like it does all the packaging.

The original babylon.d.ts file is still produced so running `gulp loaders` shouldn't produce new errors for files that have
```///.<reference.path="../../../dist/preview.release/babylon.d.ts"/>``` 

If this is no longer the correct approach, please let me know! thanks!

### DIFF
--- a/Tools/Gulp/config.json
+++ b/Tools/Gulp/config.json
@@ -5,6 +5,7 @@
     "minNoWorkerFilename": "babylon.noworker.js",
     "minCoreFilename": "babylon.core.js",
     "declarationFilename": "babylon.d.ts",
+    "declarationModuleFilename": "babylon.module.d.ts",
     "outputDirectory": "../../dist/preview release",
     "srcOutputDirectory": "../../src/"
   },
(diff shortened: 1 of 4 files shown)


byWulf:
Hey @wongwill86 ,

perhaps you can help be out. I'm trying to use the new `babylon.module.d.ts` in my angular2 project. But now I get the following error at the startup:

    babylon.module.d.ts(4672,25): error TS1039: Initializers are not allowed in ambient contexts.

Could this be a bug of the new module file or am I doing something wrong? Using latest TypeScript 2.2.1.

Thank you!

wongwill86:
hrmm that error should show up if there's declare within a declare.
i just generated a module from master but i didn't see anything like that there.

it might help if you can post your repo up for people to see what your project setup looks like.

abow:
I think you just need to update Typescript.

PR #1767, opened 2017-02-13
Added colorFilter for height map grounds

jbousquie:
```javascript
var filter = new BABYLON.Color3(1.0, 0.5, 0.2);
BABYLON.MeshBuilder.CreateGroundFromHeightMap('m', url, {colorFilter: filter});
```
The abililty to change the way the gradient to elevate the ground altitude is computed.
Default Color3(0.3, 0.59, 0.11) as usual grey gradient.

### DIFF
--- a/src/Mesh/babylon.mesh.vertexData.ts
+++ b/src/Mesh/babylon.mesh.vertexData.ts
@@ -1392,6 +1392,7 @@
             var normals = [];
             var uvs = [];
             var row, col;
+            var filter = options.colorFilter || new Color3(0.3, 0.59, 0.11);
             
             // Vertices
             for (row = 0; row <= options.subdivisions; row++) {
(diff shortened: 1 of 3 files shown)


src/Mesh/babylon.mesh.vertexData.ts
```diff
@@ -1386,13 +1386,13 @@
         /**
          * Creates the VertexData of the Ground designed from a heightmap.  
          */
-        public static CreateGroundFromHeightMap(options: { width: number, height: number, subdivisions: number, minHeight: number, maxHeight: number, buffer: Uint8Array, bufferWidth: number, bufferHeight: number }): VertexData {
+        public static CreateGroundFromHeightMap(options: { width: number, height: number, subdivisions: number, minHeight: number, maxHeight: number, colorFilter: Color3, buffer: Uint8Array, bufferWidth: number, bufferHeight: number }): VertexData {
```
deltakosh: this is a breaking change. We need to move it at the end and provide default value

jbousquie:
actually not because it's not a list of ordered parameters but a 
property of the global parameter "options"

the order doesn't matter : options.colorFilter

moreover we said once that the vertexdata was a semi-public API, (not 
the real public used by the BJS users) and it could be subject to change 
without notification as it's called by the meshBuilder class only what 
sets the default value if undefined

I can of course add a test about the hypothetical null or undefined 
value in the vertexdata, if you prefer

deltakosh:
I'm stupid....But yes just add a null test just in case to provide default values

jbousquie:
Done (although the null test was also already done at the upper level in the MeshBuilder class)

PR #1775, opened 2017-02-14
Invertion of collisionMask.

aWeirdo:
Instead of collisionMasks deciding which meshes should collide, they'll now decide which meshes shouldn't.

This change allows us to easier achieve the main goal of the original "collisionFilter' PR, having a large group of meshes not collide with each other but still collide with other meshes.

Masks will now work as following; 
 -1; // Ignore collisions with all masks, except 0. 
  0; // Always collide - Overrides -1 (default)

  1; // Ignore collisions with other meshes or camera with mask 1
  2; // Ignore collisions with mask 2
  4; // Etc
  8; 
  2 + 8 ; // Ignore collisions with masks 2 & 8, same as writing " 2 | 8 ".

http://www.babylonjs-playground.com/#20PQBI#35


aWeirdo:
something like scene.inverCollisionMasks = boolean; ?

deltakosh:
I would see it more at mesh/camera level so you can decide precisely

aWeirdo:
this is really biting me in the butt at every turn :p 

It might be better to just add collisionGroup's back in, while i'm personally not a fan because of how easily they can cause problems, they do allow both behaviors more natively.

deltakosh:
I agree mask+group is the best option to me:)

aWeirdo:
alright, i'll close this and create a new PR in a minute or two, i have them in another patch, just gonna check it's all good and ready.

PR #1777, opened 2017-02-14
Re-Addition of abstractMesh.collisionGroup.

aWeirdo:
Re-Addition of mesh.collisionGroup.

    var group_Always_Collide = -1; (default behavior)
    var group_1 = 1;
    var group_2 = 2;
    var group_3 = 4;

    ground.collisionGroup = group_Always_Collide;

    box.collisionGroup = group_1; // Box is part of group_1
    box.collisionMask = group_2; // Collide with group_2

    sphere.collisionGroup = group_2; // Sphere is part of group_2
    sphere.collisionMask = group_1 + group_3; // Collide with group_1 & group_3

    box2.collisionGroup = group_3; // Box2 is part of group_3
    box2.collisionMask = group_2; // Collide with group_2

    camera.collisionMask = group_1 + group_3; // Collide with group_1 & group_3.

so if mask == group;  collide.

http://www.babylonjs-playground.com/#20PQBI#42


src/Cameras/babylon.freeCamera.ts
```diff
@@ -104,14 +104,14 @@
         }
 
         // Collisions
-        private _collisionMask = -1;
+        private _collisionMask = 0;
```
deltakosh: should be 1 by default then

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -166,7 +166,8 @@
 
         // Collisions
         private _checkCollisions = false;
-        private _collisionMask = -1;
+        private _collisionMask = 0;
+        private _collisionGroup = 0;
```
deltakosh: same should be one by default

src/Cameras/babylon.freeCamera.ts
aWeirdo: 0 will always cause a collision when we're doing ((mask & group) === 0), it will act as no groups or masks has been set

src/Collisions/babylon.collisionCoordinator.ts
```diff
@@ -393,7 +393,7 @@ module BABYLON {
             // Check all meshes
             for (var index = 0; index < this._scene.meshes.length; index++) {
                 var mesh = this._scene.meshes[index];
-                if (mesh.isEnabled() && mesh.checkCollisions && mesh.subMeshes && mesh !== excludedMesh &&  ((collisionMask & mesh.collisionMask) !== 0)) {
+                if (mesh.isEnabled() && mesh.checkCollisions && mesh.subMeshes && mesh !== excludedMesh &&  ((collisionMask & mesh.collisionGroup) === 0)) {
```
deltakosh: this does not work
we need !== 0 (because the mask and the group are supposed to match)

aWeirdo: i kept the invertion, everything collides by default, so it made more sense to me, that masks are used to ignore collision with other masks,

http://www.babylonjs-playground.com/#20PQBI#38

abow:
this might be helpful:
http://www.babylonjs-playground.com/#1VOXGX#1

src/Collisions/babylon.collisionCoordinator.ts
deltakosh: I don't think the invertion is a good idea. By default people expect to use mask&group !== 0
If you set mask and group to 1 by default everything will collide

deltakosh: mask are working the same way on layerMask for camera for instance

aWeirdo:
@abow i have seen it, it's not the behavior i was trying to achieve, 
everything already collides in babylon, 
so it made more sense (to me atleast :P ) that if mask === group, don't collide.

aWeirdo:
👍  alright

aWeirdo:
it should be all good again.

PR #1782, opened 2017-02-15
Features/ncoder/references

ncoder:
I was reading https://doc.babylonjs.com/generals/setup_visualstudio and I knew a simpler way to get sourcemaps up and running for debugging.

Right now I've got a nice standalone project setup to build babylon with very minimal changes, and I get:

map files

All I had to do was set "Combine Javascript Output file" to dist/babylon.2.5.js (probably should have been babylon.2.5.max.js, but a small detail)
and "Generate declaration files"

That gives a nice concatenated output built by the typescript compiler instead of grunt/gulp, with proper source mappings.

Details:

All these reference nodes do is tell the typescript compiler that this file needs to be concatenated before another one. I've noticed that some errors may happen at runtime if the order is wrong. As such, a few reference paths may still be missing for some features of the engine I havn't used.

### DIFF
--- a/src/Bones/babylon.bone.ts
+++ b/src/Bones/babylon.bone.ts
@@ -1,4 +1,6 @@
-﻿module BABYLON {
+﻿/// <reference path="..\babylon.node.ts" />
+
+module BABYLON {
     export class Bone extends Node {
         public children = new Array<Bone>();
         public animations = new Array<Animation>();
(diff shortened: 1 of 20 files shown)


deltakosh:
Excellent thing. I'm using vscode so this is not required :) But really happy to see it for vs

PR #1784, opened 2017-02-16
format parameter for textures

Kesshi:
I added a parameter to Texture, CubeTexture and DynamicTexture which defines the texture format,
as discussed here: http://www.html5gamedevs.com/topic/28460-reducing-gpu-memory-usage-for-textures/

Its not so nice to have so many parameters but i know that you don't like breaking changes.
Maybe we could add an overload version of the constructor which accepts an options object to make it more user friendly.


deltakosh:
It is ok:)

we can also think about adding static helper like the CreateFromFiles one

PR #1791, opened 2017-02-20
Curve3.catmull rom

jbousquie:
As the title says ...
```javascript
var points = [vec1, vec2, vec2, ... vecN];
var nbPoints = 20;
BABYLON.Curve3.CreateCatmullRomSpline(points, nbPoints);
```

thanks to BitOfGold PG demo ;-)

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -4811,6 +4811,28 @@
         }
 
         /**
+         * Returns a Curve3 object along a CatmullRom Spline curve : 
+         * @param points (array of Vector3) the points the spline must pass through
+         * @param nbPoints (integer) the wanted number of points in the curve 
+         */
+        public static CreateCatmullRomSpline(points: Vector3[], nbPoints: number): Curve3 {
+            var totalPoints = new Array<Vector3>();
+            totalPoints.push(points[0].clone());
+            Array.prototype.push.apply(totalPoints, points);
+            totalPoints.push(points[points.length - 1].clone());
+            var catmullRom = new Array<Vector3>();
+            var step = 1.0 / nbPoints;
+            for (var i = 0; i < totalPoints.length - 3; i++) {
+                var amount = 0.0;
+                for (var c = 0; c < nbPoints; c++) {
+                    catmullRom.push( Vector3.CatmullRom(totalPoints[i], totalPoints[i + 1], totalPoints[i + 2], totalPoints[i + 3], amount) );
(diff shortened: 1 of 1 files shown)


BitOfGold:
I think this creates duplicate points now, when c==0 and when c==nbPoints. (because the <=), when the first step is the same as the last step. It was better with < nbPoints, but needs another catmullRom.push with the same parameters, to push the last point.
(Am I right?)

BitOfGold:
using the first and last point's clone is a good trick, I like it, the curve still goes through the first and last points, without 'extending' the point list. :)

jbousquie:
right
but with "<", the curve doesn't reach the last point... so I think we 
can choose to not care about the last duplicated point because it does 
the job ;-)

BitOfGold:
`catmullRom.push( Vector3.CatmullRom(totalPoints[i], totalPoints[i + 1], totalPoints[i + 2], totalPoints[i + 3], amount) );`
If you add this before the returm, the amount is already at 1.0 and adds the last point.
(with the < c, not <=c)
Now as I see adds a duplicate at every c=0.

jbousquie:
I'll check this tomorrow

my tests worked well here

jbousquie:
I'll check it tomorrow as I said before. I don't want to spend much time on this, so as the first and the last point are needed as control points although the curve don't pass through them ( https://en.wikipedia.org/wiki/Cubic_Hermite_spline#Catmull.E2.80.93Rom_spline ), what would you suggest ?
I could simply remove them from the the final Curve3 object before returning it, what would be easy to do and to understand.

BitOfGold:
I think everyone **expects** the new curve to go through the first and last points too, and the best control point is an extrapolation from thoose points:
https://github.com/mrdoob/three.js/blob/master/src/extras/curves/CatmullRomCurve3.js#L126
The code in three.js version is a bit too much, but it also extrapolates that "first" point, to a control point, as it should. cloning the first point, maybe it works and shorter code.
(As in my PG. i did the same I thing.)

BitOfGold:
I dont know how to write code here (and I'M not good in typescript), but here is the code I think would work best:

[catmullrom.txt](https://github.com/BabylonJS/Babylon.js/files/788564/catmullrom.txt)

jbousquie:
@BitOfGold was right : my "amount" value was in final just over 1.0
Fixed now according to his suggestion. Thanks.  
You can merge it now, please ;-)

PR #1798, opened 2017-02-22
Canvas2D: introduce postScale + preview builds

nockawa:
- Sprite2D now use postScale to render with the appropriate size and still having the origin computing correctly.


nockawa:
[Check Origin behavior with Sprite2D](https://trello.com/c/gn8vwD0l/98-check-origin-behavior-with-sprite2d)

PR #1806, opened 2017-02-24
fix for billboard parent rotation bug

abow:
http://www.html5gamedevs.com/topic/28662-26-alpha-error-parent-child-relations/


deltakosh:
Thanks a lot!

PR #1807, opened 2017-02-24
Canvas2D: better font support

nockawa:
- BaseLine is introduced in FontText
- BMFont are now rendering better
- Text2D is updated to support these improvements


nockawa:
[Wrong Character width with BitmapFontTexture](https://trello.com/c/BzgpLg4u/102-wrong-character-width-with-bitmapfonttexture)

PR #1808, opened 2017-02-25
Update babylon.abstractMesh.ts

dad72:
I add a function getHeightMeshAtCoordinates, because when a terrain is serialize and reload with ImportMesh, getHeightAtCoordinates that is made for when creating a field with CreatGround, no longer works.
In this case, we can use this function which will return the point Y of a coordinate.

I also add setUpdatable on the mesh, because you can only put an object updatable to the creation of the object, but once serialize and reload with ImportMesh, it is no longer updatable. With this function you can redefine updatable.

### DIFF
--- a/src/Mesh/babylon.abstractMesh.ts
+++ b/src/Mesh/babylon.abstractMesh.ts
@@ -644,19 +644,19 @@
          * @param {number} z
          * @param {number} maxHeight (default 200)
          */
-        public getHeightMeshAtCoordinates(x: number, z: number, maxHeight = 200): AbstractMesh {		
-		    var ray = new BABYLON.Ray(new BABYLON.Vector3(x, maxHeight, z), new BABYLON.Vector3(0, -1, 0), 2*maxHeight);
-		    var res = this.intersects(ray, true);
-		    return res.pickedPoint.y;
-	    }
+        public getHeightMeshAtCoordinates(x: number, z: number, maxHeight = 200): Vector3 {		
+            var ray = new BABYLON.Ray(new BABYLON.Vector3(x, maxHeight, z), new BABYLON.Vector3(0, -1, 0), 2*maxHeight);
+            var res = this.intersects(ray, true);
+            return res.pickedPoint.y;
+        }
          
         /**
          * Redefine updatable on a mesh
          */
-        public setUpdatable(updatable?: boolean): AbstractMesh {		
-		    this.setVerticesData(BABYLON.VertexBuffer.PositionKind, this.getVerticesData(BABYLON.VertexBuffer.PositionKind), updatable);
-		    this.setVerticesData(BABYLON.VertexBuffer.NormalKind, this.getVerticesData(BABYLON.VertexBuffer.NormalKind), updatable);		
-	    }
+        public setUpdatable(updatable?: boolean): void {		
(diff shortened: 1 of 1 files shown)


dad72:
I'm wondering if I should move the setUpdatable function in Mesh.ts rather than in abstractMesh. I not know much with all its classes and with tapuscript.
I'm trying to make a PR to avoid asking for features on the forum, but I'm not sure what I really do. I am not against help on this PR if it is not compliant .

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -637,6 +637,26 @@
...
+        
+        /**
+         * Return the point Y of a coordinate on a mesh 
+         * @param {number} x
+         * @param {number} z
+         * @param {number} maxHeight (default 200)
+         */
+        public getHeightMeshAtCoordinates(x: number, z: number, maxHeight = 200): Vector3 {		
+            var ray = new BABYLON.Ray(new BABYLON.Vector3(x, maxHeight, z), new BABYLON.Vector3(0, -1, 0), 2*maxHeight);
+            var res = this.intersects(ray, true);
```
deltakosh: we are not sure we will pick the mesh. It should test for hit first

deltakosh:
the setUpdatable must set the flag for all vertex buffers. I'm not convinced this should be in the framework.
It depends a lot on what user wants. Some may just want to update the position, some the normal, etc... it is not generic enough to stay like this.

What we can do else: move the function to the vertexBuffer. I can help if you want

src/Mesh/babylon.abstractMesh.ts
deltakosh: Do you think this must be in the framework? It depends a lot on the mesh. While it makes sense for the GroundMesh, I'm not sure this is useful for the abstractmesh

dad72:
Yes I want help to move the function in VertexBuffer if it does not bother. Thank you

For other functions, a terrain serialized and reload with ImportMesh becomes a mesh, and getHeightAtCoordinates for groundMesh no longer works. I thought to recreate a function that does the same for the meshes.
This function works well on my project, but I am not very comfortable with tapuScript, I think I made at least one errors.

Yes this might be useful to the framework in order to obtain the coordinate Y of a mesh as plane or a ground serialized, but maybe it should not be in abstracmesh.

else if I do like that: 
```javascript
public getHeightMeshAtCoordinates(mesh: Mesh, x: number, z: number, maxHeight = 200): Vector3 {		
            var ray = new BABYLON.Ray(new BABYLON.Vector3(x, maxHeight, z), new BABYLON.Vector3(0, -1, 0), 2*maxHeight);
            var res = mesh.intersects(ray, true);
```

deltakosh:
Ok I've created a mesh.markVerticesDataAsUpdatable and now GroundMesh should be serialized correctly:
https://github.com/BabylonJS/Babylon.js/commit/0e1cbdb0ab9e1351614b9d63e402d7b54f72f10c

dad72:
Thanks DK 👍 . By cons geometries.vertexData is not serialize from a terrain with serializeMesh.
I can make you a PG on the forum if you want?
I receive "vertexData":[]
As a result, getHeightAtCoordinates does not yet work on a mesh.
type:Mesh Is serialize it well.

deltakosh:
yes please for the pg

dad72:
Ok, done.

PR #1815, opened 2017-03-01
Changed yaw, pitch, roll to rotataionY, rotationX, rotationZ in BoneLookController to make …

abow:
…it easier for models that don't have bones aligned on the Y axis.  Fixed another issue with timing in the Bone class.

### DIFF
--- a/src/Bones/babylon.bone.ts
+++ b/src/Bones/babylon.bone.ts
@@ -203,13 +203,20 @@ module BABYLON {
 
             }else{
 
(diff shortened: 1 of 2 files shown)


abow:
I'm going to have to rethink this.  The dude model is still giving me trouble when it comes to the new rotation constraint feature, because his bones yaw on the x axis and pitch on the z axis.  Maybe users should set the yaw and pitch axis when it is something other than yaw = y and pitch = x.

PR #1816, opened 2017-03-01
Canvas2D Text improvements

nockawa:
- Compute the FontTeture in Premul Alpha
- Discard transparent pixel in fragment shader for Text2D
- When the text was white over white background the render was inaccurate, artifacts occurred. Problem solved.
- Text2D rendered with a BitmapFontTexture can now use the PremulAlpha mode
- Add debugging purpose modes...


nockawa:
[Text2D wrong blending mode with antialiasing](https://trello.com/c/32cgBcFf/103-text2d-wrong-blending-mode-with-antialiasing)

PR #1832, opened 2017-03-04
Improving Standard Rendering Pipeline

julien-moreau:
- Added serialize / Parse
- Fixed depth-of-field base color post-process
- Added blur width for depth-of-field


julien-moreau:
Forgot something, closing the issue!

PR #1841, opened 2017-03-07
WebVR Controllers

RaananW:
A lot has changed.
WebVR controllers have been implemented, using the gamepad extensions API.
WebVR view matrix is now being taken from the vrDevice for better performance. Projection matrix - the same.
WebVR camera can now teleport(!!) using the regular position vector. 
What a world to live in!


src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -132,34 +190,86 @@ module BABYLON {
...
+            // is rotation offset set? 
+            if (!Quaternion.IsIdentity(this.rotationQuaternion)) {
+                this.rotationQuaternion.toRotationMatrix(this._tempMatrix);
+                this._tempMatrix.multiplyToRef(this._webvrViewMatrix, this._webvrViewMatrix);
+            }
+
+            return this._webvrViewMatrix;
         }
 
-        public getClassName(): string {
```
deltakosh: Why removing this?

deltakosh:
Impressive!!
If you don't mind I would love to get a documentation about WebVR cameras and controllers :)

src/Cameras/VR/babylon.webVRCamera.ts
RaananW: It is there, just moved position. Line 182.

PR #1852, opened 2017-03-09
Allows y === null in drawText.

aWeirdo:
Allows leaving Y as null in the dynamicTexture.drawText function and centers the text vertically instead.

PG; http://www.babylonjs-playground.com/#1023O#47

_Side-note;_
As working with font sizes is rather difficult due to the many measuring types, it's orbviously not perfect, but it works very well for standard-use cases and with the use of PX, as is used in the turtorial, adding further support for pt, em, vw & vh, etc would be too much of a complication( in my mind), for such a simple feature.

### DIFF
--- a/src/Materials/Textures/babylon.dynamicTexture.ts
+++ b/src/Materials/Textures/babylon.dynamicTexture.ts
@@ -75,11 +75,11 @@ module BABYLON {
             }
 
             this._context.font = font;
-            if (x === null  || typeof x === "undefined") {
+            if (x === null  || x === undefined) {
                 var textSize = this._context.measureText(text);
                 x = (size.width - textSize.width) / 2;
             }
-            if (y === null || typeof y === "undefined") {
+            if (y === null || y === undefined) {
                 var fontSize = parseInt((font.replace(/\D/g,'')));;
                 y = (size.height /2) + (fontSize/3.65);
             }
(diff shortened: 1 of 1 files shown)


src/Materials/Textures/babylon.dynamicTexture.ts
```diff
@@ -79,7 +79,11 @@ module BABYLON {
                 var textSize = this._context.measureText(text);
                 x = (size.width - textSize.width) / 2;
             }
-
+            if (y === null) {
```
deltakosh: or undefined

aWeirdo: updated :)

src/Materials/Textures/babylon.dynamicTexture.ts
```diff
@@ -75,11 +75,15 @@ module BABYLON {
             }
 
             this._context.font = font;
-            if (x === null) {
+            if (x === null  || typeof x === "undefined") {
```
deltakosh: i think that x === undefined should be enough

aWeirdo: yeah, sorry, gotten the habit of using typeof everywhere, but ofc in this case neither x or y will be undeclared.. will update in a minute :)

PR #1879, opened 2017-03-12
Custom Shader Used From StandardMaterial

NasimiAsl:
easy way to make  Custom Shader Used From StandardMaterial

I need test something in Playground so this PR for that

### DIFF
--- a/src/Materials/babylon.standardMaterial.ts
+++ b/src/Materials/babylon.standardMaterial.ts
@@ -1,4 +1,10 @@
 ﻿module BABYLON {
+    
+    interface ICustomMaterialBuilder {
+        (name: string ,FragVertexMainParam: string , FragParam:string ,VertexParam:string ): string;
+     }
+
+         
     class StandardMaterialDefines extends MaterialDefines {
         public DIFFUSE = false;
         public AMBIENT = false;
@@ -65,6 +71,9 @@
     }
 
     export class StandardMaterial extends Material {
+        
+        public customMaterialBuilder: ICustomMaterialBuilder;
+        
         @serializeAsTexture()
         public diffuseTexture: BaseTexture;
 
@@ -208,9 +217,10 @@
 
         private _useLogarithmicDepth: boolean;
 
-        constructor(name: string, scene: Scene) {
+        constructor(name: string, scene: Scene, customMaterial  : ICustomMaterialBuilder) {
             super(name, scene);
-
+            
+            this.customMaterialBuilder = customMaterial;
             this._cachedDefines.BonesPerMesh = -1;
 
             this.getRenderTargetTextures = (): SmartArray<RenderTargetTexture> => {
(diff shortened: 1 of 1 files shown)


deltakosh:
Hello, I cannot validate this PR. I try to keep the Standard material under control :)
But you can inherit from the Standardmaterial and either provide a custom material in the [material lib](https://github.com/BabylonJS/Babylon.js/tree/master/materialsLibrary) or directly insude ShaderBuilder:)

NasimiAsl:
it is not depend any for shader Builder or BABYLONX
i can inherited from Standard Material  like Custom Material
any way if you accepted this i can make new PR again 
ps i feel you don't agree  with this stuff any way i like to know about that

deltakosh:
I don't agree because the StandardMaterial is defined around a specific shader. if you want to use a custom shader you can either use a ShaderMaterial directly or create a custom material.

Does it make sense? (I do not want to look like rude of course and I hope I did not upset you :))

NasimiAsl:
so we r good :D  ( am i right ) 
What about CustomMaterial ??? that is good name or not ?? i wanna inherited that from StandarMaterial 

I personally see my self like student in BABYLONJS area and you r master so help me to find a good solution please

deltakosh:
I don't clearly understand what you try to achieve :)

NasimiAsl:
i wanna make new class with standard material options and change that
so i wanna say it like custommaterial (or any name)

but i try say " i wanna commit that in Babylon not BABYLONX "

deltakosh:
Ok gotcha but this seems too advance for the core engine. The main issue is that users will have to implement in their own shaders a lot of properties (the Stdmaterial is huge) and this will cause a lot of trouble and questions on the forum if this is in the core engine ;)

NasimiAsl:
i feel your sensitive and that is correct .
 we mix shadermaterial with standardmaterial
both in core so that can be good .

if you accept we do it in babylon
or not we do it in BABYLONX 

my recommended let we do it in babylon 
that is important for me because i wanna inherited exactly from standardmaterial
so when standard material if update my work is ok always

deltakosh:
But you can still do it in BABYLONX right? I mean you can inherit with no problem.

Unless you want to provide a better shader? in this case you should do a PR for the shader it self
What do you plan to do in this customMaterial?

NasimiAsl:
i wanna append all lights and shadow and all setting  for custom shader
so i wanna help for user to make BaseColor and remove diffuse Color and diffuse Texture 
and let user write this part  i can do this now but i hacked BABYLON.StandardMaterial.IsReady

i just add one function for replace default shader with new shader (in Shaderstore ) and when we define that standard material work with all option but with new diffuse (we define that in that function )

NasimiAsl:
and we can call that customMaterial that make new shader but still is a standard material with new diffuse

deltakosh:
But why not doing something simpler?
We can expose the defaultShaderName (by default == "default") as a static property. People can then change it to their own shader (So no need for an interface and all the associated code)

NasimiAsl:
but how he/she can append scene lights and shadow and all other setting like ambiant color ....
in this new shader?

NasimiAsl:
i wanna use all we have in standard material but have my own shader :)

deltakosh:
So it is a custom material and it could be added to the material library :)

NasimiAsl:
:(  let we have that in beta version
and if you can't see good result i do that in BABYLONX 
but i am 99 percent  sure that is easy and you enjoy that after you see last result

deltakosh:
not in BABYLONX. In the material library, This is where all materials go :)

NasimiAsl:
ok win win solution 

so we r still good friend ? :)

Palmer-JC:
In your next ShaderBuilder,  could you not make a StandardMaterial subclass?  Then have BuildMaterial() use that?  I like where this might go if then you could use a skeleton & animation

NasimiAsl:
i have other method now
that is Build 
that make part of shader i need that for this

deltakosh:
lol this has never changed :) Friend forever!

NasimiAsl:
thanks  a lot @deltakosh

NasimiAsl:
but i need have very little changes in standard material ok?
let me make new PR

dad72:
In made DK, it'd be to merge a standardMaterial in ShaderBuilder with splatmap use of ShaderBuilder which can use up to 8 textures diffuse, but can have shadows, lights, fog and other... of standardMaterial.

This will create a terrain that would be evolve with more options.

deltakosh:
I think we find a way:)

Issue #1881, opened 2017-03-13
attachToBone does not work anymore


dad72:
Hello,

In the demo available here : http://doc.babylonjs.com/tutorials/how_to_use_bones_and_skeletons
 to attach an object to a bone no longer works. 
The sphere is attached to the character when I look in the inspector. but she is not render. she is not visible.

http://www.babylonjs-playground.com/#11BH6Z#18

deltakosh:
@abow will check is asap :)

dad72:
Thanks Abow

PR #1885, opened 2017-03-13
fix for attachToBone

abow:
fix for this bug:

https://github.com/BabylonJS/Babylon.js/issues/1881


deltakosh:
Rats! My optimization was buggy :D

PR #1886, opened 2017-03-14
reset release memory

wyklion:
_activeSkeletons is very big.


nockawa:
@wyklion please replace your code

`
for (var index = 0; index < this.data.length; index++) {
    this.data[index] = null;
}`


by 

`this.data.length = 0;`

it's more efficient and also release the memory
http://stackoverflow.com/questions/1232040/how-do-i-empty-an-array-in-javascript

wyklion:
right.

PR #1889, opened 2017-03-14
CustomMaterial in MaterialLibrary

NasimiAsl:


julien-moreau:
Looks excellent ! Can you provide a quick example to use it ?
For example in the index.html
Thanks !

dad72:
That sounds excellent. I think we have an example of use after the forum:
http://www.html5gamedevs.com/topic/29018-make-your-wanted-shader-in-standardmaterial-coming-soon/

julien-moreau:
The link is unfortunately dead now :(
Waiting :)

NasimiAsl:
that work  for me?

julien-moreau:
I mean http://editor.eash.space/mixedmaterial/
Then waiting for your commits =D

dad72:
@Julien Yes, link work. I just watch and it works

Palmer-JC:
I would recommend renaming the ShaderBuilder extension TS & JS files currently, 1.0.  That way when your new version, which will now be dependent on 3.0, people can migrate on their own schedule.  Thanks.

deltakosh:
Hello, I think you also need to update the gulp process right?

Palmer-JC:
The "1.0" ShaderBuilder is a single .TS file.  There is no build in the repo.  I added the .TS file directly into my QI extension Gulp [build](https://github.com/BabylonJS/Extensions/blob/master/QueuedInterpolation/Gulp/config.json).

If the motivation behind getting this subclass into the main repo is to keep ShaderBuilder a single file,  that causes 2 versions for a while.  I like the sub-class approach, but unless it is thought it will actually be used apart from shader builder, better to bundle directly into ShaderBuilder from a dependency standpoint.

My Gulp directory from QueuedInterpolation, could just be copied and config.json (link above) edited.  A bonus is Gulp significantly reduces final file size.

Palmer-JC:
Whoops, my bad.  You were talking about the main repo Gulp.

deltakosh:
Yes I would like to update this file: https://github.com/BabylonJS/Babylon.js/blob/master/Tools/Gulp/config.json#L243

NasimiAsl:
any thing i most do???

deltakosh:
Yes update the file I mentioned to add compilation for your file

NasimiAsl:
that not have any depended for shader builder
and  that use from default fragment shader so don't have any special fx

deltakosh:
but you still need to compile the ts

NasimiAsl:
so i need add js file too?

deltakosh:
yes

NasimiAsl:
i  change config.json is that ok?

Tools/Gulp/config.json
```diff
@@ -340,6 +340,12 @@
                     "../../materialsLibrary/src/sky/sky.fragment.fx"
                 ],
                 "output": "babylon.skyMaterial.js"
+            },  
+            {
+                "files": [
+                    "../../materialsLibrary/src/Custom/Babylon.CustonMaterial.ts"
```
deltakosh: custoM

Tools/Gulp/config.json
```diff
@@ -340,6 +340,12 @@
...
+                ], 
+                "output": "Babylon.CustomMaterial.js"
```
deltakosh: lowercase customMaterial

Tools/Gulp/config.json
NasimiAsl: my bad :) w8

NasimiAsl:
ok  that is ready :)

NasimiAsl:
no : i don't know what is that

NasimiAsl:
i most do it in my client or any online way to do that ?

deltakosh:
On your client

NasimiAsl:
i write it with github editor is that important to compile that in client ?

abow:
Hi NasimiAsl, This is a really helpful doc:

https://doc.babylonjs.com/generals/how_to_start

NasimiAsl:
hi abow thanks ok i need learn that too :)
i had a simple changes but i see a big now  need time to do that :(

deltakosh:
I know this could be painful but it is our collective responsibility to ensure that everything is tested before validating a PR.

NasimiAsl:
i test it but not full source  and i have js but not make it with gulp
but i think i need clone this project any way let we do this ;) thanks deltakosh

deltakosh:
It is a good learning process :) You won't regret it

julien-moreau:
@NasimiAsl I just saw you added the compilation stage, thanks!
Can you also provide an example in the materials library demos then we can understand how we can work with it ? We did the same for all other materials:
In the index.html https://github.com/BabylonJS/Babylon.js/blob/master/materialsLibrary/index.html
In the test JS files in order to configure your material https://github.com/BabylonJS/Babylon.js/tree/master/materialsLibrary/test

You have a lot of examples then you can get inspired by other materials :)

NasimiAsl:
i can't  clone the Babylonjs   in my client that failed after  download 260 mb  3 times
that is may be for my country i don't know i try get that and make complete this step and next i write that sample first i need have one output (  with babylon ) 

but for test i make that with some  hack in old version of dev.babylon.js i think that is not good example for test but i share that in this link (http://editor.eash.space/mixedmaterial/  i don't know why this not work for some place )
but i know i focus to clone project

i wanna test that in my ddc server in franc  may be that work in there 
and in i ran we close to new year so 80% my family give my task time
so need some time to do it in private  and share my last design for that 

` new BABYLON.StandardMaterial("s", scene, function (defaultShader, fragVertexP, fragP, vertexP) {

        		     var name = "def2";
        		     BABYLON.Effect.ShadersStore[name + "PixelShader"] = BABYLON.Effect.ShadersStore[defaultShader + "PixelShader"]
                         .replace(fragVertexP, '\
                                \
                                \
                                float time;\
                                ' + fragVertexP)
                     .replace(fragP, '\
                                 float afs = sin(vPositionW.x*3. +sin(vPositionW.z*3.))  ;\
                                vec3 diffuseColor = vec3(min(0.75,max(0.25,afs))); ');

        		     BABYLON.Effect.ShadersStore[name + "VertexShader"] = BABYLON.Effect.ShadersStore[defaultShader + "VertexShader"]
                     .replace(fragVertexP, '\
                                \
                                \
                                \
                                ' + fragVertexP)
                      .replace(vertexP, '\
                                vec3 newPosition = position;\
                                \
                                ' + vertexP.replace('position', 'newPosition'));


        		     return "def2";
        		 });
`

deltakosh:
You can also try to download the zip file of the repo (which should be smaller)

dad72:
It seems complicated this PR. Why not just convert the ts in js file manually. Test the JS file on the playground, and if it works, send the .ts and the .js simply. When we validate a PR, there is the test of the glup compilation. I understand all of this PR, there is so much asked gradually, is believed to have finished and no, we still have to do other things. 
 
If you need help Nasimi, I can cut you project babylon in several 50 MB zip to download from my server dedicate.

NasimiAsl:
we r in holiday for new year in iran
when i back i compile that 
sorry i don't know why some ssl connection always fail in my country

any way i test .ts file but i need make it with glup

NasimiAsl:
i start that now again :)

dad72:
Cool, Good luck then, hoping that everything goes well. 👍

dad72:
Why rename it in CustomMaterial2? CustomMaterial is not good?

NasimiAsl:
Let I make New PR :)

NasimiAsl:
just i need a cdn ( un cached ) i correct that

PR #1890, opened 2017-03-14
release memory

wyklion:
I dont think this smartArray is suit for JS...

### DIFF
--- a/src/Tools/babylon.smartArray.ts
+++ b/src/Tools/babylon.smartArray.ts
@@ -44,6 +44,7 @@
         }
 
         public reset(): void {
+            this.data.length = 0;
             this.length = 0;
             this._duplicateId++;
         }


src/Tools/babylon.smartArray.ts
```diff
@@ -44,6 +44,7 @@
         }
 
         public reset(): void {
+            this.data.length = 0;
```
deltakosh: the whole point of the smartArray is to AVOID this ;)

deltakosh:
WHy this?

src/Tools/babylon.smartArray.ts
nockawa: @deltakosh check slack's #general

deltakosh:
I did this to address memory: https://github.com/BabylonJS/Babylon.js/commit/f59cb9e1236cd750c1bdba4cfbac4a17d7acce66

wyklion:
yes. dispose is better.

PR #1893, opened 2017-03-14
Changes in inspector for cameras

ThomasVallet:
Cameras added in inspector to show their properties and change our POV to change our active camera.


deltakosh:
Are you struggling with git? :)

ThomasVallet:
Yes, first time committing there... It's okay, I sent it to @Temechon , it should be working now :)

PR #1895, opened 2017-03-14
Refactoring and adding events into Observable and ActionManager classes

yuccai:
-add two masks for onPrePointerObservable and onPointerObservable which are POINTERTAP and POINTERDOUBLETAP
-add a new trigger for ActionManager which is onDoublePickTrigger
-add a static array  for ActionManager class which classified all types of triggers registered into all actionManager instances in a unique way, it's useful in order to prevent mesh picking if it's not requiered
-equivalent methods in ActionManager such as hasPickTriggers or hasSpecificTrigger are now implemented in Observable (Mask instead of Tigger) for the same reason
-move LongPressDelay and add DoubleClickDelay into static fields of Scene class, they are in common between Obervable and ActionManager
-add a flag to Scene to enable exclusive double click mode (double click raises only double click instead of raising a simple click followed by a double click)


src/Actions/babylon.actionManager.ts
```diff
@@ -62,16 +62,17 @@
...
-        private static _OnLongPressTrigger = 7;
-        private static _OnPointerOverTrigger = 8;
-        private static _OnPointerOutTrigger = 9;
-        private static _OnEveryFrameTrigger = 10;
-        private static _OnIntersectionEnterTrigger = 11;
-        private static _OnIntersectionExitTrigger = 12;
-        private static _OnKeyDownTrigger = 13;
-        private static _OnKeyUpTrigger = 14;
-        private static _OnPickOutTrigger = 15;
+        private static _OnDoublePickTrigger = 6;
```
deltakosh: For backward compat reason I prefer not changing existing numbers

src/Actions/babylon.actionManager.ts
```diff
@@ -138,9 +143,8 @@
             return ActionManager._OnKeyUpTrigger;
         }
 
-        public static DragMovementThreshold = 10; // in pixels
```
deltakosh: Why removing them?

src/Tools/babylon.observable.ts
```diff
@@ -154,6 +154,27 @@
 
             return result;
         }
+
+        public hasPickMasks(): boolean {
```
deltakosh: Too specific for the observable class

src/Actions/babylon.actionManager.ts
deltakosh: Breaking change by the way

src/Actions/babylon.actionManager.ts
```diff
@@ -233,6 +245,39 @@
             return false;
         }
 
+        public static get HasTriggers(): boolean {
```
deltakosh: Can you add comments for all functions which are public?

src/babylon.scene.ts
```diff
@@ -963,85 +1184,137 @@
...
-                if (pickResult.hit && pickResult.pickedMesh) {
-                    if (this._pickedDownMesh != null && pickResult.pickedMesh == this._pickedDownMesh) {
-                        if (this.onPointerPick) {
-                            this.onPointerPick(evt, pickResult);
-                        }
-                        if (this.onPointerObservable.hasObservers()) {
-                            let type = PointerEventTypes.POINTERPICK;
-                            let pi = new PointerInfo(type, evt, pickResult);
-                            this.onPointerObservable.notifyObservers(pi, type);
+                    if (pickResult.pickedMesh) {
```
deltakosh: pickResult can be null here...

src/Actions/babylon.actionManager.ts
yuccai: I did this just to ensure that the hasPickTriggers and hasPointerPickTriggers are still working without changing the code. You check the range of action.trigger value which must be greater than the lowest pick trigger (onPickTrigger = 1) and lower than the greatest (onPickUpTrigger = 6 for hasPickTriggers and onPointerOutTrigger = 9 for hasPointerTriggers). So I inserted the double click trigger value inside these two ranges in order to prevent any changes for other functions.
What you want me to do ? Change the double click trigger value (16 for example),  and complete conditions in hasPickTriggers and hasPointerTriggers ? Keep my changes  because the values of triggers are not supposed to be changed and seen ?

src/Actions/babylon.actionManager.ts
yuccai: I removed them because I needed the same kind of values for Observable. 
1. I could have used ActionManager values but I thought it was not correct ; why does Observable need to rely on ActionManager values ? 
2. I also could have create the same value in Observable class but I thought it was redundant. Is it really needed to provide 2 delays for double click, one for ActionManager and the other one for Observable ?
3. So I chose to deport theses values in Scene class because I considered that it was a good place to share theses datas within both classes.
What kind of solution do you prefer ?

src/Tools/babylon.observable.ts
yuccai: What do you mean by too specific ? They shouldn't exist ? They should be private ? 
I need them in order to optimize the pickUp callback in the initClickEvent function. This function checks all functionalities that it should for the current configuration and returns a ClickInfo which warn to observables and actionManager what must be raised. The functionalities checked depends on the type of the triggers registered in ActionManager (where the necessity to know what kind of triggers are registered without processing a mesh pick thanks to the static functions HasPickTriggers and HasSpecificTrigger) and in the same way, it depends on the type of observer masks registered in PointerObservable and PrePointerObservable. I don't want to check if a double click has happened if nothing requested it.

yuccai:
Here is a big PG testing previously existing features and the new one : [http://www.babylonjs-playground.com/index.html#Y7QQJ#1](http://www.babylonjs-playground.com/index.html#Y7QQJ#1).
I put all functionalities into the same PG. You have just to comment those that you don't need for testing. I may do other PG for specific functionalities if you want.
For the moment the PG is not effective because it relies on my babylon.js build

src/Actions/babylon.actionManager.ts
deltakosh: oh yes. makes sense. so keep it like this:)

src/Actions/babylon.actionManager.ts
deltakosh: This is a breaking change I cannot validate..And this leads me to my other remark: Observable should remain generic. The hasPickMask function should not be there.

This is not technical but more philosophical :)

src/babylon.scene.ts
```diff
@@ -779,6 +844,159 @@
...
+                                    this._previousHasSwiped = clickInfo.hasSwiped;
+                                }
+                            }
+                        }
+                    }
+                    clickInfo.ignore = true;
+                    cb(clickInfo, this._currentPickResult);
+
+                    /*
+                    let checkLongPicking = obs1.hasSpecificMask(PointerEventTypes.POINTERLONGPRESS) ||
```
deltakosh: This block should be removed

src/Actions/babylon.actionManager.ts
deltakosh: Well..actually I'm wrong..You are right here. The DragMovementThreshold  will not only affect ActionManager but all pointer observable..
So this change makes sense. Can you just add it in the what's new section about breaking changes?

src/Tools/babylon.observable.ts
```diff
@@ -154,6 +154,36 @@
 
             return result;
         }
+
+        /**
+         * Does this observable has observers registered with pick mask
+         * @return {boolean} whether or not it has observers registered with pick mask
+        **/
+        public hasPickMasks(): boolean {
```
deltakosh: This one can be removed. I would like to keep the observable as generic as possible. You can just use hasSpecificMask

yuccai:
I think I'm done with the changes you requested. I'll try to do another pull request to update the documentation. Let me know if all is ok :wink:

deltakosh:
Perfect!

PR #1897, opened 2017-03-15
GLTF: Correct initialization of node rotation from quaternion.

Misterblue:
In the GLTF loader, the initialization of a node's rotation from a specified quaternion is wrong. This commit replaces the Quaternion.RotationAxis computation with a Quaternion.FromArray.


deltakosh:
Hello
Probably a stupid question but:
how do we know that the node.rotation is a quaternion and not an euler angle? (perhaps only quaternion are supported?)

deltakosh:
Ping @julien-moreau

Misterblue:
The GLTF 1.0 specification (https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#reference-node) says node "rotation" is a quaternion.

julien-moreau:
I think you are right, just going to test and approve :)
BRB

deltakosh:
Bingo!

PR #1908, opened 2017-03-17
Ability to choose the source web cam using videoId for CreateFromWebCam

avin:
ref: http://www.html5gamedevs.com/topic/28995-webrtc-support-for-videotexture/


deltakosh:
Thanks!

PR #1919, opened 2017-03-22
Webvr parenting

RaananW:
Parenting system of the cameras works correctly.

Each rig camera is a child of the main camera. This is done to avoid changing the view matrix genarated so kindly by the camera.

There is now an option to add meshes as children of the WebVR camra.
The parameter "rigParenting" will decide whether the objects stay attached to the main camera (and thus not being followed when moving the head) ot that they should be attached to the rig cameras and move together with the head rotation.

### DIFF
--- a/src/Cameras/VR/babylon.webVRCamera.ts
+++ b/src/Cameras/VR/babylon.webVRCamera.ts
@@ -52,7 +52,7 @@ module BABYLON {
 
         private _positionOffset: Vector3 = Vector3.Zero();
 
-        protected _decedents: Array<Node> = [];
+        protected _descendants: Array<Node> = [];
 
         public devicePosition = Vector3.Zero();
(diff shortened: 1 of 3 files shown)


src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -52,13 +52,17 @@ module BABYLON {
 
         private _positionOffset: Vector3 = Vector3.Zero();
 
+        protected _decedents: Array<Node> = [];
```
deltakosh: descendants ?

RaananW: Was checking to see if anyone will notice. Thumbs up! :smile:

deltakosh: LOL!

Issue #1925, opened 2017-03-24
Set levelVisible to false causes the renderer to crash


nockawa:
http://www.html5gamedevs.com/topic/29099-cannot-read-property-isdisposed-of-undefined/#comment-168464

Repro case: http://babylonjs-playground.com/#7VXUD#2

abow:
@deltakosh I think this can be closed.

deltakosh:
Thanks :)

PR #1928, opened 2017-03-24
Fixing rendering crash when a last prim was switching from hidden to visible

nockawa:
@wyklion you're bug that you couldn't repro anytime is finally fixed! 

#1925


wyklion:
I'll try the new version this week!
Finally stop overwork from now on!

PR #1930, opened 2017-03-24
Add support for two-sided lighting

bghgary:
This adds support for two-sided lighting but also fixes bugs with one-sided lighting when backfaceCulling is false.


deltakosh:
excellent
do you mind updating the what's new for 3.0?

bghgary:
Oh yeah, keeping forgetting to do that.

PR #1973, opened 2017-04-06
Move to the new API.

Jaskar:
Add a save form to add Name / Description / Tags


deltakosh:
please also update index-debug.html

PR #1982, opened 2017-04-07
Performance file to custom-build

Palmer-JC:
Only works for Firefox.  Directions on page. Right click on profile.html
file => Run with => FireFox

Some interesting dependencies, like you must have all lights.  See html
for why's.


deltakosh:
Do you think this could be done as part of the build process?

Palmer-JC:
Probably not, or at least not me as I do not know enough about Gulp / JSON.  Even if you could pass the file and Generate directly, I found I had to use the option to do special includes for my extensions.  QI subclasses Bone & Skeleton.  My scene did not use that feature, so it did not show in performance data.

Think it is a great aid.  I would have no clue what to cut out without this and give up / be way to conservative without it.  What makes matter worst, it is hard to un-delete lines in a file manually.  This tool generates the lines from the regular config.

deltakosh:
Makes sense!
Thanks for that

deltakosh:
(please update the what's new)

deltakosh:
I think you should also update this to add a mention about your tool:
https://github.com/BabylonJS/Documentation/blob/master/content/generals/General/Creating_the_Mini-fied_Version.md

PR #1985, opened 2017-04-09
Add Camera Toggle

sbtron:
Add Camera button to toggle between FreeCamera and ArcRotateCamera.
glTF files default to ArcRotateCamera. Works with glTF 1.0 files.
To make page mobile friendly help text is just shown/hidden instead of sliding in.


sandbox/index.js
```diff
@@ -68,6 +71,15 @@ if (BABYLON.Engine.isSupported()) {
                 currentScene.activeCamera.keysRight.push(69); // E
                 currentScene.activeCamera.keysRight.push(68); // D
             }
+            defaultCamera = currentScene.activeCamera;
+            arcRotateCamera = new BABYLON.ArcRotateCamera("camera", 4.71238898038469, 1.5707963267948966, 3, BABYLON.Vector3.Zero(), currentScene);
```
deltakosh: I'm wondering if the radius could be detected.
Like the same way we do this: https://github.com/BabylonJS/Babylon.js/blob/master/src/babylon.scene.ts#L3467

deltakosh:
Ping?

deltakosh:
Thanks!

PR #2001, opened 2017-04-12
Uniform Buffer Objects

CraigFeldspar:
Since this is an important engine change, I open the pull request now for early review (in case I'm going in the wrong way with data structures).

Implementation notes : 

Only StandardMaterial has been ported with UBO for now.
UBO strategy is the following : 
 + 1 UBO for the whole material parameters
 + 1 UBO for each light
 + 1 UBO for the scene transform matrix, that is shared amongst programs

default.vertex.fx and default.fragment.fx show a huge uniform block declaration, which can be scary since there is no #DEFINE. This is done on purpose, since a material will have the same UBO regardless of the mesh it's applied on. For instance, we can apply the same material with bumpTexture on a mesh which has normal information and on a mesh that doesn't. So 2 different shaders will be used, but only 1 UBO, therefore this 2 shaders must adapt to that unique UBO layout, even if the 2d one won't use the part where bump uniforms are declared. This is not a performance issue either, because defines remains in the shader code, therefore shader instructions are unchanged, they will just fetch memory on different addresses.

std140 layout has been used for uniform blocks to allow ubo sharing amongst programs (scene transform matrix for example)

Frozen materials allow a huge optimisation, since the engine will not update the uniform buffer on rendering, just bind it on the binding point. Unfrozen materials still use uniform buffers, but update them on each bind (with the same cache system used on the current non-ubo implementation).

BABYLON.UniformBuffer can be declared as static or dynamic. Static implies a WebGLBuffer that is declared as gl.STATIC_DRAW, and a cache verification prevents useless webgl calls, as discussed above. Dynamic implies frequent updates, therefore, cache verification is disabled and the underlying WebGLBuffer is declared as gl.DYNAMIC_DRAW.

Lights have their uniform buffer declared as static by default, but this is not optimal for moving lights. A nice optimisation could be the possibility to declare a light static or dynamic, thus skipping most pain with uniform cache when binding them.

The scene's transformMatrix (viewProjection in shaders) has its own SHARED uniform buffer. setTransformMatrix in Scene class triggers the UBO update, and this UBO is only bound **once** at the program creation, preventing recurrent binding on render.

TODO : 
- [x] Documentation
- [x] WebGL2 Typescript declarations (gl.UNIFORM_BUFFER)
- [x] WebGL1 Backwards compatibility
- [x] PBR port to UBO
- [x] Non-Regression tests
- [x] BABYLON.Uniform class completion with all types of uniform (mat3 etc), and check that std140 is still respected
- [x] View matrix can go inside scene UBO (for defaultMaterial)
- [x] Full Material Library compatibility with UBO

TODO for further optimisation : 
- [ ] Lights should be in push mode too, with the possibility to declare them static. Then BindLights can be deleted from the UBO version of `xxxMaterial.bind` (and lights UBO can be bound on the shader compilation, like the scene UBO)
- [ ] Uniform bases can be handled better, by reserving slots for most common uniforms (like Scene, Material, LightX etc). Is it impacting the performance a lot though ?


deltakosh:
hello!! this is really great!!!
Some questions: 
- can you measure perd improvement?
- with my recent changes to matrices you can know if a matrix need to be updated. this is useful I guess at least for viewprojection (which can change during the frame because of mirrors or renderrtargets)
- The change on createEffect should not affect current signature. Can you move the new parameter at the end?

src/Materials/babylon.uniformBuffer.ts
```diff
@@ -0,0 +1,268 @@
...
+        public updateFloat4(name: string, x: number, y: number, z: number, w: number) {
+            this._cache[0] = x;
+            this._cache[1] = y;
+            this._cache[2] = z;
+            this._cache[3] = w;
+            this.updateUniform(name, this._cache, 4);
+        }
+
+        public updateMatrix(name: string, mat: Matrix) {
+            this.updateUniform(name, mat.toArray(), 16);
```
deltakosh: on my phone but need to make sure the toArray does not instantiate an array

deltakosh:
for createEffect I actually suggest to add an "overloaded" parameter named option. this is what we did for mesh creation where you can call CreateSphere with a long list of parameter or with just a few including a big fat options one:)

Do you mind sending me an email at davca at microsoft dot com? Pretty excited by what you did here

deltakosh:
I updated your initial message to add checkboxes :D

CraigFeldspar:
I think this PR is complete now. 2 PR will follow this one : 1) changing lights to push mode and 2) converting material library to UBO (only compatibility has been done in this one).

Tests performed : 
 - Full Playground presets on Chrome (WebGL 2) and Edge (WebGL1) : OK
 - Full Material Library on Chrome and Edge : OK
 - Full PostProcess Library on Chrome and Edge : OK
 - Full Procedural Textures Library on Chrome and Edge : OK

Regarding performance, we are pretty much equal on simple materials (btw you did great work with push materials and uniform caching ! isReady and bind are very fast even without UBO), and we start to diminish greatly the CPU overhead on more complex and diverse materials - like PBR - also as we start to add more lights.

As we discussed before, maybe more things could be moved into UBO, I'm open to every idea !

deltakosh:
sounds perfect. Did you run the validation tests from vs code?

deltakosh:
other question: regarding backward compatibility, we can still use uniform directly? so people having their own material won't have an issue?

CraigFeldspar:
regarding tests, I did not, I used my own local webserver & used gulp tasks. Is there something specific to do with VSCode ?

Regarding backward compatibility, both ways are still supported. Even StandardMaterial still uses both ways. 
Users should only be careful if they are including the lightsFragment include, which has been converted to UBO, so they need to declare uniform buffers in their isReady function (see changes on the files like waterMaterial.ts)

deltakosh:
if you go to debug pane in vs code you can run th validation tests. you can also run it directly from tests/validation folder

materialsLibrary/src/lava/babylon.lavaMaterial.ts
```diff
@@ -206,25 +206,40 @@ module BABYLON {
                 // Legacy browser patch
                 var shaderName = "lava";
                 var join = defines.toString();
+
+                var uniformBuffers = [];
```
deltakosh: I'm wondering if this could be moved to the MaterialHelper

materialsLibrary/src/shadowOnly/babylon.shadowOnlyMaterial.ts
```diff
@@ -107,12 +107,29 @@ module BABYLON {
...
                 ];
                 var samplers = [];
-                    
-                MaterialHelper.PrepareUniformsAndSamplersList(uniforms, samplers, defines, 1);
+                
+                var uniformBuffers = [];
+
+                MaterialHelper.PrepareUniformsAndSamplersList(<EffectCreationOptions>{
+                    uniformsNames: uniforms, 
+                    uniformBuffersNames: uniformBuffers,
```
deltakosh: If this is = [] you should not mention it (Apply everywhere :))

src/Materials/babylon.effect.ts
```diff
@@ -595,6 +634,18 @@
             return changed;
         }
 
+        public bindUniformBuffer(buffer: WebGLBuffer, name: string): void {
+            if (Effect._baseCache[this._uniformBuffersNames[name]] === buffer) {
```
deltakosh: wondering if we can avoid the double array with Effect._baseCache[name]

src/Materials/babylon.material.ts
```diff
@@ -438,6 +441,26 @@
...
+        public bindView(effect: Effect): void {
+            if (this.getScene().getEngine().webGLVersion === 1) {
+                effect.setMatrix("view", this.getScene().getViewMatrix());
+            } else {
+                this.bindSceneUniformBuffer(effect, this.getScene().getSceneUniformBuffer());
+            }
+        }
+
+        public bindViewProjection(effect: Effect): void {
+            if (this.getScene().getEngine().webGLVersion === 1) {
```
deltakosh: we should save this value once and for all

src/Materials/babylon.materialHelper.ts
```diff
@@ -298,48 +317,39 @@
...
-            }
+            light.transferToEffect(effect, lightIndex);
         }
 
         public static BindLights(scene: Scene, mesh: AbstractMesh, effect: Effect, defines: MaterialDefines, maxSimultaneousLights = 4) {
             var lightIndex = 0;
             var depthValuesAlreadySet = false;
 
             for (var light of mesh._lightSources) {
+                var useUbo = light._uniformBuffer.useUbo;
```
deltakosh: This should be handled by the uniformBuffer itself
 and then call could be like:
light._uniformBuffer.updateColor4("vLightDiffuse", lightIndex, Tmp.Color3[0], light.range);

where lightIndex is used if there is no UBO

src/Shaders/ShadersInclude/defaultFragmentDeclaration.fx
```diff
@@ -0,0 +1,80 @@
+uniform vec4 vDiffuseColor;
```
deltakosh: just a reminder: i changed the build mechanism a bit so you will need to reference these includes for both pbrmaterial and stdmaterial in their config section :)

src/babylon.engine.ts
```diff
@@ -1305,6 +1305,64 @@
             this.wipeCaches();
         }
 
+        // UBOs
+        public createUniformBuffer(elements: number[] | Float32Array): WebGLBuffer {
+            var ubo = this._gl.createBuffer();
+            this.bindUniformBuffer(ubo);
+
+            if (elements instanceof Float32Array) {
+                this._gl.bufferData((<WebGL2RenderingContext>this._gl).UNIFORM_BUFFER, <Float32Array>elements, this._gl.STATIC_DRAW);
```
deltakosh: To avoid this I recommend updating the basic WebGLRenderingContext like I did for other webgl2 features (see mixins.ts files)

src/babylon.scene.ts
```diff
@@ -682,6 +682,8 @@
         public _activeAnimatables = new Array<Animatable>();
 
         private _transformMatrix = Matrix.Zero();
+        private _sceneUbo: UniformBuffer;
```
deltakosh: the UniformBuffer file has to be added to Core workload in the config file

src/Materials/babylon.uniformBuffer.ts
```diff
@@ -0,0 +1,584 @@
+module BABYLON {
+
+    // Pool for avoiding memory leaks
+    var MAX_UNIFORM_SIZE = 256;
```
deltakosh: should not be on BABYLON namespace but more as a private const inside the buffer

src/Materials/babylon.uniformBuffer.ts
```diff
@@ -0,0 +1,584 @@
...
+        private _data: number[];
+        private _bufferData: Float32Array;
+        private _dynamic: boolean;
+        private _uniformName: string;
+        private _uniformLocations: { [key:string]:number; };
+        private _uniformSizes: { [key:string]:number; };
+        private _uniformLocationPointer: number;
+        private _needSync: boolean;
+        private _cache: Float32Array;
+        private _noUbo: boolean;
```
deltakosh: generaly speaking three letters name should be noUBO (see HDR for instance)

src/Materials/babylon.uniformBuffer.ts
```diff
@@ -0,0 +1,584 @@
...
+         * Handles blocks of uniform on the GPU.
+         *
+         * If WebGL 2 is not available, this class falls back on traditionnal setUniformXXX calls.
+         *
+         * For more information, please refer to : 
+         * https://www.khronos.org/opengl/wiki/Uniform_Buffer_Object
+         */
+        constructor(engine: Engine, data?: number[], dynamic?: boolean) {
+            this._engine = engine;
+            this._noUbo = engine.webGLVersion === 1;
```
deltakosh: As this will be heavily used here is an idea:
we declare container function like this:
public updateMatrix3x3: (name: string, matrix: Float32Array) => Effect;

And we create two functions:
private _updateMatrix3x3ForUniform(name: string, matrix: Float32Array): Effect {
}
private _updateMatrix3x3ForEffect(name: string, matrix: Float32Array): Effect {
}


then in the constructor based on _noUBO value we affect one of the private function to the public one. So direct call, no test

materialsLibrary/src/shadowOnly/babylon.shadowOnlyMaterial.ts
CraigFeldspar: It is necessary here though, even if it is [], since `PrepareUniformsAndSamplersList` can populate this array.

src/Materials/babylon.effect.ts
CraigFeldspar: Here we cannot avoid it, since uniform buffers have binding points (see https://paroj.github.io/gltut/Positioning/UBOContextBinding.svg). Those binding points can host different uniform buffers name, so the cache must carry the index, and not the name.

I wondered too if we could set a binding point to always carry the same uniform block name, and in some cases where the UBO is heavily shared (like scene UBO or lights), this is definitely the way to go. But since a GPU usually carries less than 100 binding points, I prefer to leave the full liberty of binding for now. A PR for better handling of binding points is in my TODO list though :)

jbousquie:
Guys, I hardly understand what you're doing, but it's fascinating (I would even add "ça poutre sa race velue") and I like reading all this progess during my BJS break. Please keep on this way :-D

src/Shaders/ShadersInclude/defaultFragmentDeclaration.fx
CraigFeldspar: done

src/babylon.scene.ts
CraigFeldspar: done

CraigFeldspar:
everything fixed for this pass, we can move forward :)

src/Materials/babylon.effect.ts
deltakosh: sounds good

materialsLibrary/src/shadowOnly/babylon.shadowOnlyMaterial.ts
deltakosh: `PrepareUniformsAndSamplersList` could create it if undefined. I would like to reduce the code required to use it

src/Materials/babylon.materialHelper.ts
```diff
@@ -317,37 +316,35 @@
...
-                light._uniformBuffer.updateFloat3(useUbo ? "shadowsInfo" : "shadowsInfo" + lightIndex, shadowGenerator.getDarkness(), shadowGenerator.blurScale / shadowGenerator.getShadowMap().getSize().width, shadowGenerator.depthScale);
+                light._uniformBuffer.updateFloat3("shadowsInfo", shadowGenerator.getDarkness(), shadowGenerator.blurScale / shadowGenerator.getShadowMap().getSize().width, shadowGenerator.depthScale, lightIndex);
             }
 
             return depthValuesAlreadySet;
         }
 
         public static BindLightProperties(light: Light, effect: Effect, lightIndex: number): void {
-            light.transferToEffect(effect, lightIndex);
+            light.transferToEffect(effect, lightIndex + "");
```
deltakosh: why this?

deltakosh:
Let's MERGE :)

deltakosh:
Congrats my friend this is a WONDERFUL job
Please credit yourself in the what's new!

PR #2002, opened 2017-04-12
Custom Build refinements

Palmer-JC:
- Changed if test to use getClassName() instead of instanceof
- Improved the directions of the page
- Added a output directory text field on page

### DIFF
--- a/src/Materials/babylon.materialHelper.ts
+++ b/src/Materials/babylon.materialHelper.ts
@@ -111,11 +111,11 @@
                     defines["DIRLIGHT" + lightIndex] = false;
 
                     var type;
-                    if (light instanceof SpotLight) {
+                    if (light.getClassName() === "SpotLight") {
                         type = "SPOTLIGHT" + lightIndex;
-                    } else if (light instanceof HemisphericLight) {
+                    } else if (light.getClassName() === "HemisphericLight") {
                         type = "HEMILIGHT" + lightIndex;
-                    } else if (light instanceof PointLight) {
+                    } else if (light.getClassName() === "PointLight") {
                         type = "POINTLIGHT" + lightIndex;
                     } else {
                         type = "DIRLIGHT" + lightIndex;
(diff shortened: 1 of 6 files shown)


src/babylon.engine.ts
```diff
@@ -3362,26 +3359,32 @@
 
         // Loading screen
         public displayLoadingUI(): void {
+            if (!this._loadingScreen) this._loadingScreen = new DefaultLoadingScreen(this._renderingCanvas)
```
deltakosh: what about creating a loadingScreen property? this way you could instantiate it on first get?

src/Mesh/babylon.subMesh.ts
```diff
@@ -89,7 +89,7 @@
         public getMaterial(): Material {
             var rootMaterial = this._renderingMesh.material;
 
-            if (rootMaterial && rootMaterial instanceof MultiMaterial) {
+            if (rootMaterial && rootMaterial.getClassName() === "MultiMaterial") {
```
deltakosh: for performance reason we should just test for getSubMaterial existence

src/Materials/babylon.materialHelper.ts
```diff
@@ -111,11 +111,11 @@
                     defines["DIRLIGHT" + lightIndex] = false;
 
                     var type;
-                    if (light instanceof SpotLight) {
+                    if (light.getClassName() === "SpotLight") {
```
deltakosh: just wondering if string testing is a good idea regarding perd. Perhaps we could introduce a function  to return the type directly

src/Materials/babylon.materialHelper.ts
```diff
@@ -305,16 +305,16 @@
         }
 
         public static BindLightProperties(light: Light, effect: Effect, lightIndex: number): void {
-            if (light instanceof PointLight) {
+            if (light.getClassName() === "PointLight") {
```
deltakosh: we also should do it at light level to avoid string testing

src/babylon.engine.ts
Palmer-JC: Actually there already was one.  Done in next PR

src/Mesh/babylon.subMesh.ts
Palmer-JC: Had to cast to MultiMaterial to get getSubMaterial existence test to work, but casting is just typescript

deltakosh: yep so no perf issues

src/Materials/babylon.materialHelper.ts
Palmer-JC: for light tests, I now use getTypeID instead.  It already exists & is an integer.  Think I need to close this PR and open a new one to add another commit.

deltakosh: Ok let me merge this one then

deltakosh:
I've just did the change in subMesh as I was working on the file already

PR #999, opened 2016-02-26
Pmrem generator

sebavan:
SeamLess Cube Map + Convolve Cosine Blur (CubeMapgen Process)

PR #2004, opened 2017-04-13
Playground load

Jaskar:
When user doesn't specify an iteration (link like /#123456) , get the first iteration (/#123456#0) saved, not the last. Asked by @davrous


aWeirdo:
Sorry to bump this up, but i was thinking,

If you increase the "iteration" to a higher number than there is saved/available,
right now it will redirect to "http://www.babylonjs-playground.com/"
why not redirect to the highest available?

E.g, 
If you go to;
http://www.babylonjs-playground.com/#JZRDRD#1 
Increase it from #1 to #2 & update and it will redirect you to http://www.babylonjs-playground.com/

Thought it could be a nice and quick way to find the newest update to a PG by just having to add #999 or something and it would redirect you to the highest available for the PG.

deltakosh:
really like the idea! Fancy doing a PR?

aWeirdo:
I don't know much about the PG code tbh, but i took a quick look at the publicly available files,
However i don't see any back-end files anywhere and I think it would be most optimal to do it when querying for a snippet, 
so when querying, if the version is higher than 0 but doesn't exist, return highest available.

deltakosh:
the code is here: https://github.com/BabylonJS/Documentation-server

PR #2015, opened 2017-04-16
Playground - design rework

Temechon:
New Playground design, with : 
* Two new themes - Light and dark
* Responsive design - On small screen, the code is hidden
* The same features than before :)

Let me know if something is not ok !


Temechon:
Let me some time to update frame.html, local.html and debug.html

Temechon:
OK should be good now :)

PR #2059, opened 2017-04-24
Export multiple meshes and use de global position

felipebaltazar:
is possible export multiple meshes to obj, fot that reason i create a loop to get all meshes in array.


deltakosh:
Indentation is not good:)

deltakosh:
Sorry to bother you:( but indent level should be 4 and not 8

deltakosh:
Try with vscode perhaps?

PR #2064, opened 2017-04-25
Helper functions

RaananW:
Get left and right camera and target, allowing flexibility for the
developer.
Also - the callback is not set using a setter that checks if the
controllers were already initialized. So even if the callback is added
later, it will be called if the controllers are already set.


deltakosh:
Thanks mate!

PR #2066, opened 2017-04-26
Adding support to Babylon.JS for morph targets

BeardedGnome:
1) Fixes a compiler error in serializers (capitalization of a memeber)
2) Adds support to morph targets to have tangent data

### DIFF
--- a/src/Shaders/ShadersInclude/morphTargetsVertex.fx
+++ b/src/Shaders/ShadersInclude/morphTargetsVertex.fx
@@ -3,5 +3,10 @@
 	
 	#ifdef MORPHTARGETS_NORMAL
 	normalUpdated += (normal{X} - normal) * morphTargetInfluences[{X}];
-	#endif	
+	#endif
+
+	#ifdef MORPHTARGETS_TANGENT
+	tangentTemp = tangentUpdated.xyz + ((tangent{X} - tangent.xyz) * morphTargetInfluences[{X}]);
+	tangentUpdated = vec4(tangentTemp, tangentUpdated.w);
+	#endif
 #endif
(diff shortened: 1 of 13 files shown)


src/Shaders/ShadersInclude/morphTargetsVertex.fx
```diff
@@ -3,5 +3,10 @@
 	
 	#ifdef MORPHTARGETS_NORMAL
 	normalUpdated += (normal{X} - normal) * morphTargetInfluences[{X}];
-	#endif	
+	#endif
+
+	#ifdef MORPHTARGETS_TANGENT
+	tangentTemp = tangentUpdated.xyz + ((tangent{X} - tangent.xyz) * morphTargetInfluences[{X}]);
```
deltakosh: This seems suspicious.
why not tangentUpdated += (tangent{X} - normal) * morphTargetInfluences[{X}];

deltakosh: Because if we have have 4 targets the final formula:

tangentTemp = 4 * tangentUpdated.xyz + sum(((tangent{X} - tangent.xyz) * morphTargetInfluences[{X}]));

deltakosh: Why introducing a tangentTemp ?

BeardedGnome: Tangent from morph targets is a vec3 delta on xyz, not on w. w should stay -1/+1. Is there a better way to do that?

deltakosh: something like:
```
tangentUpdated.xyz += (tangent{X} - tangent.xyz) * morphTargetInfluences[{X}];
```

deltakosh:
Thanks a lot!

PR #2067, opened 2017-04-26
Adding support to glTF 2.0 loader for morph targets

BeardedGnome:
@howardwolosky
@bghgary 

This builds on the previous PR that adds support for morph target tangents.

### DIFF
--- a/loaders/src/glTF/2.0/babylon.glTFLoader.ts
+++ b/loaders/src/glTF/2.0/babylon.glTFLoader.ts
@@ -508,7 +508,7 @@ module BABYLON.GLTF2 {
         var indexStarts = [];
         var indexCounts = [];
 
-        var morphTargetManager;
+        var morphTargetManager = new BABYLON.MorphTargetManager();
 
         // Positions, normals and UVs
         for (var primitiveIndex = 0; primitiveIndex < mesh.primitives.length; primitiveIndex++) {
@@ -598,8 +598,6 @@ module BABYLON.GLTF2 {
 
             // Morph Targets
             if (primitive.targets !== undefined) {
-                morphTargetManager = new BABYLON.MorphTargetManager();
-
                 for (var targetsIndex = 0; targetsIndex < primitive.targets.length; targetsIndex++) {
                     var target = primitive.targets[targetsIndex];
 
@@ -674,7 +672,7 @@ module BABYLON.GLTF2 {
         babylonMesh.computeWorldMatrix(true);
 
         // Set morph target manager after all vertices data has been processed
-        if (morphTargetManager !== undefined && (<MorphTargetManager>morphTargetManager).numInfluencers != 0) {
+        if (morphTargetManager !== undefined && morphTargetManager.numInfluencers > 0) {
             babylonMesh.morphTargetManager = morphTargetManager;
         }
 
--- a/loaders/src/glTF/2.0/babylon.glTFLoader.ts
+++ b/loaders/src/glTF/2.0/babylon.glTFLoader.ts
@@ -511,11 +511,11 @@ module BABYLON.GLTF2 {
         var morphTargetManager;
 
         // Positions, normals and UVs
-        for (var index = 0; index < mesh.primitives.length; index++) {
+        for (var primitiveIndex = 0; primitiveIndex < mesh.primitives.length; primitiveIndex++) {
             // Temporary vertex data
             var tempVertexData = new VertexData();
 
-            var primitive = mesh.primitives[index];
+            var primitive = mesh.primitives[primitiveIndex];
             if (primitive.mode !== EMeshPrimitiveMode.TRIANGLES) {
                 // continue;
             }
@@ -581,8 +581,8 @@ module BABYLON.GLTF2 {
             else {
                 // Set indices on the fly
                 var indices: number[] = [];
-                for (var j = 0; j < tempVertexData.positions.length / 3; j++) {
-                    indices.push(j);
+                for (var positionsIndex = 0; positionsIndex < tempVertexData.positions.length / 3; positionsIndex++) {
+                    indices.push(positionsIndex);
                 }
 
                 tempVertexData.indices = new Int32Array(indices);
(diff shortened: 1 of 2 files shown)


loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -594,6 +596,70 @@ module BABYLON.GLTF2 {
             var material = getMaterial(runtime, primitive.material);
             multiMat.subMaterials.push(material);
 
+            // Morph Targets
+            if (primitive.targets !== undefined) {
+                morphTargetManager = new BABYLON.MorphTargetManager();
+
+                for (var index2 = 0; index2 < primitive.targets.length; index2++) {
```
HowardWolosky: is using `index2` and `index3`, etc... a common pattern throughout the Babylon codebase?  For smaller loops, anonymous variable names like `i` / `j` (or `index` / `index2`) seem reasonable, but once we get to larger loops like this, it seems that the code legibility could be improved with a more explicit name, like `morphTargetIndex` or something similar...

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -594,6 +596,70 @@ module BABYLON.GLTF2 {
...
+
+                    var weight = 0.0;
+                    if (node.weights !== undefined) {
+                        weight = node.weights[index2];
+                    }
+                    else if (mesh.weights !== undefined) {
+                        weight = mesh.weights[index2];
+                    }
+
+                    var morph = new BABYLON.MorphTarget("morph" + index2, weight);
```
HowardWolosky: It might be worthwhile to _try_ and get a reasonable name for the morph target from the file.  Right now, the only way to do that would be to grab the (optional) `name` from the `accessor` of one of the three `semantic`s.  Using a generated name like this seems like a reasonable fallback.  Can the name be changed after the `BABYLON.MorphTarget` has been created?  If not, then you might need to _set_ the data on `morph` outside of the loop.

deltakosh:
I'll just wait for @bghgary to validate it :)

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: It was used previously in this method, so I kept the style consistent.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: `name` is public, so I can set it each time we loop for each access if it has a name.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: I don't see any other index2 variables in the file. Did you mean previously in history?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -594,6 +596,70 @@ module BABYLON.GLTF2 {
...
+                    for (var semantic in target) {
+                        // Link accessor and buffer view
+                        accessor = runtime.gltf.accessors[target[semantic]];
+                        buffer = GLTFUtils.GetBufferFromAccessor(runtime, accessor);
+
+                        // glTF stores morph target information as deltas
+                        // while babylon.js expects the final data.
+                        // As a result we have to add the original data to the delta to calculate
+                        // the final data.
+                        if (semantic.toUpperCase() === "NORMAL") {
```
bghgary: Remove toUpperCase. glTF defines these to all be upper case and importMesh doesn't do this.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -594,6 +596,70 @@ module BABYLON.GLTF2 {
...
+                            var normals = new Float32Array(buffer.length);
```
bghgary: This extra allocation of a Float32Array isn't necessary. `buffer` is already a Float32Array. You should be able to just cast it.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
HowardWolosky: He did this to support thr exporter.  It's an artifact of a covariance issue I couldn't work around.  We can discuss it more tomorrow.  The short of it was I had to define target within the schema along with the three attributes, but doing so causes them to be camelCased on export due to the export rules.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: Sorry, I meant the name `index` instead of `i` or `j`.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: Here's what the exporter produces:

`          "targets": [
            {
              "position": 4,
              "normal": 3
            },
            {
              "position": 6,
              "normal": 5
            }
          ]`

Until that changes, we require `ToUpperCase`.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: I agree with Howard on this. Use better names instead of index2, etc...

BeardedGnome: Better, like smaller (j, k, l), so it's faster to load or better, like bigger (targetIndex, bufferIndex, etc) for clarity?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
HowardWolosky: Also, to be fair, the schema does _not_ define these to be uppercase.  [The schema](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#indices-and-names) does indicate that property names should be _camelCase_, but it doesn't say anywhere that the attributes should be _UPPERCASE_.

bghgary: The [morph target section](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#morph-targets) explicit states what values are supported. The [schema/property reference](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#primitivetargets) also explicitly calls this out. All values must be uppercase.

HowardWolosky: Looks like the schema/json should be updated to clarify that those values are case-sensitive, and then I can add another iteration to the [Exporter PR](https://github.com/bghgary/glTF-Tools-for-Unity/pull/9) that reads-in the .glTF file once it has been exported and just does a string-replace on those attributes to force them to upper-case.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -581,8 +581,8 @@ module BABYLON.GLTF2 {
             else {
                 // Set indices on the fly
                 var indices: number[] = [];
-                for (var j = 0; j < tempVertexData.positions.length / 3; j++) {
-                    indices.push(j);
+                for (var positionsIndex = 0; positionsIndex < tempVertexData.positions.length / 3; positionsIndex++) {
```
bghgary: positionsIndex should probably be just index?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -670,16 +674,18 @@ module BABYLON.GLTF2 {
         babylonMesh.computeWorldMatrix(true);
 
         // Set morph target manager after all vertices data has been processed
-        babylonMesh.morphTargetManager = morphTargetManager;
+        if (morphTargetManager !== undefined && (<MorphTargetManager>morphTargetManager).numInfluencers != 0) {
```
bghgary: change the declaration from `var morphTargetManager` to `var morphTargetManager: MorphTargetManager` so that you don't need this cast.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -600,63 +600,67 @@ module BABYLON.GLTF2 {
             if (primitive.targets !== undefined) {
                 morphTargetManager = new BABYLON.MorphTargetManager();
```
bghgary: I'm a bit confused. Is it intended that you are creating multiple MorphTargetManager objects per mesh?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: Didn't you just ask for a more expressive variable name than `index`?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: Good point. With the additional checks before assigning it to the mesh, I can move the object creation out a level.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: Yes, but this one is not a positionsIndex :-)

PR #2079, opened 2017-05-01
Added support to glTF loader for animated morph targets

BeardedGnome:
To accomplish this I needed two things:
1) Fix a bug in ImportMesh to add no only the active morph targets, but also the inactive ones.
2) Update ImportAnimation to handle the morph animations. This made the import code more complicated because each 'weight' animation can target up to 16 morph targets and is required to provided data for all available morph targets, even ones that aren't animated.

This also required two new getters in Babylon.

@howardwolosky
@bghgary


loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -113,19 +120,30 @@ module BABYLON.GLTF2 {
...
-                for (var j = 0; j < bufferInput.length; j++) {
+                for (var frameIndex = 0; frameIndex < bufferInput.length; frameIndex++) {
                     var value: any = null;
 
                     if (targetPath === "rotationQuaternion") { // VEC4
                         value = Quaternion.FromArray([bufferOutput[arrayOffset], bufferOutput[arrayOffset + 1], bufferOutput[arrayOffset + 2], bufferOutput[arrayOffset + 3]]);
                         arrayOffset += 4;
                     }
+                    else if (isMorph) { // FLOAT
+                        value = [];
```
bghgary: Instead of push, you can do this.

```
value = new Array<number>(numTargets);

for (...) {
    value[influence] = ...
}

Save some allocations?

BeardedGnome: I wasn't sure if operator[] would create new array elements. I'm used to STL where it would throw. I could find a clear answer when I was looking, so went with the known good option. Do you know what will happen with operator[]?

deltakosh:
Thanks!

loaders/src/glTF/2.0/babylon.glTFLoader.ts
BeardedGnome: value[] is slightly faster than value.push() on modern browsers. I didn't find any other places I could change push to [].

BeardedGnome: Oops. Looks like it's already merged.

PR #2085, opened 2017-05-02
Fix lazy variable initialization bug in animatable class

bghgary:
This causes synchronization issues with other custom code that requires timings to be accurate.


deltakosh:
thanks!

PR #2088, opened 2017-05-03
Standard Rendering pipeline now computes luminance adaptation

julien-moreau:
Set an empty string when serialized lens flare has no texture


julien-moreau:
Doc is coming :)

PR #2112, opened 2017-05-08
Adjustment to position of CannonJS HeightmapImpostor

Wingnutt:
I hope this works!  :)


src/Physics/Plugins/babylon.cannonJSPlugin.ts
```diff
@@ -375,9 +372,9 @@
                 var translation = mesh.getBoundingInfo().boundingBox.center.subtract(center).subtract(mesh.position).negate();
 
                 this._tmpPosition.copyFromFloats(translation.x, translation.y - mesh.getBoundingInfo().boundingBox.extendSize.y, translation.z);
-                //add it inverted to the delta 
-                this._tmpDeltaPosition.copyFrom(mesh.getBoundingInfo().boundingBox.center.subtract(c));
-                this._tmpDeltaPosition.y += mesh.getBoundingInfo().boundingBox.extendSize.y;
+                // comments needed - wingnut removes a .subtract(c) and then adjusts z
+                this._tmpDeltaPosition.copyFrom(mesh.getBoundingInfo().boundingBox.center);
+                this._tmpDeltaPosition.z = -mesh.getBoundingInfo().boundingBox.extendSize.z;
```
dad72: Hi,
I would have -= mesh... rather than = -mesh...

Wingnutt: Not same result, Dad72.
http://www.babylonjs-playground.com/#1RKZXB#34
(lines 43/44)  
Using -= causes mesh to sink into ground, and also no physics active on far-half of terrain.

RaananW:
Hi guys,

sorry for a late reply as always :smile:
Can you explain what was wrong with the old implementation and why you think this is the right way to go?
Cannon has z as the up vector, which is not needed until you get to very specific object types like the one we are handling here. Any playground you want to show me?

src/Physics/Plugins/babylon.cannonJSPlugin.ts
Wingnutt: Uh oh, I think I have more trouble.  I forgot to test non-square terrains.  
600 x 200 terrain fails with cannonJS error.
200 x 600 fails in my impostor positioning.
CRAP CRAP CRAP!  *sigh*
Back to the drawing board, I suppose.

Wingnutt:
Hi R, thx for asking.  I guess it was [#8 in the series](http://www.babylonjs-playground.com/#1RKZXB#8) that was first seen as an issue.  Camera aiming +x (I changed to +z aimed cam soon after).

Notice that physics impostor only active in southwest quadrant of terrain.  Notice impostor extends southwest BEYOND edges of terrain, too.  It is a 500x500 terrain, and I believe the impostor need repositioning 250 units north (+x) and 250 units east (+z)  (or +y if cannon heightmap has z vertical)

Thank god you're here!  I can use all the help I can get.  But don't let this stand in the way of real life stuff, R.  I think boundingBox.extendSize changed from returning DIAMETER-like values... to returning radius-like values (half of what they once were).  But I have no proof of that, and nobody is saying I am correct.  So, likely I'm wrong.  :)  Thanks for advice/input, R.  Sorry if I'm causing you hassles, or lost on wild goose chase.

Wingnutt:
DOA

Issue #2077, opened 2017-04-30
Change media query on playground to move features inside settings dropdown instead of just hiding them


deltakosh:

Jaskar:
https://github.com/BabylonJS/Babylon.js/pull/2120

PR #2120, opened 2017-05-10
Issue 2077 / Switch fullscreen to fullpage

Jaskar:
https://github.com/BabylonJS/Babylon.js/issues/2077
Plus, switch from Fullscreen to Fullpage, because of the Chrome weird fullscreen.
[Only on index.html for now. If it's ok, I'll modify other .html within the day]


Temechon:
The behaviour is not perfect yet. 

Can you try this:
1. Put the canvas in full page like you did
2. Force engine.FullScreen
3. Bind event on 'escape' and 'F11' to remove the full page

Thanks !

Jaskar:
@Temechon It's good now :)

PR #2126, opened 2017-05-11
Camera fixes

bghgary:
This makes the sandbox work a lot better on small models and more consistently overall

### DIFF
--- a/src/Cameras/babylon.arcRotateCamera.ts
+++ b/src/Cameras/babylon.arcRotateCamera.ts
@@ -315,11 +315,11 @@ module BABYLON {
                 this.inertialAlphaOffset *= this.inertia;
                 this.inertialBetaOffset *= this.inertia;
                 this.inertialRadiusOffset *= this.inertia;
-                if (Math.abs(this.inertialAlphaOffset) < Epsilon)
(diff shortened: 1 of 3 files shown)


src/Cameras/babylon.arcRotateCamera.ts
```diff
@@ -315,11 +315,11 @@ module BABYLON {
                 this.inertialAlphaOffset *= this.inertia;
                 this.inertialBetaOffset *= this.inertia;
                 this.inertialRadiusOffset *= this.inertia;
-                if (Math.abs(this.inertialAlphaOffset) < Epsilon)
+                if (Math.abs(this.inertialAlphaOffset) < this.speed * Epsilon)
```
deltakosh: It is a kind of a breaking change as speed == 2 by default. But I guess this is okay epsilon  x 2 is still epsilon:)

bghgary: Yes, that's what I figured. If you want, I can change it to this.speed * Epsilon * 0.5

deltakosh: This is fine

PR #2142, opened 2017-05-19
[WIP] MultiRenderTargets, geometryRenderer and SSAO V2

CraigFeldspar:
SSAO 2 using MultiRenderTargets.

The effect should be adaptable to remain fluid even on modest graphics hardware.

Live demo is here : 
https://craigfeldspar.github.io/ssao/

Usage is the following : 
```
    var ssaoRatio = {
        ssaoRatio: 0.5, // Resolution ratio of the SSAO post-process. Reducing this factor can dramatically enhance performances. Usually a 0.5 ratio (half resolution) gives good enough results, and on some light AO, a 0.25 ratio can be enough. 
        blurRatio: 0.5 // Ratio of the blur post-process. Reducing this factor can dramatically enhance performances, but artifacts can appear.
    };

    var ssao = new BABYLON.SSAO2RenderingPipeline("ssao", scene, ssaoRatio);
    // Maximum radius in scene units from where occlusion can happen
    // A smaller radius usually leads to better performances
    ssao.radius = 3.5;

    // Strength of the occlusion. Base value is 1.0 for raw occlusion.
    // On low quality settings, it can be good to set this to a low value so artifacts are less noticeable
    ssao.totalStrength = 1.0;

    // Activates bilateral filter for blurring. Gives a better shadow quality, specifically around geometry edges. Setting this to false can improve performance dramatically.
    ssao.expensiveBlur = true;

    // Number of samples for computing occlusion. More samples leads to less artifacts but impacts performance.
    ssao.samples = 16;

    // Maximum distance from camera for AO to be computed. If this value is too high, z-fighting artifacts can appear.
    ssao.maxZ = 250;

    // Register pipeline if supported
    if (ssao.isSupported) {
        scene.postProcessRenderPipelineManager.attachCamerasToRenderPipeline("ssao", camera);
    }
```
Play with these factors to get the best deal between performance and quality.
For the lower end GPUs, you can try to give a light AO feeling, by pushing both ratios to 0.25, `expensiveBlur ` to `false `, `totalStrength` to 0.5 or so, and `samples` to 8.

- [x] PG example
- [x] Readme
- [x] Live demo


deltakosh:
Hey! 
I'm oof until next week. Please keep the good work. I'll review it soo

src/babylon.scene.ts
```diff
@@ -711,6 +711,7 @@
         private _debugLayer: DebugLayer;
 
         private _depthRenderer: DepthRenderer;
+        private _geometryRenderer: GeometryRenderer;
```
deltakosh: In current implementation the renderer renders all active meshes
It could be useful to allow devs to define other  renderes with different renderlist

src/Rendering/babylon.geometryRenderer.ts
```diff
@@ -0,0 +1,164 @@
+module BABYLON {
+    export class GeometryRenderer {
+        private _scene: Scene;
```
CraigFeldspar: The thing is "MultiRenderTargetRenderer" is not precise enough.
This renderer actually renders the geometry buffer (sometimes called g-buffer) which contains depths, normals, and in the future probably albedo and some lightning information.
It uses MRT for this purpose, but it is not the end goal.

src/Rendering/babylon.geometryRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+            this._multiRenderTarget.renderParticles = false;
+            this._multiRenderTarget.renderList = null;
+            
+            // set default depth value to 1.0 (far away)
+            this._multiRenderTarget.onClearObservable.add((engine: Engine) => {
+                engine.clear(new Color4(0.0, 0.0, 0.0, 1.0), true, true, true);
+            });
+
+            // Custom render function
+            var renderSubMesh = (subMesh: SubMesh): void => {
```
CraigFeldspar: We probably can with depthRenderer, my code is widely based on this one. I'll think about some solution.
For shadows, it is also interesting to look into it, since we can now use depth textures with WebGL2

src/Materials/Textures/babylon.renderTargetTexture.ts
```diff
@@ -49,7 +49,7 @@
         */
         public onAfterUnbindObservable = new Observable<RenderTargetTexture>();
 
-        private _onAfterUnbindObserver: Observer<RenderTargetTexture>;
+        protected _onAfterUnbindObserver: Observer<RenderTargetTexture>;
```
CraigFeldspar: My implementation of BABYLON.MultiRenderTargetTexture is a child class of BABYLON.RenderTargetTexture, thus the change to protected for some properties.

src/babylon.scene.ts
CraigFeldspar: That's true and easy to change, I'm on it

src/Rendering/babylon.geometryRenderer.ts
deltakosh: well perhaps I just need to get use to it..

deltakosh: What about GeometryBufferRenderer?

src/Materials/Textures/babylon.renderTargetTexture.ts
deltakosh: Yes but perhaps I missed it but I did not see where you use it

CraigFeldspar: you're right, I used to and now I don't. I will double check what I changed to protected

src/Rendering/babylon.geometryRenderer.ts
CraigFeldspar: that works

src/PostProcess/babylon.ssao2RenderingPipeline.ts
```diff
@@ -0,0 +1,369 @@
+﻿module BABYLON {
+    export class SSAO2RenderingPipeline extends PostProcessRenderPipeline {
+        // Members
```
CraigFeldspar: IsSupported will be added, but I'm not sure about the fallback, SSAO v2 and v1 have totally different shaders, and parameter uniforms are not the same. 

Maybe just offering the possibility for devs to test that the effect is supported would be enough, what do you think ?

deltakosh: I'm fine with that. We then just need to state it in the doc

deltakosh:
Ok I'm ready to merge it.
Can you please fix the conflicts?

PR #2144, opened 2017-05-20
Adding Cell Shading Material

julien-moreau:
Adding way to compute custom lighting functions (computeCustomDiffuseLighting & computeCustomSpecularLighting) here used by cell material as example. Then added a way to keep the light intensity (ndl) in lightInfo structure following memory alignment


julien-moreau:
@deltakosh, do the modifications I did in shaders includes look ok for you ?

deltakosh:
Like it. This is pretty smart

sebavan:
Nice one !!!

PR #2145, opened 2017-05-20
First pass on motion blur support in Standard Rendering Pipeline

julien-moreau:
Will try to improve the for loop in shader, don't like to have a "if". Needed because "for" loops must have a constant limit :(


julien-moreau:
Link to the playground : http://www.babylonjs-playground.com/#MV48N3

PR #2147, opened 2017-05-21
Fixes materials library's testing page

zhutq:
On Ubuntu, material library's testing page (localhost:1338/materialsLibrary/index.html, served by gulp webserver) is broken because several paths are not found. This patch fixes these paths.


deltakosh:
Thanks a lot!

PR #2163, opened 2017-05-25
Fixes serialization for geometry with typed arrays

zhutq:
If mesh's vertex data contains Fload32Array, file serialized by `SceneSerializer.Serialize..` cannot be loaded successfully with `SceneLoader.ImportMesh` or `SceneLoader.Load`.

If position data is of type Float32Array, the serialized string looks like
`"positions":{"0":-5,"1":2,"2":-3,"3":-7,"4":-2,"5":...`
which will be parsed as `Object` instead of `Array`. `VertexData.ImportVertexData` does not handle this case.
Also, this kind of string does not conform with the specification at https://doc.babylonjs.com/generals/file_format_map_(.babylon)#vertexdata .

So this patch converts typed arrays to number arrays for serialization.

### DIFF
--- a/src/Mesh/babylon.geometry.ts
+++ b/src/Mesh/babylon.geometry.ts
@@ -639,12 +639,7 @@
             if (Array.isArray(origin)) {
                 return origin;
             } else {
-                var len = origin.length;
-                var copy = [];
-                for (var i = 0; i < len; i++) {
-                    copy.push(origin[i]);
-                }
-                return copy;
+                return Array.prototype.slice.call(origin);
             }
         }
 
--- a/src/Mesh/babylon.geometry.ts
+++ b/src/Mesh/babylon.geometry.ts
@@ -24,8 +24,8 @@
         private _vertexArrayObjects: { [key: string]: WebGLVertexArrayObject; };
 
         /**
-         *  The Bias Vector to apply on the bounding elements (box/sphere), the max extend is computed as v += v * bias.x + bias.y, the min is computed as v -= v * bias.x + bias.y 
-         * @returns The Bias Vector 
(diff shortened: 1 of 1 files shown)


sebavan:
Should the parse not convert back to ArrayBufferView ?

zhutq:
Sorry I'm not familiar with `ArrayBufferView`, but I believe it results in Object. Would you please have a look at https://www.babylonjs-playground.com/#DULLR5 ?
The file I load is just the serialization output. The imported mesh's vertex data is of type Object; and because of that `getTotalVertices()` returns `NaN`.

sebavan:
Sorry my explanation, ArrayBufferView is helper type englobing the types like float32array: https://developer.mozilla.org/en-US/docs/Web/API/ArrayBufferView

I understand the goal and it is great you did the pull request :-)

I am only wondering if we should not do the opposite during the deserialization in the parse method. converting back from array to float32array ?

zhutq:
Oh I see. Thank you! Sorry for my understanding.
I chose to change the serialization side mainly because it is much easier for me. I don't think I'm able to modify deserialization very soon..
Can I leave it to you guys?

sebavan:
np, I have just looked it up and it should be ok like this.

src/Mesh/babylon.geometry.ts
```diff
@@ -635,88 +635,101 @@
             return serializationObject;
         }
 
+        private toNumberArray(origin: Float32Array | IndicesArray) : number[] {
+            if (Array.isArray(origin)) {
+                return origin;
+            } else {
+                var len = origin.length;
```
deltakosh: this could be done like this: http://jsfiddle.net/wTJDy/ 
var aa =  Array.prototype.slice.call(a);

deltakosh:
Thanks a lot:)

PR #2164, opened 2017-05-25
fix for lower fps when reflections are disabled

abow:
http://www.html5gamedevs.com/topic/30655-disabling-reflection-texture-has-no-fps-win/


sebavan:
Thank you @adam, could you add the change on PBR as well ?

sebavan:
Thank you

PR #2210, opened 2017-06-01
PointerUp triggered when leaving Button control.

brianzinn:
Originally made change in control.ts _processObservables(), but felt that may have other unintended consequences.


deltakosh:
This is great
thanks a lot

PR #2226, opened 2017-06-03
Added faceUV and faceColors to ExtrudePolygon

BabylonJSGuide:


deltakosh:
Sounds like too much files changed to me :)

BabylonJSGuide:
OK I must have compiled to the wrong set of files will re-do it.

PR #2234, opened 2017-06-05
Set measure in px for ideal

royibernthal:
It doesn't cover TextBlock and I'm not sure if it's best practice, but it worked for me when measuring containers. I need to dive a little deeper and see how I make it work with all controls.

### DIFF
--- a/gui/src/controls/control.ts
+++ b/gui/src/controls/control.ts
@@ -768,6 +768,13 @@ module BABYLON.GUI {
         
             this._fontOffset = Control._GetFontOffset(this._font);
         }
+        
+        public measureForIdeal() {
+            this._measure();
+            
+            this.width = this._currentMeasure.width + 'px';
+			this.height = this._currentMeasure.height + 'px';
+        }
 
         // Statics
         private static _HORIZONTAL_ALIGNMENT_LEFT = 0;


gui/src/controls/control.ts
```diff
@@ -768,6 +768,13 @@ module BABYLON.GUI {
         
             this._fontOffset = Control._GetFontOffset(this._font);
         }
+        
+        public measureForIdeal() {
+            this._measure();
+            
+            this.width = this._currentMeasure.width + 'px';
```
deltakosh: Did you test it?

I don't think it will work as this.width will be reevaluated again and then scale down or up and then this._currentMeasure will be changed accordingly.

royibernthal: Basically for this purpose it shouldn't matter if currentMeasure will be changed (if I get the picture correctly).
The point is to do a 1-time measurement of the control and set its width and height in px, as if it were done externally by the user, in order for the control to be resized according to idealWidth and idealHeight later on.

I created a container and its width and height seem to have been set properly, but I can't say I thoroughly tested all cases, or that I even entirely understand what's going on.

deltakosh: Can you share a PG? I struggle understanding the goal :)

royibernthal: For some reason in PG _currentMeasure returns 0, 0 for me, while locally it works - I probably created locally a situations where it works completely by luck, as like I said I don't entirely understand what's going on in the source (well, far from it).

Anyway, I hope this PG could at least help you understand my goal:
http://www.babylonjs-playground.com/#XCPP9Y#66

deltakosh:
You cannnot call _measure directly as the container needs its parent size to evaluate its own size.
Here is a fix:
http://www.babylonjs-playground.com/#XCPP9Y#74

But still I do not see where you want to go.

deltakosh:
Closing for now

royibernthal:
I'd like to:
1) Create a function that measures any kind of control and returns the width and height in px
2) Use the measured width and height in a way that'll help automate things when dealing with idealWidth and remove some boilerplate code

I really don't know how to phrase it differently :(

Anyway, I guess I'll have to dig some more into the source and make a proper commit next time.

Issue #2239, opened 2017-06-05
Unable to load non-babylon content using the `data:` syntax


HowardWolosky:
As exemplified by this [Playground example](https://www.babylonjs-playground.com/#1AbIIF#7), instead of providing a file/Uri reference for `BABYLON.SceneLoader.Load`, you can instead combine `data:` with the JSON.stringify`d file content to directly load the file content without doing any file i/o.

The problem with this, is that in `SceneLoader`, it assumes that direct file content is of the `.babylon` type (see [babylon.sceneLoader.ts#L260](https://github.com/BabylonJS/Babylon.js/blob/master/src/Loading/babylon.sceneLoader.ts#L260)).

This prevents us from directly loading .glTF file content like in an unsaved VS Code tab as described in [this comment thread](https://github.com/AnalyticalGraphicsInc/gltf-vscode/pull/13#issuecomment-305671436).

It appears that instead of saying:

`var registeredPlugin = directLoad ? SceneLoader._getDefaultPlugin() : SceneLoader._getPluginForFilename(sceneFilename);`

we should have something like:

`var registeredPlugin = directLoad ? SceneLoader._getPluginForDirectLoad(directLoad) : SceneLoader._getPluginForFilename(sceneFilename);`

You'd then probably need to extend `ISceneLoaderPlugin` and `ISceneLoaderPluginAsync` and add a new method that is something like: `isDataForPlugin: (data: string) => boolean`, and then `_getPluginForDirectLoad` would loop through all registered plugins, asking them if they support that data.  It would return back `SceneLoader._getDefaultPlugin()` in the event that there were no hits to get back to existing behavior.

It's not clear to me how impactful it is to change the plugin interface (are there actively used plugins that are external to the Babylon source tree that would be affected?)

deltakosh:
Makes sense!

deltakosh:
Done, you just need to implement a function named canDirectLoad(data: string) => boolean in the gltf loader :)

PR #2242, opened 2017-06-06
Add canDirectLoad to glTFLoader

HowardWolosky:
Adds `canDirectLoad` to glTFLoader per the [recent change](https://github.com/BabylonJS/Babylon.js/issues/2239) to `ISceneLoaderPlugin` and `ISceneLoaderPluginAsync` so that plugins can determine if they support file content being loaded with the `data:` protocol.

Additionally fixes a bug in the [commit that resolved that issue](https://github.com/BabylonJS/Babylon.js/commit/a458241d78f10f65cf65fa6466ac06e54a28bd7f) that was passing the full protocol (including the `data:`) to the method instead of just the file content.


deltakosh:
Wonderful thanks a lot!

PR #2245, opened 2017-06-07
Attach Fragment Normal Variable 

NasimiAsl:


NasimiAsl:
other changing in progress about custom material
like atlas texture and tools attach shaderBuilder Compiler too but that take a time 
and need your opinions about that (CustomMaterial)

PR #2256, opened 2017-06-08
Removing HDRRenderingPipeline as it is replaced by StandardRenderingP…

julien-moreau:
…ipeline which is more advanced


julien-moreau:
TODO: remove from documentation

PR #2262, opened 2017-06-09
Fixe alpha of button

dad72:
I just made that alpha does not degrade increments at each hovers of mouse but still leave with its initial value. (If the "hover out" event does not trigger)

### DIFF
--- a/gui/src/controls/button.ts
+++ b/gui/src/controls/button.ts
@@ -11,13 +11,14 @@ module BABYLON.GUI {
             super(name);
             this.thickness = 1;
             this.isPointerBlocker = true;
+            this._initialAlpha = this.alpha;
 
-            this.pointerEnterAnimation = () => {
-                this.alpha -= 0.1;
+            this.pointerEnterAnimation = () => {               
+                this.alpha = (this._initialAlpha - 0.1);
             }
 
-            this.pointerOutAnimation = () => {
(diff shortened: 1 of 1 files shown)


deltakosh:
This does not work if I change alpha afterwards

dad72:
I think the ideal will be a new event when pert focus. In my case, the pointerOut event is not run, because the click on the button I have a new window that opens and prevents the pointerOut event to run. And I find myself gradually to an alpha of 0.

I do not know how.

deltakosh:
I'll do something to fix it

deltakosh:
Can you provide a repo in pg?

dad72:
I would have joined to this, a library jQuery and CSS the bug on PG.

dad72:
I try to do it later, there I have to go pick up my daughter at school

dad72:
I can not do it in the playground, he refuses my link to the jquery lib.

https://www.babylonjs-playground.com/#U9AC0N#10

deltakosh:
jquery is in the repo I think by default

dad72:
This is jquery.modal.min.js

deltakosh:
then you just need to reference it over https

dad72:
My server is not with https

dad72:
If I put https I: Failed to load resource: net::ERR_INSECURE_RESPONSE

PR #2271, opened 2017-06-09
Fixed the display of material on the water

dad72:
Fixed the display of material on the water in the case of the use of multiple camera.
I just add in the signature of the constructor the possibility to choose the camera to use.

I guess that's enough.

### DIFF
--- a/materialsLibrary/src/water/babylon.waterMaterial.ts
+++ b/materialsLibrary/src/water/babylon.waterMaterial.ts
@@ -153,9 +153,9 @@ module BABYLON {
         /**
 		* Constructor
 		*/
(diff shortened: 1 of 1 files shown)


deltakosh:
ping @julien-moreau

julien-moreau:
Thanks for PR !
My question is : when it's time to render for a given camera, how do I get the current camera and check if I'm rendering with multiple camera ?
This would allow us to bypass the camera in constructor and prevent scene.activeCamera when it's null

dad72:
Would you do it?
I'll close my PR because it is not appropriate and I do not see how to do it otherwise.
I'm too null to PR here. I will no longer do.

dad72:
Oh and what I had was not working, so it was probably not enough. I just test what I had done.

I guess I should not have tried to do, I do not master the subject. like what, everyone can not contribute.

PR #2292, opened 2017-06-12
Added binary stl loading

focomoso:


deltakosh:
Thanks a lot! Did you tested with both binary and ascii stl?

focomoso:
Yes. Attached is a screenshot of it working with both binary and ascii in the playground on my fork (one on the left is binary, right is ascii). The only thing that seemed unusual was when you called load directly on SceneLoader like this:

`    BABYLON.SceneLoader.Load("./test/", "monkey-ascii.stl", engine, function (newScene) {
        console.log(newScene);
        // newScene.activeCamera.attachControl(canvas, false);
        // engine.runRenderLoop(function () {
        //     newScene.render();
        // });
    });`

The mesh was in the newScene object, but the ascii section of the loader was trying to add the mesh to a meshes array that was null so I had to add the check on lines 25 and 67. Let me know if there's a better way to handle this.

Thanks 


screen shot 2017-06-12 at 4 32 51 pm 3

deltakosh:
Wonderful!

PR #2293, opened 2017-06-13
syncedWithParent order bug.

wyklion:
Otherwise, it'll compute always.

### DIFF
--- a/src/Mesh/babylon.abstractMesh.ts
+++ b/src/Mesh/babylon.abstractMesh.ts
@@ -1162,8 +1162,6 @@
 
             // Parent
             if (this.parent && this.parent.getWorldMatrix) {
-                this._markSyncedWithParent();
-
                 if (this.billboardMode !== AbstractMesh.BILLBOARDMODE_NONE) {
                     if (this._meshToBoneReferal) {
                         this.parent.getWorldMatrix().multiplyToRef(this._meshToBoneReferal.getWorldMatrix(), Tmp.Matrix[6]);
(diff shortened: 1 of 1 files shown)


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -1185,6 +1183,7 @@
                         this._localWorld.multiplyToRef(this.parent.getWorldMatrix(), this._worldMatrix);
                     }
                 }
+                this._markSyncedWithParent();
```
deltakosh: Don't see why it makes a difference there

wyklion: markSynced is for dont recompute.
you can add print here. and you'll see it never stop recompute.
even all meshes not moving.
I use an older version. but i think not big difference.

deltakosh:
Do you have a repro in the pg? Thanks a lot for your help

wyklion:
no... sorry for that...
take it as a remind... 

a's parent-> b
b's parent -> c

see this:
to compute a's matrix.
set a._parentRenderId = b._currentRenderId;
then compute b's matrix. that'll cause b._parentRenderId = c._currentRenderId.(make a not synced with b)
then compute c's matrix. and set c._currentRenderId = scene.RenderId.
the result, a/b/c all not synced.

deltakosh:
Can provide one? As this is pretty central we need to be sure it won't break anyting

deltakosh:
I need to understand how this could happen

wyklion:
Ok.. I'll try to make one when i have time.

deltakosh:
Thanks. I'll merge as soon as the fix is confirmed or I'll find a fix else:)

deltakosh:
I think it is ok like this after my intial testing

PR #2297, opened 2017-06-13
Custom Building using browser profiling.  Directions in self-contained

Palmer-JC:
html file.  Can read either the current master or preview config.json
files from Github.

For making builds against tagged production versions, a local
config.json file from repo required.  This is only readable by Firefox.


deltakosh:
Impressive!

PR #2323, opened 2017-06-16
Fix: Helpers.IsSystemName was not used in loop through all the meshes

Borchmann:
…from the first scene. Moreover, there was some code implementing the same functionality that the helper mentioned.


deltakosh:
Thanks

PR #2332, opened 2017-06-17
fix for ArcRotateCamera not detecting out of sync when target moves

abow:
fix for this:
http://www.html5gamedevs.com/topic/31106-arcrotatecamera-target-animation-after-tab-switch/#comment-178547

and this:
http://www.html5gamedevs.com/topic/30760-mesh-as-camera-target/


deltakosh:
Excellent !

PR #2357, opened 2017-06-21
Blender exporter: Added fog density, fog type to World panel.  Added shadow map darkness to Light panel.

robknopf:
Blender exporter panel for World, exposes Fog type and density


deltakosh:
ping @Palmer-JC for review

Palmer-JC:
All the UI works.  Pretty sure it is writing legal Fog types, since they were documented as comments prior.

Put it through.  I will then:
 

- change revision to zero from -1.  People have had enough time to play with morph.
- Generate a zip file.
- Remove 5.2

@Vinc3r there is now a Panel in World.

robknopf:
@Palmer-JC   I meant to add the second set of changes as a different pull request.  Can you do a review of them as well, please?

Palmer-JC:
darkness must be new.  Setter in Javascript is forcing it to be 0 - 1.  Change to definition to:
    default = 1, min = 0, max = 1

Also, use 'darkness' to assign to all the way through.  That is what BJS calls it.  You have both 'shadowStrength' & 'shadowDarkness'.  I did not run this code, but I think it would blow up on line 116. 
- - - - -
Not sure what change to Float precision is for.  Could allow to change precision on field by field basis in future, though.  Something like normals / UV always default, but positions over-rideable.  If that is the case, need to also pipe arg to other callers of format_f.  Positions calls format_vector_array, which calls format_vector, which calls format_f.

Dynamically doing this for every single number in export could slow it down.   I will check elapsed time before an after in log file.  Unless really bad, who cares.  This stripping of un-necessary  signs, zeros, & decimals makes the file small by a very large margin.  That is what counts.

robknopf:
Grr.. shadowStrength was what I called it during dev and renamed it at the last minute to shadowDarkness.  That's what I get for assuming that my S&R did everything.  I'll change that and resubmit.  On a side note, I called it shadowDarkness to follow the existing naming convention in the file (shadowBias, shadowMap, shadowMapSize, shadowBlurScale, etc).  Would you prefer to be consistent with that, or with 'darkness'?  Makes no nevermind to me.

Palmer-JC:
shadowDarkness is fine

Palmer-JC:
Ok, I performance checked the changes to package_level for controlling # of decimals.  File was same size, and export time almost identical.  Looks ready to go.

deltakosh:
Thanks a lot!

PR #2392, opened 2017-06-28
Update babylon.abstractMesh.ts

SvenFrankson:
Add method "RotateAround", to rotate a mesh around a given point and axis.

We implemented it because it suited our needs, but it might be useful to others.

Not sure the way we compose translation and rotation matrices to then extract their components is the best way to go, but as far as I know their is no "apply transform matrix to mesh" method already implemented. Or is there ?

### DIFF
--- a/src/Mesh/babylon.abstractMesh.ts
+++ b/src/Mesh/babylon.abstractMesh.ts
@@ -792,30 +792,26 @@
          * Returns the AbstractMesh.
          * Method is based on http://www.euclideanspace.com/maths/geometry/affine/aroundPoint/index.htm
          */
-        public rotateAround(point: BABYLON.Vector3, axis: BABYLON.Vector3, amount: number): BABYLON.AbstractMesh {
+        public rotateAround(point: Vector3, axis: Vector3, amount: number): AbstractMesh {
             axis.normalize();
             if (!this.rotationQuaternion) {
-                this.rotationQuaternion = BABYLON.Quaternion.RotationYawPitchRoll(this.rotation.y, this.rotation.x, this.rotation.z);
-                this.rotation = BABYLON.Vector3.Zero();
+                this.rotationQuaternion = Quaternion.RotationYawPitchRoll(this.rotation.y, this.rotation.x, this.rotation.z);
+                this.rotation.copyFromFloats(0, 0, 0);
             }
-            let translation: BABYLON.Vector3 = point.subtract(this.getAbsolutePosition());
-            let mTranslate: BABYLON.Matrix = BABYLON.Matrix.Translation(translation.x, translation.y, translation.z);
-            let mTranslateInvert: BABYLON.Matrix = mTranslate.clone();
-            mTranslateInvert.invert();
-            let mRotation: BABYLON.Matrix = BABYLON.Matrix.RotationAxis(axis, amount);
-            let transform: BABYLON.Matrix = mTranslateInvert.multiply(mRotation).multiply(mTranslate);
+            point.subtractToRef(this.position, Tmp.Vector3[0]);
+            Matrix.TranslationToRef(Tmp.Vector3[0].x, Tmp.Vector3[0].y, Tmp.Vector3[0].z, Tmp.Matrix[0]);
+            Tmp.Matrix[0].invertToRef(Tmp.Matrix[2]);
+            Matrix.RotationAxisToRef(axis, amount, Tmp.Matrix[1]);
+            Tmp.Matrix[2].multiplyToRef(Tmp.Matrix[1], Tmp.Matrix[2]);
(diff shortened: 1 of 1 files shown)


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -784,7 +784,39 @@
...
+         */
+        public rotateAround(point: BABYLON.Vector3, axis: BABYLON.Vector3, amount: number): BABYLON.AbstractMesh {
+            axis.normalize();
+            if (!this.rotationQuaternion) {
+                this.rotationQuaternion = BABYLON.Quaternion.RotationYawPitchRoll(this.rotation.y, this.rotation.x, this.rotation.z);
+                this.rotation = BABYLON.Vector3.Zero();
+            }
+            let translation: BABYLON.Vector3 = point.subtract(this.getAbsolutePosition());
+            let mTranslate: BABYLON.Matrix = BABYLON.Matrix.Translation(translation.x, translation.y, translation.z);
+            let mTranslateInvert: BABYLON.Matrix = mTranslate.clone();
```
deltakosh: You can directly call Matrix.InvertToRef() here

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -784,7 +784,39 @@
...
+         * Note that the property `rotationQuaternion` is then automatically updated and the property `rotation` is set to (0,0,0) and no longer used.  
+         * The passed axis is also normalized.  
+         * Returns the AbstractMesh.
+         * Method is based on http://www.euclideanspace.com/maths/geometry/affine/aroundPoint/index.htm
+         */
+        public rotateAround(point: BABYLON.Vector3, axis: BABYLON.Vector3, amount: number): BABYLON.AbstractMesh {
+            axis.normalize();
+            if (!this.rotationQuaternion) {
+                this.rotationQuaternion = BABYLON.Quaternion.RotationYawPitchRoll(this.rotation.y, this.rotation.x, this.rotation.z);
+                this.rotation = BABYLON.Vector3.Zero();
```
deltakosh: You can call this.rotation.copyFromFloats(0, 0, 0) (the idea is to reduce the number of instantiation to save GC calls)

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -784,7 +784,39 @@
...
+            }
+            let translation: BABYLON.Vector3 = point.subtract(this.getAbsolutePosition());
+            let mTranslate: BABYLON.Matrix = BABYLON.Matrix.Translation(translation.x, translation.y, translation.z);
```
deltakosh: translation vector can be avoided: you can call point.subtractToRef and use Tmp.Vector3[0]

deltakosh: then use Matrix.TranslationToRef() and save the result in one of the already instantiated values like Tmp.Vector3[1]

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -784,7 +784,39 @@
...
+            let mTranslateInvert: BABYLON.Matrix = mTranslate.clone();
+            mTranslateInvert.invert();
+            let mRotation: BABYLON.Matrix = BABYLON.Matrix.RotationAxis(axis, amount);
```
deltakosh: You get my point:) mRotation and transform have to come from tmp.Matrix :)

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -784,7 +784,39 @@
...
+            let transform: BABYLON.Matrix = mTranslateInvert.multiply(mRotation).multiply(mTranslate);
+
+            let t = BABYLON.Vector3.Zero();
+            let r = BABYLON.Quaternion.Identity();
+            let s = BABYLON.Vector3.One();
+
+            transform.decompose(s, r, t);
```
deltakosh: Here as well you can call decompose with Tmp.Vector3, TMp.QUaternion, etc..

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -784,7 +784,39 @@
...
 
+            this.position.addInPlace(t);
+            this.rotationQuaternion.multiplyInPlace(r);
+
+            return this;
```
deltakosh: I really like this function, it is a great addition! Thanks for it

SvenFrankson:
Thanks for the feedback DeltaKosh,

I get it about the GC optimization, I fixed it but some odd behaviours persist, depending uppon which Tmp.Matrix[index] I pick. I'll investigate it !

deltakosh:
Thanks a lot (a good talk for your symposium by the way;))

PR #2404, opened 2017-06-30
Shadow Shader Darkness Fix

mohamedmansour:
The darkness property was computed incorrectly, this fixes it to match the Shadow with PCF. And tested.


deltakosh:
Thanks!
@sebavan fyi

PR #2413, opened 2017-07-01
Allow configuration of Instance Divisor

benaadams:
Never thought was ever a need for this to be anything other than one; but you learn something everyday 😄

### DIFF
--- a/src/Mesh/babylon.buffer.ts
+++ b/src/Mesh/babylon.buffer.ts
@@ -6,6 +6,7 @@
         private _updatable: boolean;
         private _strideSize: number;
         private _instanced: boolean;
+        private _instanceDivisor: number;
 
         constructor(engine: any, data: number[] | Float32Array, updatable: boolean, stride: number, postponeInternalCreation?: boolean, instanced?: boolean) {
             if (engine instanceof Mesh) { // old versions of BABYLON.VertexBuffer accepted 'mesh' instead of 'engine'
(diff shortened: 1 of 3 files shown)


deltakosh:
Me neither to be honest:)

benaadams:
Half a mesh; axis multiplier -1, 1

benaadams:
Use half a mesh; reflect with axis multiplier -1, 1

benaadams:
Alas it doesn't repeat when it gets to end of 1 multipler; so either double draw or double instance buffer ;(

deltakosh:
Hey..not sure to get what is the problem :(

benaadams:
```
[v,v,v,v,v,v,v,v,v][v,v,v,v,v,v,v,v,v][v,v,v,v,v,v,v,v,v][v,v,v,v,v,v,v,v,v] // repeat divisor 0
[       pos,               pos       ][       pos,               pos       ] // repeat divisor 1 
[                 adj,                                 adj                 ] // divisor 2
```
Was hoping divisor 2 would cause divisor 1 to repeat when it got to the end. But instead just run out of vertexes :)

Was a misunderstanding on how it worked

deltakosh:
Oh..ok :) part of me was stressed that we had a bug since long time :D

benaadams:
tl;dr multiplier > 1 can be useful; but way way less useful than I was hoping (borderline not useful)

Issue #2442, opened 2017-07-06
GUI Text Block does not respect new line chars


jasonayre:
I have the need to break up text blocks manually using new line character. Currently new line is not respected, and my only option is to use textWrapping = true. Ideally, both modes would respect new line chars.

jasonayre:
http://www.babylonjs-playground.com/#XCPP9Y#142

for reference

PR #2443, opened 2017-07-06
respect new lines

jasonayre:
Adds support for new line character when setting a text block. Works for both textWrapping=true mode and standard.

screenshot 2017-07-05 19 41 39


jasonayre:
Addresses #2442

deltakosh:
Really cool:)

PR #2453, opened 2017-07-07
Fixed matrix decomposition with negative scaling

Breush:
This intends to fix an issue encountered when using `Matrix.decompose()`:
- http://www.html5gamedevs.com/topic/25222-decomposing-negative-scaling-matrices/

Fact is decomposing a matrix that invert one axis (scaling) gave wrong results.

The following fix, checking the matrix determinant, is based on this post (vagran_zulu last comment):
- http://answers.unity3d.com/questions/402280/how-to-decompose-a-trs-matrix.html

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -3283,13 +3283,13 @@
             translation.y = this.m[13];
             translation.z = this.m[14];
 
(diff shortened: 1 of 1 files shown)


deltakosh:
This is great 
Thanks

PR #2489, opened 2017-07-11
Enables inspector to display reflection probes textures

ameuleman:
The inspector is now able to display all six textures of a probe in texture tab.


deltakosh:
Ping @temechon for review

PR #2514, opened 2017-07-12
fix set textures to null

vanRepin:
currently, if you set any texture in PBRMaterial to null(if previously it wasn't), it will lead to unexpected behaviour, like reusing texture from previous material or even errrors in webgl. The reason for it is that defines for material isn't correctly reseted. Thix pr should fix it.


deltakosh:
Thanks a lot!!

PR #2521, opened 2017-07-13
Fixing PBR environment texture serialization. Override serialise in T…

julien-moreau:
…exture in order to set the base64String

@deltakosh you ok with that ? Including the name I gave to the environment texture ?

### DIFF
--- a/src/Tools/babylon.textureTools.ts
+++ b/src/Tools/babylon.textureTools.ts
@@ -60,7 +60,7 @@
 
 		public static GetEnvironmentBRDFTexture(scene: Scene) {
 			if (!scene._environmentBRDFTexture) {
-				var texture = new BABYLON.Texture(this._environmentBRDFBase64Texture, scene, true, false, BABYLON.Texture.BILINEAR_SAMPLINGMODE);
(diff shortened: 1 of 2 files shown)


src/Tools/babylon.textureTools.ts
```diff
@@ -60,7 +60,7 @@
 
 		public static GetEnvironmentBRDFTexture(scene: Scene) {
 			if (!scene._environmentBRDFTexture) {
-				var texture = new BABYLON.Texture(this._environmentBRDFBase64Texture, scene, true, false, BABYLON.Texture.BILINEAR_SAMPLINGMODE);
+				var texture = Texture.CreateFromBase64String(this._environmentBRDFBase64Texture, "EnvironmentBRDFTexture", scene, true, false, Texture.BILINEAR_SAMPLINGMODE);
```
deltakosh: this._environmentBRDFBase64Texture already starts with "data:" so this won't work right?

julien-moreau: Nope it works, CreateFromBase64String takes the given name (here "EnvironmentBRDFTexture ") and adds "data:". So "data: EnvironmentBRDFTexture".
The original data (here _environmentBRDFBase64Texture) is not modified

deltakosh: oh yeah..perfect

PR #2528, opened 2017-07-14
fixed reflection texture matrix performance issue

Kesshi:
I found a perfomance issue which happens if many materials are using the same reflection texture.
Here a playground example: http://www.babylonjs-playground.com/#3PYAME  (enable line 20)
This PR fixes this problem.


deltakosh:
Well done!

Issue #2542, opened 2017-07-18
Reflection (and refraction) seems wrong in right handed mode


deltakosh:
Ping @bghgary

PR #2544, opened 2017-07-18
Fix a few issues with handedness in a number of places

bghgary:
- createDefaultCameraOrLight now creates a camera at a same location as when using right-handed system.
- glTFLoader no longer needs an arbitrary 180 rotation on it for the default left-handed system.
- The PBR reflection map needs to be z flipped when using right-handed system.
- The build validation images are updated to match the latest behavior.


bghgary:
Fixes #2542

bghgary:
CC @sebavan

PR #2549, opened 2017-07-19
Keep OldVer From StandardMaterial

NasimiAsl:


NasimiAsl:
thanks my friend - when i can see in Babylon-playground

deltakosh:
should be the case already

PR #2551, opened 2017-07-19
Fixing materials library

julien-moreau:
Fixing all materials from materials library according to recent modifications on shadows
Removing functions already defined in LegacyPBRMaterial (declared in HelperFunctions.fx


deltakosh:
Thanks!

PR #2573, opened 2017-07-23
CreatePolygon before applyToMesh dispose of existing polygon and crea…

BabylonJSGuide:
From the stable version 3.0 BABYLON.Mesh.DOUBLESIDE no longer worked with CreatePolygon. This appears to be because vertexData.applyToMesh no longer changes an existing mesh. Hence the change which deletes the existing mesh polygon and creates a new mesh to apply new vertexData to. Also following Jerome's suggestion the equality checks ==0, ==1, == -1 have been changed to allow for possible floating point errors.

### DIFF
--- a/src/Mesh/babylon.mesh.vertexData.ts
+++ b/src/Mesh/babylon.mesh.vertexData.ts
@@ -1646,15 +1646,15 @@
             var face: number = 0;
             for (var index = 0; index < normals.length; index += 3) { 
                 //Edge Face  no. 1
-                if(Math.abs(normals[index + 1]) == 0) {
+                if(Math.abs(normals[index + 1]) < 0.001) {
                    face = 1; 
                 }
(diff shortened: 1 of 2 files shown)


deltakosh:
I do not merge not because the PR is broken but because I want first to understand why apply to mesh does not affect data to an existing mesh

BabylonJSGuide:
OK will wait to see if you can find what's happening. In case you have not already spotted it the PG that no longer works is http://playground.babylonjs.com/#4G18GY#2 and the topic where I first put the question is 
http://www.html5gamedevs.com/topic/31797-solved-createpolygon-doubleside-no-longer-works/

but why I think you may not have found it I have no idea.

deltakosh:
Ok I found the bug!

deltakosh:
I close this PR and I will merge your change on VertexData independently

PR #2580, opened 2017-07-25
Master

MackeyK24:
BABYLON.ParticleSystem animation end callback support


src/Particles/babylon.particleSystem.ts
```diff
@@ -390,6 +391,11 @@
             }
 
             this._vertexBuffer.update(this._vertexData);
+
+            // Animation End
+            if (this.onAnimationEnd != null) {
```
deltakosh: not sure this is the right place. Should go around line #375

PR #2592, opened 2017-07-27
Add Color3 array setter for ShaderMaterial

Hersir88:
Added setter to pass Color3 arrays in ShaderMaterial

### DIFF
--- a/src/Materials/babylon.effect.ts
+++ b/src/Materials/babylon.effect.ts
@@ -878,6 +878,15 @@
             return this;
         }
 
+        public setColor3Arrays(uniformName: string, colors: Color3[]): Effect {
+            this._valueCache[uniformName] = null;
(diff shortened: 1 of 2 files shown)


src/Materials/babylon.effect.ts
```diff
@@ -878,6 +878,15 @@
             return this;
         }
 
+        public setColor3Arrays(uniformName: string, colors: Color3[]): Effect {
+            this._valueCache[uniformName] = null;
+            this._engine.setArray3(this.getUniform(uniformName), colors.reduce((arr, color) => {
+                color.toArray(arr, arr.length);
```
deltakosh: I'm not a big fan of creating a new array per frame

Hersir88: My bad will fix it

PR #2608, opened 2017-07-29
Load now checks if each parsed property of .babylon file is defined before attempting to process.

mjschlot:
Load now checks if each parsed property of .babylon file is defined before attempting to process. Prior to this change exclusion of some properties (e.g. autoclear, lights, etc) would cause loading to fail.

One example use case is loading multiple scenes via SceneLoader.Append(). Without this change the .babylon file must specify settings like clearColor. My specific use case is loading multiple files using Append(). In my use case I set clearColor and other settings in my code. My .babylon files only contain a few items (materials, meshes and cameras), and not other items like 'clearColor'. I would consider using importMesh() but that does not load cameras which I need with each call to Append.


src/Loading/Plugins/babylon.babylonFileLoader.ts
```diff
@@ -269,23 +277,25 @@
...
+                        var parsedLight = parsedData.lights[index];
+                        var light = Light.Parse(parsedLight, scene);
+                        log += (index === 0 ? "\n\tLights:" : "");
+                        log += "\n\t\t" + light.toString(fullDetails);
+                    }
                 }
     
                 // Animations
-                if (parsedData.animations) {
+                if (parsedData.animations !== undefined) {
```
deltakosh: What if the value is null? We should keep the previous check

mjschlot: I'd be happy to make that fix. Does that logic apply to all property checks or is the 'animation' property unique?

My primary motivation to all checks being a strict undefined check is that there seems there could be time for some of the other properties being 1, true or similar value would result if a false positive. My secondary motivations are code clarity, slight JS performance and consistency between the checks in this block of code.

Alternatively we could keep it strict and cover both null and undefined, but not leave it open to false positives: `if (parsedData.animations !== undefined && parsedData.animations !== null) {`

This is your project, and I am not familiar enough with the code base so I rely on and appreciate your guidance. Change just "animation" back to the more forgiving condition, should I change other properties back, or add the null check to all cases?

Thanks!

deltakosh: I share your motivation so just add null everytime you were checking undefined and this will be perfect:)

mjschlot: All checks now are for undefined and null. I pushed the changes, but I have not done much in Github's PR system before. Do I need to do anything else to update this PR beyond pushing it?

I also wanted to mention that I did not change the check of "parsedData.autoAnimate" as I am not familiar with what that is. So I left it as it was _originally_:
`if (parsedData.autoAnimate) {`

I should also maybe mention that there are similar checks in importMesh() which I did not change as I didn't want to expand the scope of this PR too much.

I'd be happy to address both of those if  you feel it is a good idea. Thanks, I'm excited to be making my first PR to this Babylon.js!

deltakosh:
Please do the same for impostMesh. This is a really cool pr

mjschlot:
Implemented stricter checks inside of importMesh. I also modified the code slightly to check if the `meshesNames` is an array or not once instead of doing it over and over -- it was being done twice per mesh before.

For skeletons inside of importMesh there was an if statement that said `if (parsedMesh.skeletonId > -1 && scene.skeletons) {`, but it appears it should have been `if (parsedMesh.skeletonId > -1 && parsedData.skeletons) {`. It now says `if (parsedMesh.skeletonId > -1 && parsedData.skeletons !== undefined && parsedData.skeletons !== null) {`. But, I wanted to point it out so you could confirm I was correct in switching form `scene.skeletons` to `parsedData.skeletons`.

I accidentally committed some of the built files last night within the "preview release" directory. Can you please ignore those and only merge in the one file ("babylon.babylonFileLoader.ts") I meant to commit for this PR?

deltakosh:
Thanks!

PR #2609, opened 2017-07-31
defined both width and height for stack panel

projectSHAI:
The original code only defined width if `isVertical = false` or height if `isVertical = true` for the `StackPanel` element.

If height or width is not defined it defaults to 100%, and this is set in the `control.ts` file.

This leads to some undesirable side affects. 

### Scenario one ###

http://www.babylonjs-playground.com/#XCPP9Y#166

because height is undefined and defaults to 100% any element created before the stack panel is blocked by it. uncomment line 28 for the expected behavior.

Defining the height or width of stack panel is not ideal, since both should be dictated by its children, in case child elements are appended or deleted.

### Scenario two ###

https://www.babylonjs-playground.com/#XCPP9Y#167

because height is undefined and defaults to 100% users are unable to take advantage of `BABYLON.GUI.Control.VERTICAL_ALIGNMENT_TOP;` The same is true for horizontal alignment when `isVertical = true`.

## solution ##

Define both stackWidth and stackHeight.

lines 44-46 will find the largest width and set the stack panel's width to that width when vertical is true

and 55-57 will find the largest height and set stack panel's height to that when vertical is false

Then in lines 67-68 or 75-76 define both width and height.

> PS - I tried `gulp typescript-all` to build the project and everything went well, however, this did not update the local playground with my changes at `localhost:1338/Playground` is there a way to do that?


deltakosh:
I like the idea! we should just not do it width (or height) are already defined

projectSHAI:
I agree, if a user explicitly defines the width or height of the stack panel it should overwrite the implicitly defined width and height.

The additions I made should not get in the way of a user defining their own width or height.

Also, could you confirm the proper way to update the local playground to reflect changes to the repo? `gulp typescript-all` doesn't seem to effect the playground at all.

deltakosh:
Playground (local) should just directly pull your local changes if you debug from VS code

projectSHAI:
I did some research and it does not appear to pull the resources from my local environment.

The resources in the `/Playground/index.html` are all `<script src="https://preview.babylonjs.com/[path/to/resource]"></script>` which is a remote file.

lines 38-74 to be precise.

I'm going to manually change them, so I can test my changes with the local gui file. are these `preview.babylonjs` links fallbacks that the `BabylonLoader.js`'s `Loader.loadScripts` function is supposed to over right?

deltakosh:
No need for that. index-local works directly:
http://doc.babylonjs.com/generals/how_to_start#debug

projectSHAI:
Awesome document!

I was just making my way by reading the source, and following where all the functions were flowing towards to gain an understanding of how to build everything. It was pretty informative, though! I found a couple other things I'd like to improve upon doing so.

I'll play around with VSC and the doc you linked this weekend and make another pull request 👍

projectSHAI:
Hey @deltakosh I did some thinking and came up with a pretty good solution.

The stack panel is by default fully responsive for both width and height.

There are two new properties added to the `StackPanel` class. `manualWidth` and `manualHeight` which are booleans the user can set to true to gain control of sizing the panel to their liking.

This implementation is completely predictable now.

> Note: if we set the `width` and `height` properties in the Control class to default to `null` instead of the `_width` and `_height` properties then we would not need to use `manualWidth` or `manualHeight`. We could check to see if the user defined the `width` and `height` and toggle between responsive or static sizing without the users input making their coding experience more brief. 

The change in `control.ts` line 220 would look like this: 
image


I don't see this affecting anything in the rest of the application, is there any adverse effects you can think of with this change? The benefit of this change would be knowing if the user explicitly set the sizes, or if the `_width` and `_height` are the defaults.

deltakosh:
We cannot change the current behavior so I'm fine with manualWidth and manualHeight.
They should be automatically set when I set height or width.

projectSHAI:
Cool.

I attempted automatically setting `manualWidth` and `manualHeight` the only obvious solution I found was to check for `height` and `width`. Since these two properties default to "100%" when not set, the only logic I can think of would be:

    if(this.height !== "100%") {
        this.manualHeight = true;
    }
    if(this.width !== "100%") {
        this.manualWidth = true;
    }


This works well, however, leaves one fringe case. Manual sizing will remain false if the user explicitly defines a width or height of 100%/

deltakosh:
Why not using the set property for that? and set the _height and _width internally to avoid false positives

deltakosh:
I'll have a look at it :)

PR #2611, opened 2017-07-31
Fix materialDefines

NicolasBuecher:
Reset materialDefines if a new Material is assigned to a Mesh and subMeshes._materialEffect is already null.
Reset materialDefines if a new subMaterial is assigned to a subMesh.
When marking subMeshes as dirty, don't return if a subMesh doesn't have defines.


deltakosh:
Thanks

PR #2638, opened 2017-08-08
Instanciate function

NicolasBuecher:
Hello,

I would like to submit a small add-on for SerializationHelper, the possibility of creating a new object like if we were cloning but sharing the same properties as the source object.
Since it differs slightly from Clone static function, I reused Clone code putting it in another third function.

Thanks


NicolasBuecher:
https://github.com/BabylonJS/Babylon.js/pull/2638/commits/8a51bd014e9a9fea0f6589bf66c8cb170de80555#diff-cf29a6a784fc65669b4bc466d40d7f06R33

This line. A clone clones other objects. The instance shares the same references.
But then you can freely overwrite the properties that you don't want them to share the same references by new objects.

deltakosh:
Makes sense

PR #2653, opened 2017-08-10
Render Targets for multi-camera scene

RaananW:
When using multiple cameras (like in VR) render target textures didn't work correctly.
This solves the issue - there is now an option to render the texture either on the scene level (shadows for example), or camera level (mirror, or water for example).


deltakosh:
Hey
sorry about the stupid question
But I guess if we change this:

https://github.com/BabylonJS/Babylon.js/blob/master/materialsLibrary/src/water/babylon.waterMaterial.ts#L476 

and remove the line 476 and 477
and instead return the two textures with a getRenderTargetTextures()
I think it should fix the bug
like here in the stdmat:

https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/babylon.standardMaterial.ts#L445

PR #2654, opened 2017-08-11
Deterministic lockstep for physics and animations for frame-rate inde…

giannisantinelli:
…pendence in game logic.

Time is quantized according to PhysicsEngine timeStep, performing a maximum of lockstepMaxSteps in case of late-frames.
Two new observable events are added to plug game-logic code, via registerBeforeStep and registerAfterStep, which allow to get current stepId via scene.getStepId().


giannisantinelli:
Making a fix...

PR #2655, opened 2017-08-11
Deterministic lockstep for physics and animations

santarcade:
... to achieve for frame-rate independence in game logic.

Time is quantized according to PhysicsEngine timeStep, performing a maximum of lockstepMaxSteps in case of late-frames.

Two new observable events are added to plug game-logic code, via registerBeforeStep and registerAfterStep, which allow to get current stepId via scene.getStepId().


src/babylon.scene.ts
```diff
@@ -383,6 +383,34 @@
         public onMeshRemovedObservable = new Observable<AbstractMesh>();
 
         /**
+        * An event triggered before calculating deterministic simulation step
+        * @type {BABYLON.Observable}
+        */
+        public onBeforeStepObservable = new Observable<Scene>();
+
+        private _onBeforeStepObserver: Observer<Scene>;
+        public set beforeStep(callback: (data:Scene, state: EventState) => void) {
```
deltakosh: No need for this. This kind of construction was used just for backward compat. I would like to only use observable if possible

src/babylon.scene.ts
```diff
@@ -383,6 +383,34 @@
...
+        }
+
+        /**
+        * An event triggered after calculating deterministic simulation step
+        * @type {BABYLON.Observable}
+        */
+        public onAfterStepObservable = new Observable<Scene>();
+
+        private _onAfterStepObserver: Observer<Scene>;
+        public set afterStep(callback: (data:Scene, state: EventState) => void) {
```
deltakosh: same

src/babylon.scene.ts
```diff
@@ -1614,6 +1657,22 @@
             this.onAfterRenderObservable.removeCallback(func);
         }
 
+        public registerBeforeStep(func: () => void): void {
```
deltakosh: same for these 4 functions

src/babylon.scene.ts
```diff
@@ -2923,16 +2982,56 @@
                 this.simplificationQueue.executeNext();
             }
 
-            // Animations
-            var deltaTime = Math.max(Scene.MinDeltaTime, Math.min(this._engine.getDeltaTime(), Scene.MaxDeltaTime));
-            this._animationRatio = deltaTime * (60.0 / 1000.0);
-            this._animate();
+            if(this._engine.isDeterministicLockStep()){
+              var deltaTime = Math.max(Scene.MinDeltaTime, Math.min(this._engine.getDeltaTime(), Scene.MaxDeltaTime)) / 1000;
+              var defaultTimeStep = this._physicsEngine.getTimeStep();
```
deltakosh: what if there is no physic engine created?

abow:
Isn't Cannonjs and Oimojs already deterministic?

abow:
If what you want to do is always update the engine by the same fixed interval, couldn't this be handled right in the physics engine?

https://github.com/BabylonJS/Babylon.js/blob/master/src/Physics/babylon.physicsEngine.ts#L121

giannisantinelli:
@abow They are both deterministic, OIMO always updates by fixed interval and yes, setting useDeltaForWorldStep to false would do the trick for CannonJS.
My addition allows to use the same fixed step for both animations, game logic and physics, which is fundamental to sync each single step of a simulation.
This way if animations influences physics, and one or both influence game logic, everything will be kept in sync in same discrete steps.
In addition, my change keeps an accumulator, so if render-time lags behind, it is possible to perform multiple physics/animation/logic ticks within the same frame, to recover some lost ticks.

giannisantinelli:
Ok @deltakosh, I should have fixed according to your review.

abow:
Right now the scene calls the update of the physics engine.  It would make more sense to me to simply have the physics engine updated by a completely separate setInterval.  Maybe that won't solve your problem though.

src/babylon.scene.ts
```diff
@@ -2922,16 +2949,61 @@
...
+
+              var maxSubSteps = this._engine.getLockstepMaxSteps();
+
+              this._timeAccumulator += deltaTime;
+
+              // compute the amount of fixed steps we should have taken since the last step
+              var internalSteps = Math.floor(this._timeAccumulator / defaultTimeStep);
+              internalSteps = Math.min(internalSteps, maxSubSteps);
+
+              for(this._currentInternalStep=0; this._currentInternalStep<internalSteps; this._currentInternalStep++){
```
deltakosh: just a tiny remark: add spaces between operands and operator :)

deltakosh:
Oh and we have conflicts to be resolved before I can merge ;)

santarcade:
@abow: that would not synchronize animations, physics and game tick code.
This is why I had to modify the Scene class and add a couple of Observable.
Yet in a future there could be a way to decouple the render loop from the game tick one.

deltakosh:
They are still conflicts :(

PR #2659, opened 2017-08-11
Watermaterial VR Fix

RaananW:
To enable water material in VR mode some changes had to be made to the water material and the render target class.

### DIFF
--- a/src/Materials/Textures/babylon.renderTargetTexture.ts
+++ b/src/Materials/Textures/babylon.renderTargetTexture.ts
@@ -40,7 +40,8 @@
         public activeCamera: Camera;
         public customRenderFunction: (opaqueSubMeshes: SmartArray<SubMesh>, transparentSubMeshes: SmartArray<SubMesh>, alphaTestSubMeshes: SmartArray<SubMesh>, beforeTransparents?: () => void) => void;
         public useCameraPostProcesses: boolean;
-        
+        public ignoreCameraViewport: boolean = false;
+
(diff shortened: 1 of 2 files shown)


src/Materials/Textures/babylon.renderTargetTexture.ts
```diff
@@ -418,11 +418,7 @@
                 this._postProcessManager._prepareFrame(this._texture, this._postProcesses);
             }
             else if (!useCameraPostProcess || !scene.postProcessManager._prepareFrame(this._texture)) {
-                if (this.isCube) {
-                    engine.bindFramebuffer(this._texture, faceIndex);
-                } else {
-                    engine.bindFramebuffer(this._texture);
-                }
+                engine.bindFramebuffer(this._texture, this.isCube ? faceIndex : undefined, undefined, undefined, this.ignoreCameraViewport);
```
deltakosh: seems like there is one unwanted "undefined"

RaananW: Is there? The signature is:
`public bindFramebuffer(texture: WebGLTexture, faceIndex?: number, requiredWidth?: number, requiredHeight?: number, forceFullscreenViewport?: boolean)`
5 variables, 2 or 3 undefined, depending on isCube

deltakosh: omg..I need to sleep

PR #2667, opened 2017-08-15
 Dispose meshes in LOD levels

kyptov:
I found in my scene, that meshes I added as LOD level not disposing on masterMesh disposing. 

Maybe I am wrong. Otherwise there is no public method to get all meshes from LOD levels.

### DIFF
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -1484,6 +1484,15 @@
                     highlightLayer.removeExcludedMesh(this);
                 }
             }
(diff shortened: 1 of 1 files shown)


deltakosh:
I'm not convinced we should forcefully dispose LOD as they are real meshes that can be used elsewhere (as LOD for another mesh for instance). I would prefer providing a function to return all LOD

kyptov:
Thank you for answer

deltakosh:
Do you want to do the PR for the LOD function?

PR #2682, opened 2017-08-18
Allow bone index remapping

bghgary:
- Add support for remapping bone index.
- Add additional constructor parameters to allow local matrix to be different from base matrix.

### DIFF
--- a/src/Bones/babylon.bone.ts
+++ b/src/Bones/babylon.bone.ts
@@ -11,6 +11,10 @@ module BABYLON {
         public animations = new Array<Animation>();
         public length: number;
 
+        // Set this value to map this bone to a different index in the transform matrices.
+        // Set this value to -1 to exclude the bone from the transform matrices.
+        public _index: number;
+
         private _skeleton: Skeleton;
         private _localMatrix: Matrix;
         private _restPose: Matrix;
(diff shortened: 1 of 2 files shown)


bghgary:
@deltakosh Please review before committing. I ended up not using the sortBones function in the glTF loader, but it might be useful for others.

src/Bones/babylon.bone.ts
```diff
@@ -37,12 +41,13 @@ module BABYLON {
...
-        constructor(public name: string, skeleton: Skeleton, parentBone: Bone = null, matrix?: Matrix, restPose?: Matrix) {
+        constructor(public name: string, skeleton: Skeleton, parentBone: Bone = null, localMatrix?: Matrix, restPose?: Matrix, baseMatrix?: Matrix, index?: number) {
             super(name, skeleton.getScene());
             this._skeleton = skeleton;
-            this._localMatrix = matrix ? matrix : Matrix.Identity();
-            this._baseMatrix = this._localMatrix.clone();
+            this._localMatrix = localMatrix ? localMatrix : Matrix.Identity();
             this._restPose = restPose ? restPose : this._localMatrix.clone();
+            this._baseMatrix = baseMatrix ? baseMatrix : this._localMatrix.clone();
+            this._index = index === undefined ? undefined : index;
```
deltakosh: this._index = index should be enough

deltakosh: or instead of undefined the value should be set to -1

src/Bones/babylon.skeleton.ts
```diff
@@ -232,7 +232,10 @@
                     }
                 }
 
-                bone.getInvertedAbsoluteTransform().multiplyToArray(bone.getWorldMatrix(), targetMatrix, index * 16);
+                if (bone._index !== -1) {
```
deltakosh: I think we should add an else with:
bone.getInvertedAbsoluteTransform().multiplyToArray(bone.getWorldMatrix(), targetMatrix, index * 16);

src/Bones/babylon.bone.ts
bghgary: index === undefined is not the same as -1. undefined means it will auto calculate the index as it does before this change. -1 means it will not store the transform into the transformMatrices.

bghgary: I agree though that this._index = index is enough.

deltakosh: -1 seems important based on my other comment ;)

src/Bones/babylon.skeleton.ts
bghgary: No, that's not the intention. I'm trying to make it such that the skeleton can have more bones than the transformMatrix. There is an optimization we can make that will make the transformMatrix use less memory, but I don't have a good solution yet.

Issue #1819, opened 2017-03-01
Add support for webgl2 occlusion queries


deltakosh:

clementlevasseur:
Hey @deltakosh,

With @Breush, we'd like to work on this subject ! Is it ok for you ?

deltakosh:
obviously it is ok:)
let's just discuss here about how you want to implement it(or perhaps using a PR? you decide:))

clementlevasseur:
We want to create a POC before making a final PR !
I'll tell you when we have a something !

deltakosh:
Sounds perfect

IbraheemOsama:
@deltakosh I'll start working on this, I created a POC to test how the occlusion query works, simply I have a cube and a rectangle, the cube is rotating behind the rectangle if part of the cube is visible to the viewport then I change the color of the rectangle and if the cube is not visible in the viewport reset the rectangle color.
Here is the Poc in case you want to have a look.
http://occlusionquerypoc.azurewebsites.net/

Here is how I see it could be implemented in BabylonJs. we add 2 optional properties or add a function or constructor overload in any drawable object

1) OcclusionType an enum that has 2 values strict and optimistic
2) OcclusionRetriesNumber default is -1

here is the logic.

Before we draw any drawable we check if occlusion query is requested on this drawable 
if not {draw this object}
if yes {

Create a transparent bounding box for this object and draw the bounding box.
query the bounding box if it is available or not 
if available check result and draw or not based on the result
if the query is not available do nothing in the current draw method and just check if the query is available in the upcoming draw methods and increment the internal variable occlusionRetriesNumber and once available then draw or not.

here comes the OcclusionType and OcclusionRetriesNumber to be used in case the query took long time if object internal variable occlusionRetriesNumber is equal to OcclusionRetriesNumber then break
in case of break
if the occlusionType is strict then don't draw the object and if optimistic then draw the object.
if OcclusionRetriesNumber equals to -1 then wait for the query and never break.
I’ve a question does OcclusionType required? from my point of view yes but I wanted to check with you.
This is how I see it :) I'll fork the repo to get my head around how objects are drawn in the library.

This implementation would do some flickering if the user moved the camera in a very fast way because an object should be rendered in the current frame but we would wait for the next one because of the query. but this implementation is much better than waiting and block the requestAnimationFrame until the query result is back.

what do you think, should I start implementing this way ?

IbraheemOsama:
[...]

PR #2692, opened 2017-08-22
Occlusion queries

IbraheemOsama:
referencing issue #1819
To test the occlusion query you can use the current scene 
`var createScene = function () {
                var camera = new BABYLON.FreeCamera('camera1', new BABYLON.Vector3(0, 5, -10), scene);
                camera.attachControl(canvas, false);
                var light = new BABYLON.HemisphericLight('light1', new BABYLON.Vector3(0, 1, 0), scene);

                var plane = BABYLON.Mesh.CreatePlane('opaque1', 5, scene);
                plane.position.y = 1;
                plane.position.z = -2;

                // occlusion sphere
                sphere = BABYLON.Mesh.CreateSphere('sphere1', 16, 2, scene);
                sphere.position.y = 1;

                // this are the supported WebGl query object occlusion query algorithm. accurate is slow but accurate, CONSERVATIVE is fast but not accurate (false positive)
                //sphere.occlusionQueryAlgorithmType = BABYLON.AbstractMesh.OCCLUSION_ALGORITHM_TYPE_ACCURATE;
                sphere.occlusionQueryAlgorithmType = BABYLON.AbstractMesh.OCCLUSION_ALGORITHM_TYPE_CONSERVATIVE;

                // the initial value for the object to be execluded or not at the first render.
                sphere.isOccluded = true;

                // sometimes it takes some time for the query to return back the result so the occlusionType is definining what should happens if the query took long time. if optimistic and break happens the mesh will be rendered if strict the mesh will be rendered or not based on the previous state if it was already rendered rerender it if not keep it occluded. if the strict option will always hide the mesh in case of break and the mesh is supposed to be seen the mesh will flash because it is supposed to be shown but the query sometimes fails which is hidding the mesh and sometimes it succeed so the mesh is shown.

                //sphere.occlusionType = BABYLON.AbstractMesh.OCCLUSION_TYPE_STRICT;
                sphere.occlusionType = BABYLON.AbstractMesh.OCCLUSION_TYPE_OPTIMISITC;

                // number of retry before break the query.
                //sphere.occlusionRetryCount = 3;

                var ground = BABYLON.Mesh.CreateGround('ground1', 6, 6, 2, scene);
                return scene;
            }

            var scene = createScene();

            engine.runRenderLoop(function () {
                scene.render();
                console.log(sphere.isOccluded);
            });`

It will log the state of the Mesh if occluded or not.


IbraheemOsama:
@deltakosh will start with the documentation once the PR is approved :) and Happy Birthday

IbraheemOsama:
The FPS will be reduced because of the console.log, you can remove the console.log to get 60f/s :)

Tools/Gulp/config.json
```diff
@@ -80,7 +80,8 @@
                 "../../src/Mesh/babylon.mesh.vertexData.js",
                 "../../src/Mesh/babylon.geometry.js",
                 "../../src/PostProcess/babylon.postProcessManager.js",
-                "../../src/Tools/babylon.performanceMonitor.js" 
+                "../../src/Tools/babylon.performanceMonitor.js",
+                "../../src/Rendering/babylon.occlusionBoundingBoxRenderer.js"
```
deltakosh: Could we move it to a non core workload?

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -121,6 +127,32 @@
         // Properties
         public definedFacingForward = true; // orientation for POV movement & rotation
         public position = Vector3.Zero();
+
+        private _occlusionBoundingBoxRenderer: OcclusionBoundingBoxRenderer;
+        private _gl = this.getEngine()._gl;
```
deltakosh: by convention everything related to gl has to be handled inside the engine

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -121,6 +127,32 @@
...
+        public occlusionRetryCount = -1;
+        protected _isOccluded = false;
+        get isOccluded(): boolean {
+            return this._isOccluded;
+        }
+        set isOccluded(value: boolean) {
+            this._isOccluded = value;
+        }
+
+        public occlusionQuery = this._gl.createQuery();
```
deltakosh: same here, this has to be store inside the engine

src/Rendering/babylon.occlusionBoundingBoxRenderer.ts
```diff
@@ -0,0 +1,82 @@
+module BABYLON {
+    
+    export class OcclusionBoundingBoxRenderer {
```
deltakosh: I would prefer adding paramters to actual boundingbox renderer instead of creating a new class

IbraheemOsama: Yep, I should've used the one within the scene.

IbraheemOsama:
Glade you like it :) will apply your comments once I'm back from work.

IbraheemOsama:
@deltakosh I updated the PR with your comments :)

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -7,6 +7,12 @@
         private static _BILLBOARDMODE_Z = 4;
         private static _BILLBOARDMODE_ALL = 7;
 
+        public static OCCLUSION_TYPE_NO_VALUE = 0;
```
deltakosh: Would rename it to NONE and we're good to merge :D

deltakosh:
Thanks!!

deltakosh:
Do you mind fixing the conflicts?

IbraheemOsama:
Will merge from master to resolve the conflicts :(

deltakosh:
Now we need an update of the documentation and the what's new (in dist/preview version)

IbraheemOsama:
May I ask how to update the documentations because when I reach this page
http://doc.babylonjs.com/generals/approved_naming_conventions
and I click update documentation it is taking me to creating_the_mini-fied_version page :)

deltakosh:
woot
need to fix that :)
Just go there https://github.com/BabylonJS/Documentation

IbraheemOsama:
will update the documentation in few hours, I'll have to catch the train :D

PR #2740, opened 2017-09-05
Added logarithmic depth support to outlineRenderer

uraani:
Added usage of logarithmic depth when the sub mesh material uses it.
Added the logDepth includes to outline vertex and fragment shader.

### DIFF
--- a/src/Rendering/babylon.outlineRenderer.ts
+++ b/src/Rendering/babylon.outlineRenderer.ts
@@ -24,6 +24,13 @@
             var material = subMesh.getMaterial();
 
             engine.enableEffect(this._effect);
+
+            // Logarithmic depth
+            if(material instanceof StandardMaterial && material.useLogarithmicDepth)
+            {
+                this._effect.setFloat("logarithmicDepthConstant", 2.0 / (Math.log(scene.activeCamera.maxZ + 1.0) / Math.LN2));
+            }
+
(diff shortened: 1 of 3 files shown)


src/Rendering/babylon.outlineRenderer.ts
```diff
@@ -24,6 +24,13 @@
             var material = subMesh.getMaterial();
 
             engine.enableEffect(this._effect);
+
+            // Logarithmic depth
+            if(material instanceof StandardMaterial && material.useLogarithmicDepth)
```
deltakosh: no need to check if instanceOf..Just test if useLogarithmicDepth is true (this will work for PBR as well )

uraani: typescript compiler won't like it?

deltakosh: Cast it to <any> :)

uraani: Do you mean this:
```
var logMat = <PBRBaseMaterial> material || <StandardMaterial> material;
if(logMat && logMat.useLogarithmicDepth)
//...
```
?

deltakosh: Nope I really mean <any>

deltakosh: Sorry < any >

deltakosh: Yes:)

PR #2742, opened 2017-09-05
Now bind draw buffers correctly + removed conflict with depth rendere

julien-moreau:
r (gl.COLOR_ATTACHMENT_0 !== COLOR_ATTACHMENT0_WEBGL) + disable generate depth texture in geometry renderer


src/babylon.engine.ts
```diff
@@ -1076,17 +1077,28 @@
 
                 if (drawBuffersExtension !== null) {
                     this._caps.drawBuffersExtension = true;
-                    this._gl.drawBuffers = drawBuffersExtension.drawBuffersWEBGL;
+                    this._gl.drawBuffers = drawBuffersExtension.drawBuffersWEBGL.bind(drawBuffersExtension);
                     this._gl.DRAW_FRAMEBUFFER = this._gl.FRAMEBUFFER;
                     
                     for (var i = 0; i < 16; i++) {
-                        this._gl["COLOR_ATTACHMENT" + i] = drawBuffersExtension["COLOR_ATTACHMENT" + i + "_WEBGL"];
+                        this._gl["COLOR_ATTACHMENT" + i + "_WEBGL"] = drawBuffersExtension["COLOR_ATTACHMENT" + i + "_WEBGL"];
```
deltakosh: Why this?

julien-moreau: Because COLOR_ATTACHMENT0 is different from COLOR_ATTACHMENT0_WEBGL
And COLOR_ATTACHMENT0 is used by engine.createRenderTargetTexture. It created a conflict has COLOR_ATTACHMENTX_WEBGL is only for multiple render targets, not single

deltakosh: OK makes sense

PR #2753, opened 2017-09-08
glTF loader bug fixes

bghgary:
- Fixed an issue with texture caching that caused incorrect texture settings for textures with alpha.
- Added missing code to load alphaCutoff from materials.
- Minor cleanup.


bghgary:
@deltakosh Please wait. I found a bug.

bghgary:
Fixed. Please merge when ready.

PR #2770, opened 2017-09-11
Change Default for maxHeight for CreateGroundFromHeightMap

BabylonJSGuide:
(change line 731 - 10.0 to 1.0) In documentation default value is given as 1, in code it is 10. For default ground size of 10 x 10 I think the default size of 1 makes more sense than 10. Having done a search in Playground for MeshBuilder.CreateGroundFRomHeightMap all the PG where this appears set their own maxHeight so probably not a problem to change. However if you think this change might cause issues for users then just change default in documentation 10 instead.


deltakosh:
I'm fine with this change

PR #2785, opened 2017-09-12
Updated motion controllers to use final GLB assets.

leweaver:
Fixed pointing pose on motion controllers to read from the mesh


src/Gamepad/Controllers/babylon.windowsMotionController.ts
```diff
@@ -24,10 +23,9 @@ module BABYLON {
     }
 
     export class WindowsMotionController extends WebVRController {
-        private static readonly MODEL_BASE_URL:string = 'https://controllers.babylonjs.com/microsoft/';
+        private static readonly MODEL_BASE_URL:string = '/assets/meshes/controllers/microsoft/';
```
deltakosh: This looks suspicious :)

src/Gamepad/Controllers/babylon.windowsMotionController.ts
```diff
@@ -247,15 +246,16 @@ module BABYLON {
             let childMesh : AbstractMesh = null;
             for (let i = 0; i < meshes.length; i++) {
                 let mesh = meshes[i];
-                if (mesh.id === WindowsMotionController.MODEL_ROOT_NODE_NAME) {
-                    // There may be a parent mesh to perform the RH to LH matrix transform.
+
+                // There may be a parent mesh to perform the RH to LH matrix transform.
+                if (mesh.parent && mesh.parent.name === WindowsMotionController.GLTF_ROOT_TRANSFORM_NAME)
```
deltakosh: The parent is now returned as part of the meshes list

leweaver: Ok - so does that mean we are guaranteed to find a mesh with a null parent? In which case I can drop the .name check altogether

src/Gamepad/Controllers/babylon.windowsMotionController.ts
leweaver: Yeh accidental commit sorry - removed!

src/Gamepad/Controllers/babylon.windowsMotionController.ts
leweaver: OK I updated this section, no longer testing for 'root' and it is working great :)

deltakosh:
Perfect! thanks!

PR #2816, opened 2017-09-16
Add missing mesh scaling property to inspector

bghgary:


deltakosh:
We have a PR waiting for validation that will expose all properties

PR #2823, opened 2017-09-19
Master

MackeyK24:
BabylonJS Toolkit - Support Plugins Source Projects


deltakosh:
Lookd like a LOT of files, is it correct? do we need all sources files for dependencies?

MackeyK24:
Its All the Source For FreeImage... I thought you might people to be able to fully REBUILD that plugin... All these files are compiled and in the Toolkit/Assets/Babylon/Plugins as dlls... That i pre-built from these projects... If you want the WHOLE souces for all the plugins that DID NOT come from nuget (Ones I had to Build From Source)... These are them :)

MackeyK24:
You can cancel the Pull and we can just go with the prebuilt binaries... And i can put a URL link somewhere in docs that point to where they can download the source for FreeImage... And AjaxMin... I made that Babylon.Terrains Generator Specifically for BabylonJS... Might wanna keep that source project

deltakosh:
OK let's close it as I do not want our repo to host other projects than ours:)

Issue #2529, opened 2017-07-14
Create headless version of babylon.js


deltakosh:
A version which could be used with no DOM

nicktendo:
:-)

furcatomasz:
Great idea! I still waiting for headless version of BJS. Do you have ETA for that?

deltakosh:
Not yet unfortunately

a-galminas:
Any news on headless?

deltakosh:
Not yet. I will post here when the work will start

deltakosh:
https://github.com/BabylonJS/Babylon.js/commit/4c1e3fecf38f934d15010fbe6d28a10d7f6d8d02
http://doc.babylonjs.com/generals/nullengine

furcatomasz:
Yeah!!! :) Thanks @deltakosh

ransico:
Updated documentation link: http://doc.babylonjs.com/classes/3.1/nullengine

PR #2830, opened 2017-09-20
WebVR improvements

leweaver:
- Now handles headset connection/disconnection events
- Switches back to window render loop when headset is disconnected
- Bugfixes to allow you to enter VR more than once
- Added Observables so apps can handle cases where requestPresent doesn't return instantaneously 
- Remove support for finding headset by id (which is an anti-pattern)
- Added WebVR sample (used as reference for WebVR supprot checklist)
- Fixed camera mouse drag disabled when in VR (should be when in VR at all, not simply when the headset has pointer restriction)

Note that this has yet to be tested on Vive & Oculus. I will test on Vive tomorrow in the office, but will need someone to assist with Oculus testing.

### DIFF
--- a/src/Cameras/VR/babylon.webVRCamera.ts
+++ b/src/Cameras/VR/babylon.webVRCamera.ts
@@ -95,7 +95,23 @@ module BABYLON {
             }
 
             //enable VR
-            this.getEngine().initWebVR();
+            var engine = this.getEngine();
+            engine.initWebVR().add(() => {
+                
+                this._vrDevice = engine.getVRDevice(this.webVROptions.displayName);
+                this._vrEnabled = !!this._vrDevice;
+
+                if (!this._vrEnabled) {
+                    return;
+                }
+
+                //reset the rig parameters.
+                this.setCameraRigMode(Camera.RIG_MODE_WEBVR, { parentCamera: this, vrDisplay: this._vrDevice, frameData: this._frameData, specs: this._specsVersion });
+
+                if (this._attached) {
+                    this.getEngine().enableVR(this._vrDevice)
+                }
+            });
 
             //check specs version
             if (!window.VRFrameData) {
@@ -105,22 +121,6 @@ module BABYLON {
             } else {
                 this._frameData = new VRFrameData();
             }
-
-            this.getEngine().getVRDevice(this.webVROptions.displayName, device => {
-                if (!device) {
-                    return;
-                }
-
-                this._vrEnabled = true;               
-                this._vrDevice = device;
-
-                //reset the rig parameters.
-                this.setCameraRigMode(Camera.RIG_MODE_WEBVR, { parentCamera: this, vrDisplay: this._vrDevice, frameData: this._frameData, specs: this._specsVersion });
-
-                if (this._attached) {
-                    this.getEngine().enableVR(this._vrDevice)
-                }
-            });                
 
             /**
              * The idea behind the following lines:
--- a/src/Cameras/VR/babylon.webVRCamera.ts
+++ b/src/Cameras/VR/babylon.webVRCamera.ts
@@ -40,7 +40,7 @@ module BABYLON {
     export class WebVRFreeCamera extends FreeCamera implements PoseControlled {
         public _vrDevice = null;
         public rawPose: DevicePose = null;
-        private _vrEnabled = false;
+        private _onVREnabled: (success: boolean) => void;
         private _specsVersion: number = 1.1;
         private _attached: boolean = false;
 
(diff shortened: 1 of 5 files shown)


src/babylon.engine.ts
```diff
@@ -1578,94 +1589,106 @@
...
-
-                this._vrDisplays = devices.filter(function (device) {
-                    return device instanceof VRDisplay;
-                });
-
-                return this._vrDisplays;
+                this._vrSupported = true;
+                // note that devices may actually be an empty array. This is fine;
+                // we expect this._vrDisplay to be undefined in this case.
+                return this._vrDisplay = devices[0];
```
RaananW: Can't we expect a single host to support more than one device parallel?

src/babylon.engine.ts
```diff
@@ -1578,94 +1589,106 @@
...
-                    } else {
-                        //choose the first one
-                        callback(devices[0]);
-                    }
-                } else {
-                    Tools.Error("No WebVR devices found!");
-                    callback(null);
+            if (!this._onVrDisplayConnect) {
+                this._onVrDisplayConnect = (event) => {
+                    this._vrDisplay = event.display;
```
RaananW: I am trying to understand the need in both functions - the event based (this one) and _getVRDisplays, which also sets `this._vrDisplay` . Wouldn't it be better to choose one of the two? I know there was a problem with events back when I implemented WebVR.

src/babylon.engine.ts
```diff
@@ -1401,7 +1408,10 @@
 
             if (this._activeRenderLoops.length > 0) {
                 // Register new frame
-                this._frameHandler = Tools.QueueNewFrame(this._bindedRenderFunction, this._vrDisplayEnabled);
+                var requester = window;
```
RaananW: Would be better to keep `window` reference to a minimum, especially due to https://github.com/BabylonJS/Babylon.js/issues/2529

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -194,11 +196,11 @@ module BABYLON {
         } 
 
         public _checkInputs(): void {
-            if (this._vrEnabled) {
+            if (this._vrDevice && this._vrDevice.isPresenting) {
                 if (this._specsVersion === 1.1) {
                     this._vrDevice.getFrameData(this._frameData);
                 } else {
-                    //backwards comp
+                    // TODO: Does backwards comp need to be here any more? The Engine class doesn't support it any more.
```
RaananW: We support WebVR 1.0 only because of the GearVR browser. It was added after being completely removed. If the GearVR now supports 1.1 (I can't seem to find the information), we don't need 1.0 . if it doesn't, we need the full support.

Same goes to the changes in the engine.

src/babylon.engine.ts
leweaver: You need to support both, since a user may decide to connect a headset after the page has loaded. Additionally, they may disconnect/reconnect the headset for any reason at any time.

src/babylon.engine.ts
leweaver: perhaps I could just pass null in to the function instead, so it defaults?

src/Cameras/VR/babylon.webVRCamera.ts
leweaver: The Samsung browser on GearVR was upgraded to 1.1 a little while back now, so potentially safe to remove this. I propose full removal of 1.0 support in a subsequent PR to make things easier?

src/babylon.engine.ts
leweaver: Right now,  no current browser supports multiple concurrent connected VRDisplays. This will likely change when vendors implement WebVR 2.0.

It is an interesting topic of discussion on whether Babylon should support multiple displays for WebVR 1.1. If it does, I can see two levels of support:
- Giving user choice of which headset to render to, when attempting to Enter VR
- Support  rendering to multiple headsets concurrently

The second choice seems impossible to implement now, as there is no way to validate/test the implementation - so will ignore that.

The first choice is more interesting. I removed the existing support primarily because selecting a VRDisplay based on a name via configuration is risky behavior for cross compatibility reasons - and there is no vendor support for the feature in WebVR 1.1. If a 'choice of headset' feature is implemented, it should probably be done as part of the vrExperienceHelper class via the user interaction when they try to Enter VR. Otherwise, we will have huge compatibility issues with every developer implementing their own selection methods via device names... which will also make it harder to migrate them to a new 2.0 implementation. This being the case, perhaps instead could add a new method to Engine that the vrExperienceHelper could call prior to activating the camera. It could show the user a choice on screen if there is more than 1 display connected:

```
public chooseActiveVRDisplay(chooseFunction: (any[]) => any) {
  if (this._VRDisplay && this._VRDisplay.isPresenting) {
    // Throw error?
    // Also throw error if we are waiting for a requestPresent call to resolve?
  }
  // Store a list of all connected displays on Engine. 
  var callback = (newDisplay) => {
    if (!newDisplay) { return; }
    if (!this._vrDisplays.contains(newDisplay)) { return; }
 
    if (newDisplay !== this._VRDisplay) {
      this._VRDisplay = newDisplay; 

      // Notify watchers that the display changed
      var eventArgs = {
        vrDisplay: this._vrDisplay,
        vrSupported: this._vrSupported
      };
      this.onVRDisplayChangedObservable.notifyObservers(eventArgs);
    }
  }
  var newDisplay = chooseFunction(this._vrDisplays, callback);
}
```

Of course - this doesn't solve the issue of automatically launching VR on page load (on browsers that support this). For this case, it might be best to stick with the convention adopted by other implementers and simply enter VR on device 0.

src/Cameras/VR/babylon.webVRCamera.ts
davrous: Yes, good idea, let's try to split things. I wasn't sure for GearVR neither and wasn't able to had confirmation after looking on the web.

davrous:
On my side, things look good on the VRHelper side. Thanks for the improvements, it was on my to do list. I'm going to add also some events when entering in VR to be able to notify external code that need to do their job if wanted (like custom teleportation, whatever).

src/babylon.engine.ts
RaananW: would be great, but shouldn't prevent merging.

src/Cameras/VR/babylon.webVRCamera.ts
deltakosh: @RaananW  : Fancy doing the other PR to remove webvr 1.0?

RaananW: @deltakosh - will do.

deltakosh: Thanks!

leweaver:
Thanks all!

PR #2834, opened 2017-09-20
allow whitespace before ( in shaders

makc:
as it stands, `texture2D (` will not be converted and then browsers complain about shaders. same issue in preview branch, of course.


deltakosh:
Thanks :)

PR #2838, opened 2017-09-21
Update babylon.scene.ts

aWeirdo:
Fix for FF bug;
http://www.html5gamedevs.com/topic/32912-mousedown-action-sticks-on-canvas-after-mouseout/

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -1484,6 +1484,9 @@
             };
 
             this._onPointerUp = (evt: PointerEvent) => {
+		if (!this._isButtonPressed) {
+			return;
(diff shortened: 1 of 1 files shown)


RaananW:
Hi,

Is it only for the *up event? Are the rest of the events working fine?

I assume that the event is not detected in Firefox. Did you make sure it's not CSS related?

aWeirdo:
Hi @RaananW 
Other events seem fine,
so yes, it appears the pointerup event isn't detected after the pointer leaves the canvas in FF, even off-page, url bar, etc, so i don't believe it's CSS related.

PR #2857, opened 2017-09-22
Export alpha from PBR materials

noalak:
Export alpha from PBR materials
Minor fixes


deltakosh:
Could you rebuild the binary?

noalak:
Done

PR #2860, opened 2017-09-23
clean bone matrice weights - shadow problem

MatuskaLaszlo:


src/Loading/babylon.sceneLoader.ts
```diff
@@ -331,6 +409,7 @@
                     }
 
                     if (onSuccess) {
+                        SceneLoader._CleanMatricesIndices(scene.meshes,scene.skeletons);
```
deltakosh: why only if Onsuccess is defined? We should do it everytime

src/Loading/babylon.sceneLoader.ts
```diff
@@ -189,6 +189,82 @@
             }
         }
 
+        private static _CleanMatricesIndices(meshes:AbstractMesh[], skeletons:Skeleton[]): void {
```
deltakosh: I would like to keep it in the geometry class.
After loading a geometry we can run the cleanup.

src/Loading/babylon.sceneLoader.ts
```diff
@@ -189,6 +189,82 @@
             }
         }
 
+        private static _CleanMatricesIndices(meshes:AbstractMesh[], skeletons:Skeleton[]): void {
+            if (!SceneLoader._CleanBoneMatrixWeights) {
+                return;
+            }
+            for (var i = 0; i < skeletons.length; i++) {
+                var skeleton = skeletons[i];
+                var noInfluenceBone = new Bone("noInfluence", skeleton, null, Matrix.Identity(), Matrix.Identity());
```
deltakosh: this is done automatically by the skeleton when preparing the matrices:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Bones/babylon.skeleton.ts#L241

MatuskaLaszlo: Ok so it is always there at skeleton.bones.length, cool!

deltakosh:
up?

MatuskaLaszlo:
Sorry I'm on a vacation in beautiful Nice, France, but I will continue in a few days! 😁

MatuskaLaszlo:
I will send another PR with the changes sorry

PR #2863, opened 2017-09-24
MTL Loader didn't parse the strings into floats.

RaananW:


RaananW:
Man, you're quick :smile:

deltakosh:
Lol

RaananW:
Would be good to rebuild the library and repackage. I believe it never worked correctly until now.

deltakosh:
I can do it with Monday nightly

PR #2869, opened 2017-09-26
Texture init observables

royibernthal:


RaananW:
So, I suddenly wonder what's the use of the onAfter function. As almost everything happens async, it's practically the same. The existing onLoaded observer can be used instead.

deltakosh:
I agree, Only onBefore is useful

src/Materials/Textures/babylon.texture.ts
```diff
@@ -117,6 +117,8 @@
 
             scene = this.getScene();
 
+            scene.getEngine().onBeforeTextureInitObservable.notifyObservers(this);
```
deltakosh: we need to make sure that this.url is used later on by the constructor (and not just url)

royibernthal:
True I didn't notice that. Check my new commit.

deltakosh:
Sounds good!

royibernthal:
Great :) Any idea when you're gonna build it for dist?

deltakosh:
In the next hour :)

PR #2874, opened 2017-09-27
Xbox One Right Stick Axes Fix

MackeyK24:
Browser now properly report right stick as 2 and 3 instead of 3 and 4

### DIFF
--- a/src/Gamepad/babylon.xboxGamepad.ts
+++ b/src/Gamepad/babylon.xboxGamepad.ts
@@ -56,7 +56,7 @@
         private _isXboxOnePad: boolean = false;
 
         constructor(id: string, index: number, gamepad: any, xboxOne: boolean = false) {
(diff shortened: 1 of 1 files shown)


deltakosh:
How did you manage to get the xbox one boolean to true? All my xbox one controllers have an ID = "Xbox 360 controller"

MackeyK24:
Yeah... that a windows thing... Uses Generic 'XInput' mappings

MackeyK24:
But the current Codebase for Babylon Gamepad Xbox Once Controller Support (which i added) uses 3 and 4 for the Right Stick Axes... There is no 3 and 4 axes anymore... it reports properly as 2 and 3.. Can you please allow that small fix. Some users like MrVR are having problems because the code base is looking for right stick Axes on 3 and 4 which do not exist.

Issue #2827, opened 2017-09-19
Updates for Inspector


deltakosh:
- [x] Selecting a mesh does not select it in the list of meshes. Only properties are updated
- [x] Use real sliders (controls) when possible
- [x] We should order sub properties:
![image](https://user-images.githubusercontent.com/1306056/31021648-571e3766-a4eb-11e7-9674-03e9dfdf9308.png)
- [x] We should round numbers:
![image](https://user-images.githubusercontent.com/1306056/30620252-1dfde506-9d59-11e7-9d3f-4dda75050737.png)
- [x] Add a margin to make the list clearer:
![image](https://user-images.githubusercontent.com/1306056/30620269-2a435512-9d59-11e7-917e-5731d91afa97.png)
- [x] Weird bottom bar:
![image](https://user-images.githubusercontent.com/1306056/30620297-4f0df118-9d59-11e7-9dab-b1597f0a174d.png)
- [x] Center color box on the line:
![image](https://user-images.githubusercontent.com/1306056/30620310-5cab8e16-9d59-11e7-8e15-23b84d4d6217.png)
- [x] Hitting tab should validate an edited field (like Enter)
- [x] Hitting tab and then clicking on a field produces a weird result:
![image](https://user-images.githubusercontent.com/1306056/30620349-a6610dce-9d59-11e7-9fb8-14e55194f41d.png)
- [x] Boolean fields should be represented with check boxes
- [x] Avoid using double scroll bar:
![image](https://user-images.githubusercontent.com/1306056/30620371-d0d6616c-9d59-11e7-9f14-45ffe0edbb02.png)

bghgary:
It would be good to fix the order of components the drop down too. For example, in the first example above, the order of RGB field is B, G, R in the drop down. Some XYZs are also in the wrong order:

![image](https://user-images.githubusercontent.com/19158740/30708609-50e1d674-9eb4-11e7-9715-bd87358d3920.png)

treefort:
Not sure if a new issue would be more appropriate, but I had a few thoughts as well:

- Up/down arrows "nudge" numerical fields by +/- 1, respectively. holding down shift changes this to +/- 10.
- Ability to see a wireframe representation of physics impostors
- Proper integration into chrome's devtools (as an extension)
- Fix redundant CSS includes:
![image](https://user-images.githubusercontent.com/1132492/30813269-f92340ba-a1da-11e7-8783-b4d3e31e1765.png)

[...]

PR #2885, opened 2017-09-29
Fix issue #2827 without slider fonction

QuentinRillet:
https://github.com/BabylonJS/Babylon.js/issues/2827


deltakosh:
Really good job!
one from the comment is missing though:
image (We should order the sub properties)

Issue #2889, opened 2017-10-03
Display async error in the playground PG.


sebavan:
This can be use as a source for the repro https://www.babylonjs-playground.com/#11BH6Z#313

deltakosh:
+ Adding observables for callbacks as well

PR #2890, opened 2017-10-03
onError exception handling

RaananW:
The exception was missing as a (3rd) variable in many of the onError callbacks of the loader.
I also added it to the plugin, so we will have more consistent error handling . Now, when an error occures in the plugin's load function, no exception will be thrown if the onError is defined.

### DIFF
--- a/src/Tools/babylon.assetsManager.ts
+++ b/src/Tools/babylon.assetsManager.ts
@@ -153,7 +153,7 @@
 
     export class TextureAssetTask implements ITextureAssetTask {
         public onSuccess: (task: ITextureAssetTask) => void;
-        public onError: (task: ITextureAssetTask) => void;
+        public onError: (task: ITextureAssetTask, message?: string, exception?: any) => void;
 
         public isCompleted = false;
         public texture: Texture;
@@ -161,7 +161,7 @@
         constructor(public name: string, public url: string, public noMipmap?: boolean, public invertY?: boolean, public samplingMode: number = Texture.TRILINEAR_SAMPLINGMODE) {
         }
 
-        public run(scene: Scene, onSuccess: () => void, onError: () => void) {
+        public run(scene: Scene, onSuccess: () => void, onError: (message?: string, exception?: any) => void) {
 
             var onload = () => {
                 this.isCompleted = true;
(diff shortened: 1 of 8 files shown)


RaananW:
More changes coming up, wait with merging

RaananW:
Don't merge for now, this will also include a fix for https://github.com/BabylonJS/Babylon.js/issues/2889

RaananW:
Now fixing https://github.com/BabylonJS/Babylon.js/issues/2889 . An error message will be logged if no onError is defined. if onError is defined, the exception object can be used to analyze the reason for the error.

src/Tools/babylon.tools.ts
```diff
@@ -394,7 +394,7 @@
             return url;
         }
 
-        public static LoadImage(url: any, onload, onerror, database): HTMLImageElement {
+        public static LoadImage(url: any, onLoad, onError: (message?: string, eception?: any) => void, database): HTMLImageElement {
```
deltakosh: small typo :)

src/Tools/babylon.assetsManager.ts
```diff
@@ -263,6 +263,12 @@
         public onTaskSuccess: (task: IAssetTask) => void;
         public onTaskError: (task: IAssetTask) => void;
 
+        //Observables
+
+        public onTaskSuccessObservable = new Observable<IAssetTask>();
```
deltakosh: I would have added some on each task as well :)

RaananW: Would that make sense? since we register tasks only with the assets manager?

src/Tools/babylon.tools.ts
RaananW: was a test. you passed.

deltakosh: So proud of me :D

src/Tools/babylon.assetsManager.ts
deltakosh: yeah makes sense..It was a test, you passed!

RaananW: I want an observing-observer certificate.

PR #2893, opened 2017-10-03
clean bone matrice weights - shadow problem #2

BitOfGold:
Previous PR was [This](https://github.com/BabylonJS/Babylon.js/pull/2860).
Closed now, I deleted my repo.

I re-done it with the change requests:

- weights are cleaned in Geometry

### DIFF
--- a/src/Loading/babylon.sceneLoader.ts
+++ b/src/Loading/babylon.sceneLoader.ts
@@ -189,6 +189,37 @@
             }
         }
 
+        private static _CleanMatricesIndices(meshes:AbstractMesh[]): void {
+            if (!SceneLoader._CleanBoneMatrixWeights) {
+                return;
+            }
+            for (var i = 0; i < meshes.length; i++) {
+                var mesh = meshes[i];
+                if (mesh.skeleton) {
+                    var noinfluenceBoneIndex = mesh.skeleton.bones.length;
+                    var matricesIndices: number[] | Float32Array = mesh.getVerticesData(VertexBuffer.MatricesIndicesKind);
+                    var matricesIndicesExtra: number[] | Float32Array = mesh.getVerticesData(VertexBuffer.MatricesIndicesExtraKind);    
+                    if (matricesIndices) {
+                        let size = matricesIndices.length;
+                        for (var j = 0; j < size; j++) {
+                            if (matricesIndices[j] < 0) {
+                                matricesIndices[j] = noinfluenceBoneIndex;
+                            }
+                        }
+                        mesh.setVerticesData(VertexBuffer.MatricesIndicesKind, matricesIndices);
+                        if (matricesIndicesExtra) {
+                            for (var j = 0; j < size; j++) {
+                                if (matricesIndicesExtra[j] < 0.0) {
+                                    matricesIndicesExtra[j] = noinfluenceBoneIndex;
+                                }
+                            }
+                            mesh.setVerticesData(VertexBuffer.MatricesIndicesExtraKind, matricesIndicesExtra);
+                        }                    
+                    }    
+                }
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/Loading/babylon.sceneLoader.ts
```diff
@@ -189,6 +189,37 @@
             }
         }
 
+        private static _CleanMatricesIndices(meshes:AbstractMesh[]): void {
```
deltakosh: just wondering why you did not move this code in the geometry as well?

BitOfGold:
OK I managed to put all in geometry only (the code uses scene.getLastSkeletonByID to access the skeleton, because skeletonId is alerady avaiable in ImportGeometry)

src/Mesh/babylon.geometry.ts
```diff
@@ -959,13 +959,19 @@
...
 
                 if (parsedGeometry.matricesWeights) {
-                    Geometry._CleanMatricesWeights(parsedGeometry.matricesWeights, parsedGeometry.numBoneInfluencers);
-                    mesh.setVerticesData(VertexBuffer.MatricesWeightsKind, parsedGeometry.matricesWeights, parsedGeometry.matricesWeights._updatable);
-                }
-
-                if (parsedGeometry.matricesWeightsExtra) {                    
-                    Geometry._CleanMatricesWeights(parsedGeometry.matricesWeightsExtra, parsedGeometry.numBoneInfluencers);
-                    mesh.setVerticesData(VertexBuffer.MatricesWeightsExtraKind, parsedGeometry.matricesWeightsExtra, parsedGeometry.matricesWeights._updatable);
+                    if (parsedGeometry.skeletonId > -1) {
```
deltakosh: This is not correct actually. You can have matrices weights even if no skeleton is affected (as it could be set later).
The check for skeletonId should only be done inside the cleanMatricesWeights

src/Mesh/babylon.geometry.ts
```diff
@@ -959,13 +959,19 @@
...
+                        var skeleton = scene.getLastSkeletonByID(parsedGeometry.skeletonId);
+                        var noinfluenceBoneIndex = skeleton.bones.length;
```
deltakosh: noInfluenceBoneIndex

src/Mesh/babylon.geometry.ts
```diff
@@ -996,25 +1002,52 @@
...
-                    if (matricesWeights[i + j] > biggerWeight) {
-                        biggerWeight = matricesWeights[i + j];
-                        biggerIndex = i + j;
+            for (var i = 0; i < size; i += 4) {
+                let weight = 0.0;
+                let firstZeroWeight = -1;
+                for (var j = 0; j < 4; j++) {
+                    let w = matricesWeights[i + j];
+                    weight += w;
+                    if (w < 1e-3 && firstZeroWeight < 0) {
```
deltakosh: let's declare a const epsilon = 1e-3 at the beginning of the function

src/Mesh/babylon.geometry.ts
```diff
@@ -996,25 +1002,52 @@
...
+                        matricesWeights[i + j] *= mweight;
+                    }
+                    if (matricesWeightsExtra) {
+                        for (var j = 0; j < 4; j++) {
+                            matricesWeightsExtra[i + j] *= mweight;
+                        }    
+                    }
+                } else {
+                    if (firstZeroWeight>=4) {
+                        matricesWeightsExtra[i+firstZeroWeight-4] = 1.0 - weight;
```
deltakosh: please add spaces between all operators

BitOfGold:
no problem, it's much better code now :)
yes I tested with 2 4 and 8 influencer meshes.

PR #2900, opened 2017-10-05
Task state and error handling

RaananW:
A task now has an abstract class from which every task extends. This
class has a state object - INIT, RUNNING, DONE and ERROR.  Also, a new
errorObject was added, in case an error occures. This was done due to
the observers' architecture, which prevents us from passing more than
one variable to the callback.

A task can have en ERROR state, but still be done, if the onSuccess
callback had an exception in it. This is why isCompleted is still very
important and should be used.

The assets manager now uses those callbacks, and also reports errors, in
case they were executed during the onFinish callback.

Not a breaking change, but should be noticed - the run function is now
only implemented one time (in the abstract class) the rest have a
runTask function, which actually runs the code. The run function is
executing the local runTask method. This was done to keep backwards
compatibility.

### DIFF
--- a/src/Tools/babylon.assetsManager.ts
+++ b/src/Tools/babylon.assetsManager.ts
@@ -1,143 +1,160 @@
 ﻿module BABYLON {
+
+    export enum AssetTaskState {
+        INIT,
+        RUNNING,
+        DONE,
+        ERROR
+    }
+
     export interface IAssetTask {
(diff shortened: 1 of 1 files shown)


sebavan:
Let s check on Monday with @deltakosh bit like the idea :-)

deltakosh:
Excellent work !

PR #2907, opened 2017-10-06
Added scope option to Observer

abow:


abow:
In this example PG the name of MyObject will be printed to the console:

http://playground.babylonjs.com/#012I9K

deltakosh:
Wonderful

PR #2911, opened 2017-10-06
Support directory recursion in FilesInput class

bghgary:


bghgary:
@deltakosh Please review before committing.

deltakosh:
Sounds good at first sight. Will do more in depth testing before merging

PR #2913, opened 2017-10-08
only assign Node.uniqueId once

abow:


abow:
Sorry, there is a problem with this.  Don't merge yet.

abow:
It should be good now.

PR #2915, opened 2017-10-08
GUI - Automatically measuring TextBlock size

royibernthal:
The goal is to have TextBlocks automatically set their own width and height according to their measured text size, instead of having to manually set it for each case - which seems tedious and redundant to me.

It can be optional, but working quite a lot with GUI lately I can say this will make TextBlocks much easier and dynamic.

It's very rough as you can see, but the idea is calculating the width and height of a TextBlock every time it's rendered, and setting it to the control.

Issues:

1) Setting width and height at the moment sets the control as dirty which results in infinitely re-rendering it.
2) Doesn't seem to scale properly with idealWidth.
3) Need to calculate the height, probably something like this?

https://stackoverflow.com/questions/1134586/how-can-you-find-the-height-of-text-on-an-html-canvas

A different way to make it optional is to provide a measureText() function in TextBlock which returns accurately measured width and height, then the user can choose if/how to use this information.

I'm using this code to test the TextBlock's borders:
http://www.babylonjs-playground.com/#JR21MP#2

### DIFF
--- a/gui/src/controls/textBlock.ts
+++ b/gui/src/controls/textBlock.ts
@@ -10,6 +10,15 @@ module BABYLON.GUI {
 
         private _lines: any[];
         private _totalHeight: number;
+        private _resizeToFit: boolean = false;
+
+        get resizeToFit(): boolean {
+            return this._resizeToFit;
+        }
+
+        set resizeToFit(value: boolean) {
+            this._resizeToFit = value;
+        }
 
         public get textWrapping(): boolean {
             return this._textWrapping;
@@ -175,8 +184,10 @@ module BABYLON.GUI {
                 if (line.width > maxLineWidth) maxLineWidth = line.width;
             }
 
-            this.width = maxLineWidth + 'px';
-            this.height = rootY + this._fontOffset.height + 'px'; //TODO: need to actually measure height
+            if (this._resizeToFit) {
+                this.width = maxLineWidth + 'px';
+                this.height = this._fontOffset.height * this._lines.length + 'px';
+            }
 
             //console.log('width', maxLineWidth);
             //console.log('height', rootY, this._fontOffset.height);
--- a/gui/src/controls/textBlock.ts
+++ b/gui/src/controls/textBlock.ts
@@ -126,6 +126,32 @@ module BABYLON.GUI {
             }
         }
 
+        public getRawFontSize(): object {
+            var ignoreAdaptiveScaling: boolean = this._fontSize.ignoreAdaptiveScaling;
+
+            this._fontSize.ignoreAdaptiveScaling = false;
+            
+            var maxLineWidth: number = 0;
+
+            var _lines = this.text.split("\n").forEach(_line => {
+                //can't get context here? if so, have 2 calculations in _additionalProcessing instead? one for with idealWidth and one without
+                var lineWidth: number = this._parseLine(_line, context).width;
+
+                if (lineWidth > maxLineWidth) maxLineWidth = lineWidth;
+            });
+
+            this._prepareFont();
+
+            this._fontSize.ignoreAdaptiveScaling = ignoreAdaptiveScaling;
+
+            this._prepareFont();
+
+            return {
+                width: maxLineWidth,
+                height: this._fontOffset.height * this._lines.length + 'px'
+            }
+        }
+
         protected _parseLine(line: string='', context: CanvasRenderingContext2D): object {
           return {text: line, width: context.measureText(line).width};
         }
--- a/gui/src/controls/textBlock.ts
+++ b/gui/src/controls/textBlock.ts
@@ -86,6 +86,7 @@ module BABYLON.GUI {
                     break;
             }
 
+
             context.fillText(text, this._currentMeasure.left + x, y);
         }
 
@@ -165,10 +166,20 @@ module BABYLON.GUI {
 
             rootY += this._currentMeasure.top;
 
+            var maxLineWidth: number = 0;
+
             for (var line of this._lines) {
                 this._drawText(line.text, line.width, rootY, context);
                 rootY += this._fontOffset.height;
+
+                if (line.width > maxLineWidth) maxLineWidth = line.width;
             }
+
+            this.width = maxLineWidth + 'px';
+            this.height = rootY + this._fontOffset.height + 'px'; //TODO: need to actually measure height
+
+            //console.log('width', maxLineWidth);
+            //console.log('height', rootY, this._fontOffset.height);
         }
     }
 }
--- a/gui/src/controls/textBlock.ts
+++ b/gui/src/controls/textBlock.ts
@@ -18,6 +18,8 @@ module BABYLON.GUI {
 
         set resizeToFit(value: boolean) {
             this._resizeToFit = value;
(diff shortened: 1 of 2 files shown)


gui/src/controls/textBlock.ts
```diff
@@ -165,10 +166,20 @@ module BABYLON.GUI {
...
+            var maxLineWidth: number = 0;
+
             for (var line of this._lines) {
                 this._drawText(line.text, line.width, rootY, context);
                 rootY += this._fontOffset.height;
+
+                if (line.width > maxLineWidth) maxLineWidth = line.width;
             }
+
+            this.width = maxLineWidth + 'px';
```
deltakosh: This must be under a boolean as people may not want it
height computation must only take in account the fontOffset.height

deltakosh: Wondering if we should also add padding size in the width/height computations

deltakosh:
Another thing to check: when computing lines width, we must ensure that we are not limiting width to current control width and instead consider that the control is virtually infinite

royibernthal:
Works nice now except it doesn't scale with idealWidth, any idea why?

> Another thing to check: when computing lines width, we must ensure that we are not limiting width to current control width and instead consider that the control is virtually infinite

The whole point of this to limit the textblock's control size to be exactly the same as the text size, so I'm not sure I understand what you're trying to say.

Padding - possible - simply increase width and height by padding (width + left + right, height + top + bottom) and center the text both horizontally and vertically? (although I believe this is done by default)

deltakosh:
It could not work with idealWidth as size is not constant. idealWidth expects the user to setup a size and then use it as base for the calculation.

What I was trying to say about computing lines width is that text wrapping must be turned of:f https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/controls/textBlock.ts#L108

You are right for padding :)

royibernthal:
In spite of that, how do we still make it work with idealWidth? :)

An idea would be to measure the text at original font size (not taking into account idealWidth) and set the width and height accordingly. What do you think?

Oh now I got you, these indeed contradict. Something like that I guess?

How would padding work with percents in your opinion? (if at all)

deltakosh:
You can try for idealWidth.. Not sure how to do it yet :)

padding work in percents are actually everything is converted in pixels internally

royibernthal:
Would it be possible to create a function that returns the raw width and height of the text according to font size only while ignoring idealWidth? Or is it problematic with how things work?

Something like saving the current fontSize in the context, measuring the lines like in here:
https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/controls/textBlock.ts#L113

with the original font size, and then reverting to the saved fontSize.

By the way, where is the logic for the fontSize to be dependent on idealWidth?

Padding - so I should use ValueAndUnit/getValueInPixel()?

deltakosh:
Padding: No need to do anything. It will be done automatically
FontSize relatively to idealWidth: https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/controls/control.ts#L881 (done by the `getValue(this._host)` )

So as you can see the fontOffset is built using fontsize according to idealWidth

royibernthal:
Not really elegant at the moment, not to mention not working, but just a little piece of code so you understand my general direction.

Padding - I do have though to read the values which for percents are not returned to me automatically as pixels in this case, correct me if I'm wrong.

> simply increase width and height by padding (width + left + right, height + top + bottom) and center the text both horizontally and vertically

for instance, this would not work with percents in here:
https://github.com/royibernthal/Babylon.js/blob/572821876910248305d45aeb7b59aa3245434de6/gui/src/controls/textBlock.ts#L216

`this.width = maxLineWidth + this.paddingLeft + this.paddingRight + 'px';`

I'd have to sum the internalValues assuming they are pixels, again correct me if I'm wrong.

I think I'm missing something here which you are under the assumption that I understand but I clearly don't :)

deltakosh:
Padding - I will add a way to get value in pixels
FontSize: not sure to understand why you need the raw size. You can just set the width of the textblock with this._width.ignoreAdaptiveScaling = true

deltakosh:
You can now use paddingLeftInPixels and so on

royibernthal:
Great thanks

Font size - for idealWidth to work it needs the raw size, which is then internally multiplied by the adaptive scaling ratio.

If I set the width and height to the adapted font size, the adapted font size will then become the 'raw' size and will be adapted again.

The final goal the way I see it, is every text / font size change, calculate the raw width and height and set it to the TextBlock control. This raw size will then internally adapt according to adaptive scaling.

That way the user will need to specify the font size as it should look like when resolution matches idealWidth / idealHeight, its dimensions will be calculated automatically at ideal size, and it'll consistently adapted to stage size internally by adaptive scaling.

deltakosh:
Hence my comment about using ignoreAdaptativeScaling. Do not change font, leave it like it, just force width and height to not adapt anymore

royibernthal:
If I do that then the processed width of the textblock will ignore adaptive scaling, meaning it'll remain at the original size when the screen is resized, which means the font size will keep adapting but the width won't keep matching the adaptation. What am I missing?

deltakosh:
If you resize the screen, font will update so text width will update so you will update text width

Sounds good to me:)

royibernthal:
Just a clarification - the name getRawFontSize() might be misleading, the purpose of it is to return the measured width and height of the TextBlock when using the raw font size, not the raw font size itself. I'm not even trying to change the font size itself it'd achieve nothing as far as I can tell.

royibernthal:
So in fact I'd turn the adaptive scaling off for the text block and manually adapt the width and height instead?

deltakosh:
Correct

royibernthal:
Got it now :) I was thinking about it the other way around, it might indeed be a lot easier.

However, it looks like width and height are still being adapted, even when I'm setting width for instance to a fixed height (to make sure the problem is not in the measuring). What am I missing now?

royibernthal:
console.log('width', this.width); does return '100px', but visually it is still scaled for a reason I can't point out yet.

royibernthal:
Are you sure? Is it working for you as expected?

deltakosh:
Lol I'm sorry I miss clicked and I had to revert it back
Sorrrrry. do you mind creating the PR again?

royibernthal:
lol happens

I don't mind but I'm not sure I know how to do that.

> There isn’t anything to compare.
> BabylonJS:master is up to date with all commits from royibernthal:master. Try switching the base for your comparison.

although I can clearly see these are different:

https://github.com/royibernthal/Babylon.js/blob/572821876910248305d45aeb7b59aa3245434de6/gui/src/controls/textBlock.ts

https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/controls/textBlock.ts

deltakosh:
Try to fork it again and re-apply your changes (don't forget to save them first ;))

focomoso:
Can someone post an example of how to use this feature somewhere?

deltakosh:
http://doc.babylonjs.com/how_to/gui#resize-to-fit

deltakosh:
http://www.babylonjs-playground.com/#44KYLP#3

PR #2924, opened 2017-10-10
Fixed multitouch bug in scene.

gankun:
A second pointerUp event was not processed when multiple touch
inputs were used. Changed the boolean flag to a counter to handle
multiple inputs.


mohamedmansour:
Lgtm

deltakosh:
Awesome 
Thanks!

PR #2930, opened 2017-10-10
GUI - Automatically measuring TextBlock size

royibernthal:
Continuing this PR:
https://github.com/BabylonJS/Babylon.js/pull/2915

### DIFF
--- a/gui/src/controls/textBlock.ts
+++ b/gui/src/controls/textBlock.ts
@@ -18,8 +18,11 @@ module BABYLON.GUI {
 
         set resizeToFit(value: boolean) {
             this._resizeToFit = value;
-            this._width.ignoreAdaptiveScaling = true;
-            this._height.ignoreAdaptiveScaling = true;
+
+            if (this._resizeToFit) {
+                this._width.ignoreAdaptiveScaling = true;
+                this._height.ignoreAdaptiveScaling = true;
+            }
         }
 
         public get textWrapping(): boolean {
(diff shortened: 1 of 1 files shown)


gui/src/controls/textBlock.ts
```diff
@@ -10,6 +10,17 @@ module BABYLON.GUI {
...
         private _totalHeight: number;
+        private _resizeToFit: boolean = false;
+
+        get resizeToFit(): boolean {
+            return this._resizeToFit;
+        }
+
+        set resizeToFit(value: boolean) {
+            this._resizeToFit = value;
+            this._width.ignoreAdaptiveScaling = true;
```
deltakosh: So :) this should be turned on only if value === true

gui/src/controls/textBlock.ts
```diff
@@ -165,10 +205,25 @@ module BABYLON.GUI {
...
             for (var line of this._lines) {
                 this._drawText(line.text, line.width, rootY, context);
                 rootY += this._fontOffset.height;
+
+                if (line.width > maxLineWidth) maxLineWidth = line.width;
+            }
+
+            if (this._resizeToFit) {
+                this.width = maxLineWidth + 'px';
+                this.height = this._fontOffset.height * this._lines.length + 'px';
```
deltakosh: this is where you can add this.paddingLeftInPixels and this.paddingRIghtInPixels

royibernthal:
Here :)

What do you think about my last questions in the previous PR?

> It looks like width and height are still being adapted, even when I'm setting width for instance to a fixed height (to make sure the problem is not in the measuring). What am I missing now?
> 
> console.log('width', this.width); does return a fixed '100px' as expected when ignoring adaptive scaling, but visually it is still scaled for a reason I can't point out yet.
> 
> I guess what I'm seeing is just the canvas resizing and the text block being resized as well since it's part of it? **At any rate something is still not working there.**

deltakosh:
The canvas will resize and thus the fontsize will resize accordingly:
https://www.babylonjs-playground.com/#6C5J7E

So in turn the update you did will update the width. 

Can you repro the issue you're seeing?

royibernthal:
Try to run this PG along with this PR:
http://www.babylonjs-playground.com/#HFNBXS

The rectangle sets itself to the size of the text so we can see the borders easily, and as you can see something is not right here.

deltakosh:
It works if you do not forget to set your rect as ignoreAdaptiveScaling as well:
https://www.babylonjs-playground.com/#HFNBXS#1

royibernthal:
I had a feeling the problem will end up being with the rectangle =\

Thanks for guiding me through this, looks like it's ready for merge?

deltakosh:
Yep!

deltakosh:
Can you update the doc accordingly?

royibernthal:
Sure, how and where? :)

royibernthal:
Oh got it, on it

deltakosh:
https://github.com/BabylonJS/Documentation/blob/master/content/overviews/Standard/Gui.md
:)

PR #2931, opened 2017-10-11
Fix height canvas

QuentinRillet:


QuentinRillet:
Add fix issue https://github.com/BabylonJS/Babylon.js/issues/2827

PR #2937, opened 2017-10-11
onDrawObservable, raw text size, Watcher

royibernthal:
I found myself needing to listen to draw. I notify the observers in container to avoid doing any rewrites. Hopefully you'll be fine with this addition.

New issue I'd like to discuss (can possibly be part of this PR) -

When I'm creating a TextBlock, in many cases its container's size depends on its size, however since that can't be evaluated at the creation time when resizeToFit is set to true, there's a problem.

1) The solution would be to define the size of the container when textBlock.onDrawObservable is notified (probably consistently and not just on the first render, which would be the default behavior when listening to an observable), however it should be possible to evaluate the raw size of the width and height (sort of reverse idealWidth / idealHeight) in order for it to make sense in controls that are using idealWidth and idealHeight and need to depend on that size.

2) Since a container size can depend on more than 1 TextBlock and/or containers that themselves depend on a TextBlock's size, I was thinking of creating a watcher to aggregate observers. It'd be an all purpose watcher rather than just for this specific case of course, I imagine it something like this:

`new Watcher(controls: Control[], observable: string, callback: (control: Control) => void)`

So for instance:

`new Watcher([textBlock1, textBlock2, someContainer], 'onDrawObservable', (control) => console.log('Control Drawn', control);`

I imagine the Watcher being part of bjs rather than gui.

These solutions will of course be useful in many other cases as they're generic.

What do you think?

### DIFF
--- a/gui/src/controls/container.ts
+++ b/gui/src/controls/container.ts
@@ -144,16 +144,16 @@ module BABYLON.GUI {
                     if (child.isVisible && !child.notRenderable) {
                         child._draw(this._measureForChildren, context);
 
-                        if (child.onDrawObservable.hasObservers()) {
-                            child.onDrawObservable.notifyObservers(child);
+                        if (child.onAfterDrawObservable.hasObservers()) {
+                            child.onAfterDrawObservable.notifyObservers(child);
                         }
                     }
                 }
             }
             context.restore();
 
-            if (this.onDrawObservable.hasObservers()) {
-                this.onDrawObservable.notifyObservers(this);
+            if (this.onAfterDrawObservable.hasObservers()) {
+                this.onAfterDrawObservable.notifyObservers(this);
             }
         }
 
--- a/gui/src/controls/container.ts
+++ b/gui/src/controls/container.ts
@@ -143,10 +143,18 @@ module BABYLON.GUI {
                 for (var child of this._children) {
                     if (child.isVisible && !child.notRenderable) {
                         child._draw(this._measureForChildren, context);
+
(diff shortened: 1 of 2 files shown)


deltakosh:
Should it be onAfterDrawObservable to be coherent with the engine?
Regarding the watcher, I don't hink we will add it to the core for now as we did not use it.

royibernthal:
Sure.
Watcher - I was afraid you might say something like that, no problem.

deltakosh:
I'm working on a kind of watch for you :) stay tuned

royibernthal:
Sounds good :)

deltakosh:
Here we are: http://doc.babylonjs.com/overviews/observables#multiobserver

royibernthal:
Thanks looks good

royibernthal:
Out of curiosity, any reason you don't do something like that?

```
public static Watch<T>(observables: Observable<T>[], callback: (eventData: T, eventState: EventState) => void, mask: number = -1, scope: any = null): MultiObserver<T> {
	let result = new MultiObserver<T>();

	result._observables = observables; 
	result._observers = observables.map(observable => observable.add(callback, mask, false, scope));

	return result;
}
```

deltakosh:
No good reason for one or another:)

royibernthal:
Okay :)

royibernthal:
Hey can you please build it so it's available in preview release?

deltakosh:
I will in a couple of hours

royibernthal:
kk

PR #2942, opened 2017-10-12
Update button.ts

aWeirdo:
Added static "CreateImageWithCenterTextButton" with centered text and stretching image background.
Which i believe is a very common button setup.

https://www.babylonjs-playground.com/#XCPP9Y#288

i wasn't sure what to name it.. i think it ended up rather long, any suggestions?

### DIFF
--- a/gui/src/controls/button.ts
+++ b/gui/src/controls/button.ts
@@ -144,7 +144,7 @@ module BABYLON.GUI {
             var textBlock = new BABYLON.GUI.TextBlock(name + "_button", text);
             textBlock.textWrapping = true;
             textBlock.zIndex = 10; 
-		    textBlock.textHorizontalAlignment = BABYLON.GUI.Control.HORIZONTAL_ALIGNMENT_CENTER;
+            textBlock.textHorizontalAlignment = BABYLON.GUI.Control.HORIZONTAL_ALIGNMENT_CENTER;
             result.addControl(textBlock);   
 
             // Adding image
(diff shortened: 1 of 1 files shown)


gui/src/controls/button.ts
```diff
@@ -136,5 +136,23 @@ module BABYLON.GUI {
...
+            // Adding text
+            var textBlock = new BABYLON.GUI.TextBlock(name + "_button", text);
+            textBlock.textWrapping = true;
+            textBlock.zIndex = 10; 
+            textBlock.textHorizontalAlignment = BABYLON.GUI.Control.HORIZONTAL_ALIGNMENT_CENTER;
+            result.addControl(textBlock);   
+
+            // Adding image
+            var iconImage = new BABYLON.GUI.Image(name + "_icon", imageUrl);
+            iconImage.stretch = BABYLON.GUI.Image.STRETCH_FILL;
```
deltakosh: Should the image be behind the text?

aWeirdo: Correct, the image is a background.
Here's an example which isn't grass :)
https://raw.githubusercontent.com/aWeirdo/Babylon.js/master/button.png

deltakosh: So you should add the image first

aWeirdo: i'll swap them around if you prefer.
i just did it with zIndex as i liked keeping the textPanel as the first child.

deltakosh: The zindex will reord the children

aWeirdo: Ahh, i see, when the control is added it..

PR #2951, opened 2017-10-14
GUI Event Bubbling

royibernthal:
I was thinking of adding event bubbling, which will naturally require each child to have a reference to its parent.

Take this PG for instance:
https://www.babylonjs-playground.com/#XCPP9Y#292

A container that contains an image and a text block.

When clicking on a text block, observers of the text block are notified, but I'd like the event to keep bubbling up as long as there are more parents.

Would you be open to something like that? If so, would you mind if I try to tackle it myself or would you prefer to do it?

I was thinking of something pretty standard, an event-like object that has target, currentTarget, data, stopPropagation() and stopImmediatePropagation().

This event-like object will be notified to observers in current control, and if that control has a parent, the same observable notifying function (e.g. _onPointerDown) will be called on the parent, and so on, until no parent is found.

Since bjs uses observables, it'd look less elegant than an all-purpose event-style solution since it'd have to be manually applied for each different observable, plus it sort of mixes concepts from events and observables which is probably not good) so maybe you have a better direction.

### DIFF
--- a/gui/src/advancedDynamicTexture.ts
+++ b/gui/src/advancedDynamicTexture.ts
@@ -323,7 +323,7 @@ module BABYLON.GUI {
 
                 if (type === BABYLON.PointerEventTypes.POINTERMOVE) {
                     if (this._lastControlOver) {
-                        this._lastControlOver._onPointerOut();
+                        this._lastControlOver._onPointerOut(this._lastControlOver);
                     }
                     
                     this._lastControlOver = null;
@@ -379,7 +379,7 @@ module BABYLON.GUI {
                     this.focusedControl = null;
                 } else if (pi.type === BABYLON.PointerEventTypes.POINTERMOVE) {
                     if (this._lastControlOver) {
-                        this._lastControlOver._onPointerOut();
+                        this._lastControlOver._onPointerOut(this._lastControlOver);
                     }              
                     this._lastControlOver = null;
                 }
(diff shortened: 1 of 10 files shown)


deltakosh:
What about this:
if the current control has no observer try to use the control parent and so on
So we keep the same observable approach but you can catch the event higher in the hierarchy

no big change, no breaking change :)

royibernthal:
what if it does have an observer and you'd like to listen to the child at some place and to the parent somewhere else? (who can have many children with observables)

consider this case:

a list of buttons - each time a button is pressed you change its state as an individual button, and at the same time you listen to the list container and do something with the clicked button.

It'd be even better to allow custom events in controls and make things completely dynamic, but I guess that'd mean replacing the whole observable approach, and that's definitely a breaking change :)

deltakosh:
Yep no way we break compatibility :)
But observers already can set skip next observers flag
So we can just propagate calls up in the hierarchy and if someone set the flag then we stop

royibernthal:
Thought as much :)

Hmm I see, true we could use that.
So I guess applying the logic manually to each existing observable is the only way to go isn't it?
Would you be opposed to adding target and currentTarget to EventState as optional params? If so, where do you think they should go?

Also, on a different subject, I was just wondering, is there any reason you don't check hasObservers() inside notifyObservers() instead of checking it externally?
I just saw some inconsistency - some places check hasObservers() and others don't, and I suppose the purpose of this check is to increase performance?

deltakosh:
I let you try to add the target and current target. Let's try it:)

I agree on the need btw:)

deltakosh:
Regarding hasObservers there is no noticeable difference so it is more a developer preference to use one or the other

royibernthal:
Got it.

Something like that? :) Still need to test it properly.

royibernthal:
Did a little testing, seems like it's working as expected

http://www.babylonjs-playground.com/#IZGN7Y

royibernthal:
Everything seems to be working fine in that PG as well as far as I can see :)

PR #2961, opened 2017-10-16
SPS particle depth sort

jbousquie:
```javascript
// create a particle depth sort enabled SPS
var sps = new BABYLON.SolidParticleSystem("sps", scene, {enableDepthSort: true});
// then later only ...
sps.setParticles();   // and the particle are depth sorted each call
// We can skip the sorting at any time (or reactive it) : sps and camera not moving anymore
sps.depthSortParticles = false;  // true by default when enableDepthSort is set to true
```


deltakosh:
Woot! need a demo and a doc update :D

PR #2974, opened 2017-10-18
Include the GUI library

### LINKED ISSUE #2973: GUI not supported when zipping from Playground
GUI worked fine when coding in browser, but as soon as you zip the file it's missing https://preview.babylonjs.com/gui/babylon.gui.min.js. 

**Note**: I'm happy to fix this since it seems like a one line add to the index.html in Babylon.js/Playground/zipContent.

ashanhol:
Update the zip index.html to include the Babylon.js GUI library. Fixes #2973


deltakosh:
Thank you so much!

PR #2975, opened 2017-10-18
Addition of models for alternative Motion Controller ID

leweaver:
New models to represent the Controllers that come with the Samsung Odyssey headset. They have a slightly different form factor, as well as a different pointing pose.

No code changes to Babylon should be required.


deltakosh:
Cool thanks!

PR #3013, opened 2017-10-24
Load environment texture from .babylon

noalak:
Also update world extends to ignore meshes at infinite distance like default skybox


src/Loading/Plugins/babylon.babylonFileLoader.ts
```diff
@@ -529,6 +529,16 @@
                     }
                 }
 
+                // Environment texture
+                if (parsedData.environmentTexture !== undefined && parsedData.environmentTexture !== null) {
+                    scene.environmentTexture = Texture.Parse(parsedData.environmentTexture, scene, rootUrl);
```
deltakosh: Don't we want to call CreateFromFilteredData here?

noalak: When using CreateFromFilteredData in Sandbox, need to add "file:" before texture name.

deltakosh:
Yes agree and this should work:)
We cannot change the way loader behaves just for the sandbox 
As all other textures work in sandbox, I'm sure we can fix the CreateFromFilteredData to detect when file: need to be added

noalak:
I will have a look at it tomorrow

deltakosh:
Thanks a lot

PR #3015, opened 2017-10-25
Load environment texture from .babylon

noalak:
Also update world extends to ignore meshes at infinite distance like default skybox


deltakosh:
Thanks a lot

PR #3016, opened 2017-10-25
Impostors for child meshes without parent impostor

RaananW:
Playground for better understanding - https://playground.babylonjs.com/#PRHF00

Deals with https://github.com/BabylonJS/Babylon.js/issues/2910

Two new flags were added to the impostor option:
ignoreParent which will ignore the parent when creating the impostor (and thus making no compound)
disableBidirectionalTransformation which will not update the physics engine's body position from babylon (helps performance, disables changing an object's position using Babylon's position and rotationQuaternion)


deltakosh:
Excellent ! That deserves a Doc update:)

PR #3047, opened 2017-10-30
Added an option in GUI slider to have a rounded thumb

bviale:
Hello guys,

I use a lot Babylon GUI and I needed to have a rounded thumb in the slider control.
I have implemented a solution by adding a `isThumbCircle` property on the slider control, initialized to false for backward compatibility. 
I think this is a very nice to have feature to be able to customize the slider control.
Do not hesitate to tell me if you would have implemented the option in a different way, I'll change that.
Also, if you agree with this option, I'll update the documentation accordingly.

Cheers :)


deltakosh:
Good idea! Do you mind updating GUI Doc as well?

bviale:
Here you go :)

deltakosh:
Thanks!

PR #3053, opened 2017-10-31
 Add fallback Texture to TGA and DDS error load 

QuentinRillet:
Hi,

When you load a tga file and the tga files cannot be loaded, nothing is displayed.
It's working with a jpg files.

You can test here: http://www.babylonjs-playground.com/#FT24YX

This PR fixes this


src/Engine/babylon.engine.ts
```diff
@@ -2978,6 +2978,10 @@
                 // fallback for when compressed file not found to try again.  For instance, etc1 does not have an alpha capable type
                 if (isKTX) {
                     this.createTexture(urlArg, noMipmap, invertY, scene, samplingMode, null, onError, buffer, texture);
+                } else if (isTGA && BABYLON.Tools.UseFallbackTexture) {
```
deltakosh: I guess this should be refactored to test isTGA or isDDS :)

deltakosh: I'll do it :D

PR #3064, opened 2017-11-02
Implement touch sensibility for camera

mohamedmansour:
The current sensibility for the camera movement is slower on touch
than desktop. It seems the sensibility should be quicker on touch
to make it feel more responsive on mobile device. The value
should be set to 500 if users want it.


sebavan:
@deltakosh I was not able to figure why it feels different wondering if the H-DPI can not impact the feeling

PR #3085, opened 2017-11-06
Physics pivot fix

RaananW:
Where do I start?...

We had a few problems with the mesh-impostor from glTF files, which had a very interesting parent transformation (that is completely correct, we simply never took the parent into account).

The following changes fully take the object'S parent transformation into account when calculating the physics step of an object. A pivot can now also be set (the object doesn't have to be centered).

The following playground simulates it simply - the sphere has a parent and a pivot point set. 

Without the changes: https://playground.babylonjs.com/#I6EIED#16 . 
With: https://playground.babylonjs.com/#I6EIED#17

Here it is in action:
nov 6 2017 3-15 pm - edited


deltakosh:
You did it!!!

PR #3092, opened 2017-11-06
Fix vertex alpha for glTF loader

bghgary:
- Move transparencyMode from pbrBaseSimpleMaterial to pbrBaseMaterial
- Added Material.needAlphaBlendingForMesh() to decide whether blending should happen instead of renderGroup deciding based on mesh directly

### DIFF
--- a/loaders/src/glTF/2.0/babylon.glTFLoader.ts
+++ b/loaders/src/glTF/2.0/babylon.glTFLoader.ts
@@ -323,6 +323,7 @@ module BABYLON.GLTF2 {
             }
 
             node.babylonMesh = new Mesh(node.name || "mesh" + node.index, this._babylonScene);
+            node.babylonMesh.hasVertexAlpha = true;
 
             this._loadTransform(node);
 
@@ -1383,10 +1384,12 @@ module BABYLON.GLTF2 {
             const alphaMode = material.alphaMode || "OPAQUE";
             switch (alphaMode) {
                 case "OPAQUE": {
-                    // default is opaque
+                    babylonMaterial.transparencyMode = PBRMaterial.PBRMATERIAL_OPAQUE;
                     break;
                 }
                 case "MASK": {
+                    babylonMaterial.transparencyMode = PBRMaterial.PBRMATERIAL_ALPHATEST;
+
                     babylonMaterial.alphaCutOff = (material.alphaCutoff == null ? 0.5 : material.alphaCutoff);
 
                     if (colorFactor) {
@@ -1404,6 +1407,8 @@ module BABYLON.GLTF2 {
                     break;
                 }
                 case "BLEND": {
+                    babylonMaterial.transparencyMode = PBRMaterial.PBRMATERIAL_ALPHABLEND;
+
                     if (colorFactor) {
                         babylonMaterial.alpha = colorFactor[3];
                     }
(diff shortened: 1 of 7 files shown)


bghgary:
@deltakosh Please review

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -323,6 +323,7 @@ module BABYLON.GLTF2 {
             }
 
             node.babylonMesh = new Mesh(node.name || "mesh" + node.index, this._babylonScene);
+            node.babylonMesh.hasVertexAlpha = true;
```
deltakosh: always?

src/Materials/PBR/babylon.pbrBaseMaterial.ts
```diff
@@ -487,13 +489,48 @@
...
+        /**
+         * Specifies whether or not the meshes using this material should be rendered in alpha blend mode.
+         */
+        public needAlphaBlending(mesh: AbstractMesh): boolean {
             if (this._linkRefractionWithTransparency) {
                 return false;
             }
-            return (this.alpha < 1.0) || (this._opacityTexture != null) || this._shouldUseAlphaFromAlbedoTexture();
+
+            if (this._transparencyMode === PBRMaterial.PBRMATERIAL_OPAQUE ||
```
deltakosh: what if I set alpha = 0.5? Do we want to give the last word to the transparencyMode?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: Yes. If vertex alpha is not available, it will not use it. The name is a bit confusing.

src/Materials/PBR/babylon.pbrBaseMaterial.ts
bghgary: Yes, this is what the glTF spec says. https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#alpha-coverage

bghgary:
@sebavan has some good suggestions for not breaking existing code. Working on it.

src/Materials/PBR/babylon.pbrBaseMaterial.ts
```diff
@@ -487,13 +492,55 @@
...
+            this._forceAlphaTest = (value === PBRMaterial.PBRMATERIAL_ALPHATESTANDBLEND);
+
+            this._markAllSubMeshesAsTexturesDirty();
+        }
+
+        /**
+         * Specifies whether or not the meshes using this material should be rendered in alpha blend mode.
+         */
         public needAlphaBlending(): boolean {
+            return false;
```
deltakosh: for consistency I would prefer having this one not defined at all and rely on parent

bghgary: Can't do that if we want transparencyMode to override it unless I also check transparencyMode here.

deltakosh: Perhaps then we should do it then
I want people to be able to rely on this function if they need it

bghgary: Agreed. Fixed.

PR #3093, opened 2017-11-07
fixed TypeScript error in babylon.d.ts if strictFunctionTypes option is enabled (TypeScript 2.6)

Kesshi:


src/Tools/babylon.assetsManager.ts
```diff
@@ -186,8 +186,6 @@ module BABYLON {
     }
 
     export interface ITextureAssetTask extends IAssetTask {
-        onSuccess: (task: ITextureAssetTask) => void;
```
RaananW: Just for me to understand - is this also related to ts 2.6?

Kesshi: Yes. Typescript complains that the definition of onSuccess is not identical to the definition of onSuccess in IAssetTask. (if strictFunctionTypes is enabled)

deltakosh:
Thanks a lot

Issue #2868, opened 2017-09-26
Physics Heightmap not taking pivot transformation in account


RaananW:
See - http://www.html5gamedevs.com/topic/33094-rotating-a-heightmap-with-physics/

PR #3094, opened 2017-11-07
Another small fix, taking parents' rotation but NOT scaling 

RaananW:
prevents small rotation calculation errors when rotating the object only while setting ignoreParent to true.


RaananW:
This PR also solves https://github.com/BabylonJS/Babylon.js/issues/2868 :

https://playground.babylonjs.com/#UV2S4R#8

PR #3107, opened 2017-11-08
Heightmap Rotation can now be set before the impostor ist created.

RaananW:
Until now it was only possible to set the rotation of the impostor after, but this prevents people from using a serialized scene.


RaananW:
(Just for future reference - this should work after this is merged : https://playground.babylonjs.com/#UV2S4R#10 )

PR #3111, opened 2017-11-09
Change return type of addTextureTask to IAssetTask

ltetzlaff:
Currently it uses the more specific ITextureAssetTask which will disallow e.g. that:
```typescript
const task = this.loader.addTextureTask("download texture", "/textures/" + file)
task.onError = (task, message, exception) => { }
```
Which works for all the other tasks.

Current workaround:
typecasting

### DIFF
--- a/src/Tools/babylon.assetsManager.ts
+++ b/src/Tools/babylon.assetsManager.ts
@@ -304,7 +304,7 @@ module BABYLON {
             return task;
         }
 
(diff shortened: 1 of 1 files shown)


RaananW:
That's interesting! Probably because of the strict function types that was just introduced. I assume you are using typescript 2.6?

I think we will need a better solution that this eventually. I am merging this and will later check how we can improve this.

Thanks a lot!

ltetzlaff:
Wow, that was fast!

Yep, that's probably why.

PR #3121, opened 2017-11-10
Added facet depth sort in FacetData

jbousquie:
This feature solves the self transparency issue by sorting the mesh facets from some location (the camera position by default) just before drawing them.  
The mesh is required to be `updatable`.  
The depth sort is done on each call to `updateFacetData()`. It can be disabled at any time to save CPU cycles if the mesh and the camera don't move any more.  
Usage : 
```javascript
    var mesh = BABYLON.MeshBuilder.CreateTorusKnot("mesh", {updatable: true}, scene);
    mesh.material = mat;                         // transparent material
    mesh.mustDepthSortFacets = true;    // enable the depth sort, can be disabled at any time
    scene.registerBeforeRender(function() {
        mesh.updateFacetData();
    });
```

### DIFF
--- a/src/Mesh/babylon.abstractMesh.ts
+++ b/src/Mesh/babylon.abstractMesh.ts
@@ -36,19 +36,27 @@
         // facetData private properties
         private _facetPositions: Vector3[];             // facet local positions
         private _facetNormals: Vector3[];               // facet local normals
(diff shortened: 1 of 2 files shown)


deltakosh:
really really good. This will require a tiny doc update :) (as well as updating the what's new to credit you)

PR #3124, opened 2017-11-10
added support for shadows to GUI controls

abow:


gui/src/controls/checkbox.ts
```diff
@@ -81,10 +81,19 @@ module BABYLON.GUI {
             if (this._processMeasures(parentMeasure, context)) {
                 let actualWidth = this._currentMeasure.width - this._thickness;
                 let actualHeight = this._currentMeasure.height - this._thickness;
+
+                context.shadowColor = this.shadowColor;
```
deltakosh: I was wondering if we should test if we need to apply it with
(this.shadowBlur || this.shadowOffsetX || this.shadowOffsetY){

abow: I'm not sure if setting those adds any overhead.  If you'd like, I can do a check like that just to be safe.

deltakosh: Yeah just to be sure

PR #3130, opened 2017-11-11
Add options to compile shaders to glTF loader

bghgary:


src/Lights/Shadows/babylon.shadowGenerator.ts
```diff
@@ -605,21 +605,26 @@
         public forceCompilation(onCompiled: (generator: ShadowGenerator) => void, options?: { useInstances: boolean }): void {
             let shadowMap = this.getShadowMap();
             if (!shadowMap) {
+                onCompiled(this);
```
deltakosh: On compiled could be null (even if I forgot to set it to Nullable)

bghgary:
Wait please.

src/Lights/Shadows/babylon.shadowGenerator.ts
bghgary: Okay, I'll make both Nullable then.

PR #3141, opened 2017-11-14
facet depth sort refactor to take in account any indices array type

jbousquie:
Internal improvement.
The facet depth sort has been rewritten this way : 
The indices array (stored in the Geometry object) is no longer modified. It's only read.
Another typed array is used to store the sorted indices and to pass them to the GPU with _updateIndices()_


Temechon:
It seems to have conflicts on abstractMesh.ts...

Temechon:
Conflicts resolved with Jerome

PR #3160, opened 2017-11-16
HD screenshots

fabsharp:
BJS.Tools.CreateScreenshotWithRenderTexture has drawbacks :
- it fails generating ‘HD’ images, >2000x2000 = ‘network error’ 
- it doesn’t support antialiasing
- we can’t set the image’s name to download.

This PR :
- use HTMLCanvasElement.toBlob polyfill instead of HTMLCanvasElement.toDataURL.
- use BJS.RenderTargetTexture.addPostProcess and BABYLON.FxaaPostProcess for antialiasing
- add optional parameters for antialiasing and file’s name without breaking existing method’s signatures

Remarks :

- There’s no sense to have 2 methods to create screenshots, "the result should be the same isn’t it  ?" 
- The code c(sh)ould be refactored : 
1. Tools.DumpFramebuffer would return a CanvasRenderingContext2D (if I want to watermark a screenshot).
2. A generic Tools.ForceImageDownload could be usefull
3. BABYLON.Camera.screenshot(2000, 2000, « test.png », true) ? 
4. No doc for screenshot.

### DIFF
--- a/src/Tools/babylon.tools.ts
+++ b/src/Tools/babylon.tools.ts
@@ -948,7 +948,7 @@
             texture.renderList = null;
             texture.samples = samples;
             if(antialiasing) {
-                texture.addPostProcess(new BABYLON.FxaaPostProcess('antialising', 1.0, scene.activeCamera));
+                texture.addPostProcess(new BABYLON.FxaaPostProcess('antialiasing', 1.0, scene.activeCamera));
             }
             texture.onAfterRenderObservable.add(() => {
                 Tools.DumpFramebuffer(width, height, engine, successCallback, mimeType, fileName);
--- a/src/Tools/babylon.tools.ts
+++ b/src/Tools/babylon.tools.ts
@@ -34,7 +34,7 @@
         public static RegisteredExternalClasses: { [key: string]: Object } = {};
 
         // Used in case of a texture loading problem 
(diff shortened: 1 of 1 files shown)


deltakosh:
we have two render method because one is kept for backwards compatibility (the render target one)
I could use some help regarding the documentation :) just adding comments on the function itself could be a good start

fabsharp:
OK I will update the docs, what for is the _TargetTexture.samples_ argument  ?

Questions 
1. Should I modify Tools.CreateScreenshot to add the fileName parameters too ?
2. Successcallback is tricky because you need to set it to null if you want to force the download + I'm into CanvasRenderingContext2D because I need to put my client logo on screenshot so what do you think about this :

>if(successCallback) { successCallback(base64Image, **screenshotCanvas**) }
>if(fileName) {  screenshotCanvas.toBlob [...] }

instead of
>if(successCallback) { [] } **else** { screenshotCanvas.toBlob [...] } 

would let me :

> Tools.CreateScreenshotUsingRenderTarget(engine, camera, size,  function(base64, canvas) { canvas.getContext("2d").**drawImage(logo)** }, null, null, true, fileName)

deltakosh:
Samples is for webgl2 capable devices to enable multisamplea on the RT texture
1. Yes
2. Good idea

PR #3168, opened 2017-11-16
Added vertex colors to LineMesh and LineSystem

jbousquie:
Based on the initial code from JCPalmer, here are the per point colors for LineMesh and LineSystem
```javascript
var points = [arrayVector3];
var lineColors = [arrayColor4];
var line = BABYLON.MeshBuilder.CreateLines("l", {points: points, colors: lineColors}, scene);
// same for LineSystems
var lines = [arrayVector3[] ];
var colors = [arrayColor4[] ];
var lineSystem = BABYLON.MeshBuilder.CreateLineSystem("ls", {lines: lines, colors: colors}, scene);
```
Note1 : this works also with updatable LineMesh or LineSystem objects as the colors array can be updated live.   
Note2 : when set at construction time, the parameter `colors` overwrite the hypothetical  line or lineSystem `.color` property. If `colors` is passed but undefined, the object color falls back to the `.color` property value.

### DIFF
--- a/src/Mesh/babylon.linesMesh.ts
+++ b/src/Mesh/babylon.linesMesh.ts
@@ -32,13 +32,14 @@
         private _intersectionThreshold: number;
         private _colorShader: ShaderMaterial;
 
-        constructor(name: string, scene: Nullable<Scene> = null, parent: Nullable<Node> = null, source?: LinesMesh, doNotCloneChildren?: boolean, public useVertexColor?: boolean) {
+        constructor(name: string, scene: Nullable<Scene> = null, parent: Nullable<Node> = null, source?: LinesMesh, doNotCloneChildren?: boolean, public useVertexColor?: boolean, public useVertexAlpha?: boolean) {
             super(name, scene, parent, source, doNotCloneChildren);
 
             if (source) {
                 this.color = source.color.clone();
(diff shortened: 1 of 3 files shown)


src/Mesh/babylon.linesMesh.ts
```diff
@@ -43,15 +43,20 @@
 
             this._intersectionThreshold = 0.1;
 
+            var defines: String[] = []; 
             var options = {
                 attributes: [VertexBuffer.PositionKind],
                 uniforms: ["world", "viewProjection"],
-                needAlphaBlending: false,
+                needAlphaBlending: true,
```
deltakosh: I would not recommend turning alpha on everytime

jbousquie:
Now the alphablending is disabled, as before, for the color set with the property `.color`.
It's also disabled by defaut when used with the `colors` array passed parameter unless the optional parameter `useVertexAlpha` is set to true.

```javascript
// no alpha blending
var lineSystem = BABYLON.MeshBuilder.CreateLineSystem("ls", {lines: lines, colors: colors}, scene);
// explicit alpha blending
var lineSystem = BABYLON.MeshBuilder.CreateLineSystem("ls", {lines: lines, colors: colors, useVertexAlpha: true}, scene);
```

PR #3183, opened 2017-11-19
Master

MackeyK24:
Added json scene support for texture noMipmap parsing


src/Materials/Textures/babylon.texture.ts
```diff
@@ -431,23 +431,27 @@
             }
 
             var texture = SerializationHelper.Parse(() => {
+                var generateMipMaps: boolean = true;
+                if (parsedTexture.noMipmap) {
+                    generateMipMaps = false;
+                }
                 if (parsedTexture.mirrorPlane) {
-                    var mirrorTexture = new MirrorTexture(parsedTexture.name, parsedTexture.renderTargetSize, scene);
+                    var mirrorTexture = new MirrorTexture(parsedTexture.name, parsedTexture.renderTargetSize, scene, generateMipMaps);
```
RaananW: The fact that this is `generateMipMaps` and with `Texture`is it `!generateMipMaps` points to the fact that we have a bug (which is totally not your fault). 

One of them is wrong. The MirrorTexture is using the Texture constructor and is not negating it. Before merging this (I have no problem with the code) I want to see which one of the classes are "right". Is it "noMipMap" or "generateMipMap" on the engine level.

RaananW: Sorry, strike that - RenderTargetTexture does negate the variable. So all good in that sense. I just find it odd.

MackeyK24: Before i submitted PR, I checked them both and Mirror/Render target use !generateMipMaps and the Texture and the Texture From String both use noMipmaps... But the new Parse function handles that :)

PR #3185, opened 2017-11-19
Fixed issue when scene properties generates warning in infinite loop

Temechon:
See http://www.html5gamedevs.com/topic/34089-debuglayer-warnings-freeze-the-scene/

### DIFF
--- a/inspector/src/helpers/Helpers.ts
+++ b/inspector/src/helpers/Helpers.ts
@@ -202,12 +202,12 @@ module INSPECTOR {
          * Returns an array of string corresponding to tjhe list of properties of the object to be displayed
          * @param obj 
          */
-        public static GetAllLinesPropertiesAsString(obj: any): Array<string> {
+        public static GetAllLinesPropertiesAsString(obj: any, dontTakeThis: Array<string>): Array<string> {
             let props: Array<string> = [];
 
             for (let prop in obj) {
                 //No private and no function
-                if (prop.substring(0, 1) !== '_' && typeof obj[prop] !== 'function') {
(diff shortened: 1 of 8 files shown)


Temechon:
Added a toggle to display/hide the minimap in the Playground

inspector/src/helpers/Helpers.ts
```diff
@@ -202,12 +202,12 @@ module INSPECTOR {
          * Returns an array of string corresponding to tjhe list of properties of the object to be displayed
          * @param obj 
          */
-        public static GetAllLinesPropertiesAsString(obj: any): Array<string> {
+        public static GetAllLinesPropertiesAsString(obj: any, dontTakeThis: Array<string>): Array<string> {
```
RaananW: love the naming :)

inspector/src/helpers/Helpers.ts
```diff
@@ -202,12 +202,12 @@ module INSPECTOR {
...
          * @param obj 
          */
-        public static GetAllLinesPropertiesAsString(obj: any): Array<string> {
+        public static GetAllLinesPropertiesAsString(obj: any, dontTakeThis: Array<string>): Array<string> {
             let props: Array<string> = [];
 
             for (let prop in obj) {
                 //No private and no function
-                if (prop.substring(0, 1) !== '_' && typeof obj[prop] !== 'function') {
+                if (dontTakeThis.indexOf(prop) === -1 && prop.substring(0, 1) !== '_' && typeof obj[prop] !== 'function') {
```
RaananW: I guess this was done to prevent circular dependencies?

Temechon: Circular dependencies doesn't need to be handled, as the function is not recursive. It was created to avoid retrieving depecated properties.

Issue #3189, opened 2017-11-20
Wrong shader behaviour for sky material on mobile.


PierreLeBlond:
The sky material from the materials library behave strangely on some mobile phones.

No error is thrown, but the sky appears black :

![screenshot_skymaterial](https://user-images.githubusercontent.com/6840047/33011699-070da442-cddf-11e7-8d41-ebc3160d52b3.jpg)

instead of what is expected :

![screenshot_skymaterial_ok](https://user-images.githubusercontent.com/6840047/33012177-6fece300-cde0-11e7-9380-04b0c387bf9b.jpg)

After investigating a bit, I found out that changing line **74** of **sky.fragment.fx** from :

`return EE * max(0.0, 1.0 - exp(-((cutoffAngle - acos(zenithAngleCos))/steepness)));`

to :

`return EE * max(0.0, 1.0 - exp((-(cutoffAngle - acos(zenithAngleCos))/steepness)));`

would solve the issue.

The only difference is the position of the minus operand within the exponential.  
One can also set _steepness_ value to 1.0 to fix the bug.

I then concluded, even if it seems very unlikely to me, that there is a weird shader compilation on some devices, something like

`-(a/b) != (-a/b)`

Unfortunately I haven't been able to reproduce the bug on any playground or custom shader, so I'm now a bit confused about it.

The bug has been reproduced on a **Samsung 7** and a **Samsung Galaxy Note 4**, on Firefox as well as Chrome. Every other tested devices including Apple products does not have this issue.

deltakosh:
PingING @luaacro

PR #3191, opened 2017-11-20
Fix wrong sky material behaviour on some devices.

PierreLeBlond:
See #3189.

On some devices the sky material appears black.
Moving a single sign within the shader solve the issue, though the reason for why this fix work isn't known yet.


deltakosh:
thanks!

PR #3192, opened 2017-11-20
Allow multiple dds files for CubeMap

### LINKED ISSUE #3188: Cube Map DDS issues
Single file DDS doesn't parse correctly; for compressed texture `bbs` is always `0` so `dataOffset` doesn't advance:

https://github.com/BabylonJS/Babylon.js/blob/e3b20e41f69b3aa7f43269ae08126eb9db5d00b4/src/Tools/babylon.dds.ts#L520

Multi-file DDS ignores the `files` parameter  only loading single `rootUrl`
https://github.com/BabylonJS/Babylon.js/blob/e3b20e41f69b3aa7f43269ae08126eb9db5d00b4/src/Engine/babylon.engine.ts#L3933

benaadams:
Resolves #3188

### DIFF
--- a/src/Tools/babylon.dds.ts
+++ b/src/Tools/babylon.dds.ts
@@ -365,7 +365,7 @@
             var ext = engine.getCaps().s3tc;
 
             var header = new Int32Array(arrayBuffer, 0, headerLengthInt);
-            var fourCC: number, width: number, height: number, dataLength: number, dataOffset: number;
+            var fourCC: number, width: number, height: number, dataLength: number = 0, dataOffset: number;
             var byteArray: Uint8Array, mipmapCount: number, mip: number;
             let internalFormat = 0;
             let format = 0;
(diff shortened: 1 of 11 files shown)


benaadams:
Still has some weird artifacts I'm trying to track down

skipping

deltakosh:
OK waiting for your go here

benaadams:
Think is same issue; is because the `dataOffset` isn't advancing for each `mip` level; due to `bpp` being 0

benaadams:
K, this fixes both issues can use just root, root + extensions or files

deltakosh:
huge thanks!

src/Tools/babylon.dds.ts
```diff
@@ -517,13 +517,18 @@
                             engine._uploadCompressedDataToTexture(sampler, i, internalFormat, width, height, byteArray);
                         }
                     }
-                    dataOffset += width * height * (bpp / 8);
+                    dataOffset += dataLength;
```
deltakosh: Actually this does not work for non compressed DDS
I'll push a fix to update to:
dataOffset += bpp ? (width * height * (bpp / 8)) : dataLength;

benaadams: I never knew there was such a thing as non-compressed DDS :)

deltakosh: Pretty useful as compressed DDS is not wide supported

PR #3198, opened 2017-11-21
strictFunctionTypes:

deltakosh:


src/Cameras/babylon.arcRotateCamera.ts
```diff
@@ -634,7 +634,7 @@
                 this._targetHost = null;
                 this._target = newTarget;
                 this._targetBoundingCenter = null;
-                this.onMeshTargetChangedObservable.notifyObservers(null);
+                this.onMeshTargetChangedObservable.notifyObservers(<any>null);
```
deltakosh: This is crappy.
I will restore the Nullable on observable even if I already know it will trigger dozen of warnings

src/Materials/Textures/babylon.colorGradingTexture.ts
```diff
@@ -86,12 +86,12 @@ module BABYLON {
...
             this._texture = texture;
 
-            var callback = (text: string) => {
+            var callback = (text: string | ArrayBuffer) => {
                 var data: Nullable<Uint8Array> = null;
                 var tempData: Nullable<Float32Array> = null;
 
                 var line: string;
-                var lines = text.split('\n');
+                var lines = (<string>text).split('\n');
```
RaananW: What happens if text is an ArrayBuffer? we can test here `if(typeof text ==='string')` which will remove the ts error and will actually be safer :)

src/Tools/babylon.assetsManager.ts
```diff
@@ -28,8 +28,8 @@ module BABYLON {
             this.taskState = AssetTaskState.INIT;
         }
 
-        onSuccess: (task: this) => void;
-        onError: (task: this, message?: string, exception?: any) => void;
+        onSuccess: (task: AbstractAssetTask) => void;
```
RaananW: This is exactly what i was trying to avoid. Now the task returned from the onSuccess callback will be an abstract asset task and no the real task. This forces the user to cast the task (and we will need to change the documentation :) )

This is the reason it took me long to enable this flag. This requires a small architectural change.

src/Cameras/babylon.arcRotateCamera.ts
RaananW: I agree that this is not the best solution. WHat happens if you define the type of this Observer to `AbstractMesh | null` instead of returning the Nullable to the observers? Haven'T checked, but it might work.

src/Tools/babylon.assetsManager.ts
```diff
@@ -98,11 +82,14 @@ module BABYLON {
         }
     }
 
-    export class MeshAssetTask extends AbstractAssetTask implements IAssetTask<MeshAssetTask> {
+    export class MeshAssetTask extends AbstractAssetTask {
```
deltakosh: What about this?
I guess this fix our issues :)

deltakosh: pinging @RaananW

RaananW: Will this return MeshAssetTask  in the onSuccess callback? I will pull this soon locally and test, just wondering if you tested already.

deltakosh: I did but would like to get a second pass

RaananW: Pulling now :) I just noticed you added onSuccess and onError to all tasks, this is probably a must.

RaananW: So! Works for me. The only thing we need to consider is that some people will have the interface referenced in their code (because of the way it used to be implemented), so it is a (very small! 😄 ) backwards compatibility issue.

deltakosh: Agree. Not a big deal
I will add a note in the what's new
Lt me fix the <any> problem for observable and I'll do the final PR

src/Materials/Textures/babylon.colorGradingTexture.ts
deltakosh: fixed

deltakosh:
@RaananW : fancy doing a last quick check?

RaananW:
on it! :)

src/Materials/Textures/babylon.colorGradingTexture.ts
RaananW: Now you dont need the casting

deltakosh: Sure? I think that TS will still enforce it

deltakosh: Excellent it works
I learn something today :D

RaananW:
👍 Looks great!

PR #3219, opened 2017-11-23
Enable PP for vignetting only when required

TrevorDev:
Enable PP for vignetting only when required
Set custom imageProcessingConfiguration applyByPostProcess to true which was set to false by mistake


TrevorDev:
Ill move these to a single PR

PR #3220, opened 2017-11-23
Have initial created camera oriented like the previous camera

TrevorDev:


TrevorDev:
Ill move these to a single PR

PR #3229, opened 2017-11-24
Add Other Draw Modes

benaadams:
POINTS/LINES/LINE_LOOP/LINE_STRIP/TRIANGLE_STRIP/TRIANGLE_FAN

Just sanity checking

@deltakosh PTAL does this work?

### DIFF
--- a/src/Engine/babylon.engine.ts
+++ b/src/Engine/babylon.engine.ts
@@ -2408,41 +2408,72 @@
         }
 
         public draw(useTriangles: boolean, indexStart: number, indexCount: number, instancesCount?: number): void {
-            this.drawElementsType(useTriangles ? Engine.DrawType.TRIANGLES : Engine.DrawType.LINES, indexStart, indexCount, instancesCount);
+            this.drawElementsType(useTriangles ? Material.TriangleFillMode : Material.WireFrameFillMode, indexStart, indexCount, instancesCount);
         }
 
         public drawPointClouds(verticesStart: number, verticesCount: number, instancesCount?: number): void {
-            this.drawArraysType(Engine.DrawType.POINTS, verticesStart, verticesCount, instancesCount);
+            this.drawArraysType(Material.PointFillMode, verticesStart, verticesCount, instancesCount);
         }
 
         public drawUnIndexed(useTriangles: boolean, verticesStart: number, verticesCount: number, instancesCount?: number): void {
-            this.drawArraysType(useTriangles ? Engine.DrawType.TRIANGLES : Engine.DrawType.LINES, verticesStart, verticesCount, instancesCount);
+            this.drawArraysType(useTriangles ? Material.TriangleFillMode : Material.WireFrameFillMode, verticesStart, verticesCount, instancesCount);
         }
 
-        public drawElementsType(type: Engine.DrawType, indexStart: number, indexCount: number, instancesCount?: number): void {
+        public drawElementsType(fillMode: number, indexStart: number, indexCount: number, instancesCount?: number): void {
             // Apply states
             this.applyStates();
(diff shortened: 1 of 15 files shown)


src/Engine/babylon.engine.ts
```diff
@@ -5396,4 +5393,16 @@
             }
         }
     }
+
+    export namespace Engine {
```
deltakosh: We are not using enum as they are generating string and string comparison.
Just add statics in the Engine

src/Materials/babylon.material.ts
```diff
@@ -370,6 +402,33 @@
             this.markAsDirty(Material.MiscDirtyFlag);
         }
 
+        public static fillModeToDrawType(fillMode: number): Engine.DrawType {
```
deltakosh: Why not using the engine constants directly?

src/Mesh/babylon.mesh.ts
```diff
@@ -997,25 +997,15 @@
...
-                    }
-                    break;
-
-                default:
-                    if (this._unIndexed) {
-                        engine.drawUnIndexed(true, subMesh.verticesStart, subMesh.verticesCount, instancesCount);
-                    } else {
-                        engine.draw(true, subMesh.indexStart, subMesh.indexCount, instancesCount);
-                    }
+            var drawType: Engine.DrawType = Material.fillModeToDrawType(fillMode);
```
deltakosh: If using the engine constants we can avoid this cool

src/Engine/babylon.engine.ts
benaadams: Wanted to add strong typing rather than just `: number`; but yeah enum does add a lot of trash

deltakosh: Yeah I know :(

jbousquie:
Does this means we will have one day some `Points` public access for the final user (like we have currently for Lines) ?
maybe something like
```javascript
var cloud = BABYLON.MeshBuilder.CreatePointCloud(name, {points: Vector3[], colors?: Color4[]}, scene);
```
then the ability to update/animate them : {instance: cloud}

Would be cool for plenty of needs :-)

benaadams:
Exposes 6 additional modes

Fills for triangles (current)
* TriangleFillMode
* WireFrameFillMode
* PointFillMode

6 additional treatment for vertexes
* PointListDrawMode
* LineListDrawMode
* LineLoopDrawMode
* LineStripDrawMode
* TriangleStripDrawMode
* TriangleFanDrawMode

benaadams:
Had to do some more tweaks, which seem to work - will update PR tomorrow

[image](https://www.youtube.com/watch?v=UAvHrx-kpXU)

deltakosh:
Fantastic!!

src/Rendering/babylon.boundingBoxRenderer.ts
```diff
@@ -85,7 +85,7 @@
                     this._colorShader.bind(worldMatrix);
 
                     // Draw order
-                    engine.draw(false, 0, 24);
+                    engine.drawElementsType(Material.TriangleFillMode, 0, 24);
```
benaadams: oops

src/Mesh/babylon.linesMesh.ts
```diff
@@ -112,7 +112,7 @@
             var engine = this.getScene().getEngine();
 
             // Draw order
-            engine.draw(false, subMesh.indexStart, subMesh.indexCount);
+            engine.drawElementsType(Material.LineListDrawMode, subMesh.indexStart, subMesh.indexCount);
```
benaadams: oops

PR #3233, opened 2017-11-25
Vortex Physics Effect

bobalazek:
Hey! I'm currently working on the vortex effect, but I'm currently a little stuck. Although the solution I currently have does kinda work but requires tweaking of the parameters, to get a good result.

Any ideas to achieve better, physically based results? :)

GIF: http://recordit.co/0DpMonyiJ4

Example:
```javascript

    // Scene
    scene.enablePhysics();


    // Physics engine
    var physicsViewer = new BABYLON.Debug.PhysicsViewer();
    var physicsHelper = new BABYLON.PhysicsHelper(scene);


    // Camera
    var camera = new BABYLON.FreeCamera("camera1", new BABYLON.Vector3(0, 16, -80), scene);


    // Lightning
    light.intensity = 0.7;


    // Ground
    var ground = BABYLON.Mesh.CreateGround("ground1", 128, 128, 2, scene);
    ground.physicsImpostor = new BABYLON.PhysicsImpostor(ground, BABYLON.PhysicsImpostor.BoxImpostor, { mass: 0, restitution: 1 }, scene);


    // Boxes
    var boxSize = 2;
    var boxPadding = 4;
    var minXY = -12;
    var maxXY = 12;
    var maxZ = 8; // temporary, later set it back to 8
    var boxParams = { height: boxSize, width: boxSize, depth: boxSize };
    var boxImpostorParams = { mass: boxSize, restitution: 0, friction: 1 };
    var boxMaterial = new BABYLON.StandardMaterial("boxMaterial");
    boxMaterial.diffuseColor = new BABYLON.Color3(1, 0, 0);
    for (var x = minXY; x <= maxXY; x += boxSize + boxPadding) {
        for (var z = minXY; z <= maxXY; z += boxSize + boxPadding) {
            for (var y = boxSize / 2; y <= maxZ; y += boxSize) {
                var boxName = "box:" + x + ',' + y + ',' + z;
                var box = BABYLON.MeshBuilder.CreateBox(boxName, boxParams, scene);
                box.position = new BABYLON.Vector3(x, y, z);
                box.material = boxMaterial;
                box.physicsImpostor = new BABYLON.PhysicsImpostor(box, BABYLON.PhysicsImpostor.BoxImpostor, boxImpostorParams, scene);
                physicsViewer.showImpostor(box.physicsImpostor);
            }
        }
    }


    // Vortex
    var origin = new BABYLON.Vector3(0, 0, 0);
    var radius = 16;
    var strength = 50;
    var height = 32;

    setTimeout(function () {
        var event = physicsHelper.vortex(
            origin,
            radius,
            strength,
            height
        );
        event.enable();

        // Debug
        var eventData = event.getData();
        var cylinder = eventData.cylinder;
        addSemiTransparentMaterial(cylinder);
        cylinder.isVisible = true;
        // Debug - END
    }, 1000);


    // Helpers
    function addSemiTransparentMaterial(mesh) {
        var semiTransparentMaterial = new BABYLON.StandardMaterial("semiTransparentMaterial", scene);
        semiTransparentMaterial.alpha = 0.5;
        mesh.material = semiTransparentMaterial;
    }

    return scene;

};
```


sebavan:
Looks nice, but would prefer to add it for 3.2 as we are stabilizing everything for 3.1. So will probably get merge in a couple of weeks.

bobalazek:
Ok, I've now cleaned up the debug & removed the angularVelocity, as it wasn't used. It does work now, but may need some additional parameters in the future, for it to feel more realistic.

sebavan:
Thx @RaananW  FYI about physics

RaananW:
Hey,

As always, really awesome. Thanks a lot :) 
I am also waiting for 3.2 for the merge. 3.1 is coming pretty soon and we are working on final updates and bug fixes. we can discuss about the parameters gladly!

bobalazek:
Ok, sure thing! :)

src/Physics/babylon.physicsHelper.ts
```diff
@@ -481,20 +487,204 @@ module BABYLON {
...
+        private _strength: number;
+        private _height: number;
+        private _centripetalForceThreshold: number = 0.7; // at which distance, relative to the radius the centripetal forces should kick in
+        private _updraftMultiplier: number = 0.02;
+        private _tickCallback: any;
+        private _cylinder: Mesh;
+        private _cylinderPosition: Vector3 = Vector3.Zero(); // to keep the cylinders position, because normally the origin is in the center and not on the bottom
+        private _dataFetched: boolean = false; // check if the has been fetched the data. If not, do cleanup
+
+        constructor(physicsEngine: PhysicsEngine, scene: Scene, origin: Vector3, radius: number, strength: number, height: number) {
```
RaananW: We can get the physics engine from the scene. One less parameter in the constructor.

src/Physics/babylon.physicsHelper.ts
```diff
@@ -481,20 +487,204 @@ module BABYLON {
...
+            this._physicsEngine = physicsEngine;
+            this._scene = scene;
```
RaananW: adding "private" to the constructor parameters will do exactly that what you are doing here. Will save the next few lines of code.

bobalazek:
Sorry, wasn't a home in the last couple of days. There we go. Optimized the stuff from the review. :)

src/Physics/babylon.physicsHelper.ts
```diff
@@ -481,20 +475,194 @@ module BABYLON {
...
+        private _originTop: Vector3 = Vector3.Zero(); // the most upper part of the cylinder
+        private _centripetalForceThreshold: number = 0.7; // at which distance, relative to the radius the centripetal forces should kick in
+        private _updraftMultiplier: number = 0.02;
+        private _tickCallback: any;
+        private _cylinder: Mesh;
+        private _cylinderPosition: Vector3 = Vector3.Zero(); // to keep the cylinders position, because normally the origin is in the center and not on the bottom
+        private _dataFetched: boolean = false; // check if the has been fetched the data. If not, do cleanup
+
+        constructor(private _scene: Scene, private _origin: Vector3, private _radius: number, private _strength: number, private _height: number) {
+            this._physicsEngine = <PhysicsEngine>this._scene.getPhysicsEngine();
```
RaananW: store the physics engine in temporary variable and check if it exists. if it doesn't, make sure to let the user know that physics needs to be enabled. If it does, set the _physicsEngine variable. this way you don't need the casting, and the user is a bit safer.

src/Physics/babylon.physicsHelper.ts
```diff
@@ -481,20 +475,194 @@ module BABYLON {
...
+        private getImpostorForceAndContactPoint(impostor: PhysicsImpostor): Nullable<PhysicsForceAndContactPoint> {
+            if (impostor.mass === 0) {
+                return null;
+            }
+
+            if (!this._intersectsWithCylinder(impostor)) {
+                return null;
+            }
+
+            var impostorObject = <Mesh>impostor.object;
```
RaananW: the object is not necessarily a mesh. If the physics object interface is not enough, you can check the getClassName() function (might need to add it to the interface) and see if it contains `Mesh` . Particles (for example) can also be physics controlled.

src/Physics/babylon.physicsHelper.ts
bobalazek: https://github.com/bobalazek/Babylon.js/blob/ebd8ce7cc1c3fdb8eaeb94064c8e34b72e40f2df/src/Physics/babylon.physicsHelper.ts#L114

It does that already here

RaananW: oh, didn't see that :). why don't you get it there instead of in the constructor? this way it is checked when the user actually needs it.

bobalazek: I think the main reason I check it there is, that few lines below I also check if any impostors exist (https://github.com/bobalazek/Babylon.js/blob/ebd8ce7cc1c3fdb8eaeb94064c8e34b72e40f2df/src/Physics/babylon.physicsHelper.ts#L120). If none, the event doesn't get fired in the first place. And of course for consistency reasons. Although, now thinking about it, I think neither of these would even be needed, because I already check for the physicsEngine in the physicsHelper constructor (https://github.com/bobalazek/Babylon.js/blob/ebd8ce7cc1c3fdb8eaeb94064c8e34b72e40f2df/src/Physics/babylon.physicsHelper.ts#L8)

src/Physics/babylon.physicsHelper.ts
bobalazek: Don!

PR #3235, opened 2017-11-26
add ray length and update mesh selection predicate

brianzinn:
Update: Mesh predicate for scene.pickWithRay().  _meshSelectionPredicate would ideally call a user configurable predicate (like public meshSelectionPredicate).  Needs to be a way to pickWithRay beyond hidden/non-pickable meshes.

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -35,6 +35,7 @@ module BABYLON {
         public onExitingVR: () => void;
         public onControllerMeshLoaded: (controller: WebVRController) => void;
 
+        private _rayLength: number;
         private _useCustomVRButton: boolean = false;
         private _teleportationRequested: boolean = false;
         private _teleportationEnabledOnLeftController: boolean = false;
@@ -116,6 +117,10 @@ module BABYLON {
                     if (webVROptions.customVRButton) {
                         this._btnVR = webVROptions.customVRButton;
                     }
+                }
+
+                if (webVROptions.rayLength) {
+                    this._rayLength = webVROptions.rayLength
                 }
             }
 
@@ -414,7 +419,7 @@ module BABYLON {
                         && mesh.name.indexOf("teleportationCircle") === -1
                         && mesh.name.indexOf("torusTeleportation") === -1
                         && mesh.name.indexOf("laserPointer") === -1)) {
-                    return true;
+                    return (mesh.isEnabled() && mesh.isVisible && mesh.isPickable);
                 }
                 return false;
             }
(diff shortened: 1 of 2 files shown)


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -414,7 +419,7 @@ module BABYLON {
                         && mesh.name.indexOf("teleportationCircle") === -1
                         && mesh.name.indexOf("torusTeleportation") === -1
                         && mesh.name.indexOf("laserPointer") === -1)) {
-                    return true;
+                    return (mesh.isEnabled() && mesh.isVisible && mesh.isPickable);
```
RaananW: is this predicate configurable somehow?

brianzinn: No, see commit message - basically I think here instead of "return true" we should be calling a configurable predicate - possibly a new one, but was thinking the existing one.  The intention of two predicates appears to be to ignore specific experience helper meshes.
```javascript
public meshSelectionPredicate
```
Calling the existing predicate would affect _castRayAndSelectObject function.

The current predicate is being used *after* the pickWithRay, so return true is not workable as we cannot pick beyond an unpickable mesh.

I think another predicate change to be made as well in case floorMeshName is substring of another mesh:
```javascript
mesh.name === this._floorMeshName
```

sebavan: Like the idea of a cutom predicate could you make it part of the option to allow override the default behaavior ?

brianzinn:
@sebavan - I have added the custom predicate in a way that matches existing behavior, so just returning true.  My code works then by adding this in my project:
```javascript
let experienceHelper = new BABYLON.VRExperienceHelper(scene);
experienceHelper.raySelectionPredicate = (mesh) => {
    return mesh.isEnabled() && mesh.isVisible && mesh.isPickable;
}
```
Following the previous functionality overriding the predicate was added as public property.  You mention as an option, but it is not done as a WebVROption, although it could be easily.

Also, I think it would be more consistent and lead people down the path to success if the defaults were the same as current ray selection without a user defined predicate as I have overrode above and on Scene (https://github.com/BabylonJS/Babylon.js/blob/master/src/babylon.scene.ts#L4062).

Using the controllers in Windows Mixed Reality everything is working perfectly.  Using gaze, however, I think is still a work in progress as there is no *selection* and gaze meshes are not tracking movement without a selection.

sebavan:
Agree, I let @davrous and @TrevorDev check on Monday as I know they are both working on the VR Helper currently, and about option vs field, it is fine for me as you did.

Issue #2498, opened 2017-07-11
VR helper


deltakosh:
_Edit_
Here are some additional updates:
- [x] Remove VRHelper from scene and just return the new helper when createDefaultVRExperience is called
- [x] Expose observables instead of callbacks (onEnteringVR, etc.)
- [x] Do not use constant (like initial position)
- [x] Expose created cameras
- [x] Let the user configure the HTML button
- [x] Add support for teleportation (can be deactivated)
- [x] Add support for actions integration (simulatePointerDown)
- [x] Add support for gaze or controllers picking
- [x] Let the user define which meshes are "teleportables"
- [x] Make sure we support controllers with fallback to gamepad
- [ ] Add support for TrackPad and second pad from WMR controllers
- [x] Fix WMR models (which are upside down)
- [x] Add documentation
- [x] Enable PP for vignetting only when required
- [x] Have initial created camera oriented like the previous camera
- [x] Make sure that we teleport where the target was defined
- [x] Make sure that the gaze cursor is correctly in front of the user

noalak:
Additional update: User feedback on VR icon ?

I've read current VRExpHelper code and david rousset article about WebVR (I'm a hero now!).
Waiting for david sources for review and testing.

deltakosh:
Will close this one and create a new one for trackpad

PR #3238, opened 2017-11-27
Web vr helper fixes

TrevorDev:
-	From #2498:  
o	Enable PP for vignetting only when required
o	Make sure that we teleport where the target was defined (please check with David to repro the issue)
o	Make sure that the gaze cursor is correctly in front of the user (please check with David to repro the issue)
o	Have initial created camera oriented like the previous camera (At least the deviceOrientationCamera)


davrous:
Excellent work @TrevorDev ! Thank you :)

PR #3243, opened 2017-11-28
Fix glTF loader bugs

bghgary:
- Animation with only one key frame
- Multiple meshes pointing to same skin


bghgary:
Addresses www.html5gamedevs.com/topic/34262-bear-gltf-model-could-not-display

PR #3257, opened 2017-11-30
fix for missing glTF2 normal, tangent and vertex color attributes on glTF 2.0 mesh primitives

kcoley:


kcoley:
Changed to use the start index of the submesh to determine if the normal and tangent is the zero vector.

kcoley:
Added two glTF validation tests for skin animations (glTF 1) and generated attributes per mesh primitive (glTF 2)

PR #3262, opened 2017-11-30
Adding Viewer to the gulp process

RaananW:
The viewer will now be built using gulp and will be deployed to the cdn 
I had to update webpack and update the way the inspector is built. Also TS, webpack, gulp and other fun frameworks were not very cooperative.


RaananW:
Do not merge, I need to fix a small bug

Issue #3280, opened 2017-12-01
VRHelper usage issues


TrevorDev:
Lewis pointed out a couple issues:
- [x] if you disconnect & reconnect the controllers while in VR, the teleportation no longer works.
- [x] The helper will always creates a new DeviceOrientationCamera, which is used when not in VR. It would be great if you could provide an option such that the developer could choose to use provide their own camera rather than the Experience Helper creating one. By default the DeviceOrientationCamera is created. The option can be used to avoid this behavior, Alternatively, move the onExitingVR observable to after you set the active camera, so that the developer re-apply a custom camera on VR exit?
- [ ] Add support for TrackPad and second pad from WMR controllers
- [ ] Check if rotation and teleportation are correct. @davrous mentioned it could be wrong still
- [x] Fix Vignette flickering
- [x] http://playground.babylonjs.com/#JA1ND3#24 has an issue to align teleportation circle on stairs

sebavan:
Could you please also double check:

- [ ] multi mesh configurations for teleport
- [ ] starting the helper while in VR

THX :-)

TrevorDev:
Created new issue with remaining problems
https://github.com/BabylonJS/Babylon.js/issues/3298

PR #3288, opened 2017-12-02
Lewis vr helper issues

TrevorDev:
#3280

- Fix if you disconnect & reconnect the controllers while in VR, the teleportation no longer works.
- If a camera is already created no device orientation camera will be created
- Ability to disable the vr fallback camera


sebavan:
I ll put in the few details tonight THX

PR #3289, opened 2017-12-02
WebVRMisc

sebavan:
Small Tweak of @TrevorDev PR

#3280

Fix if you disconnect & reconnect the controllers while in VR, the teleportation no longer works.
If a camera is already created no device orientation camera will be created
Ability to disable the vr fallback camera

@davrous could you test from the branch ?


TrevorDev:
I'm testing this and needed to make a minor tweek. I'll create this as a new PR as you recommended.

PR #3291, opened 2017-12-03
Update image.ts

aWeirdo:
Added image._markAsDirty(); on update to cellId
http://www.html5gamedevs.com/topic/34365-gui-imagecellid-doesnt-update-the-image/

### DIFF
--- a/gui/src/controls/image.ts
+++ b/gui/src/controls/image.ts
@@ -159,6 +159,10 @@ module BABYLON.GUI {
             return this._cellWidth;
         }
         set cellWidth(value: number) {
+            if(this._cellWidth === value){
+                return;
+            )
+            
             this._cellWidth = value;
         }
 
@@ -166,6 +170,10 @@ module BABYLON.GUI {
             return this._cellHeight;
         }
         set cellHeight(value: number) {
+            if(this._cellHeight === value){
+                return;
+            )
+            
             this._cellHeight = value;
         }
 
--- a/gui/src/controls/image.ts
+++ b/gui/src/controls/image.ts
@@ -161,7 +161,7 @@ module BABYLON.GUI {
         set cellWidth(value: number) {
             if(this._cellWidth === value){
                 return;
-            )
+            }
             
             this._cellWidth = value;
         }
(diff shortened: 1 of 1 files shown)


gui/src/controls/image.ts
```diff
@@ -174,6 +174,7 @@ module BABYLON.GUI {
         }
         set cellId(value: number) {
             this._cellId = value;
+            this._markAsDirty();
```
sebavan: Could you check if the _cellId is different before (same pattern used elsewhere in the file like https://github.com/BabylonJS/Babylon.js/pull/3291/files#diff-6d9c4dadf5a981b552ad14470f429d4dR101)

Would be cool to set the same  pattern in cellWidth and cellHeight as well for consistency.

aWeirdo:
Travis failed on https://github.com/BabylonJS/Babylon.js/pull/3291/commits/cdda8fc1497d6fbc06dc6cb46e2672c49af0632f .. not sure why

gui/src/controls/image.ts
```diff
@@ -159,21 +159,34 @@ module BABYLON.GUI {
...
 
         get cellHeight(): number {
             return this._cellHeight;
         }
         set cellHeight(value: number) {
+            if(this._cellHeight === value){
+                return;
+            }
+            
             this._cellHeight = value;
```
sebavan: missing the _markAsDirty here

gui/src/controls/image.ts
```diff
@@ -159,21 +159,34 @@ module BABYLON.GUI {
             return this._cellWidth;
         }
         set cellWidth(value: number) {
+            if(this._cellWidth === value){
+                return;
+            }
+            
             this._cellWidth = value;
```
sebavan: missing the _markAsDirty here

aWeirdo: Didn't add it as it's usually not something changed on-the-go, but i suppose it's possible someone has different sized images in the same texture.

aWeirdo: added.

PR #3297, opened 2017-12-04
VRHelper usage issues

TrevorDev:
#3280
- [x] if you disconnect & reconnect the controllers while in VR, the teleportation no longer works.
- [x] The helper will always creates a new DeviceOrientationCamera, which is used when not in VR. It would be great if you could provide an option such that the developer could choose to use provide their own camera rather than the Experience Helper creating one. By default the DeviceOrientationCamera is created. The option can be used to avoid this behavior, Alternatively, move the onExitingVR observable to after you set the active camera, so that the developer re-apply a custom camera on VR exit?
   - Added option to disable the creation of the DeviceOrientationCamera. If disabled and an existing camera is present on the scene no DeviceOrientationCamera will be created
Also added an option to disable creation of vrDeviceOrientationCamera. If disabled the enter vr button will only appear when an HMD is connected,
- [x] Fix Vignette flickering
- [x] http://playground.babylonjs.com/#JA1ND3#24 has an issue to align teleportation circle on stairs
- [x] multi mesh configurations for teleport
   - Validated this is working

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -13,7 +13,7 @@ module BABYLON {
         private _scene: BABYLON.Scene;
         private _position: Vector3;
         private _btnVR: HTMLButtonElement;
-        private _btnVRDisplayed: Boolean;
+        private _btnVRDisplayed: boolean;
 
         // Can the system support WebVR, even if a headset isn't plugged in?
         private _webVRsupported = false;
@@ -39,8 +39,8 @@ module BABYLON {
         private _onVRRequestPresentStart: () => void;
         private _onVRRequestPresentComplete: (success: boolean) => void;
 
-        public onEnteringVR = new Observable();
-        public onExitingVR = new Observable();
+        public onEnteringVR = new Observable<VRExperienceHelper>();
+        public onExitingVR = new Observable<VRExperienceHelper>();
         public onControllerMeshLoaded = new Observable<WebVRController>();
 
         private _rayLength: number;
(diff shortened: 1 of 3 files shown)


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -4,10 +4,16 @@ module BABYLON {
...
+    export interface VRExperienceHelperOptions extends WebVROptions {
+        createDeviceOrientationCamera?: boolean; // Create a DeviceOrientationCamera to be used as your out of vr camera.
+        createFallbackVRDeviceOrientationFreeCamera?: boolean; // Create a VRDeviceOrientationFreeCamera to be used for VR when no external HMD is found
+    }
+
     export class VRExperienceHelper {
         private _scene: BABYLON.Scene;
         private _position: Vector3;
         private _btnVR: HTMLButtonElement;
+        private _btnVRDisplayed: Boolean;
```
sebavan: Small B

src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -23,18 +29,19 @@ module BABYLON {
...
         private _onKeyDown: (event: KeyboardEvent) => void;
         private _onVrDisplayPresentChange: any;
-        private _onVRDisplayChanged: (eventArgs:IDisplayChangedEventArgs) => void;
+        private _onVRDisplayChanged: (eventArgs: IDisplayChangedEventArgs) => void;
         private _onVRRequestPresentStart: () => void;
         private _onVRRequestPresentComplete: (success: boolean) => void;
-        
-        public onEnteringVR = new Observable(); 
+
+        public onEnteringVR = new Observable();
```
sebavan: generic of WebVR and trigger with this

PR #3303, opened 2017-12-05
fixed #3295

### LINKED ISSUE #3295: [Awkwardness] setParent throws an error when used with a camera, parent do not
From a total noob perspective the following is awkward (not intuitive):
BABYLON.Mesh.setParent function throws an error when parameter is a BABYLON.Camera
Error is "t.computeWorldMatrix is not a function"
Setting 'parent' attribute directly is still legal though
[Playground](http://www.babylonjs-playground.com/#WEZYAQ)

abow:


abow:
I now think I should just be calling getViewMatrix in camera.computeWorldMatrix()

deltakosh:
I think so :)

PR #3327, opened 2017-12-07
Add context to warnings and loosen some checks in glTF loader

bghgary:
- Allow byteStride === 0 and show a warning instead
- Cache default texture sampler

### DIFF
--- a/loaders/src/glTF/2.0/babylon.glTFLoader.ts
+++ b/loaders/src/glTF/2.0/babylon.glTFLoader.ts
@@ -52,6 +52,7 @@ module BABYLON.GLTF2 {
         private _parent: GLTFFileLoader;
         private _rootUrl: string;
         private _defaultMaterial: PBRMaterial;
+        private _defaultSampler = {} as IGLTFSampler;
         private _rootNode: IGLTFNode;
         private _successCallback: () => void;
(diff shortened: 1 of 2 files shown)


loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -227,7 +229,7 @@ module BABYLON.GLTF2 {
                 const buffers = this._gltf.buffers;
                 if (buffers && buffers[0] && !buffers[0].uri) {
                     const binaryBuffer = buffers[0];
-                    if (binaryBuffer.byteLength != data.bin.byteLength) {
+                    if (binaryBuffer.byteLength < data.bin.byteLength - 3 || binaryBuffer.byteLength > data.bin.byteLength) {
```
deltakosh: just for my understanding: why -3 ?

bghgary: Bin chunk must be 4-byte aligned but the buffer length does not. See https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#glb-stored-buffer and https://github.com/KhronosGroup/glTF/issues/1026.

deltakosh: ok thanks a lot!

PR #3341, opened 2017-12-07
Update loaders readme to point to usage documentation

bghgary:
@deltakosh I removed the OBJ and glTF specific readme.md and just rely on the one in loaders.


deltakosh:
Much appreciated

PR #3349, opened 2017-12-08
glTF loader exception handling

bghgary:
This change makes it such that you can now do the following:

```javascript
var onError = function (scene, message, exception) {
    if (exception instanceof BABYLON.LoadFileError) {
        if (exception.request) {
            alert(exception.name + ": " + exception.request.status);
        }
    }
};
BABYLON.SceneLoader.Append(rootUrl, sceneFilename, scene, onSuccess, onProgress, onError);
```

I have verified the `Tools.LoadFile` change in IE11, Edge, Chrome, and FireFox on Windows 10.

### DIFF
--- a/src/Tools/babylon.tools.ts
+++ b/src/Tools/babylon.tools.ts
@@ -3,6 +3,21 @@
         animations: Array<Animation>;
     }
 
+    // See https://stackoverflow.com/questions/12915412/how-do-i-extend-a-host-object-e-g-error-in-typescript
+    // and https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    export class LoadFileError extends Error {
+        // Polyfill for Object.setPrototypeOf if necessary.
+        private static _setPrototypeOf: (o: any, proto: object | null) => any =
+            (Object as any).setPrototypeOf || ((o, proto) => { o.__proto__ = proto; return o; });
+
+        constructor(message: string, public request?: XMLHttpRequest) {
+            super(message);
+            this.name = "LoadFileError";
+
+            LoadFileError._setPrototypeOf(this, LoadFileError.prototype);
+        }
+    }
+
     // Screenshots
     var screenshotCanvas: HTMLCanvasElement;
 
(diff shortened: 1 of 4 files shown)


bghgary:
@deltakosh @sebavan Perhaps should test on Safari before merging?

deltakosh:
Do you want it in 3.1?

bghgary:
Perhaps not. @sebavan Do we need this in 3.1?

sebavan:
we ll merge it just after to prevent issues in 3.1. this can wait 2-3 more days. thx Gary

src/Tools/babylon.tools.ts
```diff
@@ -3,6 +3,16 @@
         animations: Array<Animation>;
     }
 
+    // See https://stackoverflow.com/questions/12915412/how-do-i-extend-a-host-object-e-g-error-in-typescript
```
sebavan: I would prefer not inheriting from error and wrapping the original error for callstack. 

Also do not rely on __proto__ on your on you can use setPrototypeOf in Typescript

bghgary: > I would prefer not inheriting from error and wrapping the original error for callstack.

I tried that, but it has issues. Exceptions that are throw will no longer look right in the browser. It will just say the name of the class.

> Also do not rely on proto on your on you can use setPrototypeOf in Typescript

Object.setPrototypeOf is not available in the lib.es5.d.ts. Supposedly, it also does not work in IE11. Perhaps I can try to call it if it exists?

sebavan: Yup, so should we remove the inheritance ? As setting proto might be pretty slow and lead to deoptim.

bghgary: Just tested it and Object.setPrototypeOf does work in IE11.

> should we remove the inheritance?

My preference is to use it since it works much better with the debugger. Someone might be checking to see if the exception is an instance of Error, in which case not inheriting from Error would be broken.

> As setting proto might be pretty slow and lead to deoptim.

I'm not sure this is a big issue. This type of exceptions should be rare and far in between.

bghgary: I've updated the code to this locally:

```javascript
            if ((Object as any).setPrototypeOf) {
                (Object as any).setPrototypeOf(this, LoadFileError.prototype);
            }
            else {
                (this as any).__proto__ = LoadFileError.prototype;
            }
```

PR #3350, opened 2017-12-08
do not crash on dispose

TrevorDev:


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -309,8 +304,8 @@ module BABYLON {
 
             // Gamepad connection events
             this._webVRCamera.onControllerMeshLoadedObservable.add((webVRController) => this._onDefaultMeshLoaded(webVRController));
-            this._scene.gamepadManager.onGamepadConnectedObservable.add((pad) => this._onNewGamepadConnected(pad));
-            this._scene.gamepadManager.onGamepadDisconnectedObservable.add((pad) => this._onNewGamepadDisconnected(pad));
+            this._scene.gamepadManager.onGamepadConnectedObservable.add(this._onNewGamepadConnected);
```
deltakosh: you are losing the this context here no?

deltakosh: oh wait it should work as I've just seen the change on onNewGamepadConnected to arrow function

deltakosh: did you tested it with controllers and gamepads and so on?

TrevorDev: Yes, tested disposing with and without controllers in and made sure sponza and mansion still work as expected.

Issue #3352, opened 2017-12-09
Vive not working with VR helper


TrevorDev:
Suspected helper bugs:
- [x] Cannot see ray from controller
- [x] Not handling if webvrController.hand == ""
- [x] Teleporting backwards is crooked at certain rotations
- [x] Teleporting backwards sets y position too high
- [x] Pad values on vive are backwards compared to windows mixed reality controller 
- [x] Not using standing matrix for height
- [x] Teleporting does not work (This works in older version in firefox.)
- [x] Facing backwards at start (this also works as expected in older version of firefox)
- [x] Dispose vr helper when scene is disposed

Suspected bug in browser:
- [ ] Some standard materials change color when gaze cursor is selecting something.[Originally thought to be a lighting issue] (This does not occur in older firefox, this might be the webgl2 issue but this still happens in firefox)

deltakosh:
Woot
@ranaanw please test on oculus ASAP:)

RaananW:
will do

deltakosh:
Ok only lighting remains

PR #3354, opened 2017-12-09
Fix a couple vive bugs

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/3352

Validated using firefox: 57.0a1 (2017-08-29) (64-bit)


TrevorDev:
Oops, I'll fix that. I should actually test this on the wmr headset on monday before this gets merged. Ill reopen this then.

sebavan:
Np thanks again for the fix :-)

PR #3355, opened 2017-12-09
Update index.js

aWeirdo:
Playground; Wrong label on indexstable's version changer.

### DIFF
--- a/Playground/indexStable.html
+++ b/Playground/indexStable.html
@@ -110,7 +110,7 @@
         </div>
 
         <div class="category right">
-            <div class="button select"><span id="currentVersion1600">Version: Stable/span>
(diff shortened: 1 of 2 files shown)


aWeirdo:
.. my bad 😄

aWeirdo:
Okay, i had a few minutes, and it is indeed supposed to be updated at run-time.
// SetVersion
https://github.com/BabylonJS/Babylon.js/blob/master/Playground/js/index.js#L638

// setToMultipleID 
https://github.com/BabylonJS/Babylon.js/blob/master/Playground/js/index.js#L5

// setToMultipleID Called at
https://github.com/BabylonJS/Babylon.js/blob/master/Playground/js/index.js#L90

but it isn't happening, it always says "Version: Latest".

Issue #3277, opened 2017-12-01
Ability to outline groups of objects


PixelsCommander:
Ability to outline/stroke on groups of objects. Like HighlightLayer but not blurred. API wise it is possible to deliver as additional parameter to HighlightLayer `threshold` which will `smoothstep` alpha on blur texture. I will introduce a PR on this next week.

sebavan:
cool thx :-)

3dos:
Hi there, I tried it with no success. Even the playground https://playground.babylonjs.com/#K7UNC7 shows the blurry outline (tried it on chrome and firefox on OSX).

PixelsCommander:
Option property was changed to isStroke: true, instead of threshold: .025

2018-04-30 16:28 GMT+02:00 Filipe Dos Santos Mendes <
notifications@github.com>:

3dos:
Oh great thank you. Works better now :D

PR #3358, opened 2017-12-11
#3277 Implements threshold for HighlightLayer introducing threshold p…

PixelsCommander:
#3277 Implements threshold for HighlightLayer introducing threshold property on HighlightLayerOptions. Code for demoing this - https://playground.babylonjs.com/#K7UNC7


PixelsCommander:
screen shot 2017-12-11 at 2 01 21 pm

src/Layer/babylon.highlightlayer.ts
```diff
@@ -1,4 +1,19 @@
 ﻿module BABYLON {
+    BABYLON.Effect.ShadersStore["highlightLayerThresholdFragmentShader"] = "\r\n" +
```
deltakosh: shaders have to be stored as file in /src/shaders

src/Layer/babylon.highlightlayer.ts
```diff
@@ -400,14 +421,26 @@
                 });
             }
 
+            const postProcesses = [this._downSamplePostprocess, this._horizontalBlurPostprocess, this._verticalBlurPostprocess];
+
+            if (this._options.threshold) {
+                const threshold = this._options.threshold;
+                this._thresholdPostProcess = new BABYLON.PostProcess("thresold", "highlightLayerThreshold", ["screenSize", "threshold"], null, 0.25, null, BABYLON.Texture.BILINEAR_SAMPLINGMODE, this._scene.getEngine());
```
deltakosh: must be disposed

PixelsCommander: Fixed

PixelsCommander:
@deltakosh all your comments were fixed

deltakosh:
Perfect!! Will wait for 3.1 release now :)

src/Layer/babylon.highlightlayer.ts
```diff
@@ -56,6 +56,11 @@
          * The camera attached to the layer.
          */
         camera: Nullable<Camera>;
+
+        /**
+         * Threshold for cutting glow.
+         */
+        threshold: number;
```
deltakosh: should be threshold?: number

src/Layer/babylon.highlightlayer.ts
```diff
@@ -400,14 +406,26 @@
                 });
             }
 
+            const postProcesses = [this._downSamplePostprocess, this._horizontalBlurPostprocess, this._verticalBlurPostprocess];
+
+            if (this._options.threshold) {
+                const threshold = this._options.threshold;
+                this._thresholdPostProcess = new BABYLON.PostProcess("thresold", "highlightLayerThreshold", ["screenSize", "threshold"], null, 0.25, null, BABYLON.Texture.BILINEAR_SAMPLINGMODE, this._scene.getEngine());
```
deltakosh: PP name should be threshold ;)

deltakosh:
Up?

PixelsCommander:
@deltakosh all s fixed

src/Layer/babylon.highlightlayer.ts
```diff
@@ -879,6 +897,7 @@
             this._downSamplePostprocess.dispose();
             this._horizontalBlurPostprocess.dispose();
             this._verticalBlurPostprocess.dispose();
+            this._thresholdPostProcess.dispose();
```
deltakosh: Need a if here as the threshold Pp can be undefined

PixelsCommander: Done

deltakosh:
LGTM.
Waiting for @sebavan review

sebavan:
Could the threshold post process logic be integrated in glowMapMerge under a define instead ?

This would save a full screen pass improving the perfromances.

PixelsCommander:
@sebavan elaborate on "under define" approach please.

deltakosh:
Not sure we can do it actually as the HL uses only default PP (like blur).

sebavan:
Actually the shader https://github.com/BabylonJS/Babylon.js/blob/master/src/Shaders/glowMapMerge.fragment.fx#L10
is used to blit the blurred highlight information on top of the rendered scene.

So the code:

```
#ifdef THRESHOLD
float alpha = smoothstep(.0, threshold, baseColor.a);
baseColor.a = alpha;
baseColor.rgb= baseColor.rgb * alpha;
#endif
```

should be able to be inserted in the glowMapMerge shader to save one full pass.

The #ifdef THRESHOLD is the part I called Defines cause in Babylon THRESHOLD would need to be added to the defines list.

deltakosh:
Ok makes sense..This is a great idea

PixelsCommander:
@sebavan @deltakosh  And how THRESHOLD should be set from JS ?

deltakosh:
You can add a new uniform here:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Layer/babylon.highlightlayer.ts#L10

deltakosh:
up?

deltakosh:
Hello @PixelsCommander , do you have time to finish this PR?

PixelsCommander:
Currently working on another issue. Will get back to stroke/threshold once done with it.

deltakosh:
Sounds good

PixelsCommander:
@deltakosh , @sebavan elaborate please on how should I define THRESHOLD to use with #ifdef

deltakosh:
You also need to add threshold in the list of uniform.
Would you mind updating the PR to see current state?

PixelsCommander:
I do not use uniforms anymore. No reason for stroke to be tunable, I found the number of 0.5 working pretty smooth then you control width of the stroke in the same way used for highlightLayer.

So the question is still the same - any hints on why defines are lost in Effect constructor?

sebavan:
@PixelsCommander,  you added the stroke define on the wrong post process (glowBlurPostProcess instead of glowMapMergeEffect https://github.com/BabylonJS/Babylon.js/blob/master/src/Layer/babylon.highlightlayer.ts#L289)

Moving should solve your issue.

PixelsCommander:
@deltakosh @sebavan check it now, the code for demoing https://playground.babylonjs.com/#K7UNC7

sebavan:
Thank you, works nicely.

deltakosh:
Please update the what's new as well to credit you

PixelsCommander:
@deltakosh would this be a valid PR for this https://github.com/BabylonJS/Babylon.js/pull/3605 ?

PR #3368, opened 2017-12-12
Update virtualKeyboard.ts

aWeirdo:
Added shift key & shiftState
http://www.html5gamedevs.com/topic/34508-supporting-shift-key-for-virtual-keyboard/

Added a "shiftState" instead of an boolean.
Allowing a more mobile-like virtual keyboard, 
One click = Only next letter is upperCase. // Light blue key.
Two clicks = Caps-lock on. // Light blue key & border.
Three clicks = Caps-lock off. // default color & no border.

https://github.com/BabylonJS/Documentation/pull/719

### DIFF
--- a/gui/src/controls/virtualKeyboard.ts
+++ b/gui/src/controls/virtualKeyboard.ts
@@ -27,6 +27,8 @@ module BABYLON.GUI {
         public defaultButtonColor = "#DDD";
         public defaultButtonBackground = "#070707";    
         
+        public shiftState = 0;
+        
         protected _getTypeName(): string {
             return "VirtualKeyboard";
         }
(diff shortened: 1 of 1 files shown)


aWeirdo:
https://github.com/BabylonJS/Documentation/pull/719

gui/src/controls/virtualKeyboard.ts
```diff
@@ -112,6 +141,11 @@ module BABYLON.GUI {
                 }
 
                 this._connectedInputText.processKey(-1, key);
+                
+                if(this.shiftState === 1){
```
RaananW: Just wondering - shouldn't this be `> 1`? or is this a case that will never happen?

aWeirdo: Hi @RaananW 
That is only supposed to run when shiftState is 1, as in that case, only 1 letter will be capitalized and the shiftState reset to 0. 
When the shift key is clicked twice, shiftState = 2, all letters until shift is clicked again, are capitalized.

gui/src/controls/virtualKeyboard.ts
```diff
@@ -75,6 +77,26 @@ module BABYLON.GUI {
...
+                let row = this.children[i];
+                if(!row) continue;
+
+                for(var j = 0; j < row.children.length; j++){
+                    let button = row.children[j];
+                    if(!button || !button.children[0]) continue;
+                    let button_tblock = button.children[0];
+
+                    if(button_tblock.text === "\u21E7"){
+                        button.color = (shiftState ? "#7799FF" : this.defaultButtonColor);
```
deltakosh: The color must be a value that user can set

deltakosh: I'll fix it

PR #3370, opened 2017-12-12
Exposing teleportationTarget mesh

davrous:
You can now use your own mesh for the teleportation target via the... teleportationTarget property.

Demo: http://playground.babylonjs.com/#JA1ND3#28

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -61,7 +61,8 @@ module BABYLON {
         private teleportBackwardsVector = new Vector3(0, -1, -1);
         private _rotationRightAsked = false;
         private _rotationLeftAsked = false;
-        private _teleportationCircle: Mesh;
+        private _teleportationTarget: Mesh;
+        private _isDefaultTeleportationTarget = true;
         private _postProcessMove: ImageProcessingPostProcess;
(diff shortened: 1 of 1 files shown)


CarlosLanderas:
@davrous thank you very much. Just what I needed for my project. It would be amazing to have the possibility of configuring the motion controllers meshes (WebVRController.attachMesh does not seem to work anymore when working with VRHelper or I am not able to replace the meshes like some months ago)

Another thing that I have been looking for is being able to hide/display the controller meshes, while lasers can be kept.

davrous:
@CarlosLanderas this still works for me. Here's a sample: http://playground.babylonjs.com/#EQ55T5#5 

Can you confirm it works on your side? Which HMD are you using?

Good idea for the lasers, I'll think about this for 3.2 as we're closing feature for v3.1

CarlosLanderas:
Thank you very much. I have a first action scene with player arms, but I dont want them to follow the motion controllers, just being able to display the lasers so the arms can be animated with controllers button interaction.


I am using Lenovo with WMC, and your babylon playground sample is working perfect.

I was using this one 

https://www.babylonjs-playground.com/#5MV04

Thas is linked in the WebVRCamera documentation:

https://doc.babylonjs.com/how_to/webvr_camera#attaching-to-a-mesh

Thank you again @davrous

PR #3382, opened 2017-12-13
Add default value to Nullable<T> types

brianzinn:
Add default value to Nullable<T> types, which match the expected discriminated union (T | null) from TypeScript definition.  This is not backwards compatible. There are many more cases of undefined (T | null); this builds upon the strict type checking recently added.

See discussion: http://www.html5gamedevs.com/topic/34533-csg-tomesh-material-required/?tab=comments#comment-198439


deltakosh:
Thanks

PR #3384, opened 2017-12-13
HighLightLayer removeMesh bug.  _meshes not delete.

wyklion:


src/Layer/babylon.highlightlayer.ts
```diff
@@ -826,7 +826,7 @@
                 }
             }
 
-            this._meshes[mesh.uniqueId] = null;
+            delete this._meshes[mesh.uniqueId];
```
deltakosh: just move it two lines before to be sure that this._meshes[mesh.uniqueId] exists

sebavan: This actually works cause the highlight layer code expect to have null values in the list but agree it might look neater without them.

sebavan:
This actually works cause the highlight layer code expect to have null values in the list but agree it might look neater without them.

wyklion:
sorry to make two commits. I just use github website to make pull request.

PR #3399, opened 2017-12-15
Flip front face when world matrix determinant is negative

bghgary:
This is to support mirrored geometry as per glTF spec here: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#transformations.

> Implementation Note: If the determinant of the transform is a negative value, the winding order of the mesh triangle faces should be reversed. This supports negative scales for mirroring geometry.

### DIFF
--- a/src/States/babylon.depthCullingState.ts
+++ b/src/States/babylon.depthCullingState.ts
@@ -6,14 +6,16 @@
         private _isCullFaceDirty = false;
         private _isCullDirty = false;
         private _isZOffsetDirty = false;
-
+        private _isFrontFaceDirty = false;
+        
         private _depthTest: boolean;
         private _depthMask: boolean;
         private _depthFunc: Nullable<number>;
         private _cull: Nullable<boolean>;
         private _cullFace: Nullable<number>;
         private _zOffset: number;
-
+        private _frontFace: Nullable<number>;
+        
         /**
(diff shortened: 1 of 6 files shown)


deltakosh:
@sebavan do you mind reviewing this one as it touches PBR?

src/States/babylon.depthCullingState.ts
```diff
@@ -38,19 +38,6 @@
             this._isZOffsetDirty = true;
         }
 
-        public get cullFace(): Nullable<number> {
```
deltakosh: I'm fine with this PR. My only concern is this change in the public API
We need to keep cullFace property (and probably deduce it from frontFace)

bghgary: I can put this back. Is it necessary to set it at all? I can just leave it as default.

deltakosh: it was as read/write property so wee need to keep it unaltered :D

bghgary:
I have updated the PR with the changes.

bghgary:
Updated again. In early versions, `engine.cullBackFaces` was incorrectly being ignored. This has been fixed in the current version. `cullFace` property on the `depthCullingState` object is actually required because of this.

I also included a one-line change to fix #3413.

PR #3405, opened 2017-12-15
Stop Tools.LoadFile from retrying if the request is aborted

bghgary:


bghgary:
I am still testing this change. Please wait.

bghgary:
I'm moving this change to my other PR.

PR #3409, opened 2017-12-18
Remove BABYLON. references from code

RaananW:
As a preparation for module-usage, it is important to never use the BABYLON namespace inside the babylon source code (meaning. not to reference an object using BABYLON.* but using the object's name instead).

Those changes remove all references of BABYLON in the source folder.


src/babylon.mixins.ts
```diff
@@ -34,7 +34,7 @@ interface WebGLRenderingContext {
     vertexAttribDivisor(index: number, divisor: number): void;
 
     createVertexArray(): any;
-    bindVertexArray(vao: BABYLON.Nullable<WebGLVertexArrayObject>): void;
+    bindVertexArray(vao?: WebGLVertexArrayObject | null): void;
```
Hersir88: Just curious do we really need null here as vao is anyway optional, so WebGLVertexArrayObject | undefiend;

RaananW: Good question. this was added so that I wouldn't make any changed in the engine. null is being provided as a variable in some cases, and as I didn't want to make changes that might effect the engine itself, i decided to add the null.

RaananW:
(wait before merging, more are added)

deltakosh:
Good?

RaananW:
Good

PR #3410, opened 2017-12-18
Allow setting vertical line spacing with lineSpacing property in GUI Textblock control

CarlosLanderas:
When working with multiple lines, textblock is using current line measure for vertical spacing, but lines are very close as you can see in the following image:

I added in textblock the property lineSpacing so you can apply vertical spacing between lines like this:

```javascript
 var text1 = new BABYLON.GUI.TextBlock();
 text1.textWrapping = true;
 text1.lineSpacing = "80px";
```

Or using percentage:

```javascript
 text1.lineSpacing = "20%";
```

Textblock output:


gui/src/controls/textBlock.ts
```diff
@@ -20,6 +20,10 @@ module BABYLON.GUI {
             return this._resizeToFit;
         }
 
+        public set lineSpacing(value: number) {
```
RaananW: Not sure if this is the convention in the GUI, but why is there only a setter and no getter?

CarlosLanderas: Added the getter :).

gui/src/controls/textBlock.ts
```diff
@@ -81,6 +81,14 @@ module BABYLON.GUI {
             this._markAsDirty();
         }
 
+        public set lineSpacing(value: number) {
+            this._lineSpacing = value;
+        }
+
+        public get lineSpacing(): number {
+            return this._lineSpacing;
```
deltakosh: You need to call markAsDirty to force render. But make sure the value actually changed

deltakosh: Also this has to be an UnitOrValue object to allow users to use pixel or percentage

CarlosLanderas: Cool @deltakosh, thanks for the tips :D.

deltakosh:
Great. Please also update documentation and credits yourself in the what's new:
- https://github.com/BabylonJS/Documentation/blob/master/content/How_To/GUI/Gui.md
- https://github.com/BabylonJS/Babylon.js/blob/master/dist/preview%20release/what's%20new.md

CarlosLanderas:
Will do ASAP. Thanks.

PR #3419, opened 2017-12-19
Expose VRDisplay ready observable on VR Experience Helper

CarlosLanderas:
Right now VRExperienceHelper allow user to pass an HTMLElement button, or it creates the default one. Once the VRDisplay is ready it updates de button visibility so it can be shown and clicked.

I have a user interface menu, where user can select game mode (normal / VR), I need to handle code and want to initialize VR on the fly. 

If I click the button after creating the VRHelper, the VRDisplay is not ready and it fallbacks to deviceOrientationCamera or freeCamera depending on configuration.

Right now I was using a workaround with:

```javascript
scene.getEngine().onVRDisplayChangedObservable.add((e) => {
       if (e.vrDisplay) {
         vrCustomButton.click();
      }
```

But I would find interesting exposing the moment whenever the device is ready on the VR Helper:

```javascript

//custom button that will be dynamically triggered
    let vrCustomButton = createCustomVRButton();

    var vrHelper = new BABYLON.VRExperienceHelper(scene, { useCustomVRButton: true, customVRButton: vrCustomButton });
    vrHelper.onVRDeviceReadyObservable.add(e => {
        //Device is ready, and I come from a selection mode menu click, I can enter now in VRMode
        vrCustomButton.click();
    });

````

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -44,6 +44,10 @@ module BABYLON {
          */
         public onEnteringVRObservable = new Observable<VRExperienceHelper>();
 
+         /**
+         * Observable raised when VR Display is ready.
+         */
+        public onVRDeviceReadyObservable = new Observable<VRExperienceHelper>();
 
         /**
          * Observable raised when exiting VR.
@@ -435,6 +439,8 @@ module BABYLON {
             this._webVRsupported = eventArgs.vrSupported;
             this._webVRready = !!eventArgs.vrDisplay;
             this._webVRpresenting = eventArgs.vrDisplay && eventArgs.vrDisplay.isPresenting;
+
+            this._webVRready && this.onVRDeviceReadyObservable.notifyObservers(this);
 
             this.updateButtonVisibility();
         }

### CODE REVIEW

File: src/Cameras/VR/babylon.vrExperienceHelper.ts [hunk 1]
```diff
@@ -44,6 +44,10 @@ module BABYLON {
          */
         public onEnteringVRObservable = new Observable<VRExperienceHelper>();
 
+         /**
+         * Observable raised when VR Display is ready.
+         */
+        public onVRDeviceReadyObservable = new Observable<VRExperienceHelper>();
```
RaananW: maybe onVRDisplayChangedObservable? Ready might be a bit misleading here.

CarlosLanderas: Sure @RaananW , on it.

CarlosLanderas: @RaananW , VRExperienceHelper uses this._webVRready to consider it ready to go for VR. If I rename it to onVRDisplayChangedObservable I should not need to check _webVRReady is true I guess...?

CarlosLanderas: Maybe changing the location of the observable notification?

 if (webVROptions.createFallbackVRDeviceOrientationFreeCamera) {
                this.displayVRButton();
            } else {
                this._scene.getEngine().onVRDisplayChangedObservable.add((e) => {
                    
                    this.onVRDisplayChangedObservable.notifyObservers(e);

                    if (e.vrDisplay) {
                        this.displayVRButton();
                    }
                })
            }

RaananW: You mean in the constructor? looks fine.

As I see, it seems like prepending a function to the engine.onVRDisplayChangedObservable will be enough in your case won't it?

PR #3424, opened 2017-12-19
Fix progress event in loader, Tools.LoadFile abort request, and bounding box

bghgary:


loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -80,20 +80,38 @@ module BABYLON.GLTF2 {
         }
 
         // IE 11 Compatibility.
-        private static _createProgressEvent: (lengthComputable: boolean, loaded: number, total: number) => ProgressEvent =
-            (typeof (<any>window)["ProgressEvent"] === "function")
-                ? (lengthComputable, loaded, total) => {
+        private static _createProgressEvent = GLTFLoader._getCreateProgressEventFunction();
+        private static _getCreateProgressEventFunction(): (lengthComputable: boolean, loaded: number, total: number) => ProgressEvent {
+            // Use ProgressEvent if available.
+            if (typeof (<any>window)["ProgressEvent"] === "function") {
```
deltakosh: beware as this won't work in nullengine mode as window is not defined

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -80,20 +80,38 @@ module BABYLON.GLTF2 {
...
                         loaded: loaded,
                         total: total
                     });
-                }
-                : (lengthComputable, loaded, total) => {
+                };
+            }
+
+            // Fallback to initProgressEvent if available.
+            const event = document.createEvent("ProgressEvent");
```
deltakosh: we want to make sure it works event if document does not exist

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: is there a better way to check the existence of ProgressEvent class?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: what do you recommend if document doesn't exist?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
deltakosh: you can first call Tools.IsWindowObjectExist()
as a side node: why do we need to rely on a DOM element and not on our own object?

loaders/src/glTF/2.0/babylon.glTFLoader.ts
deltakosh: we should not rely on DOM element :(

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: this is supposed to match what XHR does, but maybe it's okay to change to our own object? the ISceneLoaderPluginAsync interface is currently defined to return a ProgressEvent for onProgress.

deltakosh: We can change that easily. My concern is that we now have NulLEngine in placed and used by several customers. This will break the nullengine scenario

bghgary: Okay, I'll make this change and update.

deltakosh: much appreciated:)

bghgary:
@RaananW FYI https://github.com/BabylonJS/Babylon.js/pull/3424/commits/58297ad80747a8c8d73e31480a0e712192bfcba8

`computeWorldMatrix` must happen even if there are no vertices.

deltakosh:
Thanks!

PR #3434, opened 2017-12-21
Making sure updateCache only calls update once per loop.

RaananW:
This will prevent a callstack overflow when using keyboards in webvr.

### DIFF
--- a/src/Cameras/VR/babylon.webVRCamera.ts
+++ b/src/Cameras/VR/babylon.webVRCamera.ts
@@ -325,40 +325,48 @@ module BABYLON {
         private _workingVector = Vector3.Zero();
         private _oneVector = Vector3.One();
         private _workingMatrix = Matrix.Identity();
+
+        private updateCacheCalled: boolean;
+
         public _updateCache(ignoreParentClass?: boolean): void {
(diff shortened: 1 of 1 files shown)


src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -369,6 +377,7 @@ module BABYLON {
             Quaternion.FromRotationMatrixToRef(this._workingMatrix, this.deviceRotationQuaternion);
 
             super.update();
+            this.updateCalled = false;
```
deltakosh: not sure to understand what this boolean is ?

RaananW: to make sure it is called only once. Prevents the infinite loop. It is set to false at the end of this call. Maybe the naming is not the best, but it does the job :)

deltakosh: But what is the difference with updateCacheCalled?
I see no declaration for updateCalled

RaananW: Ehhh...
This is obviously not needed. Sorry! I will remove this.

PR #3435, opened 2017-12-21
Add MSAA Support for MultiRenderTargets

PierreLeBlond:
Currently, when we enabled MSAA on MultiRenderTarget, we don't observe the expected behavior :

Each output texture share the same output, the one bind to COLOR_ATTACHMENT0.
this is similar to this [issue](https://stackoverflow.com/questions/32213369/opengl-mutlisampling-multiple-render-target-fbos#32216745).

Here's the steps I followed to fix it :

- Within a MultiRenderTarget, Each texture now share a common FBO with one RenderBuffer per texture, rather than having a different FBO.
- When unbinding the frameBuffer and performing blit operation to retrieve anti-aliased frame, each attachment is correctly copied to its corresponding render target.

- I also fixed a bug in MultiRenderTarget where it was impossible to enable unsigned byte texture type.

However, as I had to touch some core files of BJS, and as I'm not really familiar with Babylon code convention, I don't know if this feature is relevant right now and is worth the refactoring.

I'd like to hear your thoughts about this !


src/Engine/babylon.engine.ts
```diff
@@ -1906,6 +1906,62 @@
...
+
+            // If MSAA, we need to bitblt back to main texture
+            var gl = this._gl;
+
+
+            if (textures[0]._MSAAFramebuffer) {
+                gl.bindFramebuffer(gl.READ_FRAMEBUFFER, textures[0]._MSAAFramebuffer);
+                gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, textures[0]._framebuffer);
+
+                var attachments = new Array(textures.length);
```
deltakosh: We cannot create an array per frame

PierreLeBlond:
I only tested it on my own project,
I will try to set up a small playground next week !

deltakosh:
Ok I will merge then :)
Just need a small repro to make sure we are good (I will also add it to our validation tests suite ;))
Thanks again!

PierreLeBlond:
I was able to successfully test it on a simple playground within my localDev, but is there a simple way I can send you a playground which uses my babylon.js branch ?

deltakosh:
Just drop the text of the pg here:)

PierreLeBlond:
:)

```

    var camera = new BABYLON.FreeCamera("camera1", new BABYLON.Vector3(0, 0, -10), scene);
    // camera.maxZ = 200;

    // This targets the camera to scene origin

    // This attaches the camera to the canvas

    // This creates a light, aiming 0,1,0 - to the sky (non-mesh)

    // Default intensity is 1. Let's dim the light a small amount
    light.intensity = 0.7;


    var planePosition = BABYLON.MeshBuilder.CreatePlane("planePosition", {width:7, height:5}, scene);
    planePosition.material = new BABYLON.StandardMaterial("position", scene);
    planePosition.position.x -= 4;

    var planeDepth = BABYLON.MeshBuilder.CreatePlane("planeDepth", {width:7, height:5}, scene);
    planeDepth.material = new BABYLON.StandardMaterial("depth", scene);
    planeDepth.position.x += 4;

    var gBuffer = new BABYLON.GeometryBufferRenderer(scene, 1.0);

    if(gBuffer.isSupported) {

        gBuffer.enablePosition = true;
        gBuffer.getGBuffer().samples = 8;

        planePosition.material.diffuseTexture = gBuffer.getGBuffer().textures[2];
        planeDepth.material.diffuseTexture = gBuffer.getGBuffer().textures[0];

        scene.customRenderTargets.push(gBuffer.getGBuffer());
    } else {
        console.log("Error : Multiple render targets are not supported !");
    }

    return scene;
};
```

deltakosh:
Thanks a lot!

deltakosh:
Please credit yourself in the what's new:

PR #3436, opened 2017-12-21
Add onSelectedMeshUnselectedObservable in VRHelper

CarlosLanderas:
Add an observable to notify about last selected mesh being unselected when  you are working with predicates.  In my case, I would need this to interact with user interfaces, Add/remove highlights on GUI controls etc.

Usage:

```Javascript
 vrHelper.meshSelectionPredicate = (mesh) => {
      return mesh === this._mesh;
 };

vrHelper.onNewMeshSelected.add( c => console.log(` Entering ${c.name}`));
vrHelper.onSelectedMeshUnselected .add( c => console.log(` Exiting ${c.name}`));

```

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -113,6 +113,11 @@ module BABYLON {
         public onNewMeshSelected = new Observable<AbstractMesh>();
         private _circleEase: CircleEase;
 
+         /**
+         * Observable raised when current selected mesh gets unselected
+         */
+        public onSelectedMeshUnselected = new Observable<AbstractMesh>();
+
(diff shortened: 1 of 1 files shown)


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -1395,6 +1400,7 @@ module BABYLON {
                         }
                     }
                     else {
+                        this._currentMeshSelected && this.onSelectedMeshUnselected.notifyObservers(this._currentMeshSelected);
```
deltakosh: Could you replace it by a regular "if"?

CarlosLanderas: Sure @deltakosh

Issue #3340, opened 2017-12-07
Rework IDB support


deltakosh:
Rely on observables
Check current browser supports state

bghgary:
Make sure to support the IFileRequest interface when reworking this to support cancellation of loading, etc.

deltakosh:
Now with Promises ;)

deltakosh:
MOving to 3.3

PR #3442, opened 2017-12-23
Loader, scene, engine, tools dispose/abort fixes

bghgary:
* Tools.LoadFile now returns a custom request object that allows for abort method and complete event
* Disposing the SceneLoader plugin (only glTF loader right now) will now also abort the request that is created by the SceneLoader
* The scene and engine now keep track of the file requests and abort them if they are still active when disposing


bghgary:
@deltakosh Please review. Thanks!

bghgary:
Note that the IDB support for abort is missing intentionally as we plan on reworking that code anyways. I will put a comment in #3340 when this is merged.

PR #3459, opened 2017-12-29
Clean up DOM interfaces

### LINKED ISSUE #3454: Remove DOM interfaces that are declared by TypeScript
Most of the DOM interfaces exported by BJS appear to be already available in [TypeScript](https://github.com/Microsoft/TypeScript/blob/v2.6.2/lib/lib.dom.d.ts). It might be a good idea to remove them from the BJS source.

Related issue: #550

mootari:
Fixes #3454. I've also removed `MSStyleCSSProperties` as well as some legacy properties on `Window` as I couldn't find any references to them in the BJS source.

### DIFF
--- a/src/Materials/Textures/babylon.videoTexture.ts
+++ b/src/Materials/Textures/babylon.videoTexture.ts
@@ -148,7 +148,7 @@
                     if (onReady) {
                         onReady(new VideoTexture("video", video, scene, true, true));
                     }
-                }, function (e: DOMException) {
+                }, function (e: MediaStreamError) {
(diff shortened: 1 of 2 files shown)


deltakosh:
Need to deep test it but seems fine at first glance
thanks a lot

mootari:
Btw, CANNON.js has its own [declarations file](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/cannon), so `Window.CANNON` could be removed as well after integrating.

PR #3463, opened 2017-12-31
New option in scene to limit mesh candidates inside, minor optimization, minor bug fixes and added Firefox/58 to WebGL2UniformBuffersExceptionList

mjschlot:
I am working on a scene with lots and lots of meshes and needed to improve performance. These are related to that except for the first one which is an issue I just ran into. High-level summary:
* Added Firefox/58 to WebGL2UniformBuffersExceptionList
* Minor fix to set parent of node.
* Added optional argument to isEnabled to disable checking of ancestors.
* Added dispatchAllSubMeshesOfActiveMeshes to dispatch all submeshes of active meshes and bypass what can be CPU intense isInFrustum each frame. Also minor fixes and performance changes in _evaluateSubMesh.
* Optimize dispatch to reduce overhead each frame. On scenes with lots of meshes the additional work was causing low FPS.
* Added IActiveMeshCandidateProvider which provides the author a way to reduce active mesh selection work. Can greatly improves FPS on scenes with lots of meshes since the author can know the details of how their scene is setup and can bypass lots of work in determining active meshes by providing a shorter list of candidates

Except for the first bullet above the changes are for performance and had a measurable impact on FPS. The last change provided the largest impact as it let me reduce the number of meshes being checked significantly when determining active meshes. While BabylonJS needs to be programmed for many unknowns, this IActiveMeshCandidateProvider interface provides me a way to bypass a lot of work because as the author of my scene, I know when things change and when the active meshes need to be recalculated. I have attached a file to this PR to show an example implementation of IActiveMeshCandidateProvider  (please overlook the somewhat hack I am using to clear my internal cache when a node's setEnabled is called).

I apologize for this large PR. I wanted to make smaller PRs but needed to complete this work quickly to use it on an active project.

[ActiveMeshCandidateProvider.ts.txt](https://github.com/BabylonJS/Babylon.js/files/1595387/ActiveMeshCandidateProvider.ts.txt)


deltakosh:
This is good but will require documentation 
I'll review it this tuesday but if you want to tackle the doc update in the meantime ;)
(I guess this page would be the perfect fit: http://doc.babylonjs.com/how_to/optimizing_your_scene)

src/Mesh/babylon.geometry.ts
```diff
@@ -267,7 +267,7 @@
         }
 
         public getTotalVertices(): number {
-            if (!this.isReady()) {
+            if (this.isReady() === false) {
```
deltakosh: is this a real gain?

src/Mesh/babylon.abstractMesh.ts
```diff
@@ -1322,7 +1322,7 @@
             var index: number;
 
             // Action manager
-            if (this.actionManager) {
+            if (this.actionManager !== undefined && this.actionManager !== null) {
```
deltakosh: Is it a proven optimization?

src/Rendering/babylon.renderingGroup.ts
```diff
@@ -315,32 +315,39 @@
...
-            if (!material) {
+            if (material === null || material === undefined) {
                 return;
             }
 
             if (material.needAlphaBlendingForMesh(mesh)) { // Transparent
                 this._transparentSubMeshes.push(subMesh);
             } else if (material.needAlphaTesting()) { // Alpha test
-                if (material.needDepthPrePass) {
+                if (material.needDepthPrePass === true) {
```
deltakosh: not sure this is a gain

src/Mesh/babylon.geometry.ts
mjschlot: Alone no. There is a number of functions which are called thousands of times each frame. With a scene with LOTS of meshes, these small optimizations to if-statements within the functions that are called a lot do have a measurable impact. Alone this does not make enough of an impact to measure, but when this type of optimization is made to a number of places within code that is called a lot, yes, quite a bit. You will see I made this type of change in a number of functions. I was using Chrome's tools to profile frame time and then did this type of quick optimization before moving on to other optimizations I am submitting with this PR.

I measured the difference a few weeks ago when I was first working on this, but sadly did not save the results. I saw about a 8 to 15% decrease per frame when all of the small if-statement optimizations were added up. And when trying to hit 60 fps per second every millescond matters it was significant.

mjschlot:
@deltakosh  I agree with your comments. I responded to one comment, and I'll try to find some clear proof to backup the tests I did previously to demonstrate the benefits.

src/Mesh/babylon.geometry.ts
deltakosh: For instance running this: https://jsperf.com/vs-false/1
will get slightly better result with original code:
image

deltakosh:
But don't get me wrong, I'm all for all other changes (dispatch, interface, etc..)

I'm just questioning the test optimizations

src/Mesh/babylon.abstractMesh.ts
deltakosh: this one seems fine:
https://jsperf.com/davca-vs-undefined
image

src/Mesh/babylon.geometry.ts
mjschlot: I was doing the same thing, and see similar. Here is the test I was running:
https://jsperf.com/strict-equal-vs-generic-if/1

The winner is very close on my computer in both Chrome and Firefox.

deltakosh: yes so this tends to confirm my thought that we may not need this change

src/Mesh/babylon.abstractMesh.ts
mjschlot: I see the same thing here. The results I see after a few test runs are consistent with what you see, and typically are even better for the undefined/null test.

While real code is much more complex than simple tests like on jsperf, it does appear to point in the direction that switching to strict equality did not have an impact, and it was the other if-statement changes I made. I didn't test each one individually but "optimized" them all at once then tested before and after.

Should I undo all if-statements which I just added strict equality too, and keep the changes where it is more undefined / null check?

deltakosh: "Should I undo all if-statements which I just added strict equality too, and keep the changes where it is more undefined / null check?" ==> yes :)

Thank you for your help!

src/babylon.node.ts
```diff
@@ -241,17 +244,22 @@
...
-         * If the node has a parent and is enabled, the parent will be inspected as well.
+         * If the node has a parent, all ancestors will be checked and false will be returned if any are false (not enabled), otherwise will return true.
+         * @param {boolean} [checkAncestors=true] - Indicates if this method should check the ancestors. The default is to check the ancestors. If set to false, the method will return the value of this node without checking ancestors.
          * @return {boolean} whether this node (and its parent) is enabled.
          * @see setEnabled
          */
-        public isEnabled(): boolean {
-            if (!this._isEnabled) {
+        public isEnabled(checkAncestors: boolean = true): boolean {
+            if (checkAncestors === false) {
```
mjschlot: Note: This if-statement needs to use strict equality because `checkAncestors` can be undefined and should not mean "false" in this case.

deltakosh: Yep:)

mjschlot:
I think I got all the if-statements updated. Let me know if there is anything else. Once merged, I'll work on documentation later this week (or next weekend at the latest)

deltakosh:
Thanks a lot! On my phone right now but will definitely merge it asap (no latter than Tuesday:))

deltakosh:
Perfect! Now documentation time ;)

PR #3467, opened 2018-01-02
Change sandbox to play only the first glTF animation

bghgary:
Now that the [glTF sample models are updated](https://github.com/KhronosGroup/glTF-Sample-Models/pull/141), update the sandbox to play only the first animation.


deltakosh:
Cool!

PR #3470, opened 2018-01-03
Change variables of GeometryBufferRenderer protected and _geometryBuf…

xuchen-shadow:
In order to be able to use customized GeometryBufferRenderer
1. Make member variables in GeometryBufferRenderer protected.
2. Make the member variable _geometryBufferRenderer of class Scene protected.


RaananW:
I am not quite sure - what good would it do if you set a protected variable on the scene? TS 'protected' is only for the class and its descendents.
Also - why setting the _scene protected?

What exactly are you trying to achieve? maybe there is a better solution that to simply change private variables to be exposed to the outside world.

xuchen-shadow:
@RaananW , By doing so, I can inherit from GeometryBufferRenderer and can do my custom stuff like using custom shaders. Then I can derive also from Scene and override enableGeometryRenderer(). So that I can use my own shaders to render G Buffers. For example, if I would like to implement multiple clipping planes and integrate SSAO2 stuff, I need to discard the pixels not only in my color buffers but also in my G buffers.

RaananW:
I do understand why it was done, the question is - can't it be done with getters, like the rest of the classes? and - are you extending the scene? you changed a private variable in Scene, which we would very much prefer keeping private.

xuchen-shadow:
@RaananW , yes I am extending the Scene. Yes you are right we can add getters and setters for _geometryBufferRenderer. So what should I do though? I close this PR and create another one?

deltakosh:
Agree with @raananw

xuchen-shadow:
Interesting unknown server error...

src/babylon.scene.ts
```diff
@@ -930,6 +930,15 @@
 
         private _depthRenderer: Nullable<DepthRenderer>;
         private _geometryBufferRenderer: Nullable<GeometryBufferRenderer>;
+        public get geometryBufferRenderer(): Nullable<GeometryBufferRenderer> {
+            return this._geometryBufferRenderer;
+        }
+        public set geometryBufferRenderer(geometryBufferRenderer: Nullable<GeometryBufferRenderer>) {
+            this._geometryBufferRenderer = geometryBufferRenderer;
+            if (this._geometryBufferRenderer && !this._geometryBufferRenderer.isSupported) {
```
RaananW: what are you checking here? why not check it before setting the variable?

src/Rendering/babylon.geometryBufferRenderer.ts
```diff
@@ -1,13 +1,13 @@
 module BABYLON {
     export class GeometryBufferRenderer {
-        private _scene: Scene;
-        private _multiRenderTarget: MultiRenderTarget;
-        private _effect: Effect;
-        private _ratio: number;
+        protected _scene: Scene;
```
RaananW: should be a getter only

src/Rendering/babylon.geometryBufferRenderer.ts
```diff
@@ -1,13 +1,13 @@
 module BABYLON {
     export class GeometryBufferRenderer {
-        private _scene: Scene;
-        private _multiRenderTarget: MultiRenderTarget;
-        private _effect: Effect;
-        private _ratio: number;
+        protected _scene: Scene;
+        protected _multiRenderTarget: MultiRenderTarget;
```
RaananW: should be a getter only

src/babylon.scene.ts
xuchen-shadow: It is somehow the same logic as in enableGeometryBufferRenderer().

src/Rendering/babylon.geometryBufferRenderer.ts
xuchen-shadow: Sure

src/Rendering/babylon.geometryBufferRenderer.ts
xuchen-shadow: Sure

src/babylon.scene.ts
xuchen-shadow: Well I agree that when we call the setter, we should have already made sure that MRT is supported.

RaananW: you are setting `this._geometryBufferRenderer`, then checking if it is supported, and if not, setting is to null. wouldn't it be better to check for support before setting `this._geometryBufferRenderer`? something like:

```
if (geometryBufferRenderer && geometryBufferRenderer.isSupported) {
    this._geometryBufferRenderer = geometryBufferRenderer;
}
```

xuchen-shadow: Totally agree. I will modify it.

src/Rendering/babylon.geometryBufferRenderer.ts
```diff
@@ -2,13 +2,12 @@ module BABYLON {
...
         private _scene: Scene;
         private _multiRenderTarget: MultiRenderTarget;
-        private _effect: Effect;
         private _ratio: number;
-
-        private _cachedDefines: string;
-
         private _enablePosition: boolean = false;
 
+        protected _effect: Effect;
```
RaananW: final question - do you need to change and set this one? or will a getter be enough?

RaananW: Sorry, I looked at the class, it makes sense, since you want to set them in the function. All good.

src/babylon.scene.ts
```diff
@@ -930,6 +930,14 @@
 
         private _depthRenderer: Nullable<DepthRenderer>;
         private _geometryBufferRenderer: Nullable<GeometryBufferRenderer>;
+        public get geometryBufferRenderer(): Nullable<GeometryBufferRenderer> {
+            return this._geometryBufferRenderer;
+        }
+        public set geometryBufferRenderer(geometryBufferRenderer: Nullable<GeometryBufferRenderer>) {
+            if (geometryBufferRenderer && geometryBufferRenderer.isSupported) {
```
deltakosh: I do not understand this setter. Did you forget to set the value?

RaananW: The next line should be 
`this._geometryBufferRenderer = geometryBufferRenderer;`

xuchen-shadow: Oh sorry! By accident!

xuchen-shadow:
@RaananW , should I still do something for you to merge it? Thanks in advance.

RaananW:
sorry, didn't follow up :smile: . all good, merged

PR #3479, opened 2018-01-04
[VRHelper] Add onBeforeCameraTeleport and onAfterCameraTeleport observables

CarlosLanderas:
I am creating GUI instructions to show user how teleport works and detect when camera teleportation is acomplished. It would be nice to have onBeforeCameraTeleport/onAfterCameraTeleport observables so devs can handle UI or other events when using the integrated VRHelper teleportation capabilities.


RaananW:
@sebavan - we need to check why SauceLabs sends constant error. I am merging this, as this has nothing to do with visualization.

PR #3465, opened 2018-01-02
Refactored glTF exporter into class

kcoley:
Example saving multiple babylon scenes into .gltf/.glb:
```javascript
var createScene = function () {
    // This creates a basic Babylon Scene object (non-mesh)
    var scene = new BABYLON.Scene(engine);
    var camera = new BABYLON.ArcRotateCamera("Camera", 0, 0, 10, new BABYLON.Vector3(0, 0, 0), scene);
    // Positions the camera overwriting alpha, beta, radius
    camera.setPosition(new BABYLON.Vector3(0, 0, -20));
    // This attaches the camera to the canvas
    camera.attachControl(canvas, true);
    // This creates a light, aiming 0,1,0 - to the sky (non-mesh)
    var light = new BABYLON.HemisphericLight("light1", new BABYLON.Vector3(0, 1, 0), scene);
    // Default intensity is 1. Let's dim the light a small amount
    light.intensity = 0.7;
    // Our built-in 'sphere' shape. Params: name, subdivs, size, scene
    var sphere = BABYLON.Mesh.CreateSphere("sphere", 6, 3, scene);
    sphere.position.y = 2;
    // Our built-in 'ground' shape. Params: name, width, depth, subdivs, scene
    var ground = BABYLON.Mesh.CreateGround("ground1", 6, 6, 2, scene);
    ground.rotate(new BABYLON.Vector3(1,0,0), 3 * Math.PI/2);
    // Generates a .gltf and .bin file
    //var gltf = BABYLON.GLTF2Export.GLTF(scene, "test");
    // Generates a .glb file

    var gltf = BABYLON.GLTF2Export.GLB(scene, "test");
    BABYLON.GLTF2Export.downloadFiles(gltf);
    return scene;
};
```

PR #3481, opened 2018-01-04
fix node names in serializer

vanRepin:
serializer write mesh name to mesh.name field in gltf, but laoder read it from node.name
I added name to node.name in serialzier, but left name in mesh.name for compatibility, you might want to remove it probably...


vanRepin:
I guess, it should be applied after https://github.com/BabylonJS/Babylon.js/pull/3465 request :)

RaananW:
Yep, this should be added later :) let's see first what happens with #3465

deltakosh:
Let's merge it before the refactor :)

kcoley:
Thanks @vanRepin for the catch!  I'll remove the `mesh.name` field.

PR #3488, opened 2018-01-05
Add teleportationInputsEnabled to VRHelper 

CarlosLanderas:
When using VRHelper, I miss a way to enable / disable gamepad/vr controller inputs, so It would be nice to have a public property to allow changing this setting,

As rotation does not mean teleportation it stays out of the check.

In my case, in the game user needs to interact with GUI elements, and VR is not head-locked so he can stare where he wants, allowing non desired teleporation at that moment.

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -92,6 +92,7 @@ module BABYLON {
         private _rotationAllowed: boolean = true;
         private _teleportationRequestInitiated = false;
         private _teleportationBackRequestInitiated = false;
+        private _teleportationInputsEnabled = true;
         private teleportBackwardsVector = new Vector3(0, -1, -1);
         private _rotationRightAsked = false;
         private _rotationLeftAsked = false;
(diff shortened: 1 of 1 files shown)


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -155,6 +156,20 @@ module BABYLON {
             return this._teleportationTarget;
         }
 
+        /**
+        * Set if VR Controllers / Gamepad inputs are enabled to teleport camera
+        */
+        public set teleportationInputsEnabled(value: boolean) {
```
deltakosh: I'm wondering if we could not simply use this_teleportationEnabled

CarlosLanderas: @deltakosh  this is something I discussed with @RaananW. _teleportationEnabled is used as a flag to control if teleportation code was initialized inside the public method enableTeleportation. You can have collision if you set first teleportationEnabled = true and then invoke enableTeleporation as it won't initialize anything. The other flag, _teleportationAllowed, gets updated every time the target mesh collided by the ray is not the floor mesh/meshes configured so is not a flag to use in this case.

I studied reusing the existing flags but they are not for the same purpose, thats what I decided to create a flag related to the inputs, that are the ones being disabled / enabled.

I'll do whatever you consider better.

deltakosh: I would say it is just a matter of naming here
Perhaps we should rename _tepelortationEnabled to _teleportationInitialized and then reuse teleportationEnabled

CarlosLanderas:
Totally agree. I'll change asap.

PR #3487, opened 2018-01-05
Moved the downloadFiles function to an instance function

kcoley:
`BABYLON.GLTF2Export.downloadFiles(gltf)` --> `gltf.downloadFiles()`

PR #3491, opened 2018-01-05
fix gltf loader to respect vertexAlpha

vanRepin:
https://www.babylonjs-playground.com/#NNZZL9#2
this example shows, that hasVertexAlpha simply set to true when parsed by GLTF2FileLoader.
Green color - hasVertexAlpha = true, red - false. spheres on bottom are original, on top - serialzied and loaded back.
I fixed that, unfortunatly it adds O(N) to parsing time when VertexColor present, where N is vertices count, but I haven't found better universal solution. Example also demostrate problem with loader, when switching hands from right to left.(remove flipHand call, to see the difference). __root__ not helping at all.

### DIFF
--- a/loaders/src/glTF/2.0/babylon.glTFLoader.ts
+++ b/loaders/src/glTF/2.0/babylon.glTFLoader.ts
@@ -358,7 +358,6 @@ module BABYLON.GLTF2 {
             }
 
             node.babylonMesh = new Mesh(node.name || "mesh" + node.index, this._babylonScene);
-            node.babylonMesh.hasVertexAlpha = true;
 
             this._loadTransform(node);
 
@@ -426,7 +425,7 @@ module BABYLON.GLTF2 {
                 for (const primitive of primitives) {
                     vertexData.merge(primitive.vertexData);
                 }
-
+                node.babylonMesh.hasVertexAlpha = mesh.hasVertexAlpha;
                 new Geometry(node.babylonMesh.name, this._babylonScene, vertexData, false, node.babylonMesh);
 
                 // TODO: optimize this so that sub meshes can be created without being overwritten after setting vertex data.
@@ -647,6 +646,10 @@ module BABYLON.GLTF2 {
                         }
                         case "COLOR_0": {
                             vertexData.colors = this._convertToFloat4ColorArray(context, data, accessor);
+                            const hasVertexAlpha = GLTFLoader._GetNumComponents(context, accessor.type) === 4;
+                            if (!mesh.hasVertexAlpha && hasVertexAlpha) {
+                                mesh.hasVertexAlpha = hasVertexAlpha;
+                            }
                             break;
                         }
                         default: {
@@ -678,7 +681,7 @@ module BABYLON.GLTF2 {
                 });
             }
         }
-
+        
         private _createMorphTargets(context: string, node: IGLTFNode, mesh: IGLTFMesh): void {
             const primitives = mesh.primitives;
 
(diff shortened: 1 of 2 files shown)


vanRepin:
I also changed output for gltf serializer. It was changed after last refactoring, probably unintended?

bghgary:
@vanRepin 
> https://www.babylonjs-playground.com/#NNZZL9#1

I'm getting an error with this playground.

kcoley:
I changed the output for the glTF serializer to have `downloadFiles` used as an instance function on the returned output #3487.

i.e.:
```javascript
var gltf = BABYLON.GLTF2Export.GLB(scene, "test");
gltf.downloadFiles();
```

@vanRepin Do you have a preference on the previous way?

vanRepin:
@bghgary fixed, playground, made it before last API changes when live:D

vanRepin:
@kcoley that's ok, I was just confused for "_" before return class in public API and _glTFFiles - to access blob. I'll revert my change.

bghgary:
> I fixed that, unfortunatly it adds O(N) to parsing time when VertexColor present, where N is vertices count, but I haven't found better universal solution. 

You don't need to check each vertex color. To determine if the vertex color has alpha or not should just be checking if the accessor is VEC3 or VEC4. Even if all the alpha values of the vertex color is 1, it should still assume there are vertex alphas. Somehow @kcoley and I missed adding this to the original code.

bghgary:
It will also be good if you can check if build validation still succeeds.

vanRepin:
@bghgary done. Now it's more strightforward, I still don't like saving hasVertexAlpha to IGLTFMesh, but I don't want to pass babylonMesh as parameter either.

bghgary:
> don't like saving hasVertexAlpha to IGLTFMesh

Yeah, no worries. I am working on a change that cleans all that up.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -647,6 +646,10 @@ module BABYLON.GLTF2 {
                         }
                         case "COLOR_0": {
                             vertexData.colors = this._convertToFloat4ColorArray(context, data, accessor);
+                            const hasVertexAlpha = GLTFLoader._GetNumComponents(context, accessor.type) === 4;
+                            if (!mesh.hasVertexAlpha && hasVertexAlpha) {
+                              mesh.hasVertexAlpha = hasVertexAlpha;
```
bghgary: some minor tabbing issue here

loaders/src/glTF/2.0/babylon.glTFLoaderInterfaces.ts
```diff
@@ -217,6 +217,7 @@ module BABYLON.GLTF2 {
 
         // Runtime values
         index: number;
+        hasVertexAlpha:boolean;
```
bghgary: minor: missing space after the colon

kcoley:
@vanRepin, removing the '_' in front of glTFFiles would be a better idea so that it is more evident to be used publicly in the glTF serializer.  Would you mind removing the '_' in this PR?  Otherwise, I can do it on an upcoming material PR for the glTF serializer.

Thanks!

kcoley:
I can actually make the change on my end and rename `._glTFFiles` to just `.files`

loaders/src/glTF/2.0/babylon.glTFLoader.ts
vanRepin: fixed

loaders/src/glTF/2.0/babylon.glTFLoaderInterfaces.ts
vanRepin: fixed

Issue #3308, opened 2017-12-05
Transitioning between separate scenes while keeping VR active is currently not possible


AlienDeathRay:
To avoid complexity of managing separate sets of content within a single scene, It would be great to be able to maintain separate scenes and be able to switch from one to another while keeping the original VR rig active. Perhaps a 'moveScene' function on the VRExperienceHelper would be a good addition with the ability to set a new camera position within the target scene.

sebavan:
Allow transfer of other type of objects would be amazing as well.

deltakosh:
Renamed it to better align with the feature which will be released

TrevorDev:
- [x] Fix add/remove lighting issues
- [x] LoadAssetContainer method
- [ ] Update docs

PR #3494, opened 2018-01-05
Create SceneAssetContainer class and new fileLoader loadAssets method

TrevorDev:
#3308
I would like to get feedback if this is heading in the right direction. Don't merge this although this change should not effect anything as nothing references this class and method yet.


src/Loading/babylon.sceneLoader.ts
```diff
@@ -27,6 +27,7 @@
         load: (scene: Scene, data: string, rootUrl: string, onError?: (message: string, exception?: any) => void) => boolean;
```
TrevorDev: I plan on removing the importMesh and load methods and replace any existing code that references them with calls to the new loadAssets

src/babylon.scene.ts
```diff
@@ -3,6 +3,72 @@
         dispose(): void;
     }
 
+    export class SceneAssetContainer {
```
deltakosh: Would you mind moving it to a separate file?

src/babylon.scene.ts
```diff
@@ -3,6 +3,72 @@
...
+            public autoAnimateLoop:number;
+            public autoAnimateSpeed:number;
+
+            // Skybox
+            public createDefaultSkybox:boolean;
+            public environmentTexture:BaseTexture;
+            public skyboxBlurLevel:number;
+
+            // Objects
+            public cameras = new Array<Camera>();
```
TrevorDev: These are all the objects I will make sure a remove* and add* method exists for within the scene if they don't already exist.

src/babylon.scene.ts
```diff
@@ -3,6 +3,72 @@
         dispose(): void;
     }
 
+    export class SceneAssetContainer {
+            public scene: Scene;
+
+            // Config
+            public useDelayedTextureLoading:boolean;
```
deltakosh: Not sure we want to touch the scene more than just active meshes

src/babylon.scene.ts
```diff
@@ -3,6 +3,72 @@
...
+            // Config
+            public useDelayedTextureLoading:boolean;
+            public autoClear:boolean;
+            public clearColor:Color4;
+            public ambientColor:Color3;
+            public metadata:any;
+            public activeCameraID:string;
+
+            // Physics
+            public gravity:Vector3;
```
deltakosh: Same here

src/babylon.scene.ts
deltakosh: actually for me, only arrays should be in container for now

src/Loading/babylon.sceneLoader.ts
deltakosh: this will be a breaking change but  don't think it will really be impactful
We will need to document it to make sure we are not hurting existing code too much

deltakosh:
Fine to merge

src/Loading/babylon.sceneLoader.ts
TrevorDev: ok, I didn't remove them but currently any external existing loaders will need to be updated to add the loadAssets method.

src/babylon.assetContainer.ts
```diff
@@ -0,0 +1,120 @@
+module BABYLON {
+    export class AssetContainer {
```
TrevorDev: I changed this from SceneAssetContainer to AssetContainer. I can change this if anyone thinks the old name is better or has a better suggestion.

deltakosh: looks good to me

src/Loading/babylon.sceneLoader.ts
deltakosh: We should just add a warning if the plugin does not provide loadAssets nor loadAssetsAsync

src/babylon.assetContainer.ts
```diff
@@ -0,0 +1,120 @@
...
+            this.transformNodes.forEach((o)=>{
+                this.scene.removeTransformNode(o);
+            });
+            this.lensFlareSystems.forEach((o)=>{
+                this.scene.removeLensFlareSystem(o);
+            });
+            this.actionManagers.forEach((o)=>{
+                this.scene.removeActionManager(o);
+            });
+            // TODO, do shadow generators need to be removed somehow?
```
deltakosh: nope we are good as generators are hold by lights

src/Loading/babylon.sceneLoader.ts
TrevorDev: Added a warning when it doesnt exist

TrevorDev:
Works as expected now with all file types tested here after this commit:
http://playground.babylonjs.com/#JA1ND3#48

Append for .babylon files still works as expected tested with mansion demo

Note: GLTF loadAssetContainer only populates the assets that ImportMesh would load but as far as I can tell the gltf loaders do not load lights or cameras at the moment.

dad72:
I have nothing displayed on your PG

deltakosh:
Because the code is not merged yet:)

dad72:
Yes, I do not think about that. lol

deltakosh:
No worry:) we try to prepare PG to help validating and testing changes
@trevordev: would you mind adding your PG in the validation suite?

edzis:
I am so looking forward to use this!

TrevorDev:
Now that I added test to validation suite is this good to be merged?

deltakosh:
WOuld you mind to add a checklist in the original issue (including doc) ?
Thanks a lot

PR #3497, opened 2018-01-07
added optional scope to Observable.removeCallback

abow:


src/Tools/babylon.observable.ts
```diff
@@ -157,11 +157,17 @@
         /**
          * Remove a callback from the Observable object
          * @param callback the callback to remove. If it doesn't belong to this Observable, false will be returned.
+         * @param scope optional scope.  If used only the callbacks with this scope will be removed.
         */
-        public removeCallback(callback: (eventData: T, eventState: EventState) => void): boolean {
+        public removeCallback(callback: (eventData: T, eventState: EventState) => void, scope?: any): boolean {
```
Hersir88: bind() your function is not enough ?

abow:
I could use the remove function, but then my object would need to hold a reference to the observable.

abow:
I mean a reference to the observer.

RaananW:
As I said - I understand the need for that, I just find keeping a reference to the observer a better practice than keeping a reference to a function. If you once use anonymous function,o or bind, or anything that alters the execution of the function (no idea why, but some will), the removeCallback function will not work. 
But notice that I wrote it after approving this, so - no worries :smile:

abow:
My object always has a reference to the function, so it is no additional code.  Thanks

PR #3511, opened 2018-01-09
source.metadata can be null when cloning.

brianzinn:


deltakosh:
Good catch!

MackeyK24:
Yeah, good catch :)

PR #3512, opened 2018-01-09
Basic GearVR support

brianzinn:
Initial GearVR support.  Trigger and trackpad are working, so it is more usable than fallback Generic controller:
- Seems to require long button presses.
- Using Generic model for now.
- Position of pose is always null, so rays are distracting for now.

### DIFF
--- a/Tools/Gulp/config.json
+++ b/Tools/Gulp/config.json
@@ -676,7 +676,8 @@
                 "../../src/Gamepad/Controllers/babylon.oculusTouchController.js",
                 "../../src/Gamepad/Controllers/babylon.viveController.js",
                 "../../src/Gamepad/Controllers/babylon.genericController.js",
-                "../../src/Gamepad/Controllers/babylon.windowsMotionController.js"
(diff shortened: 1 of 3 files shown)


RaananW:
Thanks a lot! Looking good.
Are those the ones integrated in the VR-device itself, or external ones?

brianzinn:
hi @RaananW - this is the external controller, so you can point and click.

brianzinn:
Hi @RaananW - great review comments - thank you.  It was a copy/paste error to not load the model declared in GearVR.  I would say the somewhat convoluted mapping was more for consistency with how the WindowsMotionController (+handleButtonChange(...)) works. In the future when a 3D mesh is found the string[] would perhaps become object[].

deltakosh:
Would you mind to credit yourself in the what new? We can consider this PR as a major update

brianzinn:
Let me see about getting it working better with a proper 3d model.  I'm looking for a place to scan the gearvr and also daydream controller.  I got a pixel2 recently, so will try to get both going.

deltakosh:
I went ahead and added a line for you (just to not forget it)

PR #3515, opened 2018-01-09
executeOnceBeforeRender including optional timeout

RaananW:


src/babylon.scene.ts
```diff
@@ -1924,6 +1924,31 @@
...
+
+        /**
+         * The provided function will run before render once and will be disposed afterwards.
+         * A timeout delay can be provided so that the function will be executed in N ms.
+         * The timeout is using the browser's native setTimeout so time percision cannot be guaranteed.
+         * @param func The function to be executed.
+         * @param timeout optional delay in ms
+         */
+        public executeOnceBeforeRender(func: () => void, timeout?: number): void {
+            if (typeof timeout === 'number') {
```
deltakosh: why not just testing timeout !== undefined?

src/babylon.scene.ts
```diff
@@ -1924,6 +1924,31 @@
             this.onAfterRenderObservable.removeCallback(func);
         }
 
+        private _executeOnceBeforeRender(func: () => void): void {
+            let execFunc = () => {
+                func();
+                this.unregisterBeforeRender(execFunc);
```
deltakosh: I recommend using a setTimeout(0) for the unregister as you are updating the list of callbacks inside a loop thought the same callback list

src/babylon.scene.ts
RaananW: I just wanted to make sure that 0 can still be provided (wanted to avoid !timeout). I can change that

deltakosh: Yep the cost of typeof compared to simple eauqlity is higher

PR #3521, opened 2018-01-10
Add usage to github README.md

TrevorDev:
This is something that I couldn't find very quickly that most github/npm projects have in their readme's.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -57,6 +57,44 @@ If using TypeScript, don't forget to add 'babylonjs' to 'types' in tsconfig.json
 
 To add a module install the respected package. A list of extra packages and their installation instructions can be found on [babylonjs' user at npm](https://www.npmjs.com/~babylonjs).
 
+## Usage
+#### Import the library
+Using an HTML file:
+```
+<script src="https://cdn.babylonjs.com/babylon.js"></script>
+```
+Or in Javascript/Typescript using a bundler such as [Webpack](https://webpack.js.org/) or [Parcel](https://parceljs.org/):
+```
+import * as BABYLON from 'babylonjs';
+```
+#### Add a script
+```javascript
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -57,6 +57,44 @@ If using TypeScript, don't forget to add 'babylonjs' to 'types' in tsconfig.json
...
+#### Add a script
+```javascript
+// Create canvas html element on webpage
+var canvas = document.createElement('canvas');;
+document.body.appendChild(canvas);
+
+// Initialize Babylon scene and engine
+var engine = new BABYLON.Engine(canvas, true);
+var scene = new BABYLON.Scene(engine);
+engine.runRenderLoop(()=>{
```
deltakosh: arrow functions are not supported in IE and some mobile browsers

readme.md
```diff
@@ -57,6 +57,44 @@ If using TypeScript, don't forget to add 'babylonjs' to 'types' in tsconfig.json
 
 To add a module install the respected package. A list of extra packages and their installation instructions can be found on [babylonjs' user at npm](https://www.npmjs.com/~babylonjs).
 
+## Usage
```
deltakosh: wondering if we should point people to this doc instead: http://doc.babylonjs.com/ (and perhaps just add an anchor to the get started portion of it)

deltakosh: And perhaps mention how to import

TrevorDev: Sure, how about this? I still am showing code but I can remove that too if you think it adds too much clutter, but I think that for a lot of developers github is the first and often only place they may land and showing code strait away is a good way to get them hooked.

deltakosh: Nope this is fine :) As long as we point to the doc as well like you did

PR #3525, opened 2018-01-10
glTF loader animation fixes

### LINKED ISSUE #3478: glTF loader does not recognize STEP interpolation
As far as I can tell BabylonJS supports a fair number of easing functions, but loading a glTF model using STEP interpolation in the sandbox reports:

> Unable to load from 'InterpolationComparison.gltf': #/animations/0/samplers/2: Invalid interpolation STEP

Example: [InterpolationComparison.zip](https://github.com/BabylonJS/Babylon.js/files/1602546/InterpolationComparison.zip)

Expected output:
interpolationcomparison

From left to right, STEP, LINEAR, CUBICSPLINE, CUBICSPLINE w/ custom tangents. The fourth (yellow) cube does not animate properly in three.js yet either or may be exported incorrectly, so I'd ignore that.

bghgary:
Fixes #3478

### DIFF
--- a/loaders/src/glTF/2.0/babylon.glTFLoader.ts
+++ b/loaders/src/glTF/2.0/babylon.glTFLoader.ts
@@ -1045,8 +1045,16 @@ module BABYLON.GLTF2 {
 
                 sampler.interpolation = sampler.interpolation || "LINEAR";
 
-                let getNextKey: (frameIndex: number) => any;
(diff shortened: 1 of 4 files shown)


deltakosh:
don't you need to change something in the loader as well?

bghgary:
Yup, somehow left out that change.

bghgary:
Fixed.

PR #3529, opened 2018-01-11
Trigger onSelectedMeshUnselected observable when ray collides the floor meshes

CarlosLanderas:
Right now the VRExperienceHelper has two flows in_castRayAndSelectObject method registered in beforeRender.

When the ray collisions the floor configured meshes it executes the _moveTeleportationSelectorTo and return the execution exiting from method.

If the collided mesh is not part of the floor collection, it continues with the execution that will end triggering onNewMeshSelected / onSelectedMeshUnselected.

The problem is, if I select a mesh I want to interact to and get the onNewMeshSelected observable triggered, and then I immediately look to the floor, the onSelectedMeshUnselected observable will never trigger.

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -112,12 +112,12 @@ module BABYLON {
 
         /**
          * Observable raised when a new mesh is selected based on meshSelectionPredicate
-         */        
+         */
         public onNewMeshSelected = new Observable<AbstractMesh>();
 
         /**
          * Observable raised when a new mesh is picked based on meshSelectionPredicate
-         */        
+         */
         public onNewMeshPicked = new Observable<PickingInfo>();
 
         private _circleEase: CircleEase;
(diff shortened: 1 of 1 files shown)


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -1407,6 +1407,16 @@ module BABYLON {
                 // The object selected is the floor, we're in a teleportation scenario
                 if (this._teleportationInitialized && this._isTeleportationFloor(hit.pickedMesh) && hit.pickedPoint) {
                     // Moving the teleportation area to this targetted point
+
+                    //Raise Mesh unselected event if we trigger the floor meshes and a non-floor mesh was previously selected
+                    if (this._currentMeshSelected &&
+                        !this._isTeleportationFloor(this._currentMeshSelected) &&
+                        this.onSelectedMeshUnselected.hasObservers()) {
```
deltakosh: No need to test for observers. It is done internally

CarlosLanderas: Fixed :P. Thanks for the clarification.

PR #3546, opened 2018-01-15
Fixes SSAO2 bug when using right handed frame

CraigFeldspar:


CraigFeldspar:
If we do this, we cannot use the projection matrix in the shader, it needs negative Z values in case of right handed frame.

PR #3551, opened 2018-01-16
[Documentation] Clarified engine constructor parameters

cotejp:


src/Engine/babylon.engine.ts
```diff
@@ -818,11 +818,12 @@
...
-         * @param {HTMLCanvasElement | WebGLRenderingContext} canvasOrContext - the canvas or the webgl context to be used for rendering
-         * @param {boolean} [antialias] - enable antialias
-         * @param options - further options to be sent to the getContext function
+         * @param {HTMLCanvasElement | WebGLRenderingContext} canvasOrContext - the canvas or WebGL context to use for rendering
+         * @param {boolean} [antialias] - enable antialiasing
+         * @param {EngineOptions} [options] - further options to be sent to the getContext() function
+         * @param {boolean} [adaptToDeviceRatio] - whether to adapt to the device's viewport characteristics
          */
-        constructor(canvasOrContext: Nullable<HTMLCanvasElement | WebGLRenderingContext>, antialias?: boolean, options?: EngineOptions, adaptToDeviceRatio = false) {
+        constructor(canvasOrContext: Nullable<HTMLCanvasElement | WebGLRenderingContext>, antialias?: boolean, options?: EngineOptions, adaptToDeviceRatio?:boolean = false) {
```
deltakosh: You cannot set a question mark and a default value to a parameter

cotejp: Hmm. That's odd. So where does one specify the default value of an optional parameter? In the comment (@param) ?

deltakosh: You do not need the question mark actually
As soon as you give a default value the parameter becomes optional

deltakosh: And yes it is a good convention to also define the default value in comments

cotejp: Ok, I'll do that. Sorry for the trouble. I spotted a few errors in the documentation that I thought I could fix. I guess it would be better for me to run through typedoc before submitting.

Issue #3543, opened 2018-01-14
Switching between Stats and Mesh tabs increases number of drawcalls with each switch


deltakosh:
Mentioned in #3542

deltakosh:
It seems to also increase mat count and active meshes

Probably a problem with instrumentation not being unhooked correctly

deltakosh:
Other thing to check.
If you go to this page: http://www.babylonjs.com/demos/mansion/
Turning the debug layer on and off slow the scene down a lot

PR #3555, opened 2018-01-16
Fixed issue with inspector creating meshes at each display

Temechon:
https://github.com/BabylonJS/Babylon.js/issues/3543


inspector/src/adapters/MeshAdapter.ts
```diff
@@ -10,7 +10,6 @@ module INSPECTOR {
 
         constructor(mesh: BABYLON.Node) {
             super(mesh);
-            new BABYLON.Debug.AxesViewer(mesh.getScene());
```
Temechon: The main issue was here... I suspect a test not removed before pushing.

PR #3556, opened 2018-01-16
Viewer action manager and other changes

RaananW:
The action manager will make registering callbacks to specific events on specific items in the template much easier.

Other changes:
* ground will now be positioned below the object (was centered with certain models)
* code cleanup
* in preparation of reload / drag & drop - update the model configuration when a new model was loaded.


Viewer/src/actionManager.ts
```diff
@@ -0,0 +1,40 @@
+import { EventCallback, TemplateManager } from "./templateManager";
+
+
+export class ActionManager {
```
deltakosh: Do we want to use the same name as BABYLON.ActionManager?

RaananW: Noted. I will change it to EventManager maybe? since it is actually triggered by native js events

deltakosh: Perfect!

PR #3559, opened 2018-01-16
360 video support - In progress

DavidHGillen:
This helper is designed to make it easy to add 360 video to a project, fully featured it will have zoom and VR support.


DavidHGillen:
Currently missing desired features, documentation, and api options. Posting for feedback and requests.

src/Helpers/babylon.videoDome.ts
```diff
@@ -0,0 +1,58 @@
...
+         * The skybox material
+         */
+        private _material: StandardMaterial;
+
+        /**
+         * The surface used for the skybox
+         */
+        private _mesh: Mesh;
+
+        constructor(name: string, urlsOrVideo: string[] | HTMLVideoElement, options: {
```
deltakosh: Constructor needs comment

src/Helpers/babylon.videoDome.ts
```diff
@@ -0,0 +1,58 @@
...
+            clickToPlay?: boolean,
+            size?: number
+        }, scene: Scene) {
+            super(name, scene);
+
+            // set defaults and manage values
+            name = name || "videoDome";
+            options.clickToPlay = Boolean(options.clickToPlay);
+            options.size = Math.abs(options.size as any) || 1000;
```
deltakosh: 1000 should be deduced from camera clipbplanes

src/Helpers/babylon.videoDome.ts
```diff
@@ -0,0 +1,58 @@
...
+
+            // create
+            let material = this._material = new BABYLON.StandardMaterial(name+"_material", scene);
+            this._videoTexture = new BABYLON.VideoTexture(name+"_texture", urlsOrVideo, scene);
+            this._mesh = BABYLON.MeshBuilder.CreateBox(name+"_mesh", {size:-options.size}, scene); // needs to be inside out
+
+            // configure material
+            material.reflectionTexture = this._videoTexture;
+            material.reflectionTexture.coordinatesMode = BABYLON.Texture.FIXED_EQUIRECTANGULAR_MIRRORED_MODE; // matches src
+            material.diffuseColor = material.specularColor = new BABYLON.Color3(0, 0, 0);
```
deltakosh: I recommend using material.disableLighting = true

src/Helpers/babylon.videoDome.ts
DavidHGillen: Ran into several issues with the documentation compiler trying to write the comments I wanted to write, I'm current examining the documentation process to see if it's me or it has a couple of bugs (seems probable form the output)

deltakosh: Cc @sebavan

src/Helpers/babylon.videoDome.ts
DavidHGillen: How would I look up the current camera to make this determination? The scene does not seem to have an active camera property, the other option would be hard coupling the helper to a specific camera but I'd like to avoid that.

src/Helpers/babylon.videoDome.ts
DavidHGillen: Not convinced I wont have to write a custom material to support zooming, it definitely doesn't need lighting, but I'll make sure to make this change in the meantime.

src/Helpers/babylon.videoDome.ts
DavidHGillen: It's there in the live docs, but not the local version along with many other things. Hrm, I might have more issues with my local documentation compile than I thought. Will investigate.

src/Helpers/babylon.videoDome.ts
deltakosh: Thanks!

src/Helpers/babylon.videoDome.ts
deltakosh: What do you mean by local documentation compile? The Babylon.d.ts file?

DavidHGillen: The compile of the documentation I've done locally on my system in order to test the documentation I'm writing in the class.

deltakosh: Ok. we've just pushed a new gulp task for that: gulp typedoc-check

edzis:
@DavidHGillen I assume that with `options.pointToClick` you are targeting mobile browsers and desktop Safari where video with sound is permitted to play only on user action. To support that you might want to disable VideoTexture's `autoPlay`. It is now available on top of #3567

src/Helpers/babylon.videoDome.ts
```diff
@@ -0,0 +1,92 @@
...
+            name = name || "videoDome";
+            options.resolution = (Math.abs(options.resolution as any) | 0) || 12;
+            options.clickToPlay = Boolean(options.clickToPlay);
+            options.autoPlay = options.autoPlay === undefined ? true : Boolean(options.autoPlay);
+            options.loop = options.loop === undefined ? true : Boolean(options.loop);
+            options.size = Math.abs(options.size as any) || (scene.activeCamera ? scene.activeCamera.maxZ * 0.48 : 1000);
+
+            // create
+            let tempOptions:VideoTextureSettings = {loop: options.loop, autoPlay: options.autoPlay, autoUpdateTexture: true};
+            let material = this._material = new BABYLON.BackgroundMaterial(name+"_material", scene);
```
deltakosh: I wonder if the backgroundMaterial is more efficient than the StdMaterial in this case
I would probably prefer having the stdMaterial used here

src/Shaders/background.fragment.fx
```diff
@@ -6,6 +6,8 @@ precision highp float;
 
 #include<__decl__backgroundFragment>
 
+#define RECIPROCAL_PI2 0.15915494
```
deltakosh: why this change?

src/Shaders/ShadersInclude/reflectionFunction.fx
```diff
@@ -1,7 +1,7 @@
 ﻿vec3 computeReflectionCoords(vec4 worldPos, vec3 worldNormal)
 {
 #if defined(REFLECTIONMAP_EQUIRECTANGULAR_FIXED) || defined(REFLECTIONMAP_MIRROREDEQUIRECTANGULAR_FIXED)
-	vec3 direction = normalize(vDirectionW);
+	vec3 direction = vDirectionW;
```
deltakosh: Why this change?

src/Shaders/background.vertex.fx
```diff
@@ -78,6 +80,16 @@ void main(void) {
 
 #if defined(REFLECTIONMAP_EQUIRECTANGULAR_FIXED) || defined(REFLECTIONMAP_MIRROREDEQUIRECTANGULAR_FIXED)
     vDirectionW = normalize(vec3(finalWorld * vec4(position, 0.0)));
+
+    if (fFovMultiplier != 1.0) {
```
deltakosh: TO optimize it we would like to add a #define instead of a "if"

src/Shaders/background.fragment.fx
DavidHGillen: Because the Shader was failing to compile without this line

src/Shaders/ShadersInclude/reflectionFunction.fx
DavidHGillen: It's redundant. All other places vDirectionW is set it's set with a normalized value.

src/Helpers/babylon.videoDome.ts
DavidHGillen: All the extra facilities the Standard Material offers aren't needed for this use case. Additionally, very special functionality had to be added that wouldn't make sense in a Standard Material. Background Material seems more appropriate, is likely to perform better, and the also seemed like the best place to add the specific functionality we implemented. That's why I chose it and think it's a better fit than the Standard Material.

src/Shaders/ShadersInclude/reflectionFunction.fx
deltakosh: ok cool!

src/Shaders/background.vertex.fx
DavidHGillen: Branching on a uniform is much cheaper than on other data types, as the shader knows it's locked into a single value for its entire run. Further, it's in the vertex shader so it is not running that often. Given the power of modern GPUS and these mitigating factors, I thought it was worth having as a uniform instead of a define. As a uniform it can easily and cheaply be animated, like adjusting a pair of binoculars or smoothing the change in a video dome. Otherwise as a define it is forcing a shader recompile with every change to the value.

deltakosh: I meant only replacing the if by a define
Everything else stays and you still have an uniform for the animations

src/Helpers/babylon.videoDome.ts
deltakosh: I wanted to check that the background material was not doing extra work (like gamma / linear) but it seems that it should be ok
Just wanted you to make sure that the background material is initialized without other options (diffuse, noise, etc..). We want the straightest shader possible.

src/Shaders/background.fragment.fx
DavidHGillen: Re-confirmed. Shader fails to compile in chrome without that addition. If it's not the correct way or place for it to be added then something else is amiss as it is needed.

deltakosh: Can you pinpoint where it is used. The shaders work without it so far so it must have to do with something you added.

DavidHGillen: It's used in the `reflectionFunction.fx` file which always gets included in the backgroundMaterial. If you specify a reflection of an equirectangular nature, then it will get enabled by an ifdef. This is not code I added, but simply trying to use an equirectangular background reflection.

As for where and why I placed it, I was simply following the pattern other shaders had used for its implementation. If you would prefer it included in another file, importing a list of constants from another file, or instead wrapping it behind an ifdef too I can do so.

deltakosh: OK nope this is fine. I just wanted to understand :)

deltakosh:
Could you update the what's new in /dist/preview release?

deltakosh:
Some conflicts prevent the merge

PR #3567, opened 2018-01-18
More flexible Video Texture

edzis:
Allows to override `autoPlay` and `loop` values
Allows to use manual texture updating
Maintains backwards-compatibility

As a fallow-up on http://www.html5gamedevs.com/topic/34582-videotexture-1-frame-per-render-loop/?do=findComment&comment=201701

### DIFF
--- a/src/Materials/Textures/babylon.videoTexture.ts
+++ b/src/Materials/Textures/babylon.videoTexture.ts
@@ -1,4 +1,11 @@
 ﻿module BABYLON {
+    /**
+     * Settings for finer control over video usage
+     * @typedef {Object} VideoTextureSettings
+     * @property {boolean} [autoPlay] Applies `autoplay` to video, if specified
+     * @property {boolean} [loop] - Applies `loop` to video, if specified
+     * @property {boolean} autoUpdateTexture - Automatically updates internal texture from video at every frame in the render loop
+     */
     export interface VideoTextureSettings {
         autoPlay?: boolean;
         loop?: boolean;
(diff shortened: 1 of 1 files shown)


edzis:
@deltakosh I would have removed `name` from constructor and used just use the computed value from initial `src`, but that would be breaking backward's compatibility.

For advanced usage see

    this._scene.onBeforeRenderObservable.add(this.updateVideoTextures)

    private updateVideoTextures = (): void => {
      const activeMeshes = this._scene.getActiveMeshes()
      each(activeMeshes, mesh => {
          const material = mesh.material as StandardMaterial
          const texture = material.diffuseTexture
          if (texture instanceof VideoTexture) {
              texture.updateTexture(true)
          }
      })
    }

src/Materials/Textures/babylon.videoTexture.ts
```diff
@@ -1,44 +1,106 @@
...
+                autoPlay: true,
+                loop: true,
+                autoUpdateTexture: true,
+            }
+        ) {
             super(null, scene, !generateMipMaps, invertY);
 
-            var urls: Nullable<string[]> = null;
-            this.name = name;
+            this._engine = (<Scene>this.getScene()).getEngine();
```
deltakosh: You can use this.getScene()!.

edzis:
@deltakosh The latest CI error make no sense to me.

src/Materials/Textures/babylon.videoTexture.ts
```diff
@@ -1,44 +1,116 @@
...
+                source.src = url;
+                video.appendChild(source);
+            });
+        }
+        return video;
+    };
+
     export class VideoTexture extends Texture {
-        public video: HTMLVideoElement;
+        readonly autoUpdateTexture: boolean;
```
deltakosh: this is public by default and thus it requires a comment (and please add public keyword)

edzis:
@deltakosh I added comments on those 2 properties, but I don't get what is the issue behind errors about inner props of Settings.

deltakosh:
Oh yeah you have to comment on a per property basis instead of using the @property tag

edzis:
Finally...
Thanks for the tip!

deltakosh:
Please update the what's new to credit yourself :)

Issue #2971, opened 2017-10-17
Improve particles System


deltakosh:
- [x] More emitter shapes (cone, sphere)
- [x] Noise:  Being able to read random values from perlin noise textures instead of pure JS random code
- [x] Sub emitters: Particles should be able to spawn a new system when they die

IbraheemOsama:
Can you assign me this task :)

deltakosh:
Will do as soon as you accept the invite

IbraheemOsama:
@deltakosh yaaaaaaaaay I did it my self 😃 thanks David

IbraheemOsama:
@deltakosh hi David, I don't get what is 
_Noise: Being able to read random values from perlin noise textures instead of pure JS random code_
I assumed you want to use a similar random method like class PerlinNoiseTextures but I don't see any similar functions. !!

deltakosh:
So far the particles use random object from js
The idea would be to rely on external data loaded into an data array and use this data to generate random values.
The external data could be then loaded from an image containing perrin noise data (which can be generated by our procedural textures)

IbraheemOsama:
@deltakosh have a few question if I want to generate random values between 1.5 and 2 for example then I should fill this array previously with values between 1.5 and 2 then use a random method to choose from this array isn't that less efficient than JS rand function ??

deltakosh:
The idea is to let artists provide the random texture ( a gray level texture filled with value between 0 and 1). Then the random method will start for pixel at coordinate 0,0 and move forward to fetch new random values.
Does it make sense?

PR #3570, opened 2018-01-18
Sub Emitter

IbraheemOsama:
Sub emitters #2971


IbraheemOsama:
[subEmitterTest.txt](https://github.com/BabylonJS/Babylon.js/files/1642226/subEmitterTest.txt)

deltakosh:
Hello!

first of all thanks A LOT for your contribution (as usual).

I would like to keep things a bit simpler.

I like the way you create the API and we should stick with it. But I do not think we need to have the concept of subemitters as a class. 

When a particle die and if there is particle systems in the subEmitters property, the particle should randomly pick one, clone it and run it from its current position.. We should then track the list of activeSubEmitters if users want to do something with it.

And that's pretty much it.

obviously, calling stop or dispose on a particle system should let the users decide if he wants to call the same function on all sub emitters and so on..

IbraheemOsama:
Ok :)
Regarding random picking from the SubEmitter, what if the user wants to control which to go first then second and so on ?
Also in case of random picking I will introduce a new property to check the max number of subEmits :)

I will do it this way, if maxNumberOfSubEmits is defined I will pick a random emitter if not I will pick depending on generation, what do you think ?

deltakosh:
Wonderful :)

IbraheemOsama:
[subEmitterTest.txt](https://github.com/BabylonJS/Babylon.js/files/1646223/subEmitterTest.txt)
new file to test.
To test the random selection or incremental comment\uncomment
particleSystem.maxNumberOfSubEmits = 2;

src/Particles/babylon.particle.ts
```diff
@@ -61,11 +61,15 @@
          * Creates a new instance of @see Particle
          * @param particleSystem the particle system the particle belongs to
          */
-        constructor(private particleSystem: ParticleSystem) {
+        constructor(public particleSystem: ParticleSystem) {
```
deltakosh: As this is a public property now you need to comment it like this:
constructor(
/** ... */
public particleSystem: ParticleSystem) {

src/Particles/babylon.particleSystem.ts
```diff
@@ -333,13 +333,39 @@
...
         private _started = false;
         private _stopped = false;
         private _actualFrame = 0;
         private _scaledUpdateSpeed: number;
         private _vertexBufferSize = 11;
         private _isAnimationSheetEnabled: boolean;
 
+        // end of sheet animation
+
+        // sub emitters
```
deltakosh: Sub-emitters or sub emitters? (We need to decide on the way to write it :))

src/Particles/babylon.particleSystem.ts
```diff
@@ -333,13 +333,39 @@
...
+        /**
+         * this is the sub emitters templates that will be used to generate particle systems when the particle dies, this property is used at the root ParticleSystem only.
```
deltakosh: BY the root particle system (no 's')

src/Particles/babylon.particleSystem.ts
```diff
@@ -484,15 +497,23 @@
...
+                this.activeSubSystems = new StringDictionary<ParticleSystem>();
+                this.stockSubSystems = new StringDictionary<Array<ParticleSystem>>();
+            }
         }
 
         /**
          * Stops the particle system.
          */
-        public stop(): void {
+        public stop(stopSubEmitters = true): void {
```
deltakosh: parameter comment (hint: run gulp typedoc-check to see if you have issues with doc)

src/Particles/babylon.particleSystem.ts
```diff
@@ -540,10 +561,144 @@
...
+
+            this.activeSubSystems.forEach(subSystemName => {
+                var subSystem = this.activeSubSystems.get(subSystemName)!;
+                subSystem.stop();
+                subSystem._stoppedEmitting(true);
+            });
+            this.activeSubSystems = new StringDictionary<ParticleSystem>();
+        }
+
+        private static emitFromGeneration(rootSystem: ParticleSystem, particle: Particle, generation: number): void {
```
deltakosh: not sure why you will need a generation mechanism.
If a particle dies and the root particle system has sub emitters then spawn a new system and run it.

What else am I missing?

deltakosh: If I'm right (and I'm probably wrong :)) then we don't need all the static functions

IbraheemOsama: We don't need the static that's correct I did this way to force not using this because it was too confusing :) but we need the generation to know which sub emit we are currently at hence we can decide to stop or continue.

I will remove the static :)

deltakosh:
So I would see _emitFromGenerationUsingSystem becoming more like:
_emitFromGenerationUsingSystem(particle) only. rootSystem === this in this case

IbraheemOsama:
For the _emitFromGenerationUsingSystem(particle) I'm trying to reduce the if statements as much as possible for performance.

for the generation, the problem I'm trying to solve in cascading (not the random) is to prevent recreating new emitters with same specs and I need to save this data in a dictionary and each generation is representing an stock emitter in the subEmitters array.

The other problem is knowing when to stop emitting ? if the user wants to the sub emits to be only 4 times and he used a random emitter how can I know from the current System if should I emit or stop.

You got my point ? :)

deltakosh:
No :) (I'm old and slow so bear with me :))

- Knowing when to stop: Sub-Emitters are generated from each dying particles. This is a user decision to stop generating new particles or not. But we can provide a maxSubEmitterCount that will block generating new of the previous one are still alive. No need for generation here. Just the list of active subemitters.

- Recreating with same specs: Not sure to follow you. If not in random mode, you only need to store the current subEmitter index in the root particle systems right?

IbraheemOsama:
If you're slow then developers like me should stop developing and searching for something else to do. 😆 

- First case how can I know when to stop if there is no generation if the user provided maxSubEmitterCount =2 for example ? how can I get the info when to stop or not from the list of ActiveSubEmitter, I don't know if I'm in the first sub emit or the second if I don't know the generation ? (I think I'm missing something :) )

- I'm talking about the non-Random case when the user didn't provide maxSubEmitterCount but provided 3 SubEmitters for example, I will start with the first SubEmitter in the array and when particles dies then the second in array will start then the third and so on, so I need to separate the first created SubSystem (all of the created ones has same specs on the same level) from the created in the second subEmits and so on :) 

What do you think ?

deltakosh:
1. if maxSubEmitters == 2, then you do not generate a new one until the activeSubEmitters.count == 2. When a subEmitter stops you just need to remove it from activeSubEmitters array.
2. But for that you only need an index no?

IbraheemOsama:
I think I'm missing something here 😄 the activeSubEmitters contains all current active subSystems and this number is variable.
As a scenario in the cascading scenario (No the random) if I've system that has EmitRate =5 and 2 sub system each has manualEmitCount =5

When it start it will create around 5 particles and when each particle dies it will create a new subSystem
so now we have 5 Subsystems each with generation 1 and the activeSubEmitters.count is 5.

Now it's time for the 25(from 1st generation) particles to die mean while still activeSubEmitters.count is 5.
to check when the particles dies should I spawn a new system or now I have to check 
if( current generation <= the SubEmitters,count) then create new ones because I've 2 SubEmitters and I'm still in generation one (the first SubSystem) at this moment  activeSubEmitters.count is still 5 and the number will never predicate when to stop or to continue emitting because at each system\SubSystem the emitting rate is variable and the particles dies at different age even at same generation.

For the second point I already have the generation so why to use something else 😄 and the index number will represents the generation in the cascading case too.

Apologies for the long thread, hopefully I clear enough :)

deltakosh:
Lol no worry :)

Based on your current simulation:
- 5 particles were emitted
- When they die (let's say at the same time), they will generate 5 particle systems which will be added to the activeSubEmitters
- We don't care about what subemitters do. They can have subEmitters or not, it is not the problem of the root particle system
- At root particle system level, new particles are emitted by since activeSubEmitters.count == 5, no new subEmitter is generated until one of the subEmitters stops and remove itself from the active list
- A new sub emitter will be generated only if activeSubEmitters.count < 5

Recursively, tue subEmitters can be setup to also have subEmitters but the current algorithm still applies at their level

Does it make sense? (sorrry for that :( )

IbraheemOsama:
Now, I understand where is the confusion :) in the current implementation I allow only SubEmitter at the root level and even if the user defined SubEmitters to a subEmitter it will be useless and the rootSubEmitter is the main controller even ActiveSubEmitters is populated at the root level only :)

If I implemented it your way it means that everySubEmitter is completely a separate System (but I will share the particles across all then)  then I don't need maxNumberOfSubEmits. if I found a sub emitter I will emit and if I found more than one sub emitter I will pick one randomly and emit from it. but the problem here is the stockSubSystem will not be shared across all system will be dedicated to every single SubEmitter so reusing the SubEmitter will be less (because every SubEmitter could have totally different SubEmitters even if they are same type but different properties)

What do you think ? Should I change it or leave it as it is ?

deltakosh:
I would like it to be as close as Unity as possible. So could you please change it to align to the algorithm I mentioned?
(I was not clear at the beginning so sorry for all the back and forth)

IbraheemOsama:
No worries :) the change shouldn't be big

deltakosh:
Hey :) do you need help? I guess you should be overwhelmed right now :D

IbraheemOsama:
hehehe 😆 No I didn't start yet, will start today, I was recording a course but now I'm a free man :D

IbraheemOsama:
[subEmitterTest.txt](https://github.com/BabylonJS/Babylon.js/files/1734353/subEmitterTest.txt)
New changes requires new file for testing :)

IbraheemOsama:
@deltakosh ping

src/Particles/babylon.particleSystem.ts
```diff
@@ -491,15 +498,24 @@
...
+                this.stockSubSystems = new StringDictionary<Array<ParticleSystem>>();
+            }
         }
 
         /**
          * Stops the particle system.
+         * @param stopSubEmitters if true it will stop the current system and all created sub-Systems if false it will stop the current root system only, this param is used by the root particle system only. the default value is true.
          */
-        public stop(): void {
+        public stop(stopSubEmitters = true): void {
```
deltakosh: property not used

src/Particles/babylon.particleSystem.ts
```diff
@@ -325,13 +325,33 @@
...
+        /**
+         * this is the Sub-emitters templates that will be used to generate particle system when the particle dies, this property is used by the root particle system only.
+         */
+        public subEmitters: ParticleSystem[];
+        /**
+        * The current active Sub-systems, this property is used by the root particle system only.
+        */
+        public activeSubSystems: Array<ParticleSystem>;
+
+        private stockSubSystems: StringDictionary<Array<ParticleSystem>>;
```
deltakosh: per our previous discussion we don;t need this :)

src/Particles/babylon.particleSystem.ts
```diff
@@ -446,20 +467,6 @@
             this._indexBuffer = this._scene.getEngine().createIndexBuffer(indices);
         }
 
-        /**
-         * "Recycles" one of the particle by copying it back to the "stock" of particles and removing it from the active list.
-         * Its lifetime will start back at 0.
-         * @param particle The particle to recycle
-         */
-        public recycleParticle(particle: Particle): void {
```
deltakosh: This has to remain the same

src/Particles/babylon.particleSystem.ts
```diff
@@ -555,10 +571,139 @@
...
+            });
+            this.activeSubSystems = new Array<ParticleSystem>();
+        }
+
+        private _createParticle: () => Particle = () => {
+            return this._createParticleUsingSystem(this, this);
+        }
+
+        // to be overriden by subSystems
+        private _stoppedEmitting: () => void = () => {
```
deltakosh: why this?

src/Particles/babylon.particleSystem.ts
```diff
@@ -555,10 +571,139 @@
...
+            if (!this.subEmitters || this.subEmitters.length === 0) {
+                return;
+            }
+
+            var templateIndex = Math.floor(Math.random() * this.subEmitters.length);
+            var templateIndexString =templateIndex.toString();
+
+            if (!this.stockSubSystems.contains(templateIndexString) || (this.stockSubSystems.contains(templateIndexString) && this.stockSubSystems.get(templateIndexString)!.length === 0)) {
+                // get the current generation template and clone it to subSystem
+                var subSystem = this.subEmitters[templateIndex]._cloneToSubSystem(this, particle.position);
```
deltakosh: we only need this block and not the associated else

src/Particles/babylon.particleSystem.ts
```diff
@@ -555,10 +571,139 @@
...
+                    stockSubSystem.manualEmitCount = this.subEmitters[templateIndex].manualEmitCount;
+
+                stockSubSystem.start();
+            }
+        }
+
+        private _initSubSystem(rootParticleSystem: ParticleSystem): void {
+            this._rootParticleSystem = rootParticleSystem;
+
+            this._stoppedEmitting = () => {
```
deltakosh: subSystem are regular system. They just need to keep a ref to the originator in order to be stopped simultaneously on demand. They do not reply on it for particle emission or recycling. Every system is completely independant

src/Particles/babylon.particleSystem.ts
IbraheemOsama: Good Catch :)

src/Particles/babylon.particleSystem.ts
IbraheemOsama: Performance, If I remove it the FPS is dropping :(

src/Particles/babylon.particleSystem.ts
IbraheemOsama: But I created a function variable so the consumer won't feel the difference :) if I have to keep it a function I will do a if else inside the recycleparticle and I'm trying to avoid that as much as possible :)

src/Particles/babylon.particleSystem.ts
IbraheemOsama: I still need to dispose the last generated ParticleSystem because it was dropping FPS in a huge way :) after 10 mins to was dropped to 30.

IbraheemOsama: I will merge the main and sub into one function

src/Particles/babylon.particleSystem.ts
deltakosh: I do not get why (sorry :( ).
The stockSubSystem is used to store active ps, correct? And you use it to clone new PS instead of instancing new one.
How could this impact rendering perf?

src/Particles/babylon.particleSystem.ts
deltakosh: But you should not use root PS to create particles for sub emitters. They have to be independant

IbraheemOsama:
T-Shirt or No T-Shirt ? 😈

deltakosh:
I'll check next week. I'm moving to a new house and I'm literally drowning in a sea of moving boxes :)

deltakosh:
Congrats!!! now a bit of doc? :)

deltakosh:
https://github.com/BabylonJS/Documentation/issues/844

PR #3571, opened 2018-01-18
Explicit prettier settings

edzis:
Context:
- Babylon includes .vscode workspace settings in repo.
- It specifies values only for some properties and depends on default values in a lot of cases.
- My user settings override some of the default values.
- This messes up auto-formating

This PR does not by any means contain an excessive list of settings but just those that I had to change to make it work for me.
I assume other developers might need to override different settings, but I don't think that including all the settings is reasonable.

All of the new values are the default values, except for `printWidth` - I believe it works better with the extra characters used by typing and the current style of babylon having longer lines.


RaananW:
I must be missing something - if the values are not there, why are they being overridden? You added a few parameters, that can be set in your workspace.
Or have I misunderstood?

edzis:
When babylon does not specify these in workspace settings explicitly they are deferred from user settings and then from default values. Babylon has to accept that contributors can have user settings which override the default values that babylon expects.

I have my own user settings where I have chosen different settings than the default ones. I assume other contributors have their own user settings.
Babylon silently expects different workspace settings, which clashes against my user level settings.
For me to contribute efficiently I have to override these settings on the project level. On other projects where `.vscode` settings are not in version control I would have just added them locally and the problem would have been solved.

edzis:
To give specific example, here is the clashing excerpt from my user settings:

    ...
    "editor.tabSize": 2,
    "prettier.semi": false,
    "prettier.singleQuote": true,
    "prettier.trailingComma": "es5",
    ...

I hope you understand my desire not to change them and code in all my internal projects just so I can play nicer with babylon.

deltakosh:
We cannot get external plugin configuration in the settings of the project unfortunately 
The settings we share are required for the project and only target vscode

deltakosh:
Please do not take it personally but I will close this PR as Prettier is not a tool we are ready to support / add

edzis:
For the record, the last 2 commits should not be considered as real suggestions but as a visualization for #3573

But I rest my case.

deltakosh:
Please don't be upset :(

PR #3574, opened 2018-01-18
WebVRCamera - Consider user configured default height and allow zero …

CarlosLanderas:
WebVRCamera is not considering now the defaultHeight configured by the user when a device is reporting stage parameters and the standingMatrix is initialized. In addition, checking conditions with defaultHeight default value 0 won't allow it to be a valid value.

I used 0 as default height when user does not set it and display either, but If you consider another value would be better just tell. For example VRExperienceHelper initializes it as 1.7 if defaultHeight property not set in VR options.

I realized of this when I changed my Mixed Reality Portal headset that is not reporting sittingToStandingTransform parameters to Oculus Rift that reports it.

### DIFF
--- a/src/Cameras/VR/babylon.webVRCamera.ts
+++ b/src/Cameras/VR/babylon.webVRCamera.ts
@@ -75,7 +75,7 @@ module BABYLON {
 
         private _lightOnControllers: HemisphericLight;
 
-        private _defaultHeight = 0;
+        private _defaultHeight?: number = undefined;
         constructor(name: string, position: Vector3, scene: Scene, private webVROptions: WebVROptions = {}) {
             super(name, position, scene);
             this._cache.position = Vector3.Zero();
@@ -165,13 +165,13 @@ module BABYLON {
         }
 
         public deviceDistanceToRoomGround = () => {
-            if (this._standingMatrix) {
+            if (this._standingMatrix && this._defaultHeight === undefined) {
                 // Add standing matrix offset to get real offset from ground in room
                 this._standingMatrix.getTranslationToRef(this._workingVector);
                 return this._deviceRoomPosition.y + this._workingVector.y
-            } else {
-                return this._defaultHeight;
             }
+            //If VRDisplay does not inform stage parameters and no default height is set we fallback to zero.
+            return this._defaultHeight || 0;            
         }
 
         public useStandingMatrix = (callback = (bool: boolean) => { }) => {


TrevorDev:
Hey @CarlosLanderas, could you share why you want to use the set height when the standing matrix is available to the vrHelper? The deviceDistanceToRoomGround method is only used in the webVRHelper to place the view to the correct users height above the ground when teleporting. When would you want to use the value set instead of the actual value provided by the device?

CarlosLanderas:
@TrevorDev, I was configuring several headsets, Oculus, Lenovo using different platforms like Oculus software, Steam VR and I found the value provided by the headset when you configure for example 1.80cm height was so small that current code made the camera fall close the ground. You can enter even in a VR scene without configuring that value, so I thought it would be nice to have the chance of setting your own default distance to the floor.

CarlosLanderas:
Maybe we should have a property to scale the device height value brought by the headset display to apply it depending on the scene?

TrevorDev:
Thanks for your answer. I think maybe the confusion is that distance here was intended to be in meters and this method was created to find the device distance to the ground in meters. When using Vive in Firefox when he stage parameters are provided this method should always return the physical height of the headset in the room so ~1.7 meters when standing up and ~0 when placed on the floor. For devices where this is not able to report, it will just take the default height specified, eg. always standing height of around ~1.7m.

@CarlosLanderas is your scene setup such that 1 Babylon unit is 1cm? If so I will definitely add the units to the doc and potentially add some sort of scale parameter to this method if helpful.

CarlosLanderas:
@TrevorDev, I am not sure what unit is being used in the babylon exported file. But when VRHelper teleport camera sets to WebVRCamera Y Axis the deviceDistanceToRoomGround  output value being for example (1.6 - 1.7) it is extremelly tiny. Like 3d designers can choose their own unit and every mesh is relative to the others, like I wrote above and you suggest, I think the best thing would be adding that scale parameter so we can support both scenarios. Auto use stage parameters when available and have the chance to scale the device measure and use default height when parameters are not provided by the display.

RaananW:
Hi Guys, just jumping in. Not sure I understand - are we referring to a global scale factor for the scene? There is one in the WebVR options of the camera (https://github.com/BabylonJS/Babylon.js/blob/master/src/Cameras/VR/babylon.webVRCamera.ts#L69),  whic is taken into account when receiving new pose data (https://github.com/BabylonJS/Babylon.js/blob/master/src/Cameras/VR/babylon.webVRCamera.ts#L589). Maybe this parameter can be used here as well? I see it is not being used when setting the height, might be a nice way to globally solve it.
Or am I misunderstanding?

CarlosLanderas:
I remember using that parameter and not having the desired effect. Maybe I used it the wrong way but I remember having effects on how the camera was rendering the scene. Will try again.

TrevorDev:
If you can scale down the objects in the scene from centimeters to meters that might work too. I tried setting device scale but for some reason I cant see my controllers when I do that. I created a new issue to track this: https://github.com/BabylonJS/Babylon.js/issues/3712

CarlosLanderas:
Being the problem the teleportCamera in VRHelper, with don't we expose in VRHelper an scaling number so when it returns from distance to the ground method it sets the Y Axis to distance * multiplier? Default  value  could be 1.

RaananW:
The deviceScaleFactor currently scales the position delivered from the vr device. It can (or should?) scale the rest of the parameters as well. This is technically a "world scale".

PR #3579, opened 2018-01-18
glTF khr_lights extension support

MiiBond:


deltakosh:
Adding @bghgary as main validator here

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
```diff
@@ -0,0 +1,125 @@
...
+            return this._loadExtension<IGLTFLightReference>(context, scene, (context, extension, onComplete) => {
+                if (extension.light >= 0 && loader._gltf.extensions) {
+                    const lightInfo = loader._gltf.extensions.KHR_lights.lights[extension.light];
+                    if (lightInfo.type !== 'ambient') {
+                        return;
+                    }
+
+                    // Use monochromatic hemispheric light to simulate ambient light.
+                    const direction = new Vector3(0, 1, 0);
+                    extension.babylonLight = new BABYLON.HemisphericLight("light", direction, loader._babylonScene);
```
deltakosh: Why not using scene.ambientColor?

sebavan:
@MiiBond , looks good to me.

About the spot lights, did you consider gaussian falloff (for the angle) as it would prevent setting up innerConeAngle and outerConeAngle  ?

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
MiiBond: I thought about that but, for that to work, don't all materials have to have an ambient colour specifically set? That would seem to complicate having the glTF just load and work seamlessly.

MiiBond:
@sebavan Yeah, that's a fair question (and probably belongs in the chat about this extension - https://github.com/KhronosGroup/glTF/issues/945). My thought was that the inner/outer specification lets the user play with the shape more to try to match different spotlight designs. For example, the shape of the physical sheath around the light has a profound effect on the falloff being cast. Limiting the light to one type of falloff prevents us from matching these real-world scenarios. What do you think?

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
deltakosh: This is more about performance then. Having to compute a global light for all meshes will be a lot just for the sake of it
So we should mention that ambient is expected to be white by default on material

sebavan:
@MiiBond Makes sense to me, maybe inner could be optional and if not present a gaussian would be used ?

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
MiiBond: Weird. I had tried using `scene.ambientColor` before and couldn't get it to work without explicitly setting the ambientColor to white for a material. However, I tried it again and it's working now...

So... okay, that's done :)

MiiBond:
@sebavan I think that probably makes sense. I've posted this in the issue thread for KHR_lights.

MiiBond:
@bghgary, did you have any thoughts on this pull request? I just realized that I opened this to merge into `master` but I guess the usual flow would be to merge changes into `preview` first. Is that correct? If so, I can open a new pull request.

deltakosh:
Nope this is correct:) master is where your ant to merge 
Preview is when we publish to the cdn

bghgary:
@MiiBond 
> did you have any thoughts on this pull request?

I think it's okay. I'm working on a large change that will conflict, so I'll just let you check it in for now. I can fix it up afterwards.

bghgary:
@MiiBond FYI, IGLTF was renamed to _IGLTF as a stop gap before my changes.

MiiBond:
@bghgary Any idea why the preview deploy is failing for this pull request?

deltakosh:
No problem for netlify(it is not blocking)

MiiBond:
Is this good to be merged then?

deltakosh:
Thanks a lot!

PR #3586, opened 2018-01-19
adding depth of field to default pipeline

TrevorDev:
Open issues:
- [ ]  Depth buffer seems 1 frame behind color buffer
- [ ]  setTextureFromPostProcess does not work as originally expected
- [ ]  Not sure how to expose dials on default pipeline (option1: expose them all hanging off of default pipeline like the current bloom, option2: add new dofControl class hanging off of default pipeline)
- [ ]  Documentation for all the files
- [ ]  There is possible refactoring that could be done to move the circleOfConfusion pass into the depthOfFieldMerge pass to avoid another pass
- [ ]  Rendering optimizations such as avoiding clears

### DIFF
--- a/Tools/Gulp/config.json
+++ b/Tools/Gulp/config.json
@@ -58,6 +58,8 @@
             "additionalPostProcesses",
             "additionalPostProcess_blur",
             "additionalPostProcess_fxaa",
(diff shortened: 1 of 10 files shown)


TrevorDev:
Closing to break into smaller peices

Issue #3578, opened 2018-01-18
useStandingMatrix should use the vrDevice instead of retrieving a new one from webVR api


TrevorDev:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Cameras/VR/babylon.webVRCamera.ts#L182
should be implemented similar to 
https://github.com/BabylonJS/Babylon.js/blob/master/src/Cameras/VR/babylon.webVRCamera.ts#L285

TrevorDev:
Closing for now, see PR

PR #3589, opened 2018-01-19
try refactor useStandingMatrix

TrevorDev:
Do not merge. [Not tested]

Hey @RaananW is this along the lines of what you were thinking? 

I'm trying to keep the callback logic working the same as before but I think I'll have to make it even more complicated to do that.

#3578


src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -125,6 +128,22 @@ module BABYLON {
                 if (this._attached) {
                     this.getEngine().enableVR();
                 }
+                if(this._useStandingMatrix){
+                    if (!this._vrDevice.stageParameters.sittingToStandingTransform) {
+                        this._useStandingMatrixCallback(false);
+                    }else{
+                        this._standingMatrix = new Matrix();
```
RaananW: I believe this function should be executed anyhow, storing the standingMatrix.
if this._useStandingMatrix is set to true, it should be used. if not, then no.

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -125,6 +128,22 @@ module BABYLON {
...
+                        Matrix.FromFloat32ArrayToRefScaled(this._vrDevice.stageParameters.sittingToStandingTransform, 0, 1, this._standingMatrix);
+                        if (!this.getScene().useRightHandedSystem) {
+                            [2, 6, 8, 9, 14].forEach((num) => {
+                                if (this._standingMatrix) {
+                                    this._standingMatrix.m[num] *= -1;
+                                }
+                            });
+                        }
+                        this._useStandingMatrixCallback(true)
```
RaananW: Not sure I see the added value in calling the callback at all, but I understand this is needed since the callback needs to be called somewhere :)

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -175,26 +194,10 @@ module BABYLON {
...
-                                    this._standingMatrix.m[num] *= -1;
-                                }
-                            });
-                        }
-                        callback(true)
-                    }
-                })
+            this._useStandingMatrix = true;
+            this._useStandingMatrixCallback = callback;
+            if (!this._vrDevice) {
```
RaananW: if this._vrDevice already exists, the callback should be executed and the standing matrix should be initialized (unless it is auto-initialized like in my first comment)

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -175,26 +194,10 @@ module BABYLON {
         }
 
         public useStandingMatrix = (callback = (bool: boolean) => { }) => {
```
RaananW: is there a way to disable this? maybe make the flag public (or its getter and setter) and have this function simply change it? this way people won't really use this function and it can be deprecated.

src/Cameras/VR/babylon.webVRCamera.ts
RaananW: (I mean the entire scope after "if(this._useStandingMatrix){" should be executed

src/Cameras/VR/babylon.webVRCamera.ts
TrevorDev: The callback allows the user to tell if the vr device they are using supports the standing matrix. It currently only works when using Vive/Oculus in firefox or chrome. I agree, adding a class variable to store the callback is somewhat odd. I think I want to keep the callback in the useStandingMatrix function scope.

src/Cameras/VR/babylon.webVRCamera.ts
TrevorDev: A function with a callback was used instead of  a flag so the developer could wait until the standing matrix is set before trying to use it.

TrevorDev:
@RaananW  is this change any better, as it removes the direct webVR api calls? I think the way I was changing it to use a global callback was making it worse. Feel free to make a separate PR for this if you were thinking differently.

deltakosh:
All checks are good but there are conflicts :)

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -174,35 +295,53 @@ module BABYLON {
...
+                }
+                if(webVRInitobserver){
+                    webVRInitobserver.removeCallback(setStandingMatrix)
+                }
+            }
+
+            if(this._vrDevice){
+                setStandingMatrix();
+            }else{
+                webVRInitobserver = this.getEngine().initWebVR();
```
RaananW: I think you shouldn't init WebVR here (it is being done in the constructor already).
You can use the observable `engine.onVRDisplayChangedObservable`, which is a wonderful solution.

TrevorDev:
@RaananW  The changes I tried introduced some other bugs such as when calling useStandingMatrix multiple times. Additionally I found that the
engine.initWebVR().add(callback)
does not seem to call the callback until an additional call to initWebVR.

I ended up reverting the refactor as it works as expected now and this PR should just be for the doc changes to get them merged. Feel free to modify it if there's an cleaner change that I'm still missing.

PR #3590, opened 2018-01-19
Add suggested assetContainer.moveAllFromScene method

TrevorDev:
Helps make it easier to unload a scene to an assetContainer.

See
http://www.html5gamedevs.com/topic/35120-how-to-createadd-assets-directly-to-the-assetcontainer/

Thanks to @hololite on the forum for the code

- [x] Still need to add comments to file
- [x] Add unit test


src/babylon.assetContainer.ts
```diff
@@ -1,29 +1,109 @@
 module BABYLON {
+    /**
+     * Set of assets to keep when moving a scene into an asset container.
+     */
+    export class KeepAssets {
```
deltakosh: Not sure to understand why not lights and other asset types?

src/babylon.assetContainer.ts
```diff
@@ -115,5 +199,50 @@ module BABYLON {
...
+    
+                if (move) {
+                    targetAssets.push(asset);
+                }
+            }
+        }
+
+        /**
+         * Removes all the assets contained in the scene and adds them to the container.
+         * @param keepAssets Set of assets keep in the scene. (default: empty)
```
deltakosh: kept ? (I'm sure I'm wrong :))

TrevorDev: Will fix: Set of assets to keep in the scene. (default: empty)

src/babylon.assetContainer.ts
TrevorDev: I missed this. I will add the others and add the missing explicit public to there definitions,

PR #3594, opened 2018-01-20
refactor glTF serializer to use glTF interface declaration.

kcoley:


loaders/src/glTF/2.0/babylon.glTFLoaderInterfaces.ts
```diff
@@ -284,7 +284,7 @@ module BABYLON.GLTF2 {
         texCoord?: number;
     }
 
-    export interface IGLTF extends IGLTFProperty {
+    export interface _IGLTF extends IGLTFProperty {
```
deltakosh: Why the _ ?

kcoley: This is an intermittent change so the glTF interface @bghgary is using does not stomp on the `babylon.glTFInterface.d.ts` declaration file in the same module namespace.

deltakosh: Makes sense

PR #3608, opened 2018-01-23
VRHelper - Notify unselected mesh when ray predicate doesn't hit

CarlosLanderas:
Right now the observable onSelectedMeshUnselected is notified when there is a ray hit and the target mesh is the floor (teleport zone), or when there is a hit and the picked mesh is diferent to the previous selected and is not meshSelectionPredicate compliant.

However, if our raySelectionPredicate is very restricted and the ray hits an object making it the current and no other meshes in the predicate are close, we will never get notified about the gaze/controller losing the focus on that particular mesh.


deltakosh:
Please update what's new :)

CarlosLanderas:
Done :)

PR #3615, opened 2018-01-24
Update babylon.node.ts

pbalmasov:
Return animatable for beginAnimation method

### DIFF
--- a/src/babylon.node.ts
+++ b/src/babylon.node.ts
@@ -411,7 +411,16 @@
             return this._ranges[name];
         }
 
-        public beginAnimation(name: string, loop?: boolean, speedRatio?: number, onAnimationEnd?: () => void): Animatable {
+        /**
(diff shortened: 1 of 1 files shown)


src/babylon.node.ts
```diff
@@ -418,7 +418,7 @@
                 return;
             }
 
-            this._scene.beginAnimation(this, range.from, range.to, loop, speedRatio, onAnimationEnd);
+            return this._scene.beginAnimation(this, range.from, range.to, loop, speedRatio, onAnimationEnd);
```
RaananW: you will need to change the function's signature for that to work correctly.
see - https://travis-ci.org/BabylonJS/Babylon.js/builds/332791097

deltakosh:
Up?

pbalmasov:
Fixed

PR #3618, opened 2018-01-24
Notify with promises

RaananW:


deltakosh:
nuwatz :)

src/Tools/babylon.observable.ts
```diff
@@ -203,6 +203,59 @@
...
+            state.skipNextObservers = false;
+
+            // execute one callback after another (not using Promise.all, the order is important)
+            this._observers.forEach(obs => {
+                if (state.skipNextObservers) {
+                    return;
+                }
+                if (obs.mask & mask) {
+                    if (obs.scope) {
+                        // TODO - I can add the variable from the last function here. Requires changing callback sig
```
deltakosh: let's do it then, no?

RaananW: ah, yes :)
Gimme like 2.5 minutes.

deltakosh: I like the german precision :)

PR #3620, opened 2018-01-24
Projection Texture added to newest coset

lostink:
Done. Several lines seems to have changed nothing but still caught by github. Conflict history in babylonjs.materialhelper.ts. Need double check on that one.


deltakosh:
This looks good!
Now you need to fix comments check

sebavan:
Hi the only issue left is:
 [17:54:29] computeTextureMatrix should be Underscore Camel Case and it will be good to merge

Issue #2506, opened 2017-07-11
Add DOF (using kernel Blur) to default pipeline and create an autonomous DoF postprocess


deltakosh:

TrevorDev:
From review with Seb:
- [ ] add utility class for dof
- [ ] inherit from BlurPP instead of adding into blur
- [ ] try to get scene instead of this.getEngine().scenes[0].activeCamera;
- [ ] when using uniform near/far use float2 for the params
- [ ] check for 0 when devide by sum of weights, if so set to 0.0001 or something to make it work
- [ ] use mix for linear interp in shader to replace (original * (1.0-coc))+(blurred*coc);
- [ ] note double check the near + (far - near)*depth; might not work on certain types of camera. See shadow shader.
- [ ] doublecheck focallength 50, fstop: 4 1 meter away should be barely any blur
- [ ] validation
    - [ ]     blur
    - [ ]     dof 4 cameras different views with different focus distance, focal in one page (simple scene 3 sphere 1 plane)

deltakosh:
Please add a demo to the feature demos :) (and mention it in the what;s new)

PR #3623, opened 2018-01-25
Remove postProcessRenderPass and properly use post process render effect

TrevorDev:
Warning! This removes some public methods.
https://github.com/BabylonJS/Babylon.js/issues/2506


src/PostProcess/RenderPipeline/babylon.postProcessRenderEffect.ts
```diff
@@ -27,59 +26,28 @@ module BABYLON {
...
+                for(var ppIndex in this._postProcesses[index]){
+                    if (!this._postProcesses[index][ppIndex].isSupported) {
+                        return false;
+                    }
                 }
             }
             return true;
         }
 
         public _update(): void {
```
TrevorDev: This method is now empty. There are a couple places where this was called but it had no effect in the past as nothing was using postProcessRenderPass. Potentially logic will want to be added here in the future so I left it in.

PR #3626, opened 2018-01-25
SPS particle rotation matrix storage

jbousquie:
SPS internal modification : 
This feature now stores a simplified representation of each particle rotation matrix.
* This improves the SPS former behavior : when setting `sps.computeParticleRotation`to false, now the last rotation status of all the particles is kept.  Old discussion here : http://www.html5gamedevs.com/topic/33038-setting-material-on-sps-causes-webgl-error/?tab=comments#comment-189720
* This also opens the way to implement later (soon, I hope) the particle parenting feature.  

The cost of this feature is 9 more floats stored per created particle and 9 simple affectations per particle in the global loop.

_... and glad to be back in the contribution process_  :-)

### DIFF
--- a/src/Particles/babylon.solidParticle.ts
+++ b/src/Particles/babylon.solidParticle.ts
@@ -84,6 +84,10 @@ module BABYLON {
          * Still set as invisible in order to skip useless computations (Internal use)
          */
         public _stillInvisible: boolean = false;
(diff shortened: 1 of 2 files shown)


RaananW:
Welcome back :smile:

PR #3630, opened 2018-01-25
Remove render effect as pass and get pp

TrevorDev:
Adjusting based on Davids comments that were made after https://github.com/BabylonJS/Babylon.js/pull/3623 was merged.
- not sure we need the get process function anymore (at least not like it is now). We just need an accessor to a list of Pp
- RenderEffectAsPass can also be removed


src/PostProcess/RenderPipeline/babylon.postProcessRenderEffect.ts
```diff
@@ -1,34 +1,46 @@
...
+         * @param name The name of the effect
+         * @param getPostProcesses A function that returns a set of post processes which the effect will run in order to be run.
+         * @param singleInstance False if this post process can be run on multiple cameras. (default: true)
+         */
+        constructor(engine: Engine, name: string, getPostProcesses: () => Nullable<PostProcess | Array<PostProcess>>, singleInstance?: boolean) {
             this._name = name;
             this._singleInstance = singleInstance || true;
 
-            this._getPostProcess = getPostProcess;
+            this._getPostProcesses = getPostProcesses;
```
deltakosh: Do we need that?
 single array should be enough?

TrevorDev: Previously _postProcesses was an array because if you apply a PP to multiple cameras names 1,2 and 3 _postProcesses = {1:pp, 2:pp, 3:pp} now because I had to extend it to support multiple post processes in the effect it will be _postProcesses = {1:[pp1, pp2,...], 2:[pp1, pp2,...], 3:[pp1, pp2,...]}

deltakosh: makes sense

PR #3638, opened 2018-01-26
bug fixes for glTF serializer

kcoley:
- Fixed issue with bufferview creation and index referencing for multiple submeshes.
- fixed bug in logic with min and max calculation for left handed and right handed position coordinates
- Fixed glb chunk byte alignment for images that are not multiples of four.


kcoley:
@deltakosh The glTF serializer is already listed in the what's new file.

PR #3641, opened 2018-01-27
Set the xyz values directly

Poolminer:


deltakosh:
This was done to let you reuse left or right as result

PR #3647, opened 2018-01-29
More post process docs

TrevorDev:
Note the one potentially breaking change: Rename outputTexture to InputTexture in postprocess.ts Line 185. It looks like nothing uses that field in the playground but it is still a public field that I renamed.


deltakosh:
There are some conflicts :(

PR #3674, opened 2018-01-31
glTF loader updates and related changes

bghgary:
What's new
- Use promises instead of callbacks for all asynchronous operations in glTF loader.
- Compile all materials in parallel with downloading external resources in glTF loader (more work still needed for bones and morph targets).
- Runtime interfaces are now separated from the glTF interfaces in glTF loader.
- Add promise-based async functions to various places (e.g. SceneLoader functions, Scene.whenReadyAsync, forceCompilationAsync, etc.)
- Reorganize PBR material shader compilation such that it no longer waits for textures to complete before compiling shaders.
- Add checks to VertexData.merge to ensure data is valid before merging.
- Add Deferred wrapper class for promise with external resolve and reject.
- Add a bunch of unit tests for the glTF loader and reorganized the files a bit.

Bug Fixes
- Fix Promise error handling to reject with the exception instead of the exception message.

Breaking changes
- VertexData.merge no longer supports merging of data that do not have the same set of attributes.


loaders/src/glTF/2.0/Extensions/KHR_lights.ts
```diff
@@ -1,119 +1,107 @@
...
-        // Runtime values
-        index: number;
-    }
+    // https://github.com/MiiBond/glTF/tree/khr_lights_v1/extensions/Khronos/KHR_lights
+
+    const NAME = "KHR_lights";
 
-    interface IKHRLights {
-        lights: IGLTFLight[];
+    enum LightType {
```
sebavan: Should the type not be in the gltf typings in case the serializer wants to emit the extension information?

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
```diff
@@ -1,119 +1,107 @@
...
-                    const lightColor = lightInfo.color ? lightInfo.color : [1, 1, 1];
-                    loader._babylonScene.ambientColor.copyFromFloats(lightColor[0], lightColor[1], lightColor[2]);
+                const light = GLTFLoader._GetProperty(context, this._lights, extension.light);
+                if (light.type !== LightType.AMBIENT) {
+                    throw new Error(context + ": Only ambient lights are allowed on a scene");
                 }
-                
-                onComplete();
+
+                this._loader._babylonScene.ambientColor = light.color ? Color3.FromArray(light.color) : Color3.White();
```
sebavan: I guess default should be black to not add ambient or the entire scene would be white.

loaders/src/glTF/2.0/Extensions/KHR_materials_pbrSpecularGlossiness.ts
```diff
@@ -1,59 +1,85 @@
 /// <reference path="../../../../../dist/preview release/babylon.d.ts"/>
 
 module BABYLON.GLTF2.Extensions {
+    // https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness
+
+    const NAME = "KHR_materials_pbrSpecularGlossiness";
+
     interface IKHRMaterialsPbrSpecularGlossiness {
```
sebavan: I am also wondering if this should not be part of the typings (other places might be nice to move).

loaders/src/glTF/2.0/babylon.glTFLoader.ts
```diff
@@ -209,25 +184,113 @@ module BABYLON.GLTF2 {
...
+        private _loadNodesAsync(nodes: ILoaderNode[], ): Promise<void> {
+            const promises = new Array<Promise<void>>();
+
+            for (let node of nodes) {
+                promises.push(this._loadNodeAsync("#/nodes/" + node._index, node));
+            }
+
+            promises.push(this._loadAnimationsAsync());
+
+            return Promise.all(promises).then(() => {});
```
sebavan: Do we need the empty lambda could we not return the promise.all directly ?

src/Materials/PBR/babylon.pbrBaseMaterial.ts
```diff
@@ -686,15 +686,303 @@
                 subMesh._materialDefines = new PBRMaterialDefines();
             }
 
-            var scene = this.getScene();
-            var defines = <PBRMaterialDefines>subMesh._materialDefines;
+            const scene = this.getScene();
+            const engine = scene.getEngine();
```
sebavan: Could be move back after the check (agree it is preoptimization but happen per submesh per frame)

src/Materials/PBR/babylon.pbrBaseMaterial.ts
```diff
@@ -686,15 +686,303 @@
...
+
+            if (defines._areImageProcessingDirty) {
+                if (!this._imageProcessingConfiguration.isReady()) {
+                    return false;
+                }
+            }
+
+            if (!engine.getCaps().standardDerivatives) {
+                let bufferMesh = null;
+                if (mesh instanceof InstancedMesh) {
```
sebavan: Could we only check a property of InstancedMesh vs instanceof ? (optim)

src/Materials/PBR/babylon.pbrBaseMaterial.ts
```diff
@@ -686,15 +686,303 @@
...
+                    bufferMesh = mesh as Mesh;
+                }
+
+                if (bufferMesh && bufferMesh.geometry && bufferMesh.geometry.isReady() && !bufferMesh.geometry.isVerticesDataPresent(VertexBuffer.NormalKind)) {
+                    bufferMesh.createNormals(true);
+                    Tools.Warn("PBRMaterial: Normals have been created for the mesh: " + bufferMesh.name);
+                }
+            }
+
+            var onCompiled = (effect: Effect) => {
```
sebavan: could this be a private on the class to prevent the allocation ?

src/Materials/babylon.material.ts
```diff
@@ -1069,14 +1069,28 @@
...
         }
 
+        /**
+         * Force shader compilation.
+         * @param mesh The mesh that will use this material
+         * @param options Additional options for compiling the shaders
+         * @returns A promise that resolves when the compilation completes
+         */
+        public forceCompilationAsync(mesh: AbstractMesh, options?: Partial<{ clipPlane: boolean }>): Promise<void> {
+            return new Promise(resolve => {
```
sebavan: Should we check  compilation error ?

loaders/src/OBJ/babylon.objFileLoader.ts
```diff
@@ -264,14 +264,11 @@ module BABYLON {
             return this.importMesh(null, scene, data, rootUrl, null, null, null);
         }
 
-        public loadAssets(scene: Scene, data: string, rootUrl: string, onError?: (message: string, exception?: any) => void):Nullable<AssetContainer>{
+        public loadAssetContainer(scene: Scene, data: string, rootUrl: string, onError?: (message: string, exception?: any) => void): AssetContainer {
```
deltakosh: Why this change?

bghgary: The name change is because it was inconsistent with what the SceneLoader called it. The return value change is because there is no reason why it should return null.

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
bghgary: Good point, though right now, this is a private interface. We should do this when it becomes public.

loaders/src/glTF/2.0/Extensions/KHR_lights.ts
bghgary: Oops, yeah, that was an error. Will fix.

loaders/src/glTF/2.0/Extensions/KHR_materials_pbrSpecularGlossiness.ts
bghgary: Same answer as the other. I will make a pass at some point to expose these interfaces.

loaders/src/glTF/2.0/babylon.glTFLoader.ts
bghgary: I researched this a bit and didn't find any good answers. If I don't do this, it won't compile since the return value is expected to be Promise<void>. Promise.all returns Promise<T[]>

loaders/src/OBJ/babylon.objFileLoader.ts
deltakosh: ok!

src/Materials/PBR/babylon.pbrBaseMaterial.ts
bghgary: Yup.

src/Materials/PBR/babylon.pbrBaseMaterial.ts
bghgary: I'll check, but this code is identical to before.

bghgary: I can use getClassName() === "InstancedMesh"?

deltakosh: yep!

src/Materials/PBR/babylon.pbrBaseMaterial.ts
bghgary: Yes, but not sure it's worth it. This should only happen when something new needs to compile which should be rare. Plus, it is identical to what it was doing before.

src/Materials/babylon.material.ts
bghgary: Not sure. It didn't before. If we think this should happen, let's do it in a separate change.

bghgary:
> Please update this page with some async examples

Will do.

loaders/src/OBJ/babylon.objFileLoader.ts
TrevorDev: 👍

src/Materials/PBR/babylon.pbrBaseMaterial.ts
bghgary: It appears the extra code in the onCompiled callback was unnecessary, so now we just pass this.onCompiled directly.

PR #3675, opened 2018-01-31
missing cross origin setting

DavidHGillen:
necessary for the https hosting used on playground demos and many other places.

### DIFF
--- a/src/Helpers/babylon.videoDome.ts
+++ b/src/Helpers/babylon.videoDome.ts
@@ -44,7 +44,7 @@ module BABYLON {
          * @param options **loop=true** Automatically loop video on end.
          * @param options **size=1000** Physical radius to create the dome at, defaults to approximately half the far clip plane.
          */
-        constructor(name: string, urlsOrVideo: string[] | HTMLVideoElement, options: {
+        constructor(name: string, urlsOrVideo: string | string[] | HTMLVideoElement, options: {
             resolution?: number,
             clickToPlay?: boolean,
             autoPlay?: boolean,
@@ -75,7 +75,6 @@ module BABYLON {
             // configure material
             texture.coordinatesMode = BABYLON.Texture.FIXED_EQUIRECTANGULAR_MIRRORED_MODE; // matches orientation
             texture.wrapV = Texture.CLAMP_ADDRESSMODE; // always clamp the up/down
-            texture.video.crossOrigin = "anonymous";
             material.reflectionTexture = this._videoTexture;
             material.useEquirectangularFOV = true;
             material.fovMultiplier = 1.0;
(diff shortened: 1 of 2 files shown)


src/Helpers/babylon.videoDome.ts
```diff
@@ -75,6 +75,7 @@ module BABYLON {
             // configure material
             texture.coordinatesMode = BABYLON.Texture.FIXED_EQUIRECTANGULAR_MIRRORED_MODE; // matches orientation
             texture.wrapV = Texture.CLAMP_ADDRESSMODE; // always clamp the up/down
+            texture.video.crossOrigin = "anonymous";
```
deltakosh: Please use texture.video.crossOrigin =Tools.CorsBehavior
Also I would prefer to update the videoTexture object directly

DavidHGillen: I'll make the first change but am unsure what you mean by the second.
"Also I would prefer to update the videoTexture object directly"

deltakosh:
I meant: we should do that directly inside the video texture class when instantiating the video element

DavidHGillen:
As an upgrade to VideoTexture itself, rather than specifically for VideoDome? I can make that change for you.

deltakosh:
Yes please

src/Materials/Textures/babylon.videoTexture.ts
```diff
@@ -84,6 +84,20 @@
             }
         }
 
+        private _setCORS(src: string, element: HTMLVideoElement): void {
```
deltakosh: You can use Tools.SetCordBehavior for that

src/Materials/Textures/babylon.videoTexture.ts
```diff
@@ -92,18 +92,21 @@
...
         };
     
         private _getVideo(src: string | string[] | HTMLVideoElement): HTMLVideoElement {
             if (src instanceof HTMLVideoElement) {
+                Tools.SetCorsBehavior(src.currentSrc, src);
                 return src;
             }
             const video: HTMLVideoElement = document.createElement("video");
             if (typeof src === "string") {
+				Tools.SetCorsBehavior(src, video);
```
deltakosh: tabbing seems off

DavidHGillen: sorry about that, just updated webstorm and apparently it decided this document full of spaces should use tabs for it's tabbing on those lines. Used to behave better than that, so I forgot to check. I'll fix this a quick second.

deltakosh:
Thanks I'll create a nightly to see if this fix the PG

deltakosh:
This seems not to work : https://www.babylonjs-playground.com/#1E9JQ8#2

DavidHGillen:
It has corrected the console errors, the playground environment is the only change when compared to my working versions (it works locally and hosted on my servers). I'll do what I can to track it down.

deltakosh:
Does it work in the local playground (when using the playground debug mode defined here: http://doc.babylonjs.com/how_to/how_to_start#debug) ?

DavidHGillen:
Figured out the issue, the camera is silently failing. Fixed version with
Free https://www.babylonjs-playground.com/#1E9JQ8#3
Arc https://www.babylonjs-playground.com/#1E9JQ8#4
any preference as to which one is pushed?

deltakosh:
Arc is perfect!

Issue #3640, opened 2018-01-27
Customize webVR gaze pointer


Bulisor:
It would be nice if we had the posibility to customize more the gaze pointer, or why not, to add our texture, or gui element. Thank you.

TrevorDev:
See PR to customize the gaze tracker by setting a mesh to be used. Is that along the lines what you are looking for?

deltakosh:
Clsoing for now. Please open a new one if @TrevorDev works was not what you needed

PR #3685, opened 2018-02-02
Customize gaze tracker

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/3640
See playground for usage after this is checked in: http://playground.babylonjs.com/#15A9BF#4


deltakosh:
@TrevorDev: Please update documentation page as well

Issue #3621, opened 2018-01-24
Add Async methods to sceneLoader and webVR classes


TrevorDev:
- [x] LoadAssetContainerAsync
- [x] AppendAsync
- [x] LoadAsync
- [x] ImportMeshAsync
- [x] useStandingMatrixAsync
- [x] engine.initWebVRAsync
- [x] add somes usage examples in the promises documentation

PR #3693, opened 2018-02-03
Web vr async methods

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/3621

Note: this will the intended changes done in this PR https://github.com/BabylonJS/Babylon.js/pull/3574


CarlosLanderas:
@TrevorDev I Answered on the other PR.

src/Engine/babylon.engine.ts
```diff
@@ -1784,37 +1784,58 @@
...
+                    };
+                    this._onVrDisplayPresentChange = () => {
+                        this._vrExclusivePointerMode = this._vrDisplay && this._vrDisplay.isPresenting;
+                    }
+                    window.addEventListener('vrdisplayconnect', this._onVrDisplayConnect);
+                    window.addEventListener('vrdisplaydisconnect', this._onVrDisplayDisconnect);
+                    window.addEventListener('vrdisplaypresentchange', this._onVrDisplayPresentChange);
+                }
+    
+                this._getVRDisplays(()=>{
```
RaananW: Engine.prototype._getVRDisplays can return a promise as well (since it is using the native getVRDisplays, which is promise-based). then you don't need the entire Promise-constructor

src/Engine/babylon.engine.ts
```diff
@@ -1784,37 +1784,58 @@
...
-                    notifyObservers();
-                };
-                this._onVrDisplayPresentChange = () => {
-                    this._vrExclusivePointerMode = this._vrDisplay && this._vrDisplay.isPresenting;
+        /**
+         * Initializes a webVR display and starts listening to display change events.
+         * The onVRDisplayChangedObservable will be notified upon these changes.
+         * @returns A promise containing a VRDisplay and if vr is supported.
+         */
+        public initWebVRAsync(): Promise<{ vrDisplay: Nullable<any>, vrSupported: boolean }> {
```
RaananW: if you store the return-value of this function (for example engine.webVRInitPromise) you could always "then" it. After calling initWebVRAsync, you could use `engine.webVRInitPromise.then(() => {})` anywhere in your code. This way you make sure not to init webvr twice. you can also return this promise from this function if it already exists. Something along the line of:
```javascript
this.webVRInitPromise = this.webVRInitPromise || new Promise.....
return this.webVRInitPromise
```

CarlosLanderas:
I do like the proposal of changing the defaultHeight and standingMatrix the way it used to be and adding the distance to the ground scaling parameter we discussed here:
[Scaling parameter](https://github.com/BabylonJS/Babylon.js/pull/3574#issuecomment-363172921)

src/Engine/babylon.engine.ts
TrevorDev: 👍

src/Engine/babylon.engine.ts
TrevorDev: I'm not sure doing this is desired for situations when headsets are disconnected and connected so Ill leave this out for now.

TrevorDev:
Validated current code working as expect in sponza, hill valley and mansion scenes.

deltakosh:
Just waiting for @RaananW feedback before merging

src/Engine/babylon.engine.ts
RaananW: This should be controlled by the engine. In that case, you can store a rejected promise in this variable, and the developer will still be able to get the right state.

TrevorDev: Did I do this as expected in this commit? https://github.com/BabylonJS/Babylon.js/pull/3693/commits/7763d07f85d6d6e772877f17c7c45169f990f838

PR #3694, opened 2018-02-03
attach customization for standardMaterial

NasimiAsl:
please review parts
without serialized future for new parts

### DIFF
--- a/src/Materials/babylon.standardMaterial.ts
+++ b/src/Materials/babylon.standardMaterial.ts
@@ -106,8 +106,212 @@ module BABYLON {
             }
         }
     }
+    
+     export class StandarMaterialShaderBaseFragmentTerms{
+
+        public worldPosition  = "vPositionW"; 
+        public position  = "localPosition"; 
+        public worldNormal  = "vNormalW"; 
+        public normal  = "localNormal"; 
+        public uv  = "vDiffuseUV";  
+        public fragColor  = "color";  
+        public diffuse   = "diffuseColor";  
+        public diffuseTexture  = "baseColor"; 
+        public finalDiffuse  = "finalDiffuse";     
+        public finalSpecular  = "finalSpecular";     
+        public alpha  = "alpha";     
+        public ambientColor  = "baseAmbientColor";     
+        public reflect  = "reflectionColor";  
+        /* terms for replace shader part  */
+        public definiation  = "#define Custom_Fragment_Definiation";
+        public beforeLightMap  = "#define Custom_Fragment_Before_LightMap"; 
+        public beforeMain  = "#define Custom_Fragment_Before_Main"; 
+        public beginMain  = "#define Custom_Fragment_Begin_Main"; 
+        public updateDiffuseColor  = "#define Custom_Fragment_Update_Diffuse_Color"; 
+        public updateDiffuseTexture  = "#define Custom_Fragment_Update_Diffuse_Texture";  
+        public updateAlpha = "#define Custom_Fragment_Update_Alpha"; 
+        public beforeFragColor  = "#define Custom_Fragment_Before_Frag_Color";   
+    }
+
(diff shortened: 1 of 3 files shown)


deltakosh:
My main feedback would be to not touch the std material.
Can you instead create a new class that inherits from the std material? 
Obviously we can keep the defines in the shader (can you use uppercase like #define BEFOREMAIN)

NasimiAsl:
so where we can make it in materialLib  or extensions?
too as you see i change some parameter of default shader like vNormalW => vNormalW_helper
you wanna keep that ? or not 
witch feedback is your mind ? because it is wrong system desinged
class a  = stdMat
class b = cusMat
resource  R = defShd
a depend R & R depend a
b depend R & R depend b
b had a properties 
a has not b properties (special parameters )
some one make that property in **a** and that used in **b** class for some user ( it is **b** benefit to make any thing we don't have in **a** ) so
when you change  in (special parameters )  area  and you have changes in R you have problem
* that is happen in custom material when i start first version (that inherited directly from standard material )  so how i fix that i make a stable version of stdmat and cut dependency
so that is exactly that situation again

it is exactly what we have in custom material (the first one) we r in cycles

in my experience that  part is not make standard material complicated than now
just make it flexible so i don't see any benefit for separated that but i see a danger area again when we separate them

NasimiAsl:
this is my personal idea some percent that have change to be correct
but i think (and saw) when we separate stdmat and cusmat 
the cusmat cant be supportable because developer change the stdmat and dont notic what problem 
most be created in cusmat because thay don't see that exist  (both depend to one shader)

deltakosh:
I want to keep it as an independent class because the stdmat is already too complex. As the main maintainer of the framework this is my job to keep it under control:)

I would like to keep the shaders unchanged (just add anchors) and also I would like the CustomMaterial to be in the material library where it belongs

But I will also give you my word that I will fix any issue you may have if I break something that impacts the CustomMaterial (as long as it inherits from StdMaterial)

NasimiAsl:
look the custom material history 
https://github.com/BabylonJS/Babylon.js/commit/05d5b2b5ec67c93ee361ee3474c7a5894bd7101a#diff-1ae8612ccf7d0a25b0732557bdbefc25
we have this after 7 day  you  move isReady  function from standardmaterial  and fix custom material too 
but custom material work on that method 

that not help and i think the babylon changes is not have fixed interface in each version
mean if we have some method in interface and we know that never changed we can make independent class and we can support all old version too still you can attach a lot method but never you have complex code 

that is where we most think to how much version we gonna go if we wanna make 5 6 version 
we need some other material type so i think the custom material can do her job for now
and we can think about new material system that have all PBR and others

  am i right ?

deltakosh:
Yes I remember and I still think it is the good solution :)
Keeping a copy of the StdMaterial is not something valid on the long term. 
The changes about isReady was a big one for perf reason. I do not plan to change the way StdMaterial work anymore.

deltakosh:
In other term: I can guarantee a fixed interface for StdMaterial for now

deltakosh:
Other idea: now that we have a lot of shader code available as includes, why not creating a CustomMaterial with its own shader? The shader could reuse all includes from PBR or Stdmat but not the TS code. This way you will be really autonomous

NasimiAsl:
i need think about it more

deltakosh:
No worry my friend:)
And if this helps we can determine together if some additional shader code needs to be move to includes in order for you to write the CustomMaterial shader. 
This way we reuse the same shader code but you are free and not dependent on the StdMaterial TS code...win/win :)

NasimiAsl:
i have a idea ( maybe make new project in babylon for that :D support me )

deltakosh:
CLosing this one for now. Will continue the discussion on the new PR ;)

PR #3695, opened 2018-02-04
Fix memory leaks related to the glTF loader

bghgary:
There are probably some more to fix, but this should fix most of the cases.

### DIFF
--- a/src/Tools/babylon.promise.ts
+++ b/src/Tools/babylon.promise.ts
@@ -22,36 +22,6 @@ module BABYLON {
         private _onRejected?: (reason: any) => void;
         private _rejectWasConsumed = false;
 
-        public get state(): PromiseStates {
-            return this._state;
-        }
-
-        public get isFulfilled(): boolean {
-            return this._state === PromiseStates.Fulfilled;
-        }
-
-        public get isRejected(): boolean {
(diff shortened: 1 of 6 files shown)


bghgary:
I searched a bit and it doesn't sound like there is a way to check for memory leaks. Any suggestions on how to automate testing this would be great.

src/Tools/babylon.promise.ts
```diff
@@ -22,36 +22,6 @@ module BABYLON {
         private _onRejected?: (reason: any) => void;
         private _rejectWasConsumed = false;
 
-        public get state(): PromiseStates {
```
deltakosh: Why removing them?

bghgary: Not sure why they are needed as public. It is not part of the promise contract right?

deltakosh: Right. I used them mostly for debugging purpose. We can remove them no problem 
I was more asking to understand if it was related to memory leak

bghgary: Yeah, I figured it was something like that. I removed them to see if they were the cause of any issues. I can put them back if you want.

deltakosh: Nope this is fine like that

PR #3696, opened 2018-02-05
Fix a couple of promise bugs

bghgary:


tests/unit/babylon/src/Tools/babylon.promise.tests.ts
```diff
@@ -45,13 +45,16 @@ describe('Babylon.Promise', function () {
             }).catch(() => {
                 tempString += " to check promises";
             }).then(() => {
-                expect(tempString).to.eq("Initial message to check promises");
-                done();
+                try {
+                    expect(tempString).to.eq("Initial message to check promises");
+                    done();
+                }
+                catch (error) {
```
deltakosh: using a catch will not prevent error detection?
Why adding catch everywhere in these tests?

bghgary: Async tests using done must call done with the error or the test will timeout.

bghgary: Timeout when there is an error that is.

deltakosh: ok :) thanks for the info

bghgary: That's why promises are better for async unit tests, but we are testing promises in this test :)

PR #3711, opened 2018-02-06
Fixed #3689

### LINKED ISSUE #3689: Fix fontSize percentage support

JeanPhilippeKernel:


JeanPhilippeKernel:
To convert px ---> % we always calculate any pixels size with 16.
So 8px --> 50 %
    125% ---> 20px
and so on..

deltakosh:
we cannot change the way percentage are compiled as this is used everywhere
We need to fix it at fontsize level

JeanPhilippeKernel:
Oh :(  you right !! so I move it in control with some parsing stuff

deltakosh:
Here is a repro case: https://www.babylonjs-playground.com/#XCPP9Y#421
The text should be really large

deltakosh:
Closing for now ;)

PR #3719, opened 2018-02-07
Revert valueAndUnit, and Fixed bug #3689

JeanPhilippeKernel:
I think it should be good now :)

### DIFF
--- a/gui/src/controls/control.ts
+++ b/gui/src/controls/control.ts
@@ -290,10 +290,9 @@ module BABYLON.GUI {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


deltakosh:
Thanks a lot for that :)
But the problem was quite a bit different: https://github.com/BabylonJS/Babylon.js/commit/6539ac6eefce2fe0ee84ab86b00eb46fe650a87a

It was because the fontSize was not correctly referencing parent container

JeanPhilippeKernel:
ohh ! , cool so :)

PR #3732, opened 2018-02-09
update doc part

Vinc3r:
I've made a mystake on my previous commit about coordinatesMode & wrapU comments: I've forgotten to start with /** and my comments was below the code instead of above, sorry for that! (anyway, this was not visible in the classes doc)

### DIFF
--- a/src/Materials/Textures/babylon.baseTexture.ts
+++ b/src/Materials/Textures/babylon.baseTexture.ts
@@ -28,23 +28,8 @@
 
         @serialize()
         public coordinatesIndex = 0;
-
-        @serialize("coordinatesMode")
-        private _coordinatesMode = Texture.EXPLICIT_MODE;
-        public set coordinatesMode(value: number) {
-            if (this._coordinatesMode === value) {
-                return;
-            }
-            this._coordinatesMode = value;
-            if (this._scene) {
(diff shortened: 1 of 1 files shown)


src/Materials/Textures/babylon.baseTexture.ts
```diff
@@ -60,17 +45,30 @@
...
+            if (this._scene) {
+                this._scene.markAllMaterialsAsDirty(Material.TextureDirtyFlag);
+            }
+        }
+        public get coordinatesMode(): number {
+            return this._coordinatesMode;
+        } 
         
-        /* 
+        /**
```
deltakosh: Can you duplicate it for wrapv and wraps?

src/Materials/Textures/babylon.baseTexture.ts
```diff
@@ -28,23 +28,8 @@
...
-            if (this._scene) {
-                this._scene.markAllMaterialsAsDirty(Material.TextureDirtyFlag);
-            }
-        }
-        public get coordinatesMode(): number {
-            return this._coordinatesMode;
-        }
         
-        /*
+        /**
```
deltakosh: This Doc has to before the public set line

PR #3736, opened 2018-02-09
Gpu particles - Do not merge - ongoing

deltakosh:


deltakosh:
Pinging @benaadams as I changed where the divisor is stored (moving from Buffer to VB)

benaadams:
should work... I think

Issue #3748, opened 2018-02-12
Pressing 'mainButton in VR while using only the left controller permanently removes laserpointer


Exolun:
I noticed an issue while using only the left controller pressing the mainButton (back button on Windows MR controller) causes the laser to disappear.  

Normally this button will switch the laser between left/right hand, or when using just the right hand controller it switches the laser off/on - but if you are only using a left handed controller it will turn off the laser permanently.

Expected Result:  Pressing mainButton should have the same behavior as the right-hand contoller when used on its own.

Current Result:  Laser disappears permanently if using a left handed controller, but toggles on/off if using just a right handed one.

Steps to reproduce:
1.) Open playground
2.) Turn on your left controller
3.) Open in VR
4.) Press 'mainButton' button on the controller

Bug repro on [WebVR Tutorial Playground](https://www.babylonjs-playground.com/#IHY77X#53)

Issue #3653, opened 2018-01-30
Add flag(s) to enable both laserPointers while in VR


Exolun:
When in VR mode, the VRHelper class allows you to turn on a laser pointer for controller interactions using the pointer ray.

It's very helpful, but it would be nice if we had the option to enable both laser pointers for scenarios where the user might want to interact using both hands at once.

deltakosh:
Good idea

PR #3761, opened 2018-02-14
Dual vr controllers gui support

TrevorDev:
PR contains:
- Ability to use multiple vr controllers concurrently on separate gui elements (http://localhost:1338/Playground/index-local.html#ZI9AK7#214)
- Changed behavior so dual controllers are always used (I'm hoping the previous behavior is not desired by existing apps)
- Created new internal VRExperienceHelperGazer class to store gaze information of each controller
- Changes to pointer simulation events to add pointerId
- Removed more pointer singleton logic from control, advancedDynamicTexture classes
- BREAKING CHANGE to onCanvasPointerOutObservable in engine to pass pointerEvent instead of engine in the observable callback. (engine parameter hopefully isnt used in any public apps so I think this is ok?)

Future potential tasks:
- Let multiple curses work on color picker element
- Refactor more webVR helper controller logic into the VRExperienceHelperGazer class

https://github.com/BabylonJS/Babylon.js/issues/3748
https://github.com/BabylonJS/Babylon.js/issues/3653


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -27,6 +27,145 @@ module BABYLON {
         createFallbackVRDeviceOrientationFreeCamera?: boolean;
     }
 
+    class VRExperienceHelperGazer implements IDisposable {
```
TrevorDev: Is this how to correctly create an internal class that should not be accessible publicly?

src/Engine/babylon.engine.ts
```diff
@@ -613,7 +613,7 @@
         /**
          * Observable event triggered each time the canvas receives pointerout event
          */
-        public onCanvasPointerOutObservable = new Observable<Engine>();
+        public onCanvasPointerOutObservable = new Observable<PointerEvent>();
```
deltakosh: I'm fine with the change. Please just add it in the breaking changes list in the what'snew.md

src/babylon.scene.ts
```diff
@@ -1294,9 +1295,11 @@
         /**
          * Use this method to simulate a pointer down on a mesh
          * The pickResult parameter can be obtained from a scene.pick or scene.pickWithRay
+         * @param pickResult pickingInfo of the object wished to simulate pointer event on
+         * @param pointerEventInit pointer event state to be used when simulating the pointer event (eg. pointer id for multitouch)
          */
-        public simulatePointerDown(pickResult: PickingInfo): Scene {
-            let evt = new PointerEvent("pointerdown");
+        public simulatePointerDown(pickResult: PickingInfo, pointerEventInit: PointerEventInit | undefined): Scene {
```
RaananW: What is the difference between `pointerEventInit: PointerEventInit | undefined` and pointerEventInit?: PointerEventInit`

davrous:
Code looks good to me, I knew enabling dual interactions would imply a lot of work, this is confirmed :)

What do you mean by: "Changed behavior so dual controllers are always used" ?

src/babylon.scene.ts
TrevorDev: Looks to work the same (no idea why I wrote it that way) so I switched it to ? to be consistent.

src/Engine/babylon.engine.ts
TrevorDev: added

TrevorDev:
@davrous "Changed behavior so dual controllers are always used"

Currently the behavior is that one controller is activated when you enter vr and toggling laser will disable the other hands laser if its currently on. After this change both controllers are activated by default and toggling laser on one hand has no effect on the other hand.

RaananW:
I am fixing the netlify build error right now

PR #3770, opened 2018-02-15
loading files using Babylon's load file

RaananW:
The loading processes will be aborted is viewer is disposed.
Fixing https://github.com/BabylonJS/Babylon.js/issues/3671


deltakosh:
You are on fire!

PR #3773, opened 2018-02-15
Update Draco compression to use web workers

bghgary:
@deltakosh Please review and see if you like the way I'm using web workers and how the external Draco script location is determined.


deltakosh:
Please just update the waht's new

bghgary:
I wasn't thinking I needed to update what's new since I already added the Draco support part. Did you want me to add web worker specifically to it?

deltakosh:
This is more for the WorkerPool object actually
This is a great feature and it will also require a tiny page in the doc to explain how to use it

bghgary:
Ahh, okay. Sure, I'll update the what's new and docs.

deltakosh:
Much appreciated :)

Issue #3764, opened 2018-02-15
scene.enableDepthRenderer only uses scene.activeCamera


TrevorDev:
See: http://playground.babylonjs.com/#IDSQK2#22 

In scenes with multiple cameras, separate depth maps should be able to be created by specifying a camera.
Additionally, currently the resulting depth map is squashed to the size of the camera's view port which is not expected.

- [x]  Support depth map for multi camera
- [x]  Ignore viewport for depthmap
- [ ]  Update dof effect to specify camera when using depth renderer
- [ ]  Update validation to use multicamera
- [ ]  Create validation test that uses multicamera
- [ ]  Docs

Related; https://github.com/BabylonJS/Babylon.js/issues/2506

deltakosh:
Just add "this._depthMap.ignoreCameraViewport = true;" inside DepthRenderer constructor to ignore camera viewport

PR #3775, opened 2018-02-16
updated dof validation to use multiple cameras, added support for dof…

TrevorDev:
…/circle of confusion to specify a camera to be used

https://github.com/BabylonJS/Babylon.js/issues/3764


src/PostProcess/babylon.circleOfConfusionPostProcess.ts
```diff
@@ -23,7 +23,7 @@ module BABYLON {
         /**
          * Creates a new instance of @see CircleOfConfusionPostProcess
          * @param name The name of the effect.
-         * @param scene The scene the effect belongs to.
+         * @param minMaxZCamera The camera to be used to calculate the min/max z for the depth map
```
TrevorDev: Now this constructor takes 2 camera parameters which is kind of weird, but the second one must be null when used in the pipeline so I think this needs to be done.

src/PostProcess/RenderPipeline/Pipelines/babylon.defaultRenderingPipeline.ts
```diff
@@ -286,7 +286,7 @@
             this._reset();
 
             if(this.depthOfFieldEnabled){
-                this.depthOfField = new DepthOfFieldEffect(this._scene, this._depthOfFieldBlurLevel, this._defaultPipelineTextureType);
+                this.depthOfField = new DepthOfFieldEffect(this._scene, this._cameras[0], this._depthOfFieldBlurLevel, this._defaultPipelineTextureType);
```
deltakosh: I think we should  not force cameras[0] but use the pipeline mechanism about active cameras

TrevorDev: If the scene has multiple cameras rendering at the same time, I dont want the second camera to get the firsts depth map.

src/PostProcess/babylon.circleOfConfusionPostProcess.ts
deltakosh: I would prefer saving min and max as property of the depth texture as we know them when we create the depth texture

TrevorDev: Sounds good, changed circle of confusion and depth of field to take cameraMinZ:number, cameraMaxZ:number, depthTexture: RenderTargetTexture as parameters instead

src/PostProcess/babylon.circleOfConfusionPostProcess.ts
```diff
@@ -32,7 +33,7 @@ module BABYLON {
          * @param reusable If the post process can be reused on the same frame. (default: false)
          * @param textureType Type of textures used when performing the post process. (default: 0)
          */
-        constructor(name: string, scene: Scene, depthTexture: RenderTargetTexture, options: number | PostProcessOptions, camera: Nullable<Camera>, samplingMode?: number, engine?: Engine, reusable?: boolean, textureType: number = Engine.TEXTURETYPE_UNSIGNED_INT) {
+        constructor(name: string, cameraMinZ:number, cameraMaxZ:number, depthTexture: RenderTargetTexture, options: number | PostProcessOptions, camera: Nullable<Camera>, samplingMode?: number, engine?: Engine, reusable?: boolean, textureType: number = Engine.TEXTURETYPE_UNSIGNED_INT) {
```
deltakosh: Actually I meant to literally read these values on the depthTexture with the depthTexture.camera. So not even need to give these values to the constructor

Issue #3712, opened 2018-02-06
WebVRExperienceHelpepr controllers disappear with large positionScale parameter


TrevorDev:
See http://playground.babylonjs.com#IPSSKX#1 

Headset position looks to be working but I believe the controllers position/scale do not get updated to match.

PR #3781, opened 2018-02-16
Scale vr controllers by deviceScale when it is set in VRExperienceHelper

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/3712

Tested here: http://localhost:1338/Playground/index-local.html#IPSSKX#5


deltakosh:
Pinging @raananw

PR #3783, opened 2018-02-19
add basic Daydream controller support

brianzinn:
I have comments on how to improve both this and GearVR.  Will create a topic in forum once approved.


brianzinn:
Sorry, I did not run "gulp typedoc-all" before the commit.  Will make that part of checklist.

Looks like only code changes have a stronger validation?  ie: handleButtonChange "should be Underscore Camel Case", while that method is "protected abstract handleButtonChange" from webVRController (and works on gearVR added last month).  I have a feeling there will be cascading changes, so let me know if you wanted me to take care of that. Thanks.

deltakosh:
Please go ahead. This is a necessary change:)

deltakosh:
Also can you add an entry in the what;s new ?

PR #3785, opened 2018-02-19
iOS 10's video tags are in BGR instead of RGB

RaananW:
This solves it with a flag in the background material (and a small change in the background shader).

See - http://www.html5gamedevs.com/topic/27723-ios-video-texture-display-issues/?tab=comments#comment-205081 for a discussion about it.

It is the developer's responsibility to set this flag (as I am not sure if it is all iOS 10, or only specific versions).

Note that this is just a suggestion for a solution, if there is a cleaner/better way please comment :smile:


deltakosh:
IOS...

PR #3801, opened 2018-02-21
Adding Flag for Diffuse Texture Alpha

Pryme8:
Hopefully adding support to obj import mtl parsing to support a flag for diffuse texture having an alpha pass.
"bjs_diffuseHasAlpha 1"

### DIFF
--- a/loaders/src/OBJ/babylon.objFileLoader.ts
+++ b/loaders/src/OBJ/babylon.objFileLoader.ts
@@ -121,7 +121,14 @@ module BABYLON {
                 } else if (key === "map_d" && material) {
                     // The dissolve of the material
                     material.opacityTexture = MTLFileLoader._getTexture(rootUrl, value, scene);
-
+                } else if (key == 'bjs_diffuseHasAlpha' && ((material) && material.diffuseTexture)){
+                    //Alpha on diffuse Texture  
+                    if(value === '1'){
+                        material.diffuseTexture.hasAlpha = true;   
+                    }else{
+                        material.diffuseTexture.hasAlpha = false;
+                    }
+                }
 
                     //Options for illumination
                 } else if (key === "illum") {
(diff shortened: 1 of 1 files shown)


loaders/src/OBJ/babylon.objFileLoader.ts
```diff
@@ -121,7 +121,14 @@ module BABYLON {
                 } else if (key === "map_d" && material) {
                     // The dissolve of the material
                     material.opacityTexture = MTLFileLoader._getTexture(rootUrl, value, scene);
-
+                } else if (key == 'bjs_diffuseHasAlpha' && ((material) && material.diffuseTexture)){
```
deltakosh: So you want to go with a custom parameter? This does not seem scalable

Pryme8: Let me review the mtl file structure and and see if there is a parameter that covers the diffuse having an alpha.  I'm pretty sure there is not, so my thought would be you could edit your mtl after export.

What thoughts do you have?

deltakosh: I get your point but this is kind of not convenient
I would perhaps prefer loading the texture and trying to check if there are alpha data

But your solution is also acceptable

Pryme8: roger that! Let me whip up a method for that, then maybe we can compare and then get input from others?

deltakosh:
Excellent idea!

RaananW:
Hi @Pryme8 ,
Care to update what's new with the change? You can add it to this PR. Please don'T forget to update the documentation site (if not already done).
Thanks!!!

Pryme8:
I have no way of testing this and have just kinda given up on the idea... so y'all can delete this PR. *sigh* sorry.

PR #3807, opened 2018-02-23
Fix bug with loading skeletons in glTF loader

### LINKED ISSUE #3805: Animations do not start correctly since alpha 8
http://www.html5gamedevs.com/topic/35862-gltf-animation-is-rendered-weirdly-since-alpha8/

bghgary:
Fixes #3805 and adds a unit test.


bghgary:
The build failure is due to "what's new" which doesn't need to be updated for this bug fix.

PR #3815, opened 2018-02-24
ideal size - always take smallest

royibernthal:
I added to AdvancedDynamicTexture a boolean useSmallestIdeal, if window width is smaller than height, using idealWidth, otherwise idealHeight. I'm now working on being able to match the game to all possible screen sizes so it should actually be a quite useful feature, it looks good that way.

I had problems compiling the source as you can see in the forum, but it should work.

### DIFF
--- a/gui/src/advancedDynamicTexture.ts
+++ b/gui/src/advancedDynamicTexture.ts
@@ -28,6 +28,7 @@ module BABYLON.GUI {
         private _fullscreenViewport = new Viewport(0, 0, 1, 1);
         private _idealWidth = 0;
         private _idealHeight = 0;
+        private _useSmallestIdeal: boolean = false;
         private _renderAtIdealSize = false;
         private _focusedControl: Nullable<IFocusableControl>;
         private _blockNextFocusCheck = false;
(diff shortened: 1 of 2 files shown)


gui/src/valueAndUnit.ts
```diff
@@ -31,14 +31,12 @@ module BABYLON.GUI {
 
         public getValue(host: AdvancedDynamicTexture): number {
             if (host && !this.ignoreAdaptiveScaling && this.unit !== ValueAndUnit.UNITMODE_PERCENTAGE) {
+                var width: number = (this._value * host.getSize().width) / host.idealWidth;
```
deltakosh: What if ideal width is undefined?

gui/src/advancedDynamicTexture.ts
```diff
@@ -88,6 +89,20 @@ module BABYLON.GUI {
...
+        public get useSmallestIdeal(): boolean {
+            return this._useSmallestIdeal;
+        }
+
+        public set useSmallestIdeal(value: boolean) {
+            if (this._useSmallestIdeal === value) {
+                return;
+            }
+
+            this._useSmallestIdeal = value;
```
deltakosh: Do you mind using line returns and brackets to make it easier to read

royibernthal: Feels a bit cluttered to me but I guess something like that?

PR #3831, opened 2018-02-26
add chromatic aberration to default pipeline

TrevorDev:


dist/preview release/typedocValidationBaseline.json
```diff
@@ -8797,6 +8797,11 @@
             "NotCamelCase": true
           }
         },
+        "ChromaticAberrationPostProcessId": {
```
TrevorDev: @deltakosh any thoughts on what should be done about these ID's in the defaultRenderPipeline class? None of them use cammel case. Should all of them stay ignored or could they all be renamed?

PR #3833, opened 2018-02-27
Master

MackeyK24:
Animation Properties Override Fixes


deltakosh:
Yeah I see what you need here.
I'll do the fix

deltakosh:
Please try with the latest version
should be good now

PR #3841, opened 2018-02-27
add grain to default pipeline

TrevorDev:
image

- [x] Add ignore to cammelcase issue
- [x] Make dither take amount method, add cos gain part directly in shader
- [x] Use cos instead of softmax
- [x] Change name of controls grainEnabled, grainIntensity, grainAnimated
- [x] move + 1 for animation to outside shader

### DIFF
--- a/dist/preview release/typedocValidationBaseline.json
+++ b/dist/preview release/typedocValidationBaseline.json
@@ -38929,7 +38929,7 @@
             "MissingText": true
           }
         },
-        "DITHERING": {
+        "GRAIN": {
           "Naming": {
             "NotCamelCase": true
           }
--- a/dist/preview release/typedocValidationBaseline.json
+++ b/dist/preview release/typedocValidationBaseline.json
@@ -38928,6 +38928,11 @@
           "Comments": {
             "MissingText": true
           }
+        },
+        "DITHERING": {
+          "Naming": {
+            "NotCamelCase": true
+          }
         }
       }
     },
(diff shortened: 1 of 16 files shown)


dist/preview release/typedocValidationBaseline.json
```diff
@@ -38928,6 +38928,11 @@
           "Comments": {
             "MissingText": true
           }
+        },
+        "GRAIN": {
```
TrevorDev: Camel case error I cant fix to keep consistency again.

src/Shaders/ShadersInclude/helperFunctions.fx
```diff
@@ -70,9 +70,9 @@ float getRand(vec2 seed) {
 	return fract(sin(dot(seed.xy ,vec2(12.9898,78.233))) * 43758.5453);
 }
 
-vec3 dither(vec2 seed, vec3 color) {
+float dither(vec2 seed, vec3 color, float varianceAmount) {
```
sebavan: seems color is not needed anymore ???

sebavan:
Only one tiny comments and it is in :-)

src/Materials/babylon.imageProcessingConfiguration.ts
```diff
@@ -2,6 +2,7 @@ module BABYLON {
     /**
      * Interface to follow in your material defines to integrate easily the
      * Image proccessing functions.
+     * @ignore
```
TrevorDev: @deltakosh do you have any opinion on how to handle these validation errors in the future. Currently we have been ignoring them as they are already all uppercase and modifying them to cammelcase to pass validation might brake back compat,

deltakosh: We will ignore like that:)
All defines do not need doc as they are internal tools mostly

PR #3842, opened 2018-02-27
URL / URL.createObjectURL is missing in some environments

RaananW:
This will initialize the worker blob URL only if URL and URL.createObjectURL exist


sebavan:
Would be good to check in the constructor if the creation is possible as well ???

RaananW:
Agreed, added the init in the constructor instead of a private static, and showing an error in case it is not supported.

PR #3851, opened 2018-03-01
Computing good normals for fur material, as vNormalW is use only for …

julien-moreau:
…lighting computation


julien-moreau:
Related to https://github.com/BabylonJS/Exporters/pull/96

julien-moreau:
It looks like procedural textures and instances scenes make the tests failing
I thought it was only visual check

deltakosh:
The build is crashing because you did not update the what's new ;)

julien-moreau:
WIll know for the next time ^^

PR #3867, opened 2018-03-03
Fix node dispose bug

bghgary:
* Add dispose parameters to all relevant node-derived classes and node itself.
* Pass dispose parameters when recursing into children.
* Add unit tests.


bghgary:
I'm assuming node-derived non-mesh objects like light, bone, etc. cannot have node children.

deltakosh:
Light and camera can actually

bghgary:
Okay, then we should pass the parameters to those objects too.

PR #3881, opened 2018-03-06
set _pointingPoseNode to null when not found in loaded mesh

TrevorDev:
Fixed issue described here:
http://www.html5gamedevs.com/topic/36082-how-do-you-teleport-in-vivesteamvr/

Works now with vive/firefox


sebavan:
Sweet Thanks

PR #3889, opened 2018-03-06
Delay default pipeline shader compile

TrevorDev:


src/PostProcess/babylon.blurPostProcess.ts
```diff
@@ -69,7 +74,21 @@
...
+		/**
+         * Updates the effect with the current post process compile time values and recompiles the shader.
+         * @param defines Define statements that should be added at the beginning of the shader. (default: null)
+         * @param uniforms Set of uniform variables that will be passed to the shader. (default: null)
+         * @param samplers Set of Texture2D variables that will be passed to the shader. (default: null)
+         * @param indexParameters The index parameters to be used for babylons include syntax "#include<kernelBlurVaryingDeclaration>[0..varyingCount]". (default: undefined) See usage in babylon.blurPostProcess.ts and kernelBlur.vertex.fx
+         * @param onCompiled Called when the shader has been compiled.
+         * @param onError Called if there is an error when compiling a shader.
+         */
+        public updateEffect(defines: Nullable<string> = null, uniforms: Nullable<string[]> = null, samplers: Nullable<string[]> = null, indexParameters?: any,
```
TrevorDev: @deltakosh Could you double check that overriding this method is sane here. It seemed to me that _updateParameters below does what I expect this would do as it calls super.updateEffect at the end of it.

deltakosh: This is the right place!

PR #3903, opened 2018-03-08
Update water.vertex.fx

dad72:
Fixed a bug where the water rises to make objects below the water and goes down again to make objects above the water.

What I delete is useless and allows the water to stay at this position on the Y axis

Post here : http://www.html5gamedevs.com/topic/36197-watermaterial-level-water-up-or-down/?tab=comments#comment-207516

Fixe test here (Empty the caches to see the corectif) : https://www.babylonjs-playground.com/#1SLLOJ#439

### DIFF
--- a/materialsLibrary/src/water/water.vertex.fx
+++ b/materialsLibrary/src/water/water.vertex.fx
@@ -126,10 +126,6 @@ void main(void) {
 #endif
 
 	vec3 p = position;
-	float newY = (sin(((p.x / 0.05) + time * waveSpeed)) * waveHeight * windDirection.x * 5.0)
-			   + (cos(((p.z / 0.05) +  time * waveSpeed)) * waveHeight * windDirection.y * 5.0);
-	p.y += abs(newY);
-	
 	gl_Position = viewProjection * finalWorld * vec4(p, 1.0);
 
 #ifdef REFLECTION


deltakosh:
Pinging @luaacro for review

dad72:
It corrects the bug well, the water stays at this place (0, 0, 0)  and the waves are always present.

julien-moreau:
Not sure this is the way we should avoid waves.
Setting waveHeight to 0 will remove the artifacts

Can you add a define when wageHeight != 0 then keep the waves function instead ?

julien-moreau:
The waves you are talking about are rendering effects, but we would like to also keep the physical waves (computed by the lines you removed)

dad72:
I do not speak of the waves, I speak of the terrain that goes up and down.

Looking at the subject, Pryme has put pictures in focus that shows the problem.:
http://www.html5gamedevs.com/topic/36197-watermaterial-level-water-up-or-down/?tab=comments#comment-207516

If I place a land at xyz 0, 0, 0, and add water, it becomes xyz 0, 1, 0

julien-moreau:
like:
#ifdef WAVEHEIGHT
compute waves (newY)
#endif

And in the TypeScript:
if (this.waveHeight !== 0) { defines.WAVEHEIGHT = true; }

julien-moreau:
And it is fixed when setting waveHeight to 0 ?
https://www.babylonjs-playground.com/#1SLLOJ#440

dad72:
Yes indeed. I think that should be the default value then no?

dad72:
Ok, I understand. I will do like that. I close this PR.
Sorry for inconvenience. I really thought that this pitch that goes up is decent was not normal.

julien-moreau:
Unfortunately it is ^^
Imagine a lake with waves and your camera is flying above

julien-moreau:
Anyway thanks for reporting :)

dad72:
Yes but imagine a boat floating, it can be above the water at a time or under water with the current default value.

PR #3918, opened 2018-03-09
setTextureFromPostProcessOutput method to allow post process output to be able to be set as input to a shader

TrevorDev:
Looking to get feedback on the current direction:
- Add outputTexture field to post process pointing to where the post process has drawn (typically a texture in the next postprocess in the chain)
- Add setTextureFromPostProcessOutput method to allow post process output to be able to be set as input to a shader. The existing setTextureFromPostProcess will continue to set the texture corresponding to input texture for the passed in post process.
- Refactored dof effect to use above functions and get rid of a pass

Still under investigation:
- Figure out a way to share textures between post processes. shareOutputWith is not working for some scenarios.

https://github.com/BabylonJS/Babylon.js/issues/3921


src/PostProcess/babylon.postProcessManager.ts
```diff
@@ -147,8 +147,10 @@
             var engine = this._scene.getEngine();
 
             for (var index = 0, len = postProcesses.length; index < len; index++) {
+                var pp = postProcesses[index];
+
                 if (index < len - 1) {
-                    postProcesses[index + 1].activate(camera, targetTexture);
+                    pp._outputTexture = postProcesses[index + 1].activate(camera, targetTexture);
```
deltakosh: Only one comment here: making sure Internal texture.dispose will scan all scenes and all pp to remove link to it

src/Engine/babylon.engine.ts
```diff
@@ -5205,6 +5205,9 @@
         }
 
         public setTextureFromPostProcessOutput(channel: number, postProcess: Nullable<PostProcess>): void {
+            if(postProcess._outputTexture._disposed){
```
deltakosh: this could be late to the party
I would prefer doing it in the internalTexture dispose to do it asap

src/PostProcess/babylon.postProcessManager.ts
TrevorDev: ok, added to that logic to src/Engine/babylon.engine.ts release method. Let me know if that is how you were thinking.

deltakosh:
Perfect!

PR #3922, opened 2018-03-10
Add support for primitive modes to glTF loader

bghgary:


src/Mesh/babylon.subMesh.ts
```diff
@@ -247,6 +247,22 @@
         public intersects(ray: Ray, positions: Vector3[], indices: IndicesArray, fastCheck?: boolean): Nullable<IntersectionInfo> {
             var intersectInfo: Nullable<IntersectionInfo> = null;
```
deltakosh: I don't think we should take that in account. The way meshes are drawn should not impact geometry intersections.

bghgary: It's not that simple. The drawMode changes the way indices are stored. There is also the case where there are no indices. My change will make it such that `intersects` will return null with mesh where fillMode is not triangles. Without this change will cause an exception. Making it work will require quite a bit of code.

deltakosh: Makes sense!

src/Mesh/babylon.subMesh.ts
```diff
@@ -247,6 +247,22 @@
...
+            }
+
+            switch (material.fillMode) {
+                case Material.PointListDrawMode:
+                case Material.LineListDrawMode:
+                case Material.LineLoopDrawMode:
+                case Material.LineStripDrawMode:
+                case Material.TriangleFanDrawMode:
+                case Material.TriangleStripDrawMode:
+                    // REVIEW: support draw modes?
```
deltakosh: please just remove the comment

bghgary: done

PR #3937, opened 2018-03-13
Fix the issue that reported by http://www.html5gamedevs.com/topic/361…

xuchen-shadow:
…29-dracocompression-cause-error-when-importing-babylon-in-a-worker/?tab=comments#comment-208128


deltakosh:
Good catch!

PR #3938, opened 2018-03-13
Cannon and Oimo (and future dependencies) can now be optional

RaananW:
It is now possible to declare optional dependencies that can but not must be declared.
webpack (if used) will throw a warning that can be ignored is using externals.


RaananW:
Please don't merge, I still need to test all scenarios.

RaananW:
The current problem with merging is the playground, which has an AMD loader that only supports part of the AMD specification. Will find a solution for this

PR #3941, opened 2018-03-13
Fix bug with default draw mode in glTF loader

bghgary:
Some day I will stop making this error...


bghgary:
We have validation tests coming from the glTF asset generator so it will be caught next time.

PR #3980, opened 2018-03-19
removing the (old) reference path

RaananW:
This will force defining dependencies, but will make the declaration files more modular.

What it means is - A developer will be forced to include babylonjs in order to use babylonjs-loaders (it is actually a dependency), but would be able to use babylonjs-loaders in a native typescript project without a problem.

To make it more clear:

```
// this was valid until now, but only worked with specific project types
import * as loaders from "babylonjs-loaders"
//notice that "babylonjs" is not being loaded
```

From now it will need to look like this (which is anyhow recommended)
```
import * as BABYLON from "babylonjs";
import from "babylonjs-loaders";
```


deltakosh:
can you make sure to document it?
thanks a lot

RaananW:
This is how it is already documented, but you still had the freedom (https://www.npmjs.com/package/babylonjs-loaders)

deltakosh:
looks good to me then!

RaananW:
I will merge this later today

PR #3983, opened 2018-03-19
Added disableBlending() to Skeleton

FunFetched:


src/Bones/babylon.skeleton.ts
```diff
@@ -356,6 +356,14 @@
             });
         }
 
+        public disableBlending() {
```
deltakosh: Do using this (http://doc.babylonjs.com/babylon101/animations#overriding-properties) not enough?

FunFetched: Hmm... I missed this. I suppose that would do the trick, though I noticed that BABLON.Animatable has both an enabling and disabling function for its blending properties, so it seemed to make sense to have both here as well. I'll close this.

deltakosh:
Thanks :)

PR #3994, opened 2018-03-21
lightmap in linear space

Vinc3r:
in the same way as [linearSpace for PBR shader](https://github.com/BabylonJS/Babylon.js/commit/fdc34d17c9dce6753507d6e0028cfd9d50844b03), default should have this patch too.


Vinc3r:
ok, I'll have to remember that :)

PR #3996, opened 2018-03-21
Added Comments to DynamicTexture

BabylonJSGuide:
As I am learning as I am doing please give a thorough check before merging.
 I have added comments to this file /Babylon.js/blob/master/src/Materials/Textures/babylon.dynamicTexture.ts hope it was the correct one. 

In particular have I used @ignore correctly and in the correct places?
Also I have marked invertY as a default of true but it appears not to have a set default even though it behaves as true if you do not set it. Is there another way I should be describing it?

I was using this file as a guide - /Babylon.js/blob/master/src/Tools/babylon.assetsManager.ts and I think I found a typo please check my change is OK.

Any other issues please let me know.

### DIFF
--- a/src/Materials/Textures/babylon.dynamicTexture.ts
+++ b/src/Materials/Textures/babylon.dynamicTexture.ts
@@ -1,9 +1,23 @@
 ﻿module BABYLON {
+    /**
+     * A class extending {BABYLON.Texture} allowing drawing on a texture
+     * @see http://doc.babylonjs.com/how_to/dynamictexture
+     */
     export class DynamicTexture extends Texture {
         private _generateMipMaps: boolean;
         private _canvas: HTMLCanvasElement;
         private _context: CanvasRenderingContext2D;
         private _engine: Engine;
+
+        /**
+         * Creates a {BABYLON.DynamicTexture}
+         * @param name defines the name of the texture
+         * @param options provides 3 alternatives for width and height of texture, a canvas, object with width and height properties, number for both width and height
+         * @param scene defines the scene where you want the texture
+         * @param generateMipMaps defines the use of MinMaps or not (default is false)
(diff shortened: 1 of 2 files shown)


src/Materials/Textures/babylon.dynamicTexture.ts
```diff
@@ -35,6 +49,7 @@
             this._context = <CanvasRenderingContext2D>this._canvas.getContext("2d");
         }
 
+        /** @ignore */
```
deltakosh: AS this is public we cannot ignore it unfortunately

src/Materials/Textures/babylon.dynamicTexture.ts
```diff
@@ -66,19 +90,40 @@
             this._recreate(textureSize);
         }
 
+        /**
+         * Gets the context of the canvas used by the texture
+         */
```
deltakosh: This one needs a @returns

BabylonJSGuide:
If more changes required please let me know.

src/Materials/Textures/babylon.dynamicTexture.ts
```diff
@@ -35,6 +49,9 @@
             this._context = <CanvasRenderingContext2D>this._canvas.getContext("2d");
         }
 
+        /**
+         * Gets the current state of canRescale as true
```
deltakosh: It will not always be true as child classes can overload it and returns false

BabylonJSGuide: My lack of knowledge hits again. Will update.

deltakosh: Absolutely no problem :)

BabylonJSGuide:
If you think it is safe to let me loose on some more I noticed that many of files in the src/Actions folder need comments and would be happy to given them a go, with you reviewing them of course.

deltakosh:
Oh yeah please do :) this will help a lot!

PR #4003, opened 2018-03-22
AbstractMesh -> TransformNode inside babylon.transformNode.ts

bobalazek:


src/Mesh/babylon.transformNode.ts
```diff
@@ -26,7 +26,7 @@ module BABYLON {
         private _transformToBoneReferal: Nullable<TransformNode>;
 
         @serialize()
-        public billboardMode = AbstractMesh.BILLBOARDMODE_NONE;
+        public billboardMode = TransformNode.BILLBOARDMODE_NONE;
```
deltakosh: Good catch!

PR #4007, opened 2018-03-23
adding material alphaMode description

Vinc3r:


deltakosh:
Does it work with the new doc site?
Can't remember where you added those already

Vinc3r:
Yep, I check that before trying this patch :)
http://doc.babylonjs.com/classes/classes/babylon.basetexture.html#coordinatesmode from 
https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/Textures/babylon.baseTexture.ts#L35

(by the way I notice that anchor links doesn't works on Firefox when accessing through http://doc.babylonjs.com/api.html?basetexture urls - works in Chrome - for coordinatesMode anchor link above, I've made right-click > copy link to get it)

deltakosh:
Ok I'll check:)

deltakosh:
It works for me on firefox
do you have any error in the console?

Vinc3r:
last version of firefox (59.0.1 64b) & windows 10, newly created profile (so without addons), still the issue, and no console error (here a [video screen capture](https://i.imgur.com/vDGvDG8.mp4))

deltakosh:
Looks like a bug to me :) but in Firefox

PR #4008, opened 2018-03-23
Commented in two of the action files

BabylonJSGuide:
Thought I had better let you check before I did anymore commenting. Let me know if I have mislead people with any of my comments or any other editing needed. ActionManager.ts was already done but found a spelling error.  Now having seen the scale of the task it is a bit overwhelming. I'll plod on doing a bit here and there but it remains a drop in a very large ocean.

### DIFF
--- a/src/Actions/babylon.condition.ts
+++ b/src/Actions/babylon.condition.ts
@@ -1,6 +1,6 @@
 ﻿module BABYLON {
     /**
-     * A condition applied to an action 
+     * A Condition applied to an Action 
      */
     export class Condition {
         /**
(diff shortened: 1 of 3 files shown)


src/Actions/babylon.condition.ts
```diff
@@ -1,29 +1,67 @@
...
         public _evaluationId: number;
+
+        /**
+         * Internal only
+         * @ignore 
+         */
         public _currentResult: boolean;
 
+        /**
+         * Creates a new {BABYLON.Condition}
```
deltakosh: The {} thingy is not working. So we can just say Creates a new Condition

src/Actions/babylon.condition.ts
```diff
@@ -1,29 +1,67 @@
...
+         * @param actionManager  
+         */
         constructor(actionManager: ActionManager) {
             this._actionManager = actionManager;
         }
 
+        /**
+         * Gets the current state of isValid
```
deltakosh: Perhaps we could say something like Check if the current Condition is valid

BabylonJSGuide: Done

PR #4015, opened 2018-03-25
Return animation groups when calling SceneLoader.ImportMesh 

### LINKED ISSUE #3971: Expose animation groups in loader callback

### LINKED ISSUE #3999: Playground/scenes/Alien doesn't play its morph targets
- Bug repro on [sandbox](http://sandbox.babylonjs.com/):
Using [Alien](https://github.com/BabylonJS/Babylon.js/blob/master/Playground/scenes/Alien/Alien.gltf)
- Expected result:
Morph target and skeletal animation channels play.
- Current result:
Only skeletal animation channels play.

bghgary:
* Fixes #3971 
* Fixes #3999

Other changes
* Replace onAnimationGroupLoaded with returned animationGroups in glTFLoader.
* Add AnimationMultiTarget to optimize animations that target multiple targets.
* Update unit tests to check for morph targets, animations, and animation groups.
* Add animation groups and textures to asset container.


bghgary:
There seems to be a build order issue. I think the issue is that the viewer module depends on the loader module and they are currently building at the same time.

deltakosh:
There is a build error unfortunately

bghgary:
Yes, I'm not sure how to fix. See my previous comment.

deltakosh:
Pinging @raananw for ideas;)
I guess making sure viewer is built after loaders and bjs should make it

deltakosh:
I found a way to fix it. I'll merge tomorrow and do the fix:)

Thanks

Issue #3885, opened 2018-03-06
Add fullscreen button and switch camera button to inspector


deltakosh:
Add two new features to the inspector:
- [ ] A fullscreen button
- [ ] A dropdown list to switch between existing cameras

PR #4016, opened 2018-03-26
Inspector improvements

Bloadrick:
improvements for issue https://github.com/BabylonJS/Babylon.js/issues/3885

   Add a fullscreen button in the toolbar

   add in the scene tab radio buttons to choose the active camera

I'm waiting for @deltakosh approval for the active camera in the scene tab


deltakosh:
Why not using the camera tab?
I do think we should add cameras to a second tab while they are already on their own tab

PR #4021, opened 2018-03-26
Add support for KHR_materials_unlit to glTF loader

bghgary:
Bringing this back again. See https://github.com/KhronosGroup/glTF/pull/1163 for discussion.


bghgary:
Added a validation test.

PR #4026, opened 2018-03-27
Add support for normalized and non-float data to vertex buffers

bghgary:
Testing is covered by glTF primitive mode tests from asset generator.

### DIFF
--- a/src/Mesh/babylon.buffer.ts
+++ b/src/Mesh/babylon.buffer.ts
@@ -11,6 +11,16 @@
          */
         public readonly byteStride: number;
 
+        /**
+         * Constructor
+         * @param engine the engine
+         * @param data the data to use for this buffer
+         * @param updatable whether the data is updatable
+         * @param stride the stride (optional)
+         * @param postponeInternalCreation whether to postpone creating the internal WebGL buffer (optional)
+         * @param instanced whether the buffer is instanced (optional)
+         * @param useBytes set to true if the stride in in bytes (optional)
+         */
         constructor(engine: any, data: DataArray, updatable: boolean, stride = 0, postponeInternalCreation = false, instanced = false, useBytes = false) {
             if (engine instanceof Mesh) { // old versions of BABYLON.VertexBuffer accepted 'mesh' instead of 'engine'
                 this._engine = engine.getScene().getEngine();
@@ -96,6 +106,13 @@
             this.create(data);
         }
 
+        /**
+         * Updates the data directly.
+         * @param data the new data
+         * @param offset the new offset
+         * @param vertexCount the vertex count (optional)
+         * @param useBytes set to true if the offset is in bytes
+         */
         public updateDirectly(data: DataArray, offset: number, vertexCount?: number, useBytes = false): void {
             if (!this._buffer) {
                 return;
--- a/src/Mesh/babylon.buffer.ts
+++ b/src/Mesh/babylon.buffer.ts
@@ -72,6 +72,15 @@
             return this._buffer;
         }
 
+        /**
+         * Gets the stride in float32 units (i.e. byte stride / 4).
+         * May not be an integer if the byte stride is not divisible by 4.
+         * DEPRECATED. Use byteStride instead.
+         */
+        public getStrideSize(): number {
+            return this.byteStride / 4;
+        }
+
         // Methods
         public create(data: Nullable<DataArray> = null): void {
             if (!data && this._buffer) {
--- a/src/Mesh/babylon.buffer.ts
+++ b/src/Mesh/babylon.buffer.ts
@@ -2,12 +2,16 @@
     export class Buffer {
         private _engine: Engine;
         private _buffer: Nullable<WebGLBuffer>;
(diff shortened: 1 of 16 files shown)


src/Mesh/babylon.geometry.ts
```diff
@@ -296,18 +299,14 @@
...
-                var stride = vertexBuffer.getStrideSize();
-                this._totalVertices = data.length / stride;
-
-                this.updateBoundingInfo(updateExtends, data);
+                this._updateBoundingInfo(updateExtends, data);
             }
             this.notifyUpdate(kind);
         }
 
-        private updateBoundingInfo(updateExtends: boolean, data: Nullable<FloatArray>) {
```
deltakosh: Isn't it a breaking change?

src/Mesh/babylon.buffer.ts
```diff
@@ -34,49 +48,32 @@
...
-        public getData(): Nullable<FloatArray> {
+        public getData(): Nullable<DataArray> {
             return this._data;
         }
 
         public getBuffer(): Nullable<WebGLBuffer> {
             return this._buffer;
         }
 
-        public getStrideSize(): number {
```
bghgary: Wasn't sure what to do with this function. The best I can do is return byteStride / 4 if we want to continue to keep this function.

deltakosh: We need to keep it or else it is a breaking change

bghgary:
That's interesting. It doesn't fail locally.

bghgary:
Scratch that. It does fail locally.

src/Mesh/babylon.geometry.ts
```diff
@@ -233,12 +237,11 @@
             this._vertexBuffers[kind] = buffer;
 
             if (kind === VertexBuffer.PositionKind) {
-                var data = <FloatArray>buffer.getData();
-                var stride = buffer.getStrideSize();
-
-                this._totalVertices = data.length / stride;
+                if (totalVertices != null) {
+                    this._totalVertices = totalVertices;
```
benaadams: This should be documented as a **breaking change**.

If you set the vertex data with `setVerticesBuffer` rather than `setVerticesData` it now doesn't set `this._totalVertices` which means subMeshes don't get created; which means nothing draws.

Fix is to explicitly pass `totalVertices` into `setVerticesBuffer` when the vertex buffer is a `PositionKind` buffer; or pass `totalVertices` in for the `setIndices` call.

src/Mesh/babylon.buffer.ts
```diff
@@ -105,17 +112,24 @@
...
+         * @param useBytes set to true if the offset is in bytes
+         */
+        public updateDirectly(data: DataArray, offset: number, vertexCount?: number, useBytes = false): void {
             if (!this._buffer) {
                 return;
             }
 
             if (this._updatable) { // update buffer
-                this._engine.updateDynamicVertexBuffer(this._buffer, data, offset, (vertexCount ? vertexCount * this.getStrideSize() : undefined));
+                this._engine.updateDynamicVertexBuffer(this._buffer, data, useBytes ? offset : offset * 4, (vertexCount ? vertexCount * this.byteStride : undefined));
```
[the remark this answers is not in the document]
deltakosh: It is not a breaking change as it will behave like before if useBytes is not set
Do you think we should improve the comments?

src/Mesh/babylon.geometry.ts
bghgary: This isn't supposed to be a breaking change. I must have done something unintentional. Let me investigate.

benaadams: Fix https://github.com/BabylonJS/Babylon.js/pull/4067

src/Mesh/babylon.buffer.ts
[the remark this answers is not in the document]
a-carvallo: As of now, yes I think improving the comments would be better, unless passing `useBytes` in VertexBuffer's `updateDirectly` and Geometry's `updateVerticesDataDirectly`.

Also for `updateVerticesDataDirectly`, `data` has the type `Float32Array` but should probably be `DataArray`.

Shouldn't also `offset * 4` be replaced by `offset * this.byteStride` in Buffer's `updateDirectly` ?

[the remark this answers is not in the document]
deltakosh: I guess we could transmit it to all functions 
Fancy doing the PR?

src/Mesh/babylon.geometry.ts
bghgary:
```
                if (totalVertices == null) {
                    var data = buffer.getData()!;
                    var dataByteLength = (data instanceof Array) ? data.length * 4 : data.byteLength;
                    var bufferByteLength = dataByteLength - buffer.byteOffset;
                    this._totalVertices = bufferByteLength / buffer.byteStride;
                }
                else {
                    this._totalVertices = totalVertices;
                }
```

That's what I was going to do, but looks like you guys are on top of it.

PR #4028, opened 2018-03-27
Dof kernel fix

TrevorDev:
- Avoid multiple samples in coc fragment
- Ensure output width is used in blur calculation
- Use coc instead of depth when performing dof blur


src/PostProcess/babylon.blurPostProcess.ts
```diff
@@ -68,7 +68,11 @@
 			super(name, "kernelBlur", ["delta", "direction", "cameraMinMaxZ"], ["circleOfConfusionSampler"], options, camera, samplingMode, engine, reusable, null, textureType, "kernelBlur", {varyingCount: 0, depCount: 0}, true);
 			this._staticDefines = defines;
 			this.onApplyObservable.add((effect: Effect) => {
-				effect.setFloat2('delta', (1 / this.width) * this.direction.x, (1 / this.height) * this.direction.y);
+				if(this._outputTexture){
```
sebavan: Is this still needed ?

src/Shaders/ShadersInclude/kernelBlurFragment2.fx
```diff
@@ -1,6 +1,10 @@
 ﻿#ifdef DOF
-    sampleDepth = sampleDistance(sampleCoord{X});
-    factor = clamp(1.0-((centerSampleDepth - sampleDepth)/centerSampleDepth),0.0,1.0);
+    if(sampleCenter + delta * KERNEL_DEP_OFFSET{X} == sampleCenter){
```
sebavan: Could you use only delta compared to 0 to be more compilation friendly ?

src/PostProcess/babylon.blurPostProcess.ts
TrevorDev: yes

PR #4030, opened 2018-03-27
Beginning glow layer serialization

julien-moreau:
Need to check if "excludedMeshes" and "includedMeshes" must be deleted from the new GlowLayer instance due to the SerializationHelper.Parse(...):
delete gl.excludedMeshes;
delete gl.includedMeshes;

### DIFF
--- a/src/Layer/babylon.effectLayer.ts
+++ b/src/Layer/babylon.effectLayer.ts
@@ -664,8 +664,8 @@
         }
 
         /**
-          * Gets the class name of the effect layer.
-          * @returns The string with the class name of the effect layer.
+          * Gets the class name of the effect layer
+          * @returns the string with the class name of the effect layer
           */
          public getClassName(): string {
             return "EffectLayer";
--- a/src/Layer/babylon.effectLayer.ts
+++ b/src/Layer/babylon.effectLayer.ts
@@ -662,13 +662,5 @@
             this.onAfterComposeObservable.clear();
             this.onSizeChangedObservable.clear();
         }
-
-        /**
-         * Serializes this effect layer
-         * @returns - serialized effect layer object.
-         */
-        public serialize(): any {
-            return SerializationHelper.Serialize(this);
-        }
     }
 } 
--- a/src/Layer/babylon.effectLayer.ts
+++ b/src/Layer/babylon.effectLayer.ts
@@ -54,18 +54,27 @@
         protected _emissiveTextureAndColor: { texture: Nullable<BaseTexture>, color: Color4 } = { texture: null, color: new Color4() };
 
         /**
+         * The name of the layer
+         */
+        @serialize()
+        public name: string;
+
+        /**
          * The clear color of the texture used to generate the glow map.
          */
+        @serializeAsColor4()
         public neutralColor: Color4 = new Color4();
 
         /**
          * Specifies wether the highlight layer is enabled or not.
          */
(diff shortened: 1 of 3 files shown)


src/Layer/babylon.glowLayer.ts
```diff
@@ -115,7 +119,7 @@
          * @param scene The scene to use the layer in
          * @param options Sets of none mandatory options to use with the layer (see IGlowLayerOptions for more information)
          */
-        constructor(public name: string, scene: Scene, options?: Partial<IGlowLayerOptions>) {
+        constructor(name: string, scene: Scene, options?: Partial<IGlowLayerOptions>) {
```
deltakosh: This is a breaking change
You must comment it directly in the constructor

src/Layer/babylon.effectLayer.ts
```diff
@@ -651,5 +662,13 @@
             this.onAfterComposeObservable.clear();
             this.onSizeChangedObservable.clear();
         }
+
+        /**
+         * Serializes this effect layer
+         * @returns - serialized effect layer object.
```
deltakosh: Please do not use - but "a" in this case

src/Layer/babylon.glowLayer.ts
```diff
@@ -441,5 +445,70 @@
...
+                    }
+                }
+            }
+
+            return serializationObject;
+        }
+
+        /**
+         * Creates a Glow Layer from parsed glow layer data.
+         * @param parsedGlowLayer - Parsed glow layer data.
```
deltakosh: Please do not use "-" but "defines" in this case

src/Layer/babylon.glowLayer.ts
julien-moreau: Don't understand the breaking change :(
As we have "public name: string" in BABYLON.EffectLayer

deltakosh: Right I missed it :)

src/Layer/babylon.glowLayer.ts
```diff
@@ -441,5 +445,70 @@
...
+                    }
+                }
+            }
+
+            return serializationObject;
+        }
+
+        /**
+         * Creates a Glow Layer from parsed glow layer data.
+         * @param parsedGlowLayer Defines glow layer data.
```
deltakosh: no period at the end and no capital at the beginning (sorry :))

julien-moreau: No problems, pushing right now :)

src/Layer/babylon.glowLayer.ts
```diff
@@ -441,5 +445,78 @@
...
+                }
+            }
+
+            return serializationObject;
+        }
+
+        /**
+         * Creates a Glow Layer from parsed glow layer data
+         * @param parsedGlowLayer defines glow layer data
+         * @param scene bjs scene
```
deltakosh: defines

julien-moreau: Oh everywhere!

deltakosh: yes because of the way it is displayed in the doc :)

julien-moreau: That's what I finally guessed ^^

deltakosh:
Thanks!

Issue #3507, opened 2018-01-08
new Debug.AxisViewer (and others) cannot be used as Type


RaananW:
See inspector PhysicsViewer for example.

PR #4034, opened 2018-03-28
Revert debug-classification

RaananW:
Debug was reduced from module to class with static members to:
1) support es6 submodules
2) reduce the amount of submodules babylon has.

As we are delaying submodules, it is safe to revert this change, yet making it still es6-safe. Debug will be a variable in ES6 but its internal variables might not be accessible. Opening the debug layer will wrok in ES6, but its internal components cannot be imported individually.

Tested locally in es6 environment and using typescript compilation

Fixing - https://github.com/BabylonJS/Babylon.js/issues/3507


RaananW:
I will fix the typedoc error

RaananW:
I am not sure I understand the error:
`Module is not Upper Case Debug (id: 11976) babylon.d.ts:5524`
Debug is upper case. should it be fully upper-case?

deltakosh:
It should not:)
Pinging @sevan

RaananW:
I actually put it with ignoreNaming now, let's see

sebavan:
What convention do you want for namespace module ? I ll adapt the rule.

deltakosh:
Namespace / Module should be fine (like class names)

sebavan:
ok i ll add pascal case in the list of conventions for module/namespace.

RaananW:
Let me know when the rule updates, I will merge it here and remove the ignoreNaming

sebavan:
np, should be done in 20 minutes (in the bus atm)

RaananW:
no wifi in buses? pfff.....

sebavan:
:-) Should be all good.

Issue #4060, opened 2018-04-02
Linear/gamma conversion not working as expected when using a mirror and image processing as a post process


TrevorDev:
See: http://www.html5gamedevs.com/topic/36651-default-pipeline-depth-of-field-bloom-grain-chromatic-aberration-msaa/?do=findComment&comment=210463 

- Bug repro:
Here's a playground with a mirror and HDR set to "false" :
https://www.babylonjs-playground.com/#1YAIO7#33
=> reflection is fine
Same playground with HDR set to "true":
https://www.babylonjs-playground.com/#1YAIO7#34
=>lighting in the reflection is different (darker)
- Expected result: HDR true should work the same as HDR false
- Current result: HDR true produces a darker reflection caused by converting to linear color twice

PR #4063, opened 2018-04-02
convert back to gamma to avoid multiple toLinear conversions

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4060

Updated playground for testing: http://localhost:1338/Playground/index-local.html#1YAIO7#35
Had to modify texture type to halffloat to avoid banding:
`mirrorMaterial.reflectionTexture = new BABYLON.MirrorTexture("mirror", 1024, scene, true, BABYLON.Engine.TEXTURETYPE_HALF_FLOAT);`


src/Materials/Textures/babylon.mirrorTexture.ts
```diff
@@ -90,6 +90,7 @@
             this.ignoreCameraViewport = true;
 
             this.onBeforeRenderObservable.add(() => {
+                this.gammaSpace = !scene.imageProcessingConfiguration.isEnabled || !scene.imageProcessingConfiguration.applyByPostProcess;
```
TrevorDev: Double check that this is the correct place to set this? Should there be similar logic added to the refractionTexture?

src/Shaders/default.fragment.fx
```diff
@@ -304,13 +309,17 @@ void main(void) {
...
-#endif
+	#ifdef REFLECTIONMAP_PROJECTION
+		coords /= vReflectionUVW.z;
+	#endif
 
 	coords.y = 1.0 - coords.y;
-
-	reflectionColor = texture2D(reflection2DSampler, coords).rgb * vReflectionInfos.x;
+	#ifdef IS_REFLECTION_LINEAR
+		reflectionColor = toGammaSpace(texture2D(reflection2DSampler, coords).rgb);
```
TrevorDev: is the toGammaSpace also needed on the textureCube lookups above?

src/Materials/Textures/babylon.mirrorTexture.ts
sebavan: Could it be done on creation and in the onUpdateParameters event of the imageProcessingConfiguration ???

src/Shaders/default.fragment.fx
```diff
@@ -274,7 +274,12 @@ void main(void) {
 
 	refractionCoords.y = 1.0 - refractionCoords.y;
 
-	refractionColor = texture2D(refraction2DSampler, refractionCoords).rgb * vRefractionInfos.x;
```
sebavan: I guess we should do the same with the cube sampler a bit upper in the file.

sebavan:
adding @deltakosh for review.

PR #4064, opened 2018-04-02
Add support for Draco WebAssembly

bghgary:
* Add Draco WebAsssembly support
* Stop using web workers in Draco as it seems to slow things down
* glTF Loader: decode the same Draco buffer only once

The glTF Buggy with Draco sample model now loads in 1.36s down from 5.62s.


bghgary:
"What's new" failed in the build which is expected since this is not a new feature.

PR #4067, opened 2018-04-03
No breaking change in setVerticesBuffer

benaadams:
Set https://github.com/BabylonJS/Babylon.js/pull/4026/files#r178512410

### DIFF
--- a/src/Mesh/babylon.geometry.ts
+++ b/src/Mesh/babylon.geometry.ts
@@ -239,6 +239,11 @@
             if (kind === VertexBuffer.PositionKind) {
                 if (totalVertices != null) {
                     this._totalVertices = totalVertices;
+                } else {
+                    var data = <FloatArray>buffer.getData();
+                    if (data != null) {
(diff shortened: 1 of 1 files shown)


deltakosh:
Thanks a lot!

src/Mesh/babylon.geometry.ts
```diff
@@ -239,6 +239,11 @@
             if (kind === VertexBuffer.PositionKind) {
                 if (totalVertices != null) {
                     this._totalVertices = totalVertices;
+                } else {
+                    var data = <FloatArray>buffer.getData();
+                    if (data != null) {
+                        this._totalVertices = data.length / (buffer.byteStride * 4);
```
deltakosh: it is actually / 4 :)
I'll fix it

deltakosh: Lol

benaadams: https://github.com/BabylonJS/Babylon.js/pull/4069

PR #4074, opened 2018-04-04
Pass useBytes in updateVerticesDataDirectly in Geometry & VertexBuffer

a-carvallo:
Related to : https://github.com/BabylonJS/Babylon.js/pull/4026#pullrequestreview-108884133

### DIFF
--- a/src/Mesh/babylon.buffer.ts
+++ b/src/Mesh/babylon.buffer.ts
@@ -129,7 +129,7 @@
             }
 
             if (this._updatable) { // update buffer
-                this._engine.updateDynamicVertexBuffer(this._buffer, data, useBytes ? offset : offset * this.byteStride, (vertexCount ? vertexCount * this.byteStride : undefined));
+                this._engine.updateDynamicVertexBuffer(this._buffer, data, useBytes ? offset : offset * Float32Array.BYTES_PER_ELEMENT, (vertexCount ? vertexCount * this.byteStride : undefined));
                 this._data = null;
             }
         }
--- a/src/Mesh/babylon.buffer.ts
+++ b/src/Mesh/babylon.buffer.ts
@@ -34,7 +34,7 @@
 
             this._data = data;
 
-            this.byteStride = useBytes ? stride : stride * 4;
+            this.byteStride = useBytes ? stride : stride * Float32Array.BYTES_PER_ELEMENT;
 
             if (!postponeInternalCreation) { // by default
                 this.create();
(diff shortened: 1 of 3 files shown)


src/Mesh/babylon.vertexBuffer.ts
```diff
@@ -232,10 +232,13 @@
...
          * Updates directly the underlying WebGLBuffer according to the passed numeric array or Float32Array.  
-         * Returns the directly updated WebGLBuffer. 
+         * Returns the directly updated WebGLBuffer.
+         * @param data the new data
+         * @param offset the new offset
+         * @param useBytes set to true if the offset is in bytes
          */
-        public updateDirectly(data: DataArray, offset: number): void {
-            return this._buffer.updateDirectly(data, offset);
+        public updateDirectly(data: DataArray, offset: number, useBytes: boolean = false): void {
```
deltakosh: Just realized that it returns void so we should remove the returns comment and the  return keyword:)

src/Mesh/babylon.buffer.ts
```diff
@@ -123,13 +123,13 @@
...
          */
-        public updateDirectly(data: DataArray, offset: number, vertexCount?: number, useBytes = false): void {
+        public updateDirectly(data: DataArray, offset: number, vertexCount?: number, useBytes: boolean = false): void {
             if (!this._buffer) {
                 return;
             }
 
             if (this._updatable) { // update buffer
-                this._engine.updateDynamicVertexBuffer(this._buffer, data, useBytes ? offset : offset * 4, (vertexCount ? vertexCount * this.byteStride : undefined));
+                this._engine.updateDynamicVertexBuffer(this._buffer, data, useBytes ? offset : offset * this.byteStride, (vertexCount ? vertexCount * this.byteStride : undefined));
```
bghgary: I don't believe this is correct. If `useBytes` is false, then it should behave like before where `offset` is in float32 units which is a multiple of 4. `byteStride` has nothing to do with it.

a-carvallo: Shouldn't it be the same  for `vertexCount` then ?

deltakosh: That's right.I missed this one

deltakosh: For vertex count we are good as the bytestride is the stride in byte of the entire vertex

a-carvallo: So 4 for both. (I guess I was assuming it may be different, but actually it all ends in an array of Float32)

deltakosh: Actually only one not both (vertex count remains unchanged)

a-carvallo: Could `Float32Array.BYTES_PER_ELEMENT` be used in that case ? Unless I am wrong it's exactly what the 4 here means. (It helps understand better I think).

deltakosh: Works for me

bghgary: Perhaps update the other 4s in the file to this constant?

PR #4089, opened 2018-04-07
GUI getMeshPosition()

royibernthal:
Returns the position of a mesh for use in AdvancedDynamicTexture as a 2d point - use only x and y of the returned vector?

This is not working but I suppose it's enough to help you understand what's on my mind. What am I missing?

The projectedPosition returns x, y, z - NAN

### DIFF
--- a/gui/src/advancedDynamicTexture.ts
+++ b/gui/src/advancedDynamicTexture.ts
@@ -512,6 +512,22 @@ module BABYLON.GUI {
             this._attachToOnPointerOut(scene);
         }
 
+        public getMeshPosition(mesh: BABYLON.Mesh): BABYLON.Vector3 {
+            var scene = this.getScene();
+
+            if (!scene) {
+                return BABYLON.Vector3.Zero();
+            }
+
+            var globalViewport = this._getGlobalViewport(scene);
+            var position = mesh.getBoundingInfo().boundingSphere.center;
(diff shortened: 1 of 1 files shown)


royibernthal:
Correct. Can a Vector3 from the scene can be used as is without further translation inside an ADT? If so, why do you need to translate it to "projectedPosition" in Control/moveToVector3 for instance?

deltakosh:
You still need to project from 3d to 2d. What you did should work

royibernthal:
Then why doesn't it? :)

royibernthal:
Here's what I test locally:
http://www.babylonjs-playground.com/#BSM1NV

royibernthal:
kk thanks

deltakosh:
Fixed by https://github.com/BabylonJS/Babylon.js/commit/e7a2ea4cecc4da61576cb87e7c33983dbe70aa80
I added a adt.getProjectedPosition
To test it:
```
	scene.render();
	console.log( advancedTexture.getProjectedPosition(sphere1.getBoundingInfo().boundingSphere.center, sphere1.getWorldMatrix()) );
```

scene.render() must be called at least one to update all projection / view matrices

royibernthal:
Got it, thanks :)

What do you think about being able to project into a control's coordinates system? and/or some localToGlobal and globalToLocal functions for adt and its controls.

btw I'm working on a little something - a class that draws lines between to points / tracked controls / tracked meshes, I hope you won't be against adding such a thing to the source.

deltakosh:
As long as it is useful no problem :)

PR #4114, opened 2018-04-11
Complex Line Tracking

royibernthal:
This is still under development, need to test Vector2 and Control, and create a more convenient API.

Ignore getMeshPosition() commit in adt, I'm not sure how to undo that commit - need to research that.

Here's a PG tracking meshes:
http://www.babylonjs-playground.com/#H03KNW

Any thoughts? Do you think this could go into source after some polishing?


royibernthal:
ComplexLine (temp name) - draw lines that connect provided positions / tracked controls / tracked meshes / a mix of all of these.

Tracker - a helper class that represents a tracked item - either a position (x, y) as gui ValueAndUnit, could be percentage, etc, a control or a mesh.

For my use case I only needed to connect lines between multiple meshes at real time, but I thought I should make the solution more dynamic if I was to share it, so now it should fit more cases.

It's a more advanced and dynamic solution to Line.

royibernthal:
Imagine providing an array of (possibly mixed) vectors, controls and meshes to an instance that draws lines between all of them in real time.

gui/src/controls/complexLine.ts
```diff
@@ -0,0 +1,294 @@
+/// <reference path="../../../dist/preview release/babylon.d.ts"/>
+
+module BABYLON.GUI {
+
+    export class ComplexLine extends Control {
```
deltakosh: Perhaps MultiLine?

gui/src/controls/complexLine.ts
```diff
@@ -0,0 +1,294 @@
...
+        public set dash(value: Array<number>) {
+            if (this._dash === value) {
+                return;
+            }
+
+            this._dash = value;
+            this._markAsDirty();
+        }
+
+        getX(index: number): string | number  {
```
deltakosh: I think this should be at tracker level and we should just have getTracker(index) here

gui/src/controls/complexLine.ts
royibernthal: Yup

gui/src/tracker.ts
```diff
@@ -0,0 +1,48 @@
+/// <reference path="../../dist/preview release/babylon.d.ts"/>
+
+module BABYLON.GUI {
+
+    export class Tracker {
```
deltakosh: Perhaps segment instead of tracker ?

Tools/Gulp/config.json
```diff
@@ -1622,6 +1622,7 @@
                     "../../gui/src/measure.ts",
                     "../../gui/src/math2D.ts",
                     "../../gui/src/valueAndUnit.ts",
+					"../../gui/src/tracker.ts",
```
deltakosh: tab issue

Tools/Gulp/config.json
```diff
@@ -1636,7 +1637,8 @@
                     "../../gui/src/controls/button.ts",
                     "../../gui/src/controls/colorpicker.ts",
                     "../../gui/src/controls/inputText.ts",
-                    "../../gui/src/controls/virtualKeyboard.ts"
+                    "../../gui/src/controls/virtualKeyboard.ts",
+					"../../gui/src/controls/complexLine.ts"
```
deltakosh: tab issue

gui/src/advancedDynamicTexture.ts
```diff
@@ -527,6 +527,22 @@ module BABYLON.GUI {
             this._attachToOnPointerOut(scene);
         }
 
+        public getMeshPosition(mesh: BABYLON.Mesh): BABYLON.Vector3 {
```
deltakosh: need to revert this (as you mentioned)

gui/src/controls/complexLine.ts
royibernthal: Are you referring to get/set x/y/mesh/control? If so, I wanted to do just that but some functionality requires the control and the host. Perhaps pass the control to tracker?

gui/src/controls/complexLine.ts
```diff
@@ -0,0 +1,294 @@
...
+            }
+
+            if (tracker.control) {
+                tracker.control.onDirtyObservable.removeCallback(this.onTrackerUpdate);
+            }
+
+            tracker.control = value;
+
+            if (tracker.control) {
+                tracker.control.onDirtyObservable.add(this.onTrackerUpdate);
```
deltakosh: this may not work as "this" will not be the same in the observable. You may have to use () =>

gui/src/controls/complexLine.ts
deltakosh: Yes. You can add the multi-line hist as parameter to Segment constructor

gui/src/controls/complexLine.ts
royibernthal: I did in the definition:
private onTrackerUpdate = (): void => {
            this._markAsDirty();
        }

That way the callback reference is the same so I can remove it easily later as well

deltakosh: If private it has to start with_ then

royibernthal: No problem

royibernthal:
Thanks for the comments, I'll get to work on all of that tomorrow (it's 5 am here). Definitely, does the solution seem slow to you? If so, which parts do you think could be improved performance-wise? (You don't have to do the heavy lifting, just point me in the right direction)

deltakosh:
Nope this is fine as long as we update only when required :) and this seems to be the case so far

royibernthal:
:)

royibernthal:
Updated PG:
http://www.babylonjs-playground.com/#H03KNW#1

gui/src/segment.ts
```diff
@@ -0,0 +1,135 @@
...
+        private _translatePoint(): Vector2 {
+            if (this._mesh != null) {
+                return this._multiLine._host.getProjectedPosition(this._mesh.getBoundingInfo().boundingSphere.center, this._mesh.getWorldMatrix());
+            }
+            else if (this._control != null) {
+                return new Vector2(this._control.centerX, this._control.centerY);
+            }
+            else {
+                //TODO: extract number from px and % - eventually in order to support %
+                return new Vector2(this.x as number, this.y as number); //temp - as number
```
royibernthal: I'd like to translate Segment x and y (ValueAndUnit) to pixels. Do I need to manually pass the adt size to ValueAndUnit/getValueInPixel()? Isn't there a better way?

For instance, in a 1000x1000 canvas size, I'd like to translate { x: '50%', y: '50px' } to { x: 500, y: 50 }.

deltakosh: By the way I changed my mind regarding naming :) I prefer AnchorPoint instead of segment as it is not really a segment :)
yes for your question: you have to give adtSize to getValueInPixel()

royibernthal:
New PG combining all types successfully:
http://www.babylonjs-playground.com/#H03KNW#2

gui/src/segment.ts
royibernthal: Hmm Anchor isn't quite correct either in my opinion, let me think about it too :)

royibernthal: How about simply MultiLinePoint? Or PointInfo / PointData / DynamicPoint / AdvancedPoint?

deltakosh: MultiLinePoint is great

royibernthal:
I think that should be it, did I miss anything?

gui/src/controls/multiLine.ts
```diff
@@ -0,0 +1,215 @@
...
+        public set dash(value: Array<number>) {
+            if (this._dash === value) {
+                return;
+            }
+
+            this._dash = value;
+            this._markAsDirty();
+        }
+
+        getAt(index: number): MultiLinePoint {
```
deltakosh: getPointAt

gui/src/controls/multiLine.ts
```diff
@@ -0,0 +1,215 @@
...
+            if (!this._points[index]) {
+                this._points[index] = new MultiLinePoint(this);
+            }
+
+            return this._points[index] as MultiLinePoint;
+        }
+
+        onPointUpdate = (): void => {
```
deltakosh: make it private

gui/src/controls/multiLine.ts
```diff
@@ -0,0 +1,215 @@
...
+                    point.control = item;
+                }
+                else if (item.x != null && item.y != null) {
+                    point.x = item.x;
+                    point.y = item.y;
+                }
+            });
+        }
+
+        push(): MultiLinePoint {
```
deltakosh: general: please add public when public

gui/src/controls/multiLine.ts
```diff
@@ -0,0 +1,215 @@
...
+            if (this._lineWidth === value) {
+                return;
+            }
+
+            this._lineWidth = value;
+            this._markAsDirty();
+        }   
+
+        public set horizontalAlignment(value: number) {
+            return;
```
deltakosh: ?

gui/src/multiLinePoint.ts
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            if (this._mesh && this._meshObserver) {
+                this._mesh.getScene().onAfterCameraRenderObservable.remove(this._meshObserver);
+            }
+
+            this._mesh = value;
+
+            if (this._mesh) {
+                this._meshObserver = this._mesh.getScene().onAfterCameraRenderObservable.add(this._multiLine.onPointUpdate);
```
deltakosh: this line seems suspicious (Should it be this.,_meshObserver instead?)

gui/src/multiLinePoint.ts
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        dispose(): void {
+            this.control = null;
+            this.mesh = null;
+        }
+
+    }
+
+    export interface IValueAndUnitVector2 {
```
deltakosh: no need to declare a public interface
Just add { x: string | number, y: string | number } in the add function declaration

royibernthal: You're right, done.

gui/src/multiLinePoint.ts
royibernthal: Nope, it means if a valid mesh value was set - create an observer for it, unless I misunderstood you

gui/src/controls/multiLine.ts
royibernthal: I started this class from a copy of Line. I'm not sure what's the reason behind it:
https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/controls/line.ts#L118

gui/src/controls/multiLine.ts
royibernthal: Got it, done.

gui/src/controls/multiLine.ts
royibernthal: I need it to be public unless we change the implementation

https://github.com/royibernthal/Babylon.js/blob/4f2c9bb6e435fda2caa6e02784ceb8a74d156a92/gui/src/multiLinePoint.ts#L74

https://github.com/royibernthal/Babylon.js/blob/4f2c9bb6e435fda2caa6e02784ceb8a74d156a92/gui/src/multiLinePoint.ts#L96

gui/src/controls/multiLine.ts
deltakosh: Oh yeah true :)

gui/src/multiLinePoint.ts
deltakosh: i was expecting:
 this._meshObserver = this._mesh.getScene().onAfterCameraRenderObservable.add(this._meshObserver);

royibernthal: Observable/add() expects a callback doesn't it? meshObserver is an Observer. (that in the first place was created by passing that callback)

royibernthal: What am I missing? :)

deltakosh: ok my bad

royibernthal: :)

royibernthal:
ty, I'll make a PR for doc later tonight or tomorrow

royibernthal:
btw why did the Travis CI build fail?

Issue #4093, opened 2018-04-09
Add support for sparse accessors in glTF2Loader


sbtron:

cx20:
I confirmed that Babylon.js is now able to display [Simple Sparse Accessor](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=SimpleSparseAccessor&scale=1&type=glTF) model.
The current status is all green. congratulations!

https://github.com/cx20/gltf-test#simple-models-for-testing-individual-features
![image](https://user-images.githubusercontent.com/1833979/38704731-341d0f12-3ee2-11e8-8c1f-9d914048327c.png)

bghgary:
Thanks @cx20 for updating so fast!

PR #4120, opened 2018-04-12
Add sparse accessor support to glTF loader

bghgary:
Test coverage will happen when we implement sparse accessors in the glTF asset generator


bghgary:
Fixes #4093

Issue #4121, opened 2018-04-12
Run playground Several times leads to potential crash


sebavan:
As seen today running the same PG several times is leading to a crash: "too many webgl contexts..."

This can highlight a potential leak.

PR #4124, opened 2018-04-12
avoid creating multiple canvases in BABYLON.Engine.IsSupported, log e…

TrevorDev:
…rrors to console for easyier debugging

https://github.com/BabylonJS/Babylon.js/issues/4121

I don't know why this issue just started occurring now. Anyone have any ideas?


TrevorDev:
This is not an issue with babylon, a plugin held ref to context

PR #4128, opened 2018-04-13
Added TRS animation support to glTF serializer; refactorings

kcoley:


serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+            else {
+                if (interpolation === AnimationSamplerInterpolation.LINEAR || interpolation === AnimationSamplerInterpolation.STEP) {
+                    _GLTFAnimation._CreateLinearOrStepAnimation(babylonMesh, animation, animationChannelTargetPath, frameDelta, inputs, outputs, convertToRightHandedSystem, useQuaternion);
+
+                }
+                else if (interpolation === AnimationSamplerInterpolation.CUBICSPLINE) {
+                    _GLTFAnimation._CreateCubicSplineAnimation(babylonMesh, animation, animationChannelTargetPath, frameDelta, inputs, outputs, convertToRightHandedSystem, useQuaternion);
+                }
+                else {
+                    Tools.Error('Unsupported animation interpolation type!');
```
deltakosh: why not baking in this case?

kcoley: True, that can be done.  Just pushed a change for that

serializers/src/glTF/2.0/babylon.glTFUtilities.ts
```diff
@@ -0,0 +1,149 @@
+/// <reference path="../../../../dist/babylon.glTF2Interface.d.ts"/>
+
+module BABYLON.GLTF2 {
+    /**
+     * @hidden
+     */
+    export class _GLTFUtilities {
+        /**
+         * @ignore
```
bghgary: `@hidden` instead?

serializers/src/glTF/2.0/babylon.glTFUtilities.ts
```diff
@@ -0,0 +1,149 @@
...
+            return new Vector3(vector.x, vector.y, -vector.z);
+        }
+
+        /**
+         * @ignore
+         * 
+         * Converts a Vector3 to right-handed
+         * @param vector Vector3 to convert to right-handed
+         */
+        public static GetRightHandedVector3FromRef(vector: Vector3) {
```
bghgary: These aren't really `FromRef` functions. You probably want to call these something like ConvertToRightHandedVector3InPlace

tests/unit/babylon/serializers/babylon.glTFSerializer.tests.ts
```diff
@@ -163,5 +163,286 @@ describe('Babylon glTF Serializer', () => {
                 done();
             });
         });
+        it('should serialize single component translation animation to glTF', (done) => {
```
bghgary: what happened to the space formatting of these test functions?

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+            }
+
+            return null;
+        }
+
+        /**
+         * Math.fround is not supported in IE or PhantomJS, so this helper function is used instead
+         * @param num value to fround
+         */
+        private static fround(num: number) {
```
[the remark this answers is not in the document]
bghgary: Also, you should probably use a temp static variable to store the Float32Array so that you don't have to allocate one each time this is called.

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+                }
+                default: {
+                    Tools.Error(`Unsupported animatable property ${property[0]}`);
+                }
+            }
+            if (animationChannelTargetPath) {
+                return { animationChannelTargetPath: animationChannelTargetPath, dataAccessorType: dataAccessorType, useQuaternion: useQuaternion };
+            }
+            else {
+                Tools.Error('animation channel target path and data accessor type could be deduced');
```
bghgary: seems like this error and the error above will both be thrown? is that necessary?

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+         */
+        private static fround(num: number) {
+            return new Float32Array([num])[0];
+        }
+
+        private static _DeduceAnimationInfo(animation: Animation): Nullable<_IAnimationInfo> {
+            let animationChannelTargetPath: AnimationChannelTargetPath | undefined;
+            let dataAccessorType = AccessorType.VEC3;
+            let useQuaternion: boolean = false;
+            let property = animation.targetProperty.split('.');
```
bghgary: can be const

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+        /**
+         * Math.fround is not supported in IE or PhantomJS, so this helper function is used instead
+         * @param num value to fround
+         */
+        private static fround(num: number) {
+            return new Float32Array([num])[0];
+        }
+
+        private static _DeduceAnimationInfo(animation: Animation): Nullable<_IAnimationInfo> {
+            let animationChannelTargetPath: AnimationChannelTargetPath | undefined;
```
bghgary: probably should use Nullable here

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+                return { animationChannelTargetPath: animationChannelTargetPath, dataAccessorType: dataAccessorType, useQuaternion: useQuaternion };
+            }
+            else {
+                Tools.Error('animation channel target path and data accessor type could be deduced');
+            }
+            return null;
+        }
+
+        /**
+         * @ignore
```
bghgary: `@hidden` and remove parameters or document them

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+                            _GLTFAnimation._SetInterpolatedValue(value, time, animationChannelTargetPath, quaternionCache, inputs, outputs, convertToRightHandedSystem, useQuaternion);
+                        }
+                    }
+                }
+                if (maxUsedFrame) {
+                    minMaxFrames.max = maxUsedFrame;
+                }
+            }
+            else {
+                Tools.Warn('Frame range is too small to bake animation');
```
bghgary: it should always be possible to bake. if the range is too small, then just output one frame.

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -0,0 +1,718 @@
...
+            }
+        }
+
+        /**
+         * Checks if two values are equal to each other
+         * @param value1 First value as a Vector3 or Quaternion
+         * @param value2 Second value as a Vector3 or Quaternion
+         */
+        private static _ValuesAreEqual(value1: Vector3 | Quaternion, value2: Vector3 | Quaternion): boolean {
```
bghgary: not sure this is necessary. typescript should see that both types have an equal function

serializers/src/glTF/2.0/babylon.glTFAnimation.ts
```diff
@@ -339,100 +331,129 @@ module BABYLON.GLTF2 {
...
-                }
-                else {
-                    Quaternion.RotationYawPitchRollToRef(value.y, value.x, value.z, quaternionCache);
-                }
-                if (convertToRightHandedSystem) {
-                    quaternionCache.x *= -1;
-                    quaternionCache.y *= -1;
-                    outputs.push(quaternionCache.asArray());
-                }
+            if (value instanceof Number) {
```
bghgary: this probably doesn't work, use typeof?

PR #4132, opened 2018-04-14
Add support for cameras to glTF loader

bghgary:


src/Cameras/Inputs/babylon.freeCameraMouseInput.ts
```diff
@@ -76,7 +76,7 @@ module BABYLON {
                         var offsetX = evt.clientX - this.previousPosition.x;
                         var offsetY = evt.clientY - this.previousPosition.y;
 
-                        if (this.camera.getScene().useRightHandedSystem) {
+                        if (this.camera.getScene().useRightHandedSystem === (this.camera.parent && this.camera.parent._getWorldMatrixDeterminant() >= 0)) {
```
deltakosh: I wonder if this will  or be executed if camera.parent == null or undefined

src/Cameras/babylon.camera.ts
```diff
@@ -462,23 +462,6 @@
 
             this._refreshFrustumPlanes = true;
 
-            if (!this.parent || !this.parent.getWorldMatrix) {
```
deltakosh: We need to build worldMatrix here

src/Cameras/Inputs/babylon.freeCameraMouseInput.ts
bghgary: It doesn't because of javascriptism, but I agree it's not clear.

src/Cameras/babylon.camera.ts
bghgary: Is worldMatrix necessary every frame?

deltakosh: It is not every frame but every time we build the view

bghgary: I don't mean calculating it. I mean that getWorldMatrix is called every frame.

deltakosh: We just need to make sure that camera._worldMatrix is updated when view matrix is updated

bghgary: I've already done that with my last commit. The way the code worked before it updated the world matrix to the inverse of the local view matrix as a temp matrix. The _worldMatrix only gets updated to the correct value when getWorldMatrix is called. Anyways, I think the current changes work the way you are describing.

deltakosh: Oh ok... I'll check later then:) on my phone now

src/Cameras/babylon.camera.ts
```diff
@@ -462,29 +461,14 @@
...
-                this._markSyncedWithParent();
-            }
-
             if (this._cameraRigParams && this._cameraRigParams.vrPreViewMatrix) {
                 this._computedViewMatrix.multiplyToRef(this._cameraRigParams.vrPreViewMatrix, this._computedViewMatrix);
             }
 
             this.onViewMatrixChangedObservable.notifyObservers(this);
 
+            this._computedViewMatrix.invertToRef(this._worldMatrix);
```
deltakosh: Oh yeah I missed this line..this is cool

cx20:
I confirmed that the camera can be used with Polly model of glTF 2.0.
http://jsdo.it/cx20/m7JB

bghgary:
@cx20 Awesome! Thanks for your work!

cx20:
I also tried using the camera with VC.gltf.
http://jsdo.it/cx20/MJ1p/

It seems that helicopters and fighter planes are missing a bit. Is it a usage problem?
image

bghgary:
@cx20 As far as I can tell, these are issues with this model. It has some odd geometry and clipping issues.

bghgary:
BTW, you can choose different cameras in the inspector of Babylon.js available in the sandbox. Makes it easy to test cameras in glTF.

image

PR #4135, opened 2018-04-14
item passed to push + what's new

royibernthal:


royibernthal:
Out of curiosity, how come this error still occurs even though what's new is updated? Is there a specific way to update what's new in a way that'll pass this test?

> The command "gulp tests-whatsnew" exited with 1.

deltakosh:
Rats I missed it
You updated the wrong what's new
It is the one in dist/preview release that you have to update

PR #4136, opened 2018-04-16
Fix one bug about DynamicTexture's size

finscn:
When create a DynamicTexture with size {width:0, height:number} , the code `if (options.width) { ... }` will take a wrong result.

example : 
in babylon.gui , there is one line  like this : 
```
var result = new AdvancedDynamicTexture(name, 0, 0, scene, false, sampling);
```

When run above line , the code below will with mistake: 
```
        function DynamicTexture(name, options, scene, generateMipMaps, samplingMode, format) {
            if (scene === void 0) { scene = null; }
            if (samplingMode === void 0) { samplingMode = BABYLON.Texture.TRILINEAR_SAMPLINGMODE; }
            if (format === void 0) { format = BABYLON.Engine.TEXTUREFORMAT_RGBA; }
            var _this = _super.call(this, null, scene, !generateMipMaps, undefined, samplingMode, undefined, undefined, undefined, undefined, format) || this;
            _this.name = name;
            _this._engine = _this.getScene().getEngine();
            _this.wrapU = BABYLON.Texture.CLAMP_ADDRESSMODE;
            _this.wrapV = BABYLON.Texture.CLAMP_ADDRESSMODE;
            _this._generateMipMaps = generateMipMaps;
            if (options.getContext) {
                _this._canvas = options;
                _this._texture = _this._engine.createDynamicTexture(options.width, options.height, generateMipMaps, samplingMode);
            }
            else {
                _this._canvas = document.createElement("canvas");
                if (options.width) {
                    _this._texture = _this._engine.createDynamicTexture(options.width, options.height, generateMipMaps, samplingMode);
                }
                else {
                    _this._texture = _this._engine.createDynamicTexture(options, options, generateMipMaps, samplingMode);
                }
            }
            var textureSize = _this.getSize();
            _this._canvas.width = textureSize.width;
            _this._canvas.height = textureSize.height;
            _this._context = _this._canvas.getContext("2d");
            return _this;
        }
```
after ` var textureSize = _this.getSize();` , the `textureSize` will be 
```
 { 
  width: {width:0, height:0}, 
  height: {width:0, height:0}
}
```

### DIFF
--- a/src/Materials/Textures/babylon.dynamicTexture.ts
+++ b/src/Materials/Textures/babylon.dynamicTexture.ts
@@ -35,7 +35,7 @@
             } else {
                 this._canvas = document.createElement("canvas");
 
-                if (options.width) {
+                if (options.width || options.width === 0) {
                     this._texture = this._engine.createDynamicTexture(options.width, options.height, generateMipMaps, samplingMode);
                 } else {
                     this._texture = this._engine.createDynamicTexture(options, options, generateMipMaps, samplingMode);


finscn:
I don't know why `The Travis CI build failed` , 
but I'm sure that this PR is necessary.

finscn:
@deltakosh , in babylon-gui , yes , it is.
 the babylon-gui is official project.
so , i think the 0 is valid in babylon design

finscn:
and the method getSize could return  0 size.
So 0  is a valid value of width& height

deltakosh:
I meant: i dont think a texture with one dimension set to 0 is invalid
Min value is 1

deltakosh:
I don't think it is
1 should be the minimum

finscn:
It's another question.
whatever  `if (options.width) { ...}` is wrong in some cases.
if  you think 0 is invalid , please change 
https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/advancedDynamicTexture.ts#L599

And there some default values of width/height/size are 0 in Babylon , please change them too.

deltakosh:
Sorry I was unclear: 0 AND 0 are ok as this will let the texture be rescaled to always match fullscreen resolution
But you cannot have 0 and something else.  In this case the system consider you want 0 and 0. So the current behavior is ok

finscn:
@deltakosh , I think you don't understand me (my english is very weak).

It's that  NO User wants to use Size(0,0) ,   Size(0,0) is used in  official project: `babylon.gui`.
```
var result = new AdvancedDynamicTexture(name, 0, 0, scene, false, sampling);
```
when call the code above ,  the code below will take mistakes in some browser: 
```
            var textureSize = _this.getSize();
            _this._canvas.width = textureSize.width;
            _this._canvas.height = textureSize.height;
            _this._context = _this._canvas.getContext("2d");
```

because  textureSize will be 
```
  { 
  width: {width:0, height:0}, 
  height: {width:0, height:0}
 }
```

Then 
```
_this._canvas.width =  {width:0, height:0};
_this._canvas.height =  {width:0, height:0};
```
it takes an error : 
image

deltakosh:
Can you create a playground to reproduce your problem?

deltakosh:
because GUI works well here: https://www.babylonjs-playground.com/#T7AZBC

finscn:
in current version ,  when `options = {width:0 , height:0}`
this code will be run : 
```
_this._texture = _this._engine.createDynamicTexture({width:0 , height:0}, {width:0 , height:0}, generateMipMaps, samplingMode);
```
then ,  _this.getSize() will return 
```
  { 
  width: {width:0, height:0}, 
  height: {width:0, height:0}
 }
```
( it's should be  `{width:0, height:0}` )

deltakosh:
OOOK!! :) Gotcha!

finscn:
I can't , because I meet the error on some android devices. 

in your browser , there is no error , 
because  in most browsers,  
when you set a  non-number to canvas.width , the browser will ignore it.

But in some browsers of android , it will take an error.

Whatever whether take an error ,  
```
_this._canvas.width =  {width:0, height:0};
_this._canvas.height =  {width:0, height:0};
```
is always not correct.

finscn:
@deltakosh  , thanks. I'm sorry for my bad english.

deltakosh:
No problem :)

PR #4184, opened 2018-04-22
FIxing standard rendering pipeline + comments + serialize / parse ssa…

julien-moreau:
…o2 rendering pipeline


deltakosh:
There is a problem with _ratio cc Travis

julien-moreau:
Saw the warning in VSCode but though it was not important because my local "gulp-typescript" did not tell me anything :)

PR #4186, opened 2018-04-23
Update on the sandbox

rlouvat:
Ask by @deltakosh 

The hover on the combo was removed. Now you have to click in order to open/close the liste of animation groups. Also when you select an item, it closes the list.
The hover effect that change the items background color is still there.

The slider clickable zone is now bigger.

A wrong html balise was deleted.

### DIFF
--- a/sandbox/index.css
+++ b/sandbox/index.css
@@ -271,17 +271,13 @@ a {
     z-index: 1;
     min-width: 135px;
     width: 200px;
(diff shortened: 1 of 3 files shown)


sebavan:
Will try it asap locally before merging

sebavan:
Thx, closing the PR as I added a small change before merging

PR #4191, opened 2018-04-23
do not fire action manager pick events if clickInfo ignore is true

TrevorDev:
See http://www.html5gamedevs.com/topic/37134-sprites-pick-and-pickup-fired-twice/?tab=comments#comment-212597

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -2116,25 +2116,28 @@
                     this._processPointerUp(pickResult, evt, clickInfo);
 
                     // Sprites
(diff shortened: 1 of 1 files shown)


deltakosh:
Make sure this does not introduce issues with on double pick or Pick trigger

TrevorDev:
@deltakosh  Pick triggers still work as they did before. However it seems double pick on sprite never worked and still doesn't after this change.

deltakosh:
Yes no worry I was mostly mentioning meshes. Sprites do not have all the options

PR #4199, opened 2018-04-24
Animation reset fix

### LINKED ISSUE #4174: glTF animation channels can cross modify other channels
- Expected result: Animation channels do not influence each other when loading from a glTF model
- Actual result: Animation channels influence each other when loading from a glTF model

The glTF model is attached below:
[testModel.zip](https://github.com/BabylonJS/Babylon.js/files/1934034/testModel.zip)

bghgary:
_Engine fixes_
* Fix bug with animations not resetting back to original value (fixes #4174). I also reverted the code to allow animations to apply to multiple targets as with blending since it doesn't work in all cases.

@deltakosh Please review the animations change. I have tested blending with https://www.babylonjs-playground.com/#IQN716#9.

### DIFF
--- a/src/Animations/babylon.runtimeAnimation.ts
+++ b/src/Animations/babylon.runtimeAnimation.ts
@@ -164,8 +164,13 @@
 
         /**
          * Resets the runtime animation to the beginning
+         * @param restoreOriginal defines whether to restore the target property to the original value
          */
-        public reset(): void {
+        public reset(restoreOriginal = false): void {
+            if (restoreOriginal && this._originalValue != null) {
+                this.setValue(this._originalValue, -1);
+            }
+
             this._offsetsCache = {};
             this._highLimitsCache = {};
             this._currentFrame = 0;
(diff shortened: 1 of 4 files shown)


src/Animations/babylon.runtimeAnimation.ts
```diff
@@ -212,36 +217,19 @@
...
-        private _setValue(target: any, currentValue: any, weight = 1.0): void {
             // Set value
             var path: any;
             var destination: any;
 
             let targetPropertyPath = this._animation.targetPropertyPath
 
             if (targetPropertyPath.length > 1) {
-                var property = target[targetPropertyPath[0]];
+                var property = this._target;
```
deltakosh: On my phone but don't we miss the first part of the path?

bghgary: Oops. Will fix.

bghgary:
I've split the getVerticesData into a separate PR. @sebavan already merged the sandbox fixes.

src/Animations/babylon.runtimeAnimation.ts
```diff
@@ -272,27 +274,7 @@
                         this._originalBlendValue = originalValue;
                     }
                 }
-            }
-
-            if (weight !== -1.0) {
```
bghgary: This used to get the original value only when weight is not -1.0. Is the original value intended to be the original value before blending? My new code does not check weight because I want to be able to reset the value back to the original value before setValue is ever called. Should I use separate variables for this?

deltakosh: Unfortunately it is a bit trickier as the current value read and store in the original value is the value of the target at current time.
If an animation was already animating the target the you will get the current value computed from the animation.

This is why I think we should just get animation.keys[0].value (which was the intend of animate(0) but I think we can safely use first value

deltakosh:
Something like that:
image

bghgary:
That won't work to fix the issue I'm trying to fix. The issue I'm trying to fix is to restore the target back to the value before animation is applied.

deltakosh:
But is it the responsibility of the animation to do that?

PR #4201, opened 2018-04-24
Master

MackeyK24:
Support global environment HDRCubeTexture type.

- Added parsedData.environmentTextureType = “BABYLON.HDRCubeTexture”
option

- Added parsedData.environmentTextureSize HDR option

- Added parsedData.environmentTextureRotationY HDR option


src/Loading/Plugins/babylon.babylonFileLoader.ts
```diff
@@ -680,7 +680,16 @@
 
                 // Environment texture		
                 if (parsedData.environmentTexture !== undefined && parsedData.environmentTexture !== null) {
-                    scene.environmentTexture = CubeTexture.CreateFromPrefilteredData(rootUrl + parsedData.environmentTexture, scene);
+                    if (parsedData.environmentTextureType && parsedData.environmentTextureType === "BABYLON.HDRCubeTexture") {
+                        var hdrSize:number = (parsedData.environmentTextureSize) ? parsedData.environmentTextureSize : 128;
+                        var hdrTexture:BABYLON.HDRCubeTexture = new BABYLON.HDRCubeTexture(rootUrl + parsedData.environmentTexture, scene, hdrSize);
+                        if (parsedData.environmentTextureRotationY) {
```
sebavan: should we not apply rotation in case of prefiltered texture as well ?

sebavan: Would it not help if ppl wanted to rotate the environment lighting with prefiltered data ?

We added the support for rotationY in cubeTexture as well for this.

MackeyK24:
I guess so... didnt know you added to CubeTexture. I originally added to
HDRCubeTexture to fix the 90 degress rotation issue when use panorama hdr
images.

But your right, it should rotate via the matrix so would be fine to add
that support for general purpose reflection rotation support

PR #4202, opened 2018-04-24
Fix geometry getVerticesData when not using the default element stride

bghgary:
Also add unit tests and fix normalized math according to https://github.com/KhronosGroup/glTF/issues/1317


src/Mesh/babylon.geometry.ts
```diff
@@ -389,7 +389,7 @@
 
             const defaultStride = VertexBuffer.DeduceStride(vertexBuffer.getKind());
```
sebavan: Looks inconsistent with getSize

bghgary: fixing

PR #4212, opened 2018-04-26
Fix animation blending issue with glTF loader

bghgary:
Tested with:
http://localhost:1338/Playground/index-local.html#PSR2ZX#10
http://localhost:1338/Playground/index-local.html#BCU1XR
All glTF sample models


bghgary:
And the model from #4174

PR #4214, opened 2018-04-26
Fix SPS vertexdata serialization

a-carvallo:


src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -200,10 +200,10 @@
                 vertexData.indices = (this._depthSort) ? this._indices : this._indices32;
                 vertexData.set(this._positions32, VertexBuffer.PositionKind);
                 vertexData.set(this._normals32, VertexBuffer.NormalKind);
-                if (this._uvs32) {
+                if (this._uvs32.length > 0) {
```
deltakosh: What if uv32 is null or undefined?

a-carvallo: Fair point, a few lines before that, there is this line : 
```ts
this._uvs32 = new Float32Array(this._uvs);
```
So it should never be null or undefined.

deltakosh: Makes sense:)
Please fix the conflict and I can merge

a-carvallo: Should be good !

deltakosh:
Thanks!

PR #4230, opened 2018-04-28
Master

MackeyK24:
Support bone metadata serialization.


src/Bones/babylon.skeleton.ts
```diff
@@ -502,11 +502,15 @@
                 };
 
                 serializationObject.bones.push(serializedBone);
-
+                
                 if (bone.length) {
                     serializedBone.length = bone.length;
                 }
 
+                if (bone.metadata) {
```
deltakosh: I don't think there is a bone metadata property so far

PR #4235, opened 2018-04-29
Adding fur occlusion property to FurMaterial

julien-moreau:


julien-moreau:
Yeaaaaaaah everything is green !!

PR #4241, opened 2018-04-30
glTF exporter updates

kcoley:
- Fix left-handed/right-handed conversion
- Supports exporting primitive modes
- Supports exporting meshes with up to two uv channels (per glTF spec)
- Supports texture sampling
- Uses async functions


deltakosh:
Good to go!
Please update doc :)

PR #4252, opened 2018-05-02
Add support for KHR_texture_transform to glTF loader

### LINKED ISSUE #4149: Add support for KHR_texture_transform to glTF loader
See https://github.com/KhronosGroup/glTF/pull/1015 for proposal.

This is the original implementation. It just needs a bit of clean up, naming updates, etc.
https://github.com/bghgary/Babylon.js/tree/EXT_texture_transform

bghgary:
Includes a new validation test:
image

Fixes #4149


deltakosh:
Can you update the new what's new?:)

bghgary:
Yup, was already doing it :)

src/Materials/Textures/babylon.texture.ts
```diff
@@ -65,6 +65,24 @@
         @serialize()
         public wAng = 0;
 
+        /**
+         * Defines the center of rotation (U)
+         */
+        @serialize()
+        public uCenter = 0.5;
```
bghgary: Any opinions on the naming of this? It's just for rotation and not for scale. Should I call it uRotationCenter perhaps? It's a bit odd compared to uAng which is also for rotation.

deltakosh: I guess uRotationCenter is good.
uAng is OK as you cannot scale with an angle

bghgary: Done!

dist/preview release/what's new.md
```diff
@@ -3,6 +3,9 @@
 ## Major updates
 
 ## Updates
+- Add support for specifying the center of rotation to textures ([bghgary](http://www.github.com/bghgary))
+- Add support for KHR_texture_transform to glTF 2.0 loader ([bghgary](http://www.github.com/bghgary))
```
deltakosh: Do you mind creating a gltf section like I did for previous what's new
I guess we will have more to come:)

deltakosh:
Wonderful

PR #4271, opened 2018-05-04
Fix Help Logo & FireFox bug

SzeyinLee:
Updated font file & help logo

For Firefox
- need to use event.event.target instead of event.event.srcElement.
- fix centering issue


Viewer/src/viewer/defaultViewer.ts
```diff
@@ -85,7 +85,7 @@ export class DefaultViewer extends AbstractViewer {
 
         let pointerDown = <PointerEvent>event.event;
         if (pointerDown.button !== 0) return;
-        var element = (<HTMLElement>event.event.srcElement);
+        var element = (<HTMLElement>event.event.target);
```
deltakosh: Was it tested on edge and chrome?

SzeyinLee: Yes

deltakosh: You rock

SzeyinLee: Should check with @RaananW if this is an issue.

Issue #4268, opened 2018-05-03
Occulus go webVR usability issues


TrevorDev:
- [x] Hill valley demo had a very low frame rate and crashed once (this issue is consistent across all webvr libs)
- [x] Simpler webVR demo ran at around 40fps(lower than expected)
- [x] Controller position was attached to a head instead of where hands would normally be
- [x] Selecting GUI in webVR did not work

PR #4289, opened 2018-05-06
treat oculus go as gearVR controller and position it where a persons …

TrevorDev:
…hand would be, also updated PG webVR example to match the one here https://doc.babylonjs.com/examples/ which runs reasonably on the GO

https://github.com/BabylonJS/Babylon.js/issues/4268


deltakosh:
Thanks by the way:)

Issue #4291, opened 2018-05-06
OnPhysicsCollide unregistration is broken


antonipelka:
It is impossible to unregister previously registered callback function from PhysicsImpostor.
Passing the same arguments first to https://doc.babylonjs.com/api/classes/babylon.physicsimpostor#registeronphysicscollide and then to https://doc.babylonjs.com/api/classes/babylon.physicsimpostor#unregisteronphysicscollide should result in proper addition and removal of collision callback - however only registration works.

The bug is located here: https://github.com/BabylonJS/Babylon.js/blob/master/src/Physics/babylon.physicsImpostor.ts#L361
Each time the unregister method is invoked a new collide callback object is created. Objects are compared by their reference and in this case indexOf will not match callback object registered a few lines earlier in "registerOnPhysicsCollide" method.

- Bug repro on [playground](https://playground.babylonjs.com): https://playground.babylonjs.com/#HR7VPD
- Expected result: Successful callback unregistration
- Current result: Warning "Function to remove was not found", callback not disposed and still firing.

deltakosh:
@RaananW can you check this one?

RaananW:
I'll look into it that

RaananW:
Fix POC - https://playground.babylonjs.com/#HR7VPD#1 , waiting to be merged

PR #4292, opened 2018-05-07
Physics on collide unregister

RaananW:
Fixing https://github.com/BabylonJS/Babylon.js/issues/4291


RaananW:
it is in the bug fixes

PR #4313, opened 2018-05-11
glTF Serializer: Added procedural texture to prevent premultiplication of alpha when exporting opaque models to glTF

kcoley:
- Added procedural texture to prevent premultiplication of alpha when exporting opaque models to glTF
- Used the Babylon default material instead of omitting and using the glTF default material 
- Refactored to use promises in more functions to support the async shader call

### DIFF
--- a/serializers/src/glTF/2.0/babylon.glTFExporter.ts
+++ b/serializers/src/glTF/2.0/babylon.glTFExporter.ts
@@ -652,86 +652,83 @@ module BABYLON.GLTF2 {
          * @returns object with glb filename as key and data as value
          */
         public _generateGLBAsync(glTFPrefix: string): Promise<GLTFData> {
-            return new Promise((resolve, reject) => {
-                this.generateBinaryAsync().then(binaryBuffer => {
-                    const jsonText = this.generateJSON(true);
-                    const glbFileName = glTFPrefix + '.glb';
-                    const headerLength = 12;
-                    const chunkLengthPrefix = 8;
-                    const jsonLength = jsonText.length;
-                    let imageByteLength = 0;
+            return this.generateBinaryAsync().then(binaryBuffer => {
+                const jsonText = this.generateJSON(true);
+                const glbFileName = glTFPrefix + '.glb';
+                const headerLength = 12;
+                const chunkLengthPrefix = 8;
+                const jsonLength = jsonText.length;
+                let imageByteLength = 0;
 
-                    for (let key in this.imageData) {
-                        imageByteLength += this.imageData[key].data.byteLength;
-                    }
-                    const jsonPadding = this._getPadding(jsonLength);
-                    const binPadding = this._getPadding(binaryBuffer.byteLength);
-                    const imagePadding = this._getPadding(imageByteLength);
+                for (let key in this.imageData) {
+                    imageByteLength += this.imageData[key].data.byteLength;
+                }
+                const jsonPadding = this._getPadding(jsonLength);
+                const binPadding = this._getPadding(binaryBuffer.byteLength);
+                const imagePadding = this._getPadding(imageByteLength);
 
-                    const byteLength = headerLength + (2 * chunkLengthPrefix) + jsonLength + jsonPadding + binaryBuffer.byteLength + binPadding + imageByteLength + imagePadding;
+                const byteLength = headerLength + (2 * chunkLengthPrefix) + jsonLength + jsonPadding + binaryBuffer.byteLength + binPadding + imageByteLength + imagePadding;
 
-                    //header
-                    const headerBuffer = new ArrayBuffer(headerLength);
-                    const headerBufferView = new DataView(headerBuffer);
-                    headerBufferView.setUint32(0, 0x46546C67, true); //glTF
-                    headerBufferView.setUint32(4, 2, true); // version
-                    headerBufferView.setUint32(8, byteLength, true); // total bytes in file
+                //header
+                const headerBuffer = new ArrayBuffer(headerLength);
+                const headerBufferView = new DataView(headerBuffer);
+                headerBufferView.setUint32(0, 0x46546C67, true); //glTF
+                headerBufferView.setUint32(4, 2, true); // version
+                headerBufferView.setUint32(8, byteLength, true); // total bytes in file
 
-                    //json chunk
-                    const jsonChunkBuffer = new ArrayBuffer(chunkLengthPrefix + jsonLength + jsonPadding);
-                    const jsonChunkBufferView = new DataView(jsonChunkBuffer);
-                    jsonChunkBufferView.setUint32(0, jsonLength + jsonPadding, true);
-                    jsonChunkBufferView.setUint32(4, 0x4E4F534A, true);
+                //json chunk
+                const jsonChunkBuffer = new ArrayBuffer(chunkLengthPrefix + jsonLength + jsonPadding);
+                const jsonChunkBufferView = new DataView(jsonChunkBuffer);
+                jsonChunkBufferView.setUint32(0, jsonLength + jsonPadding, true);
+                jsonChunkBufferView.setUint32(4, 0x4E4F534A, true);
 
-                    //json chunk bytes
-                    const jsonData = new Uint8Array(jsonChunkBuffer, chunkLengthPrefix);
-                    for (let i = 0; i < jsonLength; ++i) {
-                        jsonData[i] = jsonText.charCodeAt(i);
-                    }
+                //json chunk bytes
+                const jsonData = new Uint8Array(jsonChunkBuffer, chunkLengthPrefix);
+                for (let i = 0; i < jsonLength; ++i) {
+                    jsonData[i] = jsonText.charCodeAt(i);
+                }
 
-                    //json padding
-                    const jsonPaddingView = new Uint8Array(jsonChunkBuffer, chunkLengthPrefix + jsonLength);
-                    for (let i = 0; i < jsonPadding; ++i) {
-                        jsonPaddingView[i] = 0x20;
-                    }
+                //json padding
+                const jsonPaddingView = new Uint8Array(jsonChunkBuffer, chunkLengthPrefix + jsonLength);
+                for (let i = 0; i < jsonPadding; ++i) {
+                    jsonPaddingView[i] = 0x20;
+                }
 
-                    //binary chunk
-                    const binaryChunkBuffer = new ArrayBuffer(chunkLengthPrefix);
-                    const binaryChunkBufferView = new DataView(binaryChunkBuffer);
-                    binaryChunkBufferView.setUint32(0, binaryBuffer.byteLength + imageByteLength + imagePadding, true);
-                    binaryChunkBufferView.setUint32(4, 0x004E4942, true);
+                //binary chunk
+                const binaryChunkBuffer = new ArrayBuffer(chunkLengthPrefix);
+                const binaryChunkBufferView = new DataView(binaryChunkBuffer);
+                binaryChunkBufferView.setUint32(0, binaryBuffer.byteLength + imageByteLength + imagePadding, true);
+                binaryChunkBufferView.setUint32(4, 0x004E4942, true);
 
-                    // binary padding
-                    const binPaddingBuffer = new ArrayBuffer(binPadding);
-                    const binPaddingView = new Uint8Array(binPaddingBuffer);
-                    for (let i = 0; i < binPadding; ++i) {
-                        binPaddingView[i] = 0;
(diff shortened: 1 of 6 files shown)


serializers/src/glTF/2.0/babylon.glTFExporter.ts
```diff
@@ -641,84 +651,88 @@ module BABYLON.GLTF2 {
...
-            jsonChunkBufferView.setUint32(0, jsonLength + jsonPadding, true);
-            jsonChunkBufferView.setUint32(4, 0x4E4F534A, true);
-
-            //json chunk bytes
-            const jsonData = new Uint8Array(jsonChunkBuffer, chunkLengthPrefix);
-            for (let i = 0; i < jsonLength; ++i) {
-                jsonData[i] = jsonText.charCodeAt(i);
-            }
+        public _generateGLBAsync(glTFPrefix: string): Promise<GLTFData> {
+            return new Promise((resolve, reject) => {
```
bghgary: This isn't the right way to use promises. You only need to construct a promise if you are adapting callbacks into promises. In this instance, you just need to return the result of `this.generateBinaryAsync().then(...)` and make sure the last continuation returns the `GLTFData`.

kcoley: Gotcha, made the change

bghgary: You need to fix this everywhere, not just here.

kcoley:
Ah makes sense, I’ll make a pass through everything (leaving the airport now)

kcoley:
Made the update to the other functions

serializers/src/glTF/2.0/babylon.glTFExporter.ts
```diff
@@ -591,36 +594,39 @@ module BABYLON.GLTF2 {
...
+            });
+
         }
 
         /**
          * Creates a binary buffer for glTF
          * @returns array buffer for binary data
          */
-        private generateBinary(): ArrayBuffer {
+        private generateBinaryAsync(): Promise<ArrayBuffer> {
```
bghgary: _generateBinaryAsync

serializers/src/glTF/2.0/babylon.glTFExporter.ts
```diff
@@ -937,13 +944,36 @@ module BABYLON.GLTF2 {
...
-                    uvCoordsPresent = false;
                     // go through all mesh primitives (submeshes)
                     for (const submesh of bufferMesh.subMeshes) {
+                        uvCoordsPresent = false;
+                        let babylonMaterial = submesh.getMaterial();
+
+                        let materialIndex: Nullable<number> = null;
+                        if (babylonMaterial) {
+                            if (babylonMaterial instanceof MultiMaterial) {
+                                const babylonMultiMaterial = babylonMaterial as MultiMaterial;
```
bghgary: this isn't necessary as typescript automatically makes babylonMaterial a MultiMaterial when you pass the if check

serializers/src/glTF/2.0/babylon.glTFExporter.ts
```diff
@@ -1075,62 +1052,68 @@ module BABYLON.GLTF2 {
          * @param babylonScene Babylon scene to get the mesh data from
          * @param binaryWriter Buffer to write binary data to
          */
-        private createScene(babylonScene: Scene, binaryWriter: _BinaryWriter): void {
+        private createSceneAsync(babylonScene: Scene, binaryWriter: _BinaryWriter): Promise<void> {
             if (this.setNodeTransformation.length) {
```
bghgary: I don't understand this line. What are you trying to check?

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -89,21 +89,26 @@ module BABYLON.GLTF2 {
...
-                    _GLTFMaterial._ConvertPBRMaterial(babylonMaterial, mimeType, images, textures, samplers, materials, imageData, hasTextureCoords);
+                    promises.push(_GLTFMaterial._ConvertPBRMaterialAsync(babylonMaterial, mimeType, images, textures, samplers, materials, imageData, hasTextureCoords));
                 }
                 else {
-                    Tools.Error("Unsupported material type: " + babylonMaterial.name);
+                    Tools.Warn(`Unsupported material type: ${babylonMaterial.name}`);
                 }
             }
+
+            return Promise.all(promises).then(() => {
```
bghgary: This looks like something is missing. Either put all in one line or put a comment like `// do nothing`

bghgary: Ditto everywhere else.

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -321,41 +329,48 @@ module BABYLON.GLTF2 {
             }
             if (hasTextureCoords) {
                 if (babylonStandardMaterial.diffuseTexture) {
-                    const glTFTexture = _GLTFMaterial._ExportTexture(babylonStandardMaterial.diffuseTexture, mimeType, images, textures, samplers, imageData);
-                    if (glTFTexture != null) {
-                        glTFPbrMetallicRoughness.baseColorTexture = glTFTexture;
-                    }
+                    let promise = _GLTFMaterial._ExportTextureAsync(babylonStandardMaterial.diffuseTexture, mimeType, images, textures, samplers, imageData, useAlpha).then(glTFTexture => {
+                        if (glTFTexture != null) {
```
bghgary: nit: everywhere else `if (glTFTexture)`

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -370,6 +385,40 @@ module BABYLON.GLTF2 {
...
+                    return Promise.reject(`Scene not available for texture ${texture.name}`);
+                }
+                else {
+                    const proceduralTexture = new ProceduralTexture('texture', texture.getSize(), 'setAlphaToOne', scene);
+                    if (proceduralTexture == null) {
+                        return Promise.reject(`Cannot create procedural texture for ${texture.name}!`);
+                    }
+                    else {
+                        proceduralTexture.setTexture('textureSampler', texture as Texture);
+                        return scene.whenReadyAsync().then(() => {
```
bghgary: This is an odd way to do it I think. Why not wait for just the texture to be ready?

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -370,6 +385,40 @@ module BABYLON.GLTF2 {
...
+            if (useAlpha) {
+                return Promise.resolve(texture as Texture);
+            }
+            else {
+                const scene = texture.getScene();
+                if (scene == null) {
+                    return Promise.reject(`Scene not available for texture ${texture.name}`);
+                }
+                else {
+                    const proceduralTexture = new ProceduralTexture('texture', texture.getSize(), 'setAlphaToOne', scene);
```
bghgary: Normally, you should wrap this like this`Promise.resolve().then(() => { ... })` to make sure any exceptions are captured by the promise, but if you end up using onLoadObservable, then just create a new Promise.

serializers/src/glTF/2.0/babylon.glTFExporter.ts
kcoley: I'll make that change.

serializers/src/glTF/2.0/babylon.glTFExporter.ts
kcoley: Thanks, that's good to know

serializers/src/glTF/2.0/babylon.glTFExporter.ts
kcoley: I don't know why I used that line there either...  I'm taking it out

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: Sure, I'll change those lines.

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: I'm not sure why I wrote it this way too, I'll change these lines

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: How do I wait for the texture to be ready?

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: ok, will do that

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -370,6 +383,40 @@ module BABYLON.GLTF2 {
...
+                        return Promise.reject(`Scene not available for texture ${texture.name}`);
+                    }
+                    else {
+                        const proceduralTexture = new ProceduralTexture('texture', texture.getSize(), 'setAlphaToOne', scene);
+                        if (proceduralTexture == null) {
+                            return Promise.reject(`Cannot create procedural texture for ${texture.name}!`);
+                        }
+                        else {
+                            proceduralTexture.setTexture('textureSampler', texture as Texture);
+                            return scene.whenReadyAsync().then(() => {
```
kcoley: @bghgary is there a way to know when a procedural texture has completed processing?  I only used this line since it was one way to know.   Otherwise I would get black textures on export.

bghgary: You should be able to use onLoadedObservable.

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: @bghgary made the change

bghgary: You need to put the whole function inside the new Promise, not just the observable.

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -394,29 +394,27 @@ module BABYLON.GLTF2 {
...
-                        }
-                        else {
-                            proceduralTexture.setTexture('textureSampler', texture as Texture);
-                            return scene.whenReadyAsync().then(() => {
-                                return proceduralTexture;
-                            });
-                        }
+                        proceduralTexture.setTexture('textureSampler', texture as Texture);
+                        return new Promise((resolve, reject) => {
+                            proceduralTexture.onLoadObservable.add(() => { return proceduralTexture });
```
bghgary: something needs to call resolve...

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -370,6 +383,40 @@ module BABYLON.GLTF2 {
...
+                else {
+                    const scene = texture.getScene();
+                    if (scene) {
+                        const proceduralTexture = new ProceduralTexture('texture', texture.getSize(), 'setAlphaToOne', scene);
+                        if (proceduralTexture == null) {
+                            reject(`Cannot create procedural texture for ${texture.name}!`);
+                        }
+                        else {
+                            proceduralTexture.setTexture('textureSampler', texture as Texture);
+                            proceduralTexture.onLoadObservable.add(() => { resolve(proceduralTexture) });
```
kcoley: @bghgary like this?

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -370,6 +383,41 @@ module BABYLON.GLTF2 {
...
+            return Promise.all(promises).then(() => { /* do nothing */ });
+        }
+
+        /**
+         * 
+         * @param texture Texture with alpha to overwrite to one
+         * @param useAlpha Specifies if alpha should be preserved or not
+         * @returns Promise with texture
+         */
+        public static _SetAlphaToOneAsync(texture: BaseTexture, useAlpha: boolean): Promise<Texture> {
```
kcoley: I made another slight refactor here to avoid checking for null equality explicitly ( == null)

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
bghgary: Yes.

PR #4319, opened 2018-05-14
do not merge try triggering vuild only

sebavan:


deltakosh:
Is that me or you did that from your phone??:D

sebavan:
trying to fix the build indeed :-)

Issue #3075, opened 2017-11-02
Add drag'n'drop support to viewer


deltakosh:
and then replace the sandbox code

PR #4329, opened 2018-05-16
Drag and drop

RaananW:
viewer now supports drag and drop and has it enabled per default.
That means - you can paste the model's files (including subfolders) and the viewer will display the model according to the predefined configuration.

To disable the feature you can use the param "disableDrapAndDrop" in the params of the viewer template. html example:

```html
<babylon templates.viewer.params.disable-drag-and-drop="true">...</babylon>
```

or JSON:

```javascript
{
    templates: {
        viewer: {
            params: {
                disableDragAndDrop: true
            }
        }
    }
}
```

Addressing https://github.com/BabylonJS/Babylon.js/issues/3075

### DIFF
--- a/Viewer/src/configuration/configuration.ts
+++ b/Viewer/src/configuration/configuration.ts
@@ -166,7 +166,7 @@ export interface IModelConfiguration {
     id?: string;
     url?: string;
     root?: string; //optional
-    file?: string; // is a file being loaded? root and url ignored
+    file?: string | File; // is a file being loaded? root and url ignored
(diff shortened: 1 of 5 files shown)


deltakosh:
Woot!!

sebavan:
Sweet

deltakosh:
Could it be off by default?

bghgary:
I would also recommend flipping the flag to `enableDragAndDrop`.

RaananW:
if it is off by default, then changing the key makes sense :)
I will do it.

PR #4330, opened 2018-05-16
add utility layer

TrevorDev:
See usage on this PG: #AANV2M#6

### DIFF
--- a/src/Rendering/babylon.utilityLayerRenderer.ts
+++ b/src/Rendering/babylon.utilityLayerRenderer.ts
@@ -1,0 +1,59 @@
+module BABYLON {
+    /**
+     * Renders a layer on top of an existing scene
+     */
+    export class UtilityLayerRenderer implements IDisposable {
+        /** 
+         * The scene that is rendered on top of the original scene
+         */ 
+        public utilityLayerScene:Scene;
+
+        /**
+         *  If the utility layer should automatically be rendered on top of existing scene
+        */
+        public shouldRender:boolean = true;
+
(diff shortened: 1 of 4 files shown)


src/babylon.scene.ts
```diff
@@ -1382,6 +1382,17 @@
             return this._boundingBoxRenderer;
         }
 
+        private _utilityLayerRenderer:UtilityLayerRenderer;
+        /** 
+         * Gets the UtilityLayerRenderer used to render an overlay scene layer without impacting the existing scene
+         */
+        public get utilityLayerRenderer(): UtilityLayerRenderer {
```
TrevorDev: Are you sure adding this is a good idea? Having this as a singleton on the scene may make developers mix too many things in the same utility layer. I would want to encourage multiple layers (eg. one for gizmo, one for gui, one for controllers, etc.)

src/Rendering/babylon.utilityLayerRenderer.ts
```diff
@@ -0,0 +1,67 @@
...
+         * The scene that is rendered on top of the original scene
+         */ 
+        public utilityLayerScene:Scene;
+
+        /**
+         *  If the utility layer should automatically be rendered on top of existing scene
+        */
+        public shouldRender:boolean = true;
+
+        private _afterRenderObservable:Nullable<Observer<Scene>>;
```
deltakosh: we should name them observer and not observable ;)

src/babylon.scene.ts
deltakosh: this makes sense as well. I like the idea of multiple layer actually. Keep it like that then

PR #4331, opened 2018-05-16
Adding MixMaterial to MaterialsLibrary

julien-moreau:
TODO: finish getAnimatables and Clone


dad72:
Why the PR, that is approved, has not been validated ? I can not test it today to send comments to Julien.
What is in TODO does not prevent the operation of the material and was planned I guess with another PR.

julien-moreau:
Just the code has been approved, I finalize this evening then it should be merged if
- you confirm that it works as you want
- or it’ll be merged and I’ll pull request next changes

dad72:
Ok, but I can not confirm that it works before it is merged.

dad72:
Oh ok, this because the 'WhatsNew' was not edited I guess for this new material.

dad72:
I just noticed that getAnimatables () and clone () are already present. Is this imcomplete? because when I compare with the fieldMateriel, everything seems finished.

By cons I'm not sure, but in function hasTexture(), there is only the _mixTexture1 and not the _mixTexture2. Same thing in the function dispose(), it only has the _mixTexture1. Is this normal?

I hope this PR will soon be merged.

dad72:
I tried to convert the ts file to js for testing. it works, but I have two problem:

- MixMap1 and MixMap 2 mix together creating a bad paint effect.
- And texture 4 and 8 does not give textures, but gives transparency between different layers.

I had idea to use 8 different textures. For example if on mixMap 1 I make earth, sand and rocks, then with mixMap2 I repaint snow on the peaks of mountains, I do not find myself with snow, but with a mix between the snow and rock which gives something ugly.

Basically, the textures from 1 to 8 can this mix, but not the two mixTextures. I do not know if I explain well. I hope.

Is there the possibility that mixMap1 and mixMap2 do not mix and that texture 4 and 8 create texture and not transparency.

I wonder if this should not be 6 textures that should instead of 8, because the alpha channel does not give a fourth texture, but an erase of RGB.

Otherwise I have no error, it works, apart from the mixture of 2 mixMap and alpha channel 

Maybe for alpha I have to use a different color than black which corresponds to an alpha of 1? that's what white would make an alpha of 0 ? I saw in the shader that the apha is 1.0 - mixColor.a.

Thank you Julien

julien-moreau:
That's right ! Then I should add an option for that (mixMap 2 overrides mixmap 1)

dad72:
No, I did not test like that, I would do it by testing with your patches.

julien-moreau:
@deltakosh do you want me to push the dist files as well ?

deltakosh:
No I will do it:)

Issue #4104, opened 2018-04-09
Add glTF tab to inspector


sbtron:
Add glTF in inspector with support for

- [x] <strike>Select animation group</strike> This is already done in the sandbox.
- [ ] <strike>Show validation errors</strike> This will be covered by #4094
- [x] Select which extension to load/display (e.g. use PBR_SpecGloss or Unlit) (PR #4336)
- [ ] <strike>Stats about the glTF file</strike> we have logs with some stats. can improve later if necessary.
- [x] Export glTF/glb

bghgary:
See https://bghgary.github.io/glTF/sandbox-bjs/ for example functionality for some of the items above.

PR #4336, opened 2018-05-17
Add glTF loader extension settings to GLTF tab in inspector

bghgary:
Also fixed a bug with importMesh in glTF loader and added unit tests.

See #4104


bghgary:
@deltakosh Please don't merge yet. Looking at a possible issue.

PR #4352, opened 2018-05-21
fire prePointerObservable with ray when using 6dof controller

TrevorDev:


src/babylon.scene.ts
```diff
@@ -1621,6 +1621,16 @@
          */
         public simulatePointerMove(pickResult: PickingInfo, pointerEventInit?: PointerEventInit): Scene {
             let evt = new PointerEvent("pointermove", pointerEventInit);
+            // PreObservable support
+            if (this.onPrePointerObservable.hasObservers() && !this._pointerCaptures[evt.pointerId]) {
```
deltakosh: why adding this here? It will be done inside processPointerMove (I guess, not checked)

TrevorDev: Not that I've been able to find.
Simulated path: vrHelper (Picking) -> simulatePointerMove (preObs) -> _processPointerMove
Pointer path: canvasAddListener -> _onPointerMove (preObs, picking) -> _processPointerMove

I was debating moving the preObservable logic to the _processPointerMove to avoid the duplication of code but then decided to just make all the simulatePointer methods mirror the onPointerMethods.

PR #4371, opened 2018-05-24
Bunch of small fixes

bghgary:
* Allow reloading when passing in asset url to sandbox 
* Add serializers to public sandbox to allow export to glTF in inspector 
* Add GLTF supported checks to inspector 
* Fixes for enabling/disabling glTF loader extensions in the inspector


deltakosh:
There are some conflicts

Issue #4374, opened 2018-05-24
Update Font Awesome


Gil-1:
When using @fortawesome/fontawesome >= 5.0 in the project the debug layer doesn't work as some icons are not found.

Maybe an update of fontawesome would just resolve it ?

Gil-1:
Sweet, thanks :)

Jaskar:
Hi @Gil-1 ,

I'll do a PR today.

It brokes a lot of icons, so I hope it's ok.
I you found any missing icon let me know.

Gil-1:
Great and thanks !
I will test it as soon as it's out :)

Issue #4369, opened 2018-05-24
Hash symbol appended to url clicking on a footer button in sandbox


bghgary:
Repro:
1. Open https://sandbox.babylonjs.com
2. Drag and drop an asset
3. Click on the fullscreen or debug layer button

Result:
`#` is appended to the browser's url

Expect:
no `#` is appended

Remarks:
This causes an issue when passing assetUrl to the sandbox and then refreshing the page.
For example: https://sandbox.babylonjs.com/?assetUrl=https://models.babylonjs.com/Alien.glb

PR #4381, opened 2018-05-25
Sandbox #4369 - Inspector #4374 + Edge compatibility

Jaskar:
Sandbox issue : https://github.com/BabylonJS/Babylon.js/issues/4369
-> Correction ok

Inspector issue : https://github.com/BabylonJS/Babylon.js/issues/4374
-> Replacement ok

Inspector Edge : 
When oppening on Edge, it now display something. (was empty before).

I tried to add the "Open in window" like on Chrome, but face an issue : When you use "window.open" to open a popup in Edge, you can't modify the content.
This link : https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/15755797/
say that it's fixed, but not on my version (42.17134.1.0).
Can someone test the code on the issue page on the Edge console and tell me what happen please :)


deltakosh:
I still have the issue for edge. The popup can't be changed...
Can you get back to the bug on edge site and ask them what else should be done?

PR #4389, opened 2018-05-26
gizmo manager

TrevorDev:
- Gizmo manager
- Handle occlusion and picking on the utility layer


Tools/Gulp/config.json
```diff
@@ -1028,7 +1028,8 @@
                 "../../src/Gizmos/babylon.planeRotationGizmo.js",
                 "../../src/Gizmos/babylon.positionGizmo.js",
                 "../../src/Gizmos/babylon.rotationGizmo.js",
-                "../../src/Gizmos/babylon.scaleGizmo.js"
```
deltakosh: I think we should create a workload for gizmos
Also utilitylayer should be independent and gizmos should have a ref on it

src/Gizmos/babylon.gizmoManager.ts
```diff
@@ -0,0 +1,80 @@
...
+         */
+        public dispose(){
+            this.scene.onPointerObservable.remove(this._pointerObserver);
+            this._clearGizmos();
+            this._gizmoLayer.dispose();
+        }
+
+        private _clearGizmos(){
+            for(var key in this._gizmoSet){
+                this._gizmoSet[key].positionGizmo.dispose();
```
deltakosh: Need to add OwnProperty check

PR #4390, opened 2018-05-26
Update babylon.mixMaterial.ts

dad72:
Fix this.mixTexture2 by this._mixTexture2


julien-moreau:
@deltakosh is the « what’s new » update necessary ?

dad72:
I do not think so, it has already been done.

I made you a report on the Julien forum about the problems with this material.

julien-moreau:
Yeah I read that, i’ll Take time this weekend to fix and add options. I noticed the dark glitch and I have an idea how to fix it

Issue #4339, opened 2018-05-18
Break down the viewer to importable modules


RaananW:
The viewer should be separated to a few (UMD/es6-based) modules, mainly:

* Templating / UI - for handling HTML-based UI elements (including handlebars)
* Scene Manager - for handling the scene creating and (scene-related) configuration loading 
* Model Loader and Viewer Model - model container and loaders-integration
* Labs - for experimental features and features that might change till the next stable build.
* Viewers - The abstract viewer and the default viewer (that use the other modules)

The viewer should also allow injecting your own version of Babylon (in case you use a custom build)

* [x] reduce dependencies in main classes
* [x] create assets-package that could potentially be loaded. 
* [x] make changes to the npm package (include sources), making sure preview builds stay the same.

deltakosh:
Can you evaluate a potential ETA?

RaananW:
[] reduce dependencies in main classes
[] create assets-package that could potentially be loaded. 
[] make changes to the npm package (include sources), making sure preview builds stay the same.

RaananW:
I will close the ticket after making sure the npm package published correctly

PR #4409, opened 2018-05-31
minimizing dependencies for the viewer

RaananW:
Initial PR for https://github.com/BabylonJS/Babylon.js/issues/4339

This PR divides the viewer to a few modules and minimizing the dependencies between them.
Loading the sceneManager (which in turn will configure a scene) will not load the template manager.

The next steps:
* Separating the assets to a different module and only load them when needed
* change the npm package to contain the js and d.ts files, so that the developer will be able to load individual classes and their dependencies only (tree shaking for the viewer only).

Using webpack's externals/alias the developer will also be able to use a different babylon version.


RaananW:
Working on fixing the build errors.

PR #4418, opened 2018-06-01
glTF Serializer: fix for lines mesh indices

kcoley:
- Checks if the indices of a submesh is greater than the indices of the main mesh
- Forces alpha to number in cases where it writes out as string 
- Encodes the name of the nodes to be UTF-8 friendlier


kcoley:
Closing this as it is not an issue with the exporter but with the authored Babylon scene.

PR #4424, opened 2018-06-04
Adding way to serialize all materials in SceneSerializer.SerializeMesh

julien-moreau:
This is one solution to force serializing all material types when calling SceneSerializer.SerializeMesh
Do you thing removing like 68 may not create breaking change ? It looks like materials were already serialized but only standard materials


deltakosh:
We can remove it to be fair. All material should be serialized

julien-moreau:
Cool! Updating the PR :)

julien-moreau:
Updated :)

Issue #4079, opened 2018-04-05
 DebugLayer onChange event


devAxeon:
[Here](http://www.html5gamedevs.com/topic/36846-debuglayer-onchange-event/) is the forum's link where I posted the question, but all is in the title.

Jaskar:
Hi @devAxeon ,

I don't really understand what you expect. (Maybe because english is not my native language, or because it's getting late here ...)

Can you explain it please ?
Do you have any exemple ?

Thanks :)

devAxeon:
Salut @Jaskar, 

En fait je pensais à truc dans le genre:
`BABYLON.debugLayer.onGlobalPropertyChange.add((result) => {
        console.log(result.object); // The modified object
        console.log(result.property); // The proeprty changed (as string)
        console.log(result.value); // New value after change
        console.log(result.initialValue); // Value before change
});`
C'est un Event déclenché sur les changements dans le debug layer. Si par exemple je me sert du débug layer pour changer la position Z d'un mesh, je voudrais qu'un event me prévienne et me donne le maximium d'info sur cette modification.
`{
        object: myMeshObj,
        property: 'position.z', //Il y a peut-être une meilleur façon de représenter ça...
        value: 2,
        initialValue: 1.7
}`

Jaskar:
Ping @devAxeon
https://github.com/BabylonJS/Babylon.js/pull/4460

Is this what you're looking for ? :)

devAxeon:
Exactly !! Many thanks @Jaskar !!

Jaskar:
I tested it on the Inspector test scene, but it must be tested on real scenes to be sure there are no errors.
Ping me if you find one.

PR #4460, opened 2018-06-07
Inspector : #4079 - DebugLayer onChange event

Jaskar:
You can now write :
scene.debugLayer.onGlobalPropertyChange.push((result) => {
    console.log(result.object);
    console.log("Property : " + result.property);
    console.log("New value : " + result.value);
    console.log("Old value : " + result.initialValue);
});


deltakosh:
I would have used “Changed” instead of “Change”
Also do you mind updating the what’s new and the inspector doc?

Jaskar:
https://github.com/BabylonJS/Editor/issues/48
They used "editor.core.onGlobalPropertyChange", so I used the same.

Aaaand I just notice that they used "add" instead of my "push", so I'll fix that too!

deltakosh:
It has to be an Observable so the name for bjs must be:
onPropertyChangedObservable

inspector/src/details/Property.ts
```diff
@@ -1,41 +1,84 @@
...
+                if (this._obj instanceof BABYLON.Scene) {
+                    (<BABYLON.Scene>this._obj).debugLayer.onGlobalPropertyChangeCallback({
+                        object: this._obj,
+                        property: this._property,
+                        value: newValue,
+                        initialValue: (<any>this._obj)[this._property]
+                    });
+                }
+                else {
+                    if(this._parentObj != null) {
```
bghgary: This is a confusing way of getting back to the debug layer. It would be cleaner if the observable was on the inspector and the inspector is passed into the property instead. The debug layer can then expose a separate observable that hooks to the inspector.

Jaskar:
I change it in this PR
https://github.com/BabylonJS/Babylon.js/pull/4464

inspector/src/details/Property.ts
bghgary: I'm not sure I follow. #4464 just renames the property?

bghgary:
Right, but that's the issue I'm describing. I'm saying the mechanism for getting to the debug layer from a `Property` is awkward.

PR #4476, opened 2018-06-08
glTFSerializer: Refactor alpha mode handling

kcoley:
- Simplify the alpha mode calculation
- Use `onGenerated` for `ProceduralTexture`  instead of `onLoadObservable`

### DIFF
--- a/serializers/src/glTF/2.0/babylon.glTFMaterial.ts
+++ b/serializers/src/glTF/2.0/babylon.glTFMaterial.ts
@@ -238,68 +238,12 @@ module BABYLON.GLTF2 {
          * @param babylonMaterial Babylon Material
          * @returns The Babylon alpha mode value
          */
-        public static _GetAlphaMode(babylonMaterial: Material): Nullable<MaterialAlphaMode> {
-            if (babylonMaterial instanceof StandardMaterial) {
-                const babylonStandardMaterial = babylonMaterial as StandardMaterial;
-                if ((babylonStandardMaterial.alpha !== 1.0) ||
-                    (babylonStandardMaterial.diffuseTexture != null && babylonStandardMaterial.diffuseTexture.hasAlpha) ||
-                    (babylonStandardMaterial.opacityTexture != null)) {
-                    return MaterialAlphaMode.BLEND;
-                }
-                else {
-                    return MaterialAlphaMode.OPAQUE;
-                }
+        public static _GetAlphaMode(babylonMaterial: Material): MaterialAlphaMode {
+            if (babylonMaterial.needAlphaBlending()) {
+                return babylonMaterial.needAlphaTesting() ? MaterialAlphaMode.MASK : MaterialAlphaMode.BLEND;
             }
-            else if (babylonMaterial instanceof PBRMetallicRoughnessMaterial) {
-                const babylonPBRMetallicRoughness = babylonMaterial as PBRMetallicRoughnessMaterial;
 
-                switch (babylonPBRMetallicRoughness.transparencyMode) {
-                    case PBRMaterial.PBRMATERIAL_OPAQUE: {
-                        return MaterialAlphaMode.OPAQUE;
-                    }
-                    case PBRMaterial.PBRMATERIAL_ALPHABLEND: {
-                        return MaterialAlphaMode.BLEND;
-                    }
-                    case PBRMaterial.PBRMATERIAL_ALPHATEST: {
-                        return MaterialAlphaMode.MASK;
-                    }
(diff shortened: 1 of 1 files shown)


serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -238,68 +238,12 @@ module BABYLON.GLTF2 {
...
-                        return MaterialAlphaMode.BLEND;
-                    }
-                    default: {
-                        Tools.Error("Unsupported alpha mode " + babylonPBRMetallicRoughness.transparencyMode);
-                        return null;
-                    }
-                }
+        public static _GetAlphaMode(babylonMaterial: Material): MaterialAlphaMode {
+            if (babylonMaterial.needAlphaBlending()) {
+                return babylonMaterial.needAlphaTesting() ? MaterialAlphaMode.MASK : MaterialAlphaMode.BLEND;
```
bghgary: This logic doesn't look right. If `needAlphaBlending()` returns true, it should always be blend. This logic should be identical to [this code](https://github.com/BabylonJS/Babylon.js/blob/master/src/Rendering/babylon.renderingGroup.ts#L328).

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -419,22 +363,18 @@ module BABYLON.GLTF2 {
...
-                            proceduralTexture.setTexture('textureSampler', texture as Texture);
-                            proceduralTexture.onLoadObservable.add(() => { resolve(proceduralTexture) });
-                        }
-                        else {
-                            reject(`Cannot create procedural texture for ${texture.name}!`);
-                        }
+
+                        proceduralTexture.setTexture('textureSampler', texture as Texture);
+                        proceduralTexture.onGenerated = () => {    
+                            resolve(proceduralTexture)
```
bghgary: nitpick: missing semicolon

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -419,22 +363,18 @@ module BABYLON.GLTF2 {
...
-                        
-                        if (proceduralTexture) {
-                            proceduralTexture.setTexture('textureSampler', texture as Texture);
-                            proceduralTexture.onLoadObservable.add(() => { resolve(proceduralTexture) });
-                        }
-                        else {
-                            reject(`Cannot create procedural texture for ${texture.name}!`);
-                        }
+
+                        proceduralTexture.setTexture('textureSampler', texture as Texture);
```
bghgary: There are many casts from `BaseTexture` to `Texture` in this function. Are these all necessary? Perhaps the function should take a `Texture` instead of `BaseTexture`?

kcoley: @bghgary I just refactored to use `Texture` instead

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: @bghgary I added the missing semicolon

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
kcoley: @bghgary I refactored to match the logic in the code section.

serializers/src/glTF/2.0/babylon.glTFMaterial.ts
```diff
@@ -997,7 +941,7 @@ module BABYLON.GLTF2 {
                 return null;
             }
 
-            const metallicRoughnessFactors = this._ConvertSpecularGlossinessTexturesToMetallicRoughness(babylonPBRMaterial.albedoTexture, babylonPBRMaterial.reflectivityTexture, specGloss, mimeType);
+            const metallicRoughnessFactors = this._ConvertSpecularGlossinessTexturesToMetallicRoughness(babylonPBRMaterial.albedoTexture as Texture, babylonPBRMaterial.reflectivityTexture as Texture, specGloss, mimeType);
```
bghgary: Are these textures guaranteed to be `Texture` and not something else that derives from `Texture`? Something to think about.

kcoley: Gotcha, I'll refactor this

kcoley: Just made the refactor to remove casts from BaseTexture to Texture

PR #4496, opened 2018-06-12
disable alpha mode tests for glTF Serializer

kcoley:
- Fixes a bug with unit tests for glTF alpha mode


kcoley:
@deltakosh @bghgary I will disable these alpha mode unit tests for now and investigate further.

PR #4502, opened 2018-06-12
New HDButton

RaananW:
The viewer has now a new HD button that is added per default to the navbar.
The HD button will increase (and decrease back) the hardware scaling rate so that the object will have better quality (of course with a relative degrade in performance).

It is possible to hide the HD button, exactly like the fullscreen or help button.
It is also possible to set hd to be enabled from the beginning using configuration: 
`{ engine: { hdEnabled: true } }`

Also in this PR - i18n for the buttons. The values are in:
`templates.nav-bar.params.text.{button}` where button can be fullscreenButton, helpButton, hdButton. See the default configuration for further details.


RaananW:
The HD button will be shown per default. Any public objections?

PR #4505, opened 2018-06-12
Updating meshButton3D with feedback

davrous:


gui/src/3D/controls/meshButton3D.ts
```diff
@@ -0,0 +1,75 @@
...
+            return this._currentMesh;
+        }
+
+        protected _affectMaterial(mesh: AbstractMesh) {
+        }
+
+        /**
+         * Releases all associated resources
+         */
+        public dispose() {
```
deltakosh: you can completely remove the function actually :)

PR #4531, opened 2018-06-14
feat: enable MediaStreams to handle WebRTC sound streams

menduz:
We had a hack for this.

```ts
function connectStream(stream: MediaStream){
  const sound = new BABYLON.Sound('avatar-stream', null, scene, null, {
   streaming: true,
   spatialSound: true
  })

  sound['_htmlAudioElement'] = new Audio()
  sound['_isReadyToPlay'] = true
  audioElement.srcObject = stream

  document.body.appendChild(audioElement)

  sound.play()

  return sound
}
```

This should enable us to use native WebRTC streams directly

```ts
function connectStream(stream: MediaStream){
  return new BABYLON.Sound('avatar-stream', stream, scene, null, {
   streaming: true,
   autoplay: true,
   spatialSound: true
  })
}
```


.vscode/settings.json
```diff
@@ -43,5 +43,6 @@
         "**/*.d.ts": true,
         "assets": true
     },
+    "editor.tabSize": 4,
```
deltakosh: nope :)

src/Audio/babylon.sound.ts
```diff
@@ -98,11 +98,25 @@ module BABYLON {
                         if (typeof (urlOrArrayBuffer) === "string") this._urlType = "String";
                         if (Array.isArray(urlOrArrayBuffer)) this._urlType = "Array";
                         if (urlOrArrayBuffer instanceof ArrayBuffer) this._urlType = "ArrayBuffer";
+                        if (urlOrArrayBuffer instanceof MediaStream) this._urlType = "MediaStream";
```
deltakosh: should use else here

.vscode/settings.json
menduz: Maybe this should not be part of this PR. Several files seem to use four spaces. Does it make sense?

deltakosh: Well actually you are right this is ok just to have everything aligned

src/Audio/babylon.sound.ts
menduz: I don't agree. If we remove the `if` and replace it with an `else` statement that would make `_urlType = MediaStream` in the case of not handled types like `boolean`.

By default `_urlType` is `"Unknown"`. That counts for `else`

menduz: Ohh. Should I add that to the rest of the `if` statements too?

deltakosh: Yep please:)

menduz:
@deltakosh please review

deltakosh:
Please just update the what’s new in dist/preview release?

menduz:
Done

deltakosh:
You rock!!

menduz:
🎉

Issue #4413, opened 2018-06-01
[Viewer] VR Button for the viewer


RaananW:
Add a VR button to the viewer's default template.

I will be using the VR experience helper. the ground will be the floor, so the user can teleport. Normalized models will always be the same size (1 unit / 1 meter)

Requires #4338

deltakosh:
Can we consider it done?

RaananW:
yes, initial vr support is integrated.
The next step would be interaction with the model itself (rotating, scaling etc').

PR #4537, opened 2018-06-15
initial VR support for the viewer

RaananW:
* A new VR helper will be created when the viewer was done initializing.
* For now there is no UI element to enable this.
* The function toggleVR() will toggle between the two modes.
* In VR mode the object AND environment will be scaled to fit the VR perspective.
* In VR the object will float in front of the user's eyes.
* For now there is no option to move or scale the object in VR.
* Position tracking is turned on and also teleportation on the ground. User can go around the object.

Initial commit for https://github.com/BabylonJS/Babylon.js/issues/4413

### DIFF
--- a/Viewer/src/viewer/viewer.ts
+++ b/Viewer/src/viewer/viewer.ts
@@ -1,7 +1,7 @@
 import { viewerManager } from './viewerManager';
 import { SceneManager } from '../managers/sceneManager';
 import { ConfigurationLoader } from '../configuration/loader';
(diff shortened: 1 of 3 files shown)


RaananW:
will fix tests soon

Viewer/src/viewer/viewer.ts
```diff
@@ -198,6 +198,7 @@ export abstract class AbstractViewer {
         this.onInitDoneObservable.add(() => {
             this._isInit = true;
             this.engine.runRenderLoop(this._render);
+            this._prepareVR();
```
deltakosh: Is it something we want all the time?

deltakosh: So we have to make sure not to ship it before we can control it as it will change the default behavior else

RaananW: agreed, I will continue in this branch

RaananW:
Waiting for this - https://github.com/BabylonJS/Babylon.js/pull/4585 to fix post-process darkness

PR #4578, opened 2018-06-20
Template location ignored if html if defined

RaananW:
The template location is ignored (mainly since the elimination of IDs in the viewer).
This fix provides two changes:
1) When html is an empty string, location will be used.
2) DOM Mapper will parse undefined and null correctly. so setting a parameter to "undefined" (as string) will set it to undefined (the type)


Viewer/src/configuration/mappers.ts
```diff
@@ -52,6 +52,10 @@ class HTMLMapper implements IMapper {
                         val = true;
                     } else if (val === "false") {
                         val = false;
+                    } else if (val === "undefeind") {
```
sebavan: are you sure it should not be undefined ?

RaananW: yes I am. this is the mapper, which maps strings from the dom attributed to real objects

RaananW: so people will be able to set elements to be undefined (for example - templates.nav-bar.html="undefined")

sebavan: but undefeind and not undefined looks like a typo ?

PR #4584, opened 2018-06-20
Fixing particle systems parse function

julien-moreau:
minEmitBox etc. are now part of the attached emitter (typically box emitter). These properties must be parsed only for backward compatibility and will not be present in the JSON data produced by .serialize() today

### DIFF
--- a/src/Particles/babylon.particleSystem.ts
+++ b/src/Particles/babylon.particleSystem.ts
@@ -1593,10 +1593,19 @@
                 particleSystem.preventAutoStart = parsedParticleSystem.preventAutoStart;
             }
 
-            particleSystem.minEmitBox = Vector3.FromArray(parsedParticleSystem.minEmitBox);
-            particleSystem.maxEmitBox = Vector3.FromArray(parsedParticleSystem.maxEmitBox);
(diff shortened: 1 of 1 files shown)


src/Particles/babylon.particleSystem.ts
```diff
@@ -1593,10 +1593,19 @@
                 particleSystem.preventAutoStart = parsedParticleSystem.preventAutoStart;
             }
 
-            particleSystem.minEmitBox = Vector3.FromArray(parsedParticleSystem.minEmitBox);
-            particleSystem.maxEmitBox = Vector3.FromArray(parsedParticleSystem.maxEmitBox);
-            particleSystem.direction1 = Vector3.FromArray(parsedParticleSystem.direction1);
-            particleSystem.direction2 = Vector3.FromArray(parsedParticleSystem.direction2);
+            if (parsedParticleSystem.minEmitBox) {
```
deltakosh: this should not be necessary
Only removing the 4 lines should be required. The ParticleSystem._Parse will take care of back compat

julien-moreau:
Il not sure that I should update the what’s new because it wasn’t a bug existing in the previous release

deltakosh:
It is fine:)

PR #4585, opened 2018-06-21
add/remove camera methods in the default pipeline

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4565

PR #4588, opened 2018-06-21
Setting optional fullscreen element

RaananW:
You can now set the element that will go fullscreen when clicking on the fullscreen button.
This is used when other elements were added that are not a part of the viewer element.

~Pay attention that certain css definitions relies on `viewer:fullscreen` and should be set by the user instead!~


RaananW:
before merging i want to fix the fullscreen css issue

RaananW:
Waiting for this - https://github.com/BabylonJS/Babylon.js/pull/4585 to fix post-processing darkness issue

PR #4603, opened 2018-06-23
New bounding box features

TrevorDev:


src/Rendering/babylon.utilityLayerRenderer.ts
```diff
@@ -59,6 +59,7 @@ module BABYLON {
 
                 let pointerEvent = <PointerEvent>(prePointerInfo.event);
                 if (originalScene!.isPointerCaptured(pointerEvent.pointerId)) {
+                    this._pointerCaptures[pointerEvent.pointerId] = false;
```
TrevorDev: Double check that adding this logic here is correct. It is needed because certain pointer up that were captured by the original scene left the utility layer pointer captured to true, resulting in the next pointer down on the utility layer to be ignored.

src/Gizmos/babylon.boundingBoxGizmo.ts
```diff
@@ -133,6 +163,7 @@ module BABYLON {
                         _dragBehavior.moveAttached = false;
                         box.addBehavior(_dragBehavior);
                         _dragBehavior.onDragObservable.add((event)=>{
+                            this.dragObservable.notifyObservers({});
```
TrevorDev: Is there a way to avoid the {} but also keep the option open to passing back additional data in the future?

PR #4614, opened 2018-06-26
Do not validate program if linking failed

crazyjul:
Program validation should be executed for all programs.

In previous case, the validation occurred if link phase failed. Then the error would always be "program not linked", hiding the real error.

### DIFF
--- a/src/Engine/babylon.engine.ts
+++ b/src/Engine/babylon.engine.ts
@@ -3537,7 +3537,16 @@
             var linked = context.getProgramParameter(shaderProgram, context.LINK_STATUS);
 
             if (!linked) {
-                context.validateProgram(shaderProgram);
+                var error = context.getProgramInfoLog(shaderProgram);
+                if (error) {
+                    throw new Error(error);
+                }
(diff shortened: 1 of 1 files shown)


src/Engine/babylon.engine.ts
```diff
@@ -3537,7 +3537,16 @@
...
             if (!linked) {
-                context.validateProgram(shaderProgram);
+                var error = context.getProgramInfoLog(shaderProgram);
+                if (error) {
+                    throw new Error(error);
+                }
+            }
+
+            context.validateProgram(shaderProgram);
+            var validated = context.getProgramParameter(shaderProgram, context.LINK_STATUS);
```
deltakosh: why doing it twice?

crazyjul: The first time, you get the link status, the second time, the validation status. Both might output different messages.

crazyjul: oh, I see the mistake here, it should be context.VALIDATE_STATUS

PR #4616, opened 2018-06-27
Add support for experimental EXT_lights_imageBased to glTF loader

bghgary:
cc @MiiBond


src/Tools/babylon.environmentTextureTools.ts
```diff
@@ -335,13 +335,45 @@ module BABYLON {
...
 
+            const imageData = new Array<Array<ArrayBufferView>>(mipmapsCount);
+            for (let i = 0; i < mipmapsCount; i++) {
+                imageData[i] = new Array<ArrayBufferView>(6);
+                for (let face = 0; face < 6; face++) {
+                    const imageInfo = specularInfo.mipmaps[i * 6 + face];
+                    imageData[i][face] = new Uint8Array(arrayBuffer, specularInfo.specularDataPosition! + imageInfo.position, imageInfo.length);
+
+
+                    // {
```
bghgary: will remove

src/Tools/babylon.environmentTextureTools.ts
```diff
@@ -394,24 +426,24 @@ module BABYLON {
...
                     for (let i = 0; i < mipSlices; i++) {
                         //compute LOD from even spacing in smoothness (matching shader calculation)
                         let smoothness = i / (mipSlices - 1);
                         let roughness = 1 - smoothness;
-    
+
                         let minLODIndex = offset; // roughness = 0
-                        let maxLODIndex = Scalar.Log2(info.width) * scale + offset; // roughness = 1
-    
+                        let maxLODIndex = mipmapsCount * scale + offset; // roughness = 1
```
bghgary: will fix this to `mipmapsCount - 1`

src/Engine/babylon.engine.ts
```diff
@@ -5857,30 +5857,45 @@
             this._bindTextureDirectly(this._gl.TEXTURE_CUBE_MAP, null);
 
             // this.resetTextureCache();
+
             texture.isReady = true;
+
+            texture.onLoadedObservable.notifyObservers(texture);
```
deltakosh: should not be in the update function else it will be called multiple times
(Checking isReady perhaps?)

src/Engine/babylon.engine.ts
```diff
@@ -5905,8 +5920,32 @@
...
+                            onLoad();
+                        }
+                    }, reason => {
+                        if (onError) {
+                            onError("Failed to upload raw cube texture data to the GPU", reason);
+                        }
+                    });
+                }
+                else {
+                    Tools.SetImmediate(() => {
```
deltakosh: why setImmediate ?

src/Materials/Textures/babylon.rawCubeTexture.ts
```diff
@@ -0,0 +1,46 @@
+module BABYLON {
+    export class RawCubeTexture extends CubeTexture {
```
deltakosh: no comment on the class

src/Engine/babylon.engine.ts
bghgary: hmm, perhaps. maybe i should just leave it the way it was before and not call onLoadedObservable. thoughts?

src/Engine/babylon.engine.ts
bghgary: I was originally going to use the onLoadedObservable to get notified when the texture is ready which would require updateRawCubeTexture to not fire the onLoadedObservable until after one frame to allow adding a callback in time. But I've added a onLoad callback just like createCubeTexture, so I don't need to do this anymore.

deltakosh: Perfect :)

src/Engine/babylon.engine.ts
deltakosh: onLoaded is fine. But we probalby just need to call it only if texture.isReady was false

PR #4620, opened 2018-06-27
"Fix" for boundingSphere Radius

Pryme8:
A better approximation for BS radius.

### DIFF
--- a/src/Culling/babylon.boundingSphere.ts
+++ b/src/Culling/babylon.boundingSphere.ts
@@ -2,6 +2,7 @@
     export class BoundingSphere {
         public center: Vector3;
         public radius: number;
+        public realRadius : number;
         public centerWorld: Vector3;
         public radiusWorld: number;
         public minimum: Vector3;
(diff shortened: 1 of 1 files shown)


Pryme8:
I think this should fix the problem people are having with getting a true BoundingSphere radius.

src/Culling/babylon.boundingSphere.ts
```diff
@@ -26,6 +27,21 @@
         public reConstruct(min: Vector3, max: Vector3) {
             this.minimum = min.clone();
             this.maximum = max.clone()
+            
+            var _min = Number.POSITIVE_INFINITY;
+            var _max = Number.NEGATIVE_INFINITY;
+            var _va = ['x','y','z'];
+    
+            for(var i=0; i<3; i++){
+                if(min[_va[i]]<_min){
```
deltakosh: this seems wrong to me: the distance cannot be on different axis. How the distance between x and z could work for isntance? we have to evaluate the overall distance

jbousquie:
Not sure this should be merged because of this explanation : http://www.html5gamedevs.com/topic/38508-bounding-sphere-radius-and-point-intersection/?do=findComment&comment=219771

I think the current bsphere radius is just correct.

Pryme8:
Gotcha, yeah It was a misunderstanding in what the bounding radius was.

PR #4621, opened 2018-06-27
internal radius retry

Pryme8:
Fixing some of the purposed changes, after input from others.

### DIFF
--- a/src/Culling/babylon.boundingSphere.ts
+++ b/src/Culling/babylon.boundingSphere.ts
@@ -2,6 +2,7 @@
     export class BoundingSphere {
         public center: Vector3;
         public radius: number;
+        public internalRadius : number;
         public centerWorld: Vector3;
         public radiusWorld: number;
         public minimum: Vector3;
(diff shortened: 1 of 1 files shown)


Pryme8:
http://www.html5gamedevs.com/topic/38508-bounding-sphere-radius-and-point-intersection/?do=findComment&comment=219771

based off of.

Pryme8:
but then again, I am torn between this and doing a:
max.subtract(min).scale(0.5) and having it be something you can check the radius per axis.  

Not sure which one will be of more use, but this seems to be a common request to find a "true" mesh bounding sphere.

deltakosh:
Well still this is just for spherical mesh so I don't think it should be in the framework

Pryme8:
then what about a vector3 from bb.max.subtract(bb.min).scale(0.5);

Then we could add a method like mesh.isSphere() or something where if all the axes are the same it returns true, and then have another one that is like mesh.getBoundingRadius() that checks the mesh.isSphere and returns a vector3 if its not and if it is returns a float.

I feel like this is a re-occurring request so a solution should be found.

Pryme8:
Feel free to close this one then as well, Ill re-think this and come back with a more thought out approach.
Maybe exclude it completely like you are saying and just write up a snippet page explaining how to get internal mesh radii

deltakosh:
I'm more on the snippet side:)

Pryme8:
Same, now that I kinda thought about it more.

Issue #4622, opened 2018-06-27
Running twice a PG with gizmo and default utilitylayer does not work


deltakosh:
Just run this PG twice: (with the run button)
https://www.babylonjs-playground.com/#IXNS0D

PR #4624, opened 2018-06-27
Observable handle remove during notify

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4622

### DIFF
--- a/src/Tools/babylon.observable.ts
+++ b/src/Tools/babylon.observable.ts
@@ -217,8 +217,7 @@
             var index = this._observers.indexOf(observer);
 
             if (index !== -1) {
-                observer._willBeUnregistered = true;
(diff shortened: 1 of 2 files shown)


src/Tools/babylon.observable.ts
```diff
@@ -273,7 +273,9 @@
             state.skipNextObservers = false;
             state.lastReturnValue = eventData;
 
-            for (var obs of this._observers) {
+            // Iterate over copy of array to handle removed observers during callback
+            var observerList = this._observers.slice();
```
deltakosh: This cannot be kept. We cannot create an array on every call to observables. This will put too much pressure on GC as we extensively use observables
The pattern here is to either set observer.unregisterOnNextCall = true or call addOnce

Issue #4623, opened 2018-06-27
Some bugs for gizmos


deltakosh:
- [x] Make behavior and axis meshes public
- [x] The gizmo scale should not change when moving: https://www.babylonjs-playground.com/#IXNS0D#1
- [x] The green rotation torus seems to lose the control when trying to rotate it: https://www.babylonjs-playground.com/#31M2AP#7
- [x] Scaling to the smaller size reverse the position of the control cube: https://www.babylonjs-playground.com/#DEYAQ5#47
- [x] Translation with mouse (moving the box) seems weird. Does not stay on XY axis: https://www.babylonjs-playground.com/#DEYAQ5#47
 - [x] Nested meshes not working https://www.babylonjs-playground.com/#DEYAQ5#51 
 - [x] Z-drag in sixdofdrag behavior not working as expected  https://www.babylonjs-playground.com/#DEYAQ5#51

deltakosh:
Createpickingray expects a screen coordinates so 0,0 is the upper left corner of the screen. Something more on the center of the screen should work better

TrevorDev:
Shouldn't any screen pick's ray start at or travel through the camera's position?

deltakosh:
nope as you are projected on near plane

PR #4641, opened 2018-06-29
use alternate picked point when rotating along drag plane

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4623


src/Math/babylon.math.ts
```diff
@@ -4564,11 +4564,13 @@
 
         /**
          * Extracts the rotation matrix from the current one and sets it as the given "result"  
+         * @param ignoreScaling when set, do not let negative scale matrix values modify the rotation matrix (Default: false)
```
deltakosh: Seems suspicious :)

TrevorDev: Yea, to me it was odd that scale would effect the rotation matrix. Originally I added this parameter, now I removed it but am setting scale to positive on the mesh prior to calling getRotationMatrix in the gizmo class.

src/Math/babylon.math.ts
```diff
@@ -4568,7 +4568,8 @@
          * @returns the current matrix  
          */
         public getRotationMatrixToRef(result: Matrix): Matrix {
-            var m = this.m;
+            var matrix:Matrix = this;
```
deltakosh: Why do we need matrix parameter?

PR #4657, opened 2018-07-03
Bounding box nested fixes

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4623

### DIFF
--- a/src/Cameras/VR/babylon.vrExperienceHelper.ts
+++ b/src/Cameras/VR/babylon.vrExperienceHelper.ts
@@ -921,7 +921,7 @@ module BABYLON {
                 }
 
                 this.raySelectionPredicate = (mesh) => {
-                    return mesh.isVisible;
+                    return mesh.isVisible && mesh.isPickable;
                 }
 
                 this.meshSelectionPredicate = (mesh) => {
--- a/src/Gizmos/babylon.boundingBoxGizmo.ts
+++ b/src/Gizmos/babylon.boundingBoxGizmo.ts
@@ -382,7 +382,7 @@ module BABYLON {
          * @param mesh the mesh to wrap in the bounding box mesh and make not pickable
          * @returns the bounding box mesh with the passed in mesh as a child
          */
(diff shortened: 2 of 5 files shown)


src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -921,7 +921,7 @@ module BABYLON {
                 }
 
                 this.raySelectionPredicate = (mesh) => {
-                    return mesh.isVisible;
+                    return mesh.isVisible && mesh.isPickable;
```
TrevorDev: @davrous is there a reason this was not in this predicate by default? Would this be okay to add now?

deltakosh: Agree this should be like that (with the isPickable)

src/Gizmos/babylon.boundingBoxGizmo.ts
```diff
@@ -329,5 +376,48 @@ module BABYLON {
...
+            }
+            var oldPos = mesh.position.clone()
+            var oldRot = mesh.rotationQuaternion.clone();
+            mesh.rotationQuaternion.set(0,0,0,1);
+            mesh.position.set(0,0,0)
+
+            // Update bounding dimensions/positions   
+            var box = BABYLON.MeshBuilder.CreateBox("box", {size: 1}, mesh.getScene());
+            var boundingMinMax = mesh.getHierarchyBoundingVectors();
+            box.scaling.x = boundingMinMax.max.x-boundingMinMax.min.x;
```
RaananW: out of curiosity - why not use the subtract (to reference) function?

src/Cameras/VR/babylon.vrExperienceHelper.ts
davrous: Indeed, this is better like that. Thanks!

src/Gizmos/babylon.boundingBoxGizmo.ts
TrevorDev: I think when I was writing it, I had more complex logic here, I just updated this to use subtract here.

PR #4658, opened 2018-07-03
Fixed issue where environment was not scaling correctly in VR.

atulyar:
In the VR configuration for Babylon Viewer, the user has the option to scale the object. However, the environment was not being scaled correctly with the object.


RaananW:
Thanks a lot :)

Issue #4364, opened 2018-05-23
Extract DDS, KTX or texture creation as plugin


deltakosh:

sebavan:
- [ ] Constant to readonly
- [ ] No more gl in dds tools
- [ ] No more gl in telemetry manager
- [ ] Plugin System for texture loaders
- [ ] Create plugin for env dds ktx tga.... and standard (jpg bmp...)

sebavan:
Internal vs Base texture fields are all none duplicated.

PR #4659, opened 2018-07-03
Telemetry manager no _gl

sebavan:


sebavan:
Addresses the gl access from telemetry manager part of https://github.com/BabylonJS/Babylon.js/issues/4364

PR #4661, opened 2018-07-03
1.) Allow loadFile flow to make error callback in case of file not found.

loopingLines:
1.)  Allows loadFile flow to make error callback in case of file not found, during saveFileIntoDB.

2.) Does not Save server error response to db.

### DIFF
--- a/src/Tools/babylon.database.ts
+++ b/src/Tools/babylon.database.ts
@@ -494,7 +494,7 @@ module BABYLON {
 
             var saveAndLoadFile = () => {
                 // the scene is not yet in the DB, let's try to save it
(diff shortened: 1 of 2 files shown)


loopingLines:
Created an [issue](https://github.com/BabylonJS/Babylon.js/issues/4662) related to this. Not sure if [_saveFileIntoDBAsync call ](https://github.com/BabylonJS/Babylon.js/pull/4661/files#diff-62a5fa699ed9604f122e768c83502b86L497) @497 can pass the argument useArrayBuffer.

deltakosh:
Looks good to me
Thanks a lot!

PR #4675, opened 2018-07-04
Don't generate mipmaps for HDRCubeTexture if texture float filtering is disabled

NicolasBuecher:
If OES_texture_float_linear extension is not supported, enabling mipmaps leads to:

hdrcubetexturefix

From [here at 3.8.10.5](https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf), we can read: 

> If the levelbase array was not specified with an unsized internal format from table
3.3 or a sized internal format that is both color-renderable and texture-filterable
according to table 3.13, an INVALID_OPERATION error is generated.

As far as I understand, because RGBA32F is not texture-filterable in this specific case, mipmaps generation fails.

I suggest to not generate mipmaps if OES_texture_float_linear is not supported thus.

I don't think I have to update What's new for this, do I ? (I'll do it just in case...)


deltakosh:
Looks good to me
Pinging @sebavan for a second thought

sebavan:
looks good to me

PR #4676, opened 2018-07-04
Update Comments for Scene Helpers

BabylonJSGuide:
Corrected links and have added a few more details


RaananW:
hey,

thanks!
You commited a lot of (unneeded?) d.ts files, was that on purpose?

BabylonJSGuide:
No. On the last pull a list of d.ts files appeared in the commit list . As they came from the Babylon site I thought I had better commit them.. I did get a lot of new ones when I valdated my changes but discarded those. I see that the PR failed a couple of checks. What do I need to do?

RaananW:
The PR failed because you made no changes to the what's-new file, but this is fine. 
You will need to revert those d.ts files as they are not needed. You can delete them and commit the delete, this should be fine.
Just make sure you add only the files you made changes to

PR #4684, opened 2018-07-05
Grid Mat Shadow support

Pryme8:
I have not way of testing this, so I'm not sure if this is even a valid pull.

If you see any problems or anything slightly off please reject this right away,  this was just a quick thought from:
http://www.html5gamedevs.com/topic/38623-grid-material-lacks-receiveshadows/

I would bet that its wrong, but without testing I would not know.


deltakosh:
We really need to test before everything :)

Pryme8:
I need to get back to my house to do a build to test it, and wont have that option for a few days.

So I think it may be best to just reject this for now.

deltakosh:
As you wish, I'm fine keeping it open if you want

Issue #4141, opened 2018-04-16
Add gizmos to core


deltakosh:
Add APIs to
- [x] Display gizmos for scale, rotate, translate (one or all)
- [x] Control mesh with drag and drop using the gizmos
- [x] Multi axis operations (Plane dragging for main planes, drag along viewing axis, uniform scale)
- [x] Support multiple input methods (Mouse + 6Dof controller)
- [x] six dof move behavior
- [x] multi controller scale behaviors
- [x] Axis local and world mode (In progress)
- [x] VR controllers occluded properly by gizmo layer
- [x] Boudning box improved vr interactions match closer to the cliff house
- [x] Upgrade gizmo appearance (And on selected behavior)
- [x] Snap mode: Configurable snapping with snap callback (distance/angle)
- [x] Gizmo manager modes (eg. rotation, scale, translate separate or all together. Gumball or Bounding box modes) 
- [x] Gizmo hotkeys (To enable/disable gizmos)
- [x] Gizmo skinning/sizing controls for customization

TrevorDev:
https://github.com/BabylonJS/Babylon.js/pull/4285

dad72:
I have a question/suggestion. Will the Gizmo have a : 

- Snap mode. 
- Axis local and world mode.
- An undo/redo all the manipulations.

 it would be interesting to have his possibilities too.

Thanks.

deltakosh:
Snap: yes
Local/world: yes
Undo:redo: not in v1

Eole211:
Wouldn't it be nice if we could easily override the mesh transform applied when interacting with Gizmos ?
For example, with the axisDragGizmo I'd like to be able to change easily the function called on drag. 
I could remove the current function called here ` this._dragBehavior.onDragObservable.add((event)=>{...})`, and replace it with a custom function but it looks rather dirty to me. Also the _dragBehavior parameter is private and so I couldn't do that in typescript.

Eole211:
Yeah, it would be nice to have the dragBehavior public, and maybe also the sub-gizmos ? Like the _xDrag, _yDrag and _zDrag for the Position Gizmo. It would make make easier for developers to tweak gizmo behaviors a bit without extending gizmo classes.

TrevorDev:
👍 sounds useful to me, I'll make these public.

PR #4689, opened 2018-07-05
Update gizmo manager

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4141

### DIFF
--- a/src/Gizmos/babylon.gizmo.ts
+++ b/src/Gizmos/babylon.gizmo.ts
@@ -74,7 +74,11 @@ module BABYLON {
                         this._rootMesh.position.copyFrom(this.attachedMesh.absolutePosition);
                     }
                     if(this._updateScale && this.gizmoLayer.utilityLayerScene.activeCamera && this.attachedMesh){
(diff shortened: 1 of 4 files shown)


src/Gizmos/babylon.gizmo.ts
```diff
@@ -74,7 +74,11 @@ module BABYLON {
                         this._rootMesh.position.copyFrom(this.attachedMesh.absolutePosition);
                     }
                     if(this._updateScale && this.gizmoLayer.utilityLayerScene.activeCamera && this.attachedMesh){
-                        this._rootMesh.position.subtractToRef(this.gizmoLayer.utilityLayerScene.activeCamera.position, tempVector);
+                        var cameraPosition = this.gizmoLayer.utilityLayerScene.activeCamera.position;
+                        if((<WebVRFreeCamera>this.gizmoLayer.utilityLayerScene.activeCamera).devicePosition){
```
deltakosh: are you sure it will always be a webvrcamera?

TrevorDev: If it's not a webVR camera this if won't enter

PR #4690, opened 2018-07-05
Adding TGA tests + prevent unbind in upload

sebavan:


src/Engine/babylon.engine.ts
```diff
@@ -5252,6 +5252,8 @@
             this._bindTextureDirectly(bindTarget, texture, true);
 
             this._uploadDataToTextureDirectly(texture, texture.width, texture.height, imageData, faceIndex, lod);
+
+            this._bindTextureDirectly(bindTarget, null, true);
```
deltakosh: This will have the same issue as this morning with the instance demo
Code assume that texture is still bind after loading it

sebavan: Nope as I changed the upload method accordingly to uploadDirectly (no bind/unbind) like in DDS:
https://github.com/BabylonJS/Babylon.js/pull/4690/files#diff-bad50cfda6c5c1f8fe4bba23866c0667L199

This was a typo of the previous PR :-(

I added the according test for TGA so that I am not breaking it again... I hope ;-)

deltakosh: Can you just make sure and run the instance démo as well?

sebavan: Yup Triple Checked :-)

image

PR #4696, opened 2018-07-06
Add support Atlas Texture for GUI

quocsinh:
About Atlas Texture: https://en.wikipedia.org/wiki/Texture_atlas
This is the first time i use Typescript :D
I have been trying to learn about Typescript, let me know if I have violated code rules/coding convention.
I will be soon update document.
Thanks!


dist/preview release/what's new.md
```diff
@@ -25,6 +25,7 @@
   - Added support for life time gradients. [Doc](https://doc.babylonjs.com/babylon101/particles#lifetime)
   - Added support for angular speed gradients. [Doc](https://doc.babylonjs.com/babylon101/particles#rotation)
 - Added SceneComponent to help decoupling Scene from its components ([sebavan](http://www.github.com/sebavan))
+- Added support AtlasTexture for GUI Image.
```
deltakosh: please also add your GitHub alias to get full credits :)

src/Tools/babylon.assetsManager.ts
```diff
@@ -669,6 +669,118 @@ module BABYLON {
...
+                let name = key.substring(0, key.lastIndexOf('.'));
+                this.data[name] = new AtlasTexture(name, img, this._framesData[key]);
+            });
+        }
+    }
+
+    /**
+     * Defines the interface used storage information atlas texture
+     */
+    export interface IAtlasTexture {
```
deltakosh: As this is specific to GUI it should be in GUI

src/Tools/babylon.assetsManager.ts
```diff
@@ -669,6 +669,118 @@ module BABYLON {
         }
     }
 
+    /**
+     * Define a task used by {BABYLON.AssetsManager} to load atlas texture
+     */
+    export class AtlasTextureAssetTask extends AbstractAssetTask {
```
deltakosh: We should rename it to JSONAssetTask so it could be used for several other formats

src/Tools/babylon.assetsManager.ts
```diff
@@ -669,6 +669,118 @@ module BABYLON {
...
+        rotated: boolean,
+        sourceSize: any,
+        spriteSourceSize: any,
+        trimmed: boolean,
+    }
+
+    /**
+     * The class used for source's GUI.Image
+     */
+    export class AtlasTexture {
```
deltakosh: This is not a texture but more a Rect with an imageElement? (Not a texture as in Babylon.Texture I mean)

deltakosh: As it is designed to work with GUI.Image we could name it GUI.ImageAtlas ?

dist/preview release/what's new.md
quocsinh: I forgot this :)

src/Tools/babylon.assetsManager.ts
quocsinh: OK!

src/Tools/babylon.assetsManager.ts
quocsinh: I will rename it.

src/Tools/babylon.assetsManager.ts
quocsinh: Yes, this is not a texture. Your means is create new class ImageAtlas extend Image?

deltakosh: Yep just anew ImageAtlas only

quocsinh: After thinking again, i see no necessary to create a new class ImageAtlas.
1) Use the class Image, it will be flexible when change src. Example: user get avatar (string) from server, for some reason the path does not exist, developer are easy fallback to load avatar default from AtlasTexture.
2) Actually i have created and used AtlasTexture in my project 3 months ago. So far, it still working at well, no problem.

deltakosh: Well please update the PR to see how it looks like:)

quocsinh:
@deltakosh  I finished task, please review! :)

src/Tools/babylon.assetsManager.ts
```diff
@@ -669,6 +669,71 @@ module BABYLON {
         }
     }
 
+    /**
+     * Define a task used by {BABYLON.AssetsManager} to load atlas texture
+     */
```
deltakosh: Actually I thought about this one and we could just use TextFileAssetTask directly
This one is too much specific and only useful for image atlas

dist/preview release/what's new.md
```diff
@@ -25,6 +25,7 @@
   - Added support for life time gradients. [Doc](https://doc.babylonjs.com/babylon101/particles#lifetime)
   - Added support for angular speed gradients. [Doc](https://doc.babylonjs.com/babylon101/particles#rotation)
 - Added SceneComponent to help decoupling Scene from its components ([sebavan](http://www.github.com/sebavan))
```
deltakosh: Added support for AtlasTexture for GUI Image

gui/src/2D/atlasSourceImage.ts
```diff
@@ -0,0 +1,37 @@
+module BABYLON.GUI {
+    /**
+     * Defines the interface used storage information atlas texture
```
deltakosh: Used for storing atlas information

gui/src/2D/atlasSourceImage.ts
```diff
@@ -0,0 +1,37 @@
+module BABYLON.GUI {
+    /**
+     * Defines the interface used storage information atlas texture
+     */
+    export interface IAtlasSourceData {
+        frame: any,
```
deltakosh: Every field has to be commented

gui/src/2D/atlasSourceImage.ts
```diff
@@ -0,0 +1,37 @@
...
+    export interface IAtlasSourceData {
+        frame: any,
+        pivot: any,
+        rotated: boolean,
+        sourceSize: any,
+        spriteSourceSize: any,
+        trimmed: boolean,
+    }
+
+    /**
```
deltakosh: Class used to store an element of an Atlas texture
May be renamed it to AtlasElementSourceImage

gui/src/2D/atlasSourceImageManager.ts
```diff
@@ -0,0 +1,34 @@
...
+        private _data: any = {};
+        private _sourceImage: any = {};
+
+        constructor() {
+        }
+
+        /**
+         * Parse data from task loaded
+         * @param {BABYLON.JSONAssetTask} task
+         */
```
deltakosh: This is where we should use the TextFileAssetTask directly
Also as an user I would expect to get a way to load it directly from an url

gui/src/2D/controls/image.ts
```diff
@@ -24,6 +23,9 @@ module BABYLON.GUI {
         private _cellHeight: number = 0;
         private _cellId: number = -1;
```
deltakosh: Good idea

gui/src/2D/controls/image.ts
```diff
@@ -162,69 +164,97 @@ module BABYLON.GUI {
...
          */
-        public set source(value: Nullable<string>) {
-            if (this._source === value) {
-                return;
-            }
+        public set source(value: Nullable<string | AtlasSourceImage>) {
+            if (value) {
+                let src = undefined;
+                if (typeof value === 'string') {
+                    this._sourceLeft = 0;
```
deltakosh: You should not remove the current value if source is an url. As an user I could have set them up before setting the source

gui/src/2D/controls/image.ts
```diff
@@ -162,69 +164,97 @@ module BABYLON.GUI {
...
+                    this._sourceTop = value.y;
+                    this._sourceWidth = value.w;
+                    this._sourceHeight = value.h;
+                    this.width = this._sourceWidth + 'px';
+                    this.height = this._sourceHeight + 'px';
+                    src = value.source.src;
+                }
 
-            this._loaded = false;
-            this._source = value;
```
deltakosh: This test has to be done before everything

gui/src/2D/controls/image.ts
```diff
@@ -162,69 +164,97 @@ module BABYLON.GUI {
...
+                }
 
-            this._domImage = new DOMImage();
+                this._loaded = false;
+                this._source = value;
 
-            this._domImage.onload = () => {
-                this._onImageLoaded();
-            }
-            if (value) {
```
deltakosh: The if (value) must be kept

quocsinh:
I hope this time will be okay :)

src/Tools/babylon.assetsManager.ts
```diff
@@ -318,12 +372,47 @@ module BABYLON {
...
+                    readJSONData(this.text);
+                else
+                    onSuccess();
             }, undefined, false, false, (request, exception) => {
                 if (request) {
                     onError(request.status + " " + request.statusText, exception);
                 }
             });
+
+            let readJSONData = (data: string) => {
```
deltakosh: My point was that we should have no link st all to atlas data in this file. The load should be done by an atlas loader class here.
We cannot have a dependency to an external class (the atlas element)

We should have all code related to atlas in the Gui project:)

All this code should be in Gui and should rely on the TextFileAssetTask

Something like theanager you had before (plus a way to load the data without using the task ad not everyone wants to use the assets manager)

Something like (inside GUI):
- AtlasSourceDataHelper.Load(url)
- AtlasSourceDataHelper.PrepareAssetsManager

The second one will create a TextFileAssetTask and will use it to get the data. Then it will use the same code to get everything ready and returns a list of IAtlasSourceData

Does it make sense?

src/Tools/babylon.assetsManager.ts
```diff
@@ -1,4 +1,49 @@
 module BABYLON {
+    /**
+     * Defines the interface used for storing atlas information
+     */
+    export interface IAtlasSourceData {
+        /**
+         * Coordinate of the source image
+         */
+        frame: any,
```
deltakosh: Instead of any we should say {x: number, etc...}

src/Tools/babylon.assetsManager.ts
quocsinh: Thanks for your review, i'm so stupid :( .
I will think carefully, slowly and try again.

deltakosh: Man! Don't blame yourself
You are doing great. Adding code to a framework takes time and this is a perfectly natural process.
You are about to have a wonderful PR
Take you time, there is no rush

deltakosh:
closing for now. Please create a new PR when you have a chance (no rush)

PR #4701, opened 2018-07-06
TextBlock: provide expected height

adrientetar:
r? @deltakosh

### DIFF
--- a/gui/src/2D/controls/textBlock.ts
+++ b/gui/src/2D/controls/textBlock.ts
@@ -249,29 +249,35 @@ module BABYLON.GUI {
         }
 
         protected _additionalProcessing(parentMeasure: Measure, context: CanvasRenderingContext2D): void {
-            this._lines = [];
+            this._lines = this._breakLines(this._currentMeasure.width, context);
+            this.onLinesReadyObservable.notifyObservers(this);
+        }
+
+        protected _breakLines(refWidth: number, context: CanvasRenderingContext2D): object[] {
+            var lines = [];
             var _lines = this.text.split("\n");
 
             if (this._textWrapping && !this._resizeToFit) {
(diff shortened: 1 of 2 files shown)


gui/src/2D/controls/textBlock.ts
```diff
@@ -249,37 +269,62 @@ module BABYLON.GUI {
...
-            this.onLinesReadyObservable.notifyObservers(this);
+            return lines;
         }
 
         protected _parseLine(line: string = '', context: CanvasRenderingContext2D): object {
             return { text: line, width: context.measureText(line).width };
         }
 
-        protected _parseLineWithTextWrapping(line: string = '', context: CanvasRenderingContext2D): object {
+        protected _parseLineEllipsize(line: string = '', width: number,
```
deltakosh: function name seems weird :)

adrientetar: What do you suggest?

BTW not sure why CI fails, test suite passes for me locally (running on Firefox) and I did update the what's new file afaict...

deltakosh: _parseLineEllipsis should work no?

deltakosh: For the CI error: your are not updating the right what's new :)
You want to update the one in dist/preview release

adrientetar: So the dist folder isn't all generated files? The what's new at the root of the repo is updated from the one in dist? [Seems things would be tidier w/o release files inside the repo tbh]

deltakosh: Correct, not all files :)
The one at the root is for the current stable version.

Release files are in the repo to let people grab them directly

PR #4706, opened 2018-07-08
Typescript version  0 test version for playground

NasimiAsl:


deltakosh:
It is a reallllly good start :)
Some remarks:
* The ts.js should not duplicate everything because it will be a nightmare to maintain. We should have a central index.js and a ts.js that contains only the changes needed for the TS to work
* Not a big fan of the base scene.txt. I would not expect to create my own scene, etc... Perhaps we should just provide a simpler example like the one on the doc.babylonjs.com

NasimiAsl:
ok help me to finish that 
what you suggest for ts.js?
and i like any simple scene

deltakosh:
Unfortunately I don't have that much time:(
Ideally just remove the code that is already in index.js and just keep the new function in TS.js

deltakosh:
Let me give a try I have a free hour :D

deltakosh:
I pushed it directly! Thanks a lot my friend

PR #4714, opened 2018-07-09
glTFSerializer: Fix for handling negative scale

kcoley:
- Fixes an issue where the winding order breaks when reimporting a mesh exported from the glTF serializer


kcoley:
@deltakosh spoke with @TrevorDev and we may have a better way to fix this.  Testing it now locally.

kcoley:
@deltakosh ended up going with my original fix.

PR #4718, opened 2018-07-09
Add an InputPassword control

theom:
This is the GUI password input control as [discussed in the forum](http://www.html5gamedevs.com/topic/38743-inputtext-control-types/), done using inheritance.

I added a hook into the InputText render call in order to be able to hide the output by overriding the hook in the password control. This was the simplest implementation I could come up with which preserved changes to the input password.

If this gets accepted I will then update the documentation.


deltakosh:
This is perfect.
Just two things:
* please update the what's new in dist/preview folder to credit yourself
* Please update this doc page: https://github.com/BabylonJS/Documentation/blob/master/content/How_To/GUI/Gui.md

Thanks a lot!

gui/src/2D/controls/inputPassword.ts
```diff
@@ -0,0 +1,16 @@
...
+module BABYLON.GUI {
+    /**
+     * Class used to create a password control
+     */
+    export class InputPassword extends InputText {
+        protected _beforeRenderText(text: string): string {
+            let txt = "";
+            for (let i = 0; i < text.length; i++) {
+                txt += "\u2022";
+            }
```
adrientetar: nit: you could do "\u2022".repeat(text.length)

theom: That's what I did at first but the build complained about the repeat method not being available (because of an ES5 target I presume). Hence the loop.

deltakosh:
They are conflicts :(

PR #4725, opened 2018-07-10
make gizmo managers gizmos public

TrevorDev:
http://www.html5gamedevs.com/topic/38683-gizmos-and-mesh-dragging-behaviors/?tab=comments#comment-221102


deltakosh:
This will need a Doc update:)

PR #4747, opened 2018-07-13
serializer submaterials of multimaterial

kcoley:
- Serialize sub materials of multimaterial

### DIFF
--- a/src/Tools/babylon.sceneSerializer.ts
+++ b/src/Tools/babylon.sceneSerializer.ts
@@ -71,7 +71,7 @@
                     if (!serializationObject.multiMaterials.some((mat: Material) => (mat.id === (<Material>mesh.material).id))) {
                         serializationObject.multiMaterials.push(mesh.material.serialize());
                         for (let submaterial of mesh.material.subMaterials) {
-                            if (submaterial instanceof Material) {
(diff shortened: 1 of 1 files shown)


src/Tools/babylon.sceneSerializer.ts
```diff
@@ -67,10 +67,17 @@
             if (mesh.material) {
                 if (mesh.material instanceof MultiMaterial) {
                     serializationObject.multiMaterials = serializationObject.multiMaterials || [];
+                    serializationObject.materials = serializationObject.materials || [];
                     if (!serializationObject.multiMaterials.some((mat: Material) => (mat.id === (<Material>mesh.material).id))) {
                         serializationObject.multiMaterials.push(mesh.material.serialize());
+                        for (let submaterial of mesh.material.subMaterials) {
+                            if (submaterial instanceof Material) {
```
deltakosh: Why this test?

kcoley: This was to check for nullable, but I can change it to see if it is not null

kcoley: Just made the change

deltakosh:
Great ! thanks!

Issue #4439, opened 2018-06-05
Environment Texture Follow Up


sebavan:
- [ ] Consistent Default Lod Scale matching the LYS generation tool
- [ ] Library of default environments
- [ ] Integrate in Viewer
- [ ] Review Specular AA on 512 texture generated by Lys

PR #4750, opened 2018-07-13
Viewer using the new .env format

RaananW:
https://github.com/BabylonJS/Babylon.js/issues/4439


Viewer/src/configuration/configuration.ts
```diff
@@ -94,8 +97,10 @@ export interface ViewerConfiguration {
             specular?: { r: number, g: number, b: number };
         }
         hideLoadingDelay?: number;
+        /** Deprecated */
```
sebavan: Should we keep them if not use ?

Viewer/src/configuration/interfaces/environmentMapConfiguration.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    /**
+     * Default rotation to apply to the environment map.
+     */
+    rotationY: number;
+
+    /**
+     * Tint level of the main color on the environment map.
+     */
+    tintLevel: number;
```
sebavan: There is a tint level but is there any tint color ?

Viewer/src/managers/sceneManager.ts
```diff
@@ -847,6 +842,18 @@ export class SceneManager {
         });
     }
 
+    protected _configureEnvironmentMap(environmentMapConfiguration: IEnvironmentMapConfiguration): any {
+        let rotatquatRotationionY = Quaternion.RotationAxis(Axis.Y, environmentMapConfiguration.rotationY || 0);
```
sebavan: Should this be in the below check ?

Viewer/src/configuration/configuration.ts
RaananW: yes, just for d.ts reasons (I dont want to break public projects'build process)

Viewer/src/configuration/interfaces/environmentMapConfiguration.ts
RaananW: There is main color, I should move it here instead of in scene. Will do that

Viewer/src/managers/sceneManager.ts
RaananW: yep, makes sense

sebavan:
all good to me feel free to merge

PR #4773, opened 2018-07-18
Fix gulp default command (memory consumption)

RaananW:
Our gulp default task had a meltdown on windows systems due to a problem with memory consumption.

Constant streams are known to consume a lot of memory, and we have a lot of them running in sequence.

As a solution i reduced the memory used by karma (by removing unneeded files). As another solution we can increase the memory provided to node for the task, but this will require creating an npm task for gulp.


RaananW:
Please test it before merging.
Another recommendation that came from teams with the same problem is to move to gulp 4 and not use run-sequence, which has a bit of a problem with memory management.

Issue #4793, opened 2018-07-19
Dynamic textures not getting serialized


kcoley:
Should resolve the issue flagged here: https://github.com/BabylonJS/Babylon.js/issues/4300

Sample playground: http://localhost:1338/Playground/index-local.html#TK05AV

PR #4794, opened 2018-07-20
Add serialization support for dynamic textures

kcoley:
Fixes https://github.com/BabylonJS/Babylon.js/issues/4793

### DIFF
--- a/src/Materials/Textures/babylon.dynamicTexture.ts
+++ b/src/Materials/Textures/babylon.dynamicTexture.ts
@@ -179,10 +179,15 @@
         }
 
         /**
-         * Serializes the dynamic texture
+         * Serializes the dynamic texture.  The scene should be ready before the dynamic texture is serialized
          * @returns a serialized dynamic texture object
          */
         public serialize(): any {
+            const scene = this.getScene();
+            if (scene && !scene.isReady()) {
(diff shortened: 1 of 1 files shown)


src/Materials/Textures/babylon.dynamicTexture.ts
```diff
@@ -178,6 +178,20 @@
             return newTexture;
         }
 
+        /**
+         * Serializes the dynamic texture
+         * @returns a serialized dynamic texture object
+         */
+        public serialize(): any {
+            const serializationObject = super.serialize();
```
deltakosh: Does it work on IE?

kcoley: No, this does not work in IE.  I'll work on a fix.

deltakosh: A fix or if none found a check to make sure it won't crash

kcoley:
@deltakosh `toDataURL` works on IE11, but my playground was using a promise, which didn't work.  I changed it to use observable instead to test on IE11: http://playground.babylonjs.com/#FU0ES5

I also added a warning indicating that the scene should be ready before serializing the dynamic texture.  I can add this warning to the Dynamic Texture documentation as well.

deltakosh:
Good!

Issue #4589, opened 2018-06-21
Support for previewing compressed textures in Inspector


justinctlam:
# Feature request

If you load a model with a compressed texture, for example a DXT1 texture, I want to be able to see the texture in the inspector just like the jpg or png textures. This is the texture previewer under the texture tab in the inspector.

Bloadrick:
Hi,

I am actually working on your request, can you give me more information ? 
For me DXT1 is a compression algorithm and it can give you a texture in png.

Can you share your texture or tell me what format of image you want to preview ?

justinctlam:
What specific information do you need?

You can load any DXT compression image as an example. Eg container can be dds.

PR #4798, opened 2018-07-20
Add preview for textures in the inspector

Bloadrick:
I have added a preview for different format of textures in the inspector.

The DDS format is now render in a BabylonJS Canvas so that you can see it fully if it is a cube texture.

The feature request : https://github.com/BabylonJS/Babylon.js/issues/4589


inspector/src/tabs/TextureTab.ts
```diff
@@ -196,4 +227,101 @@ export class TextureTab extends Tab {
         }
         item.active(true);
     }
+
+}
+
+class TestDDSPreview {
```
deltakosh: Not fan of having test class here

Bloadrick: Yes you are right and right, i'm fixing that asap.

I was not aware of the webGL context limit, I need to check that.

Jaskar:
I'm taking care of that this week!

deltakosh:
Still ok?

Jaskar:
Sorry, I was on holidays.
I'm looking at this before Friday!

PR #4806, opened 2018-07-21
glTFSerializer: Display warning about pivot points not being supported

kcoley:
- Add support for pivot point offsets

### DIFF
--- a/serializers/src/glTF/2.0/babylon.glTFExporter.ts
+++ b/serializers/src/glTF/2.0/babylon.glTFExporter.ts
@@ -783,6 +783,9 @@ module BABYLON.GLTF2 {
          * @param babylonTransformNode Babylon mesh used as the source for the transformation data
          */
         private setNodeTransformation(node: INode, babylonTransformNode: TransformNode): void {
+            if (!babylonTransformNode.getPivotPoint().equalsToFloats(0,0,0)) {
+                BABYLON.Tools.Warn("Pivot points are not supported in the glTF serializer");
+            }
             if (!babylonTransformNode.position.equalsToFloats(0, 0, 0)) {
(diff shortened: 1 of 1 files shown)


serializers/src/glTF/2.0/babylon.glTFExporter.ts
```diff
@@ -783,8 +783,12 @@ module BABYLON.GLTF2 {
          * @param babylonTransformNode Babylon mesh used as the source for the transformation data
          */
         private setNodeTransformation(node: INode, babylonTransformNode: TransformNode): void {
-            if (!babylonTransformNode.position.equalsToFloats(0, 0, 0)) {
-                node.translation = this._convertToRightHandedSystem ? _GLTFUtilities._GetRightHandedPositionVector3(babylonTransformNode.position).asArray() : babylonTransformNode.position.asArray();
+            const pivotPoint = babylonTransformNode.getPivotPoint();
+            if (!babylonTransformNode.position.equalsToFloats(0, 0, 0) && !pivotPoint.equalsToFloats(0,0,0)) {
```
deltakosh: the pivot point is not exactly a position.
It is more a pre translation (before rotation)
I don't think we can export it as translation

@bghgary is there a notion of Pivot in gltf?

bghgary: Not that I'm aware of.

deltakosh: Ok so we need to chat to see how to correctly support pivot point

kcoley:
@deltakosh @bghgary This update creates a parent node with the pivot translation

deltakosh:
Does it work well with @davrous sample?

kcoley:
Hmm seems to be a bug when I export his scene.  Looking into it.

kcoley:
Changed this PR to display a warning about pivot points not being supported

PR #4824, opened 2018-07-25
Moving to gulp 4, updating dependencies

RaananW:
~gulp run is not yet working, please don't merge. Just~ testing travis


RaananW:
I will take care of the node error (some package tries to load WebAssembly)

RaananW:
Please don't forget running `npm install` before your next gulp build, otherwise... well, it won't work.

PR #4839, opened 2018-07-26
Add support for WebGL2 texture formats

NicolasBuecher:
Update WebGL constants to enable support for WebGL2 texture formats.
[Discussed here.](http://www.html5gamedevs.com/topic/38584-support-webgl2-new-texture-types-formats-and-sized-internal-formats/?tab=comments#comment-220446)


src/Engine/babylon.engine.ts
```diff
@@ -359,17 +359,54 @@
...
-        /** R */
+        /** RED */
+        public static readonly TEXTUREFORMAT_RED = 6;
         public static readonly TEXTUREFORMAT_R = 6;
         /** RG */
         public static readonly TEXTUREFORMAT_RG = 7;
-
-        /** UNSIGNED_INT */
+        /** RED_INTEGER */
+        public static readonly TEXTUREFORMAT_RED_INTEGER = 8;
```
sebavan: Why do we need to add the integer here and below if it is already included in the TEXTURETYPE ?

src/Engine/babylon.engine.ts
```diff
@@ -7042,18 +7121,35 @@
...
-                    internalFormat = this._gl.RED;
-                    break;       
-                case Engine.TEXTUREFORMAT_RG:
-                    internalFormat = this._gl.RG;
-                    break;                                    
+            }
+
+            if (this._webGLVersion > 1) {
+                switch (format) {
+                    case Engine.TEXTUREFORMAT_RED_INTEGER:
```
sebavan: Is this not possible to deduce from TEXTUREFORMAT_RED and type TEXTURETYPE_INT and same below ?

NicolasBuecher: For this example, it is possible. But for some others like TEXTURETYPE_BYTE and TEXTUREFORMAT_RED. You need here to specify TEXTUREFORMAT_R_INTEGER.

NicolasBuecher:
As discussed [here](http://www.html5gamedevs.com/topic/38584-support-webgl2-new-texture-types-formats-and-sized-internal-formats/?do=findComment&comment=220258), DK said "I do not want to expose internal format." so I didn't add an extra internalSized

NicolasBuecher:
Sorry but I won't be able to do any updates before Tuesday !
No matter if you want to make some changes during this time.

sebavan:
Agree but it feels strange to me to see the format and type mixed together. This is only the integer by the way, all the rest looks good :-)

sebavan:
No problem, no rush on this, this is a major change and it is better to look at it twice :-)

sebavan:
@deltakosh, this directly maps with Webgl2 (reference: https://webgl2fundamentals.org/webgl/lessons/webgl-data-textures.html) so I wonder how we could avoid the confusion with the _INTEGER format which is truly a fromat in webgl and not mapping to a type.

Maybe we should leave it like this as it is a strong webgl match ?

deltakosh:
Well yes I think this PR is probably the best we can get providing we want to support back compat

sebavan:
So all good for me once the documentation checked issue has been fixed:
image

deltakosh:
Try to run npm install gulp@4.0.0

deltakosh:
I had to revert the PR as a lot of visual tests where failing (you can run them directly from vscode in the debug tasks (run visual tests))

NicolasBuecher:
Thanks for gulp !!
I'll take a look at that.

deltakosh:
I updated the readme :)

deltakosh:
Please start a new one;)

PR #4840, opened 2018-07-26
projectUVfromMesh

Pryme8:
trying to support a simple plane UV projection method.

### DIFF
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -1061,6 +1061,56 @@
             this._geometry.toLeftHanded();
             return this;
         }
+        
+        /**
+         * Update the current Mesh UV coords by sampling rays fired from its vertices to a sampling mesh.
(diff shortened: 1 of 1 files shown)


src/Mesh/babylon.mesh.ts
```diff
@@ -1061,6 +1061,56 @@
             this._geometry.toLeftHanded();
             return this;
         }
+        
+        /**
```
deltakosh: Hello not sure this is for the framework. I would more see it in the snippets in the Doc (gamelets or utilities)

Pryme8: Aw, did I at least do my first TS contribution correctly?

Not to pressed if it gets in or not, just wanted to see if I am matching the conventions correctly or not.

Where are the gamelets or utilities, so I can look up a reference for presenting this info.

Pryme8:
Thank you, I have been dreading learning TS but when in Rome...

deltakosh:
Appreciated!

PR #4841, opened 2018-07-26
Add ACES tone mapping and integrate image processing in particle system

sebavan:
In order to get a closer look to to other engines we integrated ACES tone mapping in Babylon.

Also we integrated image processing as part of the particles rendering (both CPU and GPU) in order to allow a one pass only bringing consistency with the rest of the scene.

### DIFF
--- a/src/Particles/babylon.particleSystem.ts
+++ b/src/Particles/babylon.particleSystem.ts
@@ -56,7 +56,6 @@
         private _particles = new Array<Particle>();
         private _epsilon: number;
         private _capacity: number;
-        private _scene: Scene;
         private _stockParticles = new Array<Particle>();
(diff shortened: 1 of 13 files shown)


src/Particles/babylon.particleSystem.ts
```diff
@@ -109,7 +109,48 @@
          */
         public getClassName(): string {
             return "ParticleSystem";
-        }        
+        }
+
+        private _imageProcessingConfigurationDefines = new ImageProcessingConfigurationDefines();
```
deltakosh: Why not having it on the base class?

sebavan: Because I did not notice there was a base class :-) I am changing it now.

Issue #4367, opened 2018-05-23
Add public contract for loading custom glTF extensions


bghgary:
See http://www.html5gamedevs.com/topic/37814-gltf-import-with-custom-extensions/?do=findComment&comment=215812

PR #4845, opened 2018-07-27
Add support for registering custom glTF loader extensions

bghgary:
See https://github.com/BabylonJS/Babylon.js/issues/4367

Testing and documentation still to come. This is just a first pass. Not everything is exposed yet.


bghgary:
Here is a playground test that replaces all the materials with StandardMaterials:
https://playground.babylonjs.com/#20XT9A

Issue #4844, opened 2018-07-27
Create equal scaling triangular gizmo


TrevorDev:
http://www.html5gamedevs.com/topic/37860-sample-code-for-features-in-babyloneditor/?page=2&tab=comments#comment-222917

- [x] Provide helper field for snapping
- [x] Triangular scale gizmo
- [x] Heavily nested bounding box perf avoidance
- [x] Separate observables in bounding box
- [x] Default pipeline removing cameras
- [x] Document:
      /**
         * Fired when a scale box is dragged
         */
        public onScaleBoxDragObservable = new Observable<{}>();
        /**
          * Fired when a scale box drag is ended
         */
        public onScaleBoxDragEndObservable = new Observable<{}>();
        /**
         * Fired when a rotation sphere is dragged
         */
        public onRotationSphereDragObservable = new Observable<{}>();
        /**
         * Fired when a rotation sphere drag is ended
         */
        public onRotationSphereDragEndObservable = new Observable<{}>();
/**
         * If child meshes should be ignored
         */
        public ignoreChildrenWorldMatrix = false;

PR #4849, opened 2018-07-28
copy array to avoid modifying passed in variable

TrevorDev:
http://www.html5gamedevs.com/topic/39097-defaultrenderingpipeline-remove-camera-in-list/?tab=comments#comment-223280

https://github.com/BabylonJS/Babylon.js/issues/4844


src/PostProcess/RenderPipeline/Pipelines/babylon.defaultRenderingPipeline.ts
```diff
@@ -367,6 +367,7 @@
         constructor(name: string = "", hdr: boolean = true, scene: Scene = BABYLON.Engine.LastCreatedScene!, cameras?: Camera[], automaticBuild = true) {
             super(scene.getEngine(), name);
             this._cameras = cameras ||  scene.cameras;
+            this._cameras = this._cameras.slice();
```
deltakosh: weird construct :)
we do not want to slice the scene.cameras but just the ones in params

PR #4853, opened 2018-07-30
[Review required] Module and filename renaming

RaananW:
A new unified naming convension for our filenames, including fallbacks to support the old files:

[modulename].js and [modulename].min.js , in [modulename] directory.

After merging there are many files to delete. If merged, it is recommended to 
1. delete all .js and all .d.ts files in /dist/preview release/  
2. Rebuild all modules
3. Commit the untracked files, remove unneeded files from the repository.


.gitignore
```diff
@@ -180,8 +180,8 @@ dist/preview release/package/
 /Viewer/dist/viewer.js
 /Viewer/dist/viewer.min.js
 dist/preview release/viewer/babylon.d.ts
-dist/preview release/viewer/babylonjs.loaders.d.ts
-dist/preview release/viewer/babylon.glTF2Interface.d.ts
+dist/preview release/viewer/loaders.d.ts
```
deltakosh: why not babylon.* ?

RaananW: i thought we remove the babylon. . [modulename].* , this is what we discussed

RaananW:
Working on fixing the tests, i just noticed one file didnt rename correctly

.gitignore
deltakosh: Well misunderstanding then :)  it must be babylon.modulename.* (I think the reason of the misunderstanding is that all modules should be babylon.* for me but the viewer is not following this rule)

PR #4874, opened 2018-08-01
Fixed issue where gaze trackers appear even after leaving VR.

atulyar:


deltakosh:
There is a conflict :)

PR #4899, opened 2018-08-07
Simplification As A SceneComponent

sebavan:


deltakosh:
Can you add a visual test for it?

PR #4905, opened 2018-08-08
Fix File Loading if hosted from `file:`-Protocol

ltetzlaff:
When loading a file using `file:`-Protocol loading does not work in a Safari-Webview because request-Status will be `0`.

I added a simple check for that edge case while not changing overall behavior.

Related changes before:
https://github.com/BabylonJS/Babylon.js/commit/63f753c3b51492ac66d81a2555f000f62d255c92#diff-5e88dca888b9e41c0e61475ba80ae286R502


deltakosh:
There is a missing code comment on the new function:)

ltetzlaff:
Ah sorry, contribution guideline seems to be outdated / deadlinked, I just checked https://github.com/BabylonJS/Babylon.js/blob/master/dist/preview%20release/typedocValidationBaseline.json#L18614 after the first build failed and assumed it would just get ignored like the others.

Will add comment! :+1:

deltakosh:
Thanks! Can you point me to the deadline? I'll fix it asap

ltetzlaff:
https://github.com/BabylonJS/Babylon.js/blob/master/contributing.md#pull-requests, `creating the minified version` is 404 and it just doesn't say `JSDoc` and `What's New` 😄

deltakosh:
Will improve that:)

PR #4909, opened 2018-08-09
Vr obserables fix

atulyar:


Viewer/src/managers/observablesManager.ts
```diff
@@ -54,6 +54,15 @@ export class ObservablesManager {
      */
     public onFrameRenderedObservable: Observable<any>;
 
+    /**
+     * Will notify when VR mode is entered.
+     */
+    public onEnteringVRObservable: Observable<any>;
```
TrevorDev: Can you confirm that this should be public (eg. external developers will use this), or would it make sense to be private and only used internally by the viewer?

Viewer/src/viewer/viewer.ts
```diff
@@ -653,6 +672,14 @@ export abstract class AbstractViewer {
 
         // create a new template manager for this viewer
         this.sceneManager = new SceneManager(this.engine, this._configurationContainer, this.observablesManager);
+        
+        this.sceneManager.onEnteringVRObservable.add(() => {
+            this.onEnteringVRObservable.notifyObservers(this);
+        });
+        
+        this.sceneManager.onExitingVRObservable.add(() => {
+            this.onExitingVRObservable.notifyObservers(this);
```
TrevorDev: Can you remind me all the places where there is an enteringVR event?

- vrHelper
- sceneManager
- observableManager
- viewer

Are them all right? Can you clarify why the sceneManager and observableManager ones are needed.

src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -256,6 +256,8 @@ module BABYLON {
         private _webVRrequesting = false;
         // Are we presenting to the headset right now?
         private _webVRpresenting = false;
+        // Have we entered VR?
+        private hasEnteredVR: boolean;
```
TrevorDev: _hasEnteredVR to be consistent naming with other private variables. Also, if you could add more of a description to this variable and _webVRpresenting, it would be good. I think _webVRpresenting is what the vrDevice says and _hasEnteredVR is the state of the vrExperienceHelper.

Viewer/src/managers/sceneManager.ts
```diff
@@ -59,6 +59,15 @@ export class SceneManager {
      */
     onVRConfiguredObservable: Observable<IPostConfigurationCallback<VRExperienceHelper, IVRConfiguration>>;
 
+    /**
+     * Will notify when VR mode is entered.
+     */
+    onEnteringVRObservable: Observable<any>;
```
RaananW: the obeservables are not needed here. Since the sceneManager has a (nullable) reference to the observers manager, use it. Dont add another observer which is the same. Notice that the observers that are added to the scene manager are all either configuration or very scene related.

Viewer/src/managers/sceneManager.ts
```diff
@@ -798,6 +809,12 @@ export class SceneManager {
                 });
             });
         }
+        this._vrHelper.onEnteringVRObservable.add(() => {
```
RaananW: are those observables working now aas expected?

src/Cameras/VR/babylon.vrExperienceHelper.ts
```diff
@@ -869,54 +870,60 @@ module BABYLON {
...
          * Attempt to exit VR, or fullscreen.
          */
         public exitVR() {
-            if (this.onExitingVRObservable) {
-                try {
-                    this.onExitingVRObservable.notifyObservers(this);
+            if (this.hasEnteredVR) {
+                if (this.onExitingVRObservable) {
+                    try {
+                        this.onExitingVRObservable.notifyObservers(this);
```
RaananW: is this really synchronous?  cant it fail async?

Viewer/src/viewer/viewer.ts
```diff
@@ -270,11 +283,13 @@ export abstract class AbstractViewer {
         }
 
         if (this.sceneManager.vrHelper && !this.sceneManager.vrHelper.isInVRMode) {
+
             // make sure the floor is set
             if (this.sceneManager.environmentHelper && this.sceneManager.environmentHelper.ground) {
                 this.sceneManager.vrHelper.addFloorMesh(this.sceneManager.environmentHelper.ground);
             }
 
+            this._vrToggled = true;
```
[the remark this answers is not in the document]
RaananW: I can probably answer that - the experience helper's flag is not consistent and cannot be read when the viewer needs it.

Viewer/src/viewer/viewer.ts
atulyar: After addressing this comment (https://github.com/BabylonJS/Babylon.js/pull/4909#discussion_r208828438), the onEnteringVR observable exists in the VRExperienceHelper and ObservablesManager only. 

When VR mode is entered through VRExperienceHelper.EnterVR(), SceneManager is notified and it then notifies the observers of ObservablesManager.onEnteringVRObservable. The Viewer is one of these observers and is therefore also notified when this happens. The same flow applies to exiting VR as well.

The ObservablesManager ones are needed because they act as the linked between the SceneManager and the Viewer.

Viewer/src/viewer/viewer.ts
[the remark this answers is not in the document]
atulyar: There appear to be occasions where isInVRMode does not agree with the VRExperienceHelper's state. So this sometimes causes exitVR() to be executed multiple times. To fix this, I have just introduced another variable that simply stores the VRExperienceHelper's state (ignoring any signals we are getting from hardware i.e. vrDisplay).

Viewer/src/managers/sceneManager.ts
atulyar: With my proposed changes exitVR() should only be executing its logic once when the user exits VR (as compared to multiple times before). So these observables should only be fired once per entry/exit.

Viewer/src/managers/observablesManager.ts
atulyar: These observables are public because the SceneManager and Viewer both need to be able to access them.

Viewer/src/managers/sceneManager.ts
atulyar: I've made the change.

src/Cameras/VR/babylon.vrExperienceHelper.ts
atulyar: I think this goes back to this discussion (https://github.com/BabylonJS/Babylon.js/pull/4909#discussion_r209035219).

Viewer/src/viewer/viewer.ts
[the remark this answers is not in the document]
sebavan: Should we not fix the root cause of the inconsistency instead ?

sebavan:
@RaananW and @TrevorDev I leave it to you for the merge as you are much more knowledgeable than myself on this area.

Viewer/src/viewer/viewer.ts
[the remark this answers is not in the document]
atulyar: Ideally yes, but I needed to fix this issue for my intern project. I am aware that this isn't actually fixing the root cause, but I've tested it locally and am confident that the changes will fix the issue. In the meantime, I will create an issue for the root cause fix, and will look into that if I have time at the end of my internship.

Issue #4929, opened 2018-08-13
undefined is not an object error when loading gltf


apptaro:
# Bugs

- Playground: http://www.babylonjs-playground.com/indexStable.html#WGZLGJ#235
  -> The bug would be reproducible by like above, but the playground does not work on my iPad.
- Expected result: GLTF model is loaded.
- Current result: The error below occurs:
  "undefined is not an object (evaluating 'this._gltf.nodes')

It seems that it's a timing issue and it occurs more often for older devices. I have an iPad with iOS 9.3.5 which almost always suffers from it.

The root cause of this issue is that GLTFLoader._gltf is cleared before GLTFLoader._getMeshes uses it. See these:
https://github.com/BabylonJS/Babylon.js/blob/master/loaders/src/glTF/2.0/babylon.glTFLoader.ts#L191
https://github.com/BabylonJS/Babylon.js/blob/master/loaders/src/glTF/2.0/babylon.glTFLoader.ts#L263

sebavan:
Nice catch @bghgary, @kcoley  could you take a look at this ?

sebavan:
Should be fix by https://github.com/BabylonJS/Babylon.js/pull/4934 Thx @bghgary

PR #4934, opened 2018-08-13
Fix glTF loader return order

bghgary:
See #4929


sebavan:
Thx I ll wait for the build and Merge

Issue #4960, opened 2018-08-17
Add Programatic Start Dragging to dragBehavior


sebavan:
About the rough idea:

The behavior would need a new function like startDragging(). 

By calling it, the user would be in drag mode without pressing the left click. 

Drag would then stop on keypress or any pointer action aside of move.

sebavan:
thanks @trevordev :-)

PR #4962, opened 2018-08-18
add startDrag to pointer behavior

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4960

Tested with: http://localhost:1338/Playground/index-local.html#9UFEBE#5


sebavan:
Sorry about the conflict, you ll need to merge as well ;-)

sebavan:
@trevordev Could you ensure that it snaps the model when start dragging is called and the pointer is outside the attachedNode ???

TrevorDev:
Previously after calling startDrag a pointer move was needed to snap the object. I just changed it such that when startDrag is called it will move the object to the last pointer position.

Issue #4953, opened 2018-08-17
BoundingBoxGizmo does not support pivot Matrix.


sebavan:
I’m seeing unexpected behavior of the BoundingBoxGizmo when I attach it to a mesh that happens to have had setPivotMatrix called on it. Here is a playground with a repro: https://www.babylonjs-playground.com/#XCPP9Y#601

sebavan:
Close with previous PR

PR #4968, opened 2018-08-20
add scalePivot to bounding box gizmo

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4953

This will allow scaling on a pivot to work for the gizmo. 
Test PG: http://localhost:1338/Playground/index-local.html#XCPP9Y#614
http://playground.babylonjs.com/#JVMFIH#6

There still does seem to be some unexpected behavior when using pivot though which is why this PR will set pivot to identity always when the mesh has a pivot. 

@deltakosh @sebavan 
See this PG: http://playground.babylonjs.com/#JVMFIH#2

1. Press a (to scale the box smaller)
2. Press s (to set the pivot)
3. observe the box changes position (I did not think this should happen according to the docs I read here)

Any idea if I am doing something wrong or if this is expected? 

I'd recommending merging this PR if this is okay, and making the above a separate bug if it is not intended.


sebavan:
@TrevorDev I do not know whether we should remove Pivot from the mesh this quite a destructive change if the author of the scene is using animation for instance ???

Would it be possible to remove the pivot as long as the gizmo is attached and put it back in at the end of the edit session ? This would probably require adding an extra translation to counter act the scale.

I am wondering if we should try this ?

TrevorDev:
@sebavan this is what I tried but because of the issue I mentioned (http://playground.babylonjs.com/#JVMFIH#2) this does not work as expected.

PR #4975, opened 2018-08-21
do not allow particle lifetime ratio to exceed 1

TrevorDev:
Fixes playground:
http://localhost:1338/Playground/index-local.html#PU4WYI#51

@deltakosh in the cpu version it does:
if (particle.age >= particle.lifeTime) { // Recycle by swapping with last particle
    this._emitFromParticle(particle);
    this.recycleParticle(particle);
    index--;
    continue;
}

gpu doesn't have this logic so age/lifetime is > 1 so I added the clamp.  Might be a good idea to have the same age >= lifetime logic in gpuUpdateParticles but I wasn't sure how/if i should do this.


deltakosh:
I will take care of it..Thanks a lot @TrevorDev

Issue #4980, opened 2018-08-22
Add cylinder emitter


deltakosh:
Like sphere or cube but as a cylinder (particles are emitter by the top and bottom caps)

PR #4984, opened 2018-08-23
Cylinder particle emitter

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/4980

Testing PG: http://localhost:1338/Playground/index-local.html#ZWE24S#11


deltakosh:
We have a tiny conflict

PR #4985, opened 2018-08-23
Improved _isSyncronized performance in computeWorldMatrix

Bolloxim:
Fixed an issue with using a getter to read rotationQuaternion vs _rotationQuaterion property which saved both performance and large GC hits due to temporaries created.   I was asked to investigate a GC and performance issue with large scenes for a project.  I created a 1000 object static scene of sphere's.  I ran tests comparing frozen matrices to _isSyncronized pass and also dynamic. 
The result showed GC heap hits on the getter for rotationQuaternion and optional parameter. This read a property which is also nullable  (unsure if its nullable that is causing a creation of a temp) .  However changing this to a direct property read increased performance 10% and drastically curbed GC. (got graph captures) up shot was a saving of 5% in off the top performance over the getter method (although frozen was 13% faster) however the GC spike disappeared which was a much larger benefit in smoothness capture over a 20s period


deltakosh:
Excellent! Thanks a lot!

PR #4994, opened 2018-08-23
Garbage collector optimizations

menduz:
This PR contains several optimizations to help the garbage collector.

In the case of `hasSpecificTriggers2` it is a specialized function to be used in the active mesh selector to avoid creating a _static_ array to pass two values, and then that array is discarded.

In several places, I replaced identity objects or zero objects with static instances tracking the code finding usages to ensure not breaking anything.
**Before:**
screen shot 2018-08-23 at 16 53 13
**After:**
screen shot 2018-08-23 at 16 53 35

It also removes a leak in the instantiation of Quaternions in the `Quaternion.Slerp` function in animations, sorry, I didn't take a screenshot of those


menduz:
Open question:

Why it is necessary to rebuild the entire bounding box/sphere in `reConstruct`? Can we reuse the instances of the vectors?

src/Animations/babylon.runtimeAnimation.ts
```diff
@@ -495,30 +512,30 @@
                         break;
                     // Quaternion
                     case Animation.ANIMATIONTYPE_QUATERNION:
-                        offsetValue = new Quaternion(0, 0, 0, 0);
+                        offsetValue = _staticOffsetValueQuaternion;
```
deltakosh: Can we create them on demand?

menduz: It is generated at the top of the file and shared with every animation, it is a private const available to this file only

deltakosh: Oh I missed that :)
Then you need to reset then to default value

menduz: It is not changed, I tracked that value and it is read-only, do you still want it to be reset?

menduz: Or what do you think about a Object.freeze? or extending the class with 
```ts
class StaticVector3 extends Vector3 { 
  get x() { return this.initialX } 
  set x(x) { throw }
  ...
}
```?

menduz:
They work #LL5BIQ#0 and #LL5BIQ#0. Tested on http://localhost:1338/Playground/#LL5BIQ#0. Is that ok?

src/Animations/babylon.runtimeAnimation.ts
deltakosh: We need to reset to default because it is used by the interpolator here:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Animations/babylon.animation.ts#L815

deltakosh:
Yes...just need to fix the default values and we are good

src/Animations/babylon.runtimeAnimation.ts
menduz: Ok, I'll do it, but those values are read-only. They are used like this

`return quatValue.addInPlace(offsetValue.scale(repeatCount));`
or
`return offsetValue * repeatCount + floatValue;`

And in every case, `T.scale(number)` returns a new `T` without mutating the original

deltakosh: This is a really good point :)
So don't change them !

LGTM!

deltakosh:
I will just wait for the CI to finish and I'll merge

menduz:
Adding Object.freeze as a double check, JIC

deltakosh:
Ha! you need to update the what's new :D

menduz:
https://github.com/BabylonJS/Babylon.js/pull/4994/files#diff-b34a47d0ad0e03caad31c0af018aad10R124 already did it

deltakosh:
I know but update it a bit to fix the CI

Issue #4974, opened 2018-08-21
GLTF2Export texture horizontally mirrored has no effect


sebavan:
Related to http://www.html5gamedevs.com/topic/39476-babylongltf2export-texture-horizontally-mirrored-has-no-effect/

PR #5024, opened 2018-08-29
glTFSerializer: Add texture transform support

kcoley:
This was originally going to be support for the KHR_texture_transform extension, but I ran into a bunch of edge cases in terms of implementing the fallback behavior.  So for now, if texture transform properties are applied to a texture, the texture file will be modified.  Should resolve https://github.com/BabylonJS/Babylon.js/issues/4974


deltakosh:
This looks good so far:
- Can you update the what's new file?
- Can you provide one visual test where we serialize a scene and load it again to make sure we have a validation?
- Would like to get @bghgary approval as well

kcoley:
Yeah sure thing

sebavan:
LGTM as well but completely agree on the validation test :-)

kcoley:
@sebavan @deltakosh does the validator visual test support promises?  It seems to exit before the serializer is able to execute since it waits for the scene to finish rendering.

deltakosh:
Yes but you have to turn on the polyfill: BABYLON.PromisePolyfill.Apply(true);

deltakosh:
Let's see how the CI feels about that :)

deltakosh:
Let's see how the CI feels about that :) (and you have a conflict)

PR #5027, opened 2018-08-29
#4589 - Support for previewing compressed textures in Inspector

Jaskar:
If inspector is closed / opened more than 15 times, it'll crash the canvas... But I don't think this is a big issue ?


deltakosh:
Can you update the what's new and make sure Travis CI is green ?

PR #5052, opened 2018-08-31
textures using square sizes

Vinc3r:
It always seems weird to me that we've got non-squared textures on the repo, looks like bad practice for me, so... :)


deltakosh:
Good job:)

PR #5054, opened 2018-08-31
new checker texture

Vinc3r:
we always need a checker texture, here my proposal (with a playground: https://www.babylonjs-playground.com/#J5E230#124 )

img


sebavan:
Yeah I love it :-)

PR #5058, opened 2018-09-02
Add support for multiple simultaneous connected InputText controls.

brianzinn:
I think the docs are OK as-is (https://doc.babylonjs.com/how_to/gui#virtualkeyboard).  I am unsure about action to take should the currently connected InputText be disconnected - it is not a defined behaviour in current version (PR works same as current version).  Have also fixed a possible issue with current connect + disconnect:
```javascript
keyboard.connect(input);
keyboard.connect(input);
keyboard.disconnect()
```

That would leave keyboard displaying when input has focus, but not redirecting keypresses.  PR ignores second connect as base case.

### DIFF
--- a/gui/src/2D/controls/virtualKeyboard.ts
+++ b/gui/src/2D/controls/virtualKeyboard.ts
@@ -28,6 +28,12 @@ export class KeyPropertySet {
     background?: string;
 }
 
+type ConnectedInputText = {
+    input: InputText,
+    onFocusObserver: Nullable<Observer<InputText>>,
+    onBlurObserver: Nullable<Observer<InputText>>
+}
+
 /**
  * Class used to create virtual keyboard
  */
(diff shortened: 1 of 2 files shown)


deltakosh:
what do you mean with "action to take should the currently connected InputText be disconnected " ?

deltakosh:
Also do you mind adding a dispose function that will call disconnect() ? thanks :)

deltakosh:
(You should also credit yourself in the what's new)

gui/src/2D/controls/virtualKeyboard.ts
```diff
@@ -151,84 +157,123 @@ export class VirtualKeyboard extends StackPanel {
...
-        if (!this._connectedInputText) {
-            return;
+    public disconnect(input?: InputText): void {
+        if (input) {
+            // .find not available on IE
+            let filtered = this._connectedInputTexts.filter(a => a.input === input);
+            if (filtered.length === 1) {
+                this._removeConnectedInputObservables(filtered[0]);
+                
+                this._connectedInputTexts = this._connectedInputTexts.filter(a => a.input !== input);
```
deltakosh: You should remove the input from the _connectedInputTexts  array as well

gui/src/2D/controls/virtualKeyboard.ts
```diff
@@ -151,84 +157,123 @@ export class VirtualKeyboard extends StackPanel {
...
     /**
      * Connects the keyboard with an input text control
+     * 
      * @param input defines the target control
      */
     public connect(input: InputText): void {
-        this.isVisible = false;
-        this._connectedInputText = input;
+        // .find not available on IE
+        let filtered = this._connectedInputTexts.filter(a => a.input === input);
```
deltakosh: can we use array.some here?

gui/src/2D/controls/virtualKeyboard.ts
```diff
@@ -151,84 +157,123 @@ export class VirtualKeyboard extends StackPanel {
...
+     * 
+     * @param input optionally defines a target control, otherwise all are disconnected
      */
-    public disconnect(): void {
-        if (!this._connectedInputText) {
-            return;
+    public disconnect(input?: InputText): void {
+        if (input) {
+            // .find not available on IE
+            let filtered = this._connectedInputTexts.filter(a => a.input === input);
```
deltakosh: array.some() ?

gui/src/2D/controls/virtualKeyboard.ts
deltakosh: Interesting. ist is faster than an indexOf + splice?

gui/src/2D/controls/virtualKeyboard.ts
brianzinn: yes, much better.  I'll switch to .some() and add dispose()

gui/src/2D/controls/virtualKeyboard.ts
brianzinn: I don't think .filter is faster, but I am trying to remove based on an object property equality, not the entire object (a => a !== input).  It became an object with extra properties for the observables to remove.  Does it still look OK?

deltakosh: Oh yeah makes sense! Thanks

deltakosh:
Don't forget the what's new:)

Issue #5047, opened 2018-08-30
Add support for subEmitters at start


deltakosh:
- [x] Give particle systems the option to spawm all the subEmitters instead of one from the list of templates
- [x] Let subEmitters be created either at the beginning or at the end. If at the beginning they could be attached to the particle position or not

PR #5068, opened 2018-09-05
sub emitters update to support attached

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/5047

test PG: http://localhost:1338/Playground/index-local.html#T0L01N#43


src/Particles/babylon.particleSystem.ts
```diff
@@ -865,11 +985,12 @@
          * Its lifetime will start back at 0.
          */
         public recycleParticle: (particle: Particle) => void = (particle) => {
-            var lastParticle = <Particle>this._particles.pop();
-            if (lastParticle !== particle) {
-                lastParticle.copyTo(particle);
+            // move particle from activeParticle list to stock particles
+            var index = this.particles.indexOf(particle);
```
TrevorDev: Note: The change for this section will remove instead of push/pop which may introduce a perf hit. 
this change was needed to ensure the same particle created is the one that gets disposed (I think it was a bug)

src/Particles/babylon.particleSystem.ts
```diff
@@ -890,6 +1011,19 @@
             } else {
                 particle = new Particle(this);
             }
+
+            // Attach emitters
+            if(this._subEmitters && this._subEmitters.length > 0){
+                var subEmitters = this._subEmitters[Math.floor(Math.random() * this._subEmitters.length)];
+                particle._attachedSubEmitters = [];
+                subEmitters.forEach((subEmitter)=>{
+                    if(subEmitter.type == SubEmitterType.ATTACHED){
```
deltakosh: ===

src/Particles/babylon.particleSystem.ts
```diff
@@ -905,16 +1039,20 @@
...
-            var templateIndex = Math.floor(Math.random() * this.subEmitters.length);
-
-            var subSystem = this.subEmitters[templateIndex].clone(this.name + "_sub", particle.position.clone());
-            subSystem._rootParticleSystem = this;
-            this.activeSubSystems.push(subSystem);
-            subSystem.start();
+            var templateIndex = Math.floor(Math.random() * this._subEmitters.length);
+            
+            this._subEmitters[templateIndex].forEach((subEmitter)=>{
+                if(subEmitter.type == SubEmitterType.END){
```
deltakosh: ===

src/Particles/babylon.particleSystem.ts
TrevorDev: Did perf test it was faster the other way. (tried 3 methods)

image


http://localhost:1338/Playground/index-local.html#6W42J5#7

PR #5088, opened 2018-09-07
compute world matrix of children

TrevorDev:
@deltakosh Double check that this looks like the right fix. It seems as though the controller mesh is not part of scene.meshes. Do you expect that the recursive world matrix logic should already exist somewhere else or would you expect scene.meshes contain the controller?


deltakosh:
Unfortunately this is not good as: it will compute wm on all nodes even if you have an octree. Also this is part of performance issue we identified (double tree traversal)

Why the mesh.computrWorldMatrix() not enough?

deltakosh:
If you can repro the issue in a PG I will make sure to fix it

deltakosh:
OK let's do that :)

PR #5094, opened 2018-09-09
Add support for glTF validation

### LINKED ISSUE #4094: Add glTF-validator to sandbox
Integrate glTF-validator with sandbox so that validation is run on any glTF file viewed in the sandbox and any issues are reported.

bghgary:
Initial version of glTF validator integration.

Use http://localhost:1338/Playground/index-local.html#PXQ6Q7 for an example.

image

image

Showing raw JSON for now.
image

Fixes #4094


sebavan:
SWEEEEET !

bghgary:
@deltakosh This is ready to go if it can pass the tests. There seems to be a failure on the documentation check that I don't think is related to this change.

PR #5160, opened 2018-09-17
Dev add remove materials observables

yuccai:
Hi @deltakosh,

First, I fixed a bug regarding "hasRenderTargetTextures" property of Material class. It should be a getter like its child classes. Unless, we have a writing error because "hasRenderTargetTextures" is a getter in the child class but it is a property set to false in Material constructor class.

The main subject of this PR is about to add two observables for notifying materials change : "onNewMaterialAddedObservable" and "onMaterialRemovedObservable".
This is similar to the existing observables : "onNewMeshAddedObservable" and "onMeshRemovedObservable".

What do you think about these kind of observables and about the way they are implemented ?
Is it relevant ? 
I also like to do the same with textures, is it also relevant ? If yes, can I keep the same PR or should I create a new one ?

### DIFF
--- a/src/Loading/Plugins/babylon.babylonFileLoader.ts
+++ b/src/Loading/Plugins/babylon.babylonFileLoader.ts
@@ -73,7 +73,6 @@
                     var parsedMaterial = parsedData.materials[index];
                     var mat = Material.Parse(parsedMaterial, scene, rootUrl);
                     container.materials.push(mat);
-                    container.scene.onNewMaterialAddedObservable.notifyObservers(mat);
                     log += (index === 0 ? "\n\tMaterials:" : "");
                     log += "\n\t\t" + mat.toString(fullDetails);
                 }
--- a/src/Loading/Plugins/babylon.babylonFileLoader.ts
+++ b/src/Loading/Plugins/babylon.babylonFileLoader.ts
@@ -73,6 +73,7 @@
                     var parsedMaterial = parsedData.materials[index];
                     var mat = Material.Parse(parsedMaterial, scene, rootUrl);
                     container.materials.push(mat);
+                    container.scene.onNewMaterialAddedObservable.notifyObservers(mat);
                     log += (index === 0 ? "\n\tMaterials:" : "");
                     log += "\n\t\t" + mat.toString(fullDetails);
                 }
--- a/src/Materials/babylon.material.ts
+++ b/src/Materials/babylon.material.ts
@@ -789,6 +789,7 @@ module BABYLON {
 
             if (!doNotAdd) {
                 this._scene.materials.push(this);
+                this._scene.onNewMaterialAddedObservable.notifyObservers(this);
             }
         }
 
--- a/src/Materials/babylon.material.ts
+++ b/src/Materials/babylon.material.ts
@@ -1,4 +1,4 @@
-module BABYLON {
+﻿module BABYLON {
     /**
      * Manages the defines for the Material
      */
@@ -1270,6 +1270,7 @@ module BABYLON {
             if (index >= 0) {
                 this._scene.materials.splice(index, 1);
             }
+            this._scene.onMaterialRemovedObservable.notifyObservers(this);
 
             // Remove from meshes
             for (index = 0; index < this._scene.meshes.length; index++) {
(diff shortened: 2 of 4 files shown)


src/Materials/babylon.material.ts
```diff
@@ -466,8 +466,10 @@
         /**
          * Gets a boolean indicating that current material needs to register RTT
          */
-        public hasRenderTargetTextures = false;
-
+        public get hasRenderTargetTextures(): boolean {
```
deltakosh: Hello not sure why this is a problem? The child can still override it with a property even if it is not a property at base class level

src/babylon.scene.ts
```diff
@@ -3220,6 +3232,7 @@
          */
         public addMaterial(newMaterial: Material): void {
             this.materials.push(newMaterial);
+            this.onNewMaterialAddedObservable.notifyObservers(newMesh);
```
deltakosh: I'm fine with these observables. that's a good idea.  You only need to clear them at dispose time

src/Materials/babylon.material.ts
yuccai: This bug occurs since version 3.3.0-beta.6.
Here is the error : 
image
The parent class define a property that could be overriden by its children. But some of its children have the same property name defined as a getter. So there is a conflict.
In the class child, when we want to set the property, it considers it as a getter and prevent setting a value because the setter doesn't exist.

deltakosh: ok makes sense..So good for this change! Thanks!

src/babylon.scene.ts
```diff
@@ -439,6 +439,16 @@
         * An event triggered when a mesh is removed
         */
         public onMeshRemovedObservable = new Observable<AbstractMesh>();
+        
+        /**
+        * An event triggered when a mesh is created
+        */
+        public onNewMaterialAddedObservable = new Observable<Material>();
```
deltakosh: Please fix the comments :)

src/babylon.scene.ts
deltakosh: Hello please do it asap because we are actually in code freeze and I should not accept any more changes for 3.3. But as you started the PR the day before I'm doing an exception.
Regarding co we are good but the content of your comments is wrong

besides that we can merge when you want. I'll keep this PR open for now waiting for the additional observables

src/babylon.scene.ts
```diff
@@ -4656,6 +4669,25 @@
             this.onBeforeRenderingGroupObservable.clear();
             this.onAfterRenderingGroupObservable.clear();
             this.onMeshImportedObservable.clear();
+            this.onBeforeCameraRenderObservable.clear();
```
deltakosh: Thanks!

src/babylon.scene.ts
deltakosh: by the way what is "newMesh" here?

src/babylon.scene.ts
yuccai: done : 269c189873e036f37218eff4a2682b2d3281ab29

src/babylon.scene.ts
yuccai: done : b264afcfacee97923918acd32f1da27fa35ae244

yuccai: I did it as soon as I can in a single commit c900ceae4819e07083fa2e69abad1584c7e929b3.
I hope I didn't forget places to notify observers.

src/Loading/Plugins/babylon.babylonFileLoader.ts
```diff
@@ -73,6 +73,7 @@
                     var parsedMaterial = parsedData.materials[index];
                     var mat = Material.Parse(parsedMaterial, scene, rootUrl);
                     container.materials.push(mat);
+                    container.scene.onNewMaterialAddedObservable.notifyObservers(mat);
```
deltakosh: I think this one should be removed as the material is not in the scene technically

yuccai: done : 52520ba30c977914c2d18bd23b1f7ffb9df2e93a

deltakosh:
thanks!

PR #5162, opened 2018-09-17
Audio Policy Unlock Fix

sebavan:


src/Audio/babylon.audioEngine.ts
```diff
@@ -97,11 +120,22 @@
          */
         public unlocked: boolean = false;
 
+        /**
+         * Defines if the audio engine relies on a custom unlocked button.
+         * In this case, the embedded button will not be displayed.
+         */
+        public useCutomUnlockedButton: boolean = false;
```
deltakosh: CuStom

sebavan: Sounds better I agree :-)

PR #5165, opened 2018-09-17
Add more configuration for arc rotate

sebavan:


src/Behaviors/Cameras/babylon.framingBehavior.ts
```diff
@@ -146,6 +146,12 @@ module BABYLON {
             return this._framingTime;
         }
 
+        /**
+         * Define if the behavior should automatically correct change the configured
```
deltakosh: remove "correct" I guess

PR #5179, opened 2018-09-18
FIxing StandardRenderingPipleine that always adds the original post-process

julien-moreau:
FIxing StandardRenderingPipleine that always adds the original post-process to the registered effects (Anti Aliasing issue)
As the pipeline will not have more post-processes in it, I don't think this "if" from evil is a problem until I pull-request for a better solution. Do you agree? :)


deltakosh:
This is fine for me!

PR #5186, opened 2018-09-18
multiLine fix

royibernthal:
Untested but should work, fairly simple changes.

### DIFF
--- a/gui/src/2D/controls/multiLine.ts
+++ b/gui/src/2D/controls/multiLine.ts
@@ -62,7 +62,7 @@ export class MultiLine extends Control {
 
     /** Function called when a point is updated */
     public onPointUpdate = (): void => {
-        this._markAsDirty();
+        if (this.isVisible) this._markAsDirty();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


gui/src/2D/controls/multiLine.ts
```diff
@@ -62,7 +62,7 @@ export class MultiLine extends Control {
 
     /** Function called when a point is updated */
     public onPointUpdate = (): void => {
-        this._markAsDirty();
+        if (this.isVisible) this._markAsDirty();
```
deltakosh: We need to add a MarkasDirty when is Visible will become true again then

deltakosh: Let me fix that one

royibernthal: I'm on it actually :)

deltakosh: I will resolve it at control level

deltakosh: You can remove it, it will be handle for all controls :)

royibernthal: Okay

royibernthal: Good call, it does make sense to handle this case for all controls :)

royibernthal:
Other than what you're doing at control level - it is of course tested now, forgot to mention

PR #5190, opened 2018-09-18
Misc docs

TrevorDev:


src/Culling/babylon.boundingInfo.ts
```diff
@@ -22,22 +22,61 @@
         return extentsOverlap(result0.min, result0.max, result1.min, result1.max);
     }
 
+    /**
+     * Interface for cullable objects see https://doc.babylonjs.com/babylon101/materials#back-face-culling
```
deltakosh: You need to use @see (after a line feed)

PR #5191, opened 2018-09-19
scene fog API comments

Vinc3r:
it seems that my ATOM configuration automatically remove useless empty spaces char, is it annoying or kind-of good practice?

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -696,7 +696,7 @@ module BABYLON {
         /**
         * Gets or sets the fog color to use
         * @see http://doc.babylonjs.com/babylon101/environment#fog
-        * Default is Color3(0.2, 0.2, 0.3)
+        * (Default is Color3(0.2, 0.2, 0.3))
         */
         public fogColor = new Color3(0.2, 0.2, 0.3);
         /**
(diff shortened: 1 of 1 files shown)


src/babylon.scene.ts
```diff
@@ -689,21 +696,25 @@ module BABYLON {
         /**
         * Gets or sets the fog color to use
         * @see http://doc.babylonjs.com/babylon101/environment#fog
+        * Default is Color3(0.2, 0.2, 0.3)
```
deltakosh: Because the Doc will concatenate lines, can you add parenthesis around the default value line?

src/babylon.scene.ts
```diff
@@ -689,21 +696,25 @@ module BABYLON {
         /**
         * Gets or sets the fog color to use
         * @see http://doc.babylonjs.com/babylon101/environment#fog
+        * (Default is Color3(0.2, 0.2, 0.3))
         */
         public fogColor = new Color3(0.2, 0.2, 0.3);
         /**
         * Gets or sets the fog density to use
         * @see http://doc.babylonjs.com/babylon101/environment#fog
+        * Default is 0.1
```
deltakosh: for ALL default value lines :)

Vinc3r: erf ok :)

Vinc3r: should be OK now... isn't it?

deltakosh:
Wonderful! thanks

PR #5193, opened 2018-09-19
FAT Objects

sebavan:


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -1,4 +1,34 @@
 ﻿module BABYLON {
+    /** @hidden */
+    class FacetData {
```
deltakosh: I would recommend using _FacetDataStorage

PR #5204, opened 2018-09-20
Fix normalize weights

Bolloxim:
Fixed an issue with badly weighted GLTF models.  (see attached tree)  On reviewing the weights it was pretty obvious the weights were all off and even had some set to 0,0,0,0 which is pretty bad data.  however on correcting for this by renormalizing the mesh (did this in a playground test, script attached).   So this has been added now to babylon.mesh.ts.  I noticed there was a cleanMatrixWeights I test this but it did not clean up the model.  However I added functions after this.    It should handle extra weights fine but I've no test data to attempt it.  I split the functionality into two private functions so that its much cleaner if something goes awry its self contained .   
Tougher implementation was adding this call to the GLTFLoader I put this in a promise right after we bind the gometry to the mesh and then I process renormalization.  It works in sandbox and playground. modes I tested both.  I had to update manually declaration modules in preview release to get loaders to rebuild.  This likely needs some review but it does clean up nicely.   i will open an issue for supporting weight / influence debugging and optimizations to reduce large influence counts, which I will like implement at somepoint. 

[tree-fix.js.txt](https://github.com/BabylonJS/Babylon.js/files/2400317/tree-fix.js.txt)
[Tree1.gltf.zip](https://github.com/BabylonJS/Babylon.js/files/2400321/Tree1.gltf.zip)

### DIFF
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -1672,13 +1672,13 @@
             }
         }
 
-        /*
-            Renormalize the mesh and patch it up if there are no weights
-            Similar to normalization by adding the weights comptue the reciprical and multiply all elements. this wil ensure that everything adds to 1. 
-            However in the case of 0 weights then we set just a single influence to 1. 
-            We check in the function for extra's present and if so we use the normalizeSkinWeightsWithExtras rather than the FourWeights version. 
+        /**
+         *   Renormalize the mesh and patch it up if there are no weights
+         *   Similar to normalization by adding the weights comptue the reciprical and multiply all elements. this wil ensure that everything adds to 1. 
+         *   However in the case of 0 weights then we set just a single influence to 1. 
+         *   We check in the function for extra's present and if so we use the normalizeSkinWeightsWithExtras rather than the FourWeights version. 
          */
-        public normalizeSkinWeights():void {
+        public normalizeSkinWeights(): void {
             if (this.isVerticesDataPresent(VertexBuffer.MatricesWeightsKind)) {
                 if (this.isVerticesDataPresent(VertexBuffer.MatricesWeightsExtraKind)) {
                     this.normalizeSkinWeightsAndExtra();
@@ -1688,8 +1688,8 @@
                 }
             }    
         }
-
-        private normalizeSkinFourWeights():void {
+        // faster 4 weight version. 
+        private normalizeSkinFourWeights(): void {
             let matricesWeights = (<FloatArray>this.getVerticesData(VertexBuffer.MatricesWeightsKind));
             let numWeights = matricesWeights.length;
             for (var a=0; a<numWeights; a+=4){
@@ -1697,7 +1697,7 @@
                 // check for invalid weight and just set it to 1.
                 if (t===0) matricesWeights[a] = 1;
                 else{
-                    // renormalize so everything adds to 1
+                    // renormalize so everything adds to 1 use reciprical
                     let recip = 1 / t;
                     matricesWeights[a]*=recip;
                     matricesWeights[a+1]*=recip;
@@ -1708,8 +1708,8 @@
             }
             this.setVerticesData(VertexBuffer.MatricesWeightsKind, matricesWeights);
         }
-
-        private normalizeSkinWeightsAndExtra():void {
+        // handle special case of extra verts.  (in theory gltf can handle 12 influences)
+        private normalizeSkinWeightsAndExtra(): void {
             let matricesWeightsExtra = (<FloatArray>this.getVerticesData(VertexBuffer.MatricesWeightsExtraKind));
             let matricesWeights = (<FloatArray>this.getVerticesData(VertexBuffer.MatricesWeightsKind));
             let numWeights = matricesWeights.length;
@@ -1719,12 +1719,13 @@
                 // check for invalid weight and just set it to 1.
                 if (t===0) matricesWeights[a] = 1;
                 else{
-                    // renormalize so everything adds to 1
+                    // renormalize so everything adds to 1 use reciprical 
                     let recip = 1 / t;
                     matricesWeights[a]*=recip;
                     matricesWeights[a+1]*=recip;
                     matricesWeights[a+2]*=recip;
                     matricesWeights[a+3]*=recip;
+                    // same goes for extras
                     matricesWeightsExtra[a]*=recip;
                     matricesWeightsExtra[a+1]*=recip;
                     matricesWeightsExtra[a+2]*=recip;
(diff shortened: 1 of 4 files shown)


deltakosh:
I think this should be being a boolean to make sure we do not slow loading down for users having correct gltf
Adding @bghgary for final review

Only one question: what was wrong with the current cleanMatricesWeights function?

Bolloxim:
Agreed on khronos spec says, but it seems some exporters have not read those specs :)

So no boolean in the loader.  I'll rework it so that there is a mesh validator in mesh that can 'test' geometry and flag it as broken rather than just changing the data.  That way it can flag it for re-export and normalization. Plus we only patch data then for those that need it.   sound good ?

as for cleanMatrixCode caused geometry to map to another point not within the tree but off to the edge and all broken geometry kind of morphed into that point.   I did not review the code though as I thought it did a bit more than normalization, I came across it in the mesh when looking for a suitable place to move it too. 

It is also worth noting that a diagnostics pass on the weights could lend to using optimized shaders for 2 and 3 influences.  We could spit that out in the validator phase. 

lastly should we enable mesh validation  and patching in the sandbox viewer by default ?  That way assets look clean and the view can give artists immediate feedback on an exported model ?

Bolloxim:
removed the GLTFloader changes.  Created a new validateSkinning function that does the diagnostics of weight counts influences etc..should be useful for the viewer.

example of how I used it in the playground code. 
```
      var skinReporting = tree.validateSkinning();
      if (skinReporting.valid==false) tree.normalizeSkinWeights();
      console.log("Skin Validation Report:\n" + skinReporting.report);
```
Here is the output from running it, I check for sorted which is useful for dynamic skinning especially at distance dropping to two influences for example.  They should be sorted output, which was the case in the broken model. it also validates the bone indices are in range. If bone indices are out of range it generates an invalid report
```
Skin Validation Report:
Number of Weights = 3566
Maximum influences = 4
Missing Weights = 246
Not Sorted = 0
Not Normalized = 3241
WeightCounts = [246,1984,438,633,265]
Number of bones = 8
Bad Bone Indices = 0
```

Bolloxim:
I figured out the reason that cleanMatrixWeights did not work on this model.   The noBoneInfluece index is used to set for models with no weights therefore although it has a weighted value of 1-epsilon it is using the base transform node for this model which is translated off the origin basically stretched all those vertices.  Where as the normalizeSkinWeights uses the root(0) index and that had the benefit of keeping the geometry relative to the root node and does not modify the indices.

deltakosh:
So I think your technique is better. Would you mind replacing your code into the cleanMatricesWeights function? So we don't end up with two functions doing the same thing

Bolloxim:
Updated the whats new doc to reflect the modification to cleanMatrixWeights.  Seperated out the validate skin line.  I am looking to update the validator in viewer for GLTF to use this. In the viewer on error detection shall we auto correct or add a clean button ? or both ? validator might have a json output option just to make merging in to the gltf validator simple.

src/Mesh/babylon.mesh.ts
```diff
@@ -1621,79 +1621,162 @@
         }
 
         /**
-         * Normalize matrix weights so that all vertices have a total weight set to 1
+         *   Renormalize the mesh and patch it up if there are no weights
+         *   Similar to normalization by adding the weights comptue the reciprical and multiply all elements. this wil ensure that everything adds to 1.
```
deltakosh: Typos

src/Mesh/babylon.mesh.ts
```diff
@@ -1621,79 +1621,162 @@
...
             }
+            this.setVerticesData(VertexBuffer.MatricesWeightsKind, matricesWeights);
+            this.setVerticesData(VertexBuffer.MatricesWeightsKind, matricesWeightsExtra);
+        }
 
-            this.setVerticesData(VertexBuffer.MatricesIndicesKind, matricesIndices);
-            if (matricesIndicesExtra) {
-                this.setVerticesData(VertexBuffer.MatricesIndicesExtraKind, matricesIndicesExtra);
+        /**
+         * ValidateSkinning is used to determin that a mesh has valid skinning data along with skin metrics, if missing weights,
```
deltakosh: Typos

src/Mesh/babylon.mesh.ts
```diff
@@ -1621,79 +1621,162 @@
...
+        public validateSkinning() : {skinned:boolean, valid:boolean, report:string} {
+            let matricesWeightsExtra = (<FloatArray>this.getVerticesData(VertexBuffer.MatricesWeightsExtraKind));
+            let matricesWeights = (<FloatArray>this.getVerticesData(VertexBuffer.MatricesWeightsKind));
+            if (matricesWeights===null || this.skeleton==null)
+            {
+                return {skinned:false, valid: true, report:"not skinned"}
+            }
+
+            let numWeights = matricesWeights.length;
+            let numberNotSorted :number = 0;
```
deltakosh: Type should be declared with Var f: type =
(the space is after the :)

We will add a tslinter soon to make sure to fix that automatically :)

deltakosh:
Regarding autocorrect I'm not sure we want to pay the validation cost each time we load a mesh. This has to be a developer decision.
But this is an excellent candidate to go into the inspector. When a mesh is selected we could add a validate/fix bone weights button in the ui

bghgary:
For glTF specifically, the glTF validator will eventually have support for (validating skins)[https://github.com/KhronosGroup/glTF-Validator/issues/58] and the sandbox will automatically pick up these errors when loading a glTF file.

src/Mesh/babylon.mesh.ts
Bolloxim: changed, and modified punctuation slightly

src/Mesh/babylon.mesh.ts
Bolloxim: I also did a lint pass manually over the code to keep it consistent with the style over all functions. hopefully that looks cleaner

src/Mesh/babylon.mesh.ts
Bolloxim: fixed

Bolloxim:
David, I was intending to mean as part of validator when it validates the skin.  As specified before auto patching should not be required per mesh but the viewer tool should be able to resolve errors I believe automatically without adding a general overhead to the systems.   I reviewed the inspector code last night.  Couldnt quite figure out the correct location to make a call validate mesh that is only in the validator callback and not in the mesh loader.  I will be looking at that more today. Feel free to push me in the right direction here :)

deltakosh:
Will merge as soon as I arrive at work. Thanks a lot
For the viewer that could be definitely an option ( off by default)

Issue #5050, opened 2018-08-31
Node Graph Recursion, Performance and Improvements 


Bolloxim:
During my endeavours to improve object rendering performance, we fixed an issue with graphing always having to update parents submission #5040 

over a 10 second window we walked we spent 1362ms walking graphs to validate they are not dirty, basically 13% overhead in graph traversal  this is not insignificant. 

Nodes use code recursive not data recursive methods, mainly because the nodes could evaluate themselves differently.  

In addition to this freezing matrices has to be done per node. This should be per branch or even tree.  If I freeze the root with all children this should explicitly set all children. This could mean we use the a set of root flags to determine frozen states for static geometry once the world matrix is computed it remains static.  Having a root object per node shortcut would then let us just skip to the head of the line to determine node properties for the tree or branch quickly thus removing the necessity to check all parents states. 

Im proposing a couple of changes here. 
1. a method to freeze all matrices in the tree by selecting the root node. Along with a method to detect that changes at the node level rapidly without code recursion.
2. a method to flag the set and test the dirty flag on the trees depth.
3. for animations a fast data recursive method to traverse the nodes rapidly looking for hierarchy changes
4. re-evaluate performance and determine if we need any additional steps.

deltakosh:
This is really great!!

I'm now really curious on how you would like to implement all these great ideas

Bolloxim:
**Node Graph Changes Proposals** 

Some fundamentals

- The node graph is code recursive. 
- Its traverses up when computing transform nodes
- Normally this means checks if cached all the way up. 
- Then a child can get rendered and we do the same thing.
- Its kind of cumbersome and inefficient.

This is due to evaluateActiveMeshes actually walking all meshes and computesWorldMatrix on each mesh.. This means we walk the tree many times.  Yep not good. 

Ideally we would want to do a run over the entire graph once and as quick as possible this means flagging the graph as to what’s changed and where. Simplest is to flag a rootNodeContainer in the scene which all subsequent children can talk to,  should for example a child node get updated it would register itself with the rootnode as being updated. 

So we create a new function ComputeAllWorldTransforms. This will walk through the tree and update all objects however we should clue list tree traversal into being somewhat smart. Why walk the entire tree if branches arent even dirty so we can easily do this just on the parent tree. 

First steps
Node structure adds value and a rootNodeReference : Node 

When we do any processing of transforms, IE change position etc we update the rootNodeReference.DirtyFlag = true;

[...]

PR #5212, opened 2018-09-20
Fix performance issue related to the rootNodes array

barroij:
All the nodes that does not have any parent are store in an array since commit  9094b709 (part of https://github.com/BabylonJS/Babylon.js/issues/5050). 

This change caused a performance regression as each time a node is switched from a "no parent" state to a "with parent" state, ther is a O(n) lookup in the array in order to remove it.  And it happens even if a parent is passed in the constructor of a Mesh, as the the addition to the array is always done in the Node base class constructor.

This PR fix this issue by storing and keeping up-to-date, in the Node, the index within the rootNodes array, so that removal can be done in O(1)

Please note that it assumes that the rootNodes array is NEVER modified from elsewhere. If so, it is up to the culprit to make sure that the indices remain up-to-date

Some discussion about this issue and the profiling that leads to this PR can be found here : http://www.html5gamedevs.com/topic/40058-linesmesh-and-createinstance/?do=findComment&comment=228801

### DIFF
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -223,10 +223,6 @@
 
             scene = this.getScene();
 
-            if (parent || (source && source.parent)) {
-                this.removeFromSceneRootNodes();
-            }
-
             if (source) {
                 // Geometry
                 if (source._geometry) {
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -223,6 +223,10 @@
 
             scene = this.getScene();
 
+            if (parent || (source && source.parent)) {
+                this.popFromSceneRootNodes();
+            }
+
             if (source) {
                 // Geometry
                 if (source._geometry) {
@@ -280,9 +284,6 @@
                 if (Tags && Tags.HasTags(source)) {
                     Tags.AddTagsTo(this, Tags.GetTags(source, true));
                 }
-
-                // Parent
-                this.parent = source.parent;
 
                 // Pivot
                 this.setPivotMatrix(source.getPivotMatrix());
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -224,7 +224,7 @@
             scene = this.getScene();
 
             if (parent || (source && source.parent)) {
-                this.popFromSceneRootNodes();
+                this.removeFromSceneRootNodes();
             }
 
             if (source) {
--- a/src/Mesh/babylon.mesh.ts
+++ b/src/Mesh/babylon.mesh.ts
@@ -281,6 +281,9 @@
                     Tags.AddTagsTo(this, Tags.GetTags(source, true));
                 }
 
+                // Parent
+                this.parent = source.parent;
+
                 // Pivot
                 this.setPivotMatrix(source.getPivotMatrix());
 
--- a/src/babylon.node.ts
+++ b/src/babylon.node.ts
@@ -169,7 +169,7 @@
             }
         }
 
-        protected removeFromSceneRootNodes() {
+        private removeFromSceneRootNodes() {
             if (this._sceneRootNodsIndex !== -1) {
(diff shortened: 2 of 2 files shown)


src/Mesh/babylon.mesh.ts
```diff
@@ -281,9 +281,6 @@
                     Tags.AddTagsTo(this, Tags.GetTags(source, true));
                 }
 
-                // Parent
-                this.parent = source.parent;
```
barroij: useless, it is set by the `Tools.DeepCopy`above :
```javascript
                 Tools.DeepCopy(source, this, ["name", "material", "skeleton", "instances", "parent", "uniqueId",
                     "source", "metadata", "hasLODLevels", "geometry", "isBlocked", "areNormalsFrozen",
                     "source", "metadata", "hasLODLevels", "geometry", "isBlocked", "areNormalsFrozen",
                     "onBeforeDrawObservable", "onBeforeRenderObservable", "onAfterRenderObservable", "onBeforeDraw"
                     "onBeforeDrawObservable", "onBeforeRenderObservable", "onAfterRenderObservable", "onBeforeDraw"
                 ],
]
````

Comment (deltakosh) [reply src/Mesh/babylon.mesh.ts]: source is in the avoid list

File: src/babylon.node.ts [hunk 2]
```diff
@@ -110,6 +110,9 @@
         /** @hidden */
         public _worldMatrixDeterminant = 0;        
 
+        /** @hidden */
+        private _sceneRootNodsIndex = -1;
```
deltakosh: NodEs

src/babylon.node.ts
```diff
@@ -135,10 +138,7 @@
                     this._parentNode._children.splice(index, 1);
                 }
 
-                if (!parent) {
-                    // Need to add this node to the rootNodes
-                    this._scene.rootNodes.push(this);
-                }
+                if (!parent) this.addToSceneRootNodes();
```
deltakosh: can you add {} and a line feed? (just cosmetic)

src/babylon.node.ts
```diff
@@ -151,14 +151,7 @@
...
 
-                if (!previousParentNode) {
-                    // Need to remove from rootNodes
-                    const rootNodeIndex = this._scene.rootNodes.indexOf(this);
-
-                    if (rootNodeIndex > -1) {
-                        this._scene.rootNodes.splice(rootNodeIndex, 1);
-                    }
-                }
+                if (!previousParentNode) this.removeFromSceneRootNodes();
```
deltakosh: can you add {} and a line feed? (just cosmetic)

Hersir88: would be nice in future to add tslint rules for such cases

src/Mesh/babylon.mesh.ts
barroij: will remove, not related to the PR anyway

barroij:
@deltakosh should be better now.

src/babylon.node.ts
```diff
@@ -110,6 +110,9 @@
         /** @hidden */
         public _worldMatrixDeterminant = 0;        
 
+        /** @hidden */
+        private _sceneRootNodesIndex = -1;
+
```
barroij: Note that if the order of children of a node does not matter, we could also keep the `_parentChildNodeIdx`to get a O(1) time removal from parent children array

src/babylon.node.ts
barroij: I agree, everything that could ensure coherent and automatic formatting is worth beeing added (tslin, prettier, pre-commit hook).

Another example, all the trailling whitespace should be removed they just add noise if your editor displays them and are likely to cause useless change in commit.

deltakosh: I agree. If someone wants to volunteer to set it up in the project, I'll appreciate the help

src/babylon.node.ts
deltakosh: Yes agree

PR #5217, opened 2018-09-21
reduce the number of created Vector3 during ExtractMinAndMax function

barroij:
During some memory profiling I notticed that there were alot of Vector3 allocation in function `ExtractMinAndMax`, which is not really needed as there are only temporary values.


barroij:
ping @deltakosh , another performance related PR for you :)

PR #5219, opened 2018-09-21
bounding box performance improvement

barroij:
the `reConstruct` function creates new instances of Vector3 and array each time it is called, which is a bit overkill (it is 27 Vector allocation + 3 array allocation).

I changed the code so that all the instances are updated in place.

side question : is it a coding standard to not get the member with default values when there are to be created in the constructor anyway ?

For instance, in the `BoundingBox ` class, only `vectorsWorld` is given a default value, the others are supposed to be initialized in the constructor (as they are not Nullable)

### DIFF
--- a/src/Culling/babylon.boundingBox.ts
+++ b/src/Culling/babylon.boundingBox.ts
@@ -1,4 +1,13 @@
 ﻿module BABYLON {
+
+    
+    /**
+     * @hidden
+     */
+    const tempVector3Diff = new Vector3();
+    const tempVector3Min = new Vector3();
+    const tempVector3Max = new Vector3();
+
     /**
      * Class used to store bounding box information
      */
@@ -30,7 +39,7 @@
         /**
          * Gets the 8 vectors representing the bounding box in world space
          */        
-        public vectorsWorld: Vector3[] = new Array<Vector3>();
+        public vectorsWorld: Vector3[];
         /**
          * Gets the minimum vector in world space
          */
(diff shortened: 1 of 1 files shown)


barroij:
side question : is it a coding standard to not get the member with default values when there are to be created in the constructor anyway ?

For instance, in the `BoundingBox ` class, only `vectorsWorld` is given a default value, the others are supposed to be initialized in the constructor (as they are not Nullable)

barroij:
also it seems that `Vector3.Zero()` is prefered over `new Vector()`, is that a generic rule too ?

src/Culling/babylon.boundingBox.ts
```diff
@@ -1,4 +1,13 @@
 ﻿module BABYLON {
+
+    
+    /**
+     * @hidden
+     */
+    const tempVector3Diff = new Vector3();
+    const tempVector3Min = new Vector3();
```
deltakosh: I would recommend using Tmp.Vector3

deltakosh:
Vector3.zero or new vector3 is fine. I think that zero is just more explicit

barroij:
@deltakosh I added similar changes in BoundingSphere too

deltakosh:
I personally prefer public v = new Vector3(); because of readability 
But both will end up exactly the same in the generated js

barroij:
So do I, I made the changes accordingly.

Do you know if there is some helper function to create an instance of Array with a given size a type ?
something like 
````
    const buildArray = <T, P extends any[]>(size:number, konstructor: new (...params: P)=>T, ...params: P) => {
        const a = new Array<T>(size);
        for (let i = 0; i < size; ++i)
            a[i] = new konstructor(...params);
    }
````

Also, is there some generic tuple types that can be used to make sure that vectors is given an array of 8 elements ?

something like 
````
const NTuple = <T extends any[]>(...args: T): T => args

interface Tuple<T, N extends number> extends Array<T> {
    readonly length: N
}

public vectors: Tuple<Vector3, 8> = NTuple(Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero());
````

?

barroij:
yes of course tuple exists, but if want an tuple of 8 Vector3, you've to right `[Vector3, Vector3, Vector3, Vector3, Vector3, Vector3, Vector3, Vector3]` which is a bit cumbersome :).

Another question regarding the `BoundingBox` and a potential issue :

when `_update` is called, the matrix that is passed as a paramter is not deeply copied, we just copy the reference : 
````
        public _update(world: Matrix): void {

           /* some stuff here */

            this._worldMatrix = world;
        }
````
It might be dangerous to keep such a reference to an external object as this function can be also called from the BoundingInfo : 
````
        public update(world: Matrix) {
            if (this._isLocked) {
                return;
            }
            this.boundingBox._update(world);
            this.boundingSphere._update(world);
        }
````
Is it on prupose ?

deltakosh:
Yes. for performance reason as we do not want to clone 16 floats per update. The risk is pretty minimal as the worldmatrix is used immediately. It is stored so you can get it later for debugging purpose

PR #5248, opened 2018-09-25
Fix linesMesh instance picking

barroij:
Now the LinesMesh support instance, but the intersection code has not eben updated accordingly, leading to javascript exception when perfomring picking on instanced Lines

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -1900,9 +1900,9 @@
          * @returns the current updated Vector3  
          */
         public minimizeInPlaceFromFloats(x: number, y: number, z: number): Vector3 {
-            this.x = Math.min(this.x, x);
-            this.y = Math.min(this.y, y);
-            this.z = Math.min(this.z, z);
+            if (x < this.x) this.x = x;
+            if (y < this.y) this.y = y;
+            if (z < this.z) this.z = z;
             return this;
         }
 
@@ -1914,9 +1914,9 @@
          * @returns the current updated Vector3
          */
         public maximizeInPlaceFromFloats(x: number, y: number, z: number): Vector3 {
-            this.x = Math.max(this.x, x);
-            this.y = Math.max(this.y, y);
-            this.z = Math.max(this.z, z);
+            if (x > this.x) this.x = x;
+            if (y > this.y) this.y = y;
+            if (z > this.z) this.z = z;
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


src/Math/babylon.math.ts
```diff
@@ -1914,9 +1914,9 @@
          * @returns the current updated Vector3
          */
         public maximizeInPlaceFromFloats(x: number, y: number, z: number): Vector3 {
-            this.x = Math.max(this.x, x);
-            this.y = Math.max(this.y, y);
-            this.z = Math.max(this.z, z);
+            if (x > this.x) this.x = x;
+            if (y > this.y) this.y = y;
+            if (z > this.z) this.z = z;
```
barroij: changes in this files are not really related to the PR subject.

I replaced `Math.min` and `Math.max` with if condition as it used to be before my other PR https://github.com/BabylonJS/Babylon.js/pull/5217.

src/Mesh/babylon.subMesh.ts
```diff
@@ -339,54 +337,68 @@
...
+
+                    if (fastCheck || !intersectInfo || currentIntersectInfo.distance < intersectInfo.distance) {
+                        intersectInfo = currentIntersectInfo;
+                        intersectInfo.faceId = index / 3;
 
-                            if (fastCheck) {
-                                break;
-                            }
+                        if (fastCheck) {
+                            break;
```
barroij: I moved the code into a function, but changed nothing

src/Mesh/babylon.subMesh.ts
```diff
@@ -339,54 +337,68 @@
...
+                var length = ray.intersectionSegment(p0, p1, intersectionThreshold);
+                if (length < 0) {
+                    continue;
+                }
+
+                if (fastCheck || !intersectInfo || length < intersectInfo.distance) {
+                    intersectInfo = new IntersectionInfo(null, null, length);
+
+                    if (fastCheck) {
+                        break;
```
barroij: I moved the code into a function, but changed nothing

src/Mesh/babylon.subMesh.ts
```diff
@@ -339,54 +337,68 @@
             }
 
             // LineMesh first as it's also a Mesh...
-            if (LinesMesh && this._mesh instanceof LinesMesh) {
-                var lineMesh = <LinesMesh>this._mesh;
+            if (LinesMesh) {
+                const mesh = this._mesh instanceof InstancedMesh ? (<InstancedMesh>this._mesh).sourceMesh : this._mesh
```
barroij: here is the real fix

PR #5249, opened 2018-09-25
reduce bounding Box memory footprint

barroij:
this PR is a proposal for reducing the memory size of a BoundingBox by removing the two Array of 8 Vector3 from the members `vectors` and `vectorsWorld`.

For backward compatibility, those are still available as getter that return a new array at each call. Those vectors can also be computed on an already existing array.

As solidParticlSystem relied on them, the code has been updated accordingly, and the API of BoundingSphere and BoundingeBow slightly updated so that an optional worldMatric can be passed to the `reconstruc` method, which is the one that must be called in case of any change of a local value (center, radius, center, ...). If my understanding is correct, `_update` should be used only if the world matrix changed.

Note that this PR might be a bit risky, but I think it is really worth it. BoundingBox memory footprint would  drop from 3 array + 27 Vector3 to 1 array + 11 Vector3


src/Culling/babylon.boundingBox.ts
```diff
@@ -182,7 +206,7 @@
          * @returns true if there is an intersection
          */
         public isInFrustum(frustumPlanes: Plane[]): boolean {
-            return BoundingBox.IsInFrustum(this.vectorsWorld, frustumPlanes);
+            return BoundingBox.IsInFrustum(this.vectorsWorldToRef(Tmp.Vector3), frustumPlanes);
```
barroij: Impact on performance here and below should be negligible as we just add a few assignements.

src/Culling/babylon.boundingSphere.ts
```diff
@@ -44,8 +44,9 @@
          * Recreates the entire bounding sphere from scratch
          * @param min defines the new minimum vector (in local space)
          * @param max defines the new maximum vector (in local space) 
+         * @param worldMatrix defines the new world matrix
          */
-        public reConstruct(min: Vector3, max: Vector3) {
+        public reConstruct(min: Vector3, max: Vector3, worldMatrix?: Matrix) {
```
barroij: adding an optional worldMatrix in case we want to update both local and world bounds

src/Math/babylon.math.ts
```diff
@@ -7187,7 +7187,7 @@
         public static Color4: Color4[] = [new Color4(0, 0, 0, 0), new Color4(0, 0, 0, 0)];
         public static Vector2: Vector2[] = [Vector2.Zero(), Vector2.Zero(), Vector2.Zero()];  // 3 temp Vector2 at once should be enough
         public static Vector3: Vector3[] = [Vector3.Zero(), Vector3.Zero(), Vector3.Zero(),
-        Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero()];    // 9 temp Vector3 at once should be enough
+        Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero()]; // 11 temp Vector3 at once should be enough
```
barroij: when typescript is upgraded to version 3.0+, you should consider adding someting like :
````
const buildArray = <T, P extends any[]>(size:number, konstructor: new (...params: P)=>T, ...params: P):Array<T> => {
    const a = new Array<T>(size);
    for (let i = 0; i < size; ++i)
        a[i] = new konstructor(...params);
    return a;
}
````

so that instead of writing
````
public static Vector3: Vector3[] = [Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero()];
````
you could write
````
public static Vector3: Vector3[] = buildArray(11, Vector3);
````

File: src/Particles/babylon.solidParticleSystem.ts [hunk 4]
```diff
@@ -965,31 +963,45 @@
...
-                            bBox.vectors[b].z = this._particle.position.z + this._cam_axisX.z * this._rotated.x + this._cam_axisY.z * this._rotated.y + this._cam_axisZ.z * this._rotated.z;
+                            v.x = this._particle.position.x + this._cam_axisX.x * this._rotated.x + this._cam_axisY.x * this._rotated.y + this._cam_axisZ.x * this._rotated.z;
+                            v.y = this._particle.position.y + this._cam_axisX.y * this._rotated.x + this._cam_axisY.y * this._rotated.y + this._cam_axisZ.y * this._rotated.z;
+                            v.z = this._particle.position.z + this._cam_axisX.z * this._rotated.x + this._cam_axisY.z * this._rotated.y + this._cam_axisZ.z * this._rotated.z;
+                            min.minimizeInPlace(v);
+                            max.maximizeInPlace(v);
                         }
-                        bBox._update(this.mesh._worldMatrix);
+
+                        bBox.reConstruct(min, max, this.mesh._worldMatrix);
```
barroij: updating the code so that the local vectors of the BBox are computing into existing temporary Tmp.Vector3.
As the localBounds are changed, calling `bBox._update` is not enough as it would only update the world stuff, meaning that `minimum`, `maximum`, `extends`, ... becomes inconsistents.
That's why I changed the code to call `reConstruct` instead.

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -965,31 +963,45 @@
...
+                    const maxBbox = Tmp.Vector3[1].copyFromFloats(
+                        this._particle._modelBoundingInfo.maximum.x * this._particle.scaling.x,
+                        this._particle._modelBoundingInfo.maximum.y * this._particle.scaling.y,
+                        this._particle._modelBoundingInfo.maximum.z * this._particle.scaling.z);
+
+                    const center = Tmp.Vector3[2].copyFrom(maxBbox).addInPlace(minBbox).scaleInPlace(0.5);
+                    const halfDiag  = Tmp.Vector3[3].copyFrom(maxBbox).subtractInPlace(minBbox).scaleInPlace(0.5).scale(this._bSphereRadiusFactor);
+                    const sphereMinBbox = Tmp.Vector3[4].copyFrom(center).subtractInPlace(halfDiag);
+                    const sphereMaxBbox = Tmp.Vector3[5].copyFrom(center).addInPlace(halfDiag);
+                    bSphere.reConstruct(sphereMinBbox, sphereMaxBbox, this.mesh._worldMatrix);
```
barroij: same here

src/Culling/babylon.boundingBox.ts
deltakosh: This is something I would like to get an evaluation because we trade 16 vector3 vs 24 assignements per submesh per frame. (this will start to be a thing if you have 4000 meshes)

I do not want to favor memory against perf.

I would prefer to have a way to use one or the other depending on user needs.

src/Math/babylon.math.ts
deltakosh: yep!!

src/Culling/babylon.boundingBox.ts
```diff
@@ -64,52 +57,82 @@
             this.reConstruct(min, max);
         }
 
+        /**
+         * Gets the 8 vectors representing the bounding box in local space
+         */
+        public get vectors(): Vector3[] {
```
deltakosh: This will also be a HUGE perf hit for Solidparticlesystem.
I definitely think we should have a specialized version of the BoundingBox (like BoundingBoxLite) that will be used by Babylonjs based on a general flag

src/Particles/babylon.solidParticleSystem.ts
deltakosh: Same as before, I'm concerned by perf here

src/Culling/babylon.boundingBox.ts
barroij: actually this getter would ideally be declared deprecated and not be used by anything within the engine.

src/Culling/babylon.boundingBox.ts
barroij: Got your point, that's something that would need to be measured. before merging. We might  consider having a single array of 24 number instead of 8 Vector3 to speed up things a bit more for such functions that are called every frame.

src/Culling/babylon.boundingBox.ts
deltakosh: We can't deprecate public API as there may be a lot of users out there

barroij:
@deltakosh I'll close it for now and keep it for me as it would require true performance figures to figure out whether there is a real perf hit or not

barroij:
no problemo, c'est normal entre anciens montrougiens ;)

deltakosh:
The Monrouge Mafia!!
Pour la blague j'habite maintenant pres d'une ville nommee Redmond :)

PR #5257, opened 2018-09-26
glTFSerializer: Remove unnecessary blob conversion

kcoley:
Should fix the issue on Edge and IE for Serializing Texture transforms to glTF

### DIFF
--- a/serializers/src/glTF/2.0/babylon.glTFMaterialExporter.ts
+++ b/serializers/src/glTF/2.0/babylon.glTFMaterialExporter.ts
@@ -479,14 +479,14 @@ module BABYLON.GLTF2.Exporter {
             return new Promise<string>((resolve, reject) => {
                 let hostingScene: Scene;
 
-                let textureType = Engine.TEXTURETYPE_UNSIGNED_INT;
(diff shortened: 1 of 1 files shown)


serializers/src/glTF/2.0/babylon.glTFMaterialExporter.ts
```diff
@@ -501,20 +501,8 @@ module BABYLON.GLTF2.Exporter {
...
-                                    hostingScene.dispose();
-                                    resolve(base64String);
-                                };
-                                fileReader.readAsDataURL(blob);
-                            }
-                            else {
-                                reject("gltfMaterialExporter: Failed to get blob from image canvas!");
-                            }
-                        });
+                        const dataURL = canvas.toDataURL();
```
sebavan: You still need to do the toblob if available to not lose color when alpha value is low on compatible browsers.

sebavan: This could be your fallback if toblob is not available.

sebavan:
LGTM

PR #5258, opened 2018-09-26
SPS optimization

barroij:
As I was trying to counter balance the loss in performance due do the removal of `vectors` and `vectorsWorld` from the BoundingBox, I made a pass on the SPS as it is the only place where we use the `vectorsWorld` property of BoundingBox in BabylonJs.

I was thinking of proposing it as "all or nothing", but it might be better to split it and propose the SPS refactor separately.

to measure performance gain, I based my measurement on https://www.babylonjs-playground.com/#WCDZS#2 that I downladed locally and modified :. 
I modified the SPS creation:
````
          var SPS = new BABYLON.SolidParticleSystem('SPS', scene, {particleIntersection:true});
          SPS.addShape(tetra, 5000);
          SPS.addShape(box, 5000);
````

and the renderLoop 
````
		let i = 0
		const frameStart = 20*30
		const frameEnd = frameStart + 30*30
		const frameCount = frameEnd-frameStart;
		let t0;
        engine.runRenderLoop(function () {
            if (scene) {
				if (i === frameEnd) {
				  const t1 = performance.now()
				  console.log(`average frame time over ${frameCount} frame : ${(t1 - t0) / frameCount} milliseconds.`)
				}
				if (i === frameStart) {
				 t0 = performance.now()
				}
				
				++i;
                scene.render();
            }
        });
````

So that I get the average frame time over 900 frames.
I used the Javascript Profiler of Chrome and I compared this branch against the master branch at commit befbf120 (as I branched from it)

My results are suggesting a performance improvement of 2.5+ms per frame.

### CODE REVIEW

**Comment by jbousquie** [comment]
About Bbox (I haven't followed the recent changes), note that it's used also in the internal ribbon instance update process : https://github.com/BabylonJS/Babylon.js/blob/master/src/Mesh/babylon.meshBuilder.ts#L214 

@barroij Could you please send me a built version of babylon.min.js from your modified SPS code so I can make stress tests and perf comparisons on some old computer ?

File: src/Mesh/babylon.meshBuilder.ts [hunk 1]
```diff
@@ -162,65 +162,54 @@ module BABYLON {
...
+                                positions[i + 1] = pathPoint.y;
+                                positions[i + 2] = pathPoint.z;
                                 i += 3;
                             }
                         }
                     }
                 };
                 var positions = <FloatArray>instance.getVerticesData(VertexBuffer.PositionKind);
                 positionFunction(positions);
-                instance._boundingInfo = new BoundingInfo(Tmp.Vector3[2], Tmp.Vector3[3]);
```
barroij: there was probably a bug here, as we passed the wrong temporary vectors

barroij:
Hi @jbousquie, here is the minified version
[babylon.zip](https://github.com/BabylonJS/Babylon.js/files/2423480/babylon.zip)

jbousquie:
@barroij 
Thank you, I'll do some stress tests before validating.
Have you also tested the modified SPS perfs on its other features : particle intersections, particle depth sort, pickability ?
I know it sounds painful but the SPS is a big stuff and fixing the code back if the optimization  has problems then would be even more painful :-(

barroij:
The only test I made so far is the one mentioned in the description (in which I set `particleIntersection` to true, is that what you mean wit testing particle intersections?), I can easilly run the test with depth sort enabled.

What I did was actually low level optimization. I tried to reduced the number of property dereferencing and use local variable instead of property when applicable. As for the picking, it seems that what we just do in the SPS is populating a map to be able to retrieve the particleId from the faceId. The "real" picking work is still done at the mesh level and I did not change that.


Anyway, I ran the tests against in other web browsers (namely Firefox and Edge), and I got different results, the frame time becomes slower :/

jbousquie:
Ok,
On va continuer en français par échange de mails si tu veux bien pour éviter de polluer ce thread. Écris moi ici sitoplé : jerome@bousquie.fr    :-)

Sinon pour le fix sur l'update du ribbon avec le calcul de la Bbox intégré, tu peux tester avec le code de cette démo (du moins la partie concernant le ribbon et virer le SPS et les facetData) : http://jerome.bousquie.fr/BJS/test/facetdata3.html
Histoire de s'assurer que la Bbox est toujours correctement calculée sur un ribbon modifié dynamiquement

src/Math/babylon.math.ts
```diff
@@ -1616,7 +1614,7 @@ module BABYLON {
          * @returns a new Quaternion object, computed from the Vector3 coordinates
          */
         public toQuaternion(): Quaternion {
-            return BABYLON.Quaternion.RotationYawPitchRoll(this.x, this.y, this.z);
+            return BABYLON.Quaternion.RotationYawPitchRoll(this.y, this.x, this.z);
```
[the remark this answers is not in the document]
jbousquie: not sure to understand : in the SPS code in TS, `_quaternionRotationYPR()` is declared as `private` 
https://github.com/BabylonJS/Babylon.js/blob/master/src/Particles/babylon.solidParticleSystem.ts#L1050

Then, I think it's just the way JS works : everything can be called or accessed from anywhere when you know its reference.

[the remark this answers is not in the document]
barroij: my point was about the order of arguments. In the SPS, yaw is `y` and pitch is `x`, adn the other way around in `toQuaternion`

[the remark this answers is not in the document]
jbousquie: oops ok
as far I know : yaw is Y and pitch is X
https://en.wikipedia.org/wiki/Aircraft_principal_axes#/media/File:Yaw_Axis_Corrected.svg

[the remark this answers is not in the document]
barroij: phew, the docs says that too :) https://doc.babylonjs.com/resources/rotation_conventions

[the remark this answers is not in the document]
barroij: I created another PR for that https://github.com/BabylonJS/Babylon.js/pull/5289

jbousquie:
I set an online test here : http://jerome.bousquie.fr/BJS/test/SPSRefactor/
You choose the particle number then the SPS version : the legacy one or the refactored one.

How to test ?
Compare, in the same browser with the same window size, both SPS versions with the same particle number.
Please don't open it in two tabs simultaneously to be sure to keep only one WebGL context instanciated at the same moment. As well, don't run two browsers with this demo simultaneously.

Please wait for some seconds (20") after the start so as the GC has finished its cleanup before comparing. Then nothing more is allocated nor deleted.

The comparison is better (more pertinent) when the model is stressed, it is to say when the number of animated particles prevents from reaching the 60 fps : it's easier to check out the gain or the loss between both versions.

What to compare ? 
The framerate of course, but also its stability along the elapsed time.

In my first tests on Linux, so with FF and Chrome only, on a strong workstation (20K particle basis) : 
FF : the legacy version is a bit faster (+17 %) than the refactored one and more stable
Chrome : the refactored version is way faster (+57 %) than the legacy one and still a bit little unstable, but nothing important (-/+ 1 fps)

I need to test with my old laptop at home.
For now, it seems that the Chrome engine was greatly improved since the SPS was published because they transformed the sps code factorization from a perf loss into a perf gain.. what seems to be a very good news imho.

Remember it's a stress test, so probably not what a real user will do with a SPS. So a regression of 17% in one browser vs a gain of 57% in another one on an extreme case ... I would think it's worth modifying the code on this lone measure.

Let's go on testing, but it already smells like this PR would get my approval :-D

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -337,50 +300,34 @@ module BABYLON {
...
-                    this._pitch = this._particle.rotation.x;
-                    this._roll = this._particle.rotation.z;
-                    this._quaternionRotationYPR();
-                }
-                this._quaternionToRotationMatrix();
-                this._rotMatrix.invertToRef(this._invertMatrix);
+                const particle = this.particles[p];
+                const shape = particle._model._shape;
+                particle.getRotationMatrix(rotMatrix);
+                rotMatrix.invertToRef(invertedRotMatrix);
```
barroij: @jbousquie Here it might be faster to retrieve the matrix of the conjugate quaternion than inverting a matrix (the result should be the same I think)

jbousquie: well, I don't know anything about conjugate quaternions. I just read how they are computed, what seems to be fast indeed. Then, I have no idea what this conjugate means and how to use it in order to "un-rotate" a given vector. My 3D maths aren't enough here but I'm still willing to learn. Do you have a link about something explaining how to use the conjugate quaternion ?

jbousquie:
About SPS tests
My tests with a 5 yo i5 laptop show the same results between Chrome, Chromium and Firefox on Linux : the legacy version is a bit faster than the refactored one with FF but far more slower with Chrome/Chromium.
The gain/loss are the same magnitude than the ones I got with my strong workstation.
I'll test tonight at home with a 10 yo laptop.

We need more feedbacks with other platforms (Edge, Safari, smartphones). I'll ask the forum community : http://www.html5gamedevs.com/topic/40380-sps-optimization-feedback-wanted/

deltakosh:
Lol ! I'll merge asap

deltakosh:
But before merging we need an update to the what's new:)

dist/preview release/what's new.md
```diff
@@ -11,6 +11,7 @@
 
 ### Core Engine
 
+- Refactor of the SolidParticleSystem code. Significant performance on Chrome when the numnber of particle becomes important (20k+) ([barroij](https://github.com/barroij))
```
barroij: @deltakosh What's new update ⬆️

deltakosh: Just a tiny typo

dist/preview release/what's new.md
```diff
@@ -11,6 +11,7 @@
 
 ### Core Engine
 
+- Refactor of the SolidParticleSystem code for performance and code quality improvement ([barroij](https://github.com/barroij))
```
barroij: @deltakosh better ?

deltakosh: Fantastic :)
Merging right now! thanks a lot

PR #5264, opened 2018-09-26
do not attach the same mesh twice in gizmo manager

TrevorDev:


deltakosh:
Would you mind fixing the conflict? (sorry :))

PR #5272, opened 2018-09-27
Enable optional scene geometry map for performance

barroij:
When adding a lot of geometries in the scene, the time spend in the `getGeometryByID` function becomes significant.

This PR adds an optional object in the scene than can be used to speed up the geometry lookup by avoiding an O(n) search for geometry each one is added.

The option can be activated by providing options to the Scene constructor


src/babylon.scene.ts
```diff
@@ -1846,7 +1863,7 @@ module BABYLON {
                 if (!checkPicking && ActionManager && ActionManager.HasPickTriggers) {
                     act = this._initActionManager(act, clickInfo);
                     if (act) {
-                        checkPicking = act.hasPickTriggers;
+                        checkPicking = act.hasPickTriggers;
```
barroij: those changes are probably Line ending change. I guess that wrong line ending have been added by the linter. (or is my config wrong ?)

deltakosh: Yes but no worry (you can now always run gulp --tsLintFix)

PR #5274, opened 2018-09-28
Add an ArrayBuilder utility function

barroij:
Add a small utility function to create Array of newly created objects, so that instead of writing
````
_tmpVecs: Vector3[] = [Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero(), Vector3.Zero()];
````

we can write
````
_tmpVecs: Vector3[] = Tools.BuildArray(10, Vector3);
````

I had to modify the `config.json` to make sure that the function is defined when static members initilization are done.

I have also a typedoc error that I cannot fix : 
````
Unrecognized tag param,param,param,returns at BuildArray (id: 34320) in Tools (id: 34212) babylon.d.ts:38966
1 error has been detected during the validation !
````
Help would be welcomed for that :)

### CODE REVIEW

File: src/Bones/babylon.bone.ts [hunk 1]
```diff
@@ -5,9 +5,9 @@ module BABYLON {
      */
     export class Bone extends Node {
 
-        private static _tmpVecs: Vector3[] = [Vector3.Zero(), Vector3.Zero()];
+        private static _tmpVecs: Vector3[] = Tools.BuildArray(2, Vector3);
         private static _tmpQuat = Quaternion.Identity();
-        private static _tmpMats: Matrix[] = [Matrix.Identity(), Matrix.Identity(), Matrix.Identity(), Matrix.Identity(), Matrix.Identity()];
+        private static _tmpMats: Matrix[] = Tools.BuildArray(5, Matrix);
```
deltakosh: Well the matrix won't be identity here

src/Math/babylon.math.ts
```diff
@@ -2034,10 +2045,7 @@ module BABYLON {
          * @returns the current updated Vector3
          */
         public copyFrom(source: Vector3): Vector3 {
-            this.x = source.x;
-            this.y = source.y;
-            this.z = source.z;
-            return this;
+            return this.copyFromFloats(source.x, source.y, source.z);
```
deltakosh: not sure this is faster

src/Tools/babylon.tools.ts
```diff
@@ -518,12 +518,8 @@ module BABYLON {
             }
 
             if (bias) {
-                minimum.x -= minimum.x * bias.x + bias.y;
-                minimum.y -= minimum.y * bias.x + bias.y;
-                minimum.z -= minimum.z * bias.x + bias.y;
-                maximum.x += maximum.x * bias.x + bias.y;
-                maximum.y += maximum.y * bias.x + bias.y;
-                maximum.z += maximum.z * bias.x + bias.y;
+                minimum.scaleInPlace(1 - bias.x).addInPlaceFromFloats(-bias.y, -bias.y, -bias.y);
```
deltakosh: same here... I think this is slower

src/Bones/babylon.bone.ts
barroij: I can change that, but we don't really care as any code that use it should not assume anything about its content.

src/Math/babylon.math.ts
barroij: Neither am I. Actually from my little javascript experience, such low level performance are very hard to predict as there are very engine dependant. On one hand, one can think that it would be faster as we avoid one function call. On the other hand, the Jitter is likely to inline the function call anyway and always going through the same function would make it considered "hot" by the engine optimizer so it is more likely to be highly optimized...

So I tend to share as much code as possible, as if it it does not really affect performance, it reduces the size of the code and avoiding code duplication tends to to makes thing easier to understand, test and maintain.

But I can revert it if you prefer, it is more "your" code base than mine, it is up to you :)

src/Bones/babylon.bone.ts
deltakosh: True:) we don't care actually

deltakosh:
Can you update the what's new?

barroij:
@deltakosh I updated the BuildArray prototype so that we can call it like this : 
````Tools.BuildArray(9, Vector3.Zero);````

The what's new is all about 3.3.0, where shall I put my stuff ?

deltakosh:
I reseted it yesterday

barroij:
@deltakosh what's new updated !

PR #5279, opened 2018-09-29
[Just for review] Object based motion blur post-process

julien-moreau:
Adding object based motion blur post-process and velocity map in geometry buffer renderer

### DIFF
--- a/src/Rendering/babylon.geometryBufferRenderer.ts
+++ b/src/Rendering/babylon.geometryBufferRenderer.ts
@@ -3,10 +3,27 @@ module BABYLON {
      * This renderer is helpfull to fill one of the render target with a geometry buffer.
      */
     export class GeometryBufferRenderer {
+        /**
+         * Constant used to retrieve the position texture index in the G-Buffer textures array
+         * using getIndex(GeometryBufferRenderer.POSITION_TEXTURE_INDEX)
+         */
+        public static readonly POSITION_TEXTURE_TYPE = 1;
+        /**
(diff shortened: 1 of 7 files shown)


src/Rendering/babylon.geometryBufferRenderer.ts
```diff
@@ -3,10 +3,22 @@ module BABYLON {
      * This renderer is helpfull to fill one of the render target with a geometry buffer.
      */
     export class GeometryBufferRenderer {
+        /**
+         * Represents the index of the position texture in the G-Buffer textures array
+         */
+        public positionIndex: number = 2;
```
julien-moreau: @deltakosh @CraigFeldspar 
Is that ok for you to expose the textures indices like that? Depth and normal textures will always be at indices 0 and 1 so I didn't expose these indices here

src/Rendering/babylon.geometryBufferRenderer.ts
```diff
@@ -3,10 +3,22 @@ module BABYLON {
...
+         */
+        public velocityIndex: number = 3;
+
         private _scene: Scene;
         private _multiRenderTarget: MultiRenderTarget;
         private _ratio: number;
         private _enablePosition: boolean = false;
+        private _enableVelocity: boolean = false;
+
+        private _previousWorldMatrices: { [index: string]: Matrix } = { };
```
julien-moreau: @deltakosh 
possibilities:
- Store a _previousWorldMatrix: Matrix on mesh class ?
- Keep this dictionary here ?

PR #5281, opened 2018-10-01
[Draft] use Pool and decorators to make the use temprary Vectors/Matrix less error-prone

barroij:
[Disclaimer] PR not meant to merged as is, it is a draft

I had some hard time debugging something because I was using already allocated vectors from `Tmp.Vector`, and it can be very easy to mess up things when we use them in a function and call another that use them as well. To make sure there is no issue with them one has to check each function implementation, which  is a bit error prone and can be a bit cumbersome :)

This PR contains an overview of what we could use to get rid of that using a Pool from which we can get and release items, and possibly with a decorator to automatically release the used items on function leave.

### DIFF
--- a/src/Culling/babylon.boundingBox.ts
+++ b/src/Culling/babylon.boundingBox.ts
@@ -117,14 +117,15 @@ module BABYLON {
          * @param factor defines the scale factor to apply
          * @returns the current bounding box
          */
+        @automaticPoolRelease(Tmp.PoolVector3)
         public scale(factor: number): BoundingBox {
-            const diff = Tmp.Vector3[0].copyFrom(this.maximum).subtractInPlace(this.minimum);
+            const diff = Tmp.PoolVector3.getTemp().copyFrom(this.maximum).subtractInPlace(this.minimum);
             let distance = diff.length() * factor;
             diff.normalize();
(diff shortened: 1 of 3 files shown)


src/Culling/babylon.boundingBox.ts
```diff
@@ -117,14 +117,15 @@ module BABYLON {
          * @param factor defines the scale factor to apply
          * @returns the current bounding box
          */
+        @automaticPoolRelease(Tmp.PoolVector3)
         public scale(factor: number): BoundingBox {
-            const diff = Tmp.Vector3[0].copyFrom(this.maximum).subtractInPlace(this.minimum);
+            const diff = Tmp.PoolVector3.getTemp().copyFrom(this.maximum).subtractInPlace(this.minimum);
```
deltakosh: Well the main problem is perf here. The math functions can be called thousands of times per frame. We do not want to add more code here.

I agree that TMP object can lead to errors and this is why we have MathTmp for instance for math functions (easier to track)

I would prefer having specific TMP objects per features (a static pool for bounding info would be fine) instead of adding more code processing to central functions

barroij: yes, it would be simpler indeed to make sure each class deals with its own pool. And better for performance

PR #5291, opened 2018-10-03
Share lines mesh shader material

barroij:
When using LineMesh, each instance have its own Shader material. Still, giving the options that can be passed to LinesMesh constructor and how ShaderMatrial are created from those options, there can be only 4 possible different ShaderMaterial for a given scene.

This PR propose to share those instances of ShaderMaterial among all the LineMesh of a scene.


barroij:
actually it does not work if lines have different colors... 🤦‍♂️

PR #5300, opened 2018-10-04
use dedicated TmpVector3 For BBox and BSphere + Prevent some Matrix inversion

barroij:
This PR contains several things :
- Instead of using `Tmp.Vector3`, `BoundingBox` and `BoundingSphere` use their own temporary Vector3 Array. This to reduce the chance of bug, as their methods are likely to be called from internal code that are already using the `Tmp.Vector3` array.

- Prevent from inverting a matrix when we know that the matrix contains only a translation or a rotation, as the inverse of a translation is the opposite translation and the inverse of a quaternion is its conjugate.

- some code cleaning in TransformNode


src/Mesh/babylon.transformNode.ts
```diff
@@ -727,19 +704,32 @@ module BABYLON {
...
+
+            translationMatrixInv.multiplyToRef(rotationMatrix, finalMatrix); // T' x R
+            finalMatrix.multiplyToRef(translationMatrix, finalMatrix);  // T' x R x T
+
+            finalMatrix.decompose(finalScale, finalRotation, finalTranslation);
 
-            this.position.addInPlace(Tmp.Vector3[1]);
-            Tmp.Quaternion[0].multiplyToRef(this.rotationQuaternion, this.rotationQuaternion);
+            this.position.addInPlace(finalTranslation);
+            finalRotation.multiplyToRef(this.rotationQuaternion, this.rotationQuaternion);
```
barroij: I use local variables and proper variable name for better understanding

src/Particles/babylon.solidParticleSystem.ts
```diff
@@ -301,13 +301,23 @@ module BABYLON {
...
+                // is equivalent to computing the matrix of the inverse quaternion, i.e of the conjugate quaternion
+                if (particle.rotationQuaternion) {
+                    particle.rotationQuaternion.conjugateToRef(quaternion);
+                }
+                else {
+                    const rotation = particle.rotation;
+                    Quaternion.RotationYawPitchRollToRef(rotation.y, rotation.x, rotation.z, quaternion);
+                    quaternion.conjugateInPlace();
+                }
+                quaternion.toRotationMatrix(invertedRotMatrix);
```
barroij: no need to call Matrix.invert to get the inverse matrix of the rotation, we can compute the matrix of the conjugate quaternion instead.

jbousquie: This one is so smart !!! Love it :-)
_(let's check in the code  that the quaternion to be conjugated is really normalized)_

I think this could be a dedicated tool in the math lib : `invertedRotationMatrixFromQuaternion()` or something like that

src/Mesh/babylon.transformNode.ts
deltakosh: yeah good idea definitely

src/Particles/babylon.solidParticleSystem.ts
deltakosh: I'll run some tests

barroij:
@deltakosh does each PR imply a what's new update ?

barroij:
Or is it because of the failing test ? I checked the log, and did not figure what is is about... will check to run it locally tomorow

deltakosh:
Nope every PR must be referenced (it is ok to change an exisiting line)

deltakosh:
There is a conflict sir :)

PR #5301, opened 2018-10-04
Speed-up addition and removal of Transfrom node from scenes

barroij:
Removing a lot of Transform Nodes from a scene can take seconds if the number of nodes is big. This PR propose to keep the Idx of the TransformNode within the `transformNodes` array of the scene.

With this PR, in my use case, removing 20k+ nodes from scenes takes roughly 2s ecs instead of 12 seconds previously


src/babylon.scene.ts
```diff
@@ -3011,10 +3012,16 @@ module BABYLON {
          * @returns the index where the transform node was in the transform node list
          */
         public removeTransformNode(toRemove: TransformNode): number {
-            var index = this.transformNodes.indexOf(toRemove);
+            var index = toRemove._indexInSceneTransformNodesArray;
             if (index !== -1) {
-                // Remove from the scene if found
-                this.transformNodes.splice(index, 1);
+                if (index !== this.transformNodes.length - 1) {
+                    const lastNode = this.transformNodes[this.transformNodes.length - 1];
```
deltakosh: You are assuming that no one touch the array directly. I'm not opposed to this idea but we need to add a warning on the transformNodes proeprty then

src/babylon.abstractScene.ts
```diff
@@ -155,6 +155,9 @@ module BABYLON {
 
         /**
         * All of the tranform nodes added to this scene
+        * In the context a the Scene, it is not supposed to be modified manually.
+        * Any addition or removal should be done using the addTransformNode and removeTransformNode Scene methods.
+        * Note also that the order of the TransformNode wihin the array is not significant and might change.
```
barroij: I added a warning in the comment. I'll update the waht's new after https://github.com/BabylonJS/Babylon.js/pull/5300 is merged

dist/preview release/what's new.md
```diff
@@ -16,7 +16,7 @@
 - Added utility function `Tools.BuildArray` for array initialisation ([barroij](https://github.com/barroij))
 - Introduced a new `IOfflineSupport` interface to hide IndexedDB ([Deltakosh](https://github.com/deltakosh))
 - `PBRMaterial` and `StandardMaterial` now use hot swapping feature for shaders. This means they can keep using a previous shader while a new one is being compiled ([Deltakosh](https://github.com/deltakosh))
-- Performance oriented change (prevent avoidable matrix inversion or square root computation) ([barroij](https://github.com/barroij))
+- Performance oriented changes (prevent avoidable matrix inversion or square root computation, enable a removal in O(1) from the `transformNodes`array of the Scene) ([barroij](https://github.com/barroij))
```
deltakosh: AS I feel this list will grow I would suggest a more bullet list oriented approach

Issue #3899, opened 2018-03-07
WebXR


deltakosh:
- [x] Explore and understand how WerbXR works
- [x] Write a doc about required dev environment
- [x] Propose an API
- [x] Implement the API
  - [x] XRSessionManager
  - [x] XRCamera
  - [x] Engine/Camera modifications
  - [x] Controllers
  - [x] DefaultXRExperience
  - [x] supportsSession
  - [x] Keep depth in sync with camera https://forum.babylonjs.com/t/webxrcamera-minz-maxz-to-set-xrsession-depthnear-far/640
- [x] Documentation - https://doc.babylonjs.com/how_to/webxr
  - [x] clerify customRenderTargets vs outputRenderTarget on camera
- [x] Samples
  - [x] https://playground.babylonjs.com/#AM07G2#4
  - [x] https://playground.babylonjs.com/#PPM311

veikkoeeva:
Further resources, discussion groups etc. from Mozilla blog: https://blog.mozilla.org/blog/2017/10/20/bringing-mixed-reality-web/. Further refined goals at a later post https://blog.mozvr.com/enabling-the-social-3d-web/ .

It appears to me this is about unifying the augmented reality experience. Concretely in near term, codewise, to me this looks like a proposal to add avatars and coordinate their actions via WebRTC (notably, by the way, MS also opened one peer server codebase).

veikkoeeva:
To continue a bit, this also looks alluring for an outsider: https://blog.mozvr.com/experimenting-with-ar-and-the-web-on-ios/.

It might be useful to see how could WASM be utilized for more compute intensive tasks now that it's coming hard to the fore, so to speak. For instance, there's Blazor template already at dotnet new -i Microsoft.AspNetCore.Blazor.Templates::0.1.0-preview1-10071 --nuget-source https://dotnet.myget.org/F/blazor-dev/api/v3/index.json (`dotnet new -i Microsoft.AspNetCore.Blazor.Templates::0.1.0-preview1-10071 --nuget-source https://dotnet.myget.org/F/blazor-dev/api/v3/index.json`) for those in .NET Standard 2.0, adding SignalR, VSIX etc. in the next few weeks apparently. This is obviously .NET centric view, though likely other ecosystems have something similar.

veikkoeeva:
Adding https://blog.mozvr.com/responsive-webxr-a-painter-xr/ too. Sorry for spamming. :)

veikkoeeva:
[...]

PR #5305, opened 2018-10-04
customAnimationFrameRequester

TrevorDev:

Usage: http://localhost:1338/Playground/index-local.html#M5J1I2#5


deltakosh:
Some checks are missing

PR #5307, opened 2018-10-05
Speed-up the disposing of large number of meshes

barroij:
This PR contains several things, some optional, to reduce the time of disposing a large number of mesh hierarchies. It makes a few assumptions that need to be confirmed by someone who has the knowledge :)

- Added `disposingLineMeshColorShader` to not go through irrelevant code when the material that is beeing disposed is a `ColorShaderMaterial` of a LinesMesh, as it is never shared nor bound to any mesh.

- Stop calling `splice` on the `scene.meshes` array. it assumes that the order in this array does not matter.

- Stop calling `splice` on the `engine._uniformBuffers` array. it assumes that the order in this array does not matter.

- Added a `_indexInSceneMaterialArray ` member to the `Material` class so that each material stores its index with the `scene.materials` array. The to allow fast O(1) lookup and removal time. It assumes that the order of the material within the array does not matter and that this array is not modified from outside of the engine.

- Added a `_indexInSourceMeshInstanceArray` member to the `InstancedMesh` class so that each instance stores its index with the `mesh.instances` array. The to allow fast O(1) lookup and removal time. It assumes that the order of the instances within the array does not matter and that this array is not modified from outside of the engine.

- Added an option `useMaterialMeshMap`in the scene constructor options. Disabled by default. When set to true, Material will have and keep up-to-date a map of bound meshes. This is to avoid browsing all the meshes of the scene to retrieve the ones bound to the current material.

- Added an option `useClonedMeshhMap `in the scene constructor options Disabled by default. When set to true, each Mesh will have and keep up-to-date a map of cloned meshes. This is to avoid browsing all the meshes of the scene to retrieve the ones that have the current mesh as source mesh.

- Added `blockfreeActiveMeshesAndRenderingGroups` getter and setter in scene, as we are currently doing with `blockMaterialDirtyMechanism`.


src/Materials/babylon.material.ts
```diff
@@ -1261,36 +1276,31 @@ module BABYLON {
...
-                                geometry._releaseVertexArrayObject(subMesh._materialEffect);
-                                if (forceDisposeEffect && subMesh._materialEffect) {
-                                    this._scene.getEngine()._releaseEffect(subMesh._materialEffect);
-                                }
-                            }
-                        } else {
-                            geometry._releaseVertexArrayObject(this._effect);
+            scene.removeMaterial(this);
+
+            if (!this.disposingLineMeshColorShader) {
```
deltakosh: This is too much coupled for me. I would prefer adding a new parameter to the function instead

barroij: Did not do that because I did not want to change the API and keep it as an implementation detail, but I can change it :)

barroij:
@deltakosh updated

src/Materials/babylon.material.ts
```diff
@@ -1259,38 +1271,34 @@ module BABYLON {
          * Disposes the material
          * @param forceDisposeEffect specifies if effects should be forcefully disposed
          * @param forceDisposeTextures specifies if textures should be forcefully disposed
+         * @param notBoundToMesh specifies if the material that is being disposed is known to be not bound to any mesh
```
barroij: @deltakosh better ?

src/Materials/babylon.material.ts
```diff
@@ -788,8 +794,11 @@ module BABYLON {
             this._useUBO = this.getScene().getEngine().supportsUniformBuffers;
 
             if (!doNotAdd) {
-                this._scene.materials.push(this);
-                this._scene.onNewMaterialAddedObservable.notifyObservers(this);
+                this._scene.addMaterial(this);
+            }
+
+            if (scene.useMaterialMeshMap) {
```
deltakosh: I fixed this one as scene should be this._scene

barroij: yes you're right, good catch !

barroij:
@deltakosh doc update here :https://github.com/BabylonJS/Documentation/pull/1283

PR #5311, opened 2018-10-05
Xr camera

TrevorDev:
#3899

Usage: http://localhost:1338/Playground/index-local.html#ATL1CS#9


src/Cameras/XR/babylon.webXRCamera.ts
```diff
@@ -0,0 +1,49 @@
+module BABYLON {
+    /**
+     * WebXR Camera which holds the views for the xrSession
+     * @see https://immersive-web.github.io/webxr/
```
deltakosh: we will probably link it to our doc and mention the spec there

PR #5313, opened 2018-10-05
Rename MotionBlurProcess to MotionBlurPostProcess

julien-moreau:
Will update the playground to


julien-moreau:
Just the what's new test failed, but I think it is not really needed for this kind of change ^^

PR #5323, opened 2018-10-06
Simplified pointerUp babylon.scene.ts

aWeirdo:
Simplified onPointerObservable code for pointerUp
It was just bugging me. 😛

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -1785,14 +1785,12 @@ module BABYLON {
 
             let type = PointerEventTypes.POINTERUP;
             if (this.onPointerObservable.hasObservers()) {
-                if (!clickInfo.ignore) {
-                    if (!clickInfo.hasSwiped) {
-                        if (clickInfo.singleClick && this.onPointerObservable.hasSpecificMask(PointerEventTypes.POINTERTAP)) {
-                            type = PointerEventTypes.POINTERTAP;
-                        }
-                        if (clickInfo.doubleClick && this.onPointerObservable.hasSpecificMask(PointerEventTypes.POINTERDOUBLETAP)) {
-                            type = PointerEventTypes.POINTERDOUBLETAP;
-                        }
+                if (!clickInfo.ignore && !clickInfo.hasSwiped) {
+                    if (clickInfo.singleClick && this.onPointerObservable.hasSpecificMask(PointerEventTypes.POINTERTAP)) {
+                        type = PointerEventTypes.POINTERTAP;
+                    }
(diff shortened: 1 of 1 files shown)


src/babylon.scene.ts
```diff
@@ -1788,24 +1788,16 @@ module BABYLON {
...
-                            this._setRayOnPointerInfo(pi);
-                            this.onPointerObservable.notifyObservers(pi, type);
+                            type = PointerEventTypes.POINTERTAP;
                         }
                         if (clickInfo.doubleClick && this.onPointerObservable.hasSpecificMask(PointerEventTypes.POINTERDOUBLETAP)) {
-                            let type = PointerEventTypes.POINTERDOUBLETAP;
-                            let pi = new PointerInfo(type, evt, pickResult);
-                            this._setRayOnPointerInfo(pi);
-                            this.onPointerObservable.notifyObservers(pi, type);
+                            type = PointerEventTypes.POINTERDOUBLETAP;
```
deltakosh: that does not work :)
There is no "else" here between the two blocks

deltakosh: That could be a function though to help reduce code complexity

aWeirdo: Hmm, Should it be able to fire multiple observers per run? 
I assumed it would either run a singleClick, doubleClick or standard pointerUp event type
http://playground.babylonjs.com/#UM43F8

deltakosh: That's a good point. Then we should add a else to make sure and do heavy testing

src/babylon.scene.ts
```diff
@@ -1785,14 +1785,12 @@ module BABYLON {
...
-                            type = PointerEventTypes.POINTERDOUBLETAP;
-                        }
+                if (!clickInfo.ignore && !clickInfo.hasSwiped) {
+                    if (clickInfo.singleClick && this.onPointerObservable.hasSpecificMask(PointerEventTypes.POINTERTAP)) {
+                        type = PointerEventTypes.POINTERTAP;
+                    }
+                    else if (clickInfo.doubleClick && this.onPointerObservable.hasSpecificMask(PointerEventTypes.POINTERDOUBLETAP)) {
+                        type = PointerEventTypes.POINTERDOUBLETAP;
                     }
                 }
```
aWeirdo: Like this you mean?

Sorry for the last minute restructure aswell..
just notised i forgot the bundled if's 

    if( x ){
      if( y ) {
        // something
      }
    }

    // instead of
    if( x && y ){
      // something
    }

deltakosh: This looks good to me. I will do.orr testing and report back here

PR #5326, opened 2018-10-08
Handle properly the LinesMesh  intersectionThreshold

barroij:
This PR is an attempt to handle properly the  `intersectionThreshold` of `LinesMesh``. This "extra-width" cannot be added to the BoundingInfo using the `boundingBias` as it is currently, because the threshold is then scaled according to the world matrix compute the world BoundingInfo. This can lead to missleading boundingBox (if the scale is 1000 and the threshold is 0.1, the wolrd bbox would get an extra extend of 100 in each direction).

To fix this issue, I removedthe code that put this `intersectionThreshold` as boundingBias into the `BoundingInfo`. Instead, I integrate this `intersectionThreshold` in the `Ray.intersectsBoxtest` and `Ray.intersectsSphere` with an extra parameter.


src/Culling/babylon.boundingBox.ts
```diff
@@ -1,4 +1,7 @@
-﻿module BABYLON {
+module BABYLON {
+
+    const _identityMatrix = Matrix.Identity();
```
barroij: I added that as there was in the BoundingSphere.
it allows an optimization when the constructor or the `reConstruct` method is called wihtout no input worldMatrix as it avoid transforming the 8 `vectors` to compute `vectorsWorld `

src/Culling/babylon.boundingBox.ts
```diff
@@ -61,20 +69,21 @@
...
-            this.reConstruct(min, max, worldMatrix);
+        constructor(min: Vector3, max: Vector3, worldMatrix?: Matrix, extraWorldExtent?: number) {
+            this.reConstruct(min, max, worldMatrix, extraWorldExtent);
         }
 
         // Methods
 
         /**
-         * Recreates the entire bounding box from scratch
+         * Recreates the entire bounding box from scratch, producing same values as if the constructor was called.
```
barroij: made it clearer :)

src/Culling/babylon.boundingBox.ts
```diff
@@ -93,7 +102,7 @@
             max.addToRef(min, this.center).scaleInPlace(0.5);
             max.subtractToRef(min, this.extendSize).scaleInPlace(0.5);
 
-            this._update(worldMatrix || this._worldMatrix || Matrix.Identity());
+            this._update(worldMatrix || _identityMatrix, extraWorldExtent || 0);
```
barroij: do not pass  `this._worldMatrix` anymore as we want this function to behave as the constructor.

src/Culling/babylon.boundingInfo.ts
```diff
@@ -167,9 +178,9 @@ module BABYLON {
 		 * Gets the world distance between the min and max points of the bounding box
 		 */
         public get diagonalLength(): number {
-            let boundingBox = this.boundingBox;
-            let size = boundingBox.maximumWorld.subtract(boundingBox.minimumWorld);
-            return size.length();
+            const boundingBox = this.boundingBox;
+            const diag = boundingBox.maximumWorld.subtractToRef(boundingBox.minimumWorld, Tmp.Vector3[0]);
```
barroij: prevent new `Vector3`

src/Culling/babylon.boundingBox.ts
```diff
@@ -131,37 +141,50 @@
          * @returns current bounding box
          */
         public setWorldMatrix(matrix: Matrix): BoundingBox {
-            this._worldMatrix.copyFrom(matrix);
+            this._worldMatrix = matrix;
```
barroij: Do not modified in place the `_worldMatrix`, for 2 reasons :
- this World matrix can be the one provided by the geometry, which should not be modified from here
- we don't want to modify `_identityMatrix`

src/Culling/babylon.boundingBox.ts
```diff
@@ -112,13 +121,14 @@
             const min = this.center.subtractToRef(newRadius, tmpVectors[1]);
             const max = this.center.addToRef(newRadius, tmpVectors[2]);
 
-            this.reConstruct(min, max);
+            this.reConstruct(min, max, this._worldMatrix, this.extraWorldExtent);
```
barroij: scale should not clear `_worldMatrix` or `extraWorldExtent`

src/Culling/babylon.boundingSphere.ts
```diff
@@ -73,18 +82,31 @@ module BABYLON {
             const min = this.center.subtractToRef(tempRadiusVector, tmpVectors[1]);
             const max = this.center.addToRef(tempRadiusVector, tmpVectors[2]);
 
-            this.reConstruct(min, max);
+            this.reConstruct(min, max, this._worldMatrix, this._extraWorldExtent);
```
barroij: scale should not clear `_worldMatrix` or `extraWorldExtent`. That's why we should keep them as members.

src/Mesh/babylon.geometry.ts
```diff
@@ -62,9 +63,6 @@ module BABYLON {
          */
         public set boundingBias(value: Vector2) {
             if (this._boundingBias) {
-                if (this._boundingBias.equals(value)) {
-                    return;
-                }
                 this._boundingBias.copyFrom(value);
```
barroij: It is probably faster to always setting the value.

barroij:
Closing this one for now As it does not make picking working... Pick is done with a ray in local coordinates, so adding the treshold to the world coordinates does not work

PR #5327, opened 2018-10-08
Handle properly the LinesMesh intersectionThreshold 

barroij:
This PR is an attempt to handle properly the `intersectionThreshold` of `LinesMesh`. This "extra-width" cannot be added to the `BoundingInfo` using the boundingBias` as it is currently, because the threshold is then scaled according to the world matrix compute the world BoundingInfo. This can lead to missleading world `BoundingBox` (if the scale is 1000 and the threshold is 0.1, the world bbox would get an extra extend of 100 in each direction).

To fix this issue, I removed the code that put this intersectionThreshold as boundingBias into the BoundingInfo. Instead, I integrate this `intersectionThreshold` in the `Ray.intersectsBox` and `Ray.intersectsSphere` with an extra parameter.


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -1335,13 +1336,11 @@ module BABYLON {
...
-                var world = this.getWorldMatrix();
-                var worldOrigin = Vector3.TransformCoordinates(ray.origin, world);
-                var direction = ray.direction.clone();
-                direction = direction.scale(intersectInfo.distance);
-                var worldDirection = Vector3.TransformNormal(direction, world);
-
-                var pickedPoint = worldOrigin.add(worldDirection);
+                const world = this.getWorldMatrix();
+                const worldOrigin = Vector3.TransformCoordinates(ray.origin, world);
+                const direction = ray.direction.scaleToRef(intersectInfo.distance, Tmp.Vector3[0]);
```
barroij: prevent new Vector3

src/Mesh/babylon.geometry.ts
```diff
@@ -120,11 +117,6 @@ module BABYLON {
 
             // applyToMesh
             if (mesh) {
-                if (mesh.getClassName() === "LinesMesh") {
-                    this.boundingBias = new Vector2(0, (<LinesMesh>mesh).intersectionThreshold);
-                    this._updateExtend();
-                }
-
```
barroij: stop changing the bias according to the intersectionThreshold

src/Culling/babylon.boundingBox.ts
```diff
@@ -1,4 +1,7 @@
-﻿module BABYLON {
+module BABYLON {
+
+    const _identityMatrix = Matrix.Identity();
```
deltakosh: is const comparable to static readonly?

src/Culling/babylon.boundingBox.ts
```diff
@@ -93,7 +96,7 @@
             max.addToRef(min, this.center).scaleInPlace(0.5);
             max.subtractToRef(min, this.extendSize).scaleInPlace(0.5);
 
-            this._update(worldMatrix || this._worldMatrix || Matrix.Identity());
+            this._update(worldMatrix || _identityMatrix);
```
deltakosh: What if I call it from the outside with reConstruct(min, max)? It will use identity instead of the current wm, right? This seems like an unwanted side effect

src/Culling/babylon.boundingBox.ts
```diff
@@ -69,7 +72,7 @@
         // Methods
 
         /**
-         * Recreates the entire bounding box from scratch
+         * Recreates the entire bounding box from scratch, producing same values as if the constructor was called.
```
deltakosh: looks like a breaking change to me

barroij: is it really unwanted as the comment said that it recreates the entine BBox from scratch ?
Also, from within the engine, it seems that the usual way to compute the world woordinates boundingBox after construction or reconstruction is to call update, that's why computing them from reconstruct is kind of extra-useless computation.

src/Culling/babylon.boundingBox.ts
barroij: nope, const just ensures at "compile time" that a variable is not changed

deltakosh: ok so I would encourage using static readonly to save memory

src/Culling/babylon.boundingBox.ts
deltakosh: It is a public API so we do not know if people use it differently hence we cannot change the way it works

src/Culling/babylon.boundingBox.ts
```diff
@@ -131,7 +135,7 @@
          * @returns current bounding box
          */
         public setWorldMatrix(matrix: Matrix): BoundingBox {
-            this._worldMatrix.copyFrom(matrix);
+            this._worldMatrix = matrix;
```
deltakosh: Nope, you don't know if people will not change the matrix externally. copy has to be kept

src/Culling/babylon.boundingBox.ts
```diff
@@ -140,22 +144,29 @@
...
-
             const vectorsWorld = this.vectorsWorld;
             const vectors = this.vectors;
-            for (let index = 0; index < 8; ++index) {
-                const v = vectorsWorld[index];
-                Vector3.TransformCoordinatesToRef(vectors[index], world, v);
-                minWorld.minimizeInPlace(v);
-                maxWorld.maximizeInPlace(v);
+
+            if (world !== _identityMatrix) {
```
deltakosh: good idea

src/Culling/babylon.boundingBox.ts
```diff
@@ -140,22 +144,29 @@
...
+            }
+            else {
+                minWorld.copyFrom(this.minimum);
+                maxWorld.copyFrom(this.maximum);
+                for (let index = 0; index < 8; ++index) {
+                    vectorsWorld[index].copyFrom(vectors[index]);
+                }
             }
 
-            // Extend
```
deltakosh: Why removing comments?

src/Culling/babylon.boundingInfo.ts
```diff
@@ -167,9 +167,9 @@ module BABYLON {
 		 * Gets the world distance between the min and max points of the bounding box
 		 */
         public get diagonalLength(): number {
-            let boundingBox = this.boundingBox;
-            let size = boundingBox.maximumWorld.subtract(boundingBox.minimumWorld);
-            return size.length();
+            const boundingBox = this.boundingBox;
+            const diag = boundingBox.maximumWorld.subtractToRef(boundingBox.minimumWorld, Tmp.Vector3[0]);
```
deltakosh: Good catch

src/Culling/babylon.ray.ts
```diff
@@ -3,11 +3,13 @@ module BABYLON {
      * Class representing a ray with position and direction
      */
     export class Ray {
-        private _edge1: Vector3;
-        private _edge2: Vector3;
-        private _pvec: Vector3;
-        private _tvec: Vector3;
-        private _qvec: Vector3;
+        private static _edge1 = Vector3.Zero();
```
deltakosh: To save memory I would have preferred keeping them lazy created

deltakosh: nevermind, I'm dumb

src/Culling/babylon.boundingBox.ts
barroij: yup, I have no preference.
In both cases the Matrix is created only once. Having a closure on an unreachable and locally declared variable has the advantage of not "polluting" the class prototype.

I did it this way because that's what we were doing in BoundingSphere

src/Culling/babylon.boundingBox.ts
barroij: ok, I'll add another function that do what I want so that I don't change the API.

src/Culling/babylon.boundingBox.ts
barroij: It is just strange to not change it in place when we call update, but change it in place when we call setWorldMatrix.
I mean, if one calls `setWorldMatrix` after an update, the matrix that will be modified will be one the one passed to the `update` function, which is the world Matrix of the Mesh. Do we really want to change the world Matrix of the change in this case ? (see our previous discussion here https://github.com/BabylonJS/Babylon.js/pull/5219#issuecomment-423580302)

src/Culling/babylon.boundingBox.ts
barroij: yes, but can only work if this._worldMatrix is never modified in place :)

src/Culling/babylon.boundingBox.ts
barroij: Actually this function should be removed IMHO. Having an accessor to it for debugging purpose is fine, but why a setter ? The only way to change it should to call `reConstruct` or `update`, as it would also update all the world coordinates stuff.

I guess some API change is allowed for a major version release such as 4.0, isn't it ? 😉

src/Culling/babylon.boundingBox.ts
barroij: Also it is strange that's the behavior `reConstruct` is not the same between BoundingBox and BoundingSphere

barroij:
@deltakosh Ok I reverted my changes to keep the API unchanged (even if some API change for a major release should be ok...)

Still, there is an inconherence between how the BoundingSphere and the BoundingBox behaves, although there are used exactly the same way and the API and the comments are the same. This is, at least, counter-intuitive. As I had did there, I would recommend to align BoundingBox on BoundingSphere, meaning that a call to `reConstruct` produces exactly same results as the constructor, and that there is no way to change the world matrix of the Bbox byt by calling `reConstruct` or `update`

I added a small change in the Matrix API so that the cached value `_isIdentity` can be properly set when the Identity matrix is constructed. I did not enable it by default because of the following 👍 The result of `isIdentity` might depend on a `considerAsTextureMatrix` parameter, which is pretty bad. It prevents a lot of optimisation when dealing with Identity matrix, as the identity-ness depends on external context which is not stored into the Matrix. So, if a Matrix has its `_isIdentity` to true, there is no way to figure whether it was computed with `considerAsTextureMatrix` to true or false. I would higly recomment to remove that and find another way to deal with 3x2 matrix. Moreover, the code seems to be just wrong : 
````javacsript
        public isIdentity(considerAsTextureMatrix = false): boolean {
            if (this._isIdentityDirty) {
                this._isIdentityDirty = false;
                if (this.m[0] !== 1.0 || this.m[5] !== 1.0 || this.m[15] !== 1.0) {
                    this._isIdentity = false;
                } else if (this.m[1] !== 0.0 || this.m[2] !== 0.0 || this.m[3] !== 0.0 ||
                    this.m[4] !== 0.0 || this.m[6] !== 0.0 || this.m[7] !== 0.0 ||
                    this.m[8] !== 0.0 || this.m[9] !== 0.0 || this.m[11] !== 0.0 ||
                    this.m[12] !== 0.0 || this.m[13] !== 0.0 || this.m[14] !== 0.0) {
                    this._isIdentity = false;
                } else {
                    this._isIdentity = true;
                }

                if (!considerAsTextureMatrix && this.m[10] !== 1.0) {
                    this._isIdentity = false;
                }
            }

            return this._isIdentity;
        }
````
the flag can be set to false even if value that are not in the 3x2 matrix are non-zero

File: src/Culling/babylon.boundingBox.ts [hunk 11]
```diff
@@ -93,7 +93,7 @@
             max.addToRef(min, this.center).scaleInPlace(0.5);
             max.subtractToRef(min, this.extendSize).scaleInPlace(0.5);
 
-            this._update(worldMatrix || this._worldMatrix || Matrix.Identity());
+            this._update(worldMatrix || this._worldMatrix || Matrix.Identity(true));
```
deltakosh: That should be always true no?

src/Math/babylon.math.ts
```diff
@@ -5075,35 +5082,48 @@ module BABYLON {
...
+            return identity;
         }
 
         /**
          * Creates a new identity matrix and stores the result in a given matrix
+         * @param markAsUpdated if set to true, the cached value _isIdentityDirty and _isIdentity will be updated
          * @param result defines the target matrix
          */
-        public static IdentityToRef(result: Matrix): void {
+        public static IdentityToRef(result: Matrix, markAsUpdated = false): void {
```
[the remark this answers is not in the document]
barroij: you're right, will remove it

deltakosh:
A tiny conflict and we can merge

PR #5328, opened 2018-10-08
Fix LinesMesh EdgeRendering + make it work for instance of LinesMesh

barroij:
There was a bug in the the linesMesh edgeRenderer as it assumes that the LinesMesh was made of a single polyline.

I also made a change that makes the LinesMeshEdgeRenderer working properly with instances of LinesMesh


src/Rendering/babylon.lineEdgesRenderer.ts
```diff
@@ -96,21 +26,13 @@ module BABYLON {
                 return;
             }
 
-            // First let's find adjacencies
```
deltakosh: Can you explain why this code was removed?

barroij: the question should be : why it has been added first ?
My gut feeling is that this implementation starts with a quick'n dirty copy/paste of the `EdgesRenderer` implementation, with as little change as possible to make it work for LinesMesh.

Unless I'm wrong and I'm missunderstanding what a `LinesMesh` is (only lines, no faces at all), there is no point having and computing something like `FaceAdjacencies` here. We already know what are the lines to be rendered, as the current mesh is made of lines only.

deltakosh: Well you are right... Just to be sure did you run some tests? can you share a playground?

barroij: I test it in the app I'm developping, but I can build a small PG for that. will do either later today or tomorow

deltakosh: Thanks so I will merge and add the test as visual validation

barroij:
playground http://www.babylonjs-playground.com/#T90MQ4#1
There is 1 "original" LinesMesh made of two crossing lines, and 1 instance of this LinesMesh
the original one is highligth in blue, and the instance in yellow 

the LineEdgesRenderer before :
image

with my fix

image

Issue #5324, opened 2018-10-07
AssetContainer is missing EffectLayers ( HighlightLayer )


ssaket:
Both addAllToScene() and removeAllFromScene() methods of AssetContainer
doesn't contain effectLayers, because of this the highlightLayer is not added back to the scene.

PG - https://www.babylonjs-playground.com/#5NFRVE#5

Not sure how the GlowLayer is working ! 
Is there any other way to load HighlightLayer ? If not I have a PR ready :)

TrevorDev:
Fixed in pr,
See updated pg: http://playground.babylonjs.com#5NFRVE#11, the reason glow was working was that it was not being added/removed from the scene, asset container already had some logic related to effect layer that seemed to be causing an issue when adding/removing the asset container.

Im not aware of another way to load highlight layer, but maybe @sebavan knows? Feel free to make a PR if you already have something so we can take a look.

ssaket:
@TrevorDev - thanks for resolving this. 
I took a different approach as I didn't know they were part of components. What I did was to add the effectLayers as member of AbstractScene, and loading them like any other (meshes, light etc). But having them in components makes more sense !

oschakravarthi:
Hi, Is it fixed and released?

sebavan:
Yes it is :-)

PR #5329, opened 2018-10-08
add effect layer to asset container

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/5324

### DIFF
--- a/src/babylon.assetContainer.ts
+++ b/src/babylon.assetContainer.ts
@@ -20,6 +20,11 @@ module BABYLON {
         constructor(scene: Scene) {
             super();
             this.scene = scene;
+            this["sounds"] = [];
+            this["effectLayers"] = [];
+            this["layers"] = [];
+            this["lensFlareSystems"] = [];
+            this["proceduralTextures"] = [];
         }
 
         /**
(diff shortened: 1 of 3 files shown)


src/babylon.assetContainer.ts
```diff
@@ -65,10 +65,9 @@ module BABYLON {
             this.textures.forEach((o) => {
                 this.scene.addTexture(o);
             });
-
-            for (let component of this.scene._serializableComponents) {
-                component.addFromContainer(this.scene);
```
TrevorDev: @sebavan it looks like you added this section that I removed. It looks like component.addFromContainer(this.scene); has logic specific to effect layer and keeping it had unexpected results. Is it needed now that I added effectLayer?

sebavan: Effect Layer should only be in components not directly in the container. It has been moved during the module work. 

What is the issue after the move ?

TrevorDev: ok it looks like there was a bug with this.scene when it should be this. Also I still want to be abloe to do asset container.effectLayers.push(layer) is there a way you recomend that?

sebavan: Is the problem the empty list on creation ?

TrevorDev: should be good now as we discussed

PR #5338, opened 2018-10-09
align BoundingBox and BoundingSphere API and behavior

barroij:
For more clarity, BoundingBox and BoundingSphere have know the same API and the same behavior when the `reConstruct` or `scale` method is called.

This PR also adds some more type-safety-ness


src/Culling/babylon.boundingInfo.ts
```diff
@@ -1,25 +1,22 @@
...
-        var r1 = Math.abs(Vector3.Dot(box.directions[1], axis)) * box.extendSize.y;
-        var r2 = Math.abs(Vector3.Dot(box.directions[2], axis)) * box.extendSize.z;
-
-        var r = r0 + r1 + r2;
-        return {
-            min: p - r,
-            max: p + r
-        };
+    const _result0 = { min: 0, max: 0};
+    const _result1 = { min: 0, max: 0};
```
barroij: prevent creating two temporary object each time axisOverlap  is called (which happen up to 15 times when BoundingInfo.intersects is called)

src/Culling/babylon.boundingInfo.ts
```diff
@@ -217,11 +214,7 @@ module BABYLON {
          * @param precise if the intersection should be done using OBB
          * @returns if the bounding info intersects
          */
-        public intersects(boundingInfo: BoundingInfo, precise: boolean): boolean {
-            if (!this.boundingSphere.centerWorld || !boundingInfo.boundingSphere.centerWorld) {
-                return false;
-            }
-
```
barroij: removing this test as the condition is always false.

src/Math/babylon.math.ts
```diff
@@ -4220,7 +4220,6 @@ module BABYLON {
      * Class used to store matrix data (4x4)
      */
     export class Matrix {
-        private static _tempQuaternion: Quaternion = new Quaternion();
```
barroij: use MathTmp.Quaternion instead

PR #5340, opened 2018-10-10
Math

barroij:
I ended up with making a global pass the math file.
Changes here are mainly esthetics, but a few things are worth mentionning: 

- I added a type `DeepImmutable<T>`. Variables of this type can be compared to C++ const reference. They cannot be modified at all. It makes sure that we are not accidentaly modyfing something and make it clear from the function signature which parameters are changed by the function, and which are not. (For instance it shows a bug in `Vector3.UnprojectFromTransform`).

- In the `Matrix` class When possible, I use the `_isIdentity` value. I changed the logic a bit so that we don't to check if `_isIdentityDirty` is true or false. 

I also have a few questions :
- What `Quaternion.AreClose` function is supposed to do ? 
````
        public static AreClose(quat0: DeepImmutable<Quaternion>, quat1: DeepImmutable<Quaternion>): boolean {
            let dot = Quaternion.Dot(quat0, quat1);

            return dot >= 0;
        }
````
It is not used internally.

- `Matrix.RotationAxisToRef` does not set the values for index 12, 13 and 14. Is it a bug ?

### DIFF
--- a/src/Math/babylon.math.ts
+++ b/src/Math/babylon.math.ts
@@ -76,11 +76,7 @@ module BABYLON {
          * @param index defines an optional index in the target array to define where to start storing values
          * @returns the current Color3 object
          */
-        public toArray(array: FloatArray, index?: number): Color3 {
-            if (index === undefined) {
-                index = 0;
-            }
-
+        public toArray(array: FloatArray, index: number = 0): Color3 {
             array[index] = this.r;
             array[index + 1] = this.g;
             array[index + 2] = this.b;
@@ -93,7 +89,7 @@ module BABYLON {
          * @param alpha defines the alpha component on the new Color4 object (default is 1)
          * @returns a new Color4 object
          */
-        public toColor4(alpha = 1): Color4 {
+        public toColor4(alpha: number = 1): Color4 {
             return new Color4(this.r, this.g, this.b, alpha);
         }
 
@@ -960,7 +956,7 @@ module BABYLON {
          * @param source defines the source Vector2
          * @returns the current updated Vector2
          */
-        public copyFrom(source: Vector2): Vector2 {
+        public copyFrom(source: DeepImmutable<Vector2>): Vector2 {
             this.x = source.x;
             this.y = source.y;
             return this;
@@ -992,7 +988,7 @@ module BABYLON {
          * @param otherVector defines the other vector
          * @returns a new Vector2 set with the addition of the current Vector2 and the given one coordinates
          */
-        public add(otherVector: Vector2): Vector2 {
+        public add(otherVector: DeepImmutable<Vector2>): Vector2 {
             return new Vector2(this.x + otherVector.x, this.y + otherVector.y);
         }
 
@@ -1002,7 +998,7 @@ module BABYLON {
          * @param result defines the target vector
          * @returns the unmodified current Vector2
          */
-        public addToRef(otherVector: Vector2, result: Vector2): Vector2 {
+        public addToRef(otherVector: DeepImmutable<Vector2>, result: Vector2): Vector2 {
             result.x = this.x + otherVector.x;
             result.y = this.y + otherVector.y;
             return this;
@@ -1013,7 +1009,7 @@ module BABYLON {
          * @param otherVector defines the other vector
          * @returns the current updated Vector2
          */
-        public addInPlace(otherVector: Vector2): Vector2 {
+        public addInPlace(otherVector: DeepImmutable<Vector2>): Vector2 {
             this.x += otherVector.x;
             this.y += otherVector.y;
             return this;
@@ -1024,7 +1020,7 @@ module BABYLON {
          * @param otherVector defines the other vector
          * @returns a new Vector2
          */
-        public addVector3(otherVector: Vector3): Vector2 {
+        public addVector3(otherVector: DeepImmutable<Vector3>): Vector2 {
             return new Vector2(this.x + otherVector.x, this.y + otherVector.y);
         }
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Math/babylon.math.ts [hunk 1]
```diff
@@ -2455,11 +2476,10 @@ module BABYLON {
          * @param vector defines the Vector3 to normalize
          * @param result defines the Vector3 where to store the result
          */
-        public static NormalizeToRef(vector: Vector3, result: Vector3): void {
+        public static NormalizeToRef(vector: DeepImmutable<Vector3>, result: Vector3): void {
             vector.normalizeToRef(result);
         }
 
-        private static _viewportMatrixCache: Matrix;
```
barroij: remove, used MathTmp.Matrix[ instead

src/Math/babylon.math.ts
```diff
@@ -387,7 +383,7 @@ module BABYLON {
          * @returns a new Color3 object
          */
         public static FromInts(r: number, g: number, b: number): Color3 {
-            return new Color3(r / 255.0, g / 255.0, b / 255.0);
+            return new Color3(r / 255, g / 255, b / 255);
```
barroij: Removed all the  useless `.0`, it only adds noise.
Also it suggest that there might be a difference between floating point value and integer in javascript, which is wrong, everything is number

src/Math/babylon.math.ts
```diff
@@ -2498,17 +2517,19 @@ module BABYLON {
...
             matrix.invert();
-            source.x = source.x / viewportWidth * 2 - 1;
-            source.y = -(source.y / viewportHeight * 2 - 1);
-            var vector = Vector3.TransformCoordinates(source, matrix);
-            var num = source.x * matrix.m[3] + source.y * matrix.m[7] + source.z * matrix.m[11] + matrix.m[15];
+            var screenSource = MathTmp.Vector3[0];
+            screenSource.x = source.x / viewportWidth * 2 - 1;
+            screenSource.y = -(source.y / viewportHeight * 2 - 1);
+            screenSource.z =  source.z;
+            var vector = Vector3.TransformCoordinates(screenSource, matrix);
```
barroij: Here we were modifying the `source` parameter, which is likely to be unexpected

src/Math/babylon.math.ts
```diff
@@ -4314,16 +4304,24 @@ module BABYLON {
...
         public determinant(): number {
-            var temp1 = (this.m[10] * this.m[15]) - (this.m[11] * this.m[14]);
-            var temp2 = (this.m[9] * this.m[15]) - (this.m[11] * this.m[13]);
-            var temp3 = (this.m[9] * this.m[14]) - (this.m[10] * this.m[13]);
-            var temp4 = (this.m[8] * this.m[15]) - (this.m[11] * this.m[12]);
-            var temp5 = (this.m[8] * this.m[14]) - (this.m[10] * this.m[12]);
-            var temp6 = (this.m[8] * this.m[13]) - (this.m[9] * this.m[12]);
+            if (this.isIdentity) {
+                return 1;
+            }
```
barroij: prevent computation wihen the matrix is identity

src/Math/babylon.math.ts
```diff
@@ -4404,68 +4402,75 @@ module BABYLON {
...
-            other.m[3] = -(((l2 * l34) - (l3 * l35)) + (l4 * l36)) * l27;
-            other.m[7] = (((l1 * l34) - (l3 * l37)) + (l4 * l38)) * l27;
-            other.m[11] = -(((l1 * l35) - (l2 * l37)) + (l4 * l39)) * l27;
-            other.m[15] = (((l1 * l36) - (l2 * l38)) + (l3 * l39)) * l27;
-
-            other._markAsUpdated();
+        public invertToRef(result: Matrix): Matrix {
+            if (this.isIdentity) {
+                Matrix.IdentityToRef(result);
+                return this;
```
barroij: prevent computation when the matrix is identity

src/Math/babylon.math.ts
```diff
@@ -4589,40 +4594,55 @@ module BABYLON {
...
+                for (var index = 0; index < 16; ++index) {
+                    result[offset + index] = otherM[index];
+                }
+                return this;
+            }
+            if (other.isIdentity) {
+                for (var index = 0; index < 16; ++index) {
+                    result[offset + index] = m[index];
+                }
+                return this;
```
barroij: prevent computation when the matrix is identity

src/Math/babylon.math.ts
```diff
@@ -4698,16 +4729,30 @@ module BABYLON {
...
+                if (translation) {
+                    translation.setAll(0);
+                }
+                if (scale) {
+                    scale.setAll(1);
+                }
+                if (rotation) {
+                    rotation.set(0, 0, 0, 1);
+                }
+                return true;
```
barroij: prevent computation when the matrix is identity

src/Math/babylon.math.ts
```diff
@@ -5130,65 +5159,35 @@ module BABYLON {
...
         public static Zero(): Matrix {
-            const zero = Matrix.FromValues(0.0, 0.0, 0.0, 0.0,
-                0.0, 0.0, 0.0, 0.0,
-                0.0, 0.0, 0.0, 0.0,
-                0.0, 0.0, 0.0, 0.0);
+            const zero = Matrix.FromValues(
+                0, 0, 0, 0,
+                0, 0, 0, 0,
+                0, 0, 0, 0,
+                0, 0, 0, 0);
```
barroij: `new Matrix();` would be enough I think

barroij:
Another think I'd like to mention : 
There are a few places where we are testing strict equality versus 0 to prevent a division by 0. It should be update to also discard value very close to 0.
example :
````
        public normalize(): Vector4 {
            var len = this.length();

            if (len === 0) {
                return this;
            }

            return this.scaleInPlace(1 / len);
        }
````

barroij:
closing for now, some tests are failing

PR #5344, opened 2018-10-10
Fix gulp mac

sebavan:


sebavan:
Let s wait for the build before merging

PR #5345, opened 2018-10-10
fix Matrix.toNormalMatrix

barroij:
discussion here : https://github.com/BabylonJS/Babylon.js/pull/3706#discussion_r223988032


barroij:
@deltakosh here is the fix for the `toNormalMatrix ` implementation

PR #5347, opened 2018-10-10
Use the Matrix isIdentity cached value to speed up thing when possible

barroij:
This PR contains some kind of optimization  when we know that a matrix is Identity .

It also contain some code polishing


barroij:
will create another PR

PR #5350, opened 2018-10-11
Use the Matrix isIdentity cached value to speed up thing when possible

barroij:


src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -320,11 +320,9 @@ module BABYLON {
                     this._standingMatrix = new Matrix();
                     Matrix.FromFloat32ArrayToRefScaled(result.vrDisplay.stageParameters.sittingToStandingTransform, 0, 1, this._standingMatrix);
                     if (!this.getScene().useRightHandedSystem) {
-                        [2, 6, 8, 9, 14].forEach((num) => {
-                            if (this._standingMatrix) {
-                                this._standingMatrix.m[num] *= -1;
-                            }
-                        });
+                        if (this._standingMatrix) {
+                            this._standingMatrix.toggleModelMatrixHandInPlace();
```
barroij: use existing method for that

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -634,9 +632,7 @@ module BABYLON {
             Matrix.FromArrayToRef(viewArray, 0, this._webvrViewMatrix);
 
             if (!this.getScene().useRightHandedSystem) {
-                [2, 6, 8, 9, 14].forEach((num) => {
-                    this._webvrViewMatrix.m[num] *= -1;
-                });
+                this._webvrViewMatrix.toggleModelMatrixHandInPlace();
```
barroij: use existing method for that

src/Cameras/VR/babylon.webVRCamera.ts
```diff
@@ -686,9 +682,7 @@ module BABYLON {
 
             //babylon compatible matrix
             if (!this.getScene().useRightHandedSystem) {
-                [8, 9, 10, 11].forEach((num) => {
-                    this._projectionMatrix.m[num] *= -1;
-                });
+                this._projectionMatrix.toggleProjectionMatrixHandInPlace();
```
barroij: use existing method for that

src/Bones/babylon.bone.ts
```diff
@@ -751,9 +737,7 @@ module BABYLON {
                 }
             }
 
-            lmat.m[12] = lx;
-            lmat.m[13] = ly;
-            lmat.m[14] = lz;
+            lmat.setTranslationFromFloats(lx, ly, lz);
```
barroij: use existing method for that

src/Bones/babylon.bone.ts
```diff
@@ -612,12 +602,8 @@ module BABYLON {
          */
         public rotate(axis: Vector3, amount: number, space = Space.LOCAL, mesh?: AbstractMesh): void {
             var rmat = Bone._tmpMats[0];
-            rmat.m[12] = 0;
-            rmat.m[13] = 0;
-            rmat.m[14] = 0;
-
+            rmat.setTranslationFromFloats(0, 0, 0);
```
barroij: use existing method for that

src/Debug/babylon.skeletonViewer.ts
```diff
@@ -65,9 +65,7 @@ module BABYLON.Debug {
             if (x !== 0 || y !== 0 || z !== 0) {
                 var tmat2 = Tmp.Matrix[1];
                 BABYLON.Matrix.IdentityToRef(tmat2);
-                tmat2.m[12] = x;
-                tmat2.m[13] = y;
-                tmat2.m[14] = z;
+                tmat2.setTranslationFromFloats(x, y, z);
```
barroij: use existing method for that

src/Math/babylon.math.ts
```diff
@@ -4244,15 +4215,18 @@ module BABYLON {
...
         public updateFlag: number;
 
+        private readonly _m: Float32Array = new Float32Array(16);
+
         /**
-         * Gets or sets the internal data of the matrix
+         * Gets the internal data of the matrix
          */
-        public m: Float32Array = new Float32Array(16);
+        public get m(): Readonly<Float32Array> { return this._m; }
```
barroij: To not break the API,  I made an accessor for that. Still, if its then modified, there while be a typescript error.

src/Math/babylon.math.ts
```diff
@@ -4415,67 +4394,87 @@ module BABYLON {
...
+         * @param index the index of the value within the matrix. between 0 and 15.
+         * @param value the value to be added
+         * @returns the current updated matrix
+         */
+        public multiplyAtIndex(index: number, value: number): Matrix {
+            this._m[index] *= value;
+            this._markAsUpdated();
+            return this;
+        }
+
```
barroij: I added those two functions for enabling some by-index modification

barroij:
@deltakosh let's wait for the build to pass to be sure, but that should OK.

Note that if this PR is accepted, it means that we can assume the value of a Matrix are not modified from the outside. If so, we can go further and cached the determinant of the Matrix. That should then enable speeding up other function such as invert.

dist/preview release/what's new.md
```diff
@@ -64,3 +65,4 @@
 - `Database.openAsync` was renamed by `Database.open` ([Deltakosh](https://github.com/deltakosh))
 - `scene.database` was renamed to `scene.offlineProvider` ([Deltakosh](https://github.com/deltakosh))
 - `BoundingBox.setWorldMatrix` was removed. `BoundingBox.getWorldMatrix` now returns a `Readonly<Matrix>` ([barroij](https://github.com/barroij))
+- `Matrix`'s accessor `m` and method `toArray` and `adArray`now returns a `Readonly<Float32Array>` as the matrix underlying array is not supposed to be modified manually from the outside of the class ([barroij](https://github.com/barroij))
```
deltakosh: typo: asArray

src/Math/babylon.math.ts
```diff
@@ -5374,22 +5296,13 @@ module BABYLON {
...
-            result.m[15] = 1.0;
+            Matrix.FromValuesToRef(
+                x, 0.0, 0.0, 0.0,
+              0.0,   y, 0.0, 0.0,
+              0.0, 0.0,   z, 0.0,
+              0.0, 0.0, 0.0, 1.0,
+              result
+          );
 
             result._markAsUpdated();
```
deltakosh: not required

src/Math/babylon.math.ts
```diff
@@ -5519,32 +5435,40 @@ module BABYLON {
...
+
             // Z axis
-            target.subtractToRef(eye, this._zAxis);
-            this._zAxis.normalize();
+            target.subtractToRef(eye, zAxis);
+            zAxis.normalize();
 
             // X axis
-            Vector3.CrossToRef(up, this._zAxis, this._xAxis);
+            Vector3.CrossToRef(up, zAxis, xAxis);
```
deltakosh: This will fail. The MathTmp.Vector3[0]  is used by the CrossToRef function

src/Math/babylon.math.ts
```diff
@@ -5570,32 +5494,40 @@ module BABYLON {
...
+
             // Z axis
-            eye.subtractToRef(target, this._zAxis);
-            this._zAxis.normalize();
+            eye.subtractToRef(target, zAxis);
+            zAxis.normalize();
 
             // X axis
-            Vector3.CrossToRef(up, this._zAxis, this._xAxis);
+            Vector3.CrossToRef(up, zAxis, xAxis);
```
deltakosh: same here, the vector3[0] is already used by CrossToRef

deltakosh: Perhaps we should move the CrossToRef to use a bigger number like MathTmp.Vector3[7]

deltakosh: Or even better create a temp just for it

barroij:
@deltakosh that's strange, all the tests pass locally ..

barroij:
Note that the modifications in this PR might result in slightly different results as if we know when a matrix is identity, we sometimes can return the value without any computation. In that case, the value we return might be different because of limited Floating point precision.

deltakosh:
Ok I'll merge tomorrow and I will update the visual tests

src/Math/babylon.math.ts
```diff
@@ -5935,25 +5892,29 @@ module BABYLON {
...
-            result.m[12] = matrix.m[3];
-            result.m[13] = matrix.m[7];
-            result.m[14] = matrix.m[11];
-            result.m[15] = matrix.m[15];
+            rm[12] = mm[3];
+            rm[13] = mm[7];
+            rm[14] = mm[11];
+            rm[15] = mm[15];
+            // identity-ness does not change when transposing
+            result._updateIdentityStatus(matrix._isIdentity, matrix._isIdentityDirty);
```
barroij: here we did not call `_markAsUpdated`

src/Math/babylon.math.ts
```diff
@@ -5889,7 +5846,9 @@ module BABYLON {
                 0.0, 0.0, zmax - zmin, 0.0,
                 cx + cw / 2.0, ch / 2.0 + cy, zmin, 1);
 
-            return world.multiply(view).multiply(projection).multiply(viewportMatrix);
+            world.multiplyToRef(view, MathTmp.Matrix[0]);
+            MathTmp.Matrix[0].multiplyToRef(projection, MathTmp.Matrix[1]);
+            return MathTmp.Matrix[1].multiply(viewportMatrix);
```
barroij: this change is to avoid creating two new temporary Matrix

barroij:
@deltakosh I did a final pass on it, 
Do you know why `RotationAxisToRef` does not set the values at idx 12, 13 and 14 (the translation slots)? it the only xxxToRef that does not set all the values in the matrix. If we need such an optimisation, we may add a parameter to the function, but the default behavior would better be consistent with the rest of the API

deltakosh:
This is a bug! I'll fix it

PR #5361, opened 2018-10-12
Web xr enter exit ui

TrevorDev:
Testing PG: http://playground.babylonjs.com#GVAEUW#96

### DIFF
--- a/src/Cameras/XR/babylon.webXREnterExitUI.ts
+++ b/src/Cameras/XR/babylon.webXREnterExitUI.ts
@@ -1,0 +1,56 @@
+module BABYLON {
+    export class WebXREnterExitUI implements IDisposable {
+        public overlay:HTMLDivElement;
+        public buttons:Array<{element:HTMLElement, initializationOptions:XRSessionCreationOptions}> = []
+        public static CreateAsync(scene:BABYLON.Scene, helper:WebXRExperienceHelper){
+            var ui = new WebXREnterExitUI(scene, helper);
+            var supportedPromises = ui.buttons.map((btn)=>{
+                return helper.supportsSession(btn.initializationOptions);
+            });
+            return Promise.all(supportedPromises).then((results)=>{
+                results.forEach((supported, i)=>{
+                    if(supported){
+                        ui.overlay.appendChild(ui.buttons[i].element)
+                        ui.buttons[i].element.onclick = async()=>{
+                            if(helper.state == BABYLON.WebXRState.IN_XR){
+                                await helper.exitXR()
+                                return
+                            }else if(helper.state == BABYLON.WebXRState.NOT_IN_XR){
+                                await helper.enterXR(ui.buttons[i].initializationOptions, "eye-level");
+                            }
+                        }
+                    }
+                })
+            })
+        }
+        private constructor(private scene:BABYLON.Scene, private helper:WebXRExperienceHelper){
+            this.overlay = document.createElement("div");
+            this.overlay.style.cssText = "z-index:11;position: absolute; right: 20px;bottom: 50px;"
+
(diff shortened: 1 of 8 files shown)


src/Cameras/XR/babylon.webXREnterExitUI.ts
```diff
@@ -0,0 +1,77 @@
...
+                        };
+                    }
+                });
+            });
+        }
+        private constructor(private scene: BABYLON.Scene, options: WebXREnterExitUIOptions) {
+            this._overlay = document.createElement("div");
+            this._overlay.style.cssText = "z-index:11;position: absolute; right: 20px;bottom: 50px;";
+
+            var hmdBtn = document.createElement("button");
```
deltakosh: we need to test this in the PG and sandbox context to be sure it works everywhere.
Also do you think we could add an option to let the user define the css?

src/Cameras/XR/babylon.webXRExperienceHelper.ts
```diff
@@ -45,27 +54,39 @@ module BABYLON {
         private _nonVRCamera: Nullable<Camera> = null;
         private _originalSceneAutoClear = true;
 
-        private _outputCanvas: HTMLCanvasElement;
-        private _outputCanvasContext: WebGLRenderingContext;
+        private _supported = false;
+
+        /**
+         * Creates the experience helper
+         * @param scene the scene to attach the exprience helper to
```
deltakosh: typo

src/Cameras/XR/babylon.webXRManagedOutputCanvas.ts
```diff
@@ -0,0 +1,62 @@
...
+     */
+    export class WebXRManagedOutputCanvas implements IDisposable {
+        private _canvas: Nullable<HTMLCanvasElement> = null;
+        /**
+         * xrpresent context of the canvas which can be used to display/mirror xr content
+         */
+        public canvasContext: Nullable<WebGLRenderingContext> = null;
+        /**
+         * Initializes the canvas to be added/removed upon entering/exiting xr
+         * @param helper the xr expreince helper used to trigger adding/removing of the canvas
```
deltakosh: typo

src/Cameras/XR/babylon.webXRManagedOutputCanvas.ts
```diff
@@ -0,0 +1,62 @@
...
+         * @param canvas The canvas to be added/removed (If not specified a full screen canvas will be created)
+         */
+        public constructor(helper: WebXRExperienceHelper, canvas?: HTMLCanvasElement) {
```
deltakosh: I'm wondering why you need to provide a canvas? Could it be jsut an engine? or could we have a canvas which is different from the main engine one?

TrevorDev: This is the canvas that the xr content is displayed on. Initially I wanted it to just use the current canvas that the engine uses but this isn't supported by the current spec (see issue I created https://github.com/immersive-web/webxr/issues/341). So right now I recommend just having a second full screen canvas, which is what this function will create if canvas isn't specified.

src/Cameras/XR/babylon.webXREnterExitUI.ts
```diff
@@ -7,6 +7,7 @@ module BABYLON {
          * Context to enter xr with
          */
         outputCanvasContext?: Nullable<WebGLRenderingContext>;
+        customButtons?: Array<{element: HTMLElement, initializationOptions: XRSessionCreationOptions}>;
```
deltakosh: XRSessionCreationOptions or WebXRSessionCreationOptions ?

PR #5364, opened 2018-10-15
Matrix det and invert

barroij:
@deltakosh  This PR is mainly esthetic as it gives proper variable names and explain what we do in the `determinant` and `invertToRef` Matrix method.

One important thing though : it takes care of non invertible Matrix, so that we never get Matrix filled with `Nan`


src/Math/babylon.math.ts
```diff
@@ -4300,19 +4300,31 @@ module BABYLON {
...
-              m[1] * (m[4] * temp1 - m[6] * temp4 + m[7] * temp5) +
-              m[2] * (m[4] * temp2 - m[5] * temp4 + m[7] * temp6) -
-              m[3] * (m[4] * temp3 - m[5] * temp5 + m[6] * temp6)
-            );
+            const m00 = m[0], m01 = m[1], m02 = m[2], m03 = m[3];
+            const m10 = m[4], m11 = m[5], m12 = m[6], m13 = m[7];
+            const m20 = m[8], m21 = m[9], m22 = m[10], m23 = m[11];
+            const m30 = m[12], m31 = m[13], m32 = m[14], m33 = m[15];
+            // https://en.wikipedia.org/wiki/Laplace_expansion
+            // to compute the derminant of a 4x4 Matrix we compute the cofactors of any row or column,
```
deltakosh: typo

PR #5379, opened 2018-10-17
Correct DEL overrun

BabylonJSGuide:
Holding down the DEL key as a quick way to delete to end of text continued to decrease the cursorOffset thus for example preventing the use of the BackSpace key. The additional condition prevents this. 

Also added  - the use of 59 as keycode for : in FF (as opposed to 186) prevented its use in inputText control so changed condition to < 60. 

Extended to < 64 to include = and ? in FF

### DIFF
--- a/gui/src/2D/controls/inputText.ts
+++ b/gui/src/2D/controls/inputText.ts
@@ -351,7 +351,7 @@ export class InputText extends Control implements IFocusableControl {
         if (key &&
             ((keyCode === -1) ||                     // Direct access
                 (keyCode === 32) ||                     // Space
-                (keyCode > 47 && keyCode < 58) ||       // Numbers
+                (keyCode > 47 && keyCode < 60) ||       // Numbers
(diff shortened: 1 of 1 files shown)


TrevorDev:
👍 Tested. I noticed a couple other issues though that are unrelated to this change. In firefox pressing backspace will cause the browser to go back also other keys like - still dont work. @deltakosh do you know why all the checks are here now? Could we remove all the OR's in the if statement and just have if(key){ . The only reason I think it's there is to stop pressing pageDown from printing out the entire world but I think it is likely ok.

deltakosh:
All the OR were here to make sure we are only checking printable characters

Also do you mind adding a evt.handled = true or a evt.preventDefault to see if it fix the backspace issue?
This could also fix the "/" key issue (which is a firefox shortcut)

TrevorDev:
@deltakosh Yep, I am already fixing those things in my PR, I think this one is good to be merged.

Issue #5375, opened 2018-10-17
Refocus on text input is not working


TrevorDev:
http://www.html5gamedevs.com/topic/40661-gui-input-text-behavior

Repro:
https://www.babylonjs-playground.com/#UWS0TS

Releated: https://github.com/BabylonJS/Babylon.js/pull/5058/files
This PR seems to have caused the issue: https://github.com/BabylonJS/Babylon.js/commit/91c00c0d9679a37aab8cc1dfaff48b8b10ff2c89

This used to work before because _initClickEvent called cb(clickInfo, this._currentPickResult); twice and the second call made control.ts _onPointerUp fire but now it doesnt get called twice due to the pr above.

control.ts _onPointerUp is not firing causing _downCount to not be set to 0 so when _onPointerDown is called, it is skipped as it is != 0. Input.ts calls control.ts _onPointerDown when it is called making it so that the focus is never set.

Issue #4236, opened 2018-04-29
Fix issue with drag'n'drop and ADT


deltakosh:
https://www.babylonjs-playground.com/#GL5SIM%234
http://www.html5gamedevs.com/topic/37297-gui-questionbug-mouse-dragging-across-gui-buttons/

PR #5382, opened 2018-10-17
Input refocus not working

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/5375

Testing PGs:
http://localhost:1338/Playground/index-local.html#WIACBP
http://localhost:1338/Playground/index-local.html#S7L7FE
http://localhost:1338/Playground/index-local.html#UWS0TS
http://localhost:1338/Playground/index-local.html#2H2LUT
http://localhost:1338/Playground/index-local.html#GL5SIM#4
http://localhost:1338/Playground/index-local.html#8MGKWK#25

This PR seems to have caused the issue: 91c00c0

This used to work before because _initClickEvent called cb(clickInfo, this._currentPickResult); twice and the second call made control.ts _onPointerUp fire but now it doesnt get called twice due to the pr above.

control.ts _onPointerUp is not firing causing _downCount to not be set to 0 so when _onPointerDown is called, it is skipped as it is != 0. Input.ts calls control.ts _onPointerDown when it is called making it so that the focus is never set.


gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -590,7 +590,7 @@ export class AdvancedDynamicTexture extends DynamicTexture {
         }
 
         this._pointerMoveObserver = scene.onPrePointerObservable.add((pi, state) => {
-            if (scene!.isPointerCaptured((<PointerEvent>(pi.event)).pointerId)) {
+            if (scene!.isPointerCaptured((<PointerEvent>(pi.event)).pointerId) && pi.type !== PointerEventTypes.POINTERUP) {
```
TrevorDev: See PR description, let me know if this change is the right one.

deltakosh: What about adding && this._lastControlOver[pointerId] so we make sure to only transmit the event if there is a potential receiver?

TrevorDev: I added 

            if(!this._lastControlOver[(<PointerEvent>(pi.event)).pointerId]){
                return;
            }

if that is what you mean't. It broke playground.babylonjs.com#S7L7FE as there is no this._lastControlOver[pointerId] ever causing it to never get focus.

TrevorDev: Another potential fix is to change when isPointerCaptured is true. scene!.isPointerCaptured is set to true after pointerdown is fired and set to false after prePointerObservables. isPointerCaptured was created as part of this pr https://github.com/BabylonJS/Babylon.js/issues/4236 . If i move the  isPointerCaptured  setting to false before prePoitnerObservable then it still works and I'd expect scene!.isPointerCaptured to be false here.

deltakosh: Well I meant:
if (scene!.isPointerCaptured((<PointerEvent>(pi.event)).pointerId) 
&& (pi.type !== PointerEventTypes.POINTERUP) || !this._lastControlOver[(<PointerEvent>(pi.event)).pointerId]) {

src/babylon.scene.ts
```diff
@@ -2083,6 +2083,8 @@ module BABYLON {
                 }
 
                 this._initClickEvent(this.onPrePointerObservable, this.onPointerObservable, evt, (clickInfo: ClickInfo, pickResult: Nullable<PickingInfo>) => {
+                    this._pointerCaptures[evt.pointerId] = false;
```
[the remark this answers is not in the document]
TrevorDev: If you mean this pg: http://playground.babylonjs.com#GL5SIM#4 then yes

PR #5383, opened 2018-10-18
Update readme example to ES6 syntax

pascalbayer:
Since ES6 is now widely adopted by all major browsers it might make sense to update the example in the readme to ES6 syntax


deltakosh:
What about IE?:(

pascalbayer:
Okay, thanks for the quick feedback anyways.

PR #5385, opened 2018-10-18
Prevent doing same operation twice when constructing a Ray for intersection

barroij:
This PR contains several things :
- it adds an `UnprojectRayToRef` static function to avoid computing and inverting the projection matrix twice when updating a `Ray`
- it adds `UnprojectFromInvertedMatrixToRef` to share and isolate some (strange) code and avoid copy-paste.
- it prevents creation of temporary Vector3 in `Ray.intersectionSegment`


barroij:
@deltakosh my bad, should be all right now

PR #5392, opened 2018-10-20
Xr camera movement

TrevorDev:


src/Math/babylon.math.ts
```diff
@@ -4054,6 +4108,54 @@ module BABYLON {
...
         }
 
+        /**
+         * Create a quaternion from Euler rotation angles
+         * @param x Pitch
+         * @param y Yaw
+         * @param z Roll
+         * @returns the new Quaternion
+         */
+        public static FromEulerAngles(x: number, y: number, z: number): Quaternion {
```
deltakosh: What is the value added of these functions?

TrevorDev: frequently i want to convert euler to quaternion and RotationYawPitchRollToRef causes confusion as parameter order is not x,y,z. Also I can never find it using autocomplete because i forget its called yawpitchroll.

deltakosh: Makes sense;)

PR #5394, opened 2018-10-20
Solve typescript 3.1.x compatibility issues for babylonjs 3.3 release

pascalbayer:
This PR will solve typing incompatibilities with typescript v3.1.3 (between `lib.dom.d.ts` and `babylon.d.ts`) which is the minimum required version of typescript for Angular 7 applications.

If possible I would like to see this PR result in a v3.3.1 of Babylon.js. Please provide further guidance if there are any other changes necessary.

### DIFF
--- a/Playground/babylon.d.txt
+++ b/Playground/babylon.d.txt
@@ -249,7 +249,6 @@ interface Document {
     webkitCancelFullScreen(): void;
     requestPointerLock(): void;
     exitPointerLock(): void;
-    fullscreen: boolean;
     mozFullScreen: boolean;
     msIsFullScreen: boolean;
     readonly webkitIsFullScreen: boolean;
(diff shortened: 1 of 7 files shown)


deltakosh:
No problem to merge in 4.0 but we  will not create a 3.3.1 for it (you can still get the tag and build your own if required but I would highly recommend take the 4.0 as it is as stable as the 3..3.)

pascalbayer:
Okay, seems like 4.0 will work for me and already has these fixes applied. However, I was a bit concerned switching to the preview versions again. I did that with 3.3 to test it early and it took almost until the final release to get the GTLF texture glitches solved there.

deltakosh:
Well we close bugs in less than 48h on the preview branch so if you have any issue just open a bug and we will fix it

PR #5400, opened 2018-10-22
rename without babylon

TrevorDev:
Script:

var fs = require('fs');
var glob = require("glob-promise")

var main = async ()=>{
   var files = await glob("src/**/*.ts")
   console.log(files[0])
   console.log(files[0].replace("babylon.", ""))
   files.forEach((file)=>{
        fs.rename(file, file.replace("babylon.", ""), (err, data)=>{
            console.log(err)
        })
   })
}
main();


sebavan:
Thx

PR #5427, opened 2018-10-26
add InstancedLinesMesh

barroij:
add a dedicated `InstancedLinesMesh` to handle instances of LinesMesh.

This is needed as different instances of a LinesMesh might have different `intersectionThreshold` value depending on their world transform.


barroij:
@deltakosh one more :)

barroij:
@deltakosh hold on, some test are failing, I'll check if it's me.

barroij:
@deltakosh I've got also failing tests when I run them locally on master

barroij:
And now the tests are sucessful... looks like something not determinist in the build 🌪

deltakosh:
Which tests where failing?

barroij:
localy, the Running Sliders (on master branch too)
image


I have the same kind of difference in "Advanced Shadow", but it does not cause the test to fail 👍 
image

deltakosh:
Sliders are for visual validation only so we're good!

PR #5432, opened 2018-10-27
Add a DeepImmutable<t> type

barroij:
This is to provide strong type checking when function input (or input) should ne considered the same way as it would be in C++ with a `pointer to const`. 
A deepImmutable object has all its property recursively readonly, so it cannot be modified.


barroij:
@deltakosh A proposal to enhance the type checking. I did not update all the files accordingly of course, only math.ts.

src/babylon.types.ts
```diff
@@ -26,4 +26,37 @@ module BABYLON {
...
+     */
+    export type Immutable<T> = T extends Primitive
+      ? T
+      : T extends Array<infer U>
+        ? ReadonlyArray<U>
+        : /* T extends Map<infer K, infer V> ? ReadonlyMap<K, V> : // es2015+ only */
+          DeepImmutable<T>;
+
+    /**
+     * Type modififier to make all the properties of an object Readonly recursively
```
deltakosh:
```suggestion
     * Type modifier to make all the properties of an object Readonly recursively
```

src/babylon.types.ts
```diff
@@ -26,4 +26,37 @@ module BABYLON {
...
     export type DataArray = number[] | ArrayBuffer | ArrayBufferView;
+
+    /**
+     * Alias type for primitive types
+     * @ignorenaming
+     */
+    type Primitive = undefined | null | boolean | string | number | Function;
+
+    /**
+     * Type modififier to make all the properties of an object Readonly
```
deltakosh:
```suggestion
     * Type modifier to make all the properties of an object Readonly
```

deltakosh:
I like it but would love to know if this has any impact on the generated JS? (I don't think so but worth checking)

Also adding @sebavan for review

barroij:
If I compare the generated babylon.max.js files in this branch against master, I can see no change because of this.

PR #5443, opened 2018-10-31
Deserialize morph target animations from animation groups

kcoley:
added support for deserializing morph target animations from animation groups

### DIFF
--- a/src/babylon.scene.ts
+++ b/src/babylon.scene.ts
@@ -3803,7 +3803,7 @@ module BABYLON {
          * @param id defines the id to search for
          * @return the found node or null if not found at all
          */
-        public getNodeByID(id: string): Nullable<Node | MorphTarget> {
+        public getNodeByID(id: string): Nullable<Node> {
             const mesh = this.getMeshByID(id);
             if (mesh) {
                 return mesh;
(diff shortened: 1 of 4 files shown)


src/babylon.scene.ts
```diff
@@ -3956,6 +3956,24 @@ module BABYLON {
             return null;
         }
 
+        /**
+         * Gets a morph target using a given id (if many are found, this function will pick the last one)
+         * @param id defines the id to search for
+         * @return the found morph target or null if not found at all.
+         */
+        public getMorphTargetByID(id: string): Nullable<MorphTarget> {
```
deltakosh:
```suggestion
        public getMorphTargetById(id: string): Nullable<MorphTarget> {
```

src/babylon.scene.ts
```diff
@@ -3956,6 +3956,24 @@ module BABYLON {
             return null;
         }
 
+        /**
+         * Gets a morph target using a given id (if many are found, this function will pick the last one)
```
deltakosh: the first one no?

kcoley: yes first one, I'll make the change

PR #5459, opened 2018-11-03
pointerDragBahavior: use AbstractMesh instead of Mesh

Poolminer:
To allow manually attaching the behavior to an AbstractMesh such as an InstancedMesh, without the type checker saying it's an error.


deltakosh:
Some checks are failing

Poolminer:
> 
> 
> Some checks are failing

I would have to add a note to the "what's new" it seems

TrevorDev:
@Poolminer Whats new is this file https://github.com/BabylonJS/Babylon.js/blob/master/dist/preview%20release/what's%20new.md which is used to document the changes for the next release

Poolminer:
I've added a note to the "what's new" file, so it should be OK now, but unfortunately _deploy/netlify_ got timed out; is there a way to retry?

deltakosh:
This is fine thanks!

PR #5475, opened 2018-11-06
Proposition: serialize and parse reflection probes

julien-moreau:
Reflection probe are standalone classes where its "cubeTexture" can be assigned to a material
When parsing a serialized material which has a reflection probe texture (typically reflectiont texture in PBR) fails.

To avoid creating multiple same reflection probes, the parse function searchs for an existing probe (identified by its name). If not found, returns a new reflection probe.

Do you agree with that kink of implementation? If yes, I can continue parsing and serialize the reflection probe (attached mesh, position, etc.)

### DIFF
--- a/src/Materials/Textures/babylon.texture.ts
+++ b/src/Materials/Textures/babylon.texture.ts
@@ -599,7 +599,7 @@ module BABYLON {
                 return parsedCustomTexture;
             }
 
-            if (parsedTexture.isCube) {
+            if (parsedTexture.isCube && !parsedTexture.isRenderTarget) {
                 return CubeTexture.Parse(parsedTexture, scene, rootUrl);
             }
 
@@ -619,8 +619,17 @@ module BABYLON {
 
                     return mirrorTexture;
                 } else if (parsedTexture.isRenderTarget) {
-                    var renderTargetTexture = new RenderTargetTexture(parsedTexture.name, parsedTexture.renderTargetSize, scene, generateMipMaps);
-                    renderTargetTexture._waitingRenderList = parsedTexture.renderList;
+                    let renderTargetTexture: Nullable<RenderTargetTexture> = null;
+                    if (parsedTexture.isCube) {
+                        const probe = ReflectionProbe.Parse(parsedTexture, scene, rootUrl);
+
+                        if (probe) {
+                            renderTargetTexture = probe.cubeTexture;
+                        }
+                    } else {
+                        renderTargetTexture = new RenderTargetTexture(parsedTexture.name, parsedTexture.renderTargetSize, scene, generateMipMaps);
+                        renderTargetTexture._waitingRenderList = parsedTexture.renderList;
+                    }
 
                     return renderTargetTexture;
                 } else {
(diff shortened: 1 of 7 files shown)


src/Probes/babylon.reflectionProbe.ts
```diff
@@ -78,7 +78,10 @@ module BABYLON {
 
                 Matrix.LookAtLHToRef(this.position, this._target, Vector3.Up(), this._viewMatrix);
 
-                scene.setTransformMatrix(this._viewMatrix, this._projectionMatrix);
+                if (scene.activeCamera) {
```
julien-moreau: Moved here as when materials are parsed, cameras can not exist for instance

deltakosh:
Checks are failing

julien-moreau:
Yes I haven't updated the what's new!
I'm still facing this issue:
- When a "cubeTexture" of a reflection probe is applied to a material for example, the function .serialize() of the RenderTargetTexture class is called. That means we loose the .position and attached mesh reference when we parse the reflection probe (parsedTexture represents the serialized data of the RenderTargetTexture)
- The only place I found to check if we should serialize reflection probe or render target texture is in .generateSerializableMember (which is not the best place ahah)

Any idea?

deltakosh:
Why not adding a section in the file for probe directly? Would be cleaner I think

julien-moreau:
Okay that looks great :)
Updating ASAP

src/Materials/Textures/babylon.texture.ts
```diff
@@ -619,8 +619,17 @@ module BABYLON {
 
                     return mirrorTexture;
                 } else if (parsedTexture.isRenderTarget) {
-                    var renderTargetTexture = new RenderTargetTexture(parsedTexture.name, parsedTexture.renderTargetSize, scene, generateMipMaps);
-                    renderTargetTexture._waitingRenderList = parsedTexture.renderList;
+                    let renderTargetTexture: Nullable<RenderTargetTexture> = null;
+                    if (parsedTexture.isCube) {
+                        const probe = ReflectionProbe.Parse(parsedTexture, scene, rootUrl);
```
deltakosh: Do we still need that? I'm a bit worried because we could save a cube without using reflection probe

src/babylon.scene.ts
```diff
@@ -3236,6 +3236,24 @@ module BABYLON {
             return index;
         }
 
+        /**
+         * Removes the given reflection probe from this scene.
+         * @param toRemove The reflection probe to remove
+         * @returns The index of the removed reflection probe
+         */
+        public removeReflectionProbe(toRemove: ReflectionProbe): number {
```
deltakosh: We need to remove any dependencies from the same so can you augment it like we did here:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Physics/babylon.physicsEngineComponent.ts

This is for the upcoming ESNext feature

julien-moreau: If I understood well, I moved these functions like the reflection probes are scene components
Kept the "AbstractScene" augmentations as it was done as this before
Am I right?

src/Materials/Textures/babylon.texture.ts
julien-moreau: Right,
So I'm now looking for an existing reflection probe instead of creating it
Other cube textures should be created/found if no reflection probe found
Is that ok for you?

deltakosh:
This looks good. Just waiting for the tests to finish!

Thanks a lot

PR #5494, opened 2018-11-10
SkyBox Fixed in FileLoader

Palmer-JC:
- also added parsed element for whether to create a PBR or Std material
for box


src/Loading/Plugins/babylon.babylonFileLoader.ts
```diff
@@ -649,7 +649,8 @@ module BABYLON {
                     if (parsedData.createDefaultSkybox === true) {
                         var skyboxScale = (scene.activeCamera !== undefined && scene.activeCamera !== null) ? (scene.activeCamera.maxZ - scene.activeCamera.minZ) / 2 : 1000;
                         var skyboxBlurLevel = parsedData.skyboxBlurLevel || 0;
-                        scene.createDefaultSkybox(undefined, true, skyboxScale, skyboxBlurLevel);
+                        var skyboxIsPBR = parsedData.skyboxIsPBR;
```
deltakosh:
```suggestion
                        var skyboxIsPBR = parsedData.skyboxIsPBR !== undefined ? arsedData.skyboxIsPBR : true;
                        scene.createDefaultSkybox(undefined, skyboxIsPBR, skyboxScale, skyboxBlurLevel);
```
undefined is equivalent to scene.environmentTexture

PR #5500, opened 2018-11-12
Fix MeshBuilder.CreatePlane using source plane

thscott:
See http://www.html5gamedevs.com/topic/41064-meshbuilder-createplane-from-sourceplane/

### DIFF
--- a/src/Mesh/babylon.meshBuilder.ts
+++ b/src/Mesh/babylon.meshBuilder.ts
@@ -694,14 +694,11 @@ module BABYLON {
             vertexData.applyToMesh(plane, options.updatable);
 
             if (options.sourcePlane) {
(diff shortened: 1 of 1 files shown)


susares:
this PR makes it impossible to create plances using sourcePlanes located at Vector3.Zero().
i.e when using Vector3.Up() as normal vector.

=>
```javascript
    planePos = new B.Vector3(0, 0, 0);
    planeN = new B.Vector3(0, 1, 0);
```
in the mentioned playground.

https://www.babylonjs-playground.com/index.html#E29R4J#8

works with the old approach but fails with the new..

only work around is slighly moving out of the origin.

deltakosh:
@bghgary @thscott can you guys have a look?

bghgary:
@susares I have a fix locally. Thanks for finding this!

PR #5511, opened 2018-11-14
Added ScrollText as GUI control

BabylonJSGuide:
Am working on documentation


gui/src/2D/controls/scrollText.ts
```diff
@@ -0,0 +1,292 @@
...
+import { Rectangle } from "./rectangle";
+import { Grid } from "./grid";
+import { Control } from "./control";
+import { TextBlock } from "./textBlock";
+import { Slider } from "./slider";
+import { ValueAndUnit } from "../valueAndUnit";
+
+/** Class used to hold the textBlock and slider in a grid
+*/
+export class ScrollText extends Rectangle {
```
deltakosh: I'm wondering if we could have a general ScrollViewer instead of a specific control only for text.
I'm not a big fan of having specific controls as they will be hard to maintain (If I add new properties to the textblock, I will also have to update this control)

BabylonJSGuide: I understand and that makes good sense but not sure **I** am up to doing it. Maybe best to put on hold. Will have a go though and let you know if I am getting anywhere. At this point I would think it would need to be more like 

```
var scrollViewer = new BABYLON.GUI.ScrollViewer();
scrollViewer.add(Control)
scrollViewer.Control.width = "200px"
```

rather than my scrollText.width setting the _textBlock width

Does that make sense?

deltakosh: Exactly! The Scrollviewer should be a pure container. So even more like that:
```
var scrollViewer = new BABYLON.GUI.ScrollViewer();
scrollViewer.add(Control)
scrollViewer.width = "200px"
```

The trick could be to provide a REALLY large measure when processing the size and then having a first step in the draw phase to remove all controls out of the current display window. I know it is a complex one but this is worth a try. I plan to work on it before release of 4.0 so do not put too much pressure on yourself

Here are random ideas i had in mind when thinking about this control:
- must be a container
- should present an "infinite" parent measure to children
- children should be offseted by the scroll position
- need a first occlusion to make sure we only render inside the scrolled view (so for instance removing controls where control.measure.left+width+scrolloffset should be checked)

BabylonJSGuide: Seems like the best idea is to close this PR. I might have another play and should I come up with anything will let you know but will leave the serious attempt to you.

deltakosh: I appreciate :)

PR #5513, opened 2018-11-15
Add support to override mesh for skeleton and link bones to transform nodes

### LINKED ISSUE #5502: Change glTF loader to group skinned meshes with the same skeleton under a common transform node
www.html5gamedevs.com/topic/41175-gltf-importer-de-parenting-skinned-meshes

The way the glTF loader works right now, there isn't an easy way to move an object with multiple skins that have the same skeleton root.

bghgary:
Fixes #5502 

Changes
* Add support to override the mesh that is used for the world matrix for a skinned mesh
* Add support to link bones to transform nodes
* Update glTF loader to use these two features so that the hierarchy is exactly as it is in the glTF and behaves like glTF

One of the nice benefits of this is that you can now use the new inspector and gizmos to move bones around.


src/Materials/babylon.materialHelper.ts
```diff
@@ -183,6 +183,10 @@ module BABYLON {
                         defines["BonesPerMesh"] = (mesh.skeleton.bones.length + 1);
                         defines["BONETEXTURE"] = materialSupportsBoneTexture ? false : undefined;
                     }
+
+                    if (mesh.skeleton.ignoreMeshTransform) {
```
deltakosh: Should be defines["IGNORETRANSFORM"] = mesh.skeleton.ignoreMeshTransform to support changing the value dynamically

bghgary: yes, that's a good point.

bghgary: no longer making changes to shader

bghgary:
@deltakosh Please review the latest changes (sorry for the force push)

bghgary:
@deltakosh This is ready to go if you approve of the changes

PR #5522, opened 2018-11-16
Fix MeshBuilder.CreatePlane when position is at origin

bghgary:
See https://github.com/BabylonJS/Babylon.js/pull/5500#issuecomment-439420992


bghgary:
I also factored out a transformNode.setDirection function from transformNode.lookAt which makes it optimal for CreatePlane.

deltakosh:
Some checks are failing

bghgary:
Fixed!

PR #5539, opened 2018-11-19
axesViewer: use scale lines parameter

TrevorDev:


src/Debug/babylon.axesViewer.ts
```diff
@@ -67,11 +67,11 @@ module BABYLON.Debug {
             this._zmesh = BABYLON.AxisDragGizmo._CreateArrow(scene, blueColoredMaterial);
 
             this._xmesh.rotationQuaternion = new BABYLON.Quaternion();
-            this._xmesh.scaling.scaleInPlace(4);
+            this._xmesh.scaling.scaleInPlace(scaleLines * 4);
```
bghgary: need to take into account scaleLines in update as well?

bghgary: or maybe change scaleLines setter to be a function and update the meshes

PR #5544, opened 2018-11-21
Fix issue with localStorage in inspector

bghgary:
This stops exceptions from being thrown when hosted in vscode.


bghgary:
Fixed conflicts

inspector/src/components/actionTabs/lineContainerComponent.tsx
```diff
@@ -8,23 +8,33 @@ interface ILineContainerComponentProps {
     closed?: boolean
 }
 
+const localStorageAvailable = function () {
```
deltakosh: why is this required? There is already a try/catch in the calling code that should take care of that. I do not want to run this test code on every click on a header

deltakosh: If you want to get rid of the typeof we could technically just remove it. The try catch already in place will make sure to turn the InMemoryStorage on

bghgary: Oh. Hmm. I didn't have your latest changes originally, hence the merge conflict, so maybe this isn't required anymore.

I am caching the boolean, so it's only called once.

I'll try it without my changes. If it works, I'll close this.

deltakosh: Ok gotcha :) I tried to do it to unblock you before leaving to Cancun :D

bghgary: Yup, it works without my changes.

PR #5547, opened 2018-11-22
Update axes viewer

bghgary:
- Support setting the renderingGroupId parameter in the constructor
- Add new createInstance function
- Update names of axes properties to better represent what they are


sebavan:
Is the constructor change not a problem regarding back compat ?

bghgary:
The constructor should behave the same as before if you don't pass additional arguments. There is a breaking change on the axis properties.

sebavan:
It should be ok as I do not think they are in use directly outside of our code (PG is returning 0 results) We ll rename back if it creates too many forum issues.

Could you log them in the readme in the breaking change section ?

bghgary:
@sebavan This is ready to go.

PR #5562, opened 2018-11-25
Inspector fixes

bghgary:
- Use flex instead of grid to avoid layout issues when hosted in vscode
- Undo canvas container when cleaning up

### DIFF
--- a/inspector/src/inspector.ts
+++ b/inspector/src/inspector.ts
@@ -88,7 +88,6 @@ export class Inspector {
             }
 
             if (!options.overlay) {
-                this._SceneExplorerHost.style.gridColumn = "1";
                 this._SceneExplorerHost.style.position = "relative";
             }
         }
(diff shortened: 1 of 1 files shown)


sebavan:
Is it still ok in PG/Sandbox/LocalDev Edge/Chrome/FF ??? (I am always scared with cross plat CSS)

bghgary:
I just checked FF. I already checked other browsers with SB and PG. They all seem okay. But I did find more issues when hosted in vscode. It's not as bad as before this change. I'm not sure why there is a difference between browser and hosted in vscode.

sebavan:
Might be related to some internal VSCode Css being in conflict with ours ???

Would be nice to check in CodePen as well as it might introduce different kind of issues ?

bghgary:
Here is a codepen: https://codepen.io/bghgary/pen/GwzMXR

Doesn't seem to exhibit any issues.

sebavan:
So I guess the last step is to try on Android and IOS :-)

deltakosh:
I'll do it

PR #5563, opened 2018-11-26
Fix solid particle bounding sphere in SPS

barroij:
Looks like something gets drop during the SPS refactoring, causing the boundingSphere of each particle to have wrong position. My bad. this PR should fix that.


barroij:
@jbousquie it shoudl fix your issue
@deltakosh FYI

jbousquie:
looks good :-)
thank you so much !

PR #5564, opened 2018-11-26
Added Mouse Wheel Scrolling to Scroll Viewer

BabylonJSGuide:


sebavan:
:-( did not see this one, I am fixing it right now.

sebavan:
Actually no, they are managed through the parent dispose :-) but is there other places we should dettach ?

gui/src/2D/controls/scrollViewer.ts
```diff
@@ -361,4 +397,28 @@ export class ScrollViewer extends Rectangle {
             this._updateScroller(windowContents);
         });
     }
+
+    /** @hidden */
+    private _attachWheel() {
+        let scene = this._host.getScene();
+        scene!.onPointerObservable.add((pi, state) => {
+            if (!this._pointerIsOver || pi.type !== BABYLON.PointerEventTypes.POINTERWHEEL) {
```
deltakosh: This is the observers I do not see disposed

sebavan: ... my bad I was looking at the constructor ones, on it

sebavan: Done And Merged, it will be in the next nightly. Thanks a ton.

PR #5571, opened 2018-11-28
change scale order in wrapInBoundingBox, remove detach camera, show …

TrevorDev:
…child elements of hidden objects in tree

http://www.html5gamedevs.com/topic/41519-bug-in-inspectordebuglayer/


inspector/src/components/sceneExplorer/entities/cameraTreeItemComponent.tsx
```diff
@@ -38,10 +38,6 @@ export class CameraTreeItemComponent extends React.Component<ICameraTreeItemComp
         const camera = this.props.camera;
         const scene = camera.getScene();
         this._onActiveCameraObserver = scene.onActiveCameraChanged.add(() => {
-            if (this.state.isActive) {
-                camera.detachControl(scene.getEngine().getRenderingCanvas()!);
```
TrevorDev: @deltakosh I removed this as this is what was causing the detach of camera drag. There seems to be no re attach code and am unsure why this is being detached in the first place. This codepath seems to get hit on every frame when scene.activeCameras has > 1 camera

Issue #3079, opened 2017-11-02
Add support for ammo.js


deltakosh:

lagauche:
+1 for Ammo support : )

D1no:
Ammo.js would just be great. The performance of wasm is outstanding and it is actively developed by the community. +1

deltakosh:
Bumping priority as latest experiences with Cannon where not satisfying

PR #5596, opened 2018-12-04
Ammo throw

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/3079

### DIFF
--- a/src/Physics/babylon.physicsImpostor.ts
+++ b/src/Physics/babylon.physicsImpostor.ts
@@ -638,7 +638,7 @@ module BABYLON {
                 this._tmpQuat.copyFrom(this.object.rotationQuaternion || new Quaternion());
             }
             if (!this._options.disableBidirectionalTransformation) {
-                this.object.rotationQuaternion && this._physicsEngine.getPhysicsPlugin().setPhysicsBodyTransformation(this, /*bInfo.boundingBox.centerWorld*/ this.object.getAbsolutePivotPoint(), this._tmpQuat);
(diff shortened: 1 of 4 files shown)


src/Physics/babylon.physicsImpostor.ts
```diff
@@ -638,7 +638,7 @@ module BABYLON {
                 this._tmpQuat.copyFrom(this.object.rotationQuaternion || new Quaternion());
             }
             if (!this._options.disableBidirectionalTransformation) {
-                this.object.rotationQuaternion && this._physicsEngine.getPhysicsPlugin().setPhysicsBodyTransformation(this, /*bInfo.boundingBox.centerWorld*/ this.object.getAbsolutePivotPoint(), this._tmpQuat);
+                this.object.rotationQuaternion && this._physicsEngine.getPhysicsPlugin().setPhysicsBodyTransformation(this, /*bInfo.boundingBox.centerWorld*/ this.object.getAbsolutePosition(), this._tmpQuat);
```
TrevorDev: @RaananW do you see any issues with this lines change?

TrevorDev: Rannan from slack:
I spent a lot of time checking pivot vs absolute positioning
The main issue, and the reason I chose pivot, was gltf files
Remember when they were trying to import the whole pool table? That's what actually fixed it.
And since they are identical 99% of the time (and all of the scenes I checked were working great with the pivot position), I left it this way
So I'm fine with this change

PR #5597, opened 2018-12-04
Adding way to add camera offset in sky material to keep horizon at a …

julien-moreau:
…specific position relative to the camera


deltakosh:
Waiting for comments before merging (as seen ofline :))

julien-moreau:
Sure :)

PR #5598, opened 2018-12-04
Fix setDirection

deltakosh:


deltakosh:
@TrevorDev  and @bghgary: I think I found the culprit. Now the axes, the gizmos and the look at work

PR #5608, opened 2018-12-07
fix for border render

ssaket:
Fix for forum link - http://www.html5gamedevs.com/topic/41694-general-gui-troubles/?tab=comments#comment-237278

### DIFF
--- a/gui/src/2D/controls/inputText.ts
+++ b/gui/src/2D/controls/inputText.ts
@@ -923,19 +923,17 @@ export class InputText extends Control implements IFocusableControl {
                 context.fillRect(highlightCursorLeft, this._currentMeasure.top + (this._currentMeasure.height - this._fontOffset.height) / 2, width, this._fontOffset.height);
                 context.globalAlpha = 1.0;
             }
-
-            context.restore();
-
-            // Border
-            if (this._thickness) {
-                if (this.color) {
-                    context.strokeStyle = this.color;
-                }
-                context.lineWidth = this._thickness;
-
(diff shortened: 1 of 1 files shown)


gui/src/2D/controls/inputText.ts
```diff
@@ -923,19 +923,17 @@ export class InputText extends Control implements IFocusableControl {
...
-            if (this._thickness) {
-                if (this.color) {
-                    context.strokeStyle = this.color;
-                }
-                context.lineWidth = this._thickness;
-
-                context.strokeRect(this._currentMeasure.left + this._thickness / 2, this._currentMeasure.top + this._thickness / 2,
-                    this._currentMeasure.width - this._thickness, this._currentMeasure.height - this._thickness);
+        }
+        context.restore();
```
deltakosh: This is probably a bug. The restore should be inside the if focused as the associated save is also inside a if focused.

You can test it with the first pg from the initial thread on the forum

ssaket:
Please don't merge, working on adding a new property onFocusBorderColor, which by default should be "blue".

gui/src/2D/controls/inputText.ts
ssaket: if I put it inside it creates incomplete border

ssaket: screenshot 2018-12-07 at 8 10 52 pm
for PG - https://www.babylonjs-playground.com/#YP180Z
@deltakosh - I don't know the cause yet.

ssaket:
@deltakosh , Sorry not sure about the context handling, Therefore, closing this pull request.

deltakosh:
Your fix was good. Just let the restore inside the if :)

PR #5610, opened 2018-12-07
Returns line index when the pickedMesh is a LinesMesh

barroij:
When Picking occurs on a classical triangle Mesh, the index of the face is returned as part of the `PickingInfo`

This PR fills the `faceId` property with the line index when the picked mesh is LinesMesh


barroij:
@deltakosh 
I hesitated between using the `faceId` property and adding a new `lineId` property in the `PickingInfo`. I chose the one that minimized the change, but if you think it can break some code, I can add a `lineId` property in the `PickingInfo`

deltakosh:
Seems fine like that :)

PR #5619, opened 2018-12-10
added input type as number

ssaket:
Inspector P2 - Better control for number input (something like up/down that we can manipulate with mouse)
Ref - https://www.w3.org/TR/2010/WD-html-markup-20101019/input.number.html
Browser support ->https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Browser_compatibility

### DIFF
--- a/inspector/src/components/actionTabs/lines/floatLineComponent.tsx
+++ b/inspector/src/components/actionTabs/lines/floatLineComponent.tsx
@@ -109,7 +109,7 @@ export class FloatLineComponent extends React.Component<IFloatLineComponentProps
                     {this.props.label}
                 </div>
                 <div className="value">
-                    <input className="numeric-input" value={this.state.value} onBlur={() => this.unlock()} onFocus={() => this.lock()} onChange={evt => this.updateValue(evt.target.value)} />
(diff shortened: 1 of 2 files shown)


deltakosh:
Well I used to have that before but I had to remove it because of negative numbers (also can you check if you can manually enter a value?)

ssaket:
Hmm.. but position can be negative right ? and for alpha I can add min="0" to avoid negative values. @deltakosh did I miss anything ?

deltakosh:
Well I can test if you want but when I used input type=number I was not able to enter negative value manually (using the keyboard)

ssaket:
yeah, solution would to add [min property](https://www.w3.org/TR/2010/WD-html-markup-20101019/input.number.html) in code, but I don't want to generalise it as for some other control, a negative value would be a valid option. That's why having separate checks is better :)

deltakosh:
ok but then this won't work for now as if I want to specify a negative number then I will not let me do it

ssaket:
I haven't added the "min" property so it will work :) 
Regarding entering -ve values manually I can say -
I can't check for edge but it's working on rest 3.
ice_video_20181211004533

deltakosh:
No thank you!

PR #5622, opened 2018-12-11
right handed and uvs - proposal

deltakosh:


deltakosh:
Here is an idea (that works for sphere and box)

@bghgary  and @kcoley: Thoughts?

deltakosh:
Well no this is wrong and just a bad patch..we should rework all the equations for all shapes.  Not sure we want to invest in it for now:)

PR #5624, opened 2018-12-11
Feature.culling strategy

jbousquie:
Now a culling strategy can be set per mesh :  
* Standard : the more accurate and standard one (exclusion test)  
* Bounding Sphere Only : faster but less accurate (exclusion test)  
* Optimistic Inclusion : mesh center inclusion test then standard exclusion test, for meshes almost always expected in the frustum. Same accuracy than the standard test.  
* Optimistic Inclusion Then bSphere Only : mesh center inclusion test, then bounding sphere exclusion test only. Same accuracy than the bSphereOnly test, interesting for almost always in the frustum meshes.

### DIFF
--- a/src/Culling/babylon.boundingInfo.ts
+++ b/src/Culling/babylon.boundingInfo.ts
@@ -146,21 +146,32 @@ module BABYLON {
         }
 
         /**
-         * Returns `true` if the bounding info is within the frustum defined by the passed array of planes.
-         * @param frustumPlanes defines the frustum to test
-         * @param strategy defines the strategy to use for the culling (default is BABYLON.Scene.CULLINGSTRATEGY_STANDARD)
-         * @returns true if the bounding info is in the frustum planes
+         * Returns `true` if the bounding info is within the frustum defined by the passed array of planes.  
+         * @param frustumPlanes defines the frustum to test  
+         * @param strategy defines the strategy to use for the culling (default is BABYLON.Scene.CULLINGSTRATEGY_BOUNDINGSPHERE_ONLY)  
+         * @returns true if the bounding info is in the frustum planes  
          */
         public isInFrustum(frustumPlanes: Array<DeepImmutable<Plane>>, strategy: number = AbstractMesh.CULLINGSTRATEGY_STANDARD): boolean {
+            let inclusionTest = (strategy === AbstractMesh.CULLINGSTRATEGY_OPTIMISTIC_INCLUSION || strategy === AbstractMesh.CULLINGSTRATEGY_OPTIMISTIC_INCLUSION_THEN_BSPHERE_ONLY);
+            let bSphereOnlyTest = (strategy === AbstractMesh.CULLINGSTRATEGY_BOUNDINGSPHERE_ONLY || strategy === AbstractMesh.CULLINGSTRATEGY_OPTIMISTIC_INCLUSION_THEN_BSPHERE_ONLY);
+            
+            if (inclusionTest) {
+                if (this.boundingSphere.isCenterInFrustum(frustumPlanes)) {
+                    return true;
+                }
+            }
+            
             if (!this.boundingSphere.isInFrustum(frustumPlanes)) {
                 return false;
             }
 
-            if (strategy === AbstractMesh.CULLINGSTRATEGY_BOUNDINGSPHERE_ONLY) {
+            if (bSphereOnlyTest) {
                 return true;
             }
+
             return this.boundingBox.isInFrustum(frustumPlanes);
         }
+
 
         /**
 		 * Gets the world distance between the min and max points of the bounding box
--- a/src/Culling/babylon.boundingInfo.ts
+++ b/src/Culling/babylon.boundingInfo.ts
@@ -146,10 +146,10 @@ module BABYLON {
         }
 
         /**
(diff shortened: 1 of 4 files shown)


src/Culling/babylon.boundingInfo.ts
```diff
@@ -146,22 +146,33 @@ module BABYLON {
         }
 
         /**
-         * Returns `true` if the bounding info is within the frustum defined by the passed array of planes.
-         * @param frustumPlanes defines the frustum to test
-         * @param strategy defines the strategy to use for the culling (default is BABYLON.Scene.CULLINGSTRATEGY_STANDARD)
-         * @returns true if the bounding info is in the frustum planes
+         * Returns `true` if the bounding info is within the frustum defined by the passed array of planes.  
+         * @param frustumPlanes defines the frustum to test  
+         * @param strategy defines the strategy to use for the culling (default is BABYLON.Scene.CULLINGSTRATEGY_BOUNDINGSPHERE_ONLY)
```
deltakosh: This is a change (unless there is no visual impact and only improvements?)

src/Culling/babylon.boundingInfo.ts
```diff
@@ -146,22 +146,33 @@ module BABYLON {
...
-         * @param strategy defines the strategy to use for the culling (default is BABYLON.Scene.CULLINGSTRATEGY_STANDARD)
-         * @returns true if the bounding info is in the frustum planes
+         * Returns `true` if the bounding info is within the frustum defined by the passed array of planes.  
+         * @param frustumPlanes defines the frustum to test  
+         * @param strategy defines the strategy to use for the culling (default is BABYLON.Scene.CULLINGSTRATEGY_BOUNDINGSPHERE_ONLY)  
+         * @returns true if the bounding info is in the frustum planes  
          */
         public isInFrustum(frustumPlanes: Array<DeepImmutable<Plane>>, strategy: number = AbstractMesh.CULLINGSTRATEGY_STANDARD): boolean {
+            let inclusionTest = (strategy === AbstractMesh.CULLINGSTRATEGY_OPTIMISTIC_INCLUSION || strategy === AbstractMesh.CULLINGSTRATEGY_OPTIMISTIC_INCLUSION_THEN_BSPHERE_ONLY);
+            let bSphereOnlyTest = (strategy === AbstractMesh.CULLINGSTRATEGY_BOUNDINGSPHERE_ONLY || strategy === AbstractMesh.CULLINGSTRATEGY_OPTIMISTIC_INCLUSION_THEN_BSPHERE_ONLY);
```
deltakosh: can be moved closer to the use (if inclusion test then no need to even evaluate it)

src/Culling/babylon.boundingSphere.ts
```diff
@@ -105,12 +105,29 @@ module BABYLON {
...
+            for (let i = 0; i < 6; i++) {
+                if (frustumPlanes[i].dotCoordinate(center) <= -radius) {
                     return false;
                 }
             }
+            return true;
+        }
 
+        /**
+         * Tests if the bounding sphere center is inbetween the frustum planes.
```
deltakosh: typo
```suggestion
         * Tests if the bounding sphere center is in between the frustum planes.
```

src/Culling/babylon.boundingInfo.ts
jbousquie: only improvements, everythings runs like it used to... sometimes faster :-)

deltakosh: But it is less accurate so you can end up having objects selected whereas they would have not with the regular strategy
So I would prefer to stick with the regular one as the default

src/Culling/babylon.boundingSphere.ts
jbousquie: mmhh... the right english spelling is really the adverb "inbetween"

src/Culling/babylon.boundingInfo.ts
jbousquie: ok, I'll change this.

src/Culling/babylon.boundingInfo.ts
jbousquie: ooopps... actually nothing is changed in the process nor in the algorithm.
 The error is only in the comments : wrong copy/paste. 
The real default behavior is still the standard strategy.

I'll also change the comments about the static vars CULLINGSTRATEGY because there's actually no perceptible loss at all when using the Optimistic modes, only some possible little gains when including immediately.

src/Culling/babylon.boundingSphere.ts
jbousquie: oops, no I'm wrong and you're right : https://en.wiktionary.org/wiki/inbetween

jbousquie:
all requested changes done

deltakosh:
We have several fails on the styling :)
Please use gulp --max-old-space-size=8192 --tsLintFix to fix them :)

src/Culling/babylon.boundingSphere.ts
deltakosh: :)

jbousquie: rhhaa beurre d'aile, je l'ai ratée celle-ci
mékékon

jbousquie:
typo fixed + linter passed
I hope this one is the right one
_(pinailleur :-)  )_

PR #5633, opened 2018-12-12
Fix #5631: Call onError even when fallback is loaded

### LINKED ISSUE #5631: Creating a texture for an image that returns a 404 doesn't report using onError callback
Code: https://github.com/BabylonJS/Babylon.js/blob/master/src/Engine/babylon.engine.ts#L4324-L4332

It goes into the if block for `UseFallbackTexture`, which returns. onError call is after this.

haroldma:


haroldma:
Remove `return` and remove pass of onError to the callback (avoid calling more than once)

deltakosh:
Thanks a lot

PR #5634, opened 2018-12-12
KhronosTextureContainer: Use DataView to read headers & headers should be Uint32

06wj:
According to the [KTX specification](https://www.khronos.org/opengles/sdk/tools/KTX/file_format_spec/)
* Use ```DataView``` to read header data, ```DataView``` native support for ```bigEndian``` and ```littleEndian``` data reading.
* The header data should be ```Uint32```.

### DIFF
--- a/src/Tools/babylon.khronosTextureContainer.ts
+++ b/src/Tools/babylon.khronosTextureContainer.ts
@@ -91,25 +91,24 @@ module BABYLON {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


deltakosh:
makes sense! thanks a lot! Do you have some ktx files to test?

06wj:
@deltakosh https://github.com/Palmer-JC/Palmer-JC.github.io/tree/master/scenes/compressedTextures/images

PR #5635, opened 2018-12-12
use uniqueId as key for geometryMap in scene

barroij:
Since https://github.com/BabylonJS/Babylon.js/pull/5476, when a geometry is pushed, we used `uniqueId` instead of `ìd` to check whether a geometry is already here.

As the optional Map I introduced in https://github.com/BabylonJS/Babylon.js/pull/5272 was for speeding up the push of geometry in the scene when the number of geometries becomes large, this PR updates the map to use `uniqueId` too, so that we can get the performance improvement back.


barroij:
@deltakosh another small update.
The use of `uniqueId` instead of `id` comes from @TrevorDev

src/babylon.scene.ts
```diff
@@ -66,7 +66,7 @@ module BABYLON {
     /** Interface defining initialization parameters for Scene class */
     export interface SceneOptions {
         /**
-         * Defines that scene should keep up-to-date a map of geometry to enable fast look-up by Id
+         * Defines that scene should keep up-to-date a map of geometry to enable fast look-up by uniqueId
          * It will improve performance when the number of geometries becomes important.
          */
         useGeometryIdsMap?: boolean;
```
barroij: I did not change the option parameter name, to preserve backward compatibilty. Should I ?
I updated the comment though.

barroij:
By the way, we used to do
`if (!force && this.getGeometryByID(geometry.id)) `, which I can undestand why.
Now we do
`if (!force && this._getGeometryByUniqueID(geometry.uniqueId)) `

but the only way to get two geometries with the same uniqueId is that a geometry is pushed twice in the scene. Does it really happen ?  If we want to enable two geometry with the same `id` in the scene, we could just set `force` to true.

src/babylon.scene.ts
deltakosh: Please change it. It was introduced during current cycle so no problem with breaking changes yet

deltakosh:
I'll do the change it is easier ;)

PR #5642, opened 2018-12-13
let every material have it's own reflection color

tauli:
Currently every Material gets the same color object as reflection color. I consider this a bug since changing the value of one color changes the colors of all materials.

### DIFF
--- a/Viewer/src/model/viewerModel.ts
+++ b/Viewer/src/model/viewerModel.ts
@@ -557,7 +557,7 @@ export class ViewerModel implements IDisposable {
                 material.disableLighting = !this._modelConfiguration.material.directEnabled;
             }
             if (this._configurationContainer && this._configurationContainer.reflectionColor) {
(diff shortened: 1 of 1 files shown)


TrevorDev:
@tauli do you have sample code where you use this? I took a quick look but and am wondering if you should be creating a new material yourself instead of cloning here.

tauli:
I'm not trying to clone a material, but just the color object. If i load a model into the viewer and try to change a color component of the reflection color with something like `scene.materials.find(elem => elem.name == "Somematerial").reflectionColor.r = 1;` then this color changes for all materials since they all share the same object. This is quite unexpected and not the case for any other color attribute i've used so far or if use just plain Babylon.js.

On a related note, i'm not quite sure when this code is executed. I'm trying to change some values for materials when loading a .glb which i can not encode in that format. Currently i'm using the onModelLoadedObservable hook for that. But it seems that some attributes like reflectionColor are overwritten after that code gets executed.

PR #5643, opened 2018-12-13
Avoid geometry copy error

Eole211:
Some meshes can have some weird vertexBuffer with no vertexData associated to, like the one imported from the obj file format : world0, world1...
This MR avoid to get an error copying the geometry and just keep copying vertexBuffers with actual vertex data


deltakosh:
Thanks!

Issue #5611, opened 2018-12-07
Add support for InvalidateRect optimization


deltakosh:
Idea would be to improve GUI performance by using the InvalidateRect approach (used by GDI) so that only the portion of the texture that has changed will be redrawn

PR #5651, opened 2018-12-14
Invalidate rect support in ADT

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/5611


deltakosh:
ok to merge? Did you run tests using samples from the doc?

TrevorDev:
@deltakosh  Yep, this should be good, I tested about 10 PGs from https://doc.babylonjs.com/how_to/gui and they all worked and fixed the one test that failed.

PR #5668, opened 2018-12-18
[WIP] stop generating Vector[] when picking

barroij:


src/Mesh/babylon.abstractMesh.ts
```diff
@@ -1348,8 +1347,11 @@ module BABYLON {
                 subMesh._trianglePlanes = [];
                 var start = subMesh.verticesStart;
                 var end = (subMesh.verticesStart + subMesh.verticesCount);
+
                 for (var i = start; i < end; i++) {
-                    subMesh._lastColliderWorldVertices.push(Vector3.TransformCoordinates(this._positions[i], transformMatrix));
+                    const v = Vector3.FromFloatArray(positionBuffer, 3*i)
```
deltakosh: This will overloads the GC with new Vector3 on every frame The goal of the pointsarray is to avoid instancing new Vector3 on each frame

src/Mesh/babylon.subMesh.ts
```diff
@@ -353,13 +355,15 @@ module BABYLON {
...
+        private _intersectLines(ray: Ray, positions: FloatArray, indices: IndicesArray, intersectionThreshold: number, fastCheck?: boolean): Nullable<IntersectionInfo> {
             var intersectInfo: Nullable<IntersectionInfo> = null;
+            const p0 = SubMesh.TmpVector3[0]
+            const p1 = SubMesh.TmpVector3[1]
 
             // Line test
             for (var index = this.indexStart; index < this.indexStart + this.indexCount; index += 2) {
-                var p0 = positions[indices[index]];
-                var p1 = positions[indices[index + 1]];
+                Vector3.FromArrayToRef(positions, 3*indices[index], p0)
```
deltakosh: My only concern is the cost of this change as we new to copy 6 values per face (and 9 for triangles later).

Can you do a benchmark to make sure we are not slower than before?

barroij: Well we'll probably be a bit slower than before as we do more operations. I'll try to measure how much slower so that we can weight the trade-off between CPU "loss" Vs Memory "gain". My gut feeling is that it should not be that slower.

@jbousquie, it seems you've got already a lot of scenes/test cases, is there one that might be used for benchmarking the picking ? @deltakosh Do you have something already ? If no I'll build one

jbousquie:
unfortunately, I have none about the picking :-(

Le 18/12/2018 à 17:12, Julien Barrois a écrit :

src/Mesh/babylon.subMesh.ts
deltakosh: I don't out of the box unfortunately

deltakosh: up?

barroij: down 😞 
Pretty busy these days, did not find time so far to test.

deltakosh: Ok  no worry  
Ping me when you have a chance to do it

deltakosh:
Should we keep this PR open?

barroij:
Let's close it for now, I'll reopen it when I write test

deltakosh:
Sounds good

PR #5692, opened 2019-01-04
Fix #5691 - After loading a scene, animations don't run because of additional data being loaded

### LINKED ISSUE #5691: After loading a scene, animations don't run because of additional data being loaded
## Scenario:
We load a scene file, once done we hide the loading screen and start loading three objects and we show a loading indicator for them (custom).

The loading spinners don't animate until everything finishes loading. After this you can add new content and the spinner will work.

## Reason:
[This block of code](https://github.com/BabylonJS/Babylon.js/blob/d3763662bb852fc010f06f469e7f196fad24d3c0/src/babylon.scene.ts#L2663-L2668).

`this._pendingData.length > 0` is true because we're loading additional data, therefore `_animationTimeLast` is not set.

## Our workaround:
Running this when we load our scene files and hide the loading screen:
`(scene as any)._animationTimeLast = BABYLON.Tools.Now;`

## Proposal:
A way to mark the scene as ready to animate when we load the scenes assets. Allowing us to load additional data that could be represented with animating spinners.

haroldma:


src/babylon.scene.ts
```diff
@@ -2653,6 +2653,14 @@ module BABYLON {
             }
         }
 
+        /**
+         * Resets the last animation time frame.
+         * Useful to override when animations start running when loading a scene for the first time.
+         */
+        public resetLastAnimationTimeFrame(): void {
+            this._animationTimeLast = BABYLON.Tools.Now;
```
bghgary: Did you perhaps mean this? `BABYLON.Tools.Now` is a function.

```suggestion
            this._animationTimeLast = BABYLON.Tools.Now();
```

deltakosh:
I'll merge it actually so it will be in next npm update

PR #5695, opened 2019-01-05
Fix case sensitive paths

mrdunk:
After following the instructions as https://doc.babylonjs.com/how_to/how_to_start
i experienced many issues relating to the case of paths when opening the links to the main Babylon entry points.
eg: http://localhost:1338/Playground/index-local.html

This pull request fixes as many of these as i could find.

I'm running on 64bit Linux.

--edit--
Re-reading this i realise it's not particularly clear what my problem is:
When building Babylon on Linux,
many of the test resources (eg: Playground) will not correctly import the required libraries due to many incorrectly specified path names.
The diff makes it clearer but in summary, on most OS-es, "Test" is not the same as "test". "Playground" is not the same as "playground".


deltakosh:
This pull request has been mentioned on **Babylon.js**. There might be relevant details there:

https://forum.babylonjs.com/t/failures-using-build-environment-resources-due-to-case-sensitive-paths/593/1

deltakosh:
This makes sense! and thanks for that (I mostly use Windows so they are issues I do not get)

PR #5697, opened 2019-01-06
Add support for color vertices in OBJ file (not part of standard)

brianzinn:
Started as a forum feature request: https://forum.babylonjs.com/t/import-vertex-colored-meshes-with-obj-files/545/11

I have issues with current PR:
1. Are we going for backwards compatibility or sensible defaults, because otherwise using available color vertices would seem more like an opt-out to me.
2. The OBJ file from the forum request has no normals, so I added an option to generate normals.  I'm happy to remove this as it was more to verify the sandbox would work like MeshLab.
3. Can you verify this is where tests belong?  I was unable to access the static member flags on ObjFileLoader.

I think the tougher question to address is why we are doing static flags like this.  These flags are read when the object model is being parsed.  If I want two models to download simultaneously with different flags then the last set flags will be used from this race condition.  I was going to add some context for parsing, but the mesh loader is called *after* the download has completed.  One option is that the loaders themselves can create a context, which is a capture of current set flags.  That would require augmenting the loader interface.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -88,6 +88,9 @@
 - Enable dragging in boundingBoxGizmo without needing a parent ([TrevorDev](https://github.com/TrevorDev))
 - Added per mesh culling strategy ([jerome](https://github.com/jbousquie))
 
+### OBJ Loader
+- Add color vertex support (not part of standard) ([brianzinn](https://github.com/brianzinn))
+
 ### glTF Loader
 
 - Added support for mesh instancing for improved performance when multiple nodes point to the same mesh ([bghgary](https://github.com/bghgary))
--- a/loaders/src/OBJ/babylon.objFileLoader.ts
+++ b/loaders/src/OBJ/babylon.objFileLoader.ts
@@ -201,10 +201,31 @@ module BABYLON {
         }
     }
 
+    type MeshObject = {
+        name: string;
+        indices?: Array<number>;
+        positions?: Array<number>;
+        normals?: Array<number>;
+        colors?: Array<number>;
+        uvs?: Array<number>;
+        materialName: string;
+    }
+
     export class OBJFileLoader implements ISceneLoaderPluginAsync {
 
         public static OPTIMIZE_WITH_UV = false;
+        /**
+         * Invert model on y-axis (does a model scaling inversion)
+         */
         public static INVERT_Y = false;
+        /**
+         * Include in meshes the vertex colors available in some OBJ files.  This is not part of OBJ standard.
+         */
+        public static IMPORT_VERTEX_COLORS = false;
+        /**
+         * Compute the normals for the model, even if normals are present in the file
+         */
+        public static COMPUTE_NORMALS = false;
         public name = "obj";
         public extensions = ".obj";
         public obj = /^o/;
@@ -212,7 +233,7 @@ module BABYLON {
         public mtllib = /^mtllib /;
         public usemtl = /^usemtl /;
         public smooth = /^s /;
-        public vertexPattern = /v( +[\d|\.|\+|\-|e|E]+)( +[\d|\.|\+|\-|e|E]+)( +[\d|\.|\+|\-|e|E]+)/;
+        public vertexPattern = /v( +[\d|\.|\+|\-|e|E]+){3,7}/;
         // vn float float float
         public normalPattern = /vn( +[\d|\.|\+|\-|e|E]+)( +[\d|\.|\+|\-|e|E]+)( +[\d|\.|\+|\-|e|E]+)/;
         // vt float float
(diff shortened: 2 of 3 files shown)


brianzinn:
@bghgary ready to discuss backwards compatibility and the flags from forum post.

deltakosh:
This pull request has been mentioned on **Babylon.js**. There might be relevant details there:

https://forum.babylonjs.com/t/import-vertex-colored-meshes-with-obj-files/545/12

deltakosh:
I would turn off the normal generation by default. Same for colors. We need to make sure people loading their obj do not get any difference

Excellent work by the way

brianzinn:
> I would turn off the normal generation by default. Same for colors.
Great.  That answers the first 2 questions.  All the flags are set to `false`, so the PR code should be 100% backwards compatible and have no effect - until a flag is changed by the application developer.

deltakosh:
So far regarding the static I'm not too overly concerned as it is a edge case.
For the tests did you make sure to have a reference to the loader?

We will need a quick update of the obj loader as well!

Thanks a lot :) and glad to see you back to contributing  :D

bghgary:
> For the tests did you make sure to have a reference to the loader?

I think the tests pull in everything, so no need to add a reference. Plus, it looks like CI has passed.

> Can you verify this is where tests belong? I was unable to access the static member flags on ObjFileLoader.

Yes, this is the right place to add this test.

> One option is that the loaders themselves can create a context, which is a capture of current set flags.

This is exactly what happens for the glTF loader which has a bunch of options as well. You can use SceneLoader.OnPluginActivatedObservable to gain access to the loader in question to set flags.

brianzinn:
Thanks @bghgary for the hints.  I will look tonight.

@deltakosh can you clarify on 
> We will need a quick update of the obj loader as well!

I will prepare tests with new flags.  Cheers.

brianzinn:
@bghgary I have looked at usage of `SceneLoader.OnPluginActivatedObservable`.  Unless I have missed something that does not appear to be able to achieve what I am after.  Were you referring to your usage of GlobalState from Inspector?  I need something available with just `import babylonjs-loaders` :)

Additionally, the SceneLoader notification to observers doesn't contain any context specific to the download:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Loading/babylon.sceneLoader.ts#L364

I was more thinking along the lines that near that call, we could have something like:
```javascript
const pluginConfigurationContext: Map<String, String> = plugin.getCurrentConfiguration();
```
We could capture the current plugin static settings before download starts and flow them through as an additional parameter in onSuccess() - for ie: mesh parsing.  As @deltakosh is not overly concerned about this edge case, I think it's best to leave changes like that out of this PR.  I think I spend too much time writing multi-threaded code that I worry too much about edge cases!! :)

bghgary:
@brianzinn 
It's more like this:
```javascript
SceneLoader.OnPluginActivatedObservable.addOnce(function (plugin) {
    if (plugin.name === "obj") {
        // plugin will be the instance of objFileLoader that was created
        // do whatever you need to in order to set the options
    }
});

BABYLON.SceneLoader.Load(/*...*/);
```

> leave changes like that out of this PR

I am totally okay if you don't this in this PR or at all if you don't want to. I just want to let you know how it can be done. 😉

brianzinn:
@bghgary -- I see that I can implement `ISceneLoaderPluginFactory` and then set instance properties via OnPluginActivatedObservable.  I am talking about a developer building an application:
```javascript
import 'babylonjs-loaders'

// these will both be parsed with INVERT_Y = false
ObjFileLoader.INVERT_Y = true;
SceneLoader.Load(..., 'file1.obj'...);
ObjFileLoader.INVERT_Y = false;
SceneLoader.Load(..., 'file2.obj'...);
```

It looks to me like you are suggesting what has been done for "gltf" in the Sandbox and Inspector, while I need something available for above code sample.  ie: where do **we** put `SceneLoader.OnPluginActivatedObservable.add(...)`?  Even in `SceneLoader.Register(...)` there is no opportunity.  I think the only place is in ObjFileLoader implementation of `ISceneLoaderPluginFactory.createPlugin()`.

deltakosh:
Merging in a few minutes unless someone screams out loud :)

brianzinn:
go for it.  i will update the docs. thanks.

deltakosh:
thanks a lot!

bghgary:
> @bghgary -- I see that I can implement ISceneLoaderPluginFactory and then set instance properties via OnPluginActivatedObservable.

Yes, this is how you would implement the plugin.

> I am talking about a developer building an application...

I'm not sure what you mean. As a user of the plugin, if you need to set different options for different calls to Load, you would do something like this:
```javascript
SceneLoader.OnPluginActivatedObservable.addOnce(function (plugin) {
    plugin.invertY = true;
});
SceneLoader.Load(..., 'file1.obj'...);

SceneLoader.OnPluginActivatedObservable.addOnce(function (plugin) {
    plugin.invertY = false;
});
SceneLoader.Load(..., 'file2.obj'...);
```

It's not necessarily the best interface. I have wanted to improve this for glTF loader.

brianzinn:
Thanks @bghgary for the follow up and further explanation!  I just wanted to make sure I wasn't missing something obvious that we could implement to handle this automatically.  From what I've seen and without making changes external to the plugin, the createPlugin() would be the best spot for this.  Cheers.

PR #5698, opened 2019-01-07
Change upperBetaLimit to avoid weird behaviours when beta is PI

a-carvallo:
Found out that ArcRotateCamera was doing weird things when beta was exactly PI :

https://www.babylonjs-playground.com/#6D0950 (zoom in/out)

In the doc it says that it is offset by 0.1 to avoid issues (https://doc.babylonjs.com/babylon101/cameras#arc-rotate-camera) but in code it's 0.01 (I have no problem with that) but more concerning, the offset was not applied for upperBetaLimit.


deltakosh:
Good catch ;)

PR #5702, opened 2019-01-08
Added InputsManager and keyboard bindings for FollowCamera.

mrdunk:
The `FollowCamera` currently has no `InputsManager`.
This change fixes that and is largely based on `ArcRotateCameraInputsManager` and it's `InputsManager`.


deltakosh:
This seems good but I need to hold this PR back for now as we will merge a gigantic PR tomorrow that may force you to do some tiny changes :)
We will introduce full ESNext modules tomorrow so stay tuned

deltakosh:
Well actually we need a bit more time so I'll merge your code
Thanks!

Tools/Gulp/config.json
```diff
@@ -46,6 +46,7 @@
             "freeCamera",
             "flyCamera",
             "arcRotateCamera",
+            "followCamera",
```
deltakosh: You have to remove it from the additionalCamera entry

deltakosh: I'll fix it for you :)

deltakosh: thanks :D

PR #5704, opened 2019-01-08
added option forceDisposeChildren to multiMaterial.dispose

danjpar:
Return an array of multi-materials that a single material is used in.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -89,6 +89,7 @@
 - Added per mesh culling strategy ([jerome](https://github.com/jbousquie))
 - Added InputsManager and keyboard bindings for FollowCamera. ([mrdunk](https://github.com))
 - Added per solid particle culling possibility : `solidParticle.isInFrustum()`  ([jerome](https://github.com/jbousquie))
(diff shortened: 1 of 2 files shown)


danjpar:
Sorry, that was my first ever pull request on a major repo.
Yes. Ultimately, I just want to know when I'm disposing of a multi material
I can iterate over its submaterials to check if they're used elsewhere and
dispose of them and their textures accordingly. Disposing of multi material
itself leaves all materials and textures referenced within the scene.

deltakosh:
No need to apologise:)
So you may prefer having an option on the multi-material.dispose to make sure to dispose its children as well?

danjpar:
only if they are absolutely unused in other multi materials and meshes. thank you !

deltakosh:
Far better :) 
I'm not sure we need the getBindedMaterials to be fair as this is easy to do with current API

Also please remove all the dist/ files from the PR

deltakosh:
This is perfect. Just waiting for the build to finish and I will merge

deltakosh:
Gonna merge manually

PR #5723, opened 2019-01-10
fixed depth on diffuse textures with alpha in GeometryRenderer

### LINKED ISSUE #5690: Using SSAO2 pipeline with hasAlpha issue
When using SSOA2 Pipeline, the SSAO rendering isn't gonna consider objects with hasAlpha as having the right depth in the z-buffer, resulting in annomalies in the rendering.

Here is an example : https://www.babylonjs-playground.com/#LFVESY

image

CraigFeldspar:
Closes #5690


deltakosh:
thanks!!

Issue #4154, opened 2018-04-18
Allow providing tessellation value when creating a cylinder impostor


tommhuth:
When using the cylinder impostor, tessellation is set to a fixed 16, so everything behaves as a "smooth" cylinder. Is there a reason for this? Seems like an oversight if Cannonjs can handle any number of sides.

tommhuth:
that would be great! Could the impostor just automatically use the tessellation from the mesh?

deltakosh:
Unfortunately we have no tesselation info on the mesh (unless for very specific cases when you create a cylinder mesh)
Do you want to submit a PR?

tommhuth:
I'd love to contribute here, but not sure if I'm cut of for this. I'll give it a go tho.

I'm just assuming the shape of the Cannonjs impostor is constructed the same as the mesh, so that any tessellation ends up the same (corners placed in the same place if I'm making any sense). If that's the case it could be as easy as supplying the tessellation as parameter. 

How should a make a PR for this? I could not get `gulp run` up and running, but I can confirm that replacing `16` with  `impostor.getParam("tessellation") || 16` works locally for me in a simple scene.

tommhuth:
Any change of getting this out in the next release (3.2)?

deltakosh:
Can you let me know in which file/line you want to add this change?

tommhuth:
babylon.cannonJSPlugin.js, line 258 :)

deltakosh:
So after discussing with the team the main issue is that we do not store the tesselation in the mesh after calling createCylinder.
Perhaps in a future update we could  update the API to provide a custom tesselation when building the cylinder impostor

tommhuth:
any chance of getting this in for the 4 release? This would be super usefull and seems like a trivial thing to fix. not all cylinders need to be perfectly round :)

deltakosh:
That's a question for @RaananW

RaananW:
I did some thinking about this - if you want to monitor the tessellation yourself and provide it when creating the impostor, we can get that to work if you add it to the nativeOptions key. It is not configured like that at the moment thou.

Reopening this, but changing the title :)

tommhuth:
@RaananW are the `radiusTop` and `radiusBottom` cannonjs props also exposed thru Babylon?

RaananW:
[...]

PR #5724, opened 2019-01-10
Allow setting the native cannonjs parameters for cylinder, overriding defaults

RaananW:
Resolving #4154
Adding radiusTop, radiusBottom, height or numSegments to the nativeOptions of the impostor will use those values instead of the default Babylon-defined cylinder parameters.


deltakosh:
Thanks!!

PR #5745, opened 2019-01-13
Follow camera mouse wheel input

mrdunk:
Tested on Chrome and Firefox under Linux.
I'm reasonably confident this will work on other browsers.


deltakosh:
Thanks!

PR #5755, opened 2019-01-14
Removed action dependency from scene and meshes

deltakosh:


src/Actions/abstractActionManager.ts
```diff
@@ -0,0 +1,218 @@
...
+ * Abstract class used to decouple action Manager from scene and meshes.
+ * Do not instantiate.
+ * @see http://doc.babylonjs.com/how_to/how_to_use_actions
+ */
+export abstract class AbstractActionManager implements IDisposable {
+    /**
+       * Nothing
+       * @see http://doc.babylonjs.com/how_to/how_to_use_actions#triggers
+       */
+    public static readonly NothingTrigger = Constants.ACTION_NothingTrigger;
```
sebavan: Would be nice to keep it constants only in action manager

PR #5789, opened 2019-01-17
Added multiMultiMaterials option to MergeMeshes

danjpar:


danjpar:
Couple things I'm considering changing:
If no material is present for a mesh it uses the first material. I am thinking of changing it to default to the scene.defaultMaterial.
If there are meshes that share the same material, all materials are still added in the new multimaterial. I should have it check if the material already exists then reference that instead.

PR #5798, opened 2019-01-18
FollowCameraPointersInput and a Base class for any CameraPointersInput.

mrdunk:
I started to itch at the amount of code duplication around these cameras...
Please look at BaseCameraPointersInput in this change. It's a Base class that any CameraPointersInput can inherit from.

If there are good reasons that i'm missing why we shouldn't subclass here, let me know and i'll refactor.


src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,328 @@
...
+            }
+        };
+
+        this._observer = this.camera.getScene().onPointerObservable.add(
+            this._pointerInput,
+            PointerEventTypes.POINTERDOWN | PointerEventTypes.POINTERUP |
+            PointerEventTypes.POINTERMOVE);
+
+        this._onLostFocus = () => {
+            this.log("FollowCameraPointersInput._onLostFocus");
```
deltakosh: this has to be removed

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,328 @@
...
+            this._onLostFocus = null;
+        }
+    }
+
+    /**
+     * Gets the class name of the current input.
+     * @returns the class name
+     */
+    public getClassName(): string {
+        return "ArcRotateCameraPointersInput";
```
deltakosh: need to be fixed with correct name

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,328 @@
...
+    /**
+     * Get the friendly name associated with the input class.
+     * @returns the input friendly name
+     */
+    public getSimpleName(): string {
+        return "pointers";
+    }
+
+    protected doDoubleTouch(type: string) {
+        this.log("FollowCameraPointersBase.doDoubleTouch(", type, ")");
```
deltakosh: we have to remove all logs for final review

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,328 @@
...
+               multiTouchPanPosition,
+               ")");
+    }
+
+    protected onContextMenu(evt: PointerEvent): void {
+        this.log("FollowCameraPointersInput.onContextMenu");
+        evt.preventDefault();
+    }
+
+    protected onGestureStart(e: PointerEvent): void {
```
deltakosh: I don't think we should introduce specific MS feature (further more now that Edge will use chromium)

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,328 @@
+import { Nullable } from "../../types";
```
deltakosh: Interesting but need to be cleaned if we want to use it as base class for all pointer inputs

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: o yes.
Let's make `getClassName()` `abstract` in the Base class to force the user to set it in the derived class.

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: agreed.
the switchable `log()` method  i have is more complicated than it has to be.

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: as above.

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: hmm, interesting...
it appears `this` in `BaseCameraPointersInput` is not able to access the derived `getClassName()`.

Learning about JS internals now. stay tuned.

mrdunk: Ok, setting a variable as `abstract` and returning that in the base `getClassName()` seems like a straight forward way to do this.

Interested to hear alternatives.

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: @deltakosh I presume your just busy but want to make sure your aware i've cleaned this up.
I'm not in any hurry at my end so whenever works for me.

cheers!
dunk.

deltakosh:
Hey:) I saw conflicts so I thought you were still working on it

mrdunk:
Hi,
apologies, i missed the big "This branch has conflicts" message.
Not used to the interface.

i'll get those tidied up.

thanks!
dunk.

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,291 @@
...
+     * Called on pointer POINTERMOVE event if multiple touches are active.
+     * Override this method to provide functionality.
+     */
+    protected doMultiTouch(pointA: Nullable<PointerTouch>,
+                           pointB: Nullable<PointerTouch>,
+                           previousPinchSquaredDistance: number,
+                           pinchSquaredDistance: number,
+                           previousMultiTouchPanPosition: Nullable<PointerTouch>,
+                           multiTouchPanPosition: Nullable<PointerTouch>): void {
+        /* if (previousPinchSquaredDistance === 0 && previousMultiTouchPanPosition === null) {
```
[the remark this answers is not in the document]
mrdunk: Just meant as documentation on how to use this method.
I tried writing it in the docstring but code was much simpler.

How do you want this?
option 1: Un-comment the `if` statement but leave the text inside the block.
option 2: Remove the whole comment; `FollowCameraPointersInput` already gives an example.

deltakosh:
Well let's start with arcrotate first ;) So we can close this PR

mrdunk:
> Well let's start with arcrotate first ;) So we can close this PR

So ArcRotate implementation was quite straight forward to do.
I'm currently working on some tests that prove my implementation and the original have the same behavior for all input modes.
You can expect a pull request in the next few days.

deltakosh:
Do you want to update this one?

mrdunk:
Here's my proposed workflow:

Since i want to test both implementations of `ArcRotateCameraPointersInput` together i was going to submit a temporary PR (not intended for merging) parented to this one with my new `ArcRotateCameraPointersInput` and the existing one along with my tests proving they are functionally identical.

If/when you agree my tests prove the new `ArcRotateCameraPointersInput` is adequate i'll overwrite the old one in this PR and scrap the temporary one.

Open to other suggestions.
We could do it all in this PR and delete the tests before submission. (the tests are only useful while there are 2 versions of `ArcRotateCameraPointersInput`.)

deltakosh:
Sounds like a good plan to me :)

mrdunk:
i don't believe the latest build failure is related to my code;

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -0,0 +1,355 @@
...
+        this._metaKey = false;
+        this._shiftKey = false;
+        this._buttonsPressed = 0;
+    }
+
+    /**
+     * Gets the class name of the current input.
+     * @returns the class name
+     */
+    public getClassName(): string {
```
deltakosh: why not overloading the function on each child?

deltakosh:
Not sure which PR you want to keep :) please close the unwanted one

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: > why not overloading the function on each child?

I already mentioned this in the "conversation" tab ~17 days ago...
..."need to be fixed with correct name"...

While overloading the function certainly looks like the correct thing to do here, it's not actually that simple.
`getClassName()` is called from the constructor of one of the parent classes. That seems to mean it does not exist in `this` of the child classes until after the constructor is called....

This is all new to me so it's possible/likely my interpretation is wrong; i'm not a JS guy.

I'm afraid my brain has dropped the exact error... i'll recreate when i get home this evening if what i'm describing here makes no sense.

deltakosh: well this is the case on all our classes:
* https://github.com/BabylonJS/Babylon.js/blob/master/src/Meshes/linesMesh.ts#L103
* https://github.com/BabylonJS/Babylon.js/blob/master/src/Meshes/groundMesh.ts#L44

src/Cameras/Inputs/BaseCameraPointersInput.ts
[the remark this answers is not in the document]
mrdunk: Was thinking about this some more;
I don't like the idea of 1. (un-commenting) as it would take up space even when minified.

That leaves 2. (Remove the whole comment).

Agree?

src/Cameras/Inputs/BaseCameraPointersInput.ts
deltakosh: sure thing:)

src/Cameras/Inputs/BaseCameraPointersInput.ts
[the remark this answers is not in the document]
deltakosh: agree on 2

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: Hrm, interesting...
so i originally wanted `getClassName()` in the base class to be virtual/abstract to force child implementations to set it.
Unfortunately setting `abstract public getClassName(): string;` in the base class causes the behavior i described; The method is undefined at compile time so the code errors at run time.

Setting `public getClassName(): string { return "something"; }` in the base class then overriding it later works; There is always a valid implementation, even if the one in Base should never be used.

I've submitted a fix that compiles and runs cleanly.
There is an argument that we should put a `console.error()` in `BaseCameraPointersInput.getClassName()`.

Thoughts?

deltakosh:
> I had fun though.
This is the only important point!

For the test, you can add them to our tests so it will just be part of the overall testing suite

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: Nope. marking it abstract breaks at run time.
Let's leave it as it is.

mrdunk:
Build failures still look unrelated to my changes.
Looking at other PRs progress the underlying issue appears fixed.
Anyone know how to re-schedule build?

(Alternatively, i'll push a null diff when i get home this evening to trigger another build.)

deltakosh:
Yes you are right..I can merge then
THanks a lot!!

PR #5799, opened 2019-01-19
inspector light gizmo

TrevorDev:


inspector/src/components/sceneExplorer/entities/sceneTreeItemComponent.tsx
```diff
@@ -10,26 +10,29 @@ import { faSyncAlt, faImage, faCrosshairs, faArrowsAlt, faCompress, faRedoAlt }
...
     private _selectedEntity: any;
 
     constructor(props: ISceneTreeItemComponentProps) {
         super(props);
-
-        const scene = this.props.scene;
         let gizmoMode = 0;
-        if (scene.reservedDataStore && scene.reservedDataStore.gizmoManager) {
-            const manager: GizmoManager = scene.reservedDataStore.gizmoManager;
+        if (props.globalState.gizmoManager) {
```
deltakosh: Why do you need to store it globally and not on the scene?

inspector/src/inspector.ts
```diff
@@ -404,6 +404,12 @@ export class Inspector {
                 this._CreateActionTabs(scene, options, parentControl);
             }
         }
+
+        // Light gizmos
+        scene.lights.forEach((l)=>{
+            debugger;
```
deltakosh: Debugger?

inspector/src/components/sceneExplorer/entities/sceneTreeItemComponent.tsx
TrevorDev: At one point I didn't have a reference to scene so I moved it to global to access it. I think all places its used now have access to the scene but I left it thinking it was better to add to globalstate than scene.reservedDataStore . I can revert if you prefer the other way.

deltakosh: Yes please, so it is more unified

PR #5828, opened 2019-01-28
Added TrailMesh class. Credit to furcatomasz

danjpar:
I have not changed much from his original code.

### DIFF
--- a/src/Meshes/trailMesh.ts
+++ b/src/Meshes/trailMesh.ts
@@ -7,6 +7,9 @@ import { VertexData } from "./mesh.vertexData";
 import { Vector3 } from "../Maths/math";
 import { Scene } from "../scene";
 
+/**
+ * Class used to create a trail following a mesh
+ */
 export class TrailMesh extends Mesh {
 	private _generator: AbstractMesh;
 	private _diameter: number;
@@ -15,7 +18,15 @@ export class TrailMesh extends Mesh {
 	private _sectionVectors: Array<Vector3>;
 	private _sectionNormalVectors: Array<Vector3>;
 
-
+    /**
+     * @constructor
+     * @param name The value used by scene.getMeshByName() to do a lookup.
+     * @param generator The mesh to generate a trail.
+     * @param scene The scene to add this mesh to.
+     * @param diameter Diameter of trailing mesh. Default is 1.
+     * @param length Length of trailing mesh. Default is 60.
+     * @param material Material to apply to trailing mesh. Defaults to scene default.
+     */
 	constructor(name: string, generator: AbstractMesh, scene: Scene, diameter: number = 1, length: number = 60, material: Material) {
 		super(name, scene);
 		scene = this.getScene();
@@ -101,7 +112,10 @@ export class TrailMesh extends Mesh {
 		data.applyToMesh(this, true);
 	}
 
-	public update = () => {
+    /**
+     * Update trailing mesh geometry.
+     */
+	public update() {
 		let positions = this.getVerticesData(VertexBuffer.PositionKind);
 		let normals = this.getVerticesData(VertexBuffer.NormalKind);
 		if (positions && normals) {
--- a/src/Meshes/trailMesh.ts
+++ b/src/Meshes/trailMesh.ts
@@ -65,31 +65,24 @@ export class TrailMesh extends Mesh {
         let positions: Array<number> = [];
         let normals: Array<number> = [];
         let indices: Array<number> = [];
-
+        let meshCenter = Vector3.Zero();
+        if (this._generator._boundingInfo) {
+            meshCenter = this._generator._boundingInfo.boundingBox.centerWorld;
+        }
         let alpha: number = 2 * Math.PI / this._sectionPolygonPointsCount;
         for (let i: number = 0; i < this._sectionPolygonPointsCount; i++) {
             positions.push(
-                Math.cos(i * alpha) * this._diameter,
-                Math.sin(i * alpha) * this._diameter,
-                -this._length
-            );
-            normals.push(
-                Math.cos(i * alpha),
-                Math.sin(i * alpha),
-                0
+                meshCenter.x + Math.cos(i * alpha) * this._diameter,
+                meshCenter.y + Math.sin(i * alpha) * this._diameter,
+                meshCenter.z
             );
         }
         for (let i: number = 1; i <= this._length; i++) {
             for (let j: number = 0; j < this._sectionPolygonPointsCount; j++) {
                 positions.push(
-                    Math.cos(j * alpha) * this._diameter,
-                    Math.sin(j * alpha) * this._diameter,
-                    -this._length + i
-                );
-                normals.push(
-                    Math.cos(j * alpha),
-                    Math.sin(j * alpha),
-                    0
+                    meshCenter.x + Math.cos(j * alpha) * this._diameter,
+                    meshCenter.y + Math.sin(j * alpha) * this._diameter,
+                    meshCenter.z
                 );
             }
             let l: number = positions.length / 3 - 2 * this._sectionPolygonPointsCount;
@@ -116,6 +109,7 @@ export class TrailMesh extends Mesh {
                 l
             );
         }
+        VertexData.ComputeNormals(positions, indices, normals);
         data.positions = positions;
         data.normals = normals;
         data.indices = indices;
@@ -128,8 +122,10 @@ export class TrailMesh extends Mesh {
     /**
      * Start trailing mesh.
      */
-    public start(mesh: TrailMesh = this): void {
-        this._beforeRenderObserver = this.getScene().onBeforeRenderObservable.add(function() { mesh.update(); });
+    public start(): void {
+        this._beforeRenderObserver = this.getScene().onBeforeRenderObservable.add(() => {
+            this.update();
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Meshes/trailMesh.ts
```diff
@@ -0,0 +1,156 @@
...
+     * @param scene The scene to add this mesh to.
+     * @param diameter Diameter of trailing mesh. Default is 1.
+     * @param length Length of trailing mesh. Default is 60.
+     * @param material Material to apply to trailing mesh. Defaults to scene default.
+     */
+	constructor(name: string, generator: AbstractMesh, scene: Scene, diameter: number = 1, length: number = 60, material: Material) {
+		super(name, scene);
+		scene = this.getScene();
+
+		this.layerMask = 2;
```
deltakosh: why this?

src/Meshes/trailMesh.ts
```diff
@@ -0,0 +1,156 @@
...
+import { Scene } from "../scene";
+
+/**
+ * Class used to create a trail following a mesh
+ */
+export class TrailMesh extends Mesh {
+	private _generator: AbstractMesh;
+	private _diameter: number;
+	private _length: number;
+	private _sectionPolygonPointsCount: number = 4;
```
deltakosh: this should be a parameter

src/Meshes/trailMesh.ts
```diff
@@ -0,0 +1,156 @@
...
+		data.positions = positions;
+		data.normals = normals;
+		data.indices = indices;
+		data.applyToMesh(this, true);
+	}
+
+    /**
+     * Update trailing mesh geometry.
+     */
+	public update() {
```
deltakosh: There are some tabs/space alignemnt issues

src/Meshes/trailMesh.ts
```diff
@@ -0,0 +1,156 @@
...
+				positions[l + 3 * i + 2] = this._sectionVectors[i].z;
+				normals[l + 3 * i] = this._sectionNormalVectors[i].x;
+				normals[l + 3 * i + 1] = this._sectionNormalVectors[i].y;
+				normals[l + 3 * i + 2] = this._sectionNormalVectors[i].z;
+			}
+			this.updateVerticesData(VertexBuffer.PositionKind, positions, true, false);
+			this.updateVerticesData(VertexBuffer.NormalKind, normals, true, false);
+		}
+	}
+}
```
deltakosh: The update mechanism should be called automatically and potentially turned off with an option if user wants to call it manually (Our golden rule is simplicity first :))

src/Meshes/trailMesh.ts
danjpar: Sorry, I saw that somewhere else in another file. Wasn't sure if it was a common thing, I'll get rid of it.

src/Meshes/trailMesh.ts
danjpar: My bad

danjpar:
It probably needs a Parse function too?

deltakosh:
Yep :) a Parse, a serialize and a clone

danjpar:
Sorry for so many commits. Still getting the hang of this..
Does everything look okay? I have not tested this feature at all, yet

danjpar:
I still need to work out some kinks in this. For example, the initial mesh is a big long ugly stick in the middle of the screen until the update function has been called enough times, and when the trail is stopped the mesh remains.

danjpar:
This one is a little tricky.. I'm struggling to figure out a good approach for this.
One thing I thought of is to copy the vertex data from the generator into the trail mesh, so it's basically a copy. As the update function runs, it checks if the position data reflects (entered length) * (generator positions length), and add more vertices each update until length is reached. Then if the trail is stopped, it would need to remove those vertices so just the initial mesh (generator copy) remains.

https://github.com/danjpar/babylonjsTesting
I don't know how to share a playground with the new class included, but here is how I've been testing. TrailMesh starts at lines 94155 in the .max.js file

danjpar:

to stop animations
```
scene.onBeforeRenderObservable.remove(observer);
newTrail.stop();
```

to start
```
observer = scene.onBeforeRenderObservable.add(animate);
newTrail.start();
```

src/Meshes/trailMesh.ts
```diff
@@ -0,0 +1,213 @@
...
+                    Math.cos(i * alpha) * this._diameter,
+                    Math.sin(i * alpha) * this._diameter,
+                    0
+                );
+                this._sectionNormalVectors[i].copyFromFloats(
+                    Math.cos(i * alpha),
+                    Math.sin(i * alpha),
+                    0
+                );
+                Vector3.TransformCoordinatesToRef(this._sectionVectors[i], this._generator.getWorldMatrix(), this._sectionVectors[i]);
```
deltakosh: For perf you should remove the call to getWorldMatrix out of the loop

danjpar:
I was messing around with resetting the animation counter so it doesn't just keep going to infinity, That's why the demo has a weird skip after a couple rounds.

deltakosh:
That looks like a good merge to me :)
Ok to write some doc as well?

danjpar:
just running gulp now to test new js file, I'll write up a doc after I push this commit

deltakosh:
Here probably:
https://github.com/BabylonJS/Documentation/tree/master/content/How_To/mesh
add a new .md and reference it from there:
https://github.com/BabylonJS/Documentation/blob/master/data/statics.json#L198

TrevorDev:
@danjpar @furcatomasz Right now only a single size parameter is support correct? For our 4.0 release video we are trying to highlight this feature. I was trying to have a trail effect on a sword/tube but the trail did not fit well to that shape. When setting the diameter to large it is visible even when the object is not moving as seen here https://playground.babylonjs.com/#L7RTFA. Any ideas on how this could be supported? Would it be possible to set top and bottom points that defines the start of the trail? Or should i limit our demo to spheres with a trail?

danjpar:
@TrevorDev I will play around with the code. I remember having the idea of making the starting points be contained within the generating mesh, but I don't remember why I didn't/couldn't implement it.

This is what creates it:
`
        var meshCenter = _Maths_math__WEBPACK_IMPORTED_MODULE_2__["Vector3"].Zero();
        if (this._generator._boundingInfo) {
            meshCenter = this._generator._boundingInfo.boundingBox.centerWorld;
        }
        var alpha = 2 * Math.PI / this._sectionPolygonPointsCount;
        for (var i = 0; i < this._sectionPolygonPointsCount; i++) {
            positions.push(meshCenter.x + Math.cos(i * alpha) * this._diameter, meshCenter.y + Math.sin(i * alpha) * this._diameter, meshCenter.z);
        }
        for (var i = 1; i <= this._length; i++) {
            for (var j = 0; j < this._sectionPolygonPointsCount; j++) {
                positions.push(meshCenter.x + Math.cos(j * alpha) * this._diameter, meshCenter.y + Math.sin(j * alpha) * this._diameter, meshCenter.z);
            }
            var l = positions.length / 3 - 2 * this._sectionPolygonPointsCount;
            for (var j = 0; j < this._sectionPolygonPointsCount - 1; j++) {
                indices.push(l + j, l + j + this._sectionPolygonPointsCount, l + j + this._sectionPolygonPointsCount + 1);
                indices.push(l + j, l + j + this._sectionPolygonPointsCount + 1, l + j + 1);
            }
            indices.push(l + this._sectionPolygonPointsCount - 1, l + this._sectionPolygonPointsCount - 1 + this._sectionPolygonPointsCount, l + this._sectionPolygonPointsCount);
            indices.push(l + this._sectionPolygonPointsCount - 1, l + this._sectionPolygonPointsCount, l);
        }
`
I think changing the Math.cos and Math.sin values will probably do the trick. I'll mess with it tomorrow.

danjpar:
After some thought, here's a few ideas,
Change diameter to dimensions array of width (x) and height (y), probably easiest way.
Using points as furthest x,y,z in each direction from center, thought this might not look good with complex shapes.
As you mentioned, user specified points.
A side note, the parameter sectionPolygonPointsCount controls how many points to use, 3 is a triangle, 4 is a square, etc.. This could also be adjusted by user with an additional constructor parameter. What do you think?

danjpar:
@TrevorDev 
https://github.com/BabylonJS/Babylon.js/pull/6171
I changed to array. Unfortunately it will still poke through if the object is stationary, but at least this one doesn't stick out the front when the object is moving

TrevorDev:
@danjpar thanks for the insight and new PR :) . I tried setting diameter x = 0 wherever an x was used and got this (still not sure why the black is visible) but its not poking through anymore but the falloff of the trail shrinks from the top and bottom which is a bit unexpected aswell.



I tried playing with _sectionPolygonPointsCount but I didn't really notice a difference. What do you expect with multiple points? A rounded tail?  

I also looked at Unity's trail mesh controls
image

It looks like they dont have points but a width instead but its similar to having 2 points for a top and bottom of an object. Ill give your branch a try. Thanks again.

danjpar:
@TrevorDev
Is backfaceculling false on the material and is it emissive?
Before, _sectionPolygonPointsCount would alter the shape protruding out, and the tail slightly. Not really worth the extra n*length vertices
I will keep testing adjustments

TrevorDev:
@danjpar I added comments to your PR for how I repo'd it. Let me know if you through your adjustments you find a solution. Thanks for your help.

PR #5830, opened 2019-01-29
Fix animation edge cases (one key, from === to)

### LINKED ISSUE #5829: glTF Asset Generator model Animation_NodeMisc_03.gltf index out of bounds
- Min repro: https://www.babylonjs-playground.com/#CGU8VD
- Expected result:  Model should load and animate without error.
- Current result:  Model loads, but animation crashes with an index-out-of-range exception.  This traces back to the _interpolate method in animation.ts, specifically to the definition of the variable endKey: var endKey = keys[key + 1];

bghgary:
Fixes #5829

### DIFF
--- a/src/Animations/animation.ts
+++ b/src/Animations/animation.ts
@@ -576,7 +576,10 @@ export class Animation {
             return highLimitValue.clone ? highLimitValue.clone() : highLimitValue;
         }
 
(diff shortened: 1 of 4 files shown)


bghgary:
Don't merge. My new unit test is failing.

PR #5847, opened 2019-02-01
Load the inspector version corresponding to the engine

Spy-Seth:
Actually the BabylonJS inspector loaded is always the same. This could lead to some incompatibility between the core engine and the inspector. This PR fix the inspector version to the same published in the engine release to mitigate this issue. 

It also add a configuration `inspectorURL ` to the debug layer to inject path of the inspector on the fly.

### DIFF
--- a/src/Debug/debugLayer.ts
+++ b/src/Debug/debugLayer.ts
@@ -1,6 +1,7 @@
 import { Tools } from "../Misc/tools";
 import { Observable } from "../Misc/observable";
 import { Scene } from "../scene";
+import { Engine } from "../Engines/engine";
 
 // declare INSPECTOR namespace for compilation issue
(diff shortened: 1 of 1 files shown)


src/Debug/debugLayer.ts
```diff
@@ -114,7 +119,7 @@ export class DebugLayer {
      * By default it uses the babylonjs CDN.
      * @ignoreNaming
      */
-    public static InspectorURL = 'https://preview.babylonjs.com/inspector/babylon.inspector.bundle.js';
+    public static InspectorURL = `https://unpkg.com/babylonjs-inspector@${Engine.Version}/babylon.inspector.bundle.js`;
```
deltakosh: That is an EXCELLENT idea!!!

PR #5867, opened 2019-02-05
add options for loading materials (skip/silently fail)

brianzinn:


brianzinn:
will update docs after PR is accepted

PR #5871, opened 2019-02-06
onAfterEnterVR

TrevorDev:
https://forum.babylonjs.com/t/when-to-use-resettocurrentrotation/954/10

testPG: https://playground.babylonjs.com/#9FWBEC


src/Cameras/VR/vrExperienceHelper.ts
```diff
@@ -316,10 +320,15 @@ export class VRExperienceHelper {
...
     /**
-     * Observable raised when entering VR.
+     * Observable raised right before entering VR.
      */
     public onEnteringVRObservable = new Observable<VRExperienceHelper>();
 
+    /**
+     * Observable raised when entering VR has completed.
+     */
+    public onAfterEnteringVRObservable = new Observable<OnAfterEnteringVRObservableEvent>();
```
deltakosh: why not just a boolean? (As the OnAfterEnteringVRObservableEvent is no exported)

TrevorDev: In case additional info is needed in the future, should this be exported in this case?

deltakosh: Yep

PR #5875, opened 2019-02-06
MeshExploder

danjpar:
~~Having trouble getting the function to split a mesh into submeshes to work, but for the time being,~~ this works

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -114,7 +114,7 @@
 - Added OnAfterEnteringVRObservable to webVRHelper ([TrevorDev](https://github.com/TrevorDev))
 - Added Support for Side By Side and Top/Bottom VR videos in the [video dome](https://doc.babylonjs.com/how_to/360videodome#video-types) ([Sebavan](https://github.com/Sebavan))
 - onActiveCameraChanged shouldn't be fired when rendering rig cameras ([TrevorDev](https://github.com/TrevorDev))
-- Added `ExplodedMesh` class ([danjpar](https://github.com/danjpar))
+- Added `MeshExploder` class ([danjpar](https://github.com/danjpar))
 
 ### OBJ Loader
 - Add color vertex support (not part of standard) ([brianzinn](https://github.com/brianzinn))
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -114,6 +114,7 @@
 - Added OnAfterEnteringVRObservable to webVRHelper ([TrevorDev](https://github.com/TrevorDev))
 - Added Support for Side By Side and Top/Bottom VR videos in the [video dome](https://doc.babylonjs.com/how_to/360videodome#video-types) ([Sebavan](https://github.com/Sebavan))
 - onActiveCameraChanged shouldn't be fired when rendering rig cameras ([TrevorDev](https://github.com/TrevorDev))
+- Added `ExplodedMesh` class ([danjpar](https://github.com/danjpar))
 
 ### OBJ Loader
 - Add color vertex support (not part of standard) ([brianzinn](https://github.com/brianzinn))
--- a/src/Meshes/explodedMesh.ts
+++ b/src/Meshes/explodedMesh.ts
@@ -31,6 +31,7 @@ export class ExplodedMesh extends Mesh {
         } else {
             this._setCenterMesh();
         }
+        this._centerMesh.computeWorldMatrix();
         if (this._centerMesh._boundingInfo) {
             this._centerOrigin = this._centerMesh._boundingInfo.boundingBox.centerWorld;
         } else {
@@ -43,6 +44,7 @@ export class ExplodedMesh extends Mesh {
             for (var index = 0; index < this._meshes.length; index++) {
                 if (this._meshes[index]) {
                     var mesh = this._meshes[index];
+                    mesh.computeWorldMatrix();
                     if (mesh._boundingInfo) {
                         var meshCenter = mesh._boundingInfo.boundingBox.centerWorld;
                         this._meshesOrigins.push(meshCenter.clone());
--- a/src/Meshes/explodedMesh.ts
+++ b/src/Meshes/explodedMesh.ts
@@ -1,0 +1,182 @@
+import { Mesh } from "../Meshes/mesh";
+import { Scene } from "../scene";
+import { Vector3 } from "../Maths/math";
+
+/**
+ * Class used to explode meshes.
+ */
+export class ExplodedMesh extends Mesh {
+    private _centerMesh: Mesh;
+    private _centerOrigin: Vector3;
+    private _meshes: Array<Mesh>;
+    private _meshesOrigins: Array<Vector3>;
+    private _originsVectors: Array<Vector3>;
+
+    /**
+     * @constructor
+     * @param name The value used by scene.getMeshByName() to do a lookup.
+     * @param meshes The meshes to explode.
+     * @param scene The scene to add this mesh to.
+     * @param centerMesh The mesh to be center of explosion.
+     */
+    constructor(name: string, meshes: Array<Mesh>, scene: Scene, centerMesh?: Mesh) {
+        super(name, scene);
+
+        this._meshes = meshes;
+        this._meshesOrigins = [];
+        this._originsVectors = [];
+
+        if (centerMesh) {
+            this._centerMesh = centerMesh;
+        } else {
+            this._setCenterMesh();
+        }
+        if (this._centerMesh._boundingInfo) {
+            this._centerOrigin = this._centerMesh._boundingInfo.boundingBox.centerWorld;
+        } else {
+            this._centerOrigin = this._centerMesh.getBoundingInfo().boundingBox.centerWorld;
+        }
+        if (this._meshes.indexOf(this._centerMesh) >= 0) {
+            this._meshes.splice(this._meshes.indexOf(this._centerMesh), 1);
+        }
+        if (this._meshes.length > 1) {
+            for (var index = 0; index < this._meshes.length; index++) {
+                if (this._meshes[index]) {
+                    var mesh = this._meshes[index];
+                    if (mesh._boundingInfo) {
+                        var meshCenter = mesh._boundingInfo.boundingBox.centerWorld;
+                        this._meshesOrigins.push(meshCenter.clone());
+                        this._originsVectors.push(meshCenter.subtract(this._centerOrigin));
+                    }
+                }
+            }
+        }
+    }
+
+    private _setCenterMesh(): void {
+        var positionCount: number = 0;
+        var averageCenter = Vector3.Zero();
+        var totalCenters = Vector3.Zero();
+        var shortestToCenter = Number.MAX_VALUE;
+        for (var index = 0; index < this._meshes.length; index++) {
+            if (this._meshes[index]) {
+                var mesh = this._meshes[index];
+                if (mesh._boundingInfo) {
+                    totalCenters.x += mesh._boundingInfo.boundingBox.centerWorld.x;
+                    totalCenters.y += mesh._boundingInfo.boundingBox.centerWorld.y;
+                    totalCenters.z += mesh._boundingInfo.boundingBox.centerWorld.z;
+                    positionCount++;
+                }
+            }
+        }
+        averageCenter.x = totalCenters.x / positionCount;
+        averageCenter.y = totalCenters.y / positionCount;
+        averageCenter.z = totalCenters.z / positionCount;
+        for (var index = 0; index < this._meshes.length; index++) {
+            if (this._meshes[index]) {
+                var mesh = this._meshes[index];
+                if (mesh._boundingInfo) {
+                    var meshCenter = mesh._boundingInfo.boundingBox.centerWorld;
+                    var vectorToCenter = meshCenter.subtract(averageCenter);
+                    var distanceToCenter = vectorToCenter.length();
+                    if (distanceToCenter < shortestToCenter) {
+                        this._centerMesh = mesh;
+                        shortestToCenter = distanceToCenter;
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * "ExplodedMesh"
+     * @returns "ExplodedMesh"
+     */
+    public getClassName(): string {
+        return "ExplodedMesh";
+    }
+
+    /**
+     * "Exploded meshes"
+     * @returns Array of meshes with the centerMesh at index 0.
+     */
+    public getMeshes(): Array<Mesh> {
+        var meshArray: Array<Mesh> = [];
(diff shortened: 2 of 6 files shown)


danjpar:
I should probably incorporate materials into this? Something like multiMultiMaterials from MergeMeshes

deltakosh:
Yep and can you provide a quick demo to evaluate the end result? With a slider to change the distance?
That could then be the PG used for the doc

Once again thanks a lot btw!

danjpar:

This is a lot more difficult than I thought it was going to be. I have left out the distance modifier, it's hard to get the correct result when passing a variable to change the distance of explosion. The current way is just doubling the distance from the mesh to center, or halving it for implode. That retains the ability to put the meshes back in their original positions. There's probably some algorithm or callback that can solve this, but I can't get the slider to work properly unless it's moved one notch at a time and not quickly/erratically. I tried a for loop to call the function a number of times, but that didn't really work either.

deltakosh:
Yes I get the problem. What about changing the way it is done to something along these lines:
- We create an ExplodedMesh class which use several source meshes and keep track of them
- This ExplodedMesh has a explosionDistance property which is used to generate the vertexbuffers out of the sourceMeshes

Does it make sense

deltakosh:
Also why do we want to work at vertices level? Why not just moving the meshes?

danjpar:
> Also why do we want to work at vertices level? Why not just moving the meshes?

Are you referring to ```mesh.translate(vectorFromCenter, distanceToCenter, Space.WORLD);``` ?
I guess it's just the first way I came up with moving a mesh in a specific direction after searching around for it. I suppose doing something like ```mesh.position = mesh.position.add(vectorFromCenter);``` would be the same result?

I will work on an ExplodedMesh class.

deltakosh:
I think so.. I would love to get more insights from @eamon-obrien

I'm wondering if we consider working with several meshes and then exploding means moving meshes or if we have only one mesh and then we need to explode it by working at vertices level

danjpar:
It will only go into the vertex level if the option separateMeshes is true. I imagine it can be a useful option, even if someone doesn't want the explosion effect, like reversing a Mesh.MergeMeshes.

eaobrien:
Great work danjpar! This is really cool to see. The scenario where this feature would be useful is for folks who have created assets in some CAD software (say a laptop with multiple parts) and want to visualize that on any device with BabylonJS. These assets often have lots of components (think an elevator, for example, that would have lots of parts) and so its the meshes that make up the overall model that would need to be exploded.

danjpar:
I will brainstorm tonight and hopefully have some fresh ideas for tomorrow. I should probably change it from a centerMesh / calculate center of all meshes, to a constructor parameter of Vector3 that defaults Zero(). That would get rid of the other vertex level operations.

src/Meshes/explodedMesh.ts
```diff
@@ -0,0 +1,184 @@
+import { Mesh } from "../Meshes/mesh";
+import { Scene } from "../scene";
+import { Vector3 } from "../Maths/math";
+
+/**
+ * Class used to explode meshes.
+ */
+export class ExplodedMesh extends Mesh {
```
deltakosh: After giving it a thought I do not think this should inherit from Mesh as this is not really a mesh but more a group of them

deltakosh: Thoughts?

danjpar: I agree. Things like position, visibility, etc. are unusable

danjpar:
Should I combine explode, implode, and reset?
Like:

```
    /**
     * Explodes mesh a given number of times.
     * @param explodeSize The number of times to move meshes away from center in increments equal to origin distance to centerMesh. Negative numbers implode. Zero resets to origin.
     */
    public explode(explodeSize: number = 1): void {
        for (var index = 0; index < this._meshes.length; index++) {
            if (this._meshes[index] && this._originsVectors[index]) {
                if (explodeSize == 0) {
                    this._meshes[index].position = this._meshesOrigins[index].clone();
                } else {
                    for (var step = 0; step < Math.abs(explodeSize); step++) {
                        if (explodeSize > 0) {
                            this._meshes[index].position.addInPlace(this._originsVectors[index]);
                        } else {
                            this._meshes[index].position.subtractInPlace(this._originsVectors[index]);
                        }
                    }
                }
            }
        }
    }
```

dist/preview release/what's new.md
```diff
@@ -114,6 +114,7 @@
 - Added OnAfterEnteringVRObservable to webVRHelper ([TrevorDev](https://github.com/TrevorDev))
 - Added Support for Side By Side and Top/Bottom VR videos in the [video dome](https://doc.babylonjs.com/how_to/360videodome#video-types) ([Sebavan](https://github.com/Sebavan))
 - onActiveCameraChanged shouldn't be fired when rendering rig cameras ([TrevorDev](https://github.com/TrevorDev))
+- Added `ExplodedMesh` class ([danjpar](https://github.com/danjpar))
```
deltakosh: the name here must be changed

src/Misc/explodeMeshes.ts
```diff
@@ -0,0 +1,146 @@
+import { Mesh } from "../Meshes/mesh";
+import { Vector3 } from "../Maths/math";
+
+/**
+ * Class used to explode meshes.
+ */
+export class ExplodeMeshes {
```
deltakosh:
```suggestion
export class MeshExploder {
```
Sounds better?

deltakosh:
yes!
and I would prefer to have a float value for the direction with something like
```
let scaledDirection = this._originsVectors[index].scale(direction);
 this._meshes[index].position.subtractInPlace(scaledDirection )
```

src/Misc/explodeMeshes.ts
danjpar: Yes

dist/preview release/what's new.md
danjpar: Oops!

danjpar:
During my commute I had thought of doing something similar to get rid of the step for loop. I wasn't sure how to easily multiply a Vector3 by a single number. I'll make these changes in the morning.

src/Misc/meshExploder.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Explodes mesh a given number of times.
+     * @param direction Number to multiply distance of each mesh's origin from center. Use a negative number to implode, or zero to reset.
+     */
+    public explode(direction: number = 1.0): void {
+        for (var index = 0; index < this._meshes.length; index++) {
+            if (this._meshes[index] && this._originsVectors[index]) {
+                let scaledDirection = this._originsVectors[index].scale(direction);
```
deltakosh: This is good
Just to save on GC, I will recommend using this._originsVectors[index].scaleToRef(direction, this._scaledDirection);
same for following line:
this._meshesOrigins[index].addToRef(scaledDirection, this._meshes[index].position);

danjpar: so create a private _scaledDirection ?

deltakosh: absolutely

src/Misc/meshExploder.ts
```diff
@@ -0,0 +1,112 @@
...
+        meshArray.unshift(this._centerMesh);
+        return meshArray;
+    }
+
+    /**
+     * Explodes mesh a given number of times.
+     * @param direction Number to multiply distance of each mesh's origin from center. Use a negative number to implode, or zero to reset.
+     */
+    public explode(direction: number = 1.0): void {
+        this._scaledDirection = Vector3.Zero();
```
deltakosh:
```suggestion
        this._scaledDirection.fromFloats(0, 0, 0);
```

The goal is to avoid all intantiations

deltakosh: this should be done jsut once when declaring it

danjpar: sorry, forgot to remove it after setting it up top

danjpar: One thing I noticed, since the centerMesh gets spliced out of the meshes array, whatever array the user passes through gets spliced. Should I change to ```this._meshes = meshes.splice();```? Or don't splice out centerMesh, and change the explode function ```if (this._meshes[index] && this._originsVectors[index]) {```?
to
```
if (this._meshes[index] && this._originsVectors[index] && this._meshes[index] !== this._centerMesh) {
```

src/Misc/meshExploder.ts
```diff
@@ -0,0 +1,105 @@
...
+    private _scaledDirection = new Vector3(0.0, 0.0, 0.0);
+
+    /**
+     * Explodes meshes from a center mesh.
+     * @param meshes The meshes to explode.
+     * @param centerMesh The mesh to be center of explosion.
+     */
+    constructor(meshes: Array<Mesh>, centerMesh?: Mesh) {
+
+        this._meshes = meshes.slice();
```
deltakosh: good idea

deltakosh:
Can you create a tiny repro on netlify for @eaobrien to validate the feature?
Looking good to merge for me!

thanks a lot

danjpar:

deltakosh:
Now with the new system we can even get rid of the step on the slider

danjpar:
Step was missing, so I added step="0.1"

danjpar:
I've been messing around with the viewer. I'm getting strange results when using the exploder.

https://vigilant-noether-ce2b24.netlify.com/

open console and run:
```
let element = document.querySelector('babylon');
let viewer2 = BabylonViewer.viewerManager.getViewerByHTMLElement(element);
let newExplode = new BABYLON.MeshExploder(viewer2.sceneManager.models[0].meshes);
newExplode.explode(0);
```
It should do nothing.. but it moves two of the meshes. I built the object in blender, tried apply rotation/scale/location, which made a little difference, but I can't get things to act proper.

deltakosh:
Sounds good!

danjpar:
Same problem when loading object in a regular scene.
I don't understand what is causing this. Meshes on blender's X axis go opposite directions.

deltakosh:
can you make sure that meshes.position remains unchanged when explode(0) ? 
This is what we expect right?

danjpar:
Yes, positions return to normal with 0

danjpar:
So, my workaround is by setting each mesh's parent to null. Since loaded objects have a "__root__" mesh as parent to all meshes. Even with passing this root mesh into the constructor as the centerMesh, the other meshes would move incorrectly.

deltakosh:
We should not break hierarchy. Instead you should be able to call mesh.setAbsolutePosition (which takes hierarchy in account)

danjpar:


https://vigilant-noether-ce2b24.netlify.com/
```
let element = document.querySelector('babylon');
let viewer = BabylonViewer.viewerManager.getViewerByHTMLElement(element);
viewer.sceneManager.models[0].meshes.forEach((o) => { o.setParent(null); } );
let newExplode = new BABYLON.MeshExploder(viewer.sceneManager.models[0].meshes, viewer.sceneManager.models[0].meshes[0]);
newExplode.explode(2);
```
Everything works after removing parent from meshes of imported objects. I will update documentation, this should be okay to merge

danjpar:
in constructor:
```
this._meshesOrigins.push(mesh.getAbsolutePosition().clone());
this._toCenterVectors.push(mesh._boundingInfo.boundingBox.centerWorld.subtract(this._centerMesh._boundingInfo.boundingBox.centerWorld));
```
and in explode:
```
this._toCenterVectors[index].scaleToRef(direction, this._scaledDirection);
this._meshes[index].setAbsolutePosition(this._meshesOrigins[index].add(this._scaledDirection));
```
Thanks!
This works. Should I add another vector3 placeholder like the scaledDirection and use addToRef instead?

danjpar:
kinda messy..
```
if (centerMesh) {
            if (!centerMesh.parent) {
                this._centerMesh = centerMesh;
            } else {
                this._parentPresent = true;
            }
        } else {
            for (var index = 0; index < this._meshes.length; index++) {
                if (this._meshes[index].parent) {
                    this._parentPresent = true;
                    break;
                }
            }
        }
        if (!this._centerMesh && !this._parentPresent) {
            this._setCenterMesh();
        } else {
            for (var index = 0; index < this._meshes.length; index++) {
                if (!this._meshes[index].parent) {
                    this._centerMesh = this._meshes[index];
                    break;
                }
            }
        }
```

deltakosh:
I would think that everything will be flat out in the meshes list. You dump there all the meshes (including children and children of children..)

> this._meshes[index].setAbsolutePosition(this._meshesOrigins[index].add(this._scaledDirection))
This should become addToRef with a property stored like the scaledDirection

danjpar:
I have written that in for my new testing. I am probably over thinking the
root-not-centerMesh issue too much. I should probably just add to
documentation that imported objects should have the root mesh entered as
second parameter or spliced out of the meshes array.

danjpar:
> for the _rootMesh_, can you elaborate? Not sure to get the issue :D
An object imported from a .obj, .babylon, .gltf, or .glb has an extra AbstractMesh that accompanies it. So far This mesh is parent to all meshes from the loaded file. When passing all loaded meshes array into the meshExploder, unless that root mesh (loaded.meshes[0].name == ```__root__```) is set as centermesh or spliced out of the meshes array, any movement to its position is reflected to all other meshes or children.

deltakosh:
Ok gotcha! But actually that should not be a problem with setAbsolutePosition as this takes the hierarchy in account 

Can you make sure to call mesh.computeWorldMatrix(true) after updating a mesh position?

danjpar:

danjpar:
https://vigilant-noether-ce2b24.netlify.com/
acting strange :/

this works as expected

deltakosh:
are they both based on the same initial file? or is there one based of gltf and not the other?

danjpar:
They are same .gltf I made in Blender and babylon.max.js I generated with gulp

src/Meshes/transformNode.ts
```diff
@@ -405,6 +405,9 @@ export class TransformNode extends Node {
             absolutePositionZ = absolutePosition.z;
         }
         if (this.parent) {
+            if (this.parent.parent && this.parent instanceof TransformNode) {
+                this.parent.setAbsolutePosition(this.parent.getAbsolutePosition());
```
deltakosh: not sure to get the point of that line :D

danjpar: I got rid of it, I was thinking the function should be run for each ancestor, which initially looked like it made things work, but I had forgotten that the issue is only when the explosion is set within the viewer.onModelLoadedObservable.add() function. Creating the explosion in console after everything is loaded, works.

danjpar:
No more changes necessary.
MeshExploder works as expected, should be safe to merge.

https://vigilant-noether-ce2b24.netlify.com/

deltakosh:
Congrats!!

PR #5879, opened 2019-02-07
Fix pinch action on FollowCameraPreviewInputs.

mrdunk:
This bug was a casualty of renaming `onMultiTouch()` method in the base class.


deltakosh:
thanks!

PR #5883, opened 2019-02-08
New BRDF

sebavan:


Vinc3r:
Nice work you're doing!
'can't wait for the doc about un/correlated BRDF 'cause I've no idea at all about what is this  :-p

sebavan:
Yup I ll do an update and if you want to understand about it now. Those are good reads:
https://twvideo01.ubm-us.net/o1/vault/gdc2017/Presentations/Hammon_Earl_PBR_Diffuse_Lighting.pdf
https://jo.dreggn.org/home/2016_microfacets.pdf

PR #5891, opened 2019-02-09
UnitTests for BaseCameraPointersInput and ArcRotateCameraPointersInput.

mrdunk:
First time writing JS UnitTests (or anything vaguely front-endey) so suggestions welcome.

I've deliberately avoided using exact values, checking instead if values are greater or less than the previous value.
I've done this to make these tests less brittle but still enforcing the correct trend.

I don't have 100% coverage here; Some of the callbacks have very little logic around when they should be called that i didn't see the value.


src/Cameras/Inputs/StubCameraPointersInput.ts
```diff
@@ -0,0 +1,236 @@
+import { Nullable } from "../../types";
```
deltakosh: I would like to avoid having test files in the main code. Tests (and associated code) must be in the test folder

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -324,7 +324,7 @@ export abstract class BaseCameraPointersInput implements ICameraInput<Camera> {
      * release.
      * Override this method to provide functionality.
      */
-    protected onButtonUp(evt: PointerEvent): void {
+    protected onButtonUp(evt: PointerEvent, buttonCount: number): void {
```
deltakosh: why that change?

tests/unit/babylon/src/Cameras/babylon.pointerInput.tests.ts
```diff
@@ -0,0 +1,1863 @@
+/**
```
deltakosh: Really good!
These tests will definitely be helpful

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: 2 reasons:
 1. Derived classes might well need to know how many buttons are down. Much easier to pass that up from the base (which already knows) than count presses and releases in the derived class.
 2. It's also a useful indicator for that multiple button press state has been cleared up.

deltakosh: sure but this information is already in the PointerEvent no?

src/Cameras/Inputs/StubCameraPointersInput.ts
mrdunk: yea, i tried but failed to get this working from within `tests/unit/babylon/src/`.
I can't `import` from there: `error TS1148: Cannot use imports, exports, or module augmentations when '--module' is 'none'.`

If i define the `StubCameraPointersInput` in `tests/unit/babylon/src/Cameras/babylon.pointerInput.tests.ts` it compiles fine but the `.js` file does not have `BABYLON` defined,
leading to `ReferenceError: BABYLON is not defined` at runtime.

Any advice on this class of issue?

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: hrm. Fair point.
So there is the subtle difference that `buttonCount` is only the number of buttons that `BaseCameraPointersInput` is aware of (maximum 2) whereas `evt: PointerEvent` presumably returns /all/ depressed buttons.
I think the user likely does not care about this subtlety though.

So, since this is only useful in testing when we care about how many buttons `BaseCameraPointersInput` is aware of ,
i think you are going to ask me to remove it....

Removing this will mean we have to either:
  1. not test how the Base class tracks button presses, or
  2. make some other attribute public (which IMO is the same as passing `buttonCount` here).

Personally i prefer to see things well tested so would like this to remain.

There is another instance of this in `onButtonDown(evt: PointerEvent, buttonCount: number)`.
If we remove it here, the same argument applies to both.

mrdunk: Actually,
a compromise would be to keep a public member variable `buttonCount: number` but not document it.

tests/unit/babylon/src/Cameras/babylon.pointerInput.tests.ts
mrdunk: yea, i should have written the tests first.
They would have made refactoring `ArcRotateCameraPointersInput` much easier.

src/Cameras/Inputs/StubCameraPointersInput.ts
deltakosh: I would recommend use plain JS here

src/Cameras/Inputs/BaseCameraPointersInput.ts
deltakosh: well not a big fan of that at all :) I really recommend removing both. the number of buttons to take care of should be something we may want in tests so let's handle it in tests where we can hack around easily without polluting the main code base

tests/unit/babylon/src/Cameras/babylon.pointerInput.tests.ts
deltakosh: I think I say that to me once a week ;D

src/Cameras/Inputs/BaseCameraPointersInput.ts
deltakosh: up?

mrdunk: ?

src/Cameras/Inputs/StubCameraPointersInput.ts
mrdunk: > I would recommend use plain JS here
I presume you mean "plain JS"  in place of `StubCameraPointersInput.ts`?
How do i include files without module loading? (Apologies for the noob question. No real JS experience here.)

I'm not sure that would work anyway;
Because `BaseCameraPointersInput` is an `abstract` class we won't be able to use it directly, either from TS or JS.
The methods we want to test are `protected` so we can't just take an existing CameraInput and override it's methods.

I think the ideal solution here would be to build `StubCameraPointersInput.ts` outside the main BABYLON bundle and include it here....

mrdunk: > How do i include files without module loading? (Apologies for the noob question. No real JS experience here.)

Never mind that question,
I can just remove the `"module": "none"` line in `tests/unit/tsconfig.json`.

src/Cameras/Inputs/BaseCameraPointersInput.ts
deltakosh: Did you remove both? I may have missed the commit  to be honest :)

deltakosh: I would like to get all stuff related to tests in the test file itself to not pollute the main code

src/Cameras/Inputs/StubCameraPointersInput.ts
deltakosh: > I think the ideal solution here would be to build `StubCameraPointersInput.ts` outside the main BABYLON bundle and include it here....
Yes as long as it is not in the main codebase

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: > Did you remove both? I may have missed the commit to be honest :)

Not yet. Been busy with other things.
soon....

src/Cameras/Inputs/StubCameraPointersInput.ts
mrdunk: How i can achieve this?

mrdunk: Well there's a solution that works, although very ugly.
Not the clean tests i was aiming for at all....

Have a look at `StubCameraInput()` in `babylon.pointerInput.tests.ts`.
It stubs out the methods of `ArcRotateCameraPointersInput()`.

anyways, sleepy time for me.
later!

deltakosh: I wonder why not simply using the code generated by your initial TS class as the JS code to use in test. It will give you the scaffolding at least

deltakosh: ok just seen the stub you created :) This is fine for me to be honest.

Also wondering if you could use scene.simualtePointerXXX ?

mrdunk: > I wonder why not simply using the code generated by your initial TS class as the JS code to use in test. It will give you the scaffolding at least.
> StubCameraPointersInput.ts has generated a .js that could be the file to use in tests directly

Yes, this what i was trying.
i could not find a way to include all the generated .js that `StubCameraPointersInput` depends on.
I'm sure it's possible, just i do not understand how JavaScript includes work.
(i'm fine debugging a C or C++ linker. Linking .js is just weird....)

> ok just seen the stub you created :) This is fine for me to be honest.

Hmmm. i think it smells bad.
Unit tests should test code in isolation.
Since my Stub depends on `ArcRotateCameraPointersInput`, the unit tests for `BaseCameraPointersInput` are dependent on future modifications of `ArcRotateCameraPointersInput` not overriding `BaseCameraPointersInput` methods.

> Also wondering if you could use scene.simualtePointerXXX ?

I looked into that. It only covers a subset of the events i needed. (Cant remember which events exactly.) If this was my day job i'd extend the simualtePointer model but i ran out of motivation.

deltakosh: Ok for now. wE can improve it later

deltakosh: Also I want to apologize, I did not want to demotivate you :( It is just that I need to keep the code as unbloated as possible. I hope you understand :(

mrdunk: ah, don't worry about that.
The whole point of this process is so we can discuss the best way of doing things. The fact we have differing opinions on some points is what makes it a discussion.

If no one was standing their ground on what gets included the project would not be the success it is. No apologies to me are required for that.

Communicating purely by text is hard when you don't know the person on the other end.
You can pretty much guarantee if you've offended me it's going to take more than comments about code and i'll let you know.

On a different tangent, here's the hacky little animation that got me playing with Babylon: https://mrdunk.github.io/karaokefox/dist/
Once i get this PR submitted i'll get that fox walking about his forest.

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: right, done.
So i've made the variables that store the latest button press events (`pointA` and `pointB`) `private` members of `BaseCameraPointersInput`.

This way we can do horrible hacky things to access them in tests without them polluting the TS namespace.

src/Cameras/Inputs/StubCameraPointersInput.ts
deltakosh: I love it! And I'm glad I did not upset you as I appreciate your work a lot!

src/Cameras/Inputs/BaseCameraPointersInput.ts
deltakosh: Well this does not solve my issue as we should not have ANYTHING that is not used in the core. I know that this is painful from your standpoint but it looks artificial (because it is :D) to have this properties but not use them

mrdunk: just to confirm, these are existing variables. The Base class needs them.

I have made them slightly more accessible by moving them outside the `attachControl()` method though.

mrdunk: And please, if that's still not acceptable, just say so.
I'll make fun of you but i won't mind.

deltakosh: lol no worries! It is tough to get me upset :)

deltakosh: Oh and yeah for the points! I'm fine with it! Ok to merge for me
Waiting for your final go

mrdunk:
This shows build as passing but
deploy/netify as "Deploy preview failed."

#5952 is showing similar symptoms.
something Build related maybe?

deltakosh:
We can ignore it for now

mrdunk:
woohoo!
Thanks for the help again!
This one took longer than expected.

deltakosh:
But again it is a good one :D

PR #5892, opened 2019-02-09
Use FirefoxHeadless instead of the obsolete PhantomJs in tests.

mrdunk:
One additional thing that could be done here is remove PhantomJS from `npm`'s `package.json` file but i don't know enough about Babylon's build environment to know if that will break anything else...

Advice welcome.

PR #5916, opened 2019-02-13
Attempt 2: Replace obsolete PhantomJs browser in tests.

mrdunk:
This work was originally attempted in https://github.com/BabylonJS/Babylon.js/pull/5892 but we are experiencing strange Build errors there.
It was decided to start again from a clean git workspace.


sebavan:
We should not use Chrome here as it falls back to its software rendering rendering version that we know having some slight diffs which will take us a long time to troubleshoot:

image

sebavan:
Closing in favor of #5923 as this succeeded on Travis. Thanks for all the help.

Issue #5913, opened 2019-02-13
Missing es6 import error messages


TrevorDev:
https://forum.babylonjs.com/t/es6-tree-shaking-experience/1365/2

sebavan:
About some of them I guess the import is missing or the code should be reviewed to not have the message like defaultRenderingPipeline could import postProcessRenderPipelineManagerScene That said we need to prevent circular dependencies.

Let s be careful on this one to make it as natural as possible :-)

PR #5918, opened 2019-02-13
add required es6 includes

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/5913

### DIFF
--- a/src/Layers/glowLayer.ts
+++ b/src/Layers/glowLayer.ts
@@ -21,6 +21,7 @@ import { _TypeStore } from '../Misc/typeStore';
 
 import "../Shaders/glowMapMerge.fragment";
 import "../Shaders/glowMapMerge.vertex";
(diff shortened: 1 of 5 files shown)


src/Layers/glowLayer.ts
```diff
@@ -21,6 +21,7 @@ import { _TypeStore } from '../Misc/typeStore';
 
 import "../Shaders/glowMapMerge.fragment";
 import "../Shaders/glowMapMerge.vertex";
+import "../Layers/effectLayerSceneComponent";
```
sebavan: Is it not creating a circular ref ?

TrevorDev: gulp circularDependencies didnt fail

sebavan: Swweeeeet perfect then

PR #5919, opened 2019-02-14
A start to softbody physics added to AmmoJSPlugin 

BabylonJSGuide:
There is a long way to go before completed but am at a stage where work needs checking before continuing. Please check that methodology of adding the soft body code is acceptable, any suggestions very welcome

A working test can be found at https://adoring-kirch-5b50fe.netlify.com/ with an alternative coding of the test at https://adoring-kirch-5b50fe.netlify.com/test.1.html with some parameters tweaked.

A major change in the AmmoJSPlugin is the addition of an extra parameter, _includeSoftBodies, as the first parameter in the constructor and so pushing back the other parameters by one. Hoping this is acceptable as AmmoJSPlugin just added in 4.0.0 alpha. If not please advise. Its default is false, ie rigid bodies only, when true both rigid and softbodies are possible.

Additions to mesh.ts are forceSharedVertices and increaseVertices.  When a mesh has repeated vertices that do not share a normal the softbody can blow apart as in https://adoring-kirch-5b50fe.netlify.com/test.2.html. Forcing a minimum number of vertices and making them share a normal with its neighboring facets joins the softbody together, as in https://adoring-kirch-5b50fe.netlify.com/test.3.html. Boxes and other meshes do not have sufficient vertices to construct a soft body.

More testing is still required. 

My intention is to add: cloth and rope impostors, anchors, air density, other features as I learn about them and to write the documentation with example playgrounds.


src/Meshes/mesh.ts
```diff
@@ -2398,6 +2398,229 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
+        }
+    }
+
+    /**
+     * Force adjacent facets to share vertices and remove any facets that have all vertices in a line
+     * This will undo any application of covertToFlatShadedMesh
+     * Introduced for use with Ammo.JSPlugin soft bodies
+     * Warning : the mesh is really modified even if not set originally as updatable. A new VertexBuffer is created under the hood each call.
+     */
+    public forceSharedVertices(): void {
```
deltakosh: Is that different from convertToFlatShadedMesh ?

src/Physics/physicsEngine.ts
```diff
@@ -33,6 +33,7 @@ export class PhysicsEngine implements IPhysicsEngine {
 
     /**
      * Creates a new Physics Engine
+     * Only AmmoJSPlugin allows soft bodies
```
deltakosh: We should not have it here for a specific function but more on the documentation site.
We can also move it to the interface declaration of new functions like getBodyPressure

src/Physics/physicsImpostor.ts
```diff
@@ -39,6 +40,23 @@ export interface PhysicsImpostorParameters {
      * Specifies if bi-directional transformations should be disabled
      */
     disableBidirectionalTransformation?: boolean;
+    /**
+     * The pressure inside the physics imposter, softbody only
```
deltakosh: This is where we should mention that only ammo supports it

src/Physics/physicsImpostor.ts
```diff
@@ -244,12 +262,97 @@ export class PhysicsImpostor {
         this._physicsEngine.getPhysicsPlugin().setBodyRestitution(this, value);
     }
 
+    /**
+     * Gets the pressure
+     */
+    get pressure(): number {
+        if (!this._physicsEngine) {
```
deltakosh:
```suggestion
        if (!this._physicsEngine && this._physicsEngine.getPhysicsPlugin().getBodyPressure) {
```

src/Meshes/mesh.ts
BabylonJSGuide: Yes. It is the opposite. It would take a flatShadedMesh and revert it to its previous state. Any flat shaded mesh would be blown apart.

src/Physics/physicsEngine.ts
BabylonJSGuide: Will do

src/Physics/physicsImpostor.ts
BabylonJSGuide: Will do

src/Physics/physicsImpostor.ts
BabylonJSGuide: Presume same request applies to other specific softbody parameters. Will make the changes.

BabylonJSGuide: Have made what I think are equivalent changes. Had problems getting past typescript error messages with your method.

src/Physics/physicsEngine.ts
BabylonJSGuide: Done

src/Physics/physicsImpostor.ts
BabylonJSGuide: Done

src/Meshes/mesh.ts
deltakosh: Makes snes then we should not even mention Ammo here as this is useful for other scenarios
Also you have a type in covertToFlat

src/Physics/physicsImpostor.ts
deltakosh: ok then my only changes would be to store the physicsPlugin jsut for the sake of not asking it twice

BabylonJSGuide: Sorry but not sure of how to store the plugin and only ask once.

src/Meshes/mesh.ts
BabylonJSGuide: Will remove Ammo comment.

BabylonJSGuide: Will remove Ammo comment.

src/Physics/physicsImpostor.ts
deltakosh: something like that:
```
    if (!this._physicsEngine) {
            return;
        }
        const plugin = this._physicsEngine.getPhysicsPlugin();
        if (!plugin .setBodyPressure) {
            return;
        }
        plugin .setBodyPressure!(this, value);
```

src/Physics/Plugins/ammoJSPlugin.ts
```diff
@@ -66,11 +70,22 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
         }
 
         // Initialize the physics world
-        this._collisionConfiguration = new this.bjsAMMO.btDefaultCollisionConfiguration();
-        this._dispatcher = new this.bjsAMMO.btCollisionDispatcher(this._collisionConfiguration);
-        this._overlappingPairCache = new this.bjsAMMO.btDbvtBroadphase();
-        this._solver = new this.bjsAMMO.btSequentialImpulseConstraintSolver();
-        this.world = new this.bjsAMMO.btDiscreteDynamicsWorld(this._dispatcher, this._overlappingPairCache, this._solver, this._collisionConfiguration);
+        if (this._includeSoftBodies) {
+            this._collisionConfiguration = new this.bjsAMMO.btSoftBodyRigidBodyCollisionConfiguration();
```
TrevorDev: I am worried about this._collisionConfiguration and this.world their type in this file was set to any as ammo does not provide typings for them but now when softbodies are enabled these fields can be of different types.

This makes writing code like this needed
```
        this.world.setGravity(this._tmpAmmoVectorA);
        if (this._includeSoftBodies) {
            this.world.getWorldInfo().set_m_gravity(this._tmpAmmoVectorA);
        }
```
Ideally typescript typings should be used to differentiate them. Maybe there should be 2 separate variables.
(eg.  this.world and this.softBodyWorld) that way we could look to add ammojs typings in the future

src/Physics/Plugins/ammoJSPlugin.ts
```diff
@@ -578,8 +769,13 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
      * @param velocity velocity to set
      */
     public setLinearVelocity(impostor: PhysicsImpostor, velocity: Vector3) {
-        this._tmpAmmoVectorA.setValue(velocity.x, velocity.y, velocity.z);
-        impostor.physicsBody.setLinearVelocity(this._tmpAmmoVectorA);
+        if (!impostor.soft) {
+            this._tmpAmmoVectorA.setValue(velocity.x, velocity.y, velocity.z);
```
TrevorDev: I see a lot of these are not supported, is this a limitation of the engine or something that could be added in the future?

src/Physics/physicsImpostor.ts
BabylonJSGuide: Will get this sorted

src/Physics/Plugins/ammoJSPlugin.ts
```diff
@@ -50,10 +53,11 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
 
     /**
      * Initializes the ammoJS plugin
+     * @param _includeSoftBodies when true you can mix tigid and soft bodies (default false)
      * @param _useDeltaForWorldStep if the time between frames should be used when calculating physics steps (Default: true)
      * @param ammoInjection can be used to inject your own ammo reference
      */
-    public constructor(private _useDeltaForWorldStep: boolean = true, ammoInjection: any = Ammo) {
+    public constructor(private _includeSoftBodies: boolean = false, private _useDeltaForWorldStep: boolean = true, ammoInjection: any = Ammo) {
```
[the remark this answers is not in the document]
BabylonJSGuide: I was hoping that as Ammo was still in alpha it wouldn't be a problem. The constructors for Oimo and Cannon append from the front 
Oimo  - 'constructor(iterations?: number, oimoInjection = OIMO)'
Cannon - 'constructor(private _useDeltaForWorldStep: boolean = true, iterations: number = 10, cannonInjection = CANNON)'
Ammo - 'constructor(private _includeSoftBodies: boolean = false, private _useDeltaForWorldStep: boolean = true, ammoInjection: any = Ammo)'

Putting it at the end would mean writing in parameters for __useDeltaForWorldStep and ammoInjection as well if the user wanted softbodies!

A possible alternative would be to do a AmmoSoftJSPlugin inheriting from AmmoJSPlugin

src/Physics/Plugins/ammoJSPlugin.ts
BabylonJSGuide: Problem here is that when soft bodies are included rigid bodies are also included in the same world so cannot have two separate variables for the world.  I am working at the boundaries of my Typescript knowledge talking about inheritance but if there was an extension of AmmoJSPlugin to AmmoJSPlugin would the mean you could have 

in AmmoJSPlugin
```
public setGravity(gravity: Vector3): void {
        this._tmpAmmoVectorA.setValue(gravity.x, gravity.y, gravity.z);
        this.world.setGravity(this._tmpAmmoVectorA);
    }
```
and in AmmoSoftJSPlugin
```
public setGravity(gravity: Vector3): void {
        this._tmpAmmoVectorA.setValue(gravity.x, gravity.y, gravity.z);
        this.world.setGravity(this._tmpAmmoVectorA);
        this.world.getWorldInfo().set_m_gravity(this._tmpAmmoVectorA);
    }
```
In fact you would leave AmmoJSPlugin as it currently is and in AmmoSoftJSPlugin where I have got

```
if (_includeSoftBodies) {
      do A
}
else {
      do B
}
```

you do 

```
do A
```

src/Physics/Plugins/ammoJSPlugin.ts
BabylonJSGuide: For some reason I thought they were not supported for soft bodies. When I went back to check I find they are. Will correct.

BabylonJSGuide:
Will do the remaining corrections and suggestions when I know whether the idea of an AmmoSoftJSPlugin inheriting from AmmoJSPlugin is the best way forward. If this is so would it then be a good idea to close this PR and await my re-write for AmmoSoftJSPlugin?

src/Physics/Plugins/ammoJSPlugin.ts
deltakosh: I'm worries that this could be complex from an user standpoint. We should not multiplicate the plugins.

Further more does that mean that soft and rigid bodies cannot be used simultaneously?

BabylonJSGuide: AmmoJSPlugin would be for rigid bodies only, whereas AmmoSoftJSPlugin would be for both soft and rigid bodies. (Could call it AmmoSoftandRigidJSPlugin but too long). I am quite happy to leave it as one plugin along the lines I am doing now. However I cannot then see away around having code of the form

```
 this.world.setGravity(this._tmpAmmoVectorA);
if (this._includeSoftBodies) {
     this.world.getWorldInfo().set_m_gravity(this._tmpAmmoVectorA);
 }
```
which Trevor was concerned about.

Currently I am using two flags to deal with soft physics. The _includeSoftBodies flag is for setting the Ammo world parameters either false for a rigid world or true for a soft and rigid world. The soft flag is a property of impostors and determines how they are treated.

src/Physics/Plugins/ammoJSPlugin.ts
BabylonJSGuide: Understand know why I thought they were not supported. For rigid bodies you use getLinearVelocity and setLinearVelocity whereas it turns out for soft bodies you use linearVelocity. Same for angular. Have now combined them in the plugin

src/Physics/physicsImpostor.ts
BabylonJSGuide: Done

src/Meshes/mesh.ts
BabylonJSGuide: Done

src/Physics/Plugins/ammoJSPlugin.ts
BabylonJSGuide: Done

src/Physics/Plugins/ammoJSPlugin.ts
deltakosh: I let @TrevorDev decide here but I'm more on the one single plugin side ;)

src/Physics/Plugins/ammoJSPlugin.ts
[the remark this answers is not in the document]
TrevorDev: There are already a good number of samples/playgrounds that might get broken so i would prefer it at the end. Passing undefined allows default parameters to be used so if softbodies are needed some one could do Ammo(undefined, undefined, ... , true). In hindsight it might have been better to have the physics plugins take an object to better support the options as they grow but I don't think it's too bad.

src/Physics/Plugins/ammoJSPlugin.ts
TrevorDev: I agree that having another plugin is undesirable. After looking at this a bit more I think the way you have it now is probably fine and if we add typings this can exist as a union type.

BabylonJSGuide: I unresolved this because I thought it depended on the answer to the above conversation but it doesn't. What ever the answer this is resolved. Apologies.

src/Physics/Plugins/ammoJSPlugin.ts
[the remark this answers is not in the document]
BabylonJSGuide: Possibly I have been over complicating things (not unusual for me). How about this for an idea - Do away with the _includeSoftBodies flag and set the Ammo physics world to always include soft and rigid bodies. I have tested out a couple of rigid body only examples in a soft and rigid body world and there seems to be no problem.
https://adoring-kirch-5b50fe.netlify.com/rigid1.html
https://adoring-kirch-5b50fe.netlify.com/rigid2.html

@deltakosh and @TrevorDev if you both agree I will submit a new commit with the world set to a soft and rigid one. If you do not agree I will move the _includesoftBodies parameter to the end.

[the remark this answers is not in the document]
deltakosh: Oh that could be great! I let @TrevorDev  weights in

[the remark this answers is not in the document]
TrevorDev: Sounds good to me.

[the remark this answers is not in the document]
BabylonJSGuide: Will get on with it ☺️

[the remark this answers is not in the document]
BabylonJSGuide: Done. Now just one physics world with both rigid and soft bodies

BabylonJSGuide:
Also added the cloth soft body - examples 
https://adoring-kirch-5b50fe.netlify.com/clothtest.1.html
https://adoring-kirch-5b50fe.netlify.com/clothtest.2.html
https://adoring-kirch-5b50fe.netlify.com/clothtest.3.html

Will finish docs during next week

deltakosh:
Woot! I love the samples!

deltakosh:
I'm ok to merge
Is it ok for you @BabylonJSGuide ?

BabylonJSGuide:
Yes please am happy to merge. Will work on the enhancements after completing the docs.

deltakosh:
Thank you sir!

PR #5935, opened 2019-02-16
Physics helper refactoring & physics raycasting

bobalazek:
It's been a while since I last contributed, but I wanted to pick up where I left off.
For that reason, I did some refactoring on the helper, as I want to implement some new features & stuff.
I know that generally breaking changes aren't good, but couldn't think of a better solution - the major change now is, that the additional options are now a separate object, instead of single attributes on the method call.

New example code:
```

    // Scene
    scene.enablePhysics();


    // Physics engine
    var physicsViewer = new BABYLON.Debug.PhysicsViewer();
    var physicsHelper = new BABYLON.PhysicsHelper(scene);


    // Camera
    var camera = new BABYLON.FreeCamera("camera1", new BABYLON.Vector3(0, 24, -64), scene);


    // Lightning
    light.intensity = 0.7;


    // Ground
    var ground = BABYLON.Mesh.CreateGround("ground1", 64, 64, 2, scene);
    ground.physicsImpostor = new BABYLON.PhysicsImpostor(ground, BABYLON.PhysicsImpostor.BoxImpostor, { mass: 0, restitution: 1 }, scene);


    // Boxes
    var boxSize = 2;
    var boxPadding = 4;
    var minXY = -12;
    var maxXY = 12;
    var maxZ = 8;
    var boxParams = { height: boxSize, width: boxSize, depth: boxSize };
    var boxImpostorParams = { mass: boxSize, restitution: 0, friction: 1 };
    var boxMaterial = new BABYLON.StandardMaterial("boxMaterial");
    boxMaterial.diffuseColor = new BABYLON.Color3(1, 0, 0);
    for (var x = minXY; x <= maxXY; x += boxSize + boxPadding) {
        for (var z = minXY; z <= maxXY; z += boxSize + boxPadding) {
            for (var y = boxSize / 2; y <= maxZ; y += boxSize) {
                var boxName = "box:" + x + ',' + y + ',' + z;
                var box = BABYLON.MeshBuilder.CreateBox(boxName, boxParams, scene);
                box.position = new BABYLON.Vector3(x, y, z);
                box.material = boxMaterial;
                box.physicsImpostor = new BABYLON.PhysicsImpostor(box, BABYLON.PhysicsImpostor.BoxImpostor, boxImpostorParams, scene);
                physicsViewer.showImpostor(box.physicsImpostor);
            }
        }
    }


    // Radial explosion impulse/force
    var radius = 8;
    var strength = 20;
    var origins = [
        new BABYLON.Vector3(-8, 6, 0),
        new BABYLON.Vector3(0, 0, 0),
    ];
    for (var i = 0; i < origins.length; i++) {
        var origin = origins[i];

        setTimeout(function (origin) {
            var event = physicsHelper.applyRadialExplosionImpulse( // or .applyRadialExplosionForce
                origin,
                {
                    radius: radius,
                    strength: strength,
                    falloff: BABYLON.PhysicsRadialImpulseFalloff.Linear, // or BABYLON.PhysicsRadialImpulseFalloff.Constant
                }
            );

            // Debug
            var eventData = event.getData();
            var debugData = showExplosionDebug(eventData);
            setTimeout(function (debugData) {
                hideExplosionDebug(debugData);
                event.dispose(); // we need to cleanup/dispose, after we don't use the data anymore
            }, 1500, debugData);
            // Debug - END
        }, i * 2000 + 1000, origin);
    }


    // Gravitational field
    var gravitationalFieldOrigin = new BABYLON.Vector3(0, 6, 10);
    setTimeout(function() {
        var event = physicsHelper.gravitationalField(
            gravitationalFieldOrigin,
            {
                radius: radius,
                strength: strength,
                falloff: BABYLON.PhysicsRadialImpulseFalloff.Linear,
            }
        );
        event.enable();

        // Debug
        var eventData = event.getData();
        var sphere = eventData.sphere;
        addMaterialToMesh(sphere);
        sphere.isVisible = true;

        setTimeout(function (sphere) {
            event.disable();
            sphere.isVisible = false;
            event.dispose(); // we need to cleanup/dispose, after we don't use the data anymore
        }, 3000, sphere);
        // Debug - END
    }, 4000);


    // Updraft
    var updraftOrigin = new BABYLON.Vector3(12, 0, 12);
    setTimeout(function() {
        var event = physicsHelper.updraft(
            updraftOrigin,
            {
                radius: 12,
                strength: 2,
                height: 20,
            }
        );
        event.enable();

        // Debug
        var eventData = event.getData();
        var cylinder = eventData.cylinder;
        addMaterialToMesh(cylinder);
        cylinder.isVisible = true;

        setTimeout(function (cylinder) {
            event.disable();
            cylinder.isVisible = false;
            event.dispose(); // we need to cleanup/dispose, after we don't use the data anymore
        }, 2000, cylinder);
        // Debug - END
    }, 6000);


    // Vortex
    var vortexOrigin = new BABYLON.Vector3(0, -8, 8);
    setTimeout(function() {
        var event = physicsHelper.vortex(
            vortexOrigin,
            {
                radius: 20,
                strength: 40,
                height: 30,
            }
        );
        event.enable();

        // Debug
        var eventData = event.getData();
        var cylinder = eventData.cylinder;
        addMaterialToMesh(cylinder);
        cylinder.isVisible = true;

        setTimeout(function (cylinder) {
            event.disable();
            cylinder.isVisible = false;
            event.dispose(); // we need to cleanup/dispose, after we don't use the data anymore
        }, 10000, cylinder);
        // Debug - END
    }, 8000);


    // Helpers
    function addMaterialToMesh(sphere) {
        var sphereMaterial = new BABYLON.StandardMaterial("sphereMaterial", scene);
        sphereMaterial.alpha = 0.5;
        sphere.material = sphereMaterial;
    }

    function showExplosionDebug(data) {
        addMaterialToMesh(data.sphere);
        data.sphere.isVisible = true;

        return {
            sphere: data.sphere,
        };
    }

    function hideExplosionDebug(debugData) {
        debugData.sphere.isVisible = false;
    }

    return scene;

};
```

### DIFF
--- a/src/Physics/physicsHelper.ts
+++ b/src/Physics/physicsHelper.ts
@@ -54,10 +54,10 @@ export class PhysicsHelper {
         }
 
         if (typeof radiusOrEventOptions === 'number') {
-          radiusOrEventOptions = new PhysicsRadialExplosionEventOptions();
-          radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
-          radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
-          radiusOrEventOptions.falloff = falloff || radiusOrEventOptions.falloff;
+            radiusOrEventOptions = new PhysicsRadialExplosionEventOptions();
+            radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
+            radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
+            radiusOrEventOptions.falloff = falloff || radiusOrEventOptions.falloff;
         }
 
         var event = new PhysicsRadialExplosionEvent(this, this._scene, radiusOrEventOptions);
@@ -96,10 +96,10 @@ export class PhysicsHelper {
         }
 
         if (typeof radiusOrEventOptions === 'number') {
-          radiusOrEventOptions = new PhysicsRadialExplosionEventOptions();
-          radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
-          radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
-          radiusOrEventOptions.falloff = falloff || radiusOrEventOptions.falloff;
+            radiusOrEventOptions = new PhysicsRadialExplosionEventOptions();
+            radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
+            radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
+            radiusOrEventOptions.falloff = falloff || radiusOrEventOptions.falloff;
         }
 
         var event = new PhysicsRadialExplosionEvent(this, this._scene, radiusOrEventOptions);
@@ -138,10 +138,10 @@ export class PhysicsHelper {
         }
 
         if (typeof radiusOrEventOptions === 'number') {
-          radiusOrEventOptions = new PhysicsRadialExplosionEventOptions();
-          radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
-          radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
-          radiusOrEventOptions.falloff = falloff || radiusOrEventOptions.falloff;
+            radiusOrEventOptions = new PhysicsRadialExplosionEventOptions();
+            radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
+            radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
+            radiusOrEventOptions.falloff = falloff || radiusOrEventOptions.falloff;
         }
 
         var event = new PhysicsGravitationalFieldEvent(this, this._scene, origin, radiusOrEventOptions);
@@ -171,11 +171,11 @@ export class PhysicsHelper {
         }
 
         if (typeof radiusOrEventOptions === 'number') {
-          radiusOrEventOptions = new PhysicsUpdraftEventOptions();
-          radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
-          radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
-          radiusOrEventOptions.height = height || radiusOrEventOptions.height;
-          radiusOrEventOptions.updraftMode = updraftMode || radiusOrEventOptions.updraftMode;
+            radiusOrEventOptions = new PhysicsUpdraftEventOptions();
+            radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
+            radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
+            radiusOrEventOptions.height = height || radiusOrEventOptions.height;
+            radiusOrEventOptions.updraftMode = updraftMode || radiusOrEventOptions.updraftMode;
         }
 
         var event = new PhysicsUpdraftEvent(this, this._scene, origin, radiusOrEventOptions);
@@ -205,10 +205,10 @@ export class PhysicsHelper {
         }
 
         if (typeof radiusOrEventOptions === 'number') {
-          radiusOrEventOptions = new PhysicsVortexEventOptions();
-          radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
-          radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
-          radiusOrEventOptions.height = height || radiusOrEventOptions.height;
+            radiusOrEventOptions = new PhysicsVortexEventOptions();
+            radiusOrEventOptions.radius = <number><any>radiusOrEventOptions;
+            radiusOrEventOptions.strength = strength || radiusOrEventOptions.strength;
+            radiusOrEventOptions.height = height || radiusOrEventOptions.height;
         }
 
         var event = new PhysicsVortexEvent(this, this._scene, origin, radiusOrEventOptions);
--- a/src/Physics/physicsHelper.ts
+++ b/src/Physics/physicsHelper.ts
@@ -35,9 +35,9 @@ export class PhysicsHelper {
 
         this.options = {...(new PhysicsHelperOptions()), ...options};
 
-        if (this.options.useNativePhysicsRaycastingIfPresent) {
+        if (this.options.useNativePhysicsRaycastingIfAvailable) {
           if (this._physicsEngine.getPhysicsPluginName() === 'OimoJSPlugin') {
-            this.options.useNativePhysicsRaycastingIfPresent = false;
+            this.options.useNativePhysicsRaycastingIfAvailable = false;
             Logger.Warn('Native raycasting not available on the OimoJSPlugin.');
           }
         }
@@ -185,7 +185,7 @@ export class PhysicsHelperOptions {
     /**
      * If set to true, it will use the native raycasting
      */
-    public useNativePhysicsRaycastingIfPresent: boolean = true;
+    public useNativePhysicsRaycastingIfAvailable: boolean = true;
 }
 
 /**
@@ -244,7 +244,7 @@ export class PhysicsRadialExplosionEvent {
         var impostorObjectCenter = impostor.getObjectCenter();
         var direction = impostorObjectCenter.subtract(origin);
 
-        if (this._physicsHelper.options.useNativePhysicsRaycastingIfPresent) {
+        if (this._physicsHelper.options.useNativePhysicsRaycastingIfAvailable) {
           var to = origin.clone().addInPlace(
             direction.clone().normalize().multiplyInPlace(new Vector3(
               this._options.radius,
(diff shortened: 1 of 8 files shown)


bobalazek:
Added also physics raycasting now - at least I started work on it. The main reason is, because I currently use the babylon raycasting for the physics stuff, so it may be better to use the physics own raycasting

deltakosh:
Regarding raycasting, the problem is that Oimo does not have one. 
Did you end up finding any evidence of perf improvement using physic raycasting?

Pinging @TrevorDev  for a second thought

bobalazek:
Oh yeah, will update the documentation too soon!
As for the physics part, I don't think that raycasting is used anywhere else, besides in the `PhysicsHelper`. For Oimo, I was thinking to have a fallback there to native raycasting (like it is now).

I'm not yet sure if/how much the physics raycast would be faster, but the main idea was rather to expose that basic functionality if somebody would use it. Will try and test it!

The next thing I was planning to do, was implementing some vehicle physics, so I was hoping that "native" physics raycasting would be more performant, as you do make heavy use for it there.

deltakosh:
Makes sense!! Just waiting for @TrevorDev approval (please just add oimo fallback to native raycast and I'm good to merge)

bobalazek:
Done!

src/Physics/physicsHelper.ts
```diff
@@ -19,29 +19,37 @@ export class PhysicsHelper {
...
      * Applies a radial explosion impulse
      * @param origin the origin of the explosion
-     * @param radius the explosion radius
-     * @param strength the explosion strength
-     * @param falloff possible options: Constant & Linear. Defaults to Constant
+     * @param eventOptions the options of radial explosion
      * @returns A physics radial explosion event, or null
      */
-    public applyRadialExplosionImpulse(origin: Vector3, radius: number, strength: number, falloff: PhysicsRadialImpulseFalloff = PhysicsRadialImpulseFalloff.Constant): Nullable<PhysicsRadialExplosionEvent> {
+    public applyRadialExplosionImpulse(origin: Vector3, eventOptions: PhysicsRadialExplosionEventOptions): Nullable<PhysicsRadialExplosionEvent> {
```
deltakosh: please mention that change in the breaking change

src/Physics/physicsHelper.ts
```diff
@@ -106,12 +112,10 @@ export class PhysicsHelper {
...
      * Creates a gravitational field
      * @param origin the origin of the explosion
-     * @param radius the explosion radius
-     * @param strength the explosion strength
-     * @param falloff possible options: Constant & Linear. Defaults to Constant
+     * @param eventOptions the options of gravitational field
      * @returns A physics gravitational field event, or null
      */
-    public gravitationalField(origin: Vector3, radius: number, strength: number, falloff: PhysicsRadialImpulseFalloff = PhysicsRadialImpulseFalloff.Constant): Nullable<PhysicsGravitationalFieldEvent> {
+    public gravitationalField(origin: Vector3, eventOptions: PhysicsRadialExplosionEventOptions): Nullable<PhysicsGravitationalFieldEvent> {
```
deltakosh: same for all other breaking changes

src/Physics/physicsHelper.ts
deltakosh: This will also require a documentation update btw

deltakosh: other option will be to NOT introduce breaking changes but instead provide a second signature:
public applyRadialExplosionImpulseEx(origin: Vector3, eventOptions: PhysicsRadialExplosionEventOptions):

and then have the first one calling that new one

src/Physics/physicsHelper.ts
```diff
@@ -174,62 +173,67 @@ export class PhysicsHelper {
...
 
         return event;
     }
 }
 
+export class PhysicsHelperOptions {
+    /**
+     * If set to true, it will use the native raycasting
+     */
+    public useNativePhysicsRaycastingIfAvailable: boolean = true;
```
deltakosh: I would prefer to have that at plugin level so users do not have to directly know about it (so each plugin can define if it supports it or not)

src/Physics/physicsHelper.ts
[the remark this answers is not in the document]
bobalazek: Yes, I know. That's what I mentioned above. The thing is, for now, extending it would work (at least for `applyRadialExplosionImpulse` & co., as it only has 4 attributes (MAYBE more in future), but the votex even now has 7 attributes, and that's probably too much, especially as you may want to edit only some.
Because of that, it makes more sense to ever event just have an options variable added, instead of the vortex event having options, but other event having each parameter separately.

bobalazek: I did mention it here (https://github.com/BabylonJS/Babylon.js/pull/5935/files#diff-683d2ac1d77ed946cbfda3917f406c8dR264), but maybe I should be more specific.
Oh yeah, will update the documentation, once we are clear that everything else is fine.

I wasn't really sure if people even use this, so that's why I was somewhat not sure, if it was that important, if it had a breaking change - but yeah, I agree, should be avoided if possible...
What about if the second parameter (here and in all other calls is `radius: number | PhysicsRadialExplosionEventOptions` and create a private second function `applyRadialExplosionImpulseEx` - would that kinda fit into the context?

deltakosh: yep I like that! it fixes the breaking change issue

bobalazek: Ok, refactored that part!

bobalazek:
Ok, so I removed the raycasting inside the physics helper (at least for now). Not because of performance issues (I haven't compared the performance yet), but because of consistency issues.
The thing is, that some a lot of calls then still use the `intersectsMesh` from the native (babylon) lib, while the raycasting of the same thing (well, finding the contact point) would be from the plugin lib. Doesn't make sense, so I'll drop that for now. May come back to that in the future.

deltakosh:
makes sense!

src/Physics/physicsHelper.ts
```diff
@@ -411,20 +436,19 @@ export class PhysicsUpdraftEvent {
...
      * @param _scene BabylonJS scene
      * @param _origin The origin position of the updraft
-     * @param _radius The radius of the updraft
-     * @param _strength The strength of the updraft
-     * @param _height The height of the updraft
-     * @param _updraftMode The mode of the updraft
+     * @param _options The options for the updraft event
      */
-    constructor(private _scene: Scene, private _origin: Vector3, private _radius: number, private _strength: number, private _height: number, private _updraftMode: PhysicsUpdraftMode) {
+    constructor(private _physicsHelper: PhysicsHelper, private _scene: Scene, private _origin: Vector3, private _options: PhysicsUpdraftEventOptions) {
```
deltakosh: This is also a breaking change (perhaps just moving the helper at the end then

src/Physics/physicsHelper.ts
```diff
@@ -694,14 +716,107 @@ export class PhysicsVortexEvent {
...
+
+    /**
+     * The mode for the the updraft.
+     */
+    updraftMode: PhysicsUpdraftMode = PhysicsUpdraftMode.Center;
+}
+
+/**
+ * Options fot the vortex event
+ * @see https://doc.babylonjs.com/how_to/using_the_physics_engine
```
deltakosh: is that the right link?

src/Physics/physicsHelper.ts
bobalazek: That was rather a consistency fix, because only the `PhysicsGravitationalFieldEvent` had the ˙PhysicsHelper` in the constructor, while other events didn't have it. Either way, it would be a breaking change, if we want consistency with the signatures.
Now, the next thing here is, that you are not really supposed to ever call that by your own, but rather via the calls (`applyRadialExplosionImpulse`, ...), as it does some of the actual processing happens there.
So I guess the best option would be to just not export/expose these classes publicly, as they aren't supposed to be public? Sorry, didn't think of that before.

src/Physics/physicsHelper.ts
bobalazek: I think so? Was copy/pasting from other classes. Still, need to update the documentation there soon, when everything else is now fine.

deltakosh: no worries :) I'm fine to makr then private then

src/Physics/physicsHelper.ts
deltakosh: no worries :) I'm fine to makr then private then

src/Physics/physicsHelper.ts
deltakosh: I think you should point to this one: https://doc.babylonjs.com/how_to/particlehelper

bobalazek: Particle helper? For what? The whole `PhysicsHelper` does not do anything on the graphics part. Only physics.
So that's what it does: https://playground.babylonjs.com/index.html#TVUDC1#4

You'll still need to do the visual effect manually.

src/Physics/physicsHelper.ts
bobalazek: Ok, that's fixed now!

src/Physics/physicsHelper.ts
deltakosh: lol bad copy /paste
https://doc.babylonjs.com/how_to/using_the_physics_engine#further-functionality-of-the-impostor-class

bobalazek: Oh, well, that one makes more sense now :)
Sure! Updated the documentation stuff in the helper!

PR #5947, opened 2019-02-20
Updated polgyonMesh.ts

artfiedler:
Changed the scene to optional and updated the parameter doc

Sorry you guys beat me to merging, looks like one last tiny change here


src/Meshes/polygonMesh.ts
```diff
@@ -174,10 +174,10 @@ export class PolygonMeshBuilder {
...
      * @param contours Path of the polygon
-     * @param scene scene to add to
+     * @param scene scene to add to when creating the mesh
      * @param earcutInjection can be used to inject your own earcut reference
      */
-    constructor(name: string, contours: Path2 | Vector2[] | any, scene: Scene, earcutInjection = earcut) {
+    constructor(name: string, contours: Path2 | Vector2[] | any, scene?: Scene, earcutInjection = earcut) {
         this.bjsEarcut = earcutInjection;
         this._name = name;
         this._scene = scene;
```
deltakosh: you should the set it to this._scene = scene || Engine.GetLastCreatedScene()

And then your _scene is not undefined

artfiedler: even though scene won't be used at all?

artfiedler:
::smh:: git can be seriously annoying sometimes

artfiedler:
scene || Engine.LastCreatedScene

engine.LastCreatedScene ends up being Nullable<Scene> same as | undefined

artfiedler:
hrmm not letting me revert back to before it updated all those files guess ill just update the current state to the previous state that should solve :/

src/Meshes/polygonMesh.ts
deltakosh: yep

deltakosh:
Ok to merge?

artfiedler:
Appears to be

PR #5956, opened 2019-02-22
Allow multiple spaces for vertex position/color definitions.

brianzinn:
(https://forum.babylonjs.com/t/invisible-imported-obj-file-with-mtl/1523/9)
This is a fix for a bug introduced in 4.0.  No docs or what's new update needed.


deltakosh:
Thanks a lot

PR #5959, opened 2019-02-22
hdrTexture availabled in all formats

Vinc3r:
At least from the official playground we should have all variants for all playground hdr textures. They are made from .hdr to .dds using IBL Baker, to .env using sandbox&inspector


deltakosh:
excellent!

PR #5961, opened 2019-02-22
STL ASCII

Pryme8:
STL Ascii Export Support.

I need to still add array iteration to this but in the state its in it should work to at least test my first TS script ^_^.


deltakosh:
Can you revert all the file from dist?

Pryme8:
how do I do that?

checkout the dist folder from the remote?

Pryme8:
I ended up copying all the files from the public repository to overwrite my local dist.

Hope that was what you were meaning!

deltakosh:
not exactly, I would like to not see the file at all in the PR (you need to git revert them before updating the PR)

Pryme8:
Ok I will do that now!

PR #5963, opened 2019-02-22
STL Export ASCII

Pryme8:
I reverted the .gitIgnore file changes.

I think this is correct. >_<

### DIFF
--- a/serializers/src/stl/stlSerializer.ts
+++ b/serializers/src/stl/stlSerializer.ts
@@ -8,7 +8,7 @@ import { Vector3 } from "babylonjs/Maths/math";
 export class STLExport {
      /**
      * Exports the geometry of a Mesh array in .STL file format (ASCII)
-     * @param mesh defines the list of meshes to serialize
+     * @param mesh defines the meshe to serialize
      * @param fileName Name of the file when downloaded.
      * @param download triggers the automatic download of the file.
      * @returns the ASCII STL format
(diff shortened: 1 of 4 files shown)


serializers/src/stl/stlSerializer.ts
```diff
@@ -0,0 +1,57 @@
...
+import { VertexBuffer } from "babylonjs/Meshes/buffer";
+import { Vector3 } from "babylonjs/Maths/math";
+
+/**
+ * Class for generating STL data from a Babylon scene.
+ */
+export class STLExport {
+     /**
+     * Exports the geometry of a Mesh array in .STL file format (ASCII)
+     * @param mesh defines the list of meshes to serialize
```
deltakosh: This is not a list right?

Pryme8: good catch, fixing now.

Pryme8:
wait whoa, so this was correct?

How would I test it on my end, I could never get it to come up on any of my builds... If I can connect this blank then I can start flying on contributions.

deltakosh:
It did :)
the best way is to create a sample code in localdev and use vsCode to debug it

deltakosh:
it will soon be available in the PG

Pryme8:
hmmm, Odd I thought I tried it in the localdev and was getting no response from it... I must have been over looking something.

deltakosh:
yeah I end up having some issues with it recently. @sebavan  will check why soon :)

PR #5973, opened 2019-02-26
Update to STL Serializer - Binary Support

Pryme8:
Now supports binary and is a little more robust.
Can have a list of meshes passed to it now.

### DIFF
--- a/serializers/src/stl/stlSerializer.ts
+++ b/serializers/src/stl/stlSerializer.ts
@@ -52,7 +52,7 @@ export class STLExport {
         if (binary) {
             for (let i = 0; i < meshes.length; i++) {
                 let mesh = meshes[i];
-                faceCount += mesh.getIndices().length || 0;
+                faceCount += (mesh.getIndices() || []).length;
             }
 
             let bufferSize = 84 + (50 * faceCount);
--- a/serializers/src/stl/stlSerializer.ts
+++ b/serializers/src/stl/stlSerializer.ts
@@ -52,8 +52,8 @@ export class STLExport {
         if (binary) {
             for (let i = 0; i < meshes.length; i++) {
                 let mesh = meshes[i];
-				let indices = mesh.getIndices();
-				faceCount +=indices ? indices.length : 0;
+                let indices = mesh.getIndices();
+                faceCount += indices ? indices.length : 0;
             }
 
             let bufferSize = 84 + (50 * faceCount);
--- a/serializers/src/stl/stlSerializer.ts
+++ b/serializers/src/stl/stlSerializer.ts
@@ -12,102 +12,91 @@ export class STLExport {
     * @param download triggers the automatic download of the file.
     * @param fileName changes the downloads fileName.
     * @param binary changes the STL to a binary type.
+	* @param isLittleEndian toggle for binary type exporter.
     * @returns the STL as UTF8 string
     */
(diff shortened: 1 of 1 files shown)


serializers/src/stl/stlSerializer.ts
```diff
@@ -37,21 +45,85 @@ export class STLExport {
...
-            return data;
+            }			
+		}	
+		
+		data += 'endsolid exportedMesh';
+		
+		}else{	
+		
+			//Adapted from https://gist.github.com/paulkaplan/6d5f0ab2c7e8fdc68a61			
+			let faceCount = 0;
```
deltakosh: identation seems off all over the file?

Pryme8: It does look odd on here which is crazy because on my local version it looks fine.
image

Pryme8: I'm not sure how else to indent it?

Also do not merge it yet! I just found a error.

deltakosh: Please use 4 spaces (does not seem the case here)

Pryme8: Ahh, ok yeah I was using tabs. Ill fix asap.

deltakosh: :D

deltakosh:
please run 'npm run build' as you have a lot of linting issues

Pryme8:
I wont be able to make the change that I wanted to make before the merge for a few days because I am out of town on business.  Is it ok that its left open or should I just close it till I get a chance to do that.

Technically its functional as there was just a few redundancies in the script that could be cleaned up.

deltakosh:
No worries. We can keep it open

Pryme8:
still have a commit to make on this, sorry jumped the gun on that last one.

Pryme8:
Im having this error come up:
Message:
    [at-loader] ..\..\serializers\src\stl\stlSerializer.ts:55:30
    TS2531: Object is possibly 'null'
and am just trying to find a work around.  I thought having a || 0 after would fix it but it seems to not have.

deltakosh:
That is a good exercice then :)

serializers/src/stl/stlSerializer.ts
```diff
@@ -52,7 +52,7 @@ export class STLExport {
         if (binary) {
             for (let i = 0; i < meshes.length; i++) {
                 let mesh = meshes[i];
-                faceCount += mesh.getIndices().length || 0;
+                faceCount += (mesh.getIndices() || []).length;
```
deltakosh: what about:
```
let indices = mesh.getIndices();
faceCount +=indices ? indices.length : 0;
```

This way we avoid instanciating an empty array

Pryme8:
That’s is very elegant. I will update that when I get home!

I also understand the reasoning behind that. Thank you for the alternative perspective.

Pryme8:
I'm not to sure what is causing the Travis CI build to fail. Any ideas?

deltakosh:
Because there is no wha'ts new update:
https://github.com/BabylonJS/Babylon.js/pull/5973/checks?check_run_id=71924985

Pryme8:
How do I write that up?

deltakosh:
You already did it so no worries

PR #5986, opened 2019-02-28
do not show outline under mesh when it becomes transparent

TrevorDev:
https://forum.babylonjs.com/t/by-reducing-the-visibility-a-green-model-appears/1190/10

https://playground.babylonjs.com#G2F8LN#34

### DIFF
--- a/src/Rendering/outlineRenderer.ts
+++ b/src/Rendering/outlineRenderer.ts
@@ -278,14 +278,7 @@ export class OutlineRenderer implements ISceneComponent {
         // Outline - step 1
         this._savedDepthWrite = this._engine.getDepthWrite();
         if (mesh.renderOutline) {
-            // Cache stencil state
-            var previousStencilBuffer = this._engine.getStencilBuffer();
-            var previousStencilFunction = this._engine.getStencilFunction();
-            var previousStencilMask = this._engine.getStencilMask();
-            var previousStencilOperationPass = this._engine.getStencilOperationPass();
-            var previousStencilOperationFail = this._engine.getStencilOperationFail();
-            var previousStencilOperationDepthFail = this._engine.getStencilOperationDepthFail();
-            var previousStencilReference = this._engine.getStencilFunctionReference();
+            this._engine.cacheStencilState();
 
             var material = subMesh.getMaterial();
             if (material && material.needAlphaBlending) {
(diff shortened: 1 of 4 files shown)


src/Rendering/outlineRenderer.ts
```diff
@@ -274,9 +278,44 @@ export class OutlineRenderer implements ISceneComponent {
         // Outline - step 1
         this._savedDepthWrite = this._engine.getDepthWrite();
         if (mesh.renderOutline) {
+            // Cache stencil state
+            var previousStencilBuffer = this._engine.getStencilBuffer();
```
deltakosh: this could be an engine function (like store / restore stencil state)

src/Rendering/outlineRenderer.ts
```diff
@@ -274,9 +278,44 @@ export class OutlineRenderer implements ISceneComponent {
...
+                this._engine.setStencilFunction(Constants.ALWAYS);
+                this._engine.setStencilMask(OutlineRenderer._StencilReference);
+                this._engine.setStencilFunctionReference(OutlineRenderer._StencilReference);
+                this.render(subMesh, batch, /* This sets offset to 0 */ true);
+            }
+
+            // Draw the outline using the above stencil if needed to avoid drawing within the mesh
+            this._engine.setColorWrite(true);
             this._engine.setDepthWrite(false);
+            this._engine.setStencilFunction(Constants.NOTEQUAL);
```
deltakosh: what if the mesh is not transparent? do we need all that stencil stuf fin that case?

src/Rendering/outlineRenderer.ts
```diff
@@ -278,14 +278,7 @@ export class OutlineRenderer implements ISceneComponent {
...
         if (mesh.renderOutline) {
-            // Cache stencil state
-            var previousStencilBuffer = this._engine.getStencilBuffer();
-            var previousStencilFunction = this._engine.getStencilFunction();
-            var previousStencilMask = this._engine.getStencilMask();
-            var previousStencilOperationPass = this._engine.getStencilOperationPass();
-            var previousStencilOperationFail = this._engine.getStencilOperationFail();
-            var previousStencilOperationDepthFail = this._engine.getStencilOperationDepthFail();
-            var previousStencilReference = this._engine.getStencilFunctionReference();
+            this._engine.cacheStencilState();
```
deltakosh: even the caching should be under needAlphaBlending

src/Rendering/outlineRenderer.ts
TrevorDev: nope, if its not transparent its fine to draw behind

deltakosh: so we should do nothing with the stencil then in that case

PR #5995, opened 2019-03-04
Added Rope as a soft body to AmmoJSPlugin

BabylonJSGuide:
Tests 
https://adoring-kirch-5b50fe.netlify.com/ropetest.1.html
https://adoring-kirch-5b50fe.netlify.com/ropetest.2.html
https://adoring-kirch-5b50fe.netlify.com/ropetest.3.html
https://adoring-kirch-5b50fe.netlify.com/ropetest.4.html

Docs will follow later in week.


src/Physics/Plugins/ammoJSPlugin.ts
```diff
@@ -219,10 +222,53 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
...
+            impostor.object = ShapeBuilder.ExtrudeShape("ext", {shape: shape, path: path, instance: <Mesh>object});
+        }
+
+    }
+
+    /**
+     * Update babylon mesh vertices vertices to match physics world softbody or cloth
+     * @param impostor imposter to match
+     */
+    public notRopeStep(impostor: PhysicsImpostor): void {
```
TrevorDev: Could this be made private or do you expect it to be called externally? Is there a better name for this method eg. _softBodyStep?

BabylonJSGuide: You are right afterSoftStep, ropeStep and notRopeStep can all be private. In the docs I use the term Soft Body to include all three types softbody (3D), cloth(2D) and rope(1D) so the name _softBody would imply rope as well. I have therefore gone for _softbodyOrClothStep  so as to be clearer.

deltakosh:
Thanks a lot John!

PR #6013, opened 2019-03-06
Add getShaderName Function

MackeyK24:
This will allow PBRMaterial subclasses to specify custom pbr shaders by overriding getShaderName


src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -774,6 +774,13 @@ export abstract class PBRBaseMaterial extends PushMaterial {
         return "PBRBaseMaterial";
     }
 
+    /**
+     * Gets the name of the material shader.
```
sebavan: @returns is missing so it won t pass the build. Aside from this it is all good

sebavan: yup: https://travis-ci.com/BabylonJS/Babylon.js/jobs/182794957#L689

PR #6020, opened 2019-03-07
add missing typedocs to fix build

TrevorDev:


src/Materials/Textures/renderTargetTexture.ts
```diff
@@ -1008,7 +1013,7 @@ export class MultiviewRenderTarget extends RenderTargetTexture {
      * @param scene scene used with the render target
      * @param size the size of the render target (used for each view)
      */
-    constructor(public scene: Scene, size: number | { width: number, height: number } | { ratio: number } = 512) {
+    constructor(private scene: Scene, size: number | { width: number, height: number } | { ratio: number } = 512) {
```
sebavan: scene should be _scene if private ?

Issue #6019, opened 2019-03-07
AssetContainer.dispose removes sounds outside of the container


nbduke:
# Bugs
## Repro steps 
Navigate to [this playground](https://www.babylonjs-playground.com/#VGJIGT). You should immediately hear the music start playing, but after three seconds, we dispose of an asset container containing _only_ the sphere and ground meshes in the scene. Unfortunately, this causes the music to be disposed as well since it's part of the serializable components of the scene.

This feels like a bug: I don't expect objects in the scene outside of the asset container to be removed/disposed when I remove/dispose the asset container.

nbduke:
@TrevorDev I was looking at your commits, and I'm concerned that `AssetContainer.dispose` will dispose the scene's `environmentTexture` even if the container did not possess that texture, which feels like a similar manifestation of the disappearing sound bug.

deltakosh:
Agree I will update that line from:
```
if (this.scene.environmentTexture) {
            this.scene.environmentTexture.dispose();
}
```

to 
```
if (this.environmentTexture) {
            this.environmentTexture.dispose();
        }
```
Thoughts?

nbduke:
That looks correct to me.

deltakosh:
Pushing a fix with the nightly

Issue #6014, opened 2019-03-06
SceneLoader.LoadAssetContainer should grab environment texture


nbduke:
# Feature request
Debatable whether this is a bug or not, so I made it a feature request. If you call `BABYLON.SceneLoader.LoadAssetContainerAsync` with a .babylon file that specifies an environment texture, this texture does not get loaded into the asset container. I'm guessing this is because `AssetContainer` does not share `Scene`'s `environmentTexture` property. 

## Possible solutions
Any of the following solutions would be acceptable. I would probably prefer (1).
1. Move the `environmentTexture` property into `AbstractScene` and make sure `SceneLoader` populates it on the `AssetContainer`.
2. If the .babylon file has an environment texture, load it into either the `textures` array or the `reflectionProbes` array of the `AssetContainer`.

nbduke:
Thinking about it some more, only solution (1) makes it straightforward to implement `addAllToScene`, since if the environment texture is just sitting in the `textures` array, it's not clear how the asset container will know to make it the environment texture of the scene.

deltakosh:
Unfortunately 1 would imply a breaking change so this is not acceptable.

I would prefer having an environment texture property on the asset container that will be used when user call addAll

@trevordev thoughts?

nbduke:
Check out #6019 for a related issue with the `AssetContainer`.

TrevorDev:
Yup, 1 sounds right to me as well

Working on fixes for these issues now

PR #6021, opened 2019-03-07
Asset container support environment texture, do not dispose all sounds in scene when AC is disposed

TrevorDev:
- **https://github.com/BabylonJS/Babylon.js/issues/6014**
test scene: https://www.babylonjs-playground.com#3YL8W4 
and add 
```
,"environmentTexture":"https://assets.babylonjs.com/environments/environmentSpecular.env"
```
to skull.babylon

 - **https://github.com/BabylonJS/Babylon.js/issues/6019**
Test scene: https://www.babylonjs-playground.com/#VGJIGT


nbduke:
@TrevorDev @deltakosh  I tried out the alpha.33 release that has the fix for this issue, and I discovered a related problem: if an `AssetContainer`'s environment texture is null/undefined and you call `addAllToScene`, it will undo the scene's environment texture.

I think the environment texture should be treated the same as the other fields (e.g. `meshes`, `textures`, etc.), where nothing in the scene changes if the field is empty in the `AssetContainer`.

TrevorDev:
@nbduke Sorry about that, sounds right to me, ill get a PR asap

PR #6033, opened 2019-03-08
Consistent Custom Shader Name Resolve Support

MackeyK24:
Add consistent custom shader support for PBRMaterialBase And StandardMaterial. Will now allow the creation of a Material Library PBRCustomMaterial with shader define tags


src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1227,8 +1233,13 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             maxSimultaneousLights: this._maxSimultaneousLights
         });
 
-        var join = defines.toString();
-        return engine.createEffect(this.getShaderName(), <EffectCreationOptions>{
+        let shaderName: string = this.getShaderName();
+        if (this.customShaderNameResolve) {
```
deltakosh: Do we need both ways then? I guess the customShaderNameResolve should be enough

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -169,7 +169,6 @@ class PBRMaterialDefines extends MaterialDefines
     public SAMPLER3DBGRMAP = false;
     public IMAGEPROCESSINGPOSTPROCESS = false;
     public EXPOSURE = false;
-    public MULTIVIEW = false;
```
deltakosh: Why are you removing the Multiview code?

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1150,10 +1160,6 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             fallbacks.addFallback(fallbackRank++, "MORPHTARGETS");
         }
 
-        if (defines.MULTIVIEW) {
```
deltakosh: ?

MackeyK24: Im not removing code... What happened is I COPIED the whole babylon js project and my my changes to the copy. Once i got it working ... I copied the whole pbrMaterialBase and standardMaterial files back to my repo then uploaded that. I think there was changes in my last copy of repo and current.

Can you please disregard this pull and i will catch up my repo and just put the changes right there and upload...

Why the getShaderName and customShaderNameResolve ???

I think its cleaner and at first "pbr" and "default" where ALWAYS hard code in the script. Just me but i think the main shader program should be hard coded like was being used in the create effect line. The getShaderName as a nice clean way around that. 

StandardMaterial as least had the customShaderNameResolve function that updated that hard coded "default" value... But i just didnt like that.

PBRMaterial never even had a customShaderNameResolve function so i added getShaderName.

I was trying to make both guys consistent and NOT have to require the subclasses to call customShaderNameResolve but simple provide a override of getShaderName... Kind how getClassName is used through out the framework. I thinks it cleaner and i dont have to create that addition builder function that you assign to customShaderNameResolve.

Eiather way... i can abstract all that a way and my subclasses can override the getShaderName i will add to my base classes

I really like my subclasses for simple to look like this (weather i am using the more complex customShaderNameResolve or not)

```
class AlbedoMaterial extends BABYLON.UniversalAlbedoMaterial {
    public constructor(name: string, scene: BABYLON.Scene) {
        super(name, scene);
    }
    public getClassName(): string {
        return "AlbedoMaterial";
    }
    public getShaderName(): string {
        return "mp_pbr";
    }
}
```


But anyways.... please dump this whole PR... i will make another cleaner PR and remove getShadreName of you dont like that. since we now will have customShaderNameResolve in both base classes... You let me know ?

PR #6037, opened 2019-03-08
keep back compat for gizmos when created on a custom utility layer

TrevorDev:
https://forum.babylonjs.com/t/gizmo-positiongizmo-arrows-are-black-v4/1782


src/Rendering/utilityLayerRenderer.ts
```diff
@@ -17,6 +17,20 @@ export class UtilityLayerRenderer implements IDisposable {
...
+    /**
+     * @hidden
+     * Light which used by gizmos to get light shading
+     */
+    public _getSharedGizmoLight(): HemisphericLight {
+        if (!this._sharedGizmoLight) {
+            this._sharedGizmoLight = new HemisphericLight("shared gizmo light", new Vector3(0, 1, 0), this.utilityLayerScene);
+            this._sharedGizmoLight.intensity = 2;
+            this._sharedGizmoLight.groundColor = Color3.Gray();
+            this._sharedGizmoLight.includedOnlyMeshes = [new AbstractMesh("", this.utilityLayerScene)];
```
deltakosh: Why creating an empty mesh ?

TrevorDev: Because it needs at least one mesh or else anything added to the scene will get the light

PR #6042, opened 2019-03-11
disable mouse interaction on orientation camera if sensor is active

TrevorDev:
https://forum.babylonjs.com/t/createdefaultvrexperience-on-android-chrome-blue-flashes-and-camera-jumps/1813

testPG: https://playground.babylonjs.com/#NWW1K1#10


src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -42,6 +51,10 @@ export class FreeCameraDeviceOrientationInput implements ICameraInput<FreeCamera
     private _beta: number = 0;
     private _gamma: number = 0;
 
+    /**
+     * @hidden
+     */
+    public _onDeviceOrientationChangedObservable = new Observable<void>();
```
deltakosh: needs to be clear at dispose time

TrevorDev: This class is not disposable so I added a clear for this when the attached camera is disposed.

Issue #6051, opened 2019-03-12
AssetContainer appears to leak textures and materials


nbduke:
# Bugs
## Repro steps
1. Navigate to this playground: https://www.babylonjs-playground.com/#FHHM51#4
2. Open the developer console and wait several seconds for the model to load, then two more seconds for it to be disposed.

The model is loaded into an `AssetContainer` using `SceneLoader.LoadAssetContainerAsync`. In the console, you'll see the numbers of meshes, materials, and textures in the scene logged before loading the container, after adding everything to the scene, and after removing/disposing everything in the container.

I would expect the numbers before and after to be exactly the same; i.e., `AssetContainer.addAllToScene()` followed by `AssetContainer.removeAllFromScene()` should leave the scene unchanged in any measurable way. This is not the case, however. Some materials and textures are left in the scene.

TrevorDev:
@nbduke thanks for the good repro. It should not leak anymore after the above PR, however on the first initialization of any pbrMaterial a cached Material will be created so after adding and removing any number of gltfs, it should show 0 materials and 1 texture.

PR #6053, opened 2019-03-13
add texture/materials to assetContainer

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/6051

TestPG: #Q96J2D

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -207,6 +207,7 @@
 - Add missing dependencies for files to support including them from a direct path (eg. import "@babylonjs/core/Helpers/sceneHelpers";) ([TrevorDev](https://github.com/TrevorDev))
 - AssetContainer should not dispose objects it doesn't contain, Support for environmentTexture ([TrevorDev](https://github.com/TrevorDev))
 - Fix `mesh.visibility` not working properly when certain material properties are set that changes the interpretation of alpha (e.g. refraction, specular over alpha, etc.) ([bghgary](https://github.com/bghgary))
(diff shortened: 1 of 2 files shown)


nbduke:
One thing I'm curious about with the `LoadAssetContainer` methods: should *any* assets be put into the scene before calling `addAllToScene` on the returned container? Or is that impossible to guarantee?

TrevorDev:
@nbduke  Are you asking if LoadAssetContainer should have any side effects before calling addAllToScene? It should not, so if you are aware of cases where this happens, let me know. Loading an assetContainer should only populate the assetContainer and adding and removing that asset container assets to the scene should only effect the assets in the assetcontainer.

nbduke:
Yeah, you can actually see that happening if you make a slight tweak to the playground I put in the bug. Check this out: https://www.babylonjs-playground.com/#FHHM51#6

TrevorDev:
@nbduke, after double checking this, this PR should make things better for this case eg. mesh materials/textures should not be be in the scene but I still do see skeleton textures populated on the scene after calling LoadAssetContainer. I could add special logic to remove these but this texture is internal to skeleton so if a user wants to add only skeletons back to the scene for some reason I'm afraid of breaking something for them so I think I'm going to leave those alone unless you have an issue with this. Those skeleton textures will be removed if the container is disposed like in your example.

nbduke:
Ok, thanks. Ideally loading assets into a container would have no side effects, but I agree that the highest priority is that AssetContainer cleans up after itself, which your PR will fix.

PR #6061, opened 2019-03-14
multiview component

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/6015


src/scene.ts
```diff
@@ -1046,7 +1047,8 @@ export class Scene extends AbstractScene implements IAnimatable {
     public proceduralTexturesEnabled = true;
 
     // Private
-    private _engine: Engine;
+    /** @hidden */
+    public _engine: Engine;
```
sebavan: Why not using getEngine() when you need engine ?

src/scene.ts
```diff
@@ -870,7 +870,8 @@ export class Scene extends AbstractScene implements IAnimatable {
     /** All of the active cameras added to this scene. */
     public activeCameras = new Array<Camera>();
 
-    private _activeCamera: Nullable<Camera>;
+    /** @hidden */
+    public _activeCamera: Nullable<Camera>;
```
sebavan: Why not relying on getActiveCamera()

src/scene.ts
```diff
@@ -1112,18 +1114,19 @@ export class Scene extends AbstractScene implements IAnimatable {
     /** @hidden */
     public _activeAnimatables = new Array<Animatable>();
 
-    private _transformMatrix = Matrix.Zero();
-    private _transformMatrixR = Matrix.Zero();
+    /** @hidden */
+    public _transformMatrix = Matrix.Zero();
```
sebavan: Not sure if the previous comment applies for matrices as well ?

src/scene.ts
```diff
@@ -2585,16 +2578,8 @@ export class Scene extends AbstractScene implements IAnimatable {
             Frustum.GetPlanesToRef(this._transformMatrix, this._frustumPlanes);
         }
 
-        if (viewR && projectionR) {
-            viewR.multiplyToRef(projectionR, Tmp.Matrix[0]);
-            Frustum.GetRightPlaneToRef(Tmp.Matrix[0], this._frustumPlanes[3]); // Replace right plane by second camera right plane
-        }
-
         if (this._multiviewSceneUbo && this._multiviewSceneUbo.useUbo) {
```
sebavan: Could this be done by callbacks like the other components instead of the test ?

src/scene.ts
```diff
@@ -4169,32 +4155,7 @@ export class Scene extends AbstractScene implements IAnimatable {
         }
 
         if (camera._useMultiviewToSingleView) {
```
sebavan: Same here could it be done by the component callbacks ?

src/scene.ts
TrevorDev: In the component I am doing this._activeCamera = camera._rigCameras[index]; if I use activeCamera instead, a camera changed notification occurs which shouldnt happen when it changes just for rendering.

sebavan: Makes total sense here 👍

src/scene.ts
TrevorDev: I am finding using the observables a bit tricky as for multiview it's not really doing something extra, it is doing something different instead. I could add something like a renderMultiviewToSingleViewObservable but then it's not any different than a function call.

sebavan:
can not reply in the conversation for some reasons. Understood for the callbacks, do not bother then

PR #6064, opened 2019-03-15
no not overwrite scene environment texture with null

TrevorDev:
https://github.com/BabylonJS/Babylon.js/pull/6021


nbduke:
Thanks, Trevor!

PR #6073, opened 2019-03-19
Arc rotate camera changes/fixes

thscott:
Pushing some modifications/fixes  I'd made in my custom camera class (which extends arcRotateCamera) to arcRotateCamera.

First, for non Y-up up-vectors, change to only calculate the rotation matrix when the up vector changes, as opposed to each time ```_getViewMatrix``` is called. This also uses a new function I've added to the Matrix class to more efficiently get the rotation matrix to align two vectors.

Second, fixed ```rebuildAnglesAndRadius``` so that it also returns correct results when the up-vector is not (0,1,0), and also fixed a divide by zero error I came across (when setting the camera position to look directly along the up axis).

In fixing the above, I can't understand the point of ever calling ```rebuildAnglesAndRadius``` without setting the first parameter to false. The default value of true means that the position, target and alpha, beta, radius will all agree after the call to ```_getViewMatrix``` (unless I've overlooked something). Made a small pg [here](https://www.babylonjs-playground.com/debug.html#I3V0GW#1).

I'm not sure if the ```serialize``` of ```parse``` functions for the camera need to be changed following these updates.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -135,6 +135,8 @@
 - Added customShaderNameResolve to PBRMaterialBase to allow subclasses to specify custom shader information [MackeyK24](https://github.com/mackeyk24))
 - Added PBRCustomMaterial to material library to allow easy subclassing of PBR materials [MackeyK24](https://github.com/mackeyk24))
 - Added `auto-exposure` support in `StandardRenderingPipeline` when `HDR` is enabled ([julien-moreau](https://github.com/julien-moreau))
+- Added `Matrix.RotationAlignToRef` method to obtain rotation matrix from one vector to another ([sable](https://github.com/thscott))
+- ArcRotateCamera will now cache the necessary matrices when modifying its upVector, instead of calculating them each time they're needed ([sable](https://github.com/thscott))
 
 ### OBJ Loader
 - Add color vertex support (not part of standard) ([brianzinn](https://github.com/brianzinn))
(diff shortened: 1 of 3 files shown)


deltakosh:
Excellent PR! Thanks a lot for that
Some linting errors though: https://github.com/BabylonJS/Babylon.js/pull/6073/checks?check_run_id=80445708

Calling `rebuildAnglesAndRadius` with true (by default) is mostly used by setTarget to ensure that the camera stay stills based on current configuration

In your PG, you set `camera.position` manually but the camera has no position per se. it only has alpha, beta and radius. To effectively change the position you need to call `camera.setPosition`:
https://www.babylonjs-playground.com/debug.html#I3V0GW#2

The last NAN is expected as target and position are at the same place hence radius = 0

thscott:
I'll fix those linting errors later today.

Regarding the NaN in the pg, was meant to set the position to along the y axis like [this](https://www.babylonjs-playground.com/debug.html#I3V0GW#3), which just demonstrates the divide by zero bug I fixed in the pr (radius is set to 0.0001 in the function anyway, so shouldn't get any NaNs after the fix). 

Regarding ```rebuildAnglesAndRadius``` with default true parameter, the function may as well return after the call to ```_getViewMatrix```, as I can't see how everything below that can change anything. 

With setTarget, there are two ways I can see the function working:
- keeping the camera position, changing the target, and then recalculating alpha/beta/radius (then checking they're within limits).
- keeping alpha/beta/radius constant and modifying the position.

The first way is how (if I remember correctly) the arcRotateCamera used to work, and it then made sense to  call ```rebuildAnglesAndRadius```.
The second is how it currently works, in which case it only (I think) makes sense to call ```_getViewMatrix```.

With the serialisation/parsing, I think the parsing of the upvector should just work (and be set correctly [here](https://github.com/BabylonJS/Babylon.js/blob/9ff5e1fc0b17cb74bb50670ab1f8983bb25f2777/src/Misc/decorators.ts#L373)).

deltakosh:
Ok for the serialization!
for rebuildANglesAndRadius: You are SO right!! I'll make sure to just remove the boolean and call getViewMatrix directly. thanks a lot for that catch!

for the setTarget, as you said it was working like that but then I changed it and I can't remember why..I will probably revert that change as the initial way of working was better

deltakosh:
Ok to merge on your side?

thscott:
Should be now, have just got around to updating what's new

deltakosh:
thanks a lot

PR #6117, opened 2019-03-26
Use web workers for draco compression

### LINKED ISSUE #4146: Investigate Draco compression with web workers
Investigate why using Draco compression with web workers ends up being slower than not using web workers.

https://github.com/bghgary/Babylon.js/tree/draco-web-workers

bghgary:
PG: https://www.babylonjs-playground.com/#ZSZ5G4#1

Good increases in speed on my dev machine (~850ms to ~250ms). On slower machine, it can make things worse if CPU resource bound, but at least it doesn't block the main thread.

Fixes #4146


bghgary:
I did originally. If I do that, I still have to limit it since there isn't enough memory to run the 20 logical processors on my machine. I'm thinking maybe to expose a static deafult variable that deafults to hardware concurrency max'd at 4.

sebavan:
Yup, I would try to limit at a certain number below the max hardware but still checking it in case of older or "smaller" device.

PR #6135, opened 2019-03-30
Inspector support bounding non leaf nodes

TrevorDev:


tests/validation/config.json
```diff
@@ -1,6 +1,12 @@
 ﻿{
   "root": "https://rawgit.com/BabylonJS/Website/master",
   "tests": [
+    {
+      "title": "Nested BBG",
+      "playgroundId": "#ZG0C8B#0",
+      "renderCount": 50,
```
bghgary: why do you need a renderCount?

inspector/src/components/sceneExplorer/entities/sceneTreeItemComponent.tsx
```diff
@@ -206,6 +206,9 @@ export class SceneTreeItemComponent extends React.Component<ISceneTreeItemCompon
                     break;
                 case 4:
                     manager.boundingBoxGizmoEnabled = true;
+                    if(manager.gizmos.boundingBoxGizmo){
```
bghgary:
```suggestion
                    if (manager.gizmos.boundingBoxGizmo) {
```

inspector/src/components/sceneExplorer/entities/sceneTreeItemComponent.tsx
```diff
@@ -206,6 +206,9 @@ export class SceneTreeItemComponent extends React.Component<ISceneTreeItemCompon
                     break;
                 case 4:
                     manager.boundingBoxGizmoEnabled = true;
+                    if(manager.gizmos.boundingBoxGizmo){
+                        manager.gizmos.boundingBoxGizmo.fixedDragMeshScreenSize = true
```
bghgary:
```suggestion
                        manager.gizmos.boundingBoxGizmo.fixedDragMeshScreenSize = true;
```

src/node.ts
```diff
@@ -765,6 +766,13 @@ export class Node implements IBehaviorAware<Node> {
         this._behaviors = [];
     }
 
+    protected _getBoudingInfoMin() {
```
bghgary: I don't understand why this is necessary.

TrevorDev: I moved getHierarchyBoundingVectors to node. But it didn't contain certain things that abstractMesh has like .subMeshes which is why I added these functions, in abstractMesh it overrides them to get the same behavior it originally had when called from an abstract mesh.

src/node.ts
```diff
@@ -779,4 +780,63 @@ export class Node implements IBehaviorAware<Node> {
...
+    public getHierarchyBoundingVectors(includeDescendants = true, predicate: Nullable<(abstractMesh: AbstractMesh) => boolean> = null): { min: Vector3, max: Vector3 } {
+        // Ensures that all world matrix will be recomputed.
+        this.getScene().incrementRenderId();
+
+        this.computeWorldMatrix(true);
+
+        let min: Vector3;
+        let max: Vector3;
+
+        let thisAbstractMesh = ((this as any) as AbstractMesh);
```
bghgary:
```suggestion
        let thisAbstractMesh = (this as Node as AbstractMesh);
```

src/node.ts
```diff
@@ -779,4 +780,63 @@ export class Node implements IBehaviorAware<Node> {
...
+        if (thisAbstractMesh.getBoundingInfo && thisAbstractMesh.subMeshes) {
+            // If this is an abstract mesh get its bounding info
+            let boundingInfo = thisAbstractMesh.getBoundingInfo();
+            min = boundingInfo.boundingBox.minimumWorld;
+            max = boundingInfo.boundingBox.maximumWorld;
+        }else {
```
bghgary:
```suggestion
        } else {
```

src/node.ts
```diff
@@ -779,4 +780,63 @@ export class Node implements IBehaviorAware<Node> {
             }
         }
     }
-}
\ No newline at end of file
+        /**
+     * Return the minimum and maximum world vectors of the entire hierarchy under current mesh
```
deltakosh:
```suggestion
     * Return the minimum and maximum world vectors of the entire hierarchy under current node
```

PR #6137, opened 2019-03-30
added image wrapping to box

BabylonJSGuide:
With ref to https://forum.babylonjs.com/t/jeromes-to-do-list/2176?u=johnk
First two suggestions
BJS CreateBox : the box mesh is one of the brick of BJS, so it could be improved by many ways

    set and return fixed float arrays (positions, normals, indices, uvs) in the VertexData constructor instead of computing them
    Babylon.js/boxBuilder.ts at master · BabylonJS/Babylon.js · GitHub 3
    it would be simpler to set once for all arrays with known values (get them from the console with the current algo)
    var positions = [1, 0, 0, 0, 1, 0, …etc ];
    idem for every array. This would be simply one line of code per array instead of the current geometry computation.
    Challenge level : very easy

    add a parameter alignSides (not sure about the name) to allow to wrap UVs from a continuous texture for 4 box sides with all the 4 images in the same orientation.
    This would solve this issue : Apply Material to Individual Faces - Babylon.js Documentation
    No need then for rotating some of the images in the texture. This would also allow to wrap standard exported UVs like the minercraft-style Mario head that was somewhere in the former forum.
    Thsi feature is expected in the forum for a long time.
    This requires to change a little the current box geometry at construction time by swapping some indices.
    Challenge level : easy

PR to implement these. 
New parameter wrap (default false backward compatible to existing CreateBox) , when true will orientate images along sides to be vertical.  Images on top and bottom match orientation of front and back images.
Also two new parameters (topBaseAt (default 1) and bottomBaseAt (default 0)) allow rotation of top and bottom images so that base of image can be set against sides  0 posz, 1 negz, 2 posx or 3 negx.

Test at https://wizardly-curran-82113b.netlify.com/
Top row of three as per existing CreateBox
Middle row, just wrap set to true
Bottom row wrap and values given to topBaseAt nd bottomBaseAt

Use of skybox not affected. Will update docs when and if PR accepted.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -96,6 +96,7 @@
 - Align `BoundingBox` and `BoundingSphere` API and behavior for clarity and simplicity. As a consequence, the `BoundingBox`'s method `setWorldMatrix` has been removed and the underlying world matrix cannot be modified but by calling `reConstruct` or `update`. ([barroij](https://github.com/barroij))
 - Make sure that `Material.markAsDirty` and all the `markXXXDirty` methods early out when `scene.blockMaterialDirtyMechanism` is true. ([barroij](https://github.com/barroij))
 - Add updateUpVectorFromRotation to target camera to allow the up vector to be computed from rotation ([TrevorDev](https://github.com/TrevorDev))
+- Added `wrap` boolean parameter to `CreateBox` options to orientate images vertically on box sides ([JohnK](https://github.com/BabylonJSGuide))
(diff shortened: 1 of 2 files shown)


deltakosh:
This is really good 
I bet @jbousquie  will appreciate :D

jbousquie:
Excellent :-D
I like the implementation and the work done, I appreciate ;-)

jbousquie:
thank you, John :-)

Le lun. 1 avr. 2019 à 02:27, David Catuhe <notifications@github.com> a
écrit :

Issue #6103, opened 2019-03-22
Material SideOrientation/Normal flipped during glb export


CadsoftBrianW:
# Bugs

- Bug repro on [playground](https://playground.babylonjs.com): 
   https://www.babylonjs-playground.com/#95MJI8#22
- Expected result: Material orientation is correct
- Current result: Material orientation is incorrect 

Forum: 
https://forum.babylonjs.com/t/material-sideorientation-normal-flipped-during-glb-export/2046

deltakosh:
This one if for @drigax

Drigax:
Looks like we currently only reverse the winding when exporting a mesh in a Left Hand coordinate system in Babylon with Clockwise triangle winding to GLTF, which is strictly Right Hand Coordinate System with Counterclockwise triangle winding.

That should explain why this only reproduces when we setup the scene with right hand coordinate system, as our materials are setup to have clockwise side orientation by default.

I believe that changing our exporter to reverse the triangle winding when we are either in a left hand coordinate system, or are working with a mesh with clockwise side orientation should fix our issue, so that we're always reversing the mesh winding when needed.

Drigax:
After my first iteration, I realized that the world space shouldn't be relevant to our winding, also I noticed that we were also inverting our normals on import as well. I believe changing our defaultMaterial constructor in glTFLoader.ts@L1561 is the culprit, creating clockwise materials for LeftHand coordinate systems, and counterclockwise materials for RightHand systems, counter to Babylon's LeftHand, CounterClockwise default.

Making the following changes appear to fix our problem on both ends.

PR #6146, opened 2019-04-02
Proposed fix to issue #6103, changed exporter conditional for reversi…

Drigax:
…ng mesh winding to reverse when encountering a Clockwise oriented material in a positive worldspace


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1195,8 +1195,8 @@ export class _Exporter {
                     if (materialIndex != null && Object.keys(meshPrimitive.attributes).length > 0) {
                         let sideOrientation = babylonMaterial.sideOrientation;
 
-                        if (this._convertToRightHandedSystem && sideOrientation === Material.ClockWiseSideOrientation) {
-                            //Overwrite the indices to be counter-clockwise
+                        // Only reverse the winding if we have a clockwise winding in a positive worldspace
+                        if (sideOrientation === Material.ClockWiseSideOrientation && babylonTransformNode._getWorldMatrixDeterminant() > 0) {
```
deltakosh: you cannot call _ function as they are considered private

bghgary: This still doesn't look right to me. It needs to be an XOR operation. The determinant being negative flips the winding order.

PR #6156, opened 2019-04-03
Fix gizmo math to account for handedness

bghgary:


deltakosh:
Tests are failing. is that expected?

bghgary:
No, I will investigate.

Issue #6163, opened 2019-04-04
GUI3d: When using touch and when cliking with 2 fingers, the buttons will scale


deltakosh:
Repro here on any touch device:
https://www.babylonjs-playground.com/#HB4C01#9

mitsufu:
adding some info: button "up" (and then click) and the associated animation should be raised only when the last manipulator is up.

TrevorDev:
I see, Ill try to see if I can move the logic I added higher up to address the event issue. Thanks for clarifying

PR #6165, opened 2019-04-04
do not scale button multiple times with multi-touch

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/6163


deltakosh:
it was fast :)

PR #6171, opened 2019-04-05
Updated trailmesh

danjpar:

I have not been able to run gulp to build the files and test that way. It keeps failing from ammojs.
Should be okay though

### DIFF
--- a/src/Meshes/trailMesh.ts
+++ b/src/Meshes/trailMesh.ts
@@ -14,7 +14,7 @@ export class TrailMesh extends Mesh {
     private _generator: AbstractMesh;
     private _autoStart: boolean;
     private _running: boolean;
-    private _diameter: number;
+    private _dimensions: Array<number>;
     private _length: number;
     private _sectionPolygonPointsCount: number = 4;
     private _sectionVectors: Array<Vector3>;
@@ -26,17 +26,17 @@ export class TrailMesh extends Mesh {
      * @param name The value used by scene.getMeshByName() to do a lookup.
      * @param generator The mesh to generate a trail.
      * @param scene The scene to add this mesh to.
-     * @param diameter Diameter of trailing mesh. Default is 1.
+     * @param dimensions Dimensions of trailing mesh. Default is [1,1].
      * @param length Length of trailing mesh. Default is 60.
      * @param autoStart Automatically start trailing mesh. Default true.
      */
-    constructor(name: string, generator: AbstractMesh, scene: Scene, diameter: number = 1, length: number = 60, autoStart: boolean = true) {
+    constructor(name: string, generator: AbstractMesh, scene: Scene, dimensions: Array<number> = [1, 1], length: number = 60, autoStart: boolean = true) {
         super(name, scene);
 
         this._running = false;
         this._autoStart = autoStart;
         this._generator = generator;
-        this._diameter = diameter;
+        this._dimensions = dimensions;
         this._length = length;
         this._sectionVectors = [];
         this._sectionNormalVectors = [];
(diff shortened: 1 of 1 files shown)


src/Meshes/trailMesh.ts
```diff
@@ -26,17 +26,17 @@ export class TrailMesh extends Mesh {
      * @param name The value used by scene.getMeshByName() to do a lookup.
      * @param generator The mesh to generate a trail.
      * @param scene The scene to add this mesh to.
-     * @param diameter Diameter of trailing mesh. Default is 1.
+     * @param dimensions Dimensions of trailing mesh. Default is [1,1].
      * @param length Length of trailing mesh. Default is 60.
      * @param autoStart Automatically start trailing mesh. Default true.
      */
-    constructor(name: string, generator: AbstractMesh, scene: Scene, diameter: number = 1, length: number = 60, autoStart: boolean = true) {
+    constructor(name: string, generator: AbstractMesh, scene: Scene, dimensions: Array<number> = [1, 1], length: number = 60, autoStart: boolean = true) {
```
TrevorDev: Would it be more useful to use Vector2 instead here?

src/Meshes/trailMesh.ts
```diff
@@ -154,8 +153,8 @@ export class TrailMesh extends Mesh {
             let alpha: number = 2 * Math.PI / this._sectionPolygonPointsCount;
             for (let i: number = 0; i < this._sectionPolygonPointsCount; i++) {
                 this._sectionVectors[i].copyFromFloats(
-                    Math.cos(i * alpha) * this._diameter,
-                    Math.sin(i * alpha) * this._diameter,
+                    Math.cos(i * alpha) * this._dimensions[0],
+                    Math.sin(i * alpha) * this._dimensions[1],
                     0
                 );
                 this._sectionNormalVectors[i].copyFromFloats(
```
TrevorDev: When changing line 161 from 
```
Math.cos(i * alpha),
```
to 
```
0,
```

It avoids it poking through on this playground https://playground.babylonjs.com/#KXWZPA but you can see the black issue described in the other thread. Toggling backfaceculling didnt seem to work.

danjpar: The normals are tricky, even using computeNormals doesn't help. Setting the material's emissive color works well, but without it the normals don't look right. Flipping them doesn't help either.

changing line 147
```
positions[i - 3 * this._sectionPolygonPointsCount] = positions[i] - normals[i] / this._length * (this._dimensions[0] > this._dimensions[1] ? this._dimensions[0] : this._dimensions[1]);
```
to
```
positions[i - 3 * this._sectionPolygonPointsCount] = positions[i] - normals[i] / this._length / (this._dimensions[0] > this._dimensions[1] ? this._dimensions[0] : this._dimensions[1]);
```
the trail doesn't taper at the end, but it looks cleaner

src/Meshes/trailMesh.ts
danjpar: After some testing and adjustments, the first number of the array doesn't really change much anyway, so perhaps I should switch it back to diameter or call it width or height.

TrevorDev: Makes sense, width sounds good to me, and you would always set the first number to 0 and this parameter would control the second number?

src/Meshes/trailMesh.ts
TrevorDev: Interesting, I think a tapering end should still be supported. I'll keep playing with the code unless you have any other ideas.

src/Meshes/trailMesh.ts
danjpar: I have set it to be width. However, the trail is static to that plane of the mesh. Changing generator mesh scaling from y to x results in:
so I need to figure out a way to construct the trail independent of the orientation of the generator mesh. This example has _sectionPolygonPointsCount set to 3. Maybe have option to set initial orientation x,y,or z and then use 2 vertices with the highest and lowest values of that plane and the generator mesh center for the 3rd point
if that makes sense?

TrevorDev: Sounds reasonable to me.

src/Meshes/trailMesh.ts
deltakosh: Should I close that PR or do you think we will work on it more?

TrevorDev: I'm fine with this being closed I ended up using spheres for a demo instead of uneven objects.
ezgif-3-50bd3ebf7ed5
so unless @danjpar feels additional modifications are required im good.

danjpar: You can go ahead and close this. I will keep trying to improve it after I finish my project and get some free time. Might be a couple weeks.

PR #6176, opened 2019-04-08
Wrap gl data

deltakosh:


deltakosh:
Pinging @benaadams as he may use some of the changing APIs

PR #6187, opened 2019-04-11
fix: Matrix.ComposeToRef() didn't update the identity status

Kesshi:


deltakosh:
OMG! good catch mate!

Thanks a lot

PR #6188, opened 2019-04-11
Cloning a Mesh might throws an uncaught exception

barroij:
The exception is `TypeError: _this.getScene(...).getPhysicsEngine is not a function`

from line 
````javascript
var physicsEngine = scene.getPhysicsEngine();
````
I guess it happens when no physics is used at all and the scene has not been enhanced when physics related methods such has `isPhysicsEnabled` or `getPhysicsEngine`


barroij:
Hi @deltakosh,

A little fix for a crash happening in my app when cloning a mesh

deltakosh:
Good catch! Thanks a lot

PR #6197, opened 2019-04-12
apply camera render target stage prior to rendering depth

TrevorDev:
https://forum.babylonjs.com/t/facecolor-glowlayer-v4/2513/7

Test playground: 
https://playground.babylonjs.com/#9FFNT7
https://www.babylonjs-playground.com/#63QN3J#4

### DIFF
--- a/src/scene.ts
+++ b/src/scene.ts
@@ -3451,6 +3451,10 @@ export class Scene extends AbstractScene implements IAnimatable {
             this._intermediateRendering = true;
             let needRebind = false;
 
+            for (let step of this._cameraDrawRenderTargetStage) {
+                needRebind = needRebind || step.action(this.activeCamera);
+            }
+
             if (this._renderTargets.length > 0) {
                 Tools.StartPerformanceCounter("Render targets", this._renderTargets.length > 0);
                 for (var renderIndex = 0; renderIndex < this._renderTargets.length; renderIndex++) {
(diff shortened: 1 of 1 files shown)


src/scene.ts
```diff
@@ -3451,6 +3451,10 @@ export class Scene extends AbstractScene implements IAnimatable {
             this._intermediateRendering = true;
             let needRebind = false;
 
+            for (let step of this._cameraDrawRenderTargetStage) {
+                needRebind = needRebind || step.action(this.activeCamera);
```
TrevorDev: I needed to move this above the loop below because having a render target like depth in the below loop causes needRebind = true so effectlayer stages don't end up getting run. Is there a reason the _cameraDrawRenderTargetStage steps should not be run if there are render targets?

sebavan: no I guess the || is wrong no need to move it...

It should be a simple | here... my bad idea is it stays true or become true but the action always run...

GG Good catch

deltakosh: I'll fix it bu using | instead of ||

deltakosh: Here is the fix I ended up using:
```
for (let step of this._cameraDrawRenderTargetStage) {
                needRebind = step.action(this.activeCamera) || needRebind;
            }
```

(Without changing the position)

PR #6205, opened 2019-04-14
Fix Spector

sebavan:


sebavan:
Closed and reopen as I thought something bad might have slept in about fallbacks but all good :-)

Issue #6215, opened 2019-04-16
Zoom inertia makes it difficult to zoom out with ArcRotateCamera


MiiBond:
We’ve been noticing that the `ArcRotateCamera` has an issue where it’s easy to zoom in and hard to zoom out.

Here’s an example:
https://www.babylonjs-playground.com/ts.html#WFAW5A#1

Try zooming in quickly and then try to zoom out at the same speed. Notice that zooming out is much more difficult. The expected behaviour, in my mind, would be for zooming in and out to take the same amount of effort.

The issue appears to be that the inertia of the camera is unaffected by the radius (i.e. `wheelDeltaPercentage`) so your zoom-in speed can easily get you in close to the model but then `wheelDeltaPercentage` prevents you from speeding up again while zooming out.
This is most obvious when the `lowerRadiusLimit` is small.
The solution would, I guess, be to take `wheelDeltaPercentage` and `pinchDeltaPercentage` into account when calculating the distance traveled under inertia.

Note that we still want `wheelDeltaPercentage` to scale the speed that the camera can zoom but it should also scale the speed applied as inertia from the previous frame.

TrevorDev:
@MiiBond after looking at the code
https://github.com/BabylonJS/Babylon.js/blob/master/src/Cameras/Inputs/arcRotateCameraMouseWheelInput.ts#L49
 it looks like the scroll wheel events are always taking the current radius in mind when computing the inertia delta to travel, so if many events are built up while zooming it, it will zoom in too far so I tried computing the target radius and used that instead which hopefully will work in your scenario.

PR #6217, opened 2019-04-17
zoom with estimated radius to avoid zooming to far

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/6215

testPG: #DQ29UL


deltakosh:
Pinging @MiiBond for validation

MiiBond:
I just tested this out. It's much easier to use now!
However, now zooming in requires a lot more turns of the wheel than zooming out does. From our usability perspective, this is much better but it still feels like zooming in and out should ideally require the same amount of effort.

MiiBond:
@deltakosh @TrevorDev 
I'm definitely seeing issues with this change when zooming in very quickly. The camera bounces back instead of forward.

https://www.babylonjs-playground.com/ts.html#WFAW5A#2

The problem seems to be that `estimatedTargetRadius` is allowed to go negative. A simple clamp might fix it.

TrevorDev:
Sure, Ill try clamping it, see PR and let me know if it fixes it for you. (I cant seem to repro with my mouse)

MiiBond:
@TrevorDev that solves it for me. Thanks!

PR #6234, opened 2019-04-20
Picking support for UnIndexed meshes

aWeirdo:
Small change, but still unsure if this should be done as a separate function or if/how much it'll impact picking perf, it does run like a million times..

PGs; (see console)
sphere is indexed, half-box is unindexed.
Without change:  https://playground.babylonjs.com/#8QNBES#3
With change; https://playground.babylonjs.com/#8QNBES#4

### DIFF
--- a/src/Meshes/subMesh.ts
+++ b/src/Meshes/subMesh.ts
@@ -372,6 +372,11 @@ export class SubMesh extends BaseSubMesh implements ICullable {
             return this._intersectLines(ray, positions, indices, (this._mesh as any).intersectionThreshold, fastCheck);
         }
 
+        // Check if mesh is unindexed
+        if (!indices.length) {
+            return this._intersectUnIndexedTriangles(ray, positions, indices, fastCheck, trianglePredicate);
+        }
+
         return this._intersectTriangles(ray, positions, indices, fastCheck, trianglePredicate);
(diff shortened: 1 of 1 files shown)


deltakosh:
There is a linting issue

src/Meshes/subMesh.ts
```diff
@@ -404,12 +404,29 @@ export class SubMesh extends BaseSubMesh implements ICullable {
...
+
+        if(!indices.length) {
+            start = this.verticesStart;
+            count = this.verticesCount;
+        }
+
+        // Triangles test
+        var p0, p1, p2;
+        for (var index = start; index < start + count; index += 3) {
+            if(!indices.length) {
```
deltakosh: I would recommend a second function for perf reason. Picking can go through complicated meshes and avoid one if per face can be great

aWeirdo: ok, lint seemed like it was just unhappy with my missing whitespaces, hope i got them all, will check once it's done running, then update to a separate function

deltakosh: sounds good!

aWeirdo:
no need for what's new, 
but i'm unsure why Viewer Tests failed

PR #6271, opened 2019-04-29
enable undinexed LinesMesh

barroij:
Some code related to the `LInesMesh` was missing to support `LinesMesh` without IndexBuffer as in the triangle `Mesh`.

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -1289,7 +1289,7 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
                     break;
                 default:
                 case Material.TriangleFillMode:
-                    indexToBind = this._unIndexed ? null : this._geometry.getIndexBuffer();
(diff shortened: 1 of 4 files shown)


src/Meshes/mesh.ts
```diff
@@ -1289,7 +1289,7 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
                     break;
                 default:
                 case Material.TriangleFillMode:
-                    indexToBind = this._unIndexed ? null : this._geometry.getIndexBuffer();
```
deltakosh: well this seems suspicious..Furthermore we are in code lock so I can only merge fixes.
What about mesh that are unindexed. They wont have an index buffer so this code will crash

barroij: we know that `this._unIndexed` as we test that line 1280 (12 lines above)

deltakosh: oh yeah.so true :)

PR #6292, opened 2019-05-01
Fix Ammo memory leak

BitOfGold:
btVector3 created on every call of  getLinearVelocity, getAngularVelocity which result in memory leak.
(I'm using wasm Ammo version with BABYLON Ammo physics plugin)


BitOfGold:
Chrome has a bug too (does not free wasm memory), this happens to me less frequent after the fix.
https://bugs.chromium.org/p/chromium/issues/detail?id=935169

After the fix, ammo chrashes only after a few hours, without it only a few refreshes (I have hundreds of physicsbodies with a lot of get and set velocity)

PR #6293, opened 2019-05-01
Added CreateTilePlane and CreateTiledBox

BabylonJSGuide:
RE: https://forum.babylonjs.com/t/jeromes-to-do-list/2176 tiled box suggestion. My approach was based on his suggestion but in the end developed differently. When and if accepted I will produce the docs on it.

 In my version I first went for a CreateTiledPlane which you need to give tileWidth and TileHeight and the width and height of the plane. You can also then assign a variety of patterns to the tiles on the plane using constants
BABYLON.Mesh.NO_FLIP, 0
BABYLON.Mesh.FLIP_TILE, 1
BABYLON.Mesh.FLIP_ROW, 3
BABYLON.Mesh.ROTATE_TILE, 2
BABYLON.Mesh.ROTATE_ROW, 4
BABYLON.Mesh.FLIP_N_ROTATE_TILE, 5
BABYLON.Mesh.FLIP_N_ROTATE_ROW 6

Where whole tiles do not fit the given width and/or height you can decide whether the whole tiles are aligned to the left, right, top, bottom or center of the plane.

BABYLON.Mesh.CENTER, 0
BABYLON.Mesh.TOP, 3
BABYLON.Mesh.BOTTOM, 4
BABYLON.Mesh.LEFT, 1
BABYLON.Mesh.RIGHT, 2

A tiled plane can use frontUVs and backUVs.

To make a tiled box six tiled planes are used. Currently all sides use tiles of the same width and height and pattern. ( A bit of internal stuff forces whole or part tiles at the top of face 1 to be the same at the bottom of face 4. This is because I always set a camera to view faces 1 and 4) A tiled box can use faceUV but NOT frontUVs and backUVs.

The JS files are in GitHub - BabylonJSGuide/Tiled
The TS files are at Babylon.js/tiledPlaneBuilder.ts at master · BabylonJSGuide/Babylon.js · GitHub and Babylon.js/tiledBoxBuilder.ts at master · BabylonJSGuide/Babylon.js · GitHub

By setting the patterns and alignments into arrays I was able to view all 63 permutations by testing using the following (I was able to spot errors in the patterns and hence in my code. There were many during development)

The numbers above the meshes show
top row, just the count through the array options
bottom row, the constants for the pattern, vertical alignment, horizontal alignment

This https://musing-joliot-0f7a2f.netlify.com/ shows the back view of the planes so left and right are swapped. You need to comment out a line of code to get the front view.

All 63 perms for a box boxes

A box made with bricks brick box

By the way it is MeshBuilder only.


BabylonJSGuide:
Added a line in What's New so do not know what the error is. What do I need to do about this and other two errors?

deltakosh:
It is a bug in our CI (pinging @sebavan fyi)
We will fix it asap and merge it as soon as 4.1 will be open

deltakosh:
Really good THANKS!

PR #6301, opened 2019-05-01
UnIndexed picking fix

aWeirdo:
Support picking for instance of unIndexed mesh,


aWeirdo:
Need to check some things, something isn't right :)

aWeirdo:
Ok, found my mistake, instanceMesh already had _unIndexed property from AbstractMesh class, it simply wasn't updated

deltakosh:
some tests are failing?

aWeirdo:
seems unrelated, not sure why they failed

deltakosh:
@sebavan is on it:)

PR #6309, opened 2019-05-03
Bug fix

bghgary:
See https://forum.babylonjs.com/t/glbasync-destroying-background-layer/2949. This was a copy & paste bug.


deltakosh:
Seems like there may be an issue

PR #6322, opened 2019-05-07
404 doc link

Vinc3r:
I've modified only the `.ts` files, is that the right way to do? (I supposed `.js` & `.d.ts` are automatically updated when someone compiled the engine, that's it?)


deltakosh:
This is correct :D thanks!

PR #6330, opened 2019-05-09
Mesh.increaseVertices preserves updatable option

darkowic:
I am new to the library and your feedback is very appreciated :)

I see the `what's new.md` file was not edited for a long time. Not sure if I should update it.

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -2587,7 +2587,7 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
             }
 
             vertex_data.indices = indices;
-            vertex_data.applyToMesh(this);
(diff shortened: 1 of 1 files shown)


src/Meshes/mesh.ts
```diff
@@ -2587,7 +2587,7 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
             }
 
             vertex_data.indices = indices;
-            vertex_data.applyToMesh(this);
+            vertex_data.applyToMesh(this, this.isVertexBufferUpdatable(VertexBuffer.PositionKind));
```
darkowic: Is there any better option to get information about mesh `updatable` option?

deltakosh: Nope this is perfect!

PR #6331, opened 2019-05-09
also remove gui layers when assetcontainer is removed

TrevorDev:
https://forum.babylonjs.com/t/assetcontainer-dispose-does-not-work/2971/3


src/Layers/layerSceneComponent.ts
```diff
@@ -127,4 +128,37 @@ export class LayerSceneComponent implements ISceneComponent {
             return this._drawRenderTargetPredicate(layer, false, this.scene.activeCamera!.layerMask, renderTarget);
         });
     }
+
+    /**
+     * Adds all the element from the container to the scene
```
deltakosh:
```suggestion
     * Adds all the elements from the container to the scene
```

PR #6333, opened 2019-05-09
Add Babylon exporter AnimationGroup serializer.

Drigax:
Fix for .babylon scene serialization where animation groups aren't serialized. Added serializer for TargetAnimation and AnimationGroup.


src/Animations/animationGroup.ts
```diff
@@ -21,6 +21,19 @@ export class TargetedAnimation {
      * Target to animate
      */
     public target: any;
+
+    public constructor(animation: Animation, target: any){
```
deltakosh: that is a major breaking change :)

src/Misc/decorators.ts
```diff
@@ -247,6 +247,16 @@ export class SerializationHelper {
                 destination.animations.push(animation.serialize());
             }
         }
+
+        if (source.animationGroups)
```
deltakosh: animation groups are only hosted by the scene. They cannot be on an object

src/Animations/animationGroup.ts
Drigax: Can you help me find a better way? I only added the constructor due to how we initialize TargetAnimation via:
https://github.com/BabylonJS/Babylon.js/pull/6333/files#diff-10c0fe63a65cb9102072d2577dc27c04L180

which doesn't seem to work now that I added serialize().

Drigax: Ah, i think i just wanted to be fancy with field initialization, calling the default constructor should work, right?

src/Misc/decorators.ts
Drigax: I understand. I think i should just modify the sceneSerializer instead :)

deltakosh:
Thanks :) I let the CI finish and I will merge

PR #6336, opened 2019-05-10
BRDF fixes

bghgary:
Reduces BRDF base64 encoded string from 47557 to 31329 and avoids a flip.

Confirmed the resulting pixels from the base64 decode are identical via code.


bghgary:
Don't merge yet.

bghgary:
Ready to go once CI passes.

PR #6345, opened 2019-05-15
Unify preparation of instance attributes.

MarkusBillharz:
https://forum.babylonjs.com/t/custom-attributes-for-instanced-mesh-to-be-used-with-custom-shaders/3164/2

- Add Array definition support to MaterialHelper.PrepareAttributesForInstances in order to avoid redundancies when pushing the required attributes where the defines are not of type MaterialDefines.
- Remove redundancies.

### DIFF
--- a/src/Materials/materialHelper.ts
+++ b/src/Materials/materialHelper.ts
@@ -577,7 +577,8 @@ export class MaterialHelper {
      * @param defines The current Defines of the effect
      */
     public static PrepareAttributesForInstances(attribs: string[], defines: any): void {
-        if (defines["INSTANCES"]) {
+        // Check if we are handling a MaterialDefine or an Array containing the #define
+        if (defines["INSTANCES"] || (defines instanceof Array && defines.indexOf("#define INSTANCES") > -1)) {
             attribs.push("world0");
             attribs.push("world1");
             attribs.push("world2");
--- a/src/Materials/materialHelper.ts
+++ b/src/Materials/materialHelper.ts
@@ -578,7 +578,7 @@ export class MaterialHelper {
      */
     public static PrepareAttributesForInstances(attribs: string[], defines: any): void {
         // Check if we are handling a MaterialDefine or an Array containing the #define
-        if (defines["INSTANCES"] || (defines instanceof Array && defines.indexOf("#define INSTANCES") > -1)) {
+        if (defines["INSTANCES"] || Array.isArray(defines)) {
             attribs.push("world0");
             attribs.push("world1");
             attribs.push("world2");
--- a/src/Materials/materialHelper.ts
+++ b/src/Materials/materialHelper.ts
@@ -14,6 +14,7 @@ import { UniformBuffer } from "./uniformBuffer";
 import { Effect, EffectFallbacks, EffectCreationOptions } from "./effect";
 import { BaseTexture } from "../Materials/Textures/baseTexture";
 import { WebVRFreeCamera } from '../Cameras/VR/webVRCamera';
+import { MaterialDefines } from "./materialDefines";
 
 /**
  * "Static Class" containing the most commonly used helper while dealing with material for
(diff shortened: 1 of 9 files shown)


src/Materials/materialHelper.ts
```diff
@@ -577,7 +577,8 @@ export class MaterialHelper {
      * @param defines The current Defines of the effect
      */
     public static PrepareAttributesForInstances(attribs: string[], defines: any): void {
-        if (defines["INSTANCES"]) {
+        // Check if we are handling a MaterialDefine or an Array containing the #define
+        if (defines["INSTANCES"] || (defines instanceof Array && defines.indexOf("#define INSTANCES") > -1)) {
```
deltakosh: We have to be cautious here
this call is executed on every frame for every mesh so we cannot afford having a indexOf because it is a slow call

MarkusBillharz: Thanks for the review! 

Are you thinking about just removing it and relying solely on the type check in case it is not a MaterialDefines, or did you have something else in mind?

I could not find a way of performing the check much faster aside from a reverse looping since we can assume the define got added somewhere at the end, but even that did not seem too great performance wise against indexOf depending on the browser used.

deltakosh: I was thinking about removing it completely

MarkusBillharz:
Just a quick side note. Since we were talking about performance, I also had a look at instanceof:
https://jsbench.me/0vjvprxwnh/1

While there seems to be something weird going on with jsbench in this case (at least on my machines), I noticed that outside of jsbench Array.isArray seems to perform around 25% faster across my test devices. Since Array.isArray is already in use elsewhere in the repo and from what I've read seems to be more solid to use in this case overall, I figured I'll switch it out just to be safe.

Hope that's all right.

src/Materials/materialHelper.ts
```diff
@@ -577,7 +577,8 @@ export class MaterialHelper {
      * @param defines The current Defines of the effect
      */
     public static PrepareAttributesForInstances(attribs: string[], defines: any): void {
-        if (defines["INSTANCES"]) {
+        // Check if we are handling a MaterialDefine or an Array containing the #define
+        if (defines["INSTANCES"] || Array.isArray(defines)) {
```
deltakosh: Not sure to get why this test :)
what if I call PrepareAttributesForInstances with a defines not containing INSTANCES?

MarkusBillharz: Haha, maybe I need to catch up on some sleep, let me try to formulate what I think is going on and see if it makes sense. :D

We have two cases of defines that could arrive in this function as of now:
One is defines as MaterialDefines which gets handled by the defines["INSTANCES"] here. 
(Coming from most Materials like shadowOnlyMaterial)
Those classes all call PrepareAttributesForInstances without checking if it should be called, so this check we can't live without.

The other source of defines is coming in as an array filled with strings.
(Coming from the newly changed cases like EffectsLayer)
Those classes all check "if (useInstances)" before calling the helper function, since world0-3 were inline, to begin with.
Initially, I wanted to check if the correct array define is present using indexOf, but due to the performance overhead, we removed it.

However, if we only check defines["INSTANCES"], it will return undefined for any of those string define arrays, so we would never get to apply the attributes in that case. 
Without indexof or a faster alternative to look inside the array we now have to rely on MaterialDefines getting checked properly and giving standard arrays a free pass so to speak, but since those are pre-checked we are currently sure it needs to be applied.

Admittedly not too pretty, that's why I was a bit confused about removing it and relying only on the array type check in my last post :)

The only other option I see as of now to keep it all in one place would be an optional parameter like forceApply, so we could call. PrepareAttributesForInstances(attribs, null, true). Not sure if this is much better though. 

I guess if we want unified attribute management globally we would need to change those classes to use real MaterialDefines anyway so we might want to avoid touching the function parameters.

In case this is too hacky let's scrap the PR and move back to the drawing board to figure out a better way or a bigger milestone that encapsulates more than just this smallish change. I meant for it to be the first change in a series of things to adjust to make the bigger picture happen anyway, like we talked about on the forums.

I'd really love to stick with this topic and get some nice features in (like the instanciated mesh morphs) if possible, but it doesn't have to be today or tomorrow :) If we can start off with this that's cool, if not, no hard feelings, I learned a lot :)

Ugh, sorry for writing a book...

deltakosh: No worries ;) I now get why so this is fine!

In this case, we should just have two functions. One with the check that will call the second one where we apply the values.

The 2nd one will then be called by those who do not need a check and performance will be happy

MarkusBillharz: Feels way better 👍 Thx!

PR #6352, opened 2019-05-16
Added Planar dragging and individual gizmo control

balupg:
https://forum.babylonjs.com/t/gizmo-manager-to-support-planar-movements/3192

1. Introduced planar position control to support moving in a plane. (Default is disabled)
2. Control using gizmoManager.gizmos.positionGizmo.planarGizmoEnabled = true/false
3. Enable/Disable for individual gizmos
    eg. gizmoManager.gizmos.positionGizmo.xGizmo.isEnabled = false;

image
[gizmo demo.txt](https://github.com/BabylonJS/Babylon.js/files/3187457/gizmo.demo.txt)


deltakosh:
@trevordev: Can you also check if the meshes and materials created by all gizmos are correctly disposed. This is probably me but I have a doubt

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -34,16 +35,19 @@ export class PlaneRotationGizmo extends Gizmo {
...
+
     /**
      * Creates a PlaneRotationGizmo
      * @param gizmoLayer The utility layer the gizmo will be added to
      * @param planeNormal The normal of the plane which the gizmo will be able to rotate on
      * @param color The color of the gizmo
      * @param tessellation Amount of tessellation to be used when creating rotation circles
      */
-    constructor(planeNormal: Vector3, color: Color3 = Color3.Gray(), gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer, tessellation = 32) {
+    constructor(planeNormal: Vector3, color: Color3 = Color3.Gray(), gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer, tessellation = 32, parent: RotationGizmo) {
```
TrevorDev: does this need to be parent?: RotationGizmo as it is following paremeters that have defaults, this needs to be optional or have a default, everywhere the parent property has been added

src/Gizmos/positionGizmo.ts
```diff
@@ -25,34 +26,67 @@ export class PositionGizmo extends Gizmo {
...
     constructor(gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer) {
         super(gizmoLayer);
-        this.xGizmo = new AxisDragGizmo(new Vector3(1, 0, 0), Color3.Red().scale(0.5), gizmoLayer);
-        this.yGizmo = new AxisDragGizmo(new Vector3(0, 1, 0), Color3.Green().scale(0.5), gizmoLayer);
-        this.zGizmo = new AxisDragGizmo(new Vector3(0, 0, 1), Color3.Blue().scale(0.5), gizmoLayer);
+        this.xGizmo = new AxisDragGizmo(new Vector3(1, 0, 0), Color3.Red().scale(0.5), gizmoLayer, this);
+        this.yGizmo = new AxisDragGizmo(new Vector3(0, 1, 0), Color3.Green().scale(0.5), gizmoLayer, this);
+        this.zGizmo = new AxisDragGizmo(new Vector3(0, 0, 1), Color3.Blue().scale(0.5), gizmoLayer, this);
 
+        this.xPlaneGizmo = new PlaneDragGizmo(new Vector3(1, 0, 0), Color3.Red().scale(0.5), this.gizmoLayer, this);
```
TrevorDev: Are these enabled by default? I don't think they should to avoid disrupting existing apps if they update

src/Gizmos/planeDragGizmo.ts
```diff
@@ -0,0 +1,173 @@
...
+     * Creates a PlaneDragGizmo
+     * @param gizmoLayer The utility layer the gizmo will be added to
+     * @param dragPlaneNormal The axis normal to which the gizmo will be able to drag on
+     * @param color The color of the gizmo
+     */
+    constructor(dragPlaneNormal: Vector3, color: Color3 = Color3.Gray(), gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer, parent: PositionGizmo) {
+        super(gizmoLayer);
+        this._parent = parent;
+        // Create Material
+        var coloredMaterial = new StandardMaterial("", gizmoLayer.utilityLayerScene);
```
TrevorDev: Yep, any materials created should be disposed in the dispose method. I actually think this is an issue for all the gizmos currently, can you fix that here, I can fix the others.

TrevorDev:
> @TrevorDev: Can you also check if the meshes and materials created by all gizmos are correctly disposed. This is probably me but I have a doubt

Yea, I remember you asking this before and I thought they were being disposed because they didn't show up in the inspector but that's because they use the utility layer scene.

src/Gizmos/positionGizmo.ts
balupg: They are not enabled by default. I have verified the behaviour through a few PG examples.
I have added a property to enable/disable individual gizmos and they are default false for the planar ones

src/Gizmos/planeRotationGizmo.ts
balupg: The original approach in the coding seemed to be controlling the attachedMesh only to active gizmos. I have maintained the same approach here as well.
In this change, I have added isEnabled for individual gizmos.
When you disable a Gizmo, attachedMesh is null and if you enable it immediately without a mesh change, I needed a way to find the correct mesh.
Hence the parent.

src/Gizmos/planeDragGizmo.ts
balupg: I will add them. :-)

balupg: @TrevorDev 
Does it makes sense for me to add the dispose in all the others as well? For consistency sake, I am modifying the rest anyway.

balupg:
Here is a PG script for updated functionality
[gizmo demo.txt](https://github.com/BabylonJS/Babylon.js/files/3189769/gizmo.demo.txt)

deltakosh:
@balupg  thanks a lot! Can you update the doc as well?

balupg:
Thanks gents. I will update the doc soon.

PR #6361, opened 2019-05-17
Fix bug with morph target data in glTF loader

bghgary:
See https://forum.babylonjs.com/t/broken-morphtarget-in-vrm-glb-model/3284


il-m-yamagishi:
ref. https://github.com/virtual-cast/babylon-vrm-loader/issues/1

PR #6363, opened 2019-05-17
do not skip updating rtt view ubo if multiple scenes are active

TrevorDev:
testPG: #01KDKN


deltakosh:
One test is failing

deltakosh:
Up? :) @TrevorDev

TrevorDev:
It appeared to be a 503 on one of the servers hit for a texture unrelated to the PR. I reran it and it passed now.

sebavan:
Looks like we are rate limited on those API, we should try to move the storage to our cdn for the scene ?

Issue #6355, opened 2019-05-16
Color's getHashCode is incorrect


juchiast:
getHashCode is incorrect for Color3 and Color4.

- Bug repro on [playground](https://playground.babylonjs.com/#R86AWY)
- Current result: Hash code for green and blue is 0

takahirox:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Maths/math.ts#L67-L72

```javascript
public getHashCode(): number {
    let hash = this.r || 0;
    hash = (hash * 397) ^ (this.g || 0);
    hash = (hash * 397) ^ (this.b || 0);
    return hash;
}
```

If I'm right, floating point number is converted to integer before bit operation. Then `hash = (hash * 397) ^ (this.g || 0);` is equal to `hash = (hash * 397) ^ (Math.floor(this.g) || 0);`. So if `.g` or `.b` is smaller than 1, it'll be `hash = (hash * 397) ^ 0;`. Therefore, hash code will be always 0 if `.r` is 0 and `.g` `.b` are smaller than 1.

Probably this `getHashCode()` behavior is away from this function author expects.

aWeirdo:
(0 ^ any number) === 0;

deltakosh:
anyone interested to do the PR?

takahirox:
If you folks don't mind, I'd like to try.

deltakosh:
Please go ahead :D

aWeirdo:
Any updates here?

takahirox:
I'm gonna do in this weekend.

aWeirdo:
Ok, thinking about it, this might need multiplication by 255 to align with RGB values, to avoid any errors with small numbers < 0.1 🙂

takahirox:
Yup, I plan to make that number big enough.

PR #6365, opened 2019-05-19
Fix getHashCode() of Colors

takahirox:
This PR fixes `getHashCode()` of `Color3/4`. See #6355 for the detail.

A bit off topic. What exactly is `|| 0` in `this.xxx || 0` for? Can `.r/g/b/a` be `null`, `undefined`, or something else?

### DIFF
--- a/src/Maths/math.ts
+++ b/src/Maths/math.ts
@@ -65,9 +65,9 @@ export class Color3 {
      * @returns an unique number that can be used to hash Color3 objects
      */
     public getHashCode(): number {
-        let hash = this.r || 0;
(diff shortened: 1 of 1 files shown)


deltakosh:
|| 0 will force the value to be converted to integer

takahirox:
Converting from float to integer? Then I guess it should be not `|| 0` but `| 0` by any chance?

```
console.log(1.5 || 0);
-> 1.5;
console.log(1.5 | 0);
-> 1;
```

deltakosh:
Perfect!

PR #6370, opened 2019-05-20
Add key listeners when already focused

Poolminer:
https://forum.babylonjs.com/t/camera-keyboard-issue/3335/6


deltakosh:
Thanks!

PR #6394, opened 2019-05-24
Ammo.js library updates

MackeyK24:
Ammo.js with **Native Bullet Physics** contact callbacks (Global Collision Detection)


deltakosh:
Thanks!

Issue #32, opened 2013-08-27
Faces


balupg:
First of all - This is a great initiative. It has taken the webGL fear out of a newbie like me. :).

Is there any way for me to loop through the faces of a mesh? Such as while trying to color all sides of a cube differently.
Regards,
Balu

PR #6406, opened 2019-05-27
clear bogus "circular references" for Rollup

tom-berend:
There are no code changes in this pull request.  But I am a newbie, this is my first pull request, and someone should check my work.

Rollup fails on several "circular references".  The only serious one is that "engine.ts" includes "material.ts" when it really only needs the types.  I followed @sebavan's suggestion of replacing  'include' with 'declare type'.  In some cases the errant file was including itself, and the 'include' could be removed.

There was already a "constants.ts" file in Engines to avoid circular references, and the references to types in "material" only needed to be replaced by references to "constants".

This doesn't yet let me use Rollup if a physics engine is declared, but that's my next step.

---------
PS:  I successfully create the Babylon libraries, but Gulp fails unit tests on my machine because of duplicate definitions in Mocha.   I'm trying to figure that out.

### DIFF
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -12,7 +12,6 @@ import { IDisplayChangedEventArgs } from "../Engines/engine";
 import { VertexBuffer } from "../Meshes/buffer";
 import { UniformBuffer } from "../Materials/uniformBuffer";
 import { Effect, EffectCreationOptions, EffectFallbacks } from "../Materials/effect";
-import { Material } from "../Materials/material";
 import { IInternalTextureLoader } from "../Materials/Textures/internalTextureLoader";
 import { InternalTexture } from "../Materials/Textures/internalTexture";
 import { BaseTexture } from "../Materials/Textures/baseTexture";
@@ -33,6 +32,7 @@ import { IPipelineContext } from './IPipelineContext';
 import { DataBuffer } from '../Meshes/dataBuffer';
 import { WebGLDataBuffer } from '../Meshes/WebGL/webGLDataBuffer';
 
+declare type Material = import("../Materials/material").Material;
 declare type PostProcess = import("../PostProcesses/postProcess").PostProcess;
 declare type Texture = import("../Materials/Textures/texture").Texture;
 declare type VideoTexture = import("../Materials/Textures/videoTexture").VideoTexture;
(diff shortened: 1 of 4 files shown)


sebavan:
So it can not compile here:

image

for instance using Material.TriangleFillMode still in Engine.

I would not consider your actual setup for rollup to be safe knowing this looks like compiling on your end and the false detection of circular dependencies. You might consider pinging the rollup team for those actually.

We can still push the other changes you made once the build will be ok.

Thanks

tom-berend:
I feel completely lost.  I missed those errors because my VSCode shows over 300 errors in ENGINE.TS.  I still haven't been able to get NPM RUN BUILD to complete on my Linux laptop after two evenings of struggle, it dies building GLTF-LOADER trying to update a file it doesn't have permission to access (wouldn't be a problem in Windows).  A handful of tests fail because of some proxy error.   I will keep at it, but the learning curve on this project is almost vertical.

sebavan:
Our build automation is running exclusively on Linux without issues. I am really sorry you have that many issues where it should be all smooth. There is definitely something I do not get about it.

tom-berend:
Thanks @sebavan.  This was a nothing code change, but it has taken me two weeks.  Happily  I can now run "NPM RUN BUILD" to the end.  

Most of my problems are just part of the learning curve - like learning that Ubuntu's package manager delivers an older version of Node that doesn't work, NPM finds different versions of libraries in different places, CHMOD does not touch hidden files, chromeheadless no longer works from root privilege (even with --no-sandbox), etc.  I get through each step and feel I've learned something.

Along the way I upgraded my Ubuntu (didn't help), and reloaded Node, NPM, and BJS several times. I have a much better understanding of the build components.  It's all good.

------
But some things just seem wrong.  Here the build process is trying to write a temp file that has a symlink to the global NPM libraries (and failing of course).   This would work fine on Windows.  I got past this by providing "/usr/lib/node_modules" write privileges to everyone, but i'm certain that isn't the right way.

Screenshot from 2019-05-29 12-17-46
 
Here test #32 fails on some kind of proxy error in test-babylon-unit (it tried several dozen times).  I can't even figure out which test this is.  

Screenshot from 2019-05-29 12-42-14

--------
I'd like to keep working on BabylonJS.  

Rollup isn't a replacement for Gulp (which is just a task runner), it is just a bundler.  It is a replacement for Webpack that standardizes on the ES2015 module format and provides dead-code-elimination ('tree-shaking').  I think that's the future, and it is exactly where TypeScript has been leading us.

But to get there, we need to clean the 'prototype inheritance' out.   For example, in the Physics section there is a program 'physicsEngineComponents.ts'.  It is lovely, a piece of poetry, and wonderfully documents the properties and methods that must be added to classes like SCENE and ABSTRACTMESH to implement a physicsImposter.

That plugin-style code might have reduced the footprint once, it doesn't help in ES2015 and modern bundlers simply eliminate the code we don't use.  Everything is static in  ES2015  That is also the way TypeScript works (can't check types that are added dynamically), and it the way VSCode works (the tools are trying to statically analyze your code).  

It would be pretty simple to start deconstructing code like 'physicsEngineComponents.ts'.  The properties and methods would be moved back to their classes, and a documentation page added to the website.

I'd like to work on that.  It's donkey work, suitable for a newbie, and will help me understand the code.  But I need a mentor.

sebavan:
The access right is needed is we rely on npm link: https://docs.npmjs.com/cli/link to double check during our tests that the npm packages we create are all ok.

About the base/base/.... it usually happen if a link is not found during the test. Could potentially be a case sensitive error on linux. This also is related to the tests phase but I wonder why I can not repro it on Ubuntu :-(

That said, rollup and webpack are mostly similar today in the way they deal with bundle tree shaking and DCE. Most of the plugin for one exists in the other. My main concern with rollup was that some of the TS behavior were weird mostly around false detection of circular dependencies and augmentation. As you hit, you had to workaround some wrongly detected circular refs by changing the code which is not reinsuring. 

Also about augmentation in general: https://www.typescriptlang.org/docs/handbook/declaration-merging.html This is a fully supported TS feature allowing the split across modules. It actually only help in ES2015 to eliminate part of the code. I agree it is at the expense of relying on side effects but if you import at the most granular level you can pretty much control what you include in the bundle.

We were not relying on it before and recently moved to it to allow tree shaking. If you reintroduce them in their parent class,  we will not be able to tree shake anymore unfortunately.

The way the TS code is emitted does not allow method level tree shaking. This is all related to inheritance in general and is well documented in one of their open issue.

I am not sure moving back the component in the main class would be of any help here. I wish I am wrong and we could remove the side effect with this without losing tree shaking.

You can give it a try and let me know the result :-)

PR #6419, opened 2019-05-29
add internal texture to texture cache when creating a EnvironmentBRDF…

TrevorDev:
…Texture for the scene

Repro: #8MGKWK#0
Run in console:
```
var ext = BABYLON.Engine.LastCreatedEngine._gl.getExtension('WEBGL_lose_context')
ext.loseContext();
setTimeout(()=>{
    ext.restoreContext()
}, 3000)
```


src/Materials/Textures/texture.ts
```diff
@@ -701,7 +701,9 @@ export class Texture extends BaseTexture {
      */
     public static CreateFromBase64String(data: string, name: string, scene: Scene, noMipmap?: boolean, invertY?: boolean, samplingMode: number = Texture.TRILINEAR_SAMPLINGMODE,
         onLoad: Nullable<() => void> = null, onError: Nullable<() => void> = null, format: number = Constants.TEXTUREFORMAT_RGBA): Texture {
-        return new Texture("data:" + name, scene, noMipmap, invertY, samplingMode, onLoad, onError, data, false, format);
+        var texture = new Texture("data:" + name, scene, noMipmap, invertY, samplingMode, onLoad, onError, data, false, format);
+        scene.getEngine()._internalTexturesCache.push(texture.getInternalTexture()!);
```
deltakosh: Why the engine is not caching the texture? because of "base64" ? In this case I would expect engine.createTexture to do the caching

TrevorDev: Yep, this is weird after debugging a bit more I found that this is already in the list so this line adds it a second time. Something else is removing it from the list before contextRestore is called.

deltakosh: ok :)

src/Misc/brdfTextureTools.ts
```diff
@@ -62,7 +62,6 @@ export class BRDFTextureTools {
 
                 // Cleanup
                 engine.restoreDefaultFramebuffer();
-                engine._releaseTexture(texture);
```
TrevorDev: @sebavan do you know why the texture was being released here as it looks like its still used (Needed when contextRestore is called). Was another texture meant to be disposed of instead?

sebavan: I think it is swapped earlier isnt it ? it might also be a bug, you could try forcing a context lost to check ?

sebavan: The expanded texture should then be in use if expanded. not releasing the original might cause a leak.

TrevorDev: Oh I think I see, engine._releaseTexture(texture); causes the texture to be removed from the cache but then expandedTexture._swapAndDie(texture); is called but texture never gets added back to the texture cache to be restored on context loss. Ill give that a try, thanks

PR #6435, opened 2019-06-02
Add MeshMorph support to the DepthRenderer.

MarkusBillharz:
So we can have SSAO and other pretty things on morphed meshes.

Test:
https://www.babylonjs-playground.com/#4ZI604

Old:
image

New:
image


julien-moreau:
@deltakosh we should add it to the geometry buffer renderer as well :)

MarkusBillharz:
> @deltakosh we should add it to the geometry buffer renderer as well :)

@julien-moreau Already on it :) Should probably be done after this goes through.
I think the outline renderer is also missing it.

julien-moreau:
You rox thanks a lot :)

deltakosh:
Excellent! thanks

Issue #6441, opened 2019-06-03
Add a visibility predicate to normalizeToUnitVector


haroldma:
https://github.com/BabylonJS/Babylon.js/blob/62f0b403319ccb75d92c1762c59695a3ae787934/src/Meshes/abstractMesh.ts#L1000

It currently includes not enabled nodes. Tho it feels like getHierarchyBoundingVectors should ignore disabled meshes?

Also should normalizeToUnitVector live in Node instead of AbstractMesh?

deltakosh:
yes for both!

haroldma:
Thanks @TrevorDev !

PR #6442, opened 2019-06-04
Particle context loss fix & move normalizeToUnitCube to node

TrevorDev:
https://github.com/BabylonJS/Babylon.js/issues/6441


src/Meshes/abstractMesh.ts
```diff
@@ -981,45 +981,12 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
-        if (ignoreRotation) {
-            if (this.rotationQuaternion && storedRotationQuaternion) {
-                this.rotationQuaternion.copyFrom(storedRotationQuaternion);
-            } else if (this.rotation && storedRotation) {
-                this.rotation.copyFrom(storedRotation);
-            }
-        }
-
-        return this;
+    public normalizeToUnitCube(includeDescendants = true, ignoreRotation = false, predicate?: Nullable<(node: AbstractMesh) => boolean>): AbstractMesh {
```
deltakosh: Why do we need this? It will be inherited directly

deltakosh: oh gotcha..to change the return type

PR #6464, opened 2019-06-08
converted physics code injections to plain TS

tom-berend:
moved the JS-style code injections from 'physics/physicsEngineComponents.ts' back into 'AbstractMesh.ts' and 'Scene.ts' in standard TS-style.  

Rollup hiccups on JS-style code injections.  This PR allows me to use TS and Rollup for very fast development cycles, at least for physics engines.  There are other places in BJS that still use prototype injections, I will slowly fix them too if you allow this PR as a model.

-----

BTW, the Mesh.setPhysicsLinkWith() function is broken in Ammo, seems to work fine in Oimo. (https://www.babylonjs-playground.com/#0BS5U0#0 doesn't run the 'links' with Ammo).  I wanted to add an automated test for this function, but need to spend more time understanding it.   Will get there eventually.


sebavan:
We can not merge this change at the moment. As suspected:

Importing scene now imports physics again so: 516055 bytes vs 509803 before just by importing scene.

You now need to find a way to ensure we could use Babylon without importing physics if we do not want to use it. In the mean time you should check with rollup to fix their TS augmentation support if this is not supported.

tom-berend:
This totally confused me.  I spent a few days convinced that I was the stupidest newbie on the planet.  If Physics methods could be added through injection then I didn't understand anything about Typescript. 

So I went back to the basic how-to-use-ES6 demo:   https://doc.babylonjs.com/features/es6_support  I followed it step-by-step, and it was happy.  

Then I tried adding some physics.  And Webpack was still happy, but the browser was unable to find "enablePhysics()" and the program would not run.  This was exactly the problem that Rollup was hiccupping on.

The immediate difference between Rollup and Webpack was that Rollup identified the problem during compile, and Webpack happily built a faulty, unusable bundle.

Screenshot from 2019-06-10 20-07-52

Then I manually copied the files modified in this PR:  'Scene', 'AbstractMesh', and 'PhysicsComponents', from Babylon.js/.temp/distES6 into node_modules/@babylonjs/core of the demo directory.  And of course Webpack now successfully bundled the demo program and it ran nicely.

Obviously, no one has ever tried to use a physics engine with ES6 bundling.  Or any of the other functions that conserve a few bytes by using prototype injection (there are some 20 files that inject their components into Scene alone, none of them will work). 

Your concern about size is _exactly_ right.  That's the whole point of ES6.  But it's up to the tree-shaker to drop stuff off that isn't needed.  It's time to get rid of the old JS-style code and move on.

sebavan:
As explained in the documentation: https://doc.babylonjs.com/features/es6_support#faq

```
The intellisense does not propose the method I normally use in the bundled version and an undefined error is raised at runtime?

This will be the case for all the methods defined by module augmentation. This means that as long as you are not importing the parent modules, the methods will not even be discoverable. This is the case for all our scene components. For enabling physics on the scene you need import "@babylonjs\core\Physics\physicsEngineComponent" to populate the scene.enablePhysics function. Please find below the list of those components for their augmented methods:
```

Basically, you need to add in your code to ensure that the physics component is not tree shooked: 
```
import "@babylonjs\core\Physics\physicsEngineComponent";
```

Both Webpack and Rollup would normally emit the same error. It is coming from Typescript itself when the typings have not been imported they should not be usable.

> Obviously, no one has ever tried to use a physics engine with ES6 bundling. Or any of the other functions that conserve a few bytes by using prototype injection (there are some 20 files that inject their components into Scene alone, none of them will work).

Actually we did and it is the main reason we dedicated a portion of the doc to this issue. I guess it is not clear enough and I would be happy to rephrase/enhance according to your suggestion.

> Your concern about size is exactly right. That's the whole point of ES6. But it's up to the tree-shaker to drop stuff off that isn't needed. It's time to get rid of the old JS-style code and move on.

As explained before, the tree shaking would not happen anymore if we reintroduce the code without augmentation. That is the whole point behind this convoluted pattern, keeping back compat whilst allowing tree shaking. More explanation could be found here: https://medium.com/@babylonjs/our-journey-into-the-module-world-74a1964fea1a

Now, if there is a better approach addresing both back compat and a side effect free code enhancing the es6 discovery I am all in.

tom-berend:
Thank you for your patience with me.   

I built the demo program from the es6_support page for both Rollup and Webpack using the NPM approach, standard except with the code from this PR in  node_modules/@babylonjs/core.  I then added a few lines of code that referenced Physics, either included or commented out.   I left it to the bundler to decide whether physics modules were required or not.

For Webpack, the size of the bundle (which included Babylon but did not include Ammo) was:

-- Without Physics:   823.7 kb 
-- With Physics:        862.1 kb

The difference - almost 40kb - represents the unused physics code pruned from the bundle.  

If the future is ES6, I conclude there is no reason to use module augmentation, and many reasons not to.  Babylon remains fully backward compatible, with no need for an ES6 user to add extra modules like 'PhysicsConstructor'.  There is no need for you to write extra documentation.   And intellisense discovery works very nicely.

The existing 2.5 mb babylon.js would not be larger either.  It has to contain all the code.  The only penalty is to users on legacy CJS / UMD  bundlers.  As you point out in your first response to this thread, importing scene will drag in physics (and soon 20 other modules) when no tree-shaking is available.

A final thought:  Rather than break BJS into smaller modules in @babylonjs/core, it probably makes ES6 sense to provide larger and fewer bundles, and let the bundler worry about what to keep.  The most backward-compatible approach would be to mirror the current modules - babylonjs, babylon-materials, etc.

---------------------
 
To my surprise, Rollup's performance is only insignificantly better.  Rollup is only 6kb smaller in both cases.

sebavan:
> For Webpack, the size of the bundle (which included Babylon but did not include Ammo) was:
> -- Without Physics: 823.7 kb
> -- With Physics: 862.1 kb
> The difference - almost 40kb - represents the unused physics code pruned from the bundle.

About this, despite the size difference, it does not mean that the tree shaking is as efficient as it should. You should try to use webpack bundle analyser to visually appreciate the difference between a client using your changes and master on the default sample.

Taking the simplest possible code:
```
import { Engine } from "@babylonjs/core/Engines/engine";
import { Scene } from "@babylonjs/core/scene";

const canvas = document.getElementById("renderCanvas") as HTMLCanvasElement;
const engine = new Engine(canvas);

var scene = new Scene(engine);
```

Without your change:
image
for a total of 490kb

With your changes:
image
for a total of 497kb (you can see the physics being dragged in)

So it is a 7kb cause now by default mesh drags physics joints and some unecessary physx method like scene.enablePhysics which now by defaut brings physicsEngine and so on. We can not introduce back those dependencies with the current state of the tree shakers everybody are using.

You need to understand that we did the opposite change for this reason. So the solution would be to find a tool or technique allowing tree shaking on those unused dependencies. As long as they stay in the output, we will not be able to merge the PR.

tom-berend:
We are working through ['Five Whys'](https://en.wikipedia.org/wiki/5_Whys), and I am slowly learning.    

You are right.  I examined the "no-physics" bundle created by Webpack, and all the code for PhysicsJoint has been dragged in by AbstractMesh.  And PhysicsJoint subsequently drags in code from IPhysicsEngine. Blech.

Webpack (and Rollup) say "Our tool statically analyzes and optimizes your imports, and will exclude anything that isn't actually used.", but all they do is throw out functions you DON'T ask for.  They aren't checking whether the code is reachable, only that it has been referenced.  Yes, that could be smarter.

The obvious path is to refactor the BJS code, for example removing all references to 'PhysicsJoint' from AbstractMesh, keeping in mind your strict policy of backward compatibility.  In this particular case, only a single static value is extracted from PhysicsJoint, and it should be possible to raise that value to a data class shared by AbstractMesh and PhysicsJoint.  In general, there is no easy way to prune a tree of program-references.

OK, let me look at this code again, and try refactoring the low-hanging fruit.

tom-berend:
@sebavan   I think I am giving up on this. You should delete this PR.    

I have tried several approaches, but none make the code better.  Some were amusing - did you know that running BJS through Rollup TWICE saves another 2 kb?  But moving references to injected code from methods into functions explodes the code size.

The good news is that I tested a number of methods in the Physics interface as a byproduct.  Several are unnecessary or hopelessly broken, and removing them will reduce the bundle size.  I submitted an issue for one earlier today, will develop demonstrations of the others.  (for example, AbstractMesh.setPhysicsLinkWith() that I mentioned at the top of the thread "works" in such a limited way that it can be safely amputated without risk of breaking backward compatibility).  

And I learned a lot in the attempt.  

In the end, it is a matter of taste.  Saving a few kb seems (to me) to be a false economy against the risk of bugs that the compiler might help prevent.  Or adding a few days to the learning curve of a newbie.  Or even having to document the non-standard methods.  And I think eventually you will remove code injections as the tools improve.  

Thanks again for your patience with me.

sebavan:
No problem, I wish I will remove it :-) but the few kb for physics ends up quite large if we include everything so we have to stick with this until better tooling is available for the clients.

Glad you at least learned from it, I went there as well and it was awkwardly interesting ;-)

PR #6488, opened 2019-06-13
Fix issue when adding and removing observers

thscott:
When adding and removing observers with the same callback in quick succession, only the first observer will be removed from the observable.

See [this pg](https://playground.babylonjs.com/debug.html#VFX196) for a fairly contrived example. I noticed it though when adding a move observer on pointerdown and removing it on pointerup, where spam clicking would sometimes result in observers not being removed (due to observer removal being deferred).


deltakosh:
Looks good! thanks a lot

PR #6494, opened 2019-06-14
Navigation

CedricGuillemet:
Navigation mesh!
Please break it :)
playground : https://www.babylonjs-playground.com/#AJTRIL#1


dist/recast.js
```diff
@@ -0,0 +1,38 @@
+
```
bghgary: Only add files to the preview release folder. Once a new version is released, it will move into dist folder.

dist/preview release/recast.js
```diff
@@ -0,0 +1,38 @@
+
```
bghgary: Where is the wasm file? Is it embedded or something?

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,135 @@
...
+    update(deltaTime: number): void;
+
+    /// Asks a particular agent to go to a destination. That destination is constrained by the navigation mesh
+    agentGoto(index: number, destination: Vector3): void;
+
+    /// dispose
+    dispose() : void;
+}
+
+export class AgentParameters {
```
bghgary: if this is supposed to be a class, probably should initialize values and put access modifier (public)

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,135 @@
...
+    collisionQueryRange: number;
+
+    /// The path visibility optimization range. [Limit: > 0]
+    pathOptimizationRange: number;
+
+    /// How aggresive the agent manager should be at avoiding collisions with this agent. [Limit: >= 0]
+    separationWeight: number;
+}
+
+export class NavMeshParameters {
```
bghgary: ditto

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,267 @@
...
+declare var Recast: any;
+
+/**
+ * RecastJS navigation plugin
+ */
+export class RecastJSPlugin implements INavigationEnginePlugin {
+    /**
+     * Reference to the Recast library
+     */
+    public bjsRECAST: any = {};
```
bghgary: nit: bjs prefix is a bit weird here with an all caps `RECAST`.  How about just recast?

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,267 @@
...
+    public navMesh: any;
+
+    /**
+     * Initializes the recastJS plugin
+     */
+    public constructor(recastInjection: any = Recast) {
+        if (typeof recastInjection === "function") {
+            recastInjection(this.bjsRECAST);
+        } else {
+            this.bjsRECAST = recastInjection();
```
bghgary: Is this possible? Looks like the usage here is expecting recastInjection to be a function, so this else case can't happen?

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+
+    /**
+     * Creates a navigation mesh
+     * @param mesh of all the geometry used to compute the navigatio mesh
+     * @param parameters bunch of parameters used to filter geometry
+     */
+    createMavMesh(mesh: AbstractMesh, parameters: NavMeshParameters): void {
+        var rc = new this.bjsRECAST.rcConfig();
```
bghgary: Use `const` if possible or use `let` otherwise. Avoid using `var` as it has scoping issues.

src/scene.ts
```diff
@@ -1345,6 +1367,8 @@ export class Scene extends AbstractScene implements IAnimatable {
         if (!options || !options.virtual) {
             this._engine.onNewSceneAddedObservable.notifyObservers(this);
         }
+
+        this.crowds = new Array<ICrowd>();
```
bghgary: nit: You can just put this directly in the declaration if you'd like

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Configures the navigation mesh creation
+ */
+export class NavMeshParameters {
+    /**
+     * The xz-plane cell size to use for fields. [Limit: > 0] [Units: wu] 
+     */
+	cs: number;
+
+	/**
```
bghgary: weird tab here?

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,284 @@
...
+    /**
+     * Tick update done by the Scene. Agent position/velocity/acceleration is updated by this function
+     * @param deltaTime in seconds
+     */
+    update(deltaTime: number): void {
+        // update crowd
+        this.recastCrowd.update(deltaTime);
+
+        // update transforms
+        var index: number;
```
bghgary: nit: this can be inside the for loop using `let`

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,284 @@
...
+        return this.bjsRECAST !== undefined;
+    }
+}
+
+export class RecastJSCrowd implements ICrowd {
+    public bjsRECASTPlugin: RecastJSPlugin;
+    public recastCrowd: any = {};
+    public transforms: TransformNode[];
+    public agents: number[];
+    private scene: Scene;
```
bghgary: private variables should have an underscore prefix

dist/preview release/recast.js
CedricGuillemet: I compiled recast to js, not wasm.

src/scene.ts
```diff
@@ -164,6 +165,27 @@ export class Scene extends AbstractScene implements IAnimatable {
      */
     public ambientColor = new Color3(0, 0, 0);
 
+    public crowds: ICrowd[];
```
deltakosh: I would love to find a way to not touch the scene at all actually.

src/scene.ts
```diff
@@ -3716,6 +3740,11 @@ export class Scene extends AbstractScene implements IAnimatable {
             this.animate();
         }
 
+        // Navigation
```
deltakosh: Instead of having crowd here, maybe we could add an observable that a crowd manager can register to?

Playground/index-local.html
```diff
@@ -18,6 +18,7 @@
     <script src="js/libs/fileSaver.js"></script>
     <!-- Dependencies -->
     <script src="../dist/preview%20release/ammo.js"></script>
+    <script src="../dist/preview%20release/recast.js"></script>
```
deltakosh: we need to do it on index, full and frame as well :)

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * Navigation plugin interface to add navigation constrained by a navigation mesh
+ */
+export interface INavigationEnginePlugin {
+    name: string;
+
+    /**
+     * Creates a navigation mesh
+     * @param mesh of all the geometry used to compute the navigatio mesh
```
deltakosh: typo

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,235 @@
...
+     * @param parameters bunch of parameters used to filter geometry
+     */
+    createMavMesh(mesh: AbstractMesh, parameters: NavMeshParameters): void;
```
deltakosh: NavMeshParameters should be INavMeshParameters

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Get a navigation mesh constrained position, within a particular radius
+     * @param position world position
+     * @param maxRadius the maximum distance to the constrained world position
+     * @returns the closest point to position constrained by the navigation mesh
+     */
+    getRandomPointAround(position: Vector3, maxRadius: number): Vector3;
+
+    ///
```
deltakosh: missing doc

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * How aggresive the agent manager should be at avoiding collisions with this agent. [Limit: >= 0]
+     */
+    separationWeight: number;
+}
+
+/**
+ * Configures the navigation mesh creation
+ */
+export class NavMeshParameters {
```
deltakosh: should be an interface

src/Navigation/INavigationEngine.ts
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Release all resources
+     */
+    dispose() : void;
+}
+
+/**
+ * Configures an agent
+ */
+export class AgentParameters {
```
deltakosh: Should be an interface

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,284 @@
...
+     * Reference to the Recast library
+     */
+    public bjsRECAST: any = {};
+    public name: string = "RecastJSPlugin";
+    public navMesh: any;
+
+    /**
+     * Initializes the recastJS plugin
+     */
+    public constructor(recastInjection: any = Recast) {
```
deltakosh: pinging @sebavan to make sure this is ok with modules

dist/preview release/recast.js
bghgary: asm.js then? Is there a plan to use wasm?

CedricGuillemet: The recastjs build system supports outputting JS or WASM. I don't know which one is preferable.

deltakosh: wasm as it is faster

src/scene.ts
CedricGuillemet: It works with an observer.

src/Navigation/Plugins/recastJSPlugin.ts
CedricGuillemet: I did the same way as ammojs. I'm not familiar yet with that. So, any code snippet is welcome :)

dist/preview release/recast.js
CedricGuillemet: I get an issue with WASM output with an undefined symbol at runtime. It seems pretty hairy.  I prefer to keep this PR using JS and update with wasm in new PR.

src/Navigation/INavigationEngine.ts
CedricGuillemet: If it becomes an interface, will I be able to create it in the playground like
`var navmeshParameters = new BABYLON.NavMeshParameters();` ?

src/Navigation/INavigationEngine.ts
CedricGuillemet: Done with Interface

src/Navigation/Plugins/recastJSPlugin.ts
bghgary: I didn't look at ammojs, but this doesn't look right to me.

bghgary: Looked at ammojs...
        if (typeof ammoInjection === "function") {
            ammoInjection(this.bjsAMMO);
        } else {
            this.bjsAMMO = ammoInjection;
        }
```

Notice the lack of `()` on the ammoInjection
```

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -0,0 +1,319 @@
...
+     */
+    public isSupported(): boolean {
+        return this.bjsRECAST !== undefined;
+    }
+}
+
+/**
+ * Recast detour crowd implementation
+ */
+export class RecastJSCrowd implements ICrowd {
```
bghgary: should this be exported? seem like you only want this to be created by the plugin and thus it doesn't need to be exported.

CedricGuillemet: It's used by the user to add/remove agents, tick crowd, request a GoTo.

bghgary: Maybe I'm not understanding something. Looks like to me ICrowd does those things, but the actual class doesn't need to be exposed. Are you expecting someone to new a RecastJSCrowd?

CedricGuillemet: no, crowd creation only by calling createCrowd on INavigationEnginePlugin

deltakosh:
Up?

CedricGuillemet:
I'm testing the PR with a more 'real' scenario : https://forum.babylonjs.com/t/how-to-create-a-path-root-like-google-maps/4364

CedricGuillemet:
I now send an array of meshes and make a compound only using positions (more reliable than mesh merge). It's quiet fast but it lacks some UI  to properly set the nav mesh creation parameters. Doing that now...

CedricGuillemet:
image
I can change the navmesh parameters in realtime now but I'm not happy with the generation. I'll expose more parameters from recastjs (only 1 region, check on climbable angle,...)

deltakosh:
Some comments are not resolved apparently

src/Navigation/Plugins/recastJSPlugin.ts
CedricGuillemet: Seen with Seb and fixed.

src/Navigation/Plugins/recastJSPlugin.ts
CedricGuillemet: same caps as bjsAMMO

bghgary:
I have one comment that isn't resolved, but otherwise, looks fine to me.

src/Navigation/Plugins/recastJSPlugin.ts
bghgary: Can you try not exporting this class? I think it will still work.

tibotiber:
This looks **really** cool @CedricGuillemet. Just tried it, an it's super easy to setup and compute the navigation mesh. Is there any way this could be used to constraint the walkable area for a UniversalCamera?

CedricGuillemet:
Hi Thibaut

It should be possible to attach a camera to a transform agent. Something like:

camera.parent = agent.transform;

But you'll have to create an agent as it's the only object constrained by the navigation automatically. Or, every frame, you can do it manually with a getClosestPoint call and set the camera position relative to that point.

tibotiber:
Thanks Cédric, I’ll try both these ideas and post back for others who may be interested. Really appreciate the answer :).

tibotiber:
Hi! So I've tested what you suggested @CedricGuillemet and some things are worth reporting for whoever attempts something similar.

**Setting the camera's parent to the agent's transform node:**
This is pretty neat. If you implement the picking to move agents to a location demonstrated in https://www.babylonjs-playground.com/#X5XCVT#9, you get a pick-to-move-camera à la Google Street View. It doesn't mix well with keyboard inputs however since the `FreeCameraKeyboardMoveInput` will update the camera position, while the agent will update the parent's position, resulting in a hard to control global position which almost always gets out of the navmesh. As long as you stay away from the keyboard or remove that input, it's pretty interesting though. To be combined with a camera rotation input using keyboard or swipe (I'll test that next).

**using `getClosestPoint` at every frame:**
This clearly is the simplest way to make sure the camera remains on the navmesh, and it's compatible out of the box with the `FreeCameraKeyboardMoveInput`. There are some weird behaviours though. For example, when you walk under some stairs which are part of the navmesh, the camera can sometimes jump on top of the stairs. I have seen other jump behaviours near edges of the navmesh. I haven't got a full multi-storey map to test further unfortunately. From the first tests, it's as if `getClosestPoint` ignored the height (y value) when picking the closest point. Is that possible? 

**Quick question also:**
Why are we disconnecting the camera in https://www.babylonjs-playground.com/#X5XCVT#9 when picking an agents' location?

tibotiber:
btw, i'm new to this community so if there are better places or ways to discuss this sort of things, feel free to point me elsewhere ;)

Thanks again

deltakosh:
Actually we recommend using forum.babylonjs.com :)

tibotiber:
Done: https://forum.babylonjs.com/t/constraining-a-universalcamera-on-a-navigation-mesh/5042

Thanks for guidance @deltakosh, and sorry I'm now seeing the contribution guidelines 😬😰.

PR #6505, opened 2019-06-18
Store glTF nodes' extras in BJS metadata

pjoe:
Adds a glTF loader extension that will store any glTF nodes' extras properties in BJS TransformNode metadata.

Also includes a simple unittest.


bghgary:
@pjoe
Thanks for this PR! I have a general question first. This is currently specific to nodes. Does it make sense to generalize this so that other objects also loads metadata from extras?

pjoe:
@bghgary I suppose it could also be useful for other parts of the glTF (materials, lights, cameras, etc.). 

But at least for us the by far most common use case has been for nodes with custom properties, e.g. for: replacing proxy objects with something else at runtime, managing dynamic visibility of objects (think things like visible from frame 200-400), designate attachment points (where e.g. accessories are parented in to a animated skeleton).

So thought I would make this part work first. Should be easy enough to enhance with e.g. also materials.

Might also be worth comparing to how the three.js glTF loader deals with this.

loaders/src/glTF/2.0/Extensions/node_extras_as_metadata.ts
```diff
@@ -0,0 +1,57 @@
...
+import { INode } from "../glTFLoaderInterfaces";
+import { IGLTFLoaderExtension } from "../glTFLoaderExtension";
+import { GLTFLoader } from "../glTFLoader";
+
+const NAME = "node_extras_as_metadata";
+
+/**
+ * Store node extras (if present) in BJS TransformNode metadata
+ */
+export class CustomProps implements IGLTFLoaderExtension {
```
bghgary: The name of the class is usually the name of the file as well in Babylon code. Perhaps rename the file to `NodeExtrasAsMetadata.ts` and rename the class to `NodeExtrasAsMetadata`.

loaders/src/glTF/2.0/Extensions/node_extras_as_metadata.ts
```diff
@@ -0,0 +1,57 @@
...
+    /** @hidden */
+    public dispose(): void {
+        delete this._loader;
+    }
+
+    /** @hidden */
+    public loadNodeAsync(
+        context: string,
+        node: INode,
+        assign: (babylonMesh: TransformNode) => void
```
bghgary:
```suggestion
        assign: (babylonTransformNode: TransformNode) => void
```

loaders/src/glTF/2.0/Extensions/node_extras_as_metadata.ts
```diff
@@ -0,0 +1,57 @@
...
+    ): Nullable<Promise<TransformNode>> {
+        return this._loader.loadNodeAsync(
+            context,
+            node,
+            (babylonMesh): void => {
```
bghgary: nit: Rename `babylonMesh` to `babylonTransformNode`

loaders/src/glTF/2.0/Extensions/node_extras_as_metadata.ts
pjoe: Renamed to `ExtrasAsMetadata` as I also added handling of other glTF objects.

loaders/src/glTF/2.0/Extensions/node_extras_as_metadata.ts
pjoe: Done. Dunno what example I got that `babylonMesh` from :)

loaders/src/glTF/2.0/Extensions/node_extras_as_metadata.ts
pjoe: Done

pjoe:
Hmm looks like I managed to break GitHub, CI failing with: `GitHub reported the error, "Internal Server Error"`

Gonna do a small comment change in a bit to force another build

loaders/src/glTF/2.0/Extensions/ExtrasAsMetadata.ts
```diff
@@ -0,0 +1,104 @@
...
+        context: string,
+        material: IMaterial,
+        babylonDrawMode: number
+    ): Nullable<Material> {
+        const babylonMaterial = this._loader.createMaterial(
+            context,
+            material,
+            babylonDrawMode
+        );
+        if (babylonMaterial) {
```
bghgary: babylonMaterial can't be null/undefined, so this check is unnecessary

loaders/src/glTF/2.0/Extensions/ExtrasAsMetadata.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    private _assignExtras(
+        babylonObject: ObjectWithMetadata,
+        gltfProp: IProperty
+    ): void {
+        if (gltfProp.extras && Object.keys(gltfProp.extras).length > 0) {
+            if (!babylonObject.metadata) {
+                babylonObject.metadata = {};
+            }
+            Object.assign(babylonObject.metadata, gltfProp.extras);
```
bghgary: I forgot about this before, but the loader currently already writes a `metadata.gltf.pointers` to created Babylon objects.

https://github.com/BabylonJS/Babylon.js/blob/e22385628759e47b1eb89bf7fc66e4a48f95930a/loaders/src/glTF/2.0/glTFLoader.ts#L1983

It might be good to put the extras metadata at `metadata.gltf.extras` so that it's far less likely to conflict with other metadata. It will also make it clearer what extras to write when exporting a glTF.

loaders/src/glTF/2.0/Extensions/ExtrasAsMetadata.ts
pjoe: Fixed

loaders/src/glTF/2.0/Extensions/ExtrasAsMetadata.ts
pjoe: Done

bghgary:
@pjoe I'm looking at this code recently. Would it be okay to make this extension turned off by default? I can turn this on by default for the sandbox and playground, but I don't think it should be on by default for typical usage.

pjoe:
@bghgary Hmm for our use case it definitely makes sense to have it on by default. IIRC back when we used Three.js their loader also handled extras by default.

What are the main reason to turn it off? I wouldn't think performance overhead would be huge - though it does add a couple of call levels per node/camera/material, it only actually does anything if these have the extras property.

bghgary:
> What are the main reason to turn it off?

I don't have a strong reason. Extras are typically not useful except in specific scenarios. If you want then to be on, it's just a few lines of code. Not that big a deal either way.

pjoe:
Guess it depends on the use case 😄 

We are just so accustomed to using Blender custom properties (which exports as glTF extras) and have those available runtime by default.

For just loading and viewing a model, I don't think extras would be very useful, but as soon as you wanna do something more interactive, being able to have artists add extra info in something like Blender custom props and have those readily available runtime is very powerful 💪

bghgary:
Ok, sounds like we should leave it on. Thanks!

Issue #6496, opened 2019-06-17
gLTF loader not setting BoundingInfo


tibotiber:
Hi, I think I've found a bug with the gLTF loader. I'm using it through the `AssetsManager` and it doesn't seem to set the bounding info object. I thought this might be related to [this issue](http://www.html5gamedevs.com/topic/27575-gltf-loader-not-setting-boundinginfo/?tab=comments#comment-158656) (although I don't think I'm using gLTF 1 here), but setting `IncrementalLoading` to `false` doesn't change anything. I've created a minimal example PG.

- Bug repro on playground: https://www.babylonjs-playground.com/#3QG2DZ
- Expected result: the bounding box should appear and `getBoundingInfo().boundingBox` should be set
- Current result: the bounding box doesn't appear (presumably size is 0), and `getBoundingInfo().boundingBox` doesn't seem to be set

Just in case, I've tried with another model (both exported from Sketchfab) and it's the same.

By the way, just got started with babylonjs and I'm blown away by how well it's built, the ES6 support, tree shaking, the overall state of the docs and example. Kudos! 🚀

bghgary:
This is by design. `getBoundingInfo` returns a shallow bounding of the mesh (i.e. does not include children). Since the code is calling `getBoundingInfo` on the root node with no mesh associated, it just returns all zeros. To get the full bounding of all children, call `getHierarchyBoundingVectors`.

https://www.babylonjs-playground.com/#3QG2DZ#2

tibotiber:
Thanks a lot @bghgary. Is there any place I should have seen this? I've been reading few hours before posting but didn't see any mention of this. Keen to improve the docs if I can.

bghgary:
@deltakosh Are you aware of any documentation for this? Maybe updating the [getBoundingBox](https://doc.babylonjs.com/api/classes/babylon.abstractmesh#getboundinginfo) function comment would help?

deltakosh:
Yeah I think that's the best place

Thanks a lot @tibotiber

tibotiber:
PR opened. Thanks guys :)

PR #6511, opened 2019-06-19
Add note on shallow bounding of getBoundingInfo

tibotiber:
follow up to https://github.com/BabylonJS/Babylon.js/issues/6496#issuecomment-503245577


sebavan:
Thanks :-)

PR #6515, opened 2019-06-20
Fix #6473

### LINKED ISSUE #6473: Add meshes metadata on export as GLB
Here is the discussion:
https://forum.babylonjs.com/t/add-metadata-on-glb-export/1046/7.
The ideea is if we export a mesh which has metadata on it, on export as GLB try to keep these metadata somewhere so on import it again to get same infos.

deltakosh:


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1259,6 +1259,11 @@ export class _Exporter {
                     glTFNodeIndex = this._nodeMap[babylonNode.uniqueId];
                     if (glTFNodeIndex !== undefined) {
                         glTFNode = this._nodes[glTFNodeIndex];
+
+                        if (babylonNode.metadata) {
+                            glTFNode.extras = babylonNode.metadata;
```
bghgary: As I noted [here](https://github.com/BabylonJS/Babylon.js/pull/6505#discussion_r295907422), it might be a good idea to only export specific parts of the metadata. I think ideally we expose a callback so that users can decide what metadata to export and default to exporting `metadata.gltf.extras`.

deltakosh: how would you like to expose it?

deltakosh: fixed!

PR #6531, opened 2019-06-25
Fix #6529

### LINKED ISSUE #6529: Problems with glow layer once parsed from a serialized scene
# Bug

- Bug repro on [playground](https://www.babylonjs-playground.com/#558RAP#7):
- Expected result: the glow layer is parsed and no error happen while loading the scene
- Current result: the scene looks loaded correctly but an error happens in the console only if a glow layer is added to the source scene being serialized.

sebavan:
Egffect layer Deserialization


julien-moreau:
You rox!

PR #6534, opened 2019-06-25
Fix Xbox One Controller Buttons

MackeyK24:
Xbox One Controller no use the same button scheme as Xbox 360 Controllers. Test with https://html5gamepad.com/

Note: Still kept separate section for xbox controllers in the gamepad update just in case things change.


deltakosh:
There is a conflict with what's new

PR #6536, opened 2019-06-26
Send request url to custom webRequest modifiers

PierreLeBlond:
It would be helpful to change web request headers according to request url, e.g. for known protected urls.


deltakosh:
A test is failing

PR #6538, opened 2019-06-26
create effect renderer. 

TrevorDev:
Test PG:  #J3FUED#2


src/Materials/effectRenderer.ts
```diff
@@ -0,0 +1,126 @@
...
+                this.engine.unBindFramebuffer(renderTo.getInternalTexture()!);
+            }
+        });
+    }
+
+    /**
+     * Disposes of the effect renderer
+     */
+    dispose() {
+        if (this._tmpScreenBuffer) {
```
sebavan: should dispose RTT and buffers

src/Materials/effectRenderer.ts
```diff
@@ -0,0 +1,126 @@
...
+     * @param outputTexture texture to draw to, if null it will render to the screen
+     */
+    render(effectWrappers: Array<EffectWrapper> | EffectWrapper, outputTexture: Nullable<Texture> = null) {
+        if (!Array.isArray(effectWrappers)) {
+            effectWrappers = [effectWrappers];
+        }
+        effectWrappers.forEach((effectWrapper, i) => {
+            var renderTo = outputTexture;
+            if ((effectWrappers as Array<EffectWrapper>).length > 1 && i != (effectWrappers as Array<EffectWrapper>).length - 1) {
+                if (!this._tmpScreenBuffer) {
```
sebavan: By creating only one texture if you have 3 effects the second will fail as you will try to write where you read. You need a ringBuffer of two textures I suppose

src/Materials/effectRenderer.ts
```diff
@@ -0,0 +1,126 @@
...
+     * The underlying effect
+     */
+    public effect: Effect;
+
+    /**
+     * Creates an effect to be renderer
+     * @param creationOptions options to create the effect
+     */
+    constructor(creationOptions: {engine: Engine, fragmentShader: string, attributeNames: Array<string>, uniformNames: Array<string>, samplerNames: Array<string>}) {
+        var shaderName = "EffectWrapperEffect" + EffectWrapper._counter++;
```
sebavan: I wonder why not using the same definition than effect where we can either pass the source or the shader store name.

It feels weird to touch the shader store if source code is provided. It could actually even be an element:

https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/effect.ts#L334

src/Materials/effectRenderer.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (this._tmpScreenBuffer) {
+            this._tmpScreenBuffer.dispose();
+        }
+    }
+}
+
+/**
+ * Wraps an effect to be used for rendering
+ */
+export class EffectWrapper {
```
sebavan: Based on the comment right after counter is not needed and we also need a dispose to be able to clean the effect ?

src/Materials/effectRenderer.ts
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * The underlying effect
+     */
+    public effect: Effect;
+
+    /**
+     * Creates an effect to be renderer
+     * @param creationOptions options to create the effect
+     */
+    constructor(creationOptions: {engine: Engine, fragmentShader: string, attributeNames: Array<string>, uniformNames: Array<string>, samplerNames: Array<string>}) {
```
sebavan: Last feedback, could this be declared as a separate interface, so that it helps with intellisense comments ?

src/Materials/Textures/texture.ts
```diff
@@ -267,8 +268,8 @@ export class Texture extends BaseTexture {
...
      * @param format define the format of the texture we are trying to load (Engine.TEXTUREFORMAT_RGBA...)
      */
-    constructor(url: Nullable<string>, scene: Nullable<Scene>, noMipmap: boolean = false, invertY: boolean = true, samplingMode: number = Texture.TRILINEAR_SAMPLINGMODE, onLoad: Nullable<() => void> = null, onError: Nullable<(message?: string, exception?: any) => void> = null, buffer: Nullable<string | ArrayBuffer | HTMLImageElement | Blob> = null, deleteBuffer: boolean = false, format?: number) {
-        super(scene);
+    constructor(url: Nullable<string>, sceneOrEngine: Nullable<Scene | Engine>, noMipmap: boolean = false, invertY: boolean = true, samplingMode: number = Texture.TRILINEAR_SAMPLINGMODE, onLoad: Nullable<() => void> = null, onError: Nullable<(message?: string, exception?: any) => void> = null, buffer: Nullable<string | ArrayBuffer | HTMLImageElement | Blob> = null, deleteBuffer: boolean = false, format?: number) {
+        super((sceneOrEngine && (sceneOrEngine as Scene).getClassName && (sceneOrEngine as Scene).getClassName() === "Scene") ? (sceneOrEngine as Scene) : null);
```
deltakosh: We should add getClassName to engine and avoid that complicated line

src/Materials/effectRenderer.ts
```diff
@@ -0,0 +1,192 @@
...
+    private _ringBufferIndex = 0;
+    private _ringScreenBuffer: Nullable<Array<Texture>> = null;
+
+    private _getNextFrameBuffer(incrementIndex = true) {
+        if (!this._ringScreenBuffer) {
+            this._ringScreenBuffer = [];
+            for (var i = 0; i < 2; i++) {
+                var internalTexture = this.engine.createRenderTargetTexture(
+                    {
+                        width: Math.floor(this.engine.getRenderWidth(true)),
```
deltakosh: why floor? Th sampling mode is nearest so any size should work

TrevorDev: I just copied this from a playground I found but it does work without it so I removed the floor

PR #6540, opened 2019-06-27
 Playground design v4

Jaskar:
Brand New design !
Media queries for big screens, middle screens, little screens and Mobile.


deltakosh:
Some feedback:
- If no change to the code we should not see the alert dialog when switching languages
- mobile version does not work for me (the code panel should not be visible by default and should popup when clicking the button):
image

deltakosh:
Any update? @Jaskar

Jaskar:
Sorry for the delay !
So, until now :
- Reviewed the mobile CSS !
- Modified the index.html but I have to do the others .html (will be quick on this one)
- Rewrote a large portion with JS Classes
- Still have to split the big "Run" function in the index.js

And, I added some "// TO DO" for things I didn't get, or have to dig a little more to comment.

deltakosh:
Do you mind sharing the current PG somewhere so patrick can review it? (Maybe using netlify from your repo?)

deltakosh:
I did it actually :)

but it does not work :(
https://peaceful-jones-9cb9ee.netlify.com/

the testPG branch will be automatically deployed so we can test

deltakosh:
oh wait this is my fault..redeploying

deltakosh:
Ok it works now

@Jaskar:
- Mobile version does not apply when I reduce the size of my desktop window
- We should not see anything (not even the splitbar when code panel is hidden

@PatrickRyanMS do you mind having a look to add more feedback: https://peaceful-jones-9cb9ee.netlify.com/

Jaskar:
- Mobile
  * Mobile version does not apply when I reduce the size of my desktop window : Actually there's three different sizes for desktop, and one for mobile. It uses media queries, so should'nt have any problems ?
  * We should not see anything (not even the splitbar when code panel is hidden : I'll fix that
  * Hambuger menu disapears sometimes : Menu ? I spotted the logo disappearing, I'll fix that too
  * No color for JS and TS : Currently, only the button / menus are changing. Do you want the full screen to change ?
  * Sub menus do not appear at the right place (in front of the main menu) : Repro step ? This works for me.
  * No highlight state when clicking on a menu : Will fix
  * fps is overlapping the header :  Repro step ?

- Desktop
  * Clicking on TS / JS does not change the color of the active one : Weird...
  * No version selector : My bad, forgot to rewrite it
  * Font sub menu on the right (not on the left) and wrong background color : If it's on the right, it can overflow the screen. Concerning the background color, I copy/past the hexa colors. White bqckground, right ?
  * Babylonjs logo should take us to bjs homepage when clicked : Will fix !
 
- Overall
  * No drop shadows on FPS and on main bar (no drop shadow at all actually) : Really ?
  * Type face is incorrect (should be semi-condensed acumin. They are defined in the redlines) : Will fix
  * Should switch to mobile version at 1024px width : Ok ! Will change that

deltakosh:
- Mobile version must apply at 1024px even on desktop
- No color for TS and TS menu icon in the hamburger menu
- Sub menu should be in front of the opening menu and not at the top (like classic sub menus):
image
- fps overlap. Just go mobile:
image

- Font sub menu should be on the right:
image. You are right for white background

- Drop shadow: Yep no drop shadows please

PR #6547, opened 2019-06-28
Sync web xr end for latest spec as of June

TrevorDev:

TestPG: #AYNQTH


src/scene.ts
```diff
@@ -3523,10 +3523,10 @@ export class Scene extends AbstractScene implements IAnimatable {
 
             this._intermediateRendering = false;
 
-            // Restore framebuffer after rendering to targets
-            if (needRebind) {
-                this._bindFrameBuffer();
-            }
+            // Bind the frame (this is needed to be done in case a subcamera has an outputRenderTarget)
+            // TODO this undoes the change that addresses this https://github.com/BabylonJS/Babylon.js/issues/6124
```
TrevorDev: @deltakosh is it fine if I add this back?

sebavan: could you not instead mark needRebind to true if the subCamera has an outputRenderTarget ? this would prevent trying to change the current frambuffer if not needed and enter in all the cache and validation checks ?

Issue #6521, opened 2019-06-21
glTF Serializer: Some assets export with additional textures


bghgary:
The asset which repros this cannot be shared publicly. If we can figure out what the issue is, maybe we can find/create an asset that can be shared publicly.

The round trip of this asset (load into a sandbox and then exported) causes extra textures to be created and the textures are bigger than the original.

Drigax:
This sounds similar to #6503. Hopefully they are related, but I'll take some time to look at these after I tackle https://github.com/BabylonJS/Exporters/issues/549

Drigax:
Sounds like a mix of a bug, and somewhat expected behavior from my limited investigation.

![image](https://user-images.githubusercontent.com/3952729/60470852-e2374f80-9c16-11e9-9161-75d8d3a69471.png)

I was able to reproduce with an asset (Dude from the meshes library) where the asset has four textures, but exports with five due to materials 35427485 and 51054032 both using texture 3.jpg, which gets duplicated. Definitely a bug.

The increase in texture size was due to the asset being imported with a compressed format (JPG), but the engine does not re-compress as we export the texture from memory, which I'd argue is expected behavior.

sebavan:
Agree for both points, we should only try to prevent the duplicate :-)

Drigax:
https://github.com/BabylonJS/Babylon.js/pull/6567

After more investigation, although we allocate multiple BabylonTextures for texture 3.jpg, they both refer to the same InternalTexture. On export, we modify the name of the texture when clashing before checking to see if we can reuse the GLTFImage, causing the dupe.

PR #6550, opened 2019-06-29
Fix texture naming on GLTF load

Drigax:
The GLTF loader renames loaded textures to "materialName ({TextureType})" on load, breaking our .babylon serializer, as the original name is lost.

This PR proposes that we change the GLTF loader to name textures with the name provided by the GLTF texture, and remove the default naming schema.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -86,6 +86,7 @@
 - Fix bug when adding and removing observers in quick succession ([sable](https://github.com/thscott))
 - Cannon and Ammo forceUpdate will no longer cause an unexpected exception ([TrevorDev](https://github.com/TrevorDev))
 - Loading the same multi-material twice and disposing one should not impact the other ([TrevorDev](https://github.com/TrevorDev))
+- GLTF loader should now preserve the texture naming ([Drigax](https://github.com/Drigax))
 
 ## Breaking changes
 - Setting mesh.scaling to a new vector will no longer automatically call forceUpdate (this should be done manually when needed) ([TrevorDev](https://github.com/TrevorDev))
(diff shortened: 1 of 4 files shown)


bghgary:
This was done on purpose as the texture names in the file are not guaranteed to be unique. Maybe we should have a flag that does one or the other.

bghgary:
Furthermore, there is not a one-to-one mapping between glTF textures and Babylon textures. Babylon textures are more like a glTF textureInfo object which does not have a name.

sebavan:
@Drigax, I think we should revert and find another way, cause this will create texture clash in the sandbox for instance, as Gary said this was the main reason of introducing the naming. Can you take care of it ?

Drigax:
That's a pretty good reason to change the names, we can add a flag to the loader, but I'm not sure how useful that would be. 

We can always try to rename at export using the babylonTexture url as the name instead.

Drigax:
Giving this more thought @sebavan why is having multiple textures in the sandbox that share names even an issue? The textures are given unique Identifiers. I noticed that #6521 causes this to happen without issue, as far as I can tell.

sebavan:
Babylon is having a cache system for textures and basically the new one will be taken from the cache instead of the given uri

sebavan:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/Textures/baseTexture.ts#L530

Drigax:
Right, but we use the image uri to check the cache, not the texture name. I'm currently unsure how we can use multiple images that share name (say we have two models that both have their own unique "texture.png" in their own local directory) but, so far my limited testing isn't showing much issue

sebavan:
So currently https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/Textures/texture.ts#L274 name and url are often the same. Maybe we can add a originURL smwhere to keep track of it.

But why do we need it in the first place ?

The main issue is to prevent broken import, why not reading the current texture pixels values with to recreate a new texture if the path is not valid ?

deltakosh:
Well url should be the originUrl :)

PR #6559, opened 2019-07-01
Added a function for calculating the intercept

horusscope:
e.g. for a top-down view RTS game
It seemed very improper to even consider collisions in this context.

### DIFF
--- a/src/Culling/ray.ts
+++ b/src/Culling/ray.ts
@@ -248,7 +248,7 @@ export class Ray {
     /**
      * Calculate the intercept of a ray on a given axis
      * @param axis to check 'x' | 'y' | 'z'
-     * @param intercept offset (i.e. an offset of 1y is intercepted above ground)
+     * @param offset from axis interception (i.e. an offset of 1y is intercepted above ground)
      * @returns a vector containing the coordinates where 'axis' is equal to zero, or false if there is no intercept.
      */
     public intersectsAxis(axis: string, offset: number = 0) {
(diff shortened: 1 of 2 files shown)


src/Culling/ray.ts
```diff
@@ -245,6 +245,24 @@ export class Ray {
             return distance;
         }
     }
+    /**
+     * Calculate the intercept of a ray on a given axis
+     * @param axis to check 'x' | 'y' | 'z'
+     * @returns a vector containing the coordinates where 'axis' is equal to zero, or false if there is no intercept.
+     */
+    public intersectsAxis(axis) {
```
deltakosh: axis should be typed as string

src/Culling/ray.ts
```diff
@@ -245,6 +245,24 @@ export class Ray {
...
     }
+    /**
+     * Calculate the intercept of a ray on a given axis
+     * @param axis to check 'x' | 'y' | 'z'
+     * @returns a vector containing the coordinates where 'axis' is equal to zero, or false if there is no intercept.
+     */
+    public intersectsAxis(axis) {
+          var t = this.origin[axis] / this.direction[axis]
+          if(t > 0) return false
+          const stepOrigin = function(t, axis) {
```
deltakosh: for performance and GC reason we should not instantiate a function every time we call intersectsAxis

src/Culling/ray.ts
horusscope: sorry about that, I only liked the readability.

deltakosh:
There is a build issue :)

PR #6564, opened 2019-07-02
Revert #6550

Drigax:


Drigax:
Closing this revert, I'm not sure if it's necessary. see #6550

bghgary:
@Drigax As per our discussion, we should still revert the change to the loader.

Drigax:
Agreed. I have some other work on the exporter side im prioritizing right
now due to low regression risk.

PR #6590, opened 2019-07-15
add disable prompt option for OculusQuest

shinyoshiaki:
In OculusQuest Using VirtualKeyboard then , babylon.js show prompt because OculusQuest is android mobile device.
When prompt shown , browser stop webvr until resolved prompt. So can not input virtualKeyboard via vr controller until resolved prompt.
So we should to add disable prompt option for OculusQuest(and other android base vr devices).


deltakosh:
Excellent!

PR #6602, opened 2019-07-17
Fix VR button placement

### LINKED ISSUE #6598: Improvements to VR Button
[The move to bottom right](https://github.com/BabylonJS/Babylon.js/blob/12a5f9b7ce49c8111cd14fea6254f19dffca5bed/src/Cameras/VR/vrExperienceHelper.ts#L925-L926) code doesn't always works. Both `canvas.top` and `offsetTop` don't always return the absolute position. In our canvas they both return zero. Instead you should rely on `getBoundingClientRect()`.

Another improvement is exposing the button itself. This will allow us to hide and show the button at will.

haroldma:
Fixes #6598

### DIFF
--- a/src/Cameras/VR/vrExperienceHelper.ts
+++ b/src/Cameras/VR/vrExperienceHelper.ts
@@ -298,7 +298,7 @@ export class OnAfterEnteringVRObservableEvent {
 export class VRExperienceHelper {
     private _scene: Scene;
     private _position: Vector3;
-    private _btnVR: HTMLButtonElement;
+    private _btnVR: Nullable<HTMLButtonElement>;
     private _btnVRDisplayed: boolean;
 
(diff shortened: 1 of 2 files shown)


src/Cameras/VR/vrExperienceHelper.ts
```diff
@@ -616,6 +616,13 @@ export class VRExperienceHelper {
         return this._vrDeviceOrientationCamera;
     }
 
+    /**
+     * The html button that is used to trigger entering into VR.
+     */
+    public get vrButton(): HTMLButtonElement | undefined {
```
deltakosh: Can you use Nullable<HTMLButtonElement > (and flat the internal button the same way?

deltakosh:
It broke the build. I';ll take care of it

PR #6615, opened 2019-07-23
fix(Scene): Dispose of timer that may not have been cleared if the scene was never ready.

zakhenry:


zakhenry:
oops this fix does not actually fix the issue as it is in a tight loop. Will revisit

PR #6620, opened 2019-07-23
 Add absolute scale and rotation to transform node

haroldma:
Our team ends up using decompose in some cases when working with scale and rotation. Since there's an absolutePosition, it made sense to do this change.

Thoughts?

### DIFF
--- a/src/Meshes/transformNode.ts
+++ b/src/Meshes/transformNode.ts
@@ -353,7 +353,6 @@ export class TransformNode extends Node {
      * Returns a Vector3.
      */
     public get absolutePosition(): Vector3 {
-        this._syncAbsoluteScalingAndRotation();
         return this._absolutePosition;
     }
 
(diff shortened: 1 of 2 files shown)


src/Meshes/transformNode.ts
```diff
@@ -1078,8 +1096,8 @@ export class TransformNode extends Node {
 
         this._afterComputeWorldMatrix();
 
-        // Absolute position
-        this._absolutePosition.copyFromFloats(this._worldMatrix.m[12], this._worldMatrix.m[13], this._worldMatrix.m[14]);
+        // Update absolute transform values
+        this._worldMatrix.decompose(this._absoluteScaling, this._absoluteRotationQuaternion, this._absolutePosition);
```
deltakosh: That is an expensive function we cannot call in the compute world matrix function (which can be called thousand of times per frame.

I would recommend having a flag in the computeWM function that indicates that absolute scaling and rotation are out of sync, then in the getters check that flag and update the _absoluteScaling and _absoluteortationquaternion before returning them

haroldma: Updated!

deltakosh:
Fantastic! I let the CI finish and I merge

PR #6625, opened 2019-07-25
Playground v4.4

Jaskar:
QRcode is back.
-> In mobile mode too ?

Version selector is here, with a config file.
-> Need help to "Unload" old BJS version. Is set BABYLON to null, but this isn't good enough.
-> Do I have to add text field so the users can enter custom version ?

Still have to find the burger button bug ....

### DIFF
--- a/Playground/css/index.css
+++ b/Playground/css/index.css
@@ -414,6 +414,13 @@ body {
 .navbar .select {
     position: relative;
 }
+.navbar .select span {
+    width: 100%;
+    height: 55px;
+    line-height: 55px;
+    font-family: "acumin-pro-extra-condensed";
+    font-size: 20px;
+}
 
 .navbar .select .subSelect {
     position: relative; 
@@ -428,6 +435,10 @@ body {
     display: none;
     text-transform: uppercase;
     background-color: #201936;
+}
+.navbar .select .toDisplay.currentVersionDisplay {
+    width: 100%;
+    display: none;
 }
 .navbar .select .toDisplay .option {
     font-size: 20px;
@@ -453,9 +464,6 @@ body {
 }
 .navbar .select .toDisplay.languageTS .option:hover {
     background-color: #ff7656;
-}
-.navbar .select .toDisplay.currentVersionDisplay {
-    width: auto;
 }
 
 .navbar .select .subSelect .toDisplaySub {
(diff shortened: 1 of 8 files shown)


PatrickRyanMS:
In looking at the preview there are a few issues that need some attention.

- [ ] Whichever tab is active (Javascript vs Typescript) should no longer be a clickable button. Clicking on the active tab will reload the page.
- [ ] The default color for the play button should be white and when hovered, the background should be white and the foreground should be the color of the top bar (based on which tab is active)
- [ ] There should be no period after "v" in the version button to the right. It should be written as "v4.0" to match the version title on the left of the bar.
- [ ] The hover state for the "v4.0" button on the right should have a white background as it currently does, but the color of the type should be the color of the top bar (based on which tab is active)
- [ ] The examples button on the right should stay as a bar color icon on a white background while the menu is open to show that the panel is active.
- [ ] In the submenu panel (Settings > Theme > Light/Dark) the active setting should be purple type on a white background and the unselected options should be white type on the darker purple background.
- [ ] If you have a submenu open (Settings > Theme > Light/Dark) and move your mouse out of the menu, the highlight behind the setting icon and the Theme button both disappear. We should probably close all menus when we lose focus on all menus rather than needing to click outside the menus to close them. 
- [ ] **Mobile:** In mobile mode, when you toggle to code view, drop the FPS counter as we only need to see it when looking at the canvas.
- [ ] **Mobile:** I checked the mobile mode on my iPhone X in safari, when selecting a menu with a submenu like Theme, the submenu does not appear. It does on desktop when scaled to mobile size.
- [ ] **Mobile:** When clicking buttons with a single action like "Inspector" or "Clear" the menu does not close after the action, but rather you need to click on the menu button or off the menu to close it.
- [ ] **Mobile:** The render order of the FPS counter should be lower than the inspector on mobile so that the FPS counter does not cover the inspector when it is open.
- [ ] **Mobile:** Add a close button (X) to the meta data panel. On a mobile device the panel takes over the whole screen and the cancel link type is small and can be overlooked. Adding a close button in the upper right corner will hopefully help people find how to get out of the panel.
- [ ] **Mobile:** If the menu is open and the user scrolls the page, we should close the menu. Right now the menu tears itself away from the hamburger button and floats where it was called. 
- [ ] **Mobile:** If the menu is open and I tap outside the menu, it stays open rather than closing from losing focus.
- [ ] **Mobile:** When the keyboard is shown , the button to switch between code and canvas is behind the keyboard (iOS) but works on Android (the button gets moved up)
- [ ] **Mobile:** We should hide the top bar with branding in landscape mode in the editor when the keyboard is invoked. We should still leave the hamburger menu so the user has access to the menu, but with the rest of the bar visible, you only see maybe one line in the editor in landscape.

PR #6630, opened 2019-07-26
Defining typescript interface for options of method to make screenshots

Dok11:
From https://forum.babylonjs.com/t/define-typescript-interface-for-options-of-method-to-make-screenshots/4713
All test are passed and build successfully complete


sebavan:
Could the interface be moved within the screenshotTools file ? This helps preventing import from tools in ES6 when not necessary.

Thx

Dok11:
@sebavan check it please https://github.com/BabylonJS/Babylon.js/pull/6630/commits/6b367047b53745174d8c4ac9a6687c42c6e67070

sebavan:
LGTM :-) Could you add an entry in dist/preview release/what's new.md ? and all the build will be green and ready to go in.

Dok11:
> add an entry in dist/preview release/what's new.md ?

Seems I cant understand.. it it written here what I must not commit dist/preview
https://doc.babylonjs.com/how_to/how_to_start#do-not-commit

sebavan:
This is utterly confusing I agree and I will fix the doc ASAP, the what s new file is the only one that needs to be updated. This is our base to create the changelog for the next release.

src/Misc/tools.ts
```diff
@@ -15,6 +15,7 @@ import { PromisePolyfill } from './promise';
 import { TimingTools } from './timingTools';
 import { InstantiationTools } from './instantiationTools';
 import { GUID } from './guid';
+import { IScreenshotSize } from './screenshotTools';
```
deltakosh: The only problem is that we do not want to import the file automatically for size reason (treeshacking). Do you mind moving the interface in its own file?

Dok11:
> Do you mind moving the interface in its own file?

I wanted itm but cant understand system for defining interfaces. In my projects I usual define interfaces in self directory called `interfaces` and use only one `export` construction per one file.
But here I dont see similar system or any other. So I make `interfaces` folder in the `/Misc`.

deltakosh:
iInspectable is already in the misc folder so it is ok to drop that file there as well
no need for a new folder

Dok11:
> iInspectable is already in the misc folder so it is ok to drop that file there as well
no need for a new folder

But in this folder already have many interfeces, these may be moved into `interfaces` folder. I dont did it because my PR not about it and not about refeactoring

deltakosh:
I agree but let's do it later then. (Also you have a doc issue)

Dok11:
> Also you have a doc issue

I thinked documentation generated automatic, it is not right? Or you about `what's new.md` file?

Dok11:
I updated file `what's news`. Is now all ok?

deltakosh:
You can see the checks here: https://github.com/BabylonJS/Babylon.js/pull/6630/checks

The documentation problem: "[18:09:46] Missing text for Interface : IScreenshotSize (id: 24186) babylon.d.ts:31576"

src/Misc/interfaces/screenshotSize.ts
```diff
@@ -0,0 +1,17 @@
+export interface IScreenshotSize {
```
deltakosh: You need to add a comment to explain the Interface

Dok11:
Aaaand did I pass this challenge?

deltakosh:
YOU DID :)
Congrats my friend!

Dok11:
Yes! Thanks man!
One contributor more haha =)

PR #6640, opened 2019-07-29
Fixed VolumetricLightScattering post-process to use a custom vertex s…

julien-moreau:
Fixed VolumetricLightScattering post-process to use a custom vertex shader instead of the depth vertex shader.
Fixes: https://forum.babylonjs.com/t/volumetric-lightscattering-post-process-not-work-latest/4790

Also fixed water material from Materials Library to implement image processing and disabling the automatic clipping.


julien-moreau:
Updating the what’s new ASAP !

deltakosh:
Ok no worries ;)

PR #6644, opened 2019-07-30
ScreenshotTools. Implementation usage of precision in combination height and width params

Dok11:
Now CreateScreenshot method provide possible to use precision with height and width parameters in any combination.
Early `precision` param ignore height and width even them was declarated.


sebavan:
Perfect, thanks a lot !!!

PR #6647, opened 2019-07-30
add positional mode to billboards

delaneyj:
Per discussions with @sebavan [on the forum](https://forum.babylonjs.com/t/lookat-vrhelper-pointing-at-feet/4764/5) it came to both our attention that billboards use the camera's orientation and not the position.  This can lead to odd behavior, especially in VR where unless the billboarded mesh is directly ahead of camera's viewport center it will [swing as shown here](https://playground.babylonjs.com/#AVXIV3).  I choose `128` just to signify its a different mode at still keep within a char in hopes v8 could optimize still.

This is my first PR to Babylon.js and may not be the most efficient way to support using positions so open to any improvements.  I did run the test validations and saw no changes in current behavior.
   
Since its an extra bit wise flag its easy to add to any current example with minimal changes to code base.  Playground example above with the extra flag
```

    // This targets the camera to scene origin

    // This attaches the camera to the canvas

    // This creates a light, aiming 0,1,0 - to the sky (non-mesh)

    // Default intensity is 1. Let's dim the light a small amount
    light.intensity = 0.7;

    // Our built-in 'sphere' shape.
    var sphere = BABYLON.MeshBuilder.CreateBox("sphere", { height: 1, width: 10, depth: 1 }, scene);
    sphere.billboardMode = BABYLON.Mesh.BILLBOARDMODE_Y | BABYLON.Mesh.BILLBOARDMODE_USE_POSITION
    // Move the sphere upward 1/2 its height
    sphere.position.y = 1;

    var ground = BABYLON.MeshBuilder.CreateGround("ground", { width: 6, height: 6 }, scene);

    return scene;

};
```

### DIFF
--- a/src/Meshes/transformNode.ts
+++ b/src/Meshes/transformNode.ts
@@ -36,6 +36,10 @@ export class TransformNode extends Node {
      * Object will rotate to face the camera
      */
     public static BILLBOARDMODE_ALL = 7;
+    /**
+     * Object will rotate to face the camera's position instead of orientation
+     */
+    public static BILLBOARDMODE_USE_POSITION = 128;
 
     private _forward = new Vector3(0, 0, 1);
     private _forwardInverted = new Vector3(0, 0, -1);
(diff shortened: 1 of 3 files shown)


sebavan:
Love it, can you also update : https://github.com/BabylonJS/Babylon.js/blob/master/dist/preview%20release/what's%20new.md

After we ll let @deltakosh review tomorrow but it should be all good !

Thx.

delaneyj:
> Love it, can you also update : https://github.com/BabylonJS/Babylon.js/blob/master/dist/preview%20release/what's%20new.md
> 
> After we ll let @deltakosh review tomorrow but it should be all good !
> 
> Thx.

Should I be updating `Babylon.js/what's new.md` and it will copy to `dist` on build/ci or editing the `dist` version directly @sebavan ?

sebavan:
You should edit exactly the one I sent from dist/preview release. It will be uploaded to release once we will ship 4.1. We chose a manual approach cause commit comments can sometimes suck :-) mostly mine.

delaneyj:
Gotcha, was making sure the `dist` contents wouldn't be blown away on a rebuild.  Thanks for all the help @sebavan!

sebavan:
It is all green for me, @deltakosh will probably only have to press the merge button :-)

src/Meshes/transformNode.ts
```diff
@@ -1050,38 +1054,54 @@ export class TransformNode extends Node {
...
-            if ((this.billboardMode & TransformNode.BILLBOARDMODE_ALL) !== TransformNode.BILLBOARDMODE_ALL) {
-                TmpVectors.Matrix[0].decompose(undefined, TmpVectors.Quaternion[0], undefined);
-                let eulerAngles = TmpVectors.Vector3[1];
-                TmpVectors.Quaternion[0].toEulerAnglesToRef(eulerAngles);
+            if (this.billboardMode & TransformNode.BILLBOARDMODE_USE_POSITION) {
+                this.lookAt(camera.position);
 
                 if ((this.billboardMode & TransformNode.BILLBOARDMODE_X) !== TransformNode.BILLBOARDMODE_X) {
-                    eulerAngles.x = 0;
+                    this.rotation.x = 0;
```
deltakosh: The problem with that (and `this.lookAt` earlier) is that it will update values which will not be used before NEXT frame

This must happen before we evaluate `this._worldMatrix`

delaneyj: Sorry I'm not seeing the right way to resolve this from your comments.  I understand the next frame issue but don't know enough about the code base to see clearly where to use the `lookAt` to effect this frame.

Looking at the `transformnode.lookAt` code its not clear that it only effects the next frame.  I'm not sure what to do from here.

deltakosh: setting this.position or this.rotation is the problem as it is used before in the computeWorldMatrix function

The idea is to move your code at the beginning of the function before we compute _worldMatrix using position and rotation

delaneyj:
@deltakosh hopefully I understood the issue correctly and doing the positional based changes prior to the call to ` this._updateCache();`.

deltakosh:
Thanks a million!

PR #6657, opened 2019-08-02
 Playground v4.6

Jaskar:
Bug fixes.
Style fixes.

/!\ Need to upload the 4.1.0 version of BJS on unpkg.com


deltakosh:
We cannot upload on unpkg.com. it is automatic
the latest version (should not be called 4.1 but "latest" should always come from preview.babylonjs.com

PR #6661, opened 2019-08-04
Fix Scene.InputManager.detachControl: Wheel event was not detached.

RamilKadyrov:
Wheel event was not detached.


deltakosh:
THanks a lot!

PR #6662, opened 2019-08-04
TexturePacker for Sprite WITH ERRORS

BabylonJSGuide:
DO NOT MERGE CONTAINS ERRORS

I need some help. Currently I am testing that the changes are backward compatible and am having a problem displaying the correct cell.

When you load this page (sprit cellIndex set to 0) https://tender-murdock-275855.netlify.com/sprites/test1.html you will see sprite cellIndex 44 upside down in the center as a sprite, after a moment the data for cellIndex 0 is used as a texture for the plane on the right.

The data for the plane texture can be seen on lines 68, 69 and 75, 76 of https://github.com/BabylonJSGuide/oddsNEdnds/blob/master/sprites/Test1.html, the console values seem also confirm the correct values.

My thinking is that I have made a stupid error in the sprite.vertex.fx (as in this PR). As far as I can see what I have done for vUV.x and vUV.y in the sprite,vertex.fx file matches what I did on line 68, 69, 75, 76 in the test1.html file. 

Any help in spotting my error (whatever it is) would be gratefully received as I have been looking at it for a couple of days now  and can not see what I have done wrong. (Own mistakes often remain invisible). Thank you.


src/Shaders/sprites.vertex.fx
```diff
@@ -40,9 +39,12 @@ void main(void) {
...
 	
 	// Texture
-	vec2 uvOffset = vec2(abs(offset.x - cellInfo.x), 1.0 - abs(offset.y - cellInfo.y));
+	vec2 uvOffset = vec2(abs(offset.x - inverts.x), abs(1.0 - offset.y - inverts.y));
+	vec2 uvPlace = cellInfo.xy;
+	vec2 uvSize = cellInfo.zw;
 
-	vUV = (uvOffset + cellInfo.zw) * uvScale;
+	vUV.x = uvPlace.x + uvSize.x * uvOffset.x;
+	vUV.y = 1.0 - (uvPlace.y + uvSize.y * uvOffset.y);
```
deltakosh: I think this is the only problem:
```
vUV.y = (uvPlace.y + uvSize.y * uvOffset.y);
```
Because the sprite texture is not loaded with invertY ;)

BabylonJSGuide: See I knew it was something stupid :)

deltakosh: Not at all!

BabylonJSGuide:
Nearly there still have a minor issue to solve. When I am happy I will close this PR and open a new one.

A point of clarification if you do not mind:

I am using assetsManager to load the json file as in my PR spriteManager.ts lines 233 to 264.

Is this a good a way as any or would there be a better way?

deltakosh:
I would recommend not loading it with the assetManager and more with a direct XHR call as users may not want to drag the AssetsManager dependency if they only use the Sprites sub system

BabylonJSGuide:
Opening a new PR for this

Issue #6624, opened 2019-07-25
Allow (VR) 180 Videos using the VideoDome


RaananW:
At the moment, the video dome is a sphere supporting full 360 videos. 180 videos should be projected on half of the dome (from part), and leave the back part texture-less.

Initial conversation and playground can be found here - 

https://forum.babylonjs.com/t/180-degree-3d-video-vr180-support/4681

deltakosh:
Thanks a lot @RaananW !!!

PR #6663, opened 2019-08-05
Video dome 180

RaananW:
closing https://github.com/BabylonJS/Babylon.js/issues/6624

The video dome will now support 180 VR videos. the parameter can be set dynamically or using the options (options.halfDomeMode) .


RaananW:
The second sphere is a black sphere with no material. It's just there to hide the back of the sphere. Nothing is rendered on it.
So, the main mesh (sphere) always renders the video. In 180 mode there is an extra black mesh to hide the back. 
There is no way around it without a second mesh sadly.  I could use a sphere with slice 0.5, but then I need to change the default uv coordinates of the sphere, which, I thought, was an overkill for this.

deltakosh:
Ok I misunderstood :) so all good for me!

PR #6706, opened 2019-08-10
Fix for case when options.cap = Mesh.NO_CAP

QuintusHegie:
**The Problem**
When options.cap was explicitly set to Mesh.NO_CAP, it was treated as Mesh.CAP_ALL (as if the default option was wanted).
https://forum.babylonjs.com/t/meshbuilder-cap-parameter-for-cylinder-extrudepolygon-sides-parameter-for-box/2959/6
**The Cause**
This is because Mesh.NO_CAP evaluates to 0 and 0 || 3 = 3 (=Mesh.CAP_ALL).
**The Solution**
With this change, the cap will remain Mesh.NO_CAP when you specify this as the value for options.cap parameter.
When you omit the options.cap parameter, the cap will keep defaulting to Mesh.CAP_ALL as desired.


deltakosh:
Excellent! Thanks a lot

PR #6712, opened 2019-08-12
Fix Right Hand coordinates with directional lights and shadows

CedricGuillemet:


src/Lights/directionalLight.ts
```diff
@@ -182,9 +182,15 @@ export class DirectionalLight extends ShadowLight {
         var xOffset = this._orthoRight - this._orthoLeft;
         var yOffset = this._orthoTop - this._orthoBottom;
 
-        Matrix.OrthoOffCenterLHToRef(this._orthoLeft - xOffset * this.shadowOrthoScale, this._orthoRight + xOffset * this.shadowOrthoScale,
-            this._orthoBottom - yOffset * this.shadowOrthoScale, this._orthoTop + yOffset * this.shadowOrthoScale,
-            this.shadowMinZ !== undefined ? this.shadowMinZ : activeCamera.minZ, this.shadowMaxZ !== undefined ? this.shadowMaxZ : activeCamera.maxZ, matrix);
+        if (this.getScene().useRightHandedSystem) {
+            Matrix.OrthoOffCenterRHToRef(this._orthoLeft - xOffset * this.shadowOrthoScale, this._orthoRight + xOffset * this.shadowOrthoScale,
```
deltakosh: This one seems legit! thanks a lot!

src/Lights/directionalLight.ts
```diff
@@ -204,10 +210,19 @@ export class DirectionalLight extends ShadowLight {
      */
     public transferToEffect(effect: Effect, lightIndex: string): DirectionalLight {
         if (this.computeTransformedInformation()) {
-            this._uniformBuffer.updateFloat4("vLightData", this.transformedDirection.x, this.transformedDirection.y, this.transformedDirection.z, 1, lightIndex);
+            if (this.getScene().useRightHandedSystem) {
+                this._uniformBuffer.updateFloat4("vLightData", -this.transformedDirection.x, -this.transformedDirection.y, -this.transformedDirection.z, 1, lightIndex);
```
deltakosh: Why should we negate light direction when using right handed system?

CedricGuillemet: Because the light computation is in view space https://github.com/BabylonJS/Babylon.js/blob/master/src/Shaders/ShadersInclude/lightFragment.fx#L14
and
https://github.com/BabylonJS/Babylon.js/blob/master/src/Shaders/default.fragment.fx
and switching to RH flips  the direction for the dots. 
More empirically, the lighting is wrong when not negating vLightData. I can clearly see why for x and z. It's more fuzzy for y.

PR #6741, opened 2019-08-20
Update videoDome.ts

RaananW:
Due to the way the video is displayed, eyes should be switched in halfDome mode.


src/Helpers/videoDome.ts
```diff
@@ -230,7 +230,12 @@ export class VideoDome extends TransformNode {
                 // Use 0.99999 to boost perf by not switching program
                 this._videoTexture.uScale = this._halfDome ? 0.99999 : 0.5;
                 this._onBeforeCameraRenderObserver = this._scene.onBeforeCameraRenderObservable.add((camera) => {
-                    this._videoTexture.uOffset = camera.isRightCamera ? 0.5 : 0.0;
+                    if(this._halfDome) {
```
sebavan: Could we keep left and right values in var ??? this would prevent the if in the observable ???

RaananW: sure

PR #6744, opened 2019-08-20
Added support of custom blur width when using lens flare post-process.

julien-moreau:
Also updated the what's new according to previous pull requests.


deltakosh:
There is a tiny conflicts

julien-moreau:
Fixed!

PR #6758, opened 2019-08-21
shader preprocess for opengl(ES) compatibility

CedricGuillemet:
Doing tests now.

### DIFF
--- a/src/Engines/Native/nativeShaderProcessor.ts
+++ b/src/Engines/Native/nativeShaderProcessor.ts
@@ -10,6 +10,15 @@ const attributeLocations: { [kind: string]: number } = {
     [VertexBuffer.MatricesIndicesKind]: 8,
     [VertexBuffer.MatricesWeightsKind]: 9,
 };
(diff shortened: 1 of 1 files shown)


src/Engines/Native/nativeShaderProcessor.ts
```diff
@@ -45,7 +54,12 @@ export class NativeShaderProcessor extends WebGL2ShaderProcessor {
                 throw new Error("Exceeded maximum custom attributes");
             }
         }
-
+        let newName = attributeBGFXName[name];
+        if (newName === undefined) {
+            throw new Error("Can't find bgfx name mapping");
+        }
+        attribute = attribute.replace(name, newName);
```
bghgary: why do you need both of these lines?

bghgary:
Don't use the draft label... Make the PR a draft instead.

bghgary:
https://github.blog/2019-02-14-introducing-draft-pull-requests/

PR #6766, opened 2019-08-22
Changes to make Babylon Native work

syntheticmagus:
Merging native branch to master. Most notable changes are as follows:
 - Add NativeEngine
 - Minor changes to shader preprocessing
 - Minor changes to texture loading
 - Assorted changes to Engine to allow NativeEngine to override behavior where required


src/Engines/engine.ts
```diff
@@ -5581,6 +5605,8 @@ export class Engine {
             return;
         }
 
+        this._boundUniforms[channel] = uniform;
```
deltakosh: Can we remove this guy?

syntheticmagus: Yep, that seems obsolete. Removed.

PR #6768, opened 2019-08-22
glue for extent parameters queries

CedricGuillemet:
glue for set extent


deltakosh:
We have a linting issue :)

PR #6787, opened 2019-08-27
Update RGBD Tools

sebavan:


Vinc3r:
hey! Have I seen RGBD lightmapping ?! Wait a minute, I need a doc' about that 🙌

sebavan:
The main goal is to do like .env a way to reduce the size of 16b or 32b pp DDS files which could be use for lightmapping.

This way, you can store hdr data in a smaller png.

Vinc3r:
yeah, and if I'm not wrong that's a way that Unity & Unreal, or in webGL side Shapespark. and playcanvas use for lightmap storring isn't it?

sebavan:
I have actually no idea :-) but it sounded like a good idea to reduce the network traffic until more things like .basis :-)

To generate the RGBD texture from a regular babylon you can use: 
https://playground.babylonjs.com/#FZS1ES

And to load them something like adding this to your code:
```
texture.isRGBD = true;
BABYLON.RGBDTextureTools.ExpandRGBDTexture(texture);
```

Like in the following playground: https://playground.babylonjs.com/#FZS1ES#3

Please note that the PG does not load RGBD encoded ones, it is just to highlight the code usage.

Vinc3r:
Nice to see my Cornell Box used as example ^^

Thanks for explanation, 'will try to going further when I have the time.

PR #6793, opened 2019-08-28
[WIP] Radiosity solver for baking global illumination on static geometry

CraigFeldspar:
This is a preview of the upcoming global illumination solver for static geometries.

It uses the progressive refinement radiosity algorithm described in this paper : 
https://developer.nvidia.com/gpugems/GPUGems2/gpugems2_chapter39.html  

You can bake either only direct light, or full solution (direct + indirect), or use "background baking", (baking a little on each frame).  
For the sake of completeness, the "smart uv mapping" algorithm from Blender has been included, thus enabling to add uvs for baking on custom geometry. Handmade uvs in a 3D software are still WAY better though, so only use it on procedural geometry, when offline uv unwrapping is not a choice.  

Playground example is here : (local url)
http://localhost:1338/Playground/index-local.html?#VRLE1V#2

TODO :

+ [x] Fix hemispherical rendering
+ [ ] Add the possibility to have 2 levels of texture resolution, one high for direct light, and one lower for indirect light
+ [ ] Consider diffuse color when bouncing light off a surface (now only monochrome light is bouncing)
+ [ ] Some code optimisation


FishOrBear:
When is it available

CraigFeldspar:
This feature is still in development, thus highly experimental. PR will be reopened when the code will be more robust

PR #6809, opened 2019-09-04
Fixes to correctly specify the texture type for native.

syntheticmagus:


src/Engines/nativeEngine.ts
```diff
@@ -1088,12 +1089,16 @@ export class NativeEngine extends Engine {
...
     }
 
-    private static _GetNativeTextureFormat(format: number): number {
-        switch (format) {
-            case Engine.TEXTUREFORMAT_RGBA:
-                return NativeTextureFormat.RGBA;
-            default:
-                throw new Error("Unexpected texture format: " + format + ".");
+    private static _GetNativeTextureFormat(format: number, type: number): number {
+        if (format == Engine.TEXTUREFORMAT_RGBA && (type == Engine.TEXTURETYPE_UNSIGNED_INT || type == Engine.TEXTURETYPE_UNSIGNED_BYTE)) {
```
bghgary:
```suggestion
        if (format == Engine.TEXTUREFORMAT_RGBA && type == Engine.TEXTURETYPE_UNSIGNED_INT) {
```

No need to check for 0 twice 😄 and thus note can be removed below.

PR #6811, opened 2019-09-04
Playground 4.1.4 : Fix some bugs.

Jaskar:
- Little CSS fix
- Libraries weren't correctly loading after version selection
- Example selection after version selection was buggy
- Refactoring


Jaskar:
Hi,
Unfortunately, we have now some errors concerning scripts that "don't want to load".
Everything seems to working fine ... But we have these errors.
I'm searching for the reason.

PR #6821, opened 2019-09-07
simpler loading for svg icons

lockphase:
Added in functionality for loading multiple svg icons from an external svg file via icon id as requested in [https://forum.babylonjs.com/t/babylonjs-gui-svg-icon-id-support/4786/3](url).

Users can now use `new BABYLON.GUI.Image("image", "sample.svg#icon_id");` without having to manually set sourceLeft, sourceTop, sourceWidth and sourceHeight variables. These variables are automatically computed from the sample.svg file itself. Pre-requisite: sample.svg must have attributes (width, height and viewbox) defined. Icon transform will use defaults if they are not available. Formulas are derived in accordance with [https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#Matrix](url) 

An observable is added to notify when the variables have been computed for users who might want to set custom icon dimensions while preserving the original svg image's aspect ratio. Note that a  HTMLObjectELement is added to document in order to provide said functionality.


lockphase:
DO NOT MERGE YET, I FOUND A BUG!

Creating multiple gui images from the same svg source will fail. The check if object alr exists in document is throwing up the object but said object is empty for the 2nd icon onwards. Suspect a race condition. Need to think up a fix.

PR #6854, opened 2019-09-16
added xml loader

Null0924:
Pull request with an XML Loader for BABYLON GUI.


gui/src/2D/xmlLoader.ts
```diff
@@ -0,0 +1,323 @@
...
+        for (let i = 0; i < value.length; i++) {
+            element = element[value[i]];
+        }
+        return element;
+
+    }
+
+    private _createGuiElement(node: any, parent: any, linkParent: boolean = true): void {
+        try {
+            let className = this._getChainElement("BABYLON.GUI." + node.nodeName, true);
```
deltakosh: That will not work if you are using ES6 modules in the client. The classes are not on the global object anymore.
To make it work you will need to update all GUI classes I'am afraid and reference them like here:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/Node/Blocks/addBlock.ts#L66

Then creating one will be easy:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/Node/nodeMaterial.ts#L1174

gui/src/2D/xmlLoader.ts
```diff
@@ -0,0 +1,323 @@
...
+                throw "GUILoader Exception : Duplicate ID, every element should have an unique ID attribute";
+            }
+            return guiNode;
+
+        } catch (e) {
+            throw "GUILoader Exception : Error parsing Control " + node.nodeName + "," + e + ".";
+        }
+    }
+
+    private _parseGrid(node: any, guiNode: any, parent: any): void {
```
deltakosh: Maybe we can move the xml parsing on each control?

deltakosh: At least for some specific controls requiring specific code

gui/src/2D/xmlLoader.ts
Null0924: Oh I was not aware. I will update the classes and change the referencing.

gui/src/2D/xmlLoader.ts
Null0924: It can be done. This would make them a little bit more tightly coupled with the xmlLoader. Maybe this would not make sense in a scenario when other loaders are implemented?

deltakosh: Well this will just add a function at control level that will do nothing and only add it in grid (only constrains is to avoid referencing the xmlRead to avoid circular dependency)

gui/src/2D/xmlLoader.ts
Null0924: This should be fine. Updated the GUI classes and changed the way it is accessed in the xmlLoader.

gui/src/2D/xmlLoader.ts
Null0924: I did look into it but I did not think of a way to set the parsing control level. The problem is that as the GUI is generated recursively, it calls the same _parseXml function several times. Not sure how we can avoid circular dependency here. Open for suggestions though if this is a must.

deltakosh: Well then just live it here. At least it will not be coupled at all :)

deltakosh:
All good! merging :) Now please update GUI doc ;)

Congrats btw!!! it is an excellent new feature

Null0924:
> All good! merging :) Now please update GUI doc ;)
> 
> Congrats btw!!! it is an excellent new feature

Thank you very much. Super happy to be able to give back to the community.

PR #6879, opened 2019-09-24
Suggested solution to scenes/engines being disposed incorrectly

RaananW:
super.dispose() was nulling the rendering canvas, which prevented scene.detachControl to work.


deltakosh:
Wonderful Thanks!

PR #6899, opened 2019-09-27
Playground : Fix an iPhone menu bug

Jaskar:
I don't know why CSS overflow property is breaking everything on iPhone ...
But let's do it another way.
(I'm not sur that this works perfectly on iPhone 5 and older.)


deltakosh:
Safari is a pile of crap :)

PR #6916, opened 2019-10-02
GLTFLoader: Allow extensions to override buffer view loading

zeux:
This adds a way for extensions to override buffer view loading by
implementing loadBufferViewAsync; this is useful for decompression
extensions.

Specifically this is useful for an (upcoming) extension, MESHOPT_compression,
that is used by gltfpack (https://github.com/zeux/meshoptimizer#gltfpack); the
extension specifies compression on the buffer view level and decompresses each
view.

I'm not sure what logic governs whether underscore should be used for function
names or not so I left it out to match the name of GLTFLoader method.

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -1397,6 +1397,11 @@ export class GLTFLoader implements IGLTFLoader {
      * @returns A promise that resolves with the loaded data when the load is complete
      */
     public loadBufferViewAsync(context: string, bufferView: IBufferView): Promise<ArrayBufferView> {
+        const extensionPromise = this._extensionsLoadBufferViewAsync(context, bufferView);
+        if (extensionPromise) {
+            return extensionPromise;
+        }
+
         if (bufferView._data) {
             return bufferView._data;
         }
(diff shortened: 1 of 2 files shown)


zeux:
Just as an example here's the proof of concept implementation for this that I wrote and confirmed that it works:

```js
        var MESHOPT_compression = (function () {
            function MESHOPT_compression(loader, decoder) {
                this.name = "MESHOPT_compression";
                this.enabled = true;
                this._loader = loader;
                this._decoder = decoder;
            }
            MESHOPT_compression.prototype.loadBufferViewAsync = function (context, bufferView) {
                if (bufferView.extensions && bufferView.extensions.MESHOPT_compression) {
                    var view = this._loader.loadBufferViewAsync(context, bufferView);
                    if (bufferView._decoded) {
                        return bufferView._decoded;
                    }

                    var decoder = this._decoder;
                    bufferView._decoded = Promise.all([view, decoder.ready]).then(function (res) {
                        var source = res[0];

                        var extensionDef = bufferView.extensions.MESHOPT_compression;
                        var count = extensionDef.count;
                        var stride = extensionDef.byteStride;

                        var result = new Uint8Array(new ArrayBuffer(count * stride));

                        switch (extensionDef.mode) {
                        case 0:
                            decoder.decodeVertexBuffer(result, count, stride, source);
                            break;

                        case 1:
                            decoder.decodeIndexBuffer(result, count, stride, source);
                            break;

                        default:
                            throw new Error('GLTFLoader: Unrecognized meshopt compression mode.');
                        }

                        return Promise.resolve(result);
                    });

                    return bufferView._decoded;
                }
                return null;
            };
            return MESHOPT_compression;
        }());
```

PR #6935, opened 2019-10-04
Added support for using HTTP range requests from a glTF binary

### LINKED ISSUE #6242: Progressive loading of GLB files
It would be great if the GLB loader could use range requests and the `MSFT_lod` extension to progressively load textures and models. My understanding is that it currently loads the entire binary file, making the MSFT_lod extension not useful for GLB's.

bghgary:
Fixes #6242

- Added ability to set request headers and get response headers to `WebRequest`.
- Factored out a `FileTools.RequestFile` function from the `FileTools.LoadFile` that only deals with URLs with a callback `onOpened` to `FileTools.RequestFile` that allows setting request headers and `onSuccess` that returns the web request to allow getting the response headers.
- Added additional functions to `ISceneLoaderPlugin` interfaces to allow plugins to change the behavior of loading data from the source files.
- Modified the glTF loader and `MSFT_lod` extension to use the additional functions in `ISceneLoaderPlugin` to use HTTP range requests to progressively load glTF binary files.

Unit tests are incoming.


src/Misc/stringTools.ts
```diff
@@ -21,4 +21,22 @@ export class StringTools {
...
         return str.indexOf(suffix) === 0;
     }
+
+    /**
+     * Decodes a buffer into a string
+     * @param buffer The buffer to decode
+     * @returns The decoded string
+     */
+    public static Decode(buffer: Uint8Array | Uint16Array): string {
+        if (typeof TextDecoder !== "undefined") {
```
sebavan: should it be a string here ?

deltakosh: Yep..this is the way it works

bghgary:
> Just a quick question about all the enabled, are we not changing the way it worked before where it seems to default to enabled ???

Yes, but I think this is the right behavior. Before, the extensions were enabled even if you don't specify them in `extensionsUsed`. The extensions will run even though they shouldn't. I will document this in the `what's new`.

PR #6949, opened 2019-10-09
WebXR changes

RaananW:
This PR takes care of a few issues:

1) The Mozilla WebXR emulator should now be working on firefox. Install the plugin and open a webxr demo. Make sure to enable on-canvas debugging using `camera._updateForDualEyeDebugging()`
2) Eye pose is now using the position and orientation when no global pose is enabled. According to the specs.
3) gamepad object is exposed in the `WebXRController` class.  Using this object you can track button pressed on the controller of choice. Notice that this object is not in charge of changing the mesh's position.
4) Bug fixing - When no grip available at all, the controller class will still initialize correctly (needed for the webxr emulator)
5) Solved a bug where a single frame was still rendering after the xr session has ended.

Tested on the oculus rift and the webxr emulator


RaananW:
Working on fixing the CI issues

PR #6961, opened 2019-10-10
Fix vignette aspect ratio when not using postProcessing

haxiomic:
If you're not using the post processing rendering pipeline (i.e. rendering directly to the canvas drawing buffer) then the vignette aspect ratio will be wrong.

This happens because when binding the `imageProcessingConfiguration` in the material, no value is provided for the aspectRatio parameter so it defaults to 1, whereas when binding via the post processing pipeline the correct aspect is provided. This change resolves this by using the current render target (or canvas drawing buffer) aspect ratio when the aspect ratio isn't supplied. And semantically, the aspect ratio parameter has been changed to `overrideAspectRatio` to clarify

OffscreenVignetteBug
In the gif, the circular vignette is correct


sebavan:
Hey @haxiomic it is fun to see you there again :-) Thanks for the fix

haxiomic:
It's been a little while @sebavan! :)

haxiomic:
@sebavan It's been great to see all the new stuff in Babylon! I'm really loving the editor. Man that makes life so much easier!

PR #6966, opened 2019-10-11
Add support for reordering glTF loader extensions

### LINKED ISSUE #6882: GLTF Extension Name Load Order Priority
Yo @bghgary ... This is a request to officially support GLTF Extension priority.

Some native babylon.js extensions **stop** the GLTF loading chain when parsing materials.
Like PBR Specular Material Does. This wont allow other extensions that need to run.

It turns out the **ExtensionNames** array specifies the loading order for GLTF Extensions.

Can you please add support to register extensions with a **priority boolean** that would **insert** into the **ExtensionNames** array instead of **pushing** them on the bottom.

Or whatever you have to do to **Allow All Extensions** to run.

Thank you so much :)

bghgary:
Fixes #6882

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -95,6 +95,7 @@
 - Added support for morph target names via `mesh.extras.targetNames` when loading a glTF ([zeux](https://github.com/zeux))
 - Added support for using HTTP range requests when loading `MSFT_lod` extension from a glTF binary. ([bghgary](https://github.com/bghgary))
 - Added a flag to enable/disable creation of instances for glTF loader. ([bghgary](https://github.com/bghgary))
+- Added an order property to glTF loader extensions to support reordering. ([bghgary](https://github.com/bghgary))
 
 ### Materials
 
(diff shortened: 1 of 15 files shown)


MackeyK24:
Yo @bghgary... so how do I reorder extensions

bghgary:
Here is a PG that shows how to register a custom extension with order, but note that the changes haven't been deployed to preview yet.

https://playground.babylonjs.com/#6XJGUP

MackeyK24:
The playground get an undefined error for me.

But basically the gltf extension has a new property called order that we should set in the construction of our custom gltf extension... is that right ???

bghgary:
> The playground get an undefined error for me.

Right. As I said, the code is not deployed yet.

> But basically the gltf extension has a new property called order that we should set in the construction of our custom gltf extension... is that right ???

Yes. You can also change the order by updating the order number [onExtensionLoaded](https://doc.babylonjs.com/api/classes/babylon.gltffileloader#onextensionloaded) event.

MackeyK24:
Sweet... Thanks Gary ... will wait for next alpha release then

PR #6968, opened 2019-10-12
gui svg icon load bugfix for Chrome

lockphase:
Started off wanting to make a PG with a sample svg to demonstrate the new functionality for loading svg assets via icon id. One thing lead to the next and while stuff worked locally in ffox, it broke in Chrome. :(

So submitting this PR for Chrome fix, tested in localDev. Note that 'npm run build' failed with 

```
[09:19:28] 'buildMin' errored after 25 s
[09:19:28] Error in plugin "webpack-stream"
Message:
    [at-loader] ..\..\src\sceneComponent.ts:215:15
    TS2343: This syntax requires an imported helper named '__spreadArrays', but module 'tslib' has no exported member '__spreadArrays'.
```

Something wrong with tslib ? Anyways, I went ahead with lint check and committed. Hope it helps.

### DIFF
--- a/Playground/textures/gui/bjs_demo1.svg
+++ b/Playground/textures/gui/bjs_demo1.svg
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:cc="http://creativecommons.org/ns#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape" width="64.0005" height="64.0007" viewBox="0 0 16.9332 16.9333" version="1.1" id="svg8" inkscape:version="0.92.4 (5da689c313, 2019-01-14)" sodipodi:docname="bjs_demo1.svg">
+<title id="title815">babylonjs demo</title>
+<defs id="defs2">
+<linearGradient inkscape:collect="always" id="linearGradient975">
+<stop style="stop-color:#808080;stop-opacity:1" offset="0" id="stop971" />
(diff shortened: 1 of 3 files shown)


deltakosh:
I had this issue as well, I had to run git clean -dxf in the root folder to make it work (Apparently linked to TS version changed)

lockphase:
So, I think the fix worked. But, I'm now see two things:

a) On Chrome, when the browser loads the first svg icon it encounters, there's a very brief flicker on screen where the entire svg doc is visible. It comes and goes. I've not had it happen on ffox like at all. No idea why since I've explicitly called 
```
svgImage.width = "0%";  
svgImage.height = "0%";
```
This will need some more work, sigh...

b) Executing via a nodejs server results in multiple file fetch calls for said svg doc. So if the client side script calls for a bunch of svg icons (as is normal during page load) that are from the same svg file, its going to choke up server bandwidth unnecessarily. I believe this is from the codes
```
        this._domImage.onload = () => {
            this._onImageLoaded();
        };
```
So sample.svg#icon1, sample.svg#icon2 will result in 2 fetch calls to the server as the codes can't see that they are actually from the same file.

Now I'm torn on this. Ideally, I want to leave the default image load codes as is, they are perfectly fine and working. But in the context of batch loading, they don't work. I'm hesitant to touch these since it might break backward compatibility. What's the best way ?

deltakosh:
a) Can you repro somewhere? Would like to see that. It should work by creating the svg by code and not attaching it to the DOM though
b) Yeah this one is a bit more complex..Not sure we can do anything easily without changing everything in the code

lockphase:
> a) Can you repro somewhere? Would like to see that. It should work by creating the svg by code and not attaching it to the DOM though

From the demo pg in forum https://playground.babylonjs.com/#E5CARD Hard refresh the page in Chrome multiple times. On my screen, there is a very brief flicker showing the entire svg doc. Likely system dependent,

lockphase:
I've tried some workarounds, none worked. getBBox requires the element to be renderable, hence attaching the svg document to DOM. If I use ajax and read the spritesheet as xml, getBBox will return all zeros even though the correct svg element is referenced. I also tried z-index = -9999 via css, nada. Toggling svg layer display, nada. So far, it seems like hiding the brief flicker behind a loading screen is the best way but that's client-side. Couldn't find the polyfill for getBBox either. So officially, I'm stumped. 

sorry, really don't like adding a feature that introduces more problems downstream. :(

deltakosh:
Maybe it is then worth mentioning the known issue in the doc?

lockphase:
Done and thanks !

Issue #6917, opened 2019-10-02
God ray demo is broken


azukaar:
# Bugs

- Bug repro on [playground](https://www.babylonjs-playground.com/#AU5641):
- Expected result: 

![image](https://user-images.githubusercontent.com/7872597/66033822-f33a7a00-e4ff-11e9-8357-06c1d8d46183.png)

- Current result:

![image](https://user-images.githubusercontent.com/7872597/66033851-fdf50f00-e4ff-11e9-9555-db453f880cc5.png)

deltakosh:
pinging @julien-moreau

julien-moreau:
I’m having a look ASAP!
Thanks for reporting

deltakosh:
any news?

julien-moreau:
Nope, the mesh is rendered but doesn’t appear in the render target used by the post-process. This is weird and I’m investigating

deltakosh:
Fixed!

azukaar:
> Fixed!

Just curious, what was the issue?

julien-moreau:
The vertex shader has been modified and the line computing the final transformation was removed :(

PR #6969, opened 2019-10-12
Fixed broken Volumetric Light Post-Process pass #6917

julien-moreau:
Finally found, the vertex shader was modified and did not compute the final screen position!


deltakosh:
LOL! well done

PR #6970, opened 2019-10-12
[DO NOT MERGE] feat: Allow CustomMaterial to use UVs > 2

azukaar:
Allow CustomMaterial to use UVs values up to UVs6.

This piggy back on existing values in VertexData, and edit slightly the StandardMaterial to use those values if they exists (Since CustomMaterial is based on StandardMaterial).

It doesnt impact performance since you wouldnt normally have uvs > 2 if you are using StandardMaterial and it checks for the existance of those before feeding them in.

[DO NOT MERGE]

There seems to be an index issue now with the standardMaterial I'm investigating, but somehow, now my projected light texture randomly mixes with normal maps all around (and it's extremmly random depending on positiion, rotation, etc...). Hopfuly i can find out why, or if someone has any idea, help welcomed! 

image

image

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -48,6 +48,7 @@
 - Added preprocessors for shaders to improve how shaders are compiled for WebGL1/2 or WebGPU ([Deltakosh](https://github.com/deltakosh/))
 - Added enterPointerlock and exitPointerlock (Separated from enterFullscreen) ([aWeirdo](https://github.com/aWeirdo/))
 - Added support for `vertexSource` and `fragmentSource` parameters to `ShaderMaterial` ([Deltakosh](https://github.com/deltakosh/))
+- Added support for UV3-UV6 in the CustomMaterial ([Azukaar](https://github.com/azukaar/))
 
 ### Inspector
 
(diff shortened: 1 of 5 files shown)


deltakosh:
Ok waiting for your feedback regarding the index issue

azukaar:
> Ok waiting for your feedback regarding the index issue

Fixed the issue we can observe in the screenshot, and the StandardMaterial works perfectly now.

I am still experiencing issues with using the new version with CustomMaterial. I am assuming there's possibility with the issue being related to the fact that I cannot perform the Gulp build completely (it crashes before the end). Going to investigate that later.

azukaar:
@deltakosh So I made some progress into understanding the error. Essentially, since the change I made,  When a material is lacking a diffuseTexture / BumpTexture, instead of being white/flat, it randomly picks textures in other materials.

Considering how small the change I made is, I really dont understand where this is coming from. Any idea what could be causing such issue?

deltakosh:
If you cancel your changes you do not have this issue? Is it an issue with StdMaterial or Custommaterial?

azukaar:
> If you cancel your changes you do not have this issue? Is it an issue with StdMaterial or Custommaterial?

- yes 
- both

deltakosh:
Can you repro in the Playground? I feel like it is not related to your code ;)

deltakosh:
Will close for now. Please reopen if you find what was going wrong

PR #6978, opened 2019-10-14
Remove circular dependency

syntheticmagus:
Removed inadvertent circular dependency created by using outdated logging mechanism.


src/Cameras/XR/webXRSessionManager.ts
```diff
@@ -6,7 +6,7 @@ import { InternalTexture, InternalTextureSource } from "../../Materials/Textures
 import { RenderTargetTexture } from "../../Materials/Textures/renderTargetTexture";
 import { WebXRRenderTarget } from './webXRTypes';
 import { WebXRManagedOutputCanvas } from './webXRManagedOutputCanvas';
-import { WebXRState, Tools } from '../../Legacy/legacy';
+import { WebXRState } from '../../Legacy/legacy';
```
sebavan: Could you import it from the file containing it ??? cause legacy will break es6 imports here

PR #6985, opened 2019-10-15
Material Node texture blocks now check rootUrl to avoid loading embed…

julien-moreau:
…ed textures with a rootUrl.


julien-moreau:
Don't hesitate if you have a better solution! For me the condition is based on the node configuration. So i have put the test here

deltakosh:
Looks good to me

PR #6997, opened 2019-10-16
Edge beta local-floor workaround

syntheticmagus:
Workaround proposal for WebXR issue where Edge beta doesn't support local-floor yet.


syntheticmagus:
Turns out this is not a missing feature, but a nuance in the spec draft that we needed to change. PR to do that following shortly.

PR #7001, opened 2019-10-16
Implement 2D Array textures 

atg:
This implements 2D [array textures](https://www.khronos.org/opengl/wiki/Array_Texture), as discussed in this [thread](https://forum.babylonjs.com/t/support-2d-array-textures-gl-texture-2d-array/6363/3). 2D array textures are a WebGL 2 feature that allows you to make a single texture containing multiple 2D texture layers, i.e. a texture atlas.

2D array textures have almost the same WebGL interface as 3D textures, so with few exceptions I just duplicated the code for 3D textures and made the appropriate adjustments.

I am still working on making a playground to demo it, in the mean time any feedback is appreciated.


src/Engines/thinEngine.ts
```diff
@@ -415,6 +416,17 @@ export class ThinEngine {
         return this._emptyTexture3D;
     }
 
+    /**
+     * Gets the default empty 2D array texture
+     */
+    public get emptyTexture2DArray(): InternalTexture {
+        if (!this._emptyTexture2DArray) {
+            this._emptyTexture2DArray = this.createRawTexture2DArray(new Uint8Array(4), 1, 1, 1, Constants.TEXTUREFORMAT_RGBA, false, false, Constants.TEXTURE_NEAREST_SAMPLINGMODE);
```
sebavan: I am wondering how this works as the depth should probably related to the usage ? keeping the depth at 1 might break on the shader side, should it not just wait for readiness in case of array ?

src/Engines/Extensions/engine.rawTexture.ts
```diff
@@ -171,6 +171,42 @@ declare module "../../Engines/engine" {
          * @param textureType defines the texture Type (Engine.TEXTURETYPE_UNSIGNED_INT, Engine.TEXTURETYPE_FLOAT...)
          */
         updateRawTexture3D(texture: InternalTexture, data: Nullable<ArrayBufferView>, format: number, invertY: boolean, compression: Nullable<string>, textureType: number): void;
+
```
sebavan: I guess this part and the implementation could be moved to a dedicated engine.rawTextureArray file to help with tree shaking ?

src/Engines/thinEngine.ts
sebavan: Actually it might clamp in the shader to the actual depth so it might be ok but worth trying in WebGL :-)

atg: Yes it's OK, the depth coord is indeed clamped to [`max(0, min(d​-1, floor(coord​+0.5)) )`](https://www.khronos.org/opengl/wiki/Array_Texture#Access_in_shaders).

atg:
[According to the spec](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glGenerateMipmap.xhtml) and MDN gl.generateMipmap should just work.

> For two-dimensional texture arrays, each layer is filtered independently.

src/Engines/Extensions/engine.rawTexture.ts
atg: OK done

sebavan:
@atg Could you add a line in the dist/preview release/what s new.md file attributing the merit to you ?

Thanks a lot it is a nice evolution.

src/Engines/Extensions/engine.rawTexture2DArray.ts
```diff
@@ -0,0 +1,115 @@
...
+         * @param format defines the data format
+         * @param invertY defines if data must be stored with Y axis inverted
+         * @param compression defines the used compression (can be null)
+         * @param textureType defines the texture Type (Engine.TEXTURETYPE_UNSIGNED_INT, Engine.TEXTURETYPE_FLOAT...)
+         */
+        updateRawTexture2DArray(texture: InternalTexture, data: Nullable<ArrayBufferView>, format: number, invertY: boolean, compression: Nullable<string>, textureType: number): void;
+    }
+}
+
+Engine.prototype.createRawTexture2DArray = function(data: Nullable<ArrayBufferView>, width: number, height: number, depth: number, format: number, generateMipMaps: boolean, invertY: boolean, samplingMode: number, compression: Nullable<string> = null, textureType: number = Constants.TEXTURETYPE_UNSIGNED_INT): InternalTexture {
```
deltakosh: I think the code here can be merged or factorize with the one for 3D textures

src/Engines/thinEngine.ts
```diff
@@ -3361,8 +3400,23 @@ export class ThinEngine {
             }
 
             this._setAnisotropicLevel(this._gl.TEXTURE_3D, texture);
-        }
-        else if (internalTexture && internalTexture.isCube) {
+        } else if (internalTexture && internalTexture.is2DArray) {
```
deltakosh: same here, the code is almost the same so jsut detect the texture and use gl.TEXTURE-3D or TEXTURE_2D_ARRAY

atg:
Yes it's not quite ready to merge, I'm still in the process of testing it, then I'll do those refactorings and add the whatsnew line.

deltakosh:
Thanks a ton!

deltakosh:
You should credit you here:

and maybe add a note there for this new webgl2 feature that we support thanks to you ;)
https://github.com/BabylonJS/Documentation/blob/master/content/features/Support/WebGL2.md

atg:
Good to go I think. I have an update to the docs ready once it's merged as I'm planning to add a link to the playground at the same time.

PR #7004, opened 2019-10-17
Use the Monaco TS Language service for the Playground

sailro:
Creating a draft pull request for now, because it needs to be tested on other browsers (only tried with Chrome Version 77.0.3865.120 (Official Build) (64-bit)).

So the main idea is to remove the custom compilation step when using the typescript language in the Playground.

If I'm not wrong, currently the code is compiled twice: by the Monaco Editor, and by a custom invocation to `ts.createProgram` as well.

Given Monaco is giving access to the JS/TS language service, it seems easier to get the generated javascript directly from it without recompiling the same stuff for nothing.

### DIFF
--- a/Playground/js/main.js
+++ b/Playground/js/main.js
@@ -40,7 +40,7 @@ compileAndRun = function(parent, fpsLabel) {
         var createEngineFunction = "createDefaultEngine";
         var createSceneFunction;
 
-        parent.monacoCreator.getRunCode(function (code) {
+        parent.monacoCreator.getRunCode().then(function (code) {
             var createDefaultEngine = function () {
                 return new BABYLON.Engine(canvas, true, { preserveDrawingBuffer: true, stencil: true });
             }
--- a/Playground/js/main.js
+++ b/Playground/js/main.js
@@ -40,7 +40,7 @@ compileAndRun = function(parent, fpsLabel) {
         var createEngineFunction = "createDefaultEngine";
         var createSceneFunction;
 
-        parent.monacoCreator.getRunCode().then(function (code) {
+        parent.monacoCreator.getRunCode().then(code => {
             var createDefaultEngine = function () {
                 return new BABYLON.Engine(canvas, true, { preserveDrawingBuffer: true, stencil: true });
             }
@@ -172,7 +172,7 @@ compileAndRun = function(parent, fpsLabel) {
                     }
                 }
             }
-        }.bind(this));
+        });
 
     } catch (e) {
         parent.utils.showError(e.message, e);
--- a/Playground/js/monacoCreator.js
+++ b/Playground/js/monacoCreator.js
@@ -157,7 +157,7 @@ class MonacoCreator {
                 worker(uri)
                 .then(function(languageService) {
                      languageService.getEmitOutput(uri.toString())
-                     .then((result) => {
+                     .then(function(result) {
                           var output = result.outputFiles[0].text;
                           var stub = "var createScene = function() { return Playground.CreateScene(engine, engine.getRenderingCanvas()); }";
                           callBack(output + stub);
--- a/Playground/js/monacoCreator.js
+++ b/Playground/js/monacoCreator.js
@@ -148,20 +148,33 @@ class MonacoCreator {
      * @param {Function} callBack : Function that will be called after retrieving the code.
      */
     getRunCode(callBack) {
-        if (this.parent.settingsPG.ScriptLanguage == "JS")
+        var parent = this.parent;
+
+        if (parent.settingsPG.ScriptLanguage == "JS")
             callBack(this.jsEditor.getValue());
-        else if (this.parent.settingsPG.ScriptLanguage == "TS") {
+        else if (parent.settingsPG.ScriptLanguage == "TS") {
             var uri = this.jsEditor.getModel().uri;
             monaco.languages.typescript.getTypeScriptWorker()
             .then(function(worker) {
                 worker(uri)
                 .then(function(languageService) {
-                     languageService.getEmitOutput(uri.toString())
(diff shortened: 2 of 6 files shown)


sebavan:
Niceeeeee !!!

deltakosh:
Excellent. Help is very appreciated

sailro:
With the previous approach:
- The code was built twice.
- The error reported by the editor was not in sync with the error reported by the custom compilation mechanism.
- The [line/column] position was wrong.
- Only mainly semantic errors were reported. Several syntactic errors like removing the last '}' in the script were not properly reported.

**[master]**
current

Reusing the TS language service allows to fix all of this:

**[this PR]**
new

**[this PR]**
image

deltakosh:
Love it! This is definitely a place where we did not put too much energy until now :)

sailro:
Firefox 69.0.3 (64-bit) | Chrome 77.0.3865.120 (64-bit) | Edge 44.18362.387.0 (EdgeHTML 18.18362)
:-------------| :-------------| :-------------
:heavy_check_mark: | :heavy_check_mark: | :red_circle:

Any idea ? master branch is working everywhere, so my change is triggering an issue with Edge only.

```
SCRIPT5022: SCRIPT5022: Out of stack space

Error: Out of stack space
   at P (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/language/typescript/lib/typescriptServices.js:56:14123)
   at k (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/language/typescript/lib/typescriptServices.js:56:12534)
   at O (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/language/typescript/lib/typescriptServices.js:65:27566)
   at e.prototype.getEmitOutput (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/language/typescript/src/worker.js:9:3869)
   at e.prototype.fmr (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/base/worker/workerMain.js#typescript:7:149450)
   at t.prototype._handleMessage (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/base/worker/workerMain.js#typescript:7:110446)
   at handleMessage (http://localhost:1338/Playground/node_modules/monaco-editor/min/vs/base/worker/workerMain.js#typescript:7:110099)
   at e.prototype._h
```

Playground/js/main.js
```diff
@@ -40,7 +40,7 @@ compileAndRun = function(parent, fpsLabel) {
         var createEngineFunction = "createDefaultEngine";
         var createSceneFunction;
 
-        parent.monacoCreator.getRunCode(function (code) {
+        parent.monacoCreator.getRunCode().then(function (code) {
```
deltakosh: I recommend using () => instead of function()

sailro: Fully agree. Given I'm a total newbie in webdev, from my understanding with Arrow functions you don't have to create "this closures" or even use bind(this) like it was done in those functions right ?

deltakosh: lol if all newbies were like you :D

yes it capture the this for you and stay in the "logical" context. No more that= this!!

sailro:
So it is now working properly in latest Chrome, Edgium, Firefox.

But I still see a strange behavior with legacy Edge. No more error in console but the loading seems stuck. Often switching back and forth to the debug window, and hitting Play will solve the issue...

Requesting advice from web experts @deltakosh @sebavan  :)

┻━┻︵ \(°□°)/ ︵ ┻━┻

Playground/js/monacoCreator.js
```diff
@@ -133,75 +145,38 @@ class MonacoCreator {
...
+
+        else if (parent.settingsPG.ScriptLanguage == "TS") {
+            const model = this.jsEditor.getModel();
+            const uri = model.uri;
+
+            const worker = await monaco.languages.typescript.getTypeScriptWorker();
+            const languageService = await worker(uri);
+
+            const uriStr = uri.toString();
+            const result = await languageService.getEmitOutput(uriStr);
```
deltakosh: Ok this line is definitely killing Monaco. The out of stack space exception comes directly from monaco code :(

Can you try updating Monaco to very latest version to see if the problem is fixed?

deltakosh:
OK just tried with latest version and it works....but....

deltakosh:
With the latest version the intelissense in JavaScript is returning the private properties and functions.

Ideally I would love to find a way to filter the value that starts with _

deltakosh:
This is what we need to avoid:
image
 (and why we are stuck on a old version of monaco)

sailro:
Then it is the perfect occasion to have a look and upgrade.

I'm going to investigate

deltakosh:
Thanks a ton sir!

sailro:
Seems we can filter internals/private members without adding our own completionProvider. I just set compiler options for javascriptDefaults. The built-in completionProvider is probably using those settings to filter babylon.d.ts entries.

Screenshots with Edge legacy for both javascript/typescript

image

image

what's new.md
```diff
@@ -64,6 +64,7 @@
 - All NPM packages have `latest` and `preview` streams ([RaananW](https://github.com/RaananW))
 - Added New Tools Tab in the inspector (env texture and screenshot tools so far) ([sebavan](http://www.github.com/sebavan))
 - Moved to gulp 4, updated dependencies to latest ([RaananW](https://github.com/RaananW))
+- Bumped Monaco Editor to 0.18.1 and improved TypeScript compilation pipeline in the Playground ([sailro](http://www.github.com/sailro))
```
deltakosh: This is not the right file (this one is for stable). This is the one to edit:

deltakosh:
Thanks mate!

PR #7024, opened 2019-10-18
PointColor.Color More Flexible

BabylonJSGuide:
Can now specify which texture from a material is used to set vertices colors.


Tools/Gulp/package.json
```diff
@@ -12,5 +12,8 @@
         "install": "cd ../../ && npm install && cd Playground/ && npm install && cd ../Viewer && npm install && cd ../Tools/Gulp/",
         "build": "gulp --max-old-space-size=8192 --tsLintFix",
         "start": "gulp run --max-old-space-size=8192"
+    },
+    "dependencies": {
+        "gulp": "^4.0.2"
```
deltakosh: This should not be modified

BabylonJSGuide: package.json restored

PR #7030, opened 2019-10-21
fix SPS insertParticlesFromArray : now the positionFunction can be run

jbousquie:
When restoring formerly saved particles (in an external storage or previously removed ones), if a positionFunction is associated to thoses particles, it can be executed back on the insertion in the SPS

+
fix on particle.copyToRef() in case of missing color in the target


jbousquie:
The whatsnew file has not been updated in purpose : it's just a tiny fix on the very last feature that is not yet documented.

jbousquie:
ok, just added 2 space caracters in the whatsnew file, so it's commited and all the tests should pass now

PR #7040, opened 2019-10-22
Cloning of ShaderMaterial should also clone _shaderPath and _options properties

Popov72:
As approved by @Deltakosh in the forum (https://forum.babylonjs.com/t/not-all-properties-cloned-when-cloning-shadermaterial/6442/7), I have updated the `ShaderMaterial.Clone` method to also clone the `_shaderPath` and `_options` properties.

_Implementation note_:

I had to cast to `<string[]>` here:
```typescript
    if (Array.isArray(propValue)) {
        (<string[]>this._options[propName]) = propValue.slice(0);
    }
```
to keep Typescript happy, but it would be better without this cast (more future foolproof if new non-string array properties are added to `_options`). 

However, I could not come up with a solution to this problem...

### DIFF
--- a/src/Materials/shaderMaterial.ts
+++ b/src/Materials/shaderMaterial.ts
@@ -775,6 +775,21 @@ export class ShaderMaterial extends Material {
         result.name = name;
         result.id = name;
 
+        // Shader code path
+        if (typeof result._shaderPath === 'object') {
+            result._shaderPath = { ...result._shaderPath };
+        }
+
+        // Options
+        this._options = { ...this._options };
+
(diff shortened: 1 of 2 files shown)


src/Materials/shaderMaterial.ts
```diff
@@ -775,6 +775,21 @@ export class ShaderMaterial extends Material {
         result.name = name;
         result.id = name;
 
+        // Shader code path
+        if (typeof result._shaderPath === 'object') {
+            result._shaderPath = { ...result._shaderPath };
+        }
+
+        // Options
+        this._options = { ...this._options };
```
deltakosh: This line looks suspicious to me :) Can you elaborate on your intend?

PR #7047, opened 2019-10-23
JS part of setviewport for BabylonNative

CedricGuillemet:


src/Engines/nativeEngine.ts
```diff
@@ -279,6 +281,9 @@ export class NativeEngine extends Engine {
     }
 
     public clear(color: Color4, backBuffer: boolean, depth: boolean, stencil: boolean = false): void {
+        if (color == null) {
```
bghgary: how can color be null?

src/Engines/nativeEngine.ts
```diff
@@ -489,6 +494,7 @@ export class NativeEngine extends Engine {
     public setViewport(viewport: Viewport, requiredWidth?: number, requiredHeight?: number): void {
         // TODO: Implement.
```
bghgary: Remove this?

src/Engines/nativeEngine.ts
```diff
@@ -489,6 +494,7 @@ export class NativeEngine extends Engine {
     public setViewport(viewport: Viewport, requiredWidth?: number, requiredHeight?: number): void {
         // TODO: Implement.
         this._cachedViewport = viewport;
```
bghgary: Should we check if the viewport changed? What's the cached value for?

src/Engines/nativeEngine.ts
CedricGuillemet: https://github.com/BabylonJS/Babylon.js/blob/master/src/scene.ts#L3916

bghgary: We should maybe do a pass on the method signatures. The main engine has a color signature like this now: `color: Nullable<IColor4Like>`. Doing this will cause TypeScript to give you an error if you don't check for `null`.

src/Engines/nativeEngine.ts
bghgary: Looks like the this._cachedViewPoint is used to get the viewport. So it's fine.

https://github.com/BabylonJS/Babylon.js/blob/master/src/Engines/thinEngine.ts#L394

src/Engines/nativeEngine.ts
bghgary: Also, this check isn't correct. If color is null, you still have to set the backBuffer, depth, and stencil.

CedricGuillemet: If we want the same behavior as BJS, yes, we have to.

PR #7054, opened 2019-10-25
Check invalid handle native

CedricGuillemet:
To be used with https://github.com/BabylonJS/BabylonNative/pull/72


src/Engines/nativeEngine.ts
```diff
@@ -71,8 +71,8 @@ interface INativeEngine {
     setFloat4(uniform: WebGLUniformLocation, x: number, y: number, z: number, w: number): void;
 
     createTexture(): WebGLTexture;
-    loadTexture(texture: WebGLTexture, buffer: ArrayBuffer | ArrayBufferView | Blob, mipMap: boolean, invertY: boolean): void;
-    loadCubeTexture(texture: WebGLTexture, data: Array<Array<ArrayBufferView>>, flipY : boolean): void;
+    loadTexture(texture: WebGLTexture, buffer: ArrayBuffer | ArrayBufferView | Blob, mipMap: boolean, invertY: boolean): any;
```
bghgary: Maybe return a book instead for these?

CedricGuillemet: What's a book?

bghgary: Aye, typing on my phone. Bool :)

CedricGuillemet: I'd like to be consistent with createIndexBuffer/createVertexBuffer

bghgary: I don't think that makes sense. The create functions create things and thus return the thing it created. The load functions load things into the first parameter :) A boolean return for load to indicate success makes the most sense to me.

PR #7056, opened 2019-10-25
clear flags and values for native

CedricGuillemet:


src/Engines/nativeEngine.ts
```diff
@@ -89,7 +90,10 @@ interface INativeEngine {
     drawIndexed(fillMode: number, indexStart: number, indexCount: number): void;
     draw(fillMode: number, vertexStart: number, vertexCount: number): void;
 
-    clear(r: number, g: number, b: number, a: number, backBuffer: boolean, depth: boolean, stencil: boolean): void;
+    clear(flags: number): void;
+    clearColor(r: number, g: number, b: number, a: number): void;
```
bghgary: Are these used?

CedricGuillemet: yes in this PR

bghgary: Ah, my bad. missed them

Issue #6665, opened 2019-08-05
Multiview does not support built-in line shaders


delaneyj:
Per [discussion on the forum](https://forum.babylonjs.com/t/vr-quest-multiview-missing-content/4920/2), MultiView does not support all built-in shaders.  As a side note; it would be ideal if the needed changes to custom shaders to support Multiview was [part of the documentation on the subject](https://doc.babylonjs.com/how_to/multiview).

# Bugs
- Bug repro on [playground](https://playground.babylonjs.com/#365m82):

- Expected result:
Visible lines in Multiview mode

- Current result:
At least of the Oculus quest it is not visible when VR mode is entered.

deltakosh:
@RaananW: Interested by this one?

PR #7059, opened 2019-10-25
Multiview shaders

RaananW:
Fixing https://github.com/BabylonJS/Babylon.js/issues/6665

Shader Material now supports multiview. To get it to work the viewProjection needs to be enabled as a uniform, and the shader needs to be written in the right way. see color.vertex.fx for a quick overview

Tested with Chrome Canary


src/Materials/shaderMaterial.ts
```diff
@@ -446,6 +447,19 @@ export class ShaderMaterial extends Material {
...
+        // global multiview
+        if (engine.getCaps().multiview &&
+            scene.activeCamera &&
+            scene.activeCamera.outputRenderTarget &&
+            scene.activeCamera.outputRenderTarget.getViewCount() > 1) {
+            this._multiview = true;
+            defines.push("#define MULTIVIEW");
+            if (this._options.uniforms.indexOf("viewProjection") !== -1 &&
+                this._options.uniforms.push("viewProjectionR") === -1) {
+                this._options.uniforms.push("viewProjectionR");
```
Popov72, line 498: I think it should be `indexOf` instead of `push` in the `if` part for `viewProjectionR`?

PR #7064, opened 2019-10-28
Add depthFunction property to the Material base class

Popov72:
See here:

https://forum.babylonjs.com/t/add-a-depthfunction-property-to-materials/6608

Allow to change the depth function per material.


Popov72:
Going to do that!

PR #7077, opened 2019-10-30
NME Trig Block Changes

Pryme8:
How would I ignore the .gitignore commit or is there a way to discard it?

### DIFF
--- a/src/Materials/Node/Blocks/trigonometryBlock.ts
+++ b/src/Materials/Node/Blocks/trigonometryBlock.ts
@@ -65,8 +65,8 @@ export class TrigonometryBlock extends NodeMaterialBlock {
     public constructor(name: string) {
         super(name, NodeMaterialBlockTargets.Neutral);
 
-        this.registerInput("input", NodeMaterialBlockConnectionPointTypes.Float);
-        this.registerOutput("output", NodeMaterialBlockConnectionPointTypes.Float);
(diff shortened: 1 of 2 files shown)


deltakosh:
You must discard the .gitignore change :)

src/Materials/Node/Blocks/trigonometryBlock.ts
```diff
@@ -65,8 +65,11 @@ export class TrigonometryBlock extends NodeMaterialBlock {
     public constructor(name: string) {
         super(name, NodeMaterialBlockTargets.Neutral);
 
-        this.registerInput("input", NodeMaterialBlockConnectionPointTypes.Float);
-        this.registerOutput("output", NodeMaterialBlockConnectionPointTypes.Float);
+        this.registerInput("input", NodeMaterialBlockConnectionPointTypes.AutoDetect);
+        this.registerOutput("output", NodeMaterialBlockConnectionPointTypes.BasedOnInput);
+
+        this._outputs[0]._typeConnectionSource = this._inputs[0];
+        this._linkConnectionTypes(0, 1);
```
deltakosh: This is not required (Only used if you have two inputs that must have the same type)

Pryme8: Ok so should I just discard the commit that has that part?

PR #7080, opened 2019-10-31
setInt used internaly for setBool

CedricGuillemet:


CedricGuillemet:
> Actually, I'm confused. The `setBool` function was removed from the native side. Should the same happen here?
setbool from BJS side calls setint on the engine.

bghgary:
> setbool from BJS side calls setint on the engine. So, not needed on nativenengine JS or native engine

Right, so we need to remove it from nativeEngine.ts right?

https://github.com/BabylonJS/Babylon.js/blob/5b0ba794647dd8f46efabf4b214c7f68ef5067ca/src/Engines/nativeEngine.ts#L782-L788

PR #7100, opened 2019-11-04
Update engine.views.ts

Pryme8:
fix unRegisterView


src/Engines/Extensions/engine.views.ts
```diff
@@ -81,7 +81,7 @@ Engine.prototype.unRegisterView = function(canvas: HTMLCanvasElement): Engine {
     }
 
     for (var view of this.views) {
-        if (view.target = canvas) {
+        if (view.target === canvas) {
```
deltakosh: woot! good catch!

PR #7103, opened 2019-11-05
Use correct non-uniform normal transform

haxiomic:
This is best explained by @graphitemaster who raised awareness of the bug in the graphics community, his explanation and derivation:
https://github.com/graphitemaster/normals_revisited

More detailed derivation and discussion in this document (from 1987!)
https://www.glassner.com/wp-content/uploads/2014/04/On-the-Transformation-of-Surface-Normals.pdf

The traditional method for transforming normals with `transpose(inverse(m))` is incorrect since it doesn't fully preserve normal orientation. Here's a demo (change `#define method` to see incorrect example)
https://www.shadertoy.com/view/3s33zj

Fortunately the correct method, which uses the [`adjugate`](https://en.wikipedia.org/wiki/Adjugate_matrix) matrix is cheaper to calculate

In this PR I've replaced instances of `transpose(inverse(m))` with `adjugate` and added an implementation of `adjugate` to helperFunctions.fx

**Don't merge yet, want to see test results**

### DIFF
--- a/src/Shaders/ShadersInclude/helperFunctions.fx
+++ b/src/Shaders/ShadersInclude/helperFunctions.fx
@@ -43,6 +43,20 @@ mat3 inverseMat3(mat3 inMatrix) {
               b21, (-a21 * a00 + a01 * a20), (a11 * a00 - a01 * a10)) / det;
 }
 
+/**
+    Given a transform matrix, the adjugate matrix gives us the equivalent transform matrix for transforming surface normals
+    The cofactor matrix is equivalent to transpose(adjugate(m))
+
+    See https://github.com/graphitemaster/normals_revisited
+**/
+mat3 adjugate(in mat3 m) {
+  return mat3(
+      cross(m[1], m[2]), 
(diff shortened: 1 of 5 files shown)


sebavan:
Lovin it !!! @haxiomic you rock

deltakosh:
Ok to merge?

haxiomic:
After some testing I think there might be more to change to prevent changing the existing behavior

If you set the x-scale to -3, then the normals _should_ be flipped in principle, but often cases we want to un-flip them so we still get nice surface reflections. The `transpose(inverse(m))` doesn't flip the normals here but in this case it works out to what we usually want anyway (however in other cases we _do_ want the normals to flip for negative scales)

To avoid changing the current behaviour we need to flip the normals for negative scales by default (and maybe add a flag to enable flipping w.r.t scale)

I need to do some thinking on that, not got a lot of free bandwidth at the moment so I'll close and revisit this in the future

deltakosh:
Ok please keep us posted!

PR #7116, opened 2019-11-06
Path3D (bi)normal computation edge case fix

Poolminer:
Regarding: the computation of normals and binormals of Path3D

The computation:
For each segment, the tangent (forward axis) is taken and crossed with the previous binormal (up axis) to get a normal.

The problem:
If the previous binormal is along the same axis as the tangent, the normal will be zero, as will the derived binormal.

The fix:
Use the normal of the previous segment


deltakosh:
Pinging @jbousquie / @BabylonJSGuide  for review

PR #7119, opened 2019-11-07
sensibility factor for axisScaleGizmo

CedricGuillemet:


bigtimebuddy:
Wouldn’t a better name for this be called sensitivity instead of sensibility? they are basic the synonyms but slight preference for sensitivity.

Thanks for the PR will be super helpful.

PR #7138, opened 2019-11-11
New VRExperienceHelper Teleportation options

LeoRodz:
1) Add option to change the teleportation duration in the VRExperienceHelper class.
The property teleportationTime was added to the vrTeleportationOptions interface.
2) Add support to teleport the camera at constant speed in the VRExperienceHelper class.
The property teleportationSpeed was added to the vrTeleportationOptions interface.
The property vrTeleportationOptions.teleportationMode is used to change the teleportation mode between constant time (default) and constant speed.
Two public static properties were added to the VRExperienceHelper class:
BABYLON.VRExperienceHelper.TELEPORTATIONMODE_CONSTANTTIME and
BABYLON.VRExperienceHelper.TELEPORTATIONMODE_CONSTANTSPEED

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -150,6 +150,9 @@
 - Preparing to deprecate supportsSession in favor of isSupportedSession ([RaananW](https://github.com/RaananW/))
 - Added onControllerModelLoaded observable for WebXR ([RaananW](https://github.com/RaananW/))
 - UI Button has options to set different session mode and reference type ([RaananW](https://github.com/RaananW/))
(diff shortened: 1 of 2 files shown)


LeoRodz:
> 
> 
> This is really good
> Just one question: does it behave the same as before by default?

Yes, it does.
The default teleportation mode is TELEPORTATIONMODE_CONSTANTTIME as before.
The default teleportationTime is 122ms, calculated as round(11/90*1000), where 90 is the FPS and 11 is the end frame used in the original teleportation animations.

deltakosh:
Congratulations for making a perfect PR!

PR #7142, opened 2019-11-11
Updating deterministic lockstep to work with 144Hz monitors

sacharobarts:
Fixed bug where frame-rates higher than 60 fps would cause physics simulations to run at increased speeds.
Added the ability to set the time step between frames when passed with the EngineOptions interface.


deltakosh:
Thanks!

deltakosh:
You may want to credits yourself here:

PR #7148, opened 2019-11-12
Fix possible crash when an effect is not ready

barroij:
In the `Mesh.render` function, we test if the effective material is ready only on the first call. If it us called twice, and the material is still not ready, its ready-ness is not checked and there is an exception thereafter.

This PR fixes that bug by making sure that a non-ready material keeps beeing tested for ready-ness untill it is ready and then promoted as the "effective material"


barroij:
@deltakosh PR here

I tested in local in my application and it works. Don't if this change can cause any other side effect though.

PR #7157, opened 2019-11-13
Implement support for KHR_mesh_quantization

zeux:
This change advertises support for ~KHR_quantized_geometry~ KHR_mesh_quantization
and allows to specify it as part of required extensions when loading glTF files.

The reason why the actual extension class is empty is that the extension
doesn't define any new JSON metadata, and merely allows the use of
various quantized attributes in glTF file - and as far as I can tell
Babylon.js already supports these quantized formats either by directly
using the encoded data, or by decoding it into floating-point arrays.

Extension spec: https://github.com/KhronosGroup/glTF/pull/1673
Validator support: https://github.com/KhronosGroup/glTF-Validator/pull/124
Example model: [BrainStem.glb.zip](https://github.com/BabylonJS/Babylon.js/files/3844599/BrainStem.glb.zip)

Note that based on prior testing, Babylon.js de-facto supports this. This change is required because the extension is non-optional and Babylon.js refuses to load files with required extensions that aren't explicitly supported.


loaders/src/glTF/2.0/Extensions/KHR_quantized_geometry.ts
```diff
@@ -0,0 +1,34 @@
...
+     * The name of this extension.
+     */
+    public readonly name = NAME;
+
+    /**
+     * Defines whether this extension is enabled.
+     */
+    public enabled: boolean;
+
+    private _loader: GLTFLoader;
```
bghgary: There is no need to save the loader since it's only used in the constructor.

loaders/src/glTF/2.0/Extensions/KHR_quantized_geometry.ts
```diff
@@ -0,0 +1,34 @@
...
+    /** @hidden */
+    constructor(loader: GLTFLoader) {
+        this._loader = loader;
+        this.enabled = this._loader.isExtensionUsed(NAME);
+    }
+
+    /** @hidden */
+    public dispose() {
+        delete this._loader;
+    }
```
bghgary: Can be removed once the private _loader field is removed.

loaders/src/glTF/2.0/Extensions/KHR_quantized_geometry.ts
```diff
@@ -0,0 +1,34 @@
+import { IGLTFLoaderExtension } from "../glTFLoaderExtension";
+import { GLTFLoader } from "../glTFLoader";
+
+const NAME = "KHR_quantized_geometry";
+
+/**
+ * [Specification](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_quantized_geometry)
```
bghgary: Can't verify since the spec hasn't been merged.

loaders/src/glTF/2.0/Extensions/KHR_quantized_geometry.ts
zeux: I've tried to remove this actually and I can't - the TypeScript compiler (iirc) requires that there's an implementation of dispose. It can be empty though, I'll fix this.

bghgary: Ahh, yes, the interface function is required.

Issue #7158, opened 2019-11-13
Camera isn't in same position after entering XR


nbduke:
# Bugs
Repro playground: https://www.babylonjs-playground.com/#THMWLW#2

If you enter XR mode, you'll notice that the `WebXRCamera` is not where the previous (non-XR) camera had been, which feels really jarring. You can see their respective positions printed to the console.

Interestingly, this didn't repro in WebVR - flip the `useXR` variable to false and try again in a different browser. The `WebVRCamera` is where you'd expect.

deltakosh:
adding @RaananW

nbduke:
@RaananW I'm not convinced this is fixed (or maybe there's been a regression). In the Playground example, you still see two different positions printed to the console before and after entering XR.

RaananW:
I will look into that and let you know

RaananW:
Implemented with https://github.com/BabylonJS/Babylon.js/issues/7239 , documentation is coming soon

PR #7164, opened 2019-11-14
Camera position when entering XR

RaananW:
Make sure position is copied from the main scene camera when entering XR

Fixing https://github.com/BabylonJS/Babylon.js/issues/7158


RaananW:
The Y is not being taken from the old camera. Only X and Z are set. Y is taken from the first XR Frame

PR #7169, opened 2019-11-15
Hook monaco completion provider to filter members starting with underscores

sailro:
_Given a non filtered babylon.d.ts_

Before:
before

After:
after

### DIFF
--- a/Playground/babylon.d.txt
+++ b/Playground/babylon.d.txt
@@ -107,7 +107,8 @@ declare module BABYLON {
          * Defines the current scope used to restore the JS context
          */
         scope: any;
-        /** @hidden */protected _willBeUnregistered: boolean;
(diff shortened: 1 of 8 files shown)


sailro:
@deltakosh, after merging you need to update the babylon.d.ts with the regular one, and remove any auto-post-processing-step if needed.

deltakosh:
yep yep it is done:)

sebavan:
So thankful to @sailro as it helps removing some build steps :-)

sailro:
Do we still need this ?

https://github.com/BabylonJS/Babylon.js/blob/master/Tools/Gulp/tasks/gulpTasks-intellisense.js#L15

deltakosh:
Nope, it is removed in my branch

PR #7178, opened 2019-11-17
expose BasisTools class for configuring JSModuleURL and WasmModuleURL

jasonayre:
Expose BasisTools so that JSModuleURL and WasmModuleURL can be configured, so local apps relying heavily on basis textures i.e. mine, don't break when server goes down/connection is lost.

Also: I don't need to commit ANY of the changed files from the preview release folder, correct? Docs listed specific files to ignore but doing a build modified almost everything in the preview release folder.


deltakosh:
Hello I feel like this PR is incomplete as I see no real changes

jasonayre:
@deltakosh It's a simple change, `export * from "./basis";` -- There is no way to configure JSModuleURL or WasmModuleURL at the moment because BABYLON.Basis is not exposed, this exposes it

deltakosh:
Ok understood :)

PR #7183, opened 2019-11-19
ScrollViewer with ImageScrollBar

BabylonJSGuide:
Help needed as in https://forum.babylonjs.com/t/customize-scrollbar/6956/18?u=johnk

Thank you for saying you will help.


deltakosh:
So what problems do you face so far?

BabylonJSGuide:
Gijs came up with a solution for image scroll bars in https://www.babylonjs-playground.com/#13CF95#66

It has something I liked, the thinness of the images within the bars, plus a couple of issues 

1. It calls Image.drawImage directly
2. It only works for a vertical bar.

When attempting to modify the ScrollViewer code I had the following problems 

1. getting a thin image
2. working with one image in both a vertical and horizontal orientation.

Here is a working example of what I have so far  [Scroller](https://tender-murdock-275855.netlify.com/view/test1.html) 

You will notice that the thumb fills the height of the background unlike the Gijs example. Also the background image is more stretched across the height then the Gijs example.

What I have done it so far

i. Created a new ImageScrollBar class 

ii. Added an extra parameter to ScrollViewer which when true allows images for the scroll bars see Line 97

iii. Forced the thumb to be in proportion to the image used and then extend the width using the thumbSize variable lines 378 / 379 in ScrollViewer

iv. Set the measures for the bar background and bar (thumb) in lines 92 - 117 in the ImageScrollBar which then calls Image._draw

v. The Image._draw call goes to Image.ts at line 615 where the drawImage parameters are set and called depending on a switch state line 648. The ScrollViewer for both ScrollBar and ImageScrollBar go through Line 653 STRETCH_FILL and the private function _drawImage is called Line 598

I experimented with creating different switch calls named IMAGE_VERTICAL, IMAGE_HORIZONTAL but couldn’t work out the parameter values to use to get the same effect as Gijs, so deleted this attempt. 

I tried using transform and rotate on the canvas image but got completely lost and could not work out the values to use.

Now Gijs has come up with some further code though this again calls drawImage directly rather than the private _draw but perhaps this is sufficient. His use of rotate only partially solves the vertical horizontal bar issue as it does not take into account whether width > height or vice versa.

I will keep experimenting and would be very grateful for any advise you could give.

BabylonJSGuide:
Am close to a solution of sizing the images for the thumb and the bar. No where near for rotating on vertical or horizontal.

deltakosh:
Here are some thoughts:
- I'm not sure we should have a background image. Or it should be optional.
- Your scroller example looks really good to me.
- The thumbImage width should be set to bar width when vertical and height should be set to bar height when horizontal
- _getThumbThickness should return bar.width * image aspect ratio when vertical (and same thing for height when vertical)

BabylonJSGuide:
Am making progress.  

So far I have that the background image is optional, along the horizontal or vertical it fills the space available it height across the space can be varied with barImageHeight, range 0 to 1.

There are also two new properties, thumbLength and thumbHeight (range 0 to 1) that controls the controls the size of the thumb.

A little more checking and I should be able to PR this stage and give examples.

What I really would like to do is if your image for the thumb (or background) is say 100 x 200 then it fits best in the vertical bar space and I would like to rotate the image to put it into the horizontal bar space and vice versa for an image 200 x 100. This I am have major problems with.

deltakosh:
Ok for that you can call something like that before rendering the image:
https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/2D/controls/control.ts#L1283

It does a central rotation (do not forget to call context.save() before and context.restore() after drawing the image)

BabylonJSGuide:
Have this new problem on npm run build following the last pull

[17:34:04] Finished 'typescript-libraries-importLint' after 8.02 s
[17:34:04] Finished 'importLint' after 8.03 s
[17:34:04] Starting 'circularDependencies'...
[17:34:04] Starting 'core-circularDependencies'...
[17:35:12] 'core-circularDependencies' errored after 1.12 min
[17:35:12] TypeError: Cannot read property 'summary' of undefined
    at dependencies (C:\Users\User\Documents\GitHub\Babylon.js\Tools\Gulp\tasks\gulpTasks-dependencies.js:40:29)
    at C:\Users\User\Documents\GitHub\Babylon.js\Tools\Gulp\tasks\gulpTasks-dependencies.js:64:9
    at taskWrapper (C:\Users\User\Documents\GitHub\Babylon.js\Tools\Gulp\node_modules\undertaker\lib\set-task.js:13:15)
    at bound (domain.js:395:14)
    at runBound (domain.js:408:12)
    at asyncRunner (C:\Users\User\Documents\GitHub\Babylon.js\Tools\Gulp\node_modules\async-done\index.js:55:18)
    at process._tickCallback (internal/process/next_tick.js:61:11)
[17:35:12] 'circularDependencies' errored after 1.12 min
[17:35:12] 'default' errored after 1.58 min
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! babylonjs@0.0.0 build: `gulp --max-old-space-size=8192 --tsLintFix`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the babylonjs@0.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\User\AppData\Roaming\npm-cache\_logs\2019-11-19T17_35_12_665Z-debug.log

Packing up for now will check back later.

deltakosh:
Run npm install in tools/gulp folder

BabylonJSGuide:
If you are happy with what I have done so far then I am happy to have this merged. I will work on the rotation issue and do a new PR  later, when I have it working.

Working example (now showing properties lines 90 to 95) https://tender-murdock-275855.netlify.com/view/test1.html

PR #7196, opened 2019-11-22
AmmoJS - Custom Impostor Shapes

MackeyK24:
AmmoJS - Update to allow the use of custom impostor shapes within the native babylon.js physics plugin system.


MackeyK24:
Hey @deltakosh .

I added this because they are times when you will want to get down and dirty with the physics system and need to create a custom physics impostor shape and still use the physics entity in the native babylon physics ecosystem (As apposed to creating your custom physics entity **OUTSIDE** the babylon environment)

In my case i have create a native C++ sub class of **btTriangleMesh** called **btSmoothTriangleMesh** that needs to get instantiated for mesh impostors. By allowing us to create our own **onCreateCustomShape** factory function on the AmmoJSPlugin, we can at least be able to use our custom shapes for native babylon physics impostors.

So please lets add if no objections :)

PR #7199, opened 2019-11-23
Support for deprecated members in the Playground

sailro:
Set markers for deprecated members in the Editor:
markers

Tag completion entries as deprecated in the auto-complete popup:
completion


sailro:
@deltakosh please note that you cannot test correctly this PR with the preview deployment because of this:
https://github.com/BabylonJS/Babylon.js/blob/c9ec56d16f48029cf940a82f9e8fb273b252b896/Playground/js/monacoCreator.js#L56

I.e. for now the playground is always using the published definition and not the one related to the PR.

PR #7214, opened 2019-11-26
ScrollViewer Update Images in Vertical Rotated

BabylonJSGuide:
Have solved the rotation issue for vertical image scroll bar. 

Example https://tender-murdock-275855.netlify.com/view/test4.html


gui/src/2D/controls/sliders/imageScrollBar.ts
```diff
@@ -8,53 +8,97 @@ import { Measure } from "../../measure";
...
 
-        this._thumbImage = value;
+        this._thumbBaseImage = value;
 
-        if (value && !value.isLoaded) {
-            value.onImageLoadedObservable.addOnce(() => this._markAsDirty());
+        if (this.isVertical) {
+            if (value && !value.isLoaded) {
+                value.onImageLoadedObservable.addOnce(() => {
+                    var rotatedValue = value._rotate90(-1);
```
deltakosh: I'm concerned about the memory cost here. Won't it be better to render it rotated? Or I bet your did that for perf reason? Which is totally fine as well

gui/src/2D/controls/image.ts
```diff
@@ -260,6 +260,28 @@ export class Image extends Control {
...
+
+        const context = canvas.getContext('2d')!;
+        const width = this._domImage.width;
+        const height = this._domImage.height;
+
+        canvas.width = height;
+        canvas.height = width;
+
+        context.translate(canvas.width / 2, canvas.height / 2);
+        context.rotate(n * Math.PI / 2);
```
deltakosh: if n == 2 then width and height should not be switched...It works only if n is odd

gui/src/2D/controls/sliders/imageScrollBar.ts
BabylonJSGuide: To be honest I could not work out the values to pass to rotate the image inside the already rotated vertical scroll bar. After 3 days of failing to work out the values I decided to create an image that was a rotation of the horizontal one. I knew there was an increased memory cost but went for it anyway. My rational was generally such images would be small. However I do agree just rotating the given value into place would be more pleasing and I would be pleased if you or anyone else could show the values to pass.

gui/src/2D/controls/image.ts
BabylonJSGuide: Only values used are -1 and 1. The function is only intended to be used internally.

deltakosh: Ok makes sense

Issue #7219, opened 2019-11-27
Taking a screenshot causes shadows of instanced meshes to disappear


nbduke:
# Bugs
Repro playground: https://www.babylonjs-playground.com/#IFYDRS#48

In the above playground, I create a box mesh and an instance of that mesh. The instance is added as a shadow caster. A second after scene load, I create a new camera and take a screenshot of the scene using `Tools.CreateScreenshotUsingRenderTargetAsync`. You see a brief blink of the scene and after that, the shadow of the instanced box is gone.

sebavan:
Min repro: https://www.babylonjs-playground.com/#IFYDRS#50

This is not only linked to screenshot but looks like it could impact all RTT. Narrowed down to a caching issue. I am still looking into a fix.

PR #7221, opened 2019-11-27
Set the scene render mode to intermediate before texture.render

ladevieq:
When i tried to do screenshot on scenes containing instances, the first screenshot worked well, everything was displayed, as you can see below.

Test_rals-Version-191127154654667

But when i tried to do screenshot a second time some meshes disappeared, all of them are instances.

Test_rals-Version-191127154125626

When doing a screenshot using render target for the first time, the scene
intermediate rendering is set to false at the end of the
texture.render().

Next time the function is called it makes all instances being avoided
during the render.

This PR is essentially a hotfix and should be edited in order to fix the bug in a correct way.

### DIFF
--- a/src/Misc/screenshotTools.ts
+++ b/src/Misc/screenshotTools.ts
@@ -151,6 +151,7 @@ export class ScreenshotTools {
         const renderToTexture = () => {
             scene.incrementRenderId();
             scene.resetCachedMaterial();
+            scene._intermediateRendering = true;
             texture.render(true);
             texture.dispose();
 


ladevieq:
It looks like the bug is fixed with 4.1.0 beta4

deltakosh:
Ok thanks! closing then ;)

ladevieq:
I did some more testing and just realized it's not fixed in beta-4 could you reopen it @deltakosh ?

sebavan:
It should be from the latest nightly and so in the next npm version. We worked on the same issue yesterday with @deltakosh #7219

PR #7223, opened 2019-11-27
[Playground] Debounce calls to analyzecode

sailro:
Prevent recomputing deprecated markers on every keystroke (thus preventing flickering).


sailro:
@deltakosh please squash-merge this one after review. Thanks!

PR #7231, opened 2019-11-29
Add global exclusion list for compressed textures

abogartz:
**Issue:** 

Once compressed texture formats are set, there is no way to exclude certain formats, such as .dds or certain image files from attempting to load the `.ktx` version. 

This allows a new method, `setCompressedTextureExclusions` to provide an array of file extensions or file names to exclude.

### DIFF
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -1475,7 +1475,7 @@ export class ThinEngine {
         return dataBuffer;
     }
 
-    protected _normalizeIndexData(indices: IndicesArray): Uint16Array | Uint32Array 
+    protected _normalizeIndexData(indices: IndicesArray): Uint16Array | Uint32Array
     {
         if (indices instanceof Uint16Array) {
             return indices;
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -174,7 +174,7 @@ export class ThinEngine {
     /**
     * Gets or sets the textures that the engine should not attempt to load as compressed
     */
-    protected excludedCompressedTextures: string[] = [];
+    protected _excludedCompressedTextures: string[] = [];
 
     /**
      * Filters the compressed texture formats to only include
@@ -186,7 +186,7 @@ export class ThinEngine {
      */
     public excludedCompressedTextureFormats(url: Nullable<string>, textureFormatInUse: Nullable<string>): Nullable<string> {
         const skipCompression = (): boolean => {
-            return this.excludedCompressedTextures.some((entry) => {
+            return this._excludedCompressedTextures.some((entry) => {
                 const strRegExPattern: string = '\\b' + entry + '\\b';
                 return (url && (url === entry || url.match(new RegExp(strRegExPattern, 'g'))));
             });
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -171,6 +171,29 @@ export class ThinEngine {
         Effect.ShadersRepository = value;
     }
 
+     /**
+     * Gets or sets the textures that the engine should not attempt to load as compressed
+     */
+    public static ExcludedCompressedTextures: string[] = [];
+
+    /**
+     * Filters the compressed texture formats to only include
+     * files that are not included in the skippable list
+     *
+     * @param url the current extension
+     * @param textureFormatInUse the current compressed texture format
+     * @returns "format" string
+     */
+    public static ExcludedCompressedTextureFormats(url: Nullable<string>, textureFormatInUse: Nullable<string>): Nullable<string> {
+        const skipCompression = (): boolean => {
+            return ThinEngine.ExcludedCompressedTextures.some((entry) => {
+                const strRegExPattern: string = '\\b' + entry + '\\b';
+                return (url && url) === entry || (url && url.match(new RegExp(strRegExPattern, 'g')));
+            });
+        };
+        return skipCompression() ? null : textureFormatInUse;
+    }
+
     // Public members
 
     /** @hidden */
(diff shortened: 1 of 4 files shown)


abogartz:
Some possible improvements:

Make the provided array allow for regex strings to add flexibility.

src/Engines/thinEngine.ts
```diff
@@ -171,6 +171,29 @@ export class ThinEngine {
...
+     * Filters the compressed texture formats to only include
+     * files that are not included in the skippable list
+     *
+     * @param url the current extension
+     * @param textureFormatInUse the current compressed texture format
+     * @returns "format" string
+     */
+    public static ExcludedCompressedTextureFormats(url: Nullable<string>, textureFormatInUse: Nullable<string>): Nullable<string> {
+        const skipCompression = (): boolean => {
+            return ThinEngine.ExcludedCompressedTextures.some((entry) => {
```
sebavan: not sure we should use some for older browser support as we do not enforce polyfill use ???

src/Engines/thinEngine.ts
```diff
@@ -171,6 +171,29 @@ export class ThinEngine {
...
+                const strRegExPattern: string = '\\b' + entry + '\\b';
+                return (url && url) === entry || (url && url.match(new RegExp(strRegExPattern, 'g')));
```
sebavan: (url && url) === entry could we just not use url === entry ?

src/Engines/thinEngine.ts
```diff
@@ -171,6 +171,29 @@ export class ThinEngine {
         Effect.ShadersRepository = value;
     }
 
+     /**
+     * Gets or sets the textures that the engine should not attempt to load as compressed
+     */
+    public static ExcludedCompressedTextures: string[] = [];
```
sebavan: Could this be not static ???

This would prevent 2 instances in the same page to require the same setup. This would follow the same than textureFormatToUse

sebavan:
A final one, Could you also use the exclusion in the engine.CubeTexture file to ensure it works also for environments ?

src/Engines/thinEngine.ts
abogartz: I can do that, but I think the Typescript transpiler should handle that. But it seems like browser support for it is pretty universal:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

Can you confirm?

src/Engines/thinEngine.ts
abogartz: We don't want this method to return false if there are null values, so we only want to do this test if both `url` and the exclusion string are populated.

src/Engines/thinEngine.ts
abogartz: I don't think it would require two instances. This array is populated but the `setCompressedTextureExclusions` and only needs to be called once. Can you explain?

abogartz:
> A final one, Could you also use the exclusion in the engine.CubeTexture file to ensure it works also for environments ?

Like here?

https://github.com/BabylonJS/Babylon.js/pull/7231/files#diff-5a64960f6ed8d802dcfb24220fb7a285R236

src/Engines/thinEngine.ts
sebavan: Looks like it is already everywhere we need it so you are fine :-)

src/Engines/thinEngine.ts
sebavan: but what does it change for (url && url) === entry I am still confuse I am sorry :-) 

isnt (null && null) equal to null and any none null value equal to themselves ?

abogartz: Oh shoot I think the paren got moved. I'll update, sorry.

src/Engines/thinEngine.ts
sebavan: Yes but setCompressedTextureExclusions is calling into ThinEngine.ExcludedCompressedTextures = skippedFiles;

meaning for any different instance of Engine (if I have several canvases in my page and each of them have different engine) the value will be the same.

In the set method you want to use this.excludedCompressedTextures = skippedFiles;

so all of them can use a different one. Think about it like this one has been done
```
// Hardware supported Compressed Textures
protected _texturesSupported = new Array<string>();
```

src/Engines/thinEngine.ts
```diff
@@ -171,6 +171,29 @@ export class ThinEngine {
...
+     *
+     * @param url the current extension
+     * @param textureFormatInUse the current compressed texture format
+     * @returns "format" string
+     */
+    public static ExcludedCompressedTextureFormats(url: Nullable<string>, textureFormatInUse: Nullable<string>): Nullable<string> {
+        const skipCompression = (): boolean => {
+            return ThinEngine.ExcludedCompressedTextures.some((entry) => {
+                const strRegExPattern: string = '\\b' + entry + '\\b';
+                return (url && url === entry) || (url && url.match(new RegExp(strRegExPattern, 'g')));
```
sebavan:
```
(url && (url === entry || url.match(new RegExp(strRegExPattern, 'g'));
```

not sure it changes anything but win a couple of bytes :-) ???

sebavan:
So mostly missing the static to instance conversion and will be all good :-)

src/Engines/thinEngine.ts
abogartz: Done :-)

src/Engines/thinEngine.ts
abogartz: Ah, I see now. Ok, makes more sense, thank you!

abogartz: Done!

src/Engines/thinEngine.ts
```diff
@@ -171,6 +171,29 @@ export class ThinEngine {
         Effect.ShadersRepository = value;
     }
 
+    /**
+    * Gets or sets the textures that the engine should not attempt to load as compressed
+    */
+    protected excludedCompressedTextures: string[] = [];
```
sebavan: needs to start with _ as it is a protected member _excludedCompressedTextures or the build will fail but other than that all looks good !!!!

sebavan:
All perfect !!! waiting for the build and I ll merge it in thanks a lot !!!

PR #7235, opened 2019-11-30
added optional `initialTab` .. debug layer config option

ycw:
``` 
scene.debugLayer.show({
  initialTab: "statistics"
});
```

`initialTab` is one of 
- `DebugLayer.ACTION_TAB.PROPERTIES`
- `DebugLayer.ACTION_TAB.DEBUG`
- `DebugLayer.ACTION_TAB.TOOLS`
- `DebugLayer.ACTION_TAB.STATISTICS`
- `DebugLayer.ACTION_TAB.SETTINGS`

corresponding string values are 
- `"properties"`
- `"debug"`
- `"tools"`
- `"statistics"`
- `"settings"`

default is `"properties"`

### DIFF
--- a/inspector/src/components/actionTabs/actionTabsComponent.tsx
+++ b/inspector/src/components/actionTabs/actionTabsComponent.tsx
@@ -24,8 +24,19 @@ interface IActionTabsComponentProps {
     popupMode?: boolean,
     onPopup?: () => void,
     onClose?: () => void,
-    globalState?: GlobalState
+    globalState?: GlobalState,
+    initialTab?: string
 }
+
+const ACTION_TAB_INDEX: {
+    [k: string]: number
+} = {
+    PROPERTIES: 0,
+    DEBUG: 1,
+    STATISTICS: 2,
+    TOOLS: 3,
+    SETTINGS: 4
+};
 
 export class ActionTabsComponent extends React.Component<IActionTabsComponentProps, { selectedEntity: any, selectedIndex: number }> {
     private _onSelectionChangeObserver: Nullable<Observer<any>>;
@@ -35,13 +46,19 @@ export class ActionTabsComponent extends React.Component<IActionTabsComponentPro
     constructor(props: IActionTabsComponentProps) {
         super(props);
 
-        let initialIndex = 0;
+        let initialIndex = ACTION_TAB_INDEX.PROPERTIES;
+        if (props.initialTab !== undefined) {
+            let index = ACTION_TAB_INDEX[props.initialTab.toUpperCase()];
+            if (index !== undefined) {
+                initialIndex = index;
+            }
+        }
 
         if (this.props.globalState) {
             const validationResutls = this.props.globalState.validationResults;
             if (validationResutls) {
                 if (validationResutls.issues.numErrors || validationResutls.issues.numWarnings) {
-                    initialIndex = 3;
+                    initialIndex = ACTION_TAB_INDEX.TOOLS;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


inspector/src/components/actionTabs/actionTabsComponent.tsx
```diff
@@ -24,9 +24,20 @@ interface IActionTabsComponentProps {
     popupMode?: boolean,
     onPopup?: () => void,
     onClose?: () => void,
-    globalState?: GlobalState
+    globalState?: GlobalState,
+    initialTab?: string
```
sebavan: Why not relying on an enum ?

sebavan:
I wonder if we should not use an enum instead to ensure only the available config are supported.

Asking @deltakosh  For review ?

inspector/src/components/actionTabs/actionTabsComponent.tsx
ycw: i found that /src/tsconfig.json restricted 'rootDir' to './', implied that there's no way to 
> export an `enum` from /inspector/src/..

and then 

> import it to /src/Debug/debugLayer.ts 

so i chose to use `string`

src/Debug/debugLayer.ts
```diff
@@ -121,6 +125,17 @@ export class DebugLayer {
      */
     public static InspectorURL = `https://unpkg.com/babylonjs-inspector@${Engine.Version}/babylon.inspector.bundle.js`;
 
+    /**
+     * Action tab kinds
+     */
+    public static ACTION_TAB = {
```
deltakosh: why not exporting this as an enum (export enum DebugLayerTabs...) so you do not need to recreate it in the action tab class

ycw: do you mean that export enum from debugLayer.ts, then inspector.ts import that enum?

ycw: > 
> 
> why not exporting this as an enum (export enum DebugLayerTabs...) so you do not need to recreate it in the action tab class

imo, the debuglayer is a consumer of inspector.. what tabs are available in debuglayer should be the tabs exporting from inspector.. should not be opposite, what do you think?

ycw: > 
> 
> > why not exporting this as an enum (export enum DebugLayerTabs...) so you do not need to recreate it in the action tab class
> 
> imo, the debuglayer is a consumer of inspector.. what tabs are available in debuglayer should be the tabs exporting from inspector.. should not be opposite, what do you think?

the reason why i can't do that has mentioned at https://github.com/BabylonJS/Babylon.js/pull/7235#discussion_r352292106 :D

deltakosh: Yes this is weird I agree but we are already doing it. It is ok to expose it and import it from the inspector

ycw: ok.

ycw:
usage

```
scene.debugLayer.show({
  initialTab: DebugLayerTab.STATISTICS
});
```

`initialTab` options:
- `DebugLayerTab.PROPERTIES` (default)
- `DebugLayerTab.DEBUG`
- `DebugLayerTab.STATISTICS`
- `DebugLayerTab.TOOLS`
- `DebugLayerTab.SETTINGS`

src/Debug/debugLayer.ts
```diff
@@ -108,6 +112,32 @@ Object.defineProperty(Scene.prototype, "debugLayer", {
...
 });
 
+/**
+ * Enum of inspector action tab
+ */
+export enum DebugLayerTab {
+    /**
+     * Properties tag (default)
+     */
+    PROPERTIES = 0,
```
deltakosh: Enums are not upper case:
https://github.com/BabylonJS/Babylon.js/blob/a407a04631d848ac89535d64282af004f0ec2cea/src/Materials/Node/Enums/nodeMaterialSystemValues.ts#L4

deltakosh: But not a big deal..I will fix that after the merge ;)

Thanks a ton!

PR #7245, opened 2019-12-02
IE 11 array.find not available.

sebavan:
Waiting for the build before anything.


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -440,8 +440,8 @@ export class GLTFLoader implements IGLTFLoader {
     private _checkExtensions(): void {
         if (this._gltf.extensionsRequired) {
             for (const name of this._gltf.extensionsRequired) {
-                const extension = this._extensions.find((extension) => extension.name === name);
-                if (!extension || !extension.enabled) {
+                const available = this._extensions.some((extension) => extension.name === name && extension.enabled);
+                if (!available) {
                     throw new Error(`Require extension ${name} is not available`);
```
bghgary: Do you mind fixing this too?
```suggestion
                    throw new Error(`Required extension ${name} is not available`);
```

PR #7247, opened 2019-12-02
Clear coat conservation Factor

sebavan:
Do not merge so far, waiting for the build results.

This should improve the furnace test results.


sebavan:
sounds like it is a keeper :-)

Issue #6349, opened 2019-05-15
Double __root__ on export as glb a gltf object


Bulisor:
Discusion here:
https://forum.babylonjs.com/t/double-root-when-saving-out-glb-from-sandbox/2608

Drigax:
I initially looked at this change: https://github.com/BabylonJS/Babylon.js/pull/7344/files
[Reverted via this PR](https://github.com/BabylonJS/Babylon.js/pull/7366/files)

This change added some extra checking of nodes to be exported and inverted between coordinate systems, and appears to work properly for Left Handed coordinate systems.  Unfortunately, we detect whether or not a node is used as a coordnate system transformation via isNodeConvertingToLeftHanded() this particular line seems somewhat suspect:
https://github.com/BabylonJS/Babylon.js/pull/7344/files#diff-27fc9e54be12c9c3784cf3eb48540e70R1315

In that it works when we use a left handed coordinate system, however, round tripped right handed scenes would still have an encapsulating root node, but would have an Identity transform matrix:
![image](https://user-images.githubusercontent.com/3952729/74177948-92c6f200-4bef-11ea-92fc-165daaa6b98a.png)

I'll add some additional logic to use the proper transform matrix depending on the scene handedness.

Drigax:
In addition, #7377 also indicates some issue with the encapsulating node used to represent punctual lights in glTF, the logic is similar, we need to prevent duplication of a light's node as well:

![image](https://user-images.githubusercontent.com/3952729/74178176-f8b37980-4bef-11ea-9b19-e54030da43be.png)

PR #7257, opened 2019-12-03
Fix the double __root__ issue for glb

Bloadrick:
Fix for the the double root issue: https://github.com/BabylonJS/Babylon.js/issues/6349

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -414,7 +414,8 @@ export class GLTFLoader implements IGLTFLoader {
                 }
             }
 
-            const rootNode = this._createRootNode();
+            const rootNode = this._createRootNode(this._gltf.nodes);
+
             for (const node of this._gltf.nodes) {
                 const parentIndex = nodeParents[node.index];
                 node.parent = parentIndex === undefined ? rootNode : this._gltf.nodes[parentIndex];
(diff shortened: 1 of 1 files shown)


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
 
-        const rootNode: INode = {
-            _babylonTransformNode: this._rootBabylonMesh,
-            index: -1
-        };
+        var indexScene = this._gltf.scene || 0;
+        var scene = this._gltf.scenes[indexScene];
+
+        // Root node must be unique
+        if (scene.nodes.length != 1) {
```
deltakosh:
```suggestion
        if (scene.nodes.length !== 1) {
```

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
+            return null;
+        }
+        else if(this._gltf.nodes != null) {
```
deltakosh:
```suggestion
        else if(this._gltf.nodes !== null) {
```

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
+            var indexRoot = scene.nodes[0];
+            var root = this._gltf.nodes[indexRoot];
+
+            // Check rotation and scale
+            if(root.rotation != undefined && root.scale != undefined){
```
deltakosh: ok...stopping here but please fix all triple equals ;)

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
+                if (root.rotation.join(',') != "0,0,0,1" || root.scale.join(',') != "1,1,-1") {
+                    return null;
+                }
+            }
+
+            // Check name
+            if (root.name != "__root__") {
```
deltakosh: not sure we want to check this one. As long as we have the correct transforms we are fine

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
+                return null;
+            }
+        }else{
```
deltakosh: spaces

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
+            return null;
+        }
+
+        return root;
+    }
+
+    private _createRootNode(nodes : INode[]): INode {
+
+        var ExistingRootNode = this._checkForRootNode();
```
deltakosh: no capital first letter

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -453,21 +454,76 @@ export class GLTFLoader implements IGLTFLoader {
...
+        var RootAlreadyImplemented = false;
+        var INodeToReturn;
```
deltakosh: need a type

bghgary:
This change is not in the right place. The double root should be removed on export and not on import.

deltakosh:
Ok I was under the same impression

Closing this PR then

Bloadrick:
Hi, I think we are both right here after thinking about it again.

So 3ds max and maya are right handed, they export in Babylon which is left handed, no root here, we are changing the geometry and the transform.
Then if we export in GLTF we are doing the same operation between Babylon and GLTF.

Current state:

After that if we put a GLTF in the sandbox we need to add a root to put it in left handed coordinates.
For the moment when exporting to GLB we keep the root added previously without adding an other one.
Then if we put this glb in the sandbox, it adds another root which is not needed (that's where I got confused for this issue).

What to do:

When we export in GLB from the sandbox we need to delete that root to keep it right handed. 
I think what I did is still usefull for legacy purpose, if a glb has been exported with multiple root it would accept it and format it in the right coordinates.

If you are ok with that I can add the delete of the root node on export and keep this PR with the changes asked implemented.
It would prevent the multiple roots and keep it to one when a GLTF/GLB is loaded.

deltakosh:
I agree we should not add a root node if not necessary. But I struggle to convince myself that we will have the cae. When will this happen if we fix the serializer?

Bloadrick:
If a GLB has already been exported, in my test files some of them have up to 4 or 5 root nodes.
I was thinking about someone who already used it and have already a root node in his GLB if not checked in the loader it is going to add a second one.

deltakosh:
Ok that makes to me
@bghgary  ?

bghgary:
I don't think the loader should do anything different.

The loader must do something in order to convert the data from right-handed (glTF) to left-handed (Babylon assuming useRightHandedSystem is false). I don't think the loader should muck with the hierarchy of the glTF itself.

If the existing glTF already has a bunch of unnecessary root nodes, then it should be up to the author of the glTF to fix the problem. The main fix is to make sure the exporter does not perpetuate the issue.

deltakosh:
Yeah you are right. We always have to switch to left...so yeah we only need the serializer part :D

PR #7260, opened 2019-12-04
AR Pointer event simulation on screen target ray mode

RaananW:
'screen' target ray mode (mainly mobile) will now have emulated pointer events that will allow interaction with the scene

### DIFF
--- a/src/Cameras/XR/webXREnterExitUI.ts
+++ b/src/Cameras/XR/webXREnterExitUI.ts
@@ -117,6 +117,11 @@ export class WebXREnterExitUI implements IDisposable {
         });
     }
 
+    /**
+     *
+     * @param scene babylon scene object to use
+     * @param options read-only version of the options passed to this UI
+     */
     private constructor(private scene: Scene, public readonly options: WebXREnterExitUIOptions) {
         this._overlay = document.createElement("div");
         this._overlay.style.cssText = "z-index:11;position: absolute; right: 20px;bottom: 50px;";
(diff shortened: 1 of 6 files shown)


deltakosh:
We still have a doc issue ;(

RaananW:
It is all there, but it seems like the doc-check doesnt like readonly definitions in the constructor.
I will remove the readonly

deltakosh:
Still ;D

deltakosh:
image

RaananW:
I keep on searching, but it is there - 

image

checking further :-)

src/Cameras/XR/webXREnterExitUI.ts
```diff
@@ -117,7 +117,12 @@ export class WebXREnterExitUI implements IDisposable {
         });
     }
 
-    private constructor(private scene: Scene, options: WebXREnterExitUIOptions) {
+    /**
+     *
+     * @param scene babylon scene object to use
+     * @param options (read-only) version of the options passed to this UI
+     */
+    private constructor(private scene: Scene, public options: WebXREnterExitUIOptions) {
```
deltakosh: correct one is:
/**
     * BLA BLA BLA
     * @param scene babylon scene object to use
     * @param options (read-only) version of the options passed to this UI
     */
    private constructor(private scene: Scene, 
      /** BBLA BLA BLA */
      public options: WebXREnterExitUIOptions) {

src/Cameras/XR/webXRSessionManager.ts
```diff
@@ -62,7 +66,7 @@ export class WebXRSessionManager implements IDisposable {
      * Constructs a WebXRSessionManager, this must be initialized within a user action before usage
      * @param scene The scene which the session should be created for
      */
-    constructor(private scene: Scene) {
+    constructor(public scene: Scene) {
```
deltakosh: same here the scene parameter needs a comment

deltakosh:
Yeahhh! ;)

PR #7266, opened 2019-12-05
Creation working canvas only if needed (CubeTexture)

kyptov:
If images with right size  there is no need in additional working canvas.
Added Logger.Warn, silent return took some time for investigation.

This change makes possible to use CubeTexture in a worker environment in case if images don`t need resizing.


deltakosh:
You only have a tiny linting issue. Just run npm run build to get it fixed :)

deltakosh:
Actually I will fix it :) no worries and thanks!

PR #7268, opened 2019-12-05
1st pass implementation of Cascaded Shadow Maps

lockphase:
So, this PR is a 1st pass working prototype of the cascaded shadow maps. I tried to inject codes and/or copy paste codes with minimal impact to existing shadowGenerator functionalities. I hope I did not break anything. Notables:

getShadowMap returns null when CSM is in use. Files that use getShadowMap() and not yet updated for CSM are (as of writing):

Playground/scripts/proceduralTexture.js
Playground/scripts/instanced bones.js
materialsLibrary/index.html
proceduralTexturesLibrary/index.html
Viewer/src/managers/sceneManager.ts
src/PostProcesses/RenderPipeline/Pipelines/standardRenderingPipeline.ts

All observables in _initializeCascadedShadowMaps() are copy pasted, unverified. I have no idea what they do exactly, so did not touch. Similarly for _renderSubMeshForCSMs(), largely copy pasted.

Filters, serialization and parsing not implemented for CSM.

Pls check!


dist/preview release/what's new.md
```diff
@@ -45,11 +45,16 @@
 - Added `AnimationGroup.onAnimationGroupLoopObservable` ([Deltakosh](https://github.com/deltakosh/))
 - Supports custom materials to generate glow through `referenceMeshToUseItsOwnMaterial` in the `GlowLayer` ([sebavan](http://www.github.com/sebavan))
 - Added `RawTexture2DArray` to enable use of WebGL2 2D array textures by custom shaders ([atg](https://github.com/atg))
+<<<<<<< HEAD
```
sebavan: Looks like a merge leftover

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,100 @@ export class ShadowGenerator implements IShadowGenerator {
         if (this._shadowMap2) {
             return this._shadowMap2;
         }
-
         return this._shadowMap;
```
sebavan: not sure if this should also handle CSM

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,100 @@ export class ShadowGenerator implements IShadowGenerator {
...
+        return this._useCSM;
+    }
+    /**
+     * Sets the CSM mode.
+     */
+    public set useCSM(value: boolean) {
+        if (this._useCSM === value) {
+            return;
+        }
+        this._useCSM = value;
```
sebavan: this should recreate the internal textures/post pro like the other modes I guess

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,100 @@ export class ShadowGenerator implements IShadowGenerator {
...
+    }
+
+    private _csmFrustumLength: number;
+    /**
+     * Gets the csmFrustumLength value: furthest range of the frustum for the CSM mode.
+     */
+    public get csmFrustumLength(): number {
```
sebavan: Why not relying on the camera or light minShadowZ  and MaxShadowZ like the rest of the shadows ?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,100 @@ export class ShadowGenerator implements IShadowGenerator {
...
+    }
+
+    private _cascadeShadowMaps: Nullable<Array<RenderTargetTexture>>;
+
+    /**
+     * Gets the RTT containing the cascaded shadow map (usually storing depth from the light point of view).
+	 * @param index index number of the cascaded shadow map
+     * @returns The render target texture if present otherwise, null
+     */
+    public getCSM(index: number): Nullable<RenderTargetTexture> {
```
sebavan: Not sure we need this one, we might simply all the time use the array to prevent extra validation in a loop for instance

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,100 @@ export class ShadowGenerator implements IShadowGenerator {
...
+            return null;
+        }
+    }
+
+    /**
+     * Gets the array containing all the shadow maps (usually storing depth from the light point of view).
+     * @returns The array containing the cascaded shadow maps if present otherwise, empty array
+     */
+    public getAllCSMs(): Array<RenderTargetTexture> {
+        if (this.useCSM && this._cascadeShadowMaps && this._cascadeShadowMaps.length !== 0) {
```
sebavan: If we have CSM the do we need to validate the array ? it should always be created and not empty ?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,100 @@ export class ShadowGenerator implements IShadowGenerator {
...
+            return this._cascadeShadowMaps;
+        } else {
+            return [];
```
sebavan: Could we still return this._cascadedShadowMaps and initialize it with [] to prevent GC ?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -675,21 +790,37 @@ export class ShadowGenerator implements IShadowGenerator {
      * @returns the Shadow Generator itself
      */
     public addShadowCaster(mesh: AbstractMesh, includeDescendants = true): ShadowGenerator {
-        if (!this._shadowMap) {
+        if (this.useCSM && this._cascadeShadowMaps) {
+            if (this._cascadeShadowMaps.length === 0) {
```
sebavan: not sure how we could end up in this case ?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -776,15 +926,30 @@ export class ShadowGenerator implements IShadowGenerator {
      * @param mapSize The size of the texture what stores the shadows. Example : 1024.
      * @param light The light object generating the shadows.
      * @param usefulFloatFirst By default the generator will try to use half float textures but if you need precision (for self shadowing for instance), you can use this option to enforce full float texture.
+	 * @param cascades Number of cascades for CSM. Values between MIN_CASCADES and MAX_CASCADES.
+	 * @param csmFrustumLength The distance of the furthest frustum of the camera for CSM.
      */
-    constructor(mapSize: number, light: IShadowLight, usefulFloatFirst?: boolean) {
+    constructor(mapSize: number, light: IShadowLight, usefulFloatFirst?: boolean, cascades?: number, csmFrustumLength?: number) {
```
sebavan: not sure the arguments should cotains csm options as it is not the case for the rest of the shadow parameters.

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -776,15 +926,30 @@ export class ShadowGenerator implements IShadowGenerator {
...
-    constructor(mapSize: number, light: IShadowLight, usefulFloatFirst?: boolean) {
+    constructor(mapSize: number, light: IShadowLight, usefulFloatFirst?: boolean, cascades?: number, csmFrustumLength?: number) {
         this._mapSize = mapSize;
         this._light = light;
         this._scene = light.getScene();
         light._shadowGenerator = this;
 
         ShadowGenerator._SceneComponentInitialization(this._scene);
 
+        // sets cascade value and the csmFrustumLength if any, valid only for directional lights
```
sebavan: This could be done in the rebuild function to ensure that if the type of shadow changes we would do the same

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -814,9 +979,110 @@ export class ShadowGenerator implements IShadowGenerator {
...
+        } else if (camera.fovMode === 1) {
+            nearx = camera.minZ * Math.tan(camera.fov / 2);
+            neary = camera.minZ * Math.tan(camera.fov / 2) * engine.getAspectRatio(camera);
+            farx = this._csmFrustumLength * Math.tan(camera.fov / 2);
+            fary = this._csmFrustumLength * Math.tan(camera.fov / 2) * engine.getAspectRatio(camera);
+        }
+
+        // populate the viewSpaceFrustums array
+        for (let i = 0; i < this.cascades + 1; i++) {
+            let temp = [];
```
sebavan: why not pushing all the way directly to the array to prevent the temp ???

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -814,9 +979,110 @@ export class ShadowGenerator implements IShadowGenerator {
...
+        let engine = this._scene.getEngine();
+        this._viewSpaceFrustums = [];
+        this._viewSpaceBoundingSpheres = [];
+        let nearx = 0;
+        let neary = 0;
+        let farx = 0;
+        let fary = 0;
+
+        // get all internal camera cascaded frustum points
+        let breaks = this._frustumSplit(this.cascades, camera.minZ, this._csmFrustumLength, 0.5);
```
sebavan: 0.5 should be the default value for lambda but kept configurable as a property

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -888,6 +1154,88 @@ export class ShadowGenerator implements IShadowGenerator {
...
+        this._CSMtransformMatrices = [];
+        // for each cascade, create and set a light and shadowmap to said cascade
+        for (let index = 0; index < this.cascades; index++) {
+            if (engine.webGLVersion > 1) {
+                this._cascadeShadowMaps[index] = new RenderTargetTexture(this._light.name + "_cascadeShadowMap" + index, this._mapSize, this._scene, false, true, this._textureType, this._light.needCube(), undefined, false, false);
+                this._cascadeShadowMaps[index].createDepthStencilTexture(Constants.LESS, true);
+            } else {
+                this._cascadeShadowMaps[index] = new RenderTargetTexture(this._light.name + "_cascadeShadowMap" + index, this._mapSize, this._scene, false, true, this._textureType, this._light.needCube());
+            }
+            this._cascadeShadowMaps[index].wrapU = Texture.CLAMP_ADDRESSMODE;
```
sebavan: This creation should depend of the type of shadows e.g. simple/poisson/PCF and so on ?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -1514,6 +2090,12 @@ export class ShadowGenerator implements IShadowGenerator {
             this._shadowMap.dispose();
             this._shadowMap = null;
         }
+        if (this.useCSM) {
```
sebavan: every resources should be disposed

src/Materials/materialHelper.ts
```diff
@@ -340,11 +341,21 @@ export class MaterialHelper {
         if (mesh && mesh.receiveShadows && scene.shadowsEnabled && light.shadowEnabled) {
             var shadowGenerator = light.getShadowGenerator();
             if (shadowGenerator) {
-                const shadowMap = shadowGenerator.getShadowMap();
-                if (shadowMap) {
-                    if (shadowMap.renderList && shadowMap.renderList.length > 0) {
-                        state.shadowEnabled = true;
-                        shadowGenerator.prepareDefines(defines, lightIndex);
+                if (shadowGenerator.useCSM) {
```
sebavan: This looks pretty similar I guess we could facto a bit

src/Meshes/abstractMesh.ts
```diff
@@ -1624,13 +1624,27 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
             // Shadow generators
             var generator = light.getShadowGenerator();
             if (generator) {
-                var shadowMap = generator.getShadowMap();
-
-                if (shadowMap && shadowMap.renderList) {
-                    meshIndex = shadowMap.renderList.indexOf(this);
+                if (generator.useCSM) {
```
sebavan: same here, I wonder if the array in none CSM could contain only one map to simplify those blocks

src/Shaders/ShadersInclude/lightFragment.fx
```diff
@@ -162,6 +162,29 @@
...
                 shadow = computeShadowCube(light{X}.vLightData.xyz, shadowSampler{X}, light{X}.shadowsInfo.x, light{X}.depthValues);
+			#elif defined(SHADOWCSM{X})
+				int index = -1;
+				for (int i = 0; i < num_cascades{X}; i++) {
+					if (vPositionFromCameraCSM{X}.z <= ViewFrustumZCSM{X}[i]){
+						index = i;
+						break;
+					}
+				}
+				if (index == 0){
```
sebavan: If we lock down CSM to webgl2 is it possible to access through dynamic index in the array ?

dist/preview release/what's new.md
lockphase: My bad, fixed!

src/Lights/Shadows/shadowGenerator.ts
lockphase: No idea what shadowMap2 was used for, so did not touch it. Can always KIV.

src/Lights/Shadows/shadowGenerator.ts
lockphase: Are you suggesting the user implement csm with smthg like 
```
var shadowgenerator = new ShadowGenerator(512, light);
shadowgenerator.useCSM = true;
```
This was my initial take but I found it inefficient. Reason is in the comments on the constructor.

src/Lights/Shadows/shadowGenerator.ts
lockphase: afaik, they mean different things.
a) camera.maxZ vs csmFrustumLength
Using racing game as example, scenery from driver seat falls within camera minZ and maxZ limits. So a far building say, 1000 units away does not need to be in CSM range but the competing car in the next lane should be in csmFrustumLength range. This allows the user to configure its scene appropriately. If user wishes to set csmFrustumLength = camera.maxZ, that is still fine but if the user does not wish to, he/she must have the option not to.

b) light.shadowMaxZ vs csmFrustumLength
How far a light should cast its shadows and the csm frustum length are independent from each other. Imagine camera, a sphere 100 units away, and a light behind the sphere 5 units away all in a straight line: camera ---100--- sphere -5- light. If light..shadowMaxZ is 20, then the camera should see the sphere's shadow, right ? But if the csmFrustumLength is also 20, the camera will see a sphere with no shadow.

src/Lights/Shadows/shadowGenerator.ts
lockphase: Thanks for the catch! A whole bunch of this were put in during testing. Now removed !

src/Lights/Shadows/shadowGenerator.ts
lockphase: Right ! Changed to ```if (this.useCSM)```.

src/Lights/Shadows/shadowGenerator.ts
lockphase: Right, fixed! Now ```private _cascadeShadowMaps: Array<RenderTargetTexture>``` looks much better

src/Lights/Shadows/shadowGenerator.ts
lockphase: Fixed!

src/Lights/Shadows/shadowGenerator.ts
lockphase: So this is the part where I initially considered to code as:
```
Line1: var shadowgenerator = new ShadowGenerator(1024, light);
Line2: shadowgenerator.useCSM = true;
shadowgenerator.<insert filters blah blah>
```
I think that Line1 would create a shadowMap immediately even if the user did not have the intention to use it. And then the useCSM would have to dispose it before initializing csm RTTs. Basically, wasted CPU cycles. Then I thought we could just treat Line1 as a special case of cascade = 1 but I found it a hassle to have to track/rename and reconfigure said map. So I decided to go with:
```
Default
var shadowgenerator = new ShadowGenerator(1024, light);
shadowgenerator.<insert filters blah blah>

CSM
var shadowgenerator = new ShadowGenerator(1024, light, false, 4, 100);
shadowgenerator.<insert filters blah blah>
```
This is very much open for discussion as things could go either way and the difference is minimal. From user perspective. bjs is offering two modes of the basic shadows and it doesn't break back compat. I recognize that it does raise other concerns, biggest of which is whether there will be other types of basic shadows in future to be introduced.

Alternatively, we could just leave it for this PR until its clear that filters or other CSM related improvements do require a new CascadedShadowGenerator class or somesuch...

src/Lights/Shadows/shadowGenerator.ts
lockphase: Nice catch! Fixed!

src/Lights/Shadows/shadowGenerator.ts
lockphase: ok, fixed to ```private _csmLambda = 0.5;```.

src/Lights/Shadows/shadowGenerator.ts
lockphase: No idea, copy pasted from the ```_initializeShadowMap``` function, lol.

src/Lights/Shadows/shadowGenerator.ts
lockphase: ok , changed to 
```
if (this.useCSM) {
		for (let i = 0; i < this._cascadeShadowMaps.length; i++){
			this._cascadeShadowMaps[i].dispose();
		}
		this._viewSpaceFrustums = [];
		this._viewSpaceBoundingSpheres = [];
            this._CSMtransformMatrices = [];
        }
```
_cascadeShadowMaps are no longer nullable, cf, above fix. BoundingSphere, Matrix, Vector3 cannot be disposed nor set to null. Dunno what bjs standard for dispose should be, lemme know if this isn't sufficient.

src/Materials/materialHelper.ts
lockphase: Done!
```
var shadowMap;
if (shadowGenerator.useCSM) {
	shadowMap = shadowGenerator.getAllCSMs()[0];
} else {
	shadowMap = shadowGenerator.getShadowMap();
}
if (shadowMap) {
	if (shadowMap.renderList && shadowMap.renderList.length > 0) {
	state.shadowEnabled = true;
	shadowGenerator.prepareDefines(defines, lightIndex);
	}
}
```

src/Meshes/abstractMesh.ts
lockphase: erm, if I try to refactor this, one is a RTT, the other is an array. The compiler throws errors and the resulting code is longer than this. Gonna leave it for now.

src/Shaders/ShadersInclude/lightFragment.fx
lockphase: Yh, same thoughts, these codes were initially super easy and short, like so:
```
shadow = computeShadow(vPositionFromLightCSM{X}[i], vDepthMetricCSM{X}[i], shadowSamplerArrayCSM{X}[i], light{X}.shadowsInfo.x, light{X}.shadowsInfo.w);
```
Guess what ?
```
BJS - [15:24:32]: Error: C:\fakepath(157,12-35): warning X3550: sampler array index must be a literal expression, forcing loop to unroll
```
plus a whole bunch of errors. I guess this (https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Opaque_arrays) is in effect.

src/Lights/Shadows/shadowGenerator.ts
sebavan: This is the one used by the materials as it might come from a post process in case of blur applied to the map for instance.

We need to handle it in case pp are applied to the shadows.

src/Lights/Shadows/shadowGenerator.ts
sebavan: Agree so you can keep the ctor params but we still need to handle the dynamic case in case the configuration changes later on.

src/Lights/Shadows/shadowGenerator.ts
sebavan: Basically CSM is node a kind of shadow but is complementarty to the chosen shadow types.

You could for instance use PCF and cascades or Poisson sampling and CSM.

The initialization and setup should depend on it as well. If the function is a full copy besides the loop we should find a way to factorize it.

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -656,10 +681,85 @@ export class ShadowGenerator implements IShadowGenerator {
...
+        if (!this._scene || !this._scene.activeCamera) {
+            return;
+        }
+        if (this._csmFrustumLength === value || value < this._scene.activeCamera.minZ || value > this._scene.activeCamera.maxZ) {
+            return;
+        }
+        this._csmFrustumLength = value;
+    }
+
+    private _cascadeShadowMaps: Array<RenderTargetTexture>;
```
sebavan: This should be initialized to [] to prevent undefined exception (just in case ppl call in getCSM functions without being in CSM mode)

src/Lights/Shadows/shadowGenerator.ts
sebavan: I kind of like the idea of the separate class to be honest as it sounds you have "if CSM" everywhere in this one. It might also be much easier to handle the specific parameter and we could easily limit the mode to Simple and PCF for instance.

What do you think ???

src/Lights/Shadows/shadowGenerator.ts
sebavan: Good but would still need a public property to be able to modify it ?

This would dynamically recreate the matrices and so on.

src/Lights/Shadows/shadowGenerator.ts
sebavan: Only the GL ones need to be disposed or the ones retaining memory. it is perfect like this.

src/Meshes/abstractMesh.ts
sebavan: could extract the inside to a function ?

src/Shaders/ShadersInclude/lightFragment.fx
sebavan: Is it the same if you manually unroll ?

As we do for lights actually.

src/Lights/Shadows/shadowGenerator.ts
lockphase: ah, I see it now. ok fixed by adding ```private _cascadeShadowMaps2: Array<RenderTargetTexture>;``` and
```
public getCSMForRendering(index: number): RenderTargetTexture {
     if (this._cascadeShadowMaps2[index]) {
          return this._cascadeShadowMaps2[index];
     }
     return this._cascadeShadowMaps[index];
}
```
and in the onAfterUnbindObservable, fixed to
```let shadowMap = this.getCSMForRendering(index);```

File: src/Lights/Shadows/shadowGenerator.ts (hunk 12 above)
Comment (lockphase) [reply src/Lights/Shadows/shadowGenerator.ts]: ok, fixed by adding in public getters and setters.

File: src/Lights/Shadows/shadowGenerator.ts (hunk 9 above)
Comment (lockphase) [reply src/Lights/Shadows/shadowGenerator.ts]: I needed to get this first pass out and into people's hand so as to
a) get feedback, bug reports etc
b) get more use case scenarios
c) get an idea of demand and what params users want exposed for tweaking
If you look closely at the csmtransformmatrix calcs, there are major differences compared to the default case. CSM ignores position of the directional light, the lookat transform has no margins atm, casters and receivers are identical for all n maps etc... Plus based on the refs thrown at me, only pcf and vsm are the filtering techniques, csm don't seem to have the same artifact problems plaguing the usual case. I had a sneaky suspicion that it *might* be better off as a separate class. But how to get there ?

Hence, 'if CSM' came about. If CSM users could at least 'play' with it using existing function calls, we would have a route by which a), b) and c) could be fulfilled. That would pave the way for a future CascadedShadowGenerator class. 

To summarize, I have no problems with a new class, but my main concern right now is to get more testing and feedback.

File: src/Lights/Shadows/shadowGenerator.ts (hunk 18 above)
Comment (lockphase) [reply src/Lights/Shadows/shadowGenerator.ts]: oh, you are right. fixed to ```private _cascadeShadowMaps: Array<RenderTargetTexture> = [];```

File: src/Shaders/ShadersInclude/lightFragment.fx (hunk 17 above)
Comment (lockphase) [reply src/Shaders/ShadersInclude/lightFragment.fx]: I've tried ```#pragma unroll 4``` (and whatever other syntax) before the loop prior to PR, same result. Also tried searching the repo for pragma or unroll, but I don't see it at all. Unless you are not doing it this way, then I need an example.

File: src/Meshes/abstractMesh.ts (hunk 16 above)
Comment (lockphase) [reply src/Meshes/abstractMesh.ts]: ok, added a ```private removeMesh``` function and factored a little.

File: src/Lights/Shadows/shadowGenerator.ts (hunk 13 above)
Comment (lockphase) [reply src/Lights/Shadows/shadowGenerator.ts]: ok, not entirely sure if this is what you wanted. But I factored out the six lines where the texture properties were inited with ```this._initializeRTTProperties(this._cascadeShadowMaps[index]);```, ditto for the default case.

File: src/Lights/Shadows/shadowGenerator.ts (hunk 3 above)
Comment (lockphase) [reply src/Lights/Shadows/shadowGenerator.ts]: ok, added an init and dispose checks.

File: src/Lights/Shadows/shadowGenerator.ts (hunk 10 above)
Comment (lockphase) [reply src/Lights/Shadows/shadowGenerator.ts]: ok, pulled out into a separate ```private _checkAndInitCascades``` function.

**Comment by lockphase** [comment]
As replied [here](https://github.com/BabylonJS/Babylon.js/pull/7268#discussion_r355093880), I have no problems with a separate class. But the first hurdle is to get this 1st pass codes out and let the user pool try/break/feedback etc. Then we can use this groundwork as a springboard for the launch of the class in, what 4.2 release?

**Comment by Popov72** [comment]
Sorry to chime in, but as I'm too playing with CSM on my own, I thought I would give my two cents about this.

I think a separate class would really be a must have, and delaying it to a next version of Babylon will be a waste of time. The hard part is getting all of the CSM code working, and it seems you have it. Reorganizing the code should not be the biggest problem.

Having a flat `ShadowGenerator` class with all possible options to handle all possible rendering technics will make it more and more difficult to maintain / extend in the future. Plus it feels wrong to have some properties having a meaning only based on some other property/ies... What if/when SDSM or Stencil shadows is supported? Still more code in the `ShadowGenerator` class.

The current `IShadowGenerator` / `ShadowGenerator` is in fact a `IShadowMapGenerator` / `ShadowMapGenerator` as it implements the specific shadow map rendering technic. It's not fitting for a stencil shadow technic implementation, for eg (I think - I don't really know stencil shadows).

What we could have is:

```typescript
interface IShadowGenerator {
    isReady(subMesh: SubMesh, useInstances: boolean): boolean;
    prepareDefines(defines: MaterialDefines, lightIndex: number): void;
    bindShadowLight(lightIndex: string, effect: Effect): void;
    forceCompilation(onCompiled?: (generator: ShadowGenerator) => void, options?: Partial<{ useInstances: boolean }>): void;
    forceCompilationAsync(options?: Partial<{ useInstances: boolean }>): Promise<void>;
    recreate(): void;
    serialize(): any;
    dispose(): void;
}
```
it's the current `IShadowGenerator` interface but stripped from all the shadow map specificities (`getShadowMap`, `getShadowMapForRendering`, `getTransformMatrix`, `recreateShadowMap`). Note that we need a `recreate` method (`PointLight` calls it when switching between cube / not cube): it's the `recreateShadowMap` method but renamed to be more generic.

The current `ShadowGenerator` class would not change (it would be better to be renamed `ShadowMapGenerator`, but that would be a breaking change...).

```typescript
class ShadowCSMGenerator implements IShadowGenerator {
    ...
    private _cascades: Array<ShadowMapGenerator>;
    ...
}
```
The CSM implementation would hold an array of shadow map generators and would not be **a** `ShadowMapGenerator` (as it is when implementing it inside `ShadowGenerator`). It seems more right to me.

Benefits:
* No change to `ShadowGenerator`, so no possibility of breaking things
* The code is not all packed inside a single class (`ShadowGenerator`). Specificities of CSM rendering will be in the `ShadowCSMGenerator` class, not in `ShadowMapGenerator` (`ShadowGenerator` if we don't want breaking changes). When we implement SDSM, we will create a new `ShadowSDSMGenerator` class that extends `ShadowCSMGenerator` and that implements specificities of SDSM over plain CSM.
* We can use a different settings for each map of the cascade. For eg, one could set a smaller width/height for the last map of the cascade than for the first one(s). We can also set a filtering technic less GPU intensive for the farthest maps.

In `MaterialHelper` we have:
```typescript
        if (mesh && mesh.receiveShadows && scene.shadowsEnabled && light.shadowEnabled) {
            var shadowGenerator = light.getShadowGenerator();
            if (shadowGenerator) {
                const shadowMap = shadowGenerator.getShadowMap();
                if (shadowMap) {
                    if (shadowMap.renderList && shadowMap.renderList.length > 0) {
                        state.shadowEnabled = true;
                        shadowGenerator.prepareDefines(defines, lightIndex);
                    }
                }
            }
        }
```
I think we could replace it with (we need to get rid of `.getShadowMap` that does not exist in `IShadowGenerator` anymore):
```typescript
        if (mesh && mesh.receiveShadows && scene.shadowsEnabled && light.shadowEnabled) {
            var shadowGenerator = light.getShadowGenerator();
            if (shadowGenerator && shadowGenerator.mustRender) {
                        state.shadowEnabled = true;
                        shadowGenerator.prepareDefines(defines, lightIndex);
            }
        }
```
so by adding a `mustRender` (or any other name) property in `IShadowGenerator` that will return `true` if the shadow generator must indeed be in effect for the rendering.

In `AbstrachMesh.dispose`, we have:
```typescript
            var generator = light.getShadowGenerator();
            if (generator) {
                var shadowMap = generator.getShadowMap();

                if (shadowMap && shadowMap.renderList) {
                    meshIndex = shadowMap.renderList.indexOf(this);

                    if (meshIndex !== -1) {
                        shadowMap.renderList.splice(meshIndex, 1);
                    }
                }
            }
```
I think we need to have the `renderList` property defined on the `IShadowGenerator` interface, and so replace the code like this:
```typescript
            var generator = light.getShadowGenerator();
            if (generator && generator.renderList) {
                    meshIndex = generator.renderList.indexOf(this);

                    if (meshIndex !== -1) {
                        generator.renderList.splice(meshIndex, 1);
                    }
            }
```
of course, `ShadowMapGenerator` (`ShadowGenerator`) can implement this property by proxying to `RenderTargetTexture.renderList`.

`ShadowCSMGenerator` would implement its own `renderList` property that could be mirrored to all `RenderTargetTexture.renderList` of the cascades, but could also provide specific methods to handle the `renderList` property of each cascade separately and so provides finer level of control over shadow casters for each cascade <= another benefit over the current implementation proposal

It's a long post, but in fact I think there's not a lot of changes to perform in the current code to achieve something like this (I basically listed all there is to do, or sort of).

What do you think?

Popov72:
Te be complete, here's the final `IShadowGenerator` proposal:

```typescript
interface IShadowGenerator {
    isReady(subMesh: SubMesh, useInstances: boolean): boolean;
    prepareDefines(defines: MaterialDefines, lightIndex: number): void;
    bindShadowLight(lightIndex: string, effect: Effect): void;
    forceCompilation(onCompiled?: (generator: ShadowGenerator) => void, options?: Partial<{ useInstances: boolean }>): void;
    forceCompilationAsync(options?: Partial<{ useInstances: boolean }>): Promise<void>;
    recreate(): void;
    mustRender: boolean;
    renderList: Array<AbstrachMesh>;
    serialize(): any;
    dispose(): void;
}
```

lockphase:
@Popov72 Thanks for the feedback, I am not a coder by trade. The maths for CSM were (imo) trivial, in total, it took like only a couple of days to get it right. But I don't think I'm the right person to create the new class, as seen in this PR (lol). If you wish to, you can simply strip the maths and submit the PR for the new class. I will be happy to close this one. Lemme know, thks.

lockphase:
I did a 'npm run build' to resolve the documentation check failing. But npm threw:
```
[10:39:17] TypeError: Cannot read property 'summary' of undefined
    at dependencies (C:\Users\shinj\Babylon.js\Tools\Gulp\tasks\gulpTasks-dependencies.js:40:29)
    at C:\Users\shinj\Babylon.js\Tools\Gulp\tasks\gulpTasks-dependencies.js:64:9
    at taskWrapper (C:\Users\shinj\Babylon.js\node_modules\undertaker\lib\set-task.js:13:15)
    at bound (domain.js:395:14)
    at runBound (domain.js:408:12)
    at asyncRunner (C:\Users\shinj\Babylon.js\node_modules\async-done\index.js:55:18)
    at process._tickCallback (internal/process/next_tick.js:61:11)
[10:39:17] 'circularDependencies' errored after 27 s
```
erm, no idea what this is, need a little help here.

src/Lights/Shadows/shadowGenerator.ts
sebavan: We won t be able to push it before 4.2 as we are already in beta and it is a huge change. At least in a separate class this would isolate the changes and leave more room for improvements. I can propose to extract your code as a separate class during the Christmas vacations.

Agree with the filtering technique and I would only consider PCF and PCSS as it might be easier.

src/Shaders/ShadersInclude/lightFragment.fx
sebavan: We do it for light via special macros that are expanded before compilation.

#include<lightFragment>[0..maxSimultaneousLights]

sebavan:
I can definitely do the extraction of the code during the christmas break as it sounds pretty fun :-) and I was having CSM on the roadmap for 4.2 anyway. I ll basically copy paste most of your code in if you wish ?

lockphase:
@sebavan I'm going overseas from the end of the week to 2020, so this sounds great. I will close this PR. You have been a great and patient teacher for my bumbling self. This work would not have been possible if not for your guidance and the help from many others. I look forward to seeing the new class soon. Thank you very much !

sebavan:
I think you should reopen it as **a draft** https://github.blog/2019-02-14-introducing-draft-pull-requests/ so that I keep it mostly as it is and you will find back your code simply in a separate file when you come back. And you ll be able to add the missing features you wanted too.

I would love to collaborate with you on it as it is definitely Your Great Work :-)

Popov72:
> I can definitely do the extraction of the code during the christmas break as it sounds pretty fun :-) and I was having CSM on the roadmap for 4.2 anyway. I ll basically copy paste most of your code in if you wish ?

I'm actually working on CSM implementation for some time now, I will have something to release on next week, so you may want to wait a bit before diving in (also, enjoy your holidays - hope you have some !).

Issue #7270, opened 2019-12-05
WebXR teleportation uses wrong controls (Windows MR)


nbduke:
# Bugs
Repro playground: https://playground.babylonjs.com/#AM07G2#33

I'm using an HP Windows Mixed Reality Headset. In WebVR, teleportation worked using the thumbstick and felt very predictable/controllable. In WebXR, however, you teleport whenever you touch the directional pad, which is much more difficult to control.

# Feature request
Along with fixing the above bug, I'd like to request a feature regarding WebXR teleportation. It would be nice to be able to enable/disable teleportation, either by disposing the `WebXRControllerTeleportation` instance or by setting some property (e.g. `isEnabled`) on that instance. Currently, once you create an instance of that object, there's no way to turn teleportation off.

PR #7275, opened 2019-12-09
WebXR Teleportation 

RaananW:
closing https://github.com/BabylonJS/Babylon.js/issues/7270

Added an `enabled` flag to the loop. if teleportation.enabled set to false, the teleportation will stop working. Notice that if set enabled WHILE teleporting, results can be unexpected.


RaananW:
fixing docs

Issue #7264, opened 2019-12-04
WebXR: No stereoscopic 3D effect when using MODE_TOPBOTTOM


mattdmorgan:
0) Open https://www.babylonjs-playground.com/#ZM8UQE#4 in a WebXR-enabled browser (e.g. Chrome Canary).
1) Enter VR.
Result: video plays without a stereoscopic effect -- you see a complete image (not top/bottom stripes), but the same image is shown to each eye.
Expect: stereoscopic 3D effect in video.

Works as expected in WebVR-only browsers (e.g. old Edge).

RaananW:
This happens because the rig camera's isRightCamera is not set to true

PR #7276, opened 2019-12-09
VideoDome in WebXR

RaananW:
Fixing https://github.com/BabylonJS/Babylon.js/issues/7264

isRightCamera was never set, which is important for the videoDome


RaananW:
Closing https://github.com/BabylonJS/Babylon.js/issues/7264 when merged

PR #7281, opened 2019-12-09
spriteMap es 100 support

Pryme8:


src/Shaders/spriteMap.fragment.fx
```diff
@@ -37,20 +37,21 @@ mat4 getFrameData(float frameID){
...
-
-    for(int i = 0; i < LAYERS; i++){	
+    vec2 stageUnits = 1. / stageSize;        
     
-        float frameID = texture(tileMaps[i], (tileID + 0.5) / stageSize, 0.).x;		
-
+    for(int i = 0; i < LAYERS; i++){
+        float frameID;
+            switch(i){
+                #define LAYER_ID_SWITCH
```
deltakosh: This seems very weird (ant not used at all)

Pryme8: it's used the #define is replaced when the the correct cases on compilation.

Pryme8: https://github.com/BabylonJS/Babylon.js/pull/7281/files#diff-31bc45eaba6a2d1949cb8de69fdcea23R224

Pryme8: If I do it all inline it will it make it more performant?  I would have to restructure the Injection script to be a lot longer, and then the shader will compound in length per layer.  But if you think that is better, then Ill do that.

deltakosh: Oh so you replace the define with a string right?

Pryme8: Yes sir,
```
let shaderString: string = Effect.ShadersStore['spriteMapPixelShader'];
    let layerSampleString: string = '';

    for (let i = 0; i < this.options.layerCount; i++) {
        layerSampleString += 'case ' + i + ' : frameID = texture(tileMaps[' + i + '], (tileID + 0.5) / stageSize, 0.).x;';
        layerSampleString += 'break;';
    }

    Effect.ShadersStore['spriteMap' + this.name + 'PixelShader'] = shaderString.replace('#define 
    LAYER_ID_SWITCH',  layerSampleString);
```

Issue #7210, opened 2019-11-26
Sandbox fails to load glTF with textures relatively pathed to same folder


Drigax:
Loading a glTF image that references an image texture that is relatively pathed to the same directory causes the scene to fail loading.

Bug only repros when scene is manually added to sandbox. Loading the scene programmatically works without issue.

- Expected result:
Model should load into sandbox without error. This playground loads the repro model using the BABYLON.SceneLoader.LoadAssetContainer API entry point without issue:
https://playground.babylonjs.com/#JMZX3H#1

Likewise manually loading a scene into the sandbox with a relative path other than the current folder succeeds as expected (TestImageURIPathing-Subfolder.gltf):
![image](https://user-images.githubusercontent.com/3952729/69593187-4afbdb80-0fad-11ea-8a6c-d21c7b8ccc50.png)

Also, scene loads without issue in the gltf Viewer extension for VSCode in all engines

- Current result:
Current result causes scene load to fail. 
![image](https://user-images.githubusercontent.com/3952729/69593060-e17bcd00-0fac-11ea-9346-17720e11f297.png)

Attached are the repro (TestImageURIPathing-SameFolder.gltf) and non-repro (TestImageURIPathing-Subfolder.gltf) test scenes:

[TestImageURIPathing-Samefolder.zip](https://github.com/BabylonJS/Babylon.js/files/3889664/TestImageURIPathing-Samefolder.zip)
[TestImageURIPathing-Subfolder.zip](https://github.com/BabylonJS/Babylon.js/files/3889662/TestImageURIPathing-Subfolder.zip)

PR #7283, opened 2019-12-10
Sandbox will now load assets relatively path-ed to same folder

belfortk:
Fix for [#7210](https://github.com/BabylonJS/Babylon.js/issues/7210). Updated FileTools.LoadFiles to preprocess the url string before decoding. This pre-processing removes the "file:/" prefix from the file URL and handles the case when the files are relatively path-ed to the same folder.

### DIFF
--- a/src/Misc/fileTools.ts
+++ b/src/Misc/fileTools.ts
@@ -87,7 +87,11 @@ export class FileTools {
      * Gets or sets a function used to pre-process url before using them to load assets
      */
     public static PreprocessUrl = (url: string) => {
-        return url;
+        let processedURL = url.substring(5).toLowerCase();
+        if (processedURL.indexOf('./') === 0) {
+            processedURL = processedURL.substring(2);
+          }
+        return processedURL;
     }
 
     /**
@@ -150,7 +154,6 @@ export class FileTools {
         }
         else {
             url = this._CleanUrl(input);
-            url = this.PreprocessUrl(input);
         }
 
         if (typeof Image === "undefined") {
@@ -301,7 +304,8 @@ export class FileTools {
     public static LoadFile(url: string, onSuccess: (data: string | ArrayBuffer, responseURL?: string) => void, onProgress?: (ev: ProgressEvent) => void, offlineProvider?: IOfflineProvider, useArrayBuffer?: boolean, onError?: (request?: WebRequest, exception?: LoadFileError) => void): IFileRequest {
         // If file and file input are set
         if (url.indexOf("file:") !== -1) {
-            const fileName = decodeURIComponent(url.substring(5).toLowerCase());
+            const processedUrl = this.PreprocessUrl(url);
+            const fileName = decodeURIComponent(processedUrl);
             const file = FilesInputStore.FilesToLoad[fileName];
             if (file) {
                 return this.ReadFile(file, onSuccess, onProgress, useArrayBuffer, onError ? (error) => onError(undefined, new LoadFileError(error.message, error.file)) : undefined);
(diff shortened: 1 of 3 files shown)


Drigax:
Can you add a rough description of your fix in the descripton, and change your title to explain what you changed?

also, GitHub only properly links this PR to the issue externally if you reference it in a description or comment like this #7210

src/Misc/fileTools.ts
```diff
@@ -87,7 +87,11 @@ export class FileTools {
      * Gets or sets a function used to pre-process url before using them to load assets
      */
     public static PreprocessUrl = (url: string) => {
-        return url;
+        let processedURL = url.substring(5).toLowerCase();
```
sebavan: This can break back compat and the default behaviour can not be change for all the urls ???

Drigax: I agree, I don't think we want to do remove the "file:" prefix here, but just remove "./" from the front of our url

deltakosh: This function is actually for the external users and used during the requestFile processing. Meaning that if baseUrl is set to "" then we will strip off the moniker and it will not work anymore if the content is not a file

I recommend to do it only when url.indexOf("file) is > -1 (inside the loadFile function)

src/Misc/fileTools.ts
```diff
@@ -87,7 +87,11 @@ export class FileTools {
      * Gets or sets a function used to pre-process url before using them to load assets
      */
     public static PreprocessUrl = (url: string) => {
-        return url;
+        let processedURL = url.substring(5).toLowerCase();
+        if (processedURL.indexOf('./') === 0) {
+            processedURL = processedURL.substring(2);
+          }
```
bghgary: Careful with spacing:
```suggestion
        }
```

src/Misc/fileTools.ts
Drigax: Ah, I wasn't aware that PreprocessUrl was a reserved function, we can also make another function "resolve redundant relative paths" to handle just this case as well in the fileLoader if we don't want to override this function.

It may also be more appropriate to add this check to _CleanUrl as well

bghgary: Putting it in _CleanUrl seems right to me.

deltakosh: Agree

belfortk: After talking with @sebavan, we decided that doing a simple if check for the "./" before checking the store would be the best approach

belfortk: reverted the change

src/Misc/fileTools.ts
```diff
@@ -87,7 +87,7 @@ export class FileTools {
      * Gets or sets a function used to pre-process url before using them to load assets
      */
     public static PreprocessUrl = (url: string) => {
-        return url;
```
bghgary: erroneous space removal?

what's new.md
```diff
@@ -205,6 +205,7 @@
 - Sometimes duplicate controller models are loaded in VR ([TrevorDev](https://github.com/TrevorDev))
 - Particle emit rate and start size over time do not reset on every particle system start ([TrevorDev](https://github.com/TrevorDev))
 - SSAO2 now correctly takes diffuse texture alpha value into account ([CraigFeldspar](https://github.com/CraigFeldspar))
+- Sandbox will now load assets relatively path-ed to same folder([Kyle Belfort](https://github.com/belfortk))
```
bghgary: It's not obvious. You want to update the one in `dist/preview release` instead of this one. I'm not sure why we have a `what's new.md` at the root.

src/Misc/fileTools.ts
belfortk: Is there a prettier config or auto-formatter that I am missing?

what's new.md
belfortk: My bad. I missed that in Nick's comment.

belfortk: Updated!

src/Misc/fileTools.ts
belfortk: updated!

Issue #5100, opened 2018-09-10
GLTF serializer: Add bones and morph target animation support


deltakosh:
Follow-up of #4131

hobogenized:
Does this issue mean that BabylonJS currently does not support loading a skeleton/bones from GLTF scenes?

deltakosh:
Nope it means we do not export them from a babylon.js scene
From a loading standpoint we support EVERYTHING from a gltf file :)

Drigax:
Continuing from #6628, this PG:
https://playground.babylonjs.com/#T087A8#7

reproduces the issue.

Drigax:
I think I have a good plan of action here:
Looking at the current result of exporting the alien scene:
[alien_babylon_glTF2_export.zip](https://github.com/BabylonJS/Babylon.js/files/3866399/alien_babylon_glTF2_export.zip)

We can see that our current joint structure is preserved (as we iterate through all nodes in the babylon scene):
![Screenshot from 2019-11-19 14-43-47](https://user-images.githubusercontent.com/3952729/69193318-5053b480-0adb-11ea-8cb8-b91acd52d38c.png)

well, for the most part anyways...due to #7152 we lose all non-mesh leaf nodes in the scene: 
![image](https://user-images.githubusercontent.com/3952729/69193799-775eb600-0adc-11ea-82d6-a6b1504b0f35.png)

But this should be a trivial fix of removing this conditional we use to evaluate what nodes get exported when traversing the scene tree:
https://github.com/BabylonJS/Babylon.js/blob/v4.0.0/serializers/src/glTF/2.0/glTFExporter.ts#L1342

I'll check this change in as a part of this enhancement.

Now, back to implementation details, looks like we should be able to shim this functionality into createSceneAsync after we convert materials, and create our node map (mainly so that we already have a babylon scene node id to gltf node id to work with)

I should be able to add createSkinsAsync and createMorphTargetsAsync functions (Sticking with the asynchronous convention even though the export is performed synchronously?)

Only major concerns I have right now is in exporting the reverse bind matrix for my skin joints, I fear that LHC/RHC setting will cause some issues, but that a problem for Future Drigax™

deltakosh:
I wish Future Drigax™ good luck :D

Drigax:
[...]

PR #7299, opened 2019-12-11
Modify Animation Group Serialization to include all animated nodes in scene

Drigax:
Partially addressing #5100. This first fix was needed to be able to test the skin serialization in the babylon playground.

This change changes the node animation serialization logic to export animation tracks for all animated node transforms in scene.


serializers/src/glTF/2.0/glTFAnimation.ts
```diff
@@ -241,13 +240,13 @@ export class _GLTFAnimation {
                 for (let targetAnimation of animationGroup.targetedAnimations) {
                     let target = targetAnimation.target;
                     let animation = targetAnimation.animation;
-                    if (target instanceof Mesh || target.length === 1 && target[0] instanceof Mesh) { // TODO: Update to support bones
+                    if (target instanceof TransformNode || target.length === 1 && target[0] instanceof TransformNode) {
```
Drigax: @sebavan , I'm a bit confused about the second condition we check for, why do we want to check target.length?

deltakosh: because targetedAnimations can contain a target or an array of targets

PR #7302, opened 2019-12-11
Fix PG Zip

sebavan:


Vinc3r:
some assets seems to not be downloaded.

use case : open Cornell box example playground from https://doc.babylonjs.com/examples/ : https://www.babylonjs-playground.com/#G3GZA1#49 : 

image

It may happen because it's a `BABYLON.CubeTexture` and not a `BABYLON.Texture`?

Issue #7295, opened 2019-12-10
Only the first scene is rendered after multiple scenes are created


belfortk:
After adding a second scene to the engine and returning the second scene in my createScene function, only the first one is rendered

Bug repro on [playground](https://playground.babylonjs.com/#JMZX3H#6)

-Expected:
The new scene (the scene returned in the createScene function) to be rendered

Current Result:
only the first scene added to the engine is rendered

PR #7311, opened 2019-12-11
Playground will now render the returned scene from createScene() when there are multiple scenes added to engine

belfortk:
Fix for #[7295](https://github.com/BabylonJS/Babylon.js/issues/7295). 

update the scene used in engine.runRenderLoop callback from engine.scenes[0] to the one returned from createScene() in the playground script.


belfortk:
I feel like my implementation is not very DRY.

I tried to make it cleaner in this [commit](https://github.com/belfortk/Babylon.js/commit/d72365a7df26217fd89bef0c5372a4be42a22d80#diff-0f38524fed9cda05e0565578fd1ea1c6R201) but it fails to render the default scene script. Any pointers/ideas would be welcomed! =)

Playground/js/main.js
```diff
@@ -138,23 +138,43 @@ compileAndRun = function(parent, fpsLabel) {
...
-                var scene = engine.scenes[0];
-
-                if (scene.activeCamera || scene.activeCameras.length > 0) {
-                    scene.render();
-                }
-
-                fpsLabel.innerHTML = engine.getFps().toFixed() + " fps";
-            }.bind(this));
+            var sceneToRender;
+            if (scene.then) {
```
sebavan: No need to dupplicate so we ll check tomorrow together :-)

deltakosh: Well this is really important to consider that your code is public and must be clean and up to great standards.

belfortk:
Made a smarter PR [here](https://github.com/BabylonJS/Babylon.js/pull/7318)

PR #7338, opened 2019-12-16
Delete of the root node on GLB export

Bloadrick:
We have fixed this issue:  https://github.com/BabylonJS/Babylon.js/issues/6349

The root node on GLB export is successfuly removed for the export. 

I have 2 things to point out about this implementation:

- We are toggling the coordinate system  (x = !x), would you prefer to set an explicit value  (x=true) in case the user manually edit the 'useRightHandedSystem' ?

- The asynchronous export causes the scene to have its normals inverted during the export, would you prefer we made it synchronous with a loading screen ?

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -269,6 +269,7 @@
 - Playground will now render the returned scene from createScene() when there are multiple scenes added to engine ([Kyle Belfort](https://github.com/belfortk))
 - Fixed bug so Playground will now download .env texture files to ./textures in .zip  ([Kyle Belfort](https://github.com/belfortk))
 - It was not possible to change the gaze and laser color in VR ([#7323](https://github.com/BabylonJS/Babylon.js/issues/7323)) ([RaananW](https://github.com/RaananW/))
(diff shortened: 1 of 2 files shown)


bghgary:
I don't think this change works well. We should not be mucking with the scene content during export. If we have to, we at least need to reset it when the export is complete.

Here is what I believe should be implemented:

- Check [`_convertToRightHandedSystem`](https://github.com/BabylonJS/Babylon.js/blob/master/serializers/src/glTF/2.0/glTFExporter.ts#L280).
- Check the scene for a `__root__` node.
- If both `_convertToRightHandedSystem` is true and there is a `__root__` node, check if the transform for the node converts from right-handed to left-handed. If so, remove the __root__ node and set `_convertToRightHandedSystem` to false.
- If there is not, do what it does now.

You might have to make `_convertToRightHandedSystem` a variable passed to each function instead of a member field for this to work properly, if there are multiple root nodes.

noalak:
Gary's solution is implemented in PR https://github.com/BabylonJS/Babylon.js/pull/7344.

Issue #7239, opened 2019-12-02
WebXR camera positioning API


RaananW:
It should be possible to position the XR camera in any 3D position in space. 

The current state allows a developer to re-position the webxr container (the parent of the camera and the controllers), but in order to do that correctly the user must also know the height of the user and other factors. There is also an issue when querying the current camera's position and direction, as the camera itself will always return 0,0,0 as its local position due to the way the rig-cameras are working. 

This is a proposed API extension to the WebXR camera that will be able to position the container based on floor meshes, ignoring the user's current height. The camera should be the "single point of truth" when it comes to its own position.

The API will allow:

1) Simple (re)positioning of a user to a point of reference, with an optional floor / height map to use for height:

```javascript
function setPosition(pointInSpace: Vector3, pointOfReference?: Vector3 | Node, options? : {
  space: Space = Space.LOCAL,
  setHeightFromXRFrame: boolean = true,
  // further options TBD
}): void;
```

2) Get the camera's current global and local position using the XFrame information sent by XR (average of both eyes)
3) The ability to query the global position of a single eye
4) The ability to check the camera's current direction 
5) Active position-copy from native (non-XR) to XR cameras and vice versa

RaananW:
@nbduke - please comment if anything is missing here

nbduke:
So if I call `setPosition(point, Vector3.Zero(), { space: Space.LOCAL })`, is that just the same as `setPosition(point, undefined, { space: Space.WORLD })`?

RaananW:
@nbduke - Sorry for a late answer!

Yes, world coordinates will be including the container, local will change the camera alone. Internally, both will change the container but either relative to the current position or not.

RaananW:
Teleporting (position changes of the user) will be done using https://www.w3.org/TR/webxr/#xrreferencespace-interface 
the `setPOsition` function will create a new relative reference space.

RaananW:
merged and tested

stefanvanherwijnen:
@RaananW 
Using `xr.baseExperience.camera.leftCamera` and `xr.baseExperience.camera.rightCamera` does not seem to work anymore. Which is strange, because `xr.baseExperience.camera.rigCameras` returns an Array with two cameras, but `xr.baseExperience.camera.rigCameras.length` returns 1 :confused: .

https://www.babylonjs-playground.com/#LUYXFU

RaananW:
Hi @stefanvanherwijnen , just got back and looking into this.

[...]

PR #7343, opened 2019-12-17
WebXR camera API

RaananW:
Closing #7239 

This removes the container element in the webxr experience helper and adds support to setting and getting the camera's position and rotation (and view / world matrix)

Changes:

- experienceHelper.camera is the head position / base position of the user
- camera.position can be changed and re-set
- camera.rotationQuaternion can be changed and re-set
- The rest of the TargetCamera API works as expected
- the camera's position will be set when entering XR - the x and z position and the Y-Axis rotation will be taken from the non-xr camera 
- Rig camera do not have the camera as their parent, as they receive the full transformation and not relative to the head position.
- a new observable - `onInitialXRPoseSetObservable` will allow you to alter the position and rotationQuaternion on your own (for example when setting a height other than 0 as ground level)


deltakosh:
Some tests are failing :(

RaananW:
The two validation tests. are we running webxr visualisation tests?

RaananW:
@deltakosh - before merging - old playgrounds using the container will probably throw some errors in the console. The container is usually used when entering XR (async), so the playground will probably work correctly.

deltakosh:
Seems not a big deal

PR #7345, opened 2019-12-17
small fixes and features for Babylon Native

CedricGuillemet:
exit function
glue with document/window
fix attributes replacement bug

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -95,6 +95,8 @@ interface INativeEngine {
     clearDepth(depth: number): void;
     clearStencil(stencil: number): void;
 
+    exit(exitCode: number): void;
+
     getRenderWidth(): number;
     getRenderHeight(): number;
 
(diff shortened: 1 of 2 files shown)


src/Engines/nativeEngine.ts
```diff
@@ -95,6 +95,8 @@ interface INativeEngine {
     clearDepth(depth: number): void;
     clearStencil(stencil: number): void;
 
+    exit(exitCode: number): void;
```
bghgary: We discussed this and this will be removed and moved into the TestApp in BabylonNative.

src/Engines/nativeEngine.ts
```diff
@@ -301,6 +303,22 @@ export class NativeEngine extends Engine {
...
+     * Gets host window
+     * @returns the host window object
+     */
+    public getHostWindow(): Nullable<Window> {
+        return window;
+    }
+
+    public updateDynamicTexture(texture: Nullable<InternalTexture>, canvas: HTMLCanvasElement, invertY: boolean, premulAlpha: boolean = false, format?: number): void {
+        // TODO
+    }
```
bghgary: Not sure about this one. It has the HTML canvas which will not be available in native.

bghgary: Maybe we should use TexImageSource here and only support ImageData

CedricGuillemet:
Closing for now.

PR #7349, opened 2019-12-17
Add DirectionalLight.autoCalcShadowZBounds property

Popov72:
The min/max X/Y best fit bounds are calculated when `autoUpdateExtends` is `true` but not the min/max Z.

When `autoCalcShadowZBounds` is set to `true`, min/max Z best fit will also be calculated (the values being set to `shadowMinZ` / `shadowMaxZ`).

Note I had to switch `private` to `protected` for the `ShadowLight.shadowMin/MaxZ` properties because calling their setter leads to the projection matrix being recomputed, something we don't want because we are already in the function that computes this matrix when we set those properties!

This auto calculation allows better fit in Z, so better precision in the shadow map:

image image

Left: depth map with zmin/zmax values I set manually (that I thought were good...)
Right: depth map with auto-calculated zmin/zmax. You can see there's a better range of Z values

Also, maybe more importantly, it frees the user from setting the `shadowMinZ` / `shadowMaxZ` manually. Generally, those properties are left untouched, and so the camera near / far Z values are used instead (which is generally awful). When they are set, setting the best values can be difficult and will change when objects / light move.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -51,6 +51,7 @@
 - Allow setting of `BABYLON.Basis.JSModuleURL` and `BABYLON.Basis.WasmModuleURL`, for hosting the Basis transcoder locally ([JasonAyre])(https://github.com/jasonyre))
 - PNG support for browsers not supporting SVG ([RaananW](https://github.com/RaananW/))
 - Device orientation event permissions for iOS 13+ ([RaananW](https://github.com/RaananW/))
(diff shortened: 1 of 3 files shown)


deltakosh:
You rock buddy!

deltakosh:
This needs a doc update then :D

Popov72:
> This needs a doc update then :D

Ah, yes! But I'm a bit short of time those days as I would like to finish my project (or at least reach a milestone) before christmas, so I will update the docs beginning of next year (I have never done it before, so I think it will take a little time to setup - even if I know there's a doc explaining how to do it).

deltakosh:
Sure no rush..just keep it in mind

Issue #7352, opened 2019-12-18
Safari texture export is flipped Y


Drigax:
Issue was created on behalf of forum thread:
https://forum.babylonjs.com/t/textures-flipped-in-export-when-using-safari/7607/4

When exporting a scene in Safari, the textures are exported with inverted-Y.

This is reproducable using https://playground.babylonjs.com/#T79ZIN#2
- Expected result:
Mesh is round tripped through glTF with proper texture orientation ⬆
- Current result:
Mesh is round tripped with inverted texture Y ⬇

Drigax:
A bit of research appears to show that this issue is somewhat well known, and seems unaddressed. I may want to file a bug with Apple regarding this as well.

https://github.com/x3dom/x3dom/issues/937#issuecomment-464327561
https://github.com/pixijs/pixi.js/issues/2283
https://stackoverflow.com/questions/31082912/canvaselement-todataurl-for-webgl-canvases
https://www.html5gamedevs.com/topic/24109-canvas-screenshot-is-taken-upsidedown/

PR #7353, opened 2019-12-18
Fix Safari inverting exported textures

Drigax:
Fix for issue where Safari canvas is drawn Y down, workaround was discovered where premultiplying alpha causes correct canvas orientation. Added utility method to detect Safari browser to use this alternate engine initializing path.

Addressing #7352

### DIFF
--- a/serializers/src/glTF/2.0/glTFExporter.ts
+++ b/serializers/src/glTF/2.0/glTFExporter.ts
@@ -327,7 +327,7 @@ export class _Exporter {
     }
 
     /**
-     * Lazy load a local engine with premultiplied alpha set to false
+     * Lazy load a local engine
      */
     public _getLocalEngine(): Engine {
         if (!this._localEngine) {
@@ -335,11 +335,7 @@ export class _Exporter {
             localCanvas.id = "WriteCanvas";
             localCanvas.width = 2048;
             localCanvas.height = 2048;
-            if (Tools.IsSafari()) {
-                this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: true, preserveDrawingBuffer: true });
-            } else {
-                this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: false, preserveDrawingBuffer: true });
-            }
+            this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: Tools.IsSafari(), preserveDrawingBuffer: true });
             this._localEngine.setViewport(new Viewport(0, 0, 1, 1));
         }
 
--- a/serializers/src/glTF/2.0/glTFExporter.ts
+++ b/serializers/src/glTF/2.0/glTFExporter.ts
@@ -335,7 +335,11 @@ export class _Exporter {
             localCanvas.id = "WriteCanvas";
             localCanvas.width = 2048;
             localCanvas.height = 2048;
-            this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: false, preserveDrawingBuffer: true });
+            if (Tools.IsSafari()) {
+                this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: true, preserveDrawingBuffer: true });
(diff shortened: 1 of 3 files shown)


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -335,7 +335,11 @@ export class _Exporter {
             localCanvas.id = "WriteCanvas";
             localCanvas.width = 2048;
             localCanvas.height = 2048;
-            this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: false, preserveDrawingBuffer: true });
+            if (Tools.IsSafari()) {
```
Drigax: @deltakosh, I noticed in my local build that I'm not able to "see" the new Tools utility method I added, causing the build to fail here. Is there anything I need to do to make sure that this method is exported properly?

src/Misc/tools.ts
```diff
@@ -1215,6 +1215,14 @@ export class AsyncLoop {
             }
         }, callback);
     }
+
+    /**
+     * Utility function to detect if the current user agent is Safari
+     * @returns whether or not the current user agent is safari
+     */
+    public static IsSafari() : boolean {
+        return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
Drigax: kudos of some logic in the thingine that is used to detect certain bad desktop OSes. 😉

I opted to add this logic to Tools as well, but perhaps I should modify thinEngine to reuse the logic there instead?

serializers/src/glTF/2.0/glTFExporter.ts
deltakosh: Well yeah I know, you need to build the .d.ts first and then recompile but no worries the CI will take care of it

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -335,7 +335,11 @@ export class _Exporter {
             localCanvas.id = "WriteCanvas";
             localCanvas.width = 2048;
             localCanvas.height = 2048;
-            this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: false, preserveDrawingBuffer: true });
+            if (Tools.IsSafari()) {
+                this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: true, preserveDrawingBuffer: true });
```
deltakosh: can we just do something like
```
 this._localEngine = new Engine(localCanvas, true, { premultipliedAlpha: Tools.IsSafari(), preserveDrawingBuffer: true });
```

src/Misc/tools.ts
deltakosh: We need to keep thinengine with no dependency so let's keep it like that for now

deltakosh:
Also CI is failing big time ;)

serializers/src/glTF/2.0/glTFExporter.ts
Drigax: Looks like the CI is failing for the same reason:
https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=3030&view=logs&j=182f86de-37f5-5af4-c789-e92a7c8fa3dc&t=14c2bff8-fe23-5e9c-07aa-9b09ac155f63&l=169

I'm not sure what I'm doing wrong here

deltakosh: Are you sure this is the correct Tools (the one from bjs?)

Drigax: I made the change to src/Misc/Tools.ts...

Nevermind I'm just being silly. I accidentally put the change as a part of AsyncLoop

serializers/src/glTF/2.0/glTFExporter.ts
Drigax: sure, made the appropriate change.

PR #7356, opened 2019-12-18
Texture packer

Pryme8:
First PR for the initialization of a Texture Packer

TODO:
~~Enable the Power2 Layout and~Column Count Layout modes~~.
~~Enable the Extend Padding and Color Padding modes~~.
~~Add the ability to copy the content of a DynamicTextures instead of just Textures~~.
~~Create Function for Merging Similar Materials after compilation~~. -> this might be better to be left to the user.
~~Start Implementing PBR channels.~~
~~Export Method~~
~~Dispose Method~~
~~Load from JSON Method~~
~~Public method to update mesh lists to frame assignments~~
~~Finish Comments~~
~~Delete all Console.logs~~
~~Check spacing between all operators~~

~~Documentation~~
~~Example Scenes (Usage, Export, Load)~~
~~Performance Comparison Scene~~
~~What's New~~


src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+        //Make it an even padding Number.
+        if (this._paddingValue % 2 !== 0) {
+            this._paddingValue++;
+        }
+
+        /**
+        * Create the promise and then run through the materials on the meshes.
+        */
+        this.promise = new Promise ((resolve, reject) => {
+            console.log("Promise Start");
```
deltakosh: no console.log (ultimately I mean)

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+                                    (this.sets as any)[this.options.map[j]] = true;
+                                }
+
+                                if (this.options.disposeSources) {
+                                    this._disposeList.push(t);
+                                }
+                            }
+                        }
+
+                        if (done == this.meshes.length) {
```
deltakosh: ===

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+                            this._createFrames(resolve);
+                        }
+                    }
+                };
+
+                for (let i = 0; i < this.meshes.length; i++) {
+                    let mesh = this.meshes[i];
+                    let material: Nullable<Material> | Material = mesh.material;
```
deltakosh: should only be: Nullable<Material>

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+                    if (!material) {
+                        return new Error('Mesh has no Material Assigned!');
```
deltakosh:
```suggestion
                        return new Error('Mesh has no Material assigned!');
```

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+                        doneCheck((material as Material));
+                    });
+
+                }
+            }catch (e) {
+                return reject(e);
+            }
+        });
+
+    return;
```
deltakosh: we have several tabs issues

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+    }
+
+    /**
+    * Starts the package process
+    * @param resolve The promises resolution function
+    * @returns TexturePacker
+    */
+    private _createFrames(resolve: Function) {
```
deltakosh:
```suggestion
    private _createFrames(resolve: () => void) {
```

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+            dt.update();
+            (this.sets as any)[setName] = dt;
+        }
+
+        let baseSize = this.options.frameSize || 256;
+        let padding = this._paddingValue;
+        let tcs = baseSize + (2 * padding);
+
+        const done = () => {
+            console.log("Compilation Done");
```
deltakosh: no console

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+            let m = this.meshes[i];
+            let mat = m.material;
+
+            //Check if the material has the texture
+            //Create a temporary canvas the same size as 1 frame
+            //Then apply the texture to the center and the 8 offsets
+            //Copy the Context and place in the correct frame on the DT
+
+            for (let j = 0; j < sKeys.length; j++) {
+                let tempTexture = new DynamicTexture('temp', tcs, this.scene, true);
```
deltakosh: needs to be dispose and I wonder if we can reuse it (and only rescale it))

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,551 @@
...
+
+        const done = () => {
+            console.log("Compilation Done");
+            this._calculateMeshUVFrames(baseSize, padding, dtSize, dtUnits, this.options.updateInputMeshes || false);
+        };
+
+        //Update the Textures
+        for (let i = 0; i < meshLength; i++) {
+            let m = this.meshes[i];
+            let mat = m.material;
```
deltakosh: What if they share the same material, are you not doing it multiple times?

Pryme8: No, because the texture then already exists on the sets and so I just make sure the mesh gets set to the correct frames for that material.

Pryme8:
I didn't mean to start a review...

deltakosh:
We have some check issues :)

src/Materials/Textures/Packer/frame.ts
```diff
@@ -0,0 +1,57 @@
...
+    /**
+	 * The Frames offset
+	 */
+    offset: Vector2;
+
+}
+
+/**
+ * This is a support class for frame Data on texture packer sets.
+ * @see #TODO ADD THIS
```
deltakosh: warning here

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+import { TexturePackerFrame } from "./frame";
+import { ITexturePackerJSON } from "./packerLoader";
+
+/**
+ * Defines the basic options interface of a TexturePacker
+ */
+export interface ITexturePackerOptions{
+
+    /**
+	 * Custom targets for the channels of a texture packer.  Default is all the channels of the Standard Material
```
deltakosh: tabs seems off in the comments

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+	 */
+    map?: string[];
+
+    /**
+	 * the UV input targets, as a single value for all meshes or an array of values that matches the mesh count.  Defaults to VertexBuffer.UVKind
+	 */
+    uvsIn?: string; // | string[];
```
deltakosh: please remove not used comments

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+                    }
+                };
+
+                for (let i = 0; i < this.meshes.length; i++) {
+
+                    let mesh = this.meshes[i];
+                    let material: Nullable< Material > = mesh.material;
+
+                    if (!material) {
+                        return new Error('Mesh has no Material assigned!');
```
deltakosh: maybe we can just ignore it instead

src/Materials/Textures/Packer/frame.ts
Pryme8: Opps good catch!

Pryme8:
Ill make those fixes as soon as I get back from dropping my wife off at work!

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+        mesh.setVerticesData(this.options.uvsOut || VertexBuffer.UVKind, uvOut);
+    }
+
+    /**
+    * Updates a Meshs materials to use the texture packer channels
+    * @param m is the mesh to target
+    * @param force all channels on the packer to be set.
+    * @returns void
+    */
+    private _updateTextureRefrences(m: AbstractMesh, force: boolean = false): void {
```
deltakosh: Typo

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+        }
+
+        mesh.setVerticesData(this.options.uvsOut || VertexBuffer.UVKind, uvOut);
+    }
+
+    /**
+    * Updates a Meshs materials to use the texture packer channels
+    * @param m is the mesh to target
+    * @param force all channels on the packer to be set.
+    * @returns void
```
deltakosh: no need if no return value

Pryme8: weird, I fixed that last night, let me see if I did not push the most recent version.

src/Materials/Textures/Packer/packerLoader.ts
```diff
@@ -0,0 +1,71 @@
...
+    /**
+	 * The frame data of the Packer
+	 */
+    frames: Array<number>;
+
+}
+
+/**
+ * This is a support class to load JSON texture packer data.
+ * @see #TODO ADD THIS
```
deltakosh: todo

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+    /**
+	 * If in SUBUV_COLOR padding mode what color to use.
+	 */
+    paddingColor?: Color3 | Color4;
+
+}
+
+/**
+ * This is a support class that generates a series of packed texture sets.
+ * @see #TODO ADD THIS
```
deltakosh: todo

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+    public promise: Promise< TexturePacker | string >;
+
+    /** The Container object for the channel sets that are generated */
+    public sets: object;
+
+    /** The Container array for the frames that are generated */
+    public frames: TexturePackerFrame[];
+
+    /** The List of textures to purge from memory after compilation */
+    private _disposeList: (Texture | DynamicTexture)[];
```
deltakosh: I do not see where this is used

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+    }
+
+    /**
+    * Calculates the UV data for the frames.
+    * @param baseSize the base frameSize
+    * @param padding the base frame padding
+    * @param dtSize size of the Dynamic Texture for that channel
+    * @param dtUnits is 1/dtSize
+    * @param update flag to update the input meshes
+    * @returns Void
```
deltakosh: no need

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,761 @@
...
+            this.options.paddingColor = this.options.paddingColor || new Color4(0, 0, 0, 1.0);
+        }
+
+        this.sets = {};
+        this.frames = [];
+
+        /**
+        * Create the promise and then run through the materials on the meshes.
+        */
+        this.promise = new Promise ((resolve, reject) => {
```
deltakosh: I wonder if instead of having a then function which is basically a way to give resolve/error to this promise we should not have a texturePacker.processAsync() which will return a promise. So the user can use then or catch or anything related to a promise like using Promise.all if he/she has multiple packers

src/Materials/Textures/Packer/packer.ts
Pryme8: hmmm I guess I ended up not using it, in my cleanup method.

Let me get rid of it and change that to just a expected count, because that is needed

src/Materials/Textures/Packer/packerLoader.ts
```diff
@@ -0,0 +1,71 @@
...
+export class TexturePackerLoader{
+    /**
+    * Initializes a XML request for the JSON file.
+    * @param jsonURL The URL of the texturePackage.json
+    * @param scene that the packer is scoped to.
+    * @returns TexturePacker
+    */
+    constructor(jsonURL: string, scene: Scene) {
+        if (!scene || !jsonURL || typeof jsonURL !== 'string') {
+            return false;
```
deltakosh: you cannot return false here (I'm surprised TS let you do it). A constructor returns nothing

src/Materials/Textures/Packer/packer.ts
Pryme8: sure?  If you think that is better I kinda thought the promise let it resolve in async like a doing a setTimeout hack.

src/Materials/Textures/Packer/packerLoader.ts
Pryme8: oh crap, that is weird.

src/Materials/Textures/Packer/packerLoader.ts
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                };
+                xml.open("GET", jsonURL);
+                xml.send();
+                }catch (e) {
+                    error(e);
+                }
+            });
+
+            return pack;
```
deltakosh: Not sure to understand how the user is supposed to get the pack variable.

Overall I feel like we do not want this class.

Pryme8: I was accessing it in the "Then" function to make sure it was compiled.

Pryme8: What would be an alternative off the top of your head?

If we change it from a promise then to a async Promise then I will have a onComplete method correct?

Pryme8:
I will have all this fixed today.

src/Materials/Textures/Packer/packer.ts
deltakosh: I still want the promise but it should not be set in the constructor and "awaited" by a fake then function. It should instead be created and returned by a processAsync function . In a nutshell:
- remove the then function
- add a processAsync function and move the promise there
- return the promise

src/Materials/Textures/Packer/packerLoader.ts
deltakosh: The alternative it to use the loadFromJSON directly and let the user deal with where the data comes from

src/Materials/Textures/Packer/packer.ts
Pryme8: Ok I'll have to look up how to do that, should be EZPZ though!

So the user is going to have to say when to start packing basically?

src/Materials/Textures/Packer/packerLoader.ts
Pryme8:
```
pack = new BABYLON.TexturePackerLoader('./textures/TestPack_texurePackage.json', scene)    
    pack.then(()=>{
        console.log("Pack Parsed")
        console.log("texturePack", pack)        
        let packMat = new BABYLON.StandardMaterial('packMat', scene )
        output.material = packMat;
        pack.setMeshToFrame(output, 2, true)
    }, (error)=>{
        console.log(error)
    }) 
```

Was working on accessing the pack variable, but maybe I am doing this incorrect though from what you are saying.  Let me review everything you have said and make sure Im not being dumb.

src/Materials/Textures/Packer/packer.ts
deltakosh: Yes and I usually prefer that. Creating an object should not start any heavy duty process:)

Pryme8: I will have that done this morning.

Pryme8: I will have that done this morning.

Pryme8: So basically the user will create the TexturePacker, but will have to manually call `pack.processAsync()` which is where most the stuff I am doing in the constructor will get moved?

That will also mean then I get rid of the Loader class like mentioned and then just have them access that class directly which I can then add an argument to have that start its async Compilation as well.

Am I understanding you correctly?

deltakosh: totally

Pryme8: hmm I've never used an await before, let me do some reading on this.

deltakosh: no need for await (And technically we cannot use them as this is not supported on IE and we have no polyfill for it)

Pryme8: Ok cool! I was kinda in full research mode for a second and then was just about to ask about compatibility.

src/Materials/Textures/Packer/packer.ts
Pryme8: I think this is resolved, let me test it really quick.

Pryme8: getting this resolved now, then will do lint fixes

Pryme8:
Doing a build then pushing the fixes.

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,827 @@
...
+    }
+
+    /**
+    * Starts the async promise to compile the texture packer.
+    * @param success callback for the load promise
+    * @param error callback for the load promise
+    */
+    public processAsync(success = (): void => {}, error = (err : string): void => {}): void {
+        setTimeout(() => {
+            this.promise = new Promise ((resolve, reject) => {
```
deltakosh: Here is what I had in mind:

public processAsync(): Promise<void> {
    return new Promise ((resolve, reject) => {
...
})
}

deltakosh: the user can then do:
```
pack.processAsync().then(() => {
// my stuff
}).catch(e => {
// handling error
});
```

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,827 @@
...
+    }
+
+    /**
+    * Public method to load a texturePacker JSON file.
+    * @param url location of the JSON file
+    * @param success callback for the load promise
+    * @param error callback for the load promise
+    */
+    public updateFromJSON(url: string , success = (): void => {}, error = (err : string): void => {}): void {
+        setTimeout(() => {
```
deltakosh: No need for all of that. the updateFRomJSON must except the json data 
It is not up to us to do the xhr

Pryme8: ok cool.

src/Materials/Textures/Packer/packer.ts
Pryme8: so then they do what like
```
let packPromise = pack.processAsync()

packPromise.then(success, error)
```
?

Pryme8: and then I can just drop the this.promise

deltakosh: packPromise.then(() => success()).catch(() => error())

Pryme8: I did not see your comment my bad!

Pryme8: I did not see your comment my bad!

src/Materials/Textures/Packer/packer.ts
```diff
@@ -764,64 +753,40 @@ export class TexturePacker{
...
     * Public method to load a texturePacker JSON file.
-    * @param url location of the JSON file
-    * @param success callback for the load promise
-    * @param error callback for the load promise
+    * @param data of the JSON file in string format.
+    * @returns Promise<void>
     */
-    public updateFromJSON(url: string , success = (): void => {}, error = (err : string): void => {}): void {
-        setTimeout(() => {
+    public updateFromJSON(data: string): Promise<void> {
```
deltakosh: This does not need to be a promise as nothing is asynchronous

Pryme8:
For sure I'll fix that if get back to the house tonight.

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+    */
+    layout?: number;
+
+    /**
+    * number of columns if using custom column count layout(2).  This defaults to 4.
+    */
+    colnum?: number;
+
+    /**
+	 * flag to update the input meshes to the new packed texture after compilation. Defaults to true.
```
deltakosh: Still tabs issue in this file :(

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+}
+
+/**
+* This is a support class that generates a series of packed texture sets.
+* @see https://doc.babylonjs.com/babylon101/materials
+*/
+export class TexturePacker{
+
+    /** Packer Layout Constant 0 */
+    public static readonly LAYOUT_STRIP = Constants.LAYOUT_STRIP;
```
deltakosh: Actually I give that a thought and we do not need to expose the constants in the constant file as they are only used by the packer. So we can keep the declartion here (with the direct value instead of a link to Constants)

Sorry buddy about that last one

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+            this.options.paddingColor = this.options.paddingColor || new Color4(0, 0, 0, 1.0);
+        }
+
+        this.sets = {};
+        this.frames = [];
+
+        /**
+        * Create the promise and then run through the materials on the meshes.
+        */
+        this.promise = null;
```
deltakosh: Do we still need a promise variable?

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+
+                    if (this.options.fillBlanks) {
+                        tcx.fillStyle = (this.options.customFillColor as string);
+                    }
+
+                    tcx.fillRect(0, 0, tcs, tcs);
+
+                    updateDt();
+
+                }else {
```
deltakosh: space

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+
+        mesh.setVerticesData(this.options.uvsOut || VertexBuffer.UVKind, uvOut);
+    }
+
+    /**
+    * Updates a Meshes materials to use the texture packer channels
+    * @param m is the mesh to target
+    * @param force all channels on the packer to be set.
+    */
+    private _updateTextureRefrences(m: AbstractMesh, force: boolean = false): void {
```
deltakosh: typo on the function name

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+    /**
+    * Starts the async promise to compile the texture packer.
+    * @returns Promise<void>
+    */
+    public processAsync(): Promise<void> {
+            return new Promise ((resolve, reject) => {
+                try {
+                    if (this.meshes.length === 0) {
+                        //Must be a JSON load!
+                        resolve();
```
deltakosh: you must return then after the resolve

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+                                return this._createFrames(resolve);
+                            }
+                            continue;
+                        }
+
+                        material.forceCompilationAsync(mesh).then(() => {
+                            doneCheck((material as Material));
+                        });
+                    }
+                }catch (e) {
```
deltakosh: space before catch

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+                    return reject(e);
+                }
+            });
+    }
+
+    /**
+    * Updates a Meshes materials to use the texture packer channels
+    * @param m is the mesh to target
```
deltakosh: wrong comment

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,792 @@
...
+        });
+    }
+
+    /**
+    * Public method to load a texturePacker JSON file.
+    * @param data of the JSON file in string format.
+    * @returns Promise<void>
+    */
+    public updateFromJSON(data: string): Promise<void> {
+        return new Promise ((resolve, reject) => {
```
deltakosh: no need for a promse at all
code is synchronous

src/Materials/Textures/Packer/packer.ts
Pryme8: That is really weird my lint is not catching that....

src/Materials/Textures/Packer/packer.ts
Pryme8: So is there something wrong with my lint process?  I feel like stuff like that should not be coming up.

Pryme8: So is there something wrong with my lint process?  I feel like stuff like that should not be coming up.

Pryme8:
Something odd happend to my last commit, things I did were not propigated over.  Sorry to make you review that last commit, it was not correct.  

I will resubmit the correct one ASAP.

src/Materials/Textures/Packer/packer.ts
Pryme8: Ok this is weird, all this was deleted on my last revision.  I must have done something wrong and reverted changes somehow...

src/Materials/Textures/Packer/packer.ts
Pryme8: so return promise.then?

deltakosh: `if (this.meshes.length === 0) {
                        //Must be a JSON load!
                        resolve();
return;
}
`

Pryme8:
Still need to do the return after resolve then that should be it.  I'll do that before I go to bed tonight.

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,783 @@
...
+            _a.remove();
+        });
+    }
+
+    /**
+    * Public method to load a texturePacker JSON file.
+    * @param data of the JSON file in string format.
+    */
+    public updateFromJSON(data: string): void {
+            try {
```
deltakosh: tab issues

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,783 @@
...
+                this.frames.push(frame);
+                }
+
+                let channels = Object.keys(parsedData.sets);
+
+                for (let i = 0; i < channels.length; i++) {
+                    let _t = new Texture(parsedData.sets[channels[i]], this.scene, false, false);
+                    (this.sets as any)[channels[i]] = _t;
+                }
+            } catch (e) {
```
deltakosh: we should not swallow errors

src/Materials/Textures/Packer/packer.ts
```diff
@@ -0,0 +1,783 @@
...
+            (this.sets as any)[channel].dispose();
+        }
+    }
+
+    /**
+    * Starts the download process for all the channels converting them to base64 data and embedding it all in a JSON file.
+    * @param imageType is the image type to use.
+    * @param quality of the image if downloading as jpeg, Ranges from >0 to 1.
+    */
+    public download(imageType: string = 'png', quality: number = 1): void {
```
deltakosh: This function should return the download promise so user can take action when download is done

src/Materials/Textures/Packer/packer.ts
Pryme8: so return e?

src/Materials/Textures/Packer/packer.ts
Pryme8: will do

src/Materials/Textures/Packer/packer.ts
deltakosh: nope just not catch

Pryme8: or do I use the logger?

src/Materials/Textures/Packer/packer.ts
Pryme8: actually thinking about this one, I want control of what happens to the then.

Otherwise I would have to expect the user to pack it into the JSON format that I have setup.

Pryme8: Maybe not even use a promise, but go ahead and use a setTimeout hack?

deltakosh: Well this works for now..Will see if more is required when people get their hands on it :)

deltakosh:
Congrats buddy!!!

Now last step is to provide examples in the doc :)

Issue #7290, opened 2019-12-10
Move Independent WebXR features to the feature manager


RaananW:
Due to the unstable nature of WebXR, it will be good to move all currently implemented features to the features manager in order to stay backwards-compatible but still move fast with the development of the specs.

Features that can move:

* Teleportation (separated to gamepad, hands, mobile, if possible)
* Gaze and Pointer system (supporting all device inputs)

Features manager commited here - https://github.com/BabylonJS/Babylon.js/pull/7288

PR #7361, opened 2019-12-19
WebXR Motion Controllers 

RaananW:
* Motion controllers (the VR controllers) are now separated from the gamepad class
* No more pose-enabled VR controllers
* Standard schema was used using the profile registry (https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/registry/profiles)
* Supported controllers: Generic Trigger, Microsoft Mixed Reality, Oculus (touch 1 and quest), Vive
* If the controller is not supported the generic model will be loaded
* Model classes are loaded using imports, so specific controllers can be addressed if the developer wants. The generic controller is always included.
* WebXR Input doesn't have and relations to the WebXR Experience helper so it can be used independently
* Now using the new xrInput.profiles to detect the controller type (instead of gamepad.id which is deprecated (an empty string)
* (temporarily) removed gaze support in the vr helper when XR is used. This will return when teleportation will move to the feature manager
* xr input has now options object that can override default behavior (at the moment can be used to prevent loading the mesh)


src/Cameras/VR/vrExperienceHelper.ts
```diff
@@ -844,18 +844,18 @@ export class VRExperienceHelper {
...
-
-                                if (controller.inputSource.handedness === "right" || (this._leftController && this._leftController.webVRController != webVRController)) {
-                                    this._rightController = localController;
-                                } else {
-                                    this._leftController = localController;
-                                }
-
-                                this._tryEnableInteractionOnController(localController);
-                            }
+                            // var webVRController = controller.gamepadController;
```
deltakosh: Comments?

src/Cameras/XR/motionController/webXRGenericMotionController.ts
```diff
@@ -0,0 +1,80 @@
...
+            "buttons": [
+                "xr-standard-trigger"
+            ],
+            "axes": []
+        }
+    }
+
+};
+
+// TODO support all generic models with xr-standard mapping at:
```
deltakosh: Can we capture that as an issue?

src/Cameras/VR/vrExperienceHelper.ts
RaananW: Commented on that in the description - (temporarily) removed gaze support in the vr helper when XR is used. This will return when teleportation will move to the feature manager

src/Cameras/XR/motionController/webXRGenericMotionController.ts
RaananW: Already did, #7362

src/Cameras/VR/vrExperienceHelper.ts
deltakosh: Makes sense!

RaananW: (when this is closed - https://github.com/BabylonJS/Babylon.js/issues/7290)
The main issue is that the webxr motion controller is no longer a gamepad object, so the old gaze suddenly doesnt support it.

stefanvanherwijnen:
@RaananW 
Is there a way to completely hide the pointer? `controller.pointer.dispose()` somewhat works, but the 'dot' (the end of the ray) does not seem to be removed.
https://www.babylonjs-playground.com/#KEWJFN

RaananW:
I am currently working on moving the teleportation and pointer features to the features manager so you can disable them whenever you want. There is currently no easy way of doing that. As a workaround you can use this - 

https://www.babylonjs-playground.com/#KEWJFN#1

RaananW:
@stefanvanherwijnen - the latest PR (https://github.com/BabylonJS/Babylon.js/pull/7393) will show you how to detach or attach the feature. Once merged, you will have both features in the default experience helper.
Full doc is coming till the end of the month, until then use the code doc or just ask (forum, here) if you need any help.

stefanvanherwijnen:
Thanks, great work.

PR #7373, opened 2019-12-24
Transparent meshes are not displayed when calling geometryBufferRenderer customRenderFunction before rendering the scene

ladevieq:
In my project since babylon 4.1 my transparent instances were not
rendered.

After some investigation if found that i was using a rendertarget that
renders my submeshes with this function.

As this function doesn't render the transparent meshes, those meshes
`_internalAbstractMeshDataInfo._isActiveIntermediate` property keep
containing true.

So when rendering the next frame, during `Scene._evaluteActiveMeshes`
those transparent instances are considered as style active when calling
`InstancedMesh._activate`.

This cause the meshes to not being dispatched in `RenderingGroups` and not
being rendered.

This fix work but you may take a closer look at the reason why only
alpha and opaque meshes were initially rendered.

If this solution is not optimal, it is also possible to do that, it is also possible to check and disable each active states in `Mesh.render` but that will set the mesh as inactive the first time the scene is rendered through the rendertarget.

```JS
if (this._internalAbstractMeshDataInfo._isActiveIntermediate) {
    this._internalAbstractMeshDataInfo._isActiveIntermediate = false;
}
if (this._internalAbstractMeshDataInfo._isActive) {
    this._internalAbstractMeshDataInfo._isActive = false;
}
```

### DIFF
--- a/src/Rendering/geometryBufferRenderer.ts
+++ b/src/Rendering/geometryBufferRenderer.ts
@@ -431,6 +431,10 @@ export class GeometryBufferRenderer {
             for (index = 0; index < alphaTestSubMeshes.length; index++) {
                 renderSubMesh(alphaTestSubMeshes.data[index]);
             }
+
+            for (index = 0; index < transparentSubMeshes.length; index++) {
+                renderSubMesh(transparentSubMeshes.data[index]);
+            }
         };
     }
 


deltakosh:
Looks ok to me

deltakosh:
Just to understand: why do you need transparent objects in the geometry buffer? what is your use case?

ladevieq:
@deltakosh There is no real reason for this.
The main problem is that when my instances material become a transparent one, those instances are not being rendered anymore.
My fix seems to be a workaround, nothing more.

deltakosh:
Can you share a Plyaground reproducing your issue? I need to understand the relationship between your change and the original issue ;)

ladevieq:
@deltakosh So here is the [playground](https://playground.babylonjs.com/#ILUEHM#5)

Took me a will to understand which high level components are causing this behavior.

It the combination of the `SSAORenderPipeline2`, event if it is not attached to the camera, and a `HighlighLayer`.

src/Rendering/geometryBufferRenderer.ts
```diff
@@ -431,6 +431,10 @@ export class GeometryBufferRenderer {
             for (index = 0; index < alphaTestSubMeshes.length; index++) {
                 renderSubMesh(alphaTestSubMeshes.data[index]);
             }
+
```
deltakosh: I think this is ok but should not be on by default
So I would suggest adding a property to enable it

Thoughts?

deltakosh:
You know what I will add the boolean for you :) So I can merge your PR!

PR #7378, opened 2020-01-02
ammoJSPlugin fix

bobalazek:
I'm currently trying to implement ammo into my app. Experiencing some issues, as it's not working right with typescript and webpack.
If I try like in the documentation (https://doc.babylonjs.com/features/es6_support#ammo):
```javascript
import * as Ammo from 'ammo.js';
...
scene.enablePhysics(
  new BABYLON.Vector3(0, -9.82, 0),
  new BABYLON.AmmoJSPlugin(true, Ammo)
);
```
I get: `babylon.js:16 Uncaught TypeError: this.bjsAMMO.btSoftBodyRigidBodyCollisionConfiguration is not a constructor`

That is then fixed if I replace `new BABYLON.AmmoJSPlugin(true, Ammo)` with `new BABYLON.AmmoJSPlugin(true, Ammo.default())`, but then I get a `babylon.js:16 Uncaught ReferenceError: Ammo is not defined`, so that PR should fix this issue.

Still need to actually check if that will fix everything, as I haven't tested it yet. Just assuming it should work.

### DIFF
--- a/src/Physics/Plugins/ammoJSPlugin.ts
+++ b/src/Physics/Plugins/ammoJSPlugin.ts
@@ -413,7 +413,7 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
             if (impostor.soft) {
                 colShape.get_m_cfg().set_collisions(0x11);
                 colShape.get_m_cfg().set_kDP(impostor.getParam("damping"));
-                Ammo.castObject(colShape, Ammo.btCollisionObject).getCollisionShape().setMargin(impostor.getParam("margin"));
+                this.bjsAMMO.castObject(colShape, this.bjsAMMO.btCollisionObject).getCollisionShape().setMargin(impostor.getParam("margin"));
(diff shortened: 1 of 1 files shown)


sebavan:
This is not the right fix as the rest of the code looks correct, could you send a repro of the issue on the forum and I ll have a look at it, the import should work.

bobalazek:
Will create a repro, but why is then `new this.bjsAMMO. ...` used in the constructor, but not in other methods?
https://github.com/bobalazek/Babylon.js/blob/master/src/Physics/Plugins/ammoJSPlugin.ts#L80-L99

sebavan:
Yup I can see also direct ref to ammo which should not be there cause only bjsAMMO should be use to ensure that the injection works.

I will create an issue for this one.

sebavan:
Ok so you PR is totally ok.... I guess the vacation did not help.... Thanks a lot. I thought it was all going to AMMO instead of bjsAmmo...

Such a rough start :-)

bobalazek:
Hah, no worries! :D

PR #7400, opened 2020-01-08
CSM improvement with min/max depth redux

Popov72:
Implements the first step of SDSM (http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.360.3911&rep=rep1&type=pdf and https://software.intel.com/en-us/articles/sample-distribution-shadow-maps).

It works by computing the min and max values from the depth buffer and using them to tightly split the frustum.

For faster performances, the reduction is done on the GPU, as done here: https://github.com/TheRealMJP/Shadows.

It can greatly enhance shadow quality:

image

Same settings for both sides, except `autoCalcDepthBounds=true` on the right.


Popov72:
It's not really possible to do everything on the GPU, see discussion here by MJP: https://mynameismjp.wordpress.com/2013/09/10/shadow-maps/, section "GPU-DRIVEN CASCADE SETUP AND SCENE SUBMISSION".

Even if it's now from some years ago, I don't think we can do it in WebGL, it's quite involved and use compute shaders. Also, I'm not sure it could easily fit in the Babylonjs pipeline.

I'm going to try the 1-frame latency readback.

Regarding:
> and an options to prevent rendering on every frame

Do you mean preventing the min/max redux on every frame? I think I can do it easily through a call to the `refreshRate` function of the depth RTT, but it may introduce rendering artifacts (like the 1-frame delay readback).

sebavan:
Yup exactly what I mean, exposing the Refresh rate :-)

And Yes, agree it might be for later for the read back.

PR #7401, opened 2020-01-08
Fix black screen when presenting in multiview

RaananW:
Small changes made between beta 15 and 16 prevented multiview from presenting at all.

issue - https://forum.babylonjs.com/t/oculus-quest-with-scene-createdefaultvrexperience-with-usemultiview-fails/7917

### DIFF
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -3094,7 +3094,7 @@ export class ThinEngine {
             return this._gl.TEXTURE_CUBE_MAP;
         } else if (texture.is3D) {
             return this._gl.TEXTURE_3D;
-        } else if (texture.is2DArray) {
+        } else if (texture.is2DArray || texture.isMultiview) {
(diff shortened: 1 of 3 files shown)


src/Engines/thinEngine.ts
```diff
@@ -3627,7 +3627,7 @@ export class ThinEngine {
         this._activeChannel = channel;
         const target = this._getTextureTarget(internalTexture);
         if (needToBind) {
-            this._bindTextureDirectly(target, internalTexture, isPartOfTextureArray);
+            this._bindTextureDirectly((internalTexture && internalTexture.isMultiview) ? this._gl.TEXTURE_2D_ARRAY : target, internalTexture, isPartOfTextureArray);
```
deltakosh: Why not doing it inside the getTextureTarget?

RaananW: It was there before, but it makes sense :-)

let me fix that

PR #7404, opened 2020-01-09
restore Mesh.clone return type

barroij:
In current beta version `Mesh.clone` return type is `Nullable<AbstractMesh>` whereas the "true" return type is `Mesh`

Having `clone` returning a nullable causes all the code using it to be updated to prevent typescript complaining about possibly null stuff where it is actually never null.


barroij:
@deltakosh FYI and happy new year !

deltakosh:
Happy new year :) can you run npm run build to fix the linting issue?

deltakosh:
Thanks!

PR #7423, opened 2020-01-10
Fix #7418 GLTF Loader and finishes samplers refacto.

### LINKED ISSUE #7418: Some GLB textures render with artifacts
# Bugs
Repro playground: https://www.babylonjs-playground.com/#XYZVIV

If you load that playground, you'll see what looks like Moire pattern on the bear model. It didn't use to look this way. If you point the playground at v4.0.3, for example, the pattern disappears.

I can't say for certain when this was introduced, but I know that the model rendered normally as of 4.1.0-alpha.27.

sebavan:


sebavan:
Fix #7418

PR #7428, opened 2020-01-12
Added support of Screen Space Reflection Post-Process.

julien-moreau:
Also set renderTransparentMeshes on GeometryBufferRenderer to true by default.

### DIFF
--- a/src/Shaders/geometry.fragment.fx
+++ b/src/Shaders/geometry.fragment.fx
@@ -1,13 +1,23 @@
 #extension GL_EXT_draw_buffers : require
 
+#if defined(BUMP) || !defined(NORMAL)
+#extension GL_OES_standard_derivatives : enable
+#endif
+
 precision highp float;
 precision highp int;
 
+#ifdef BUMP
+varying mat4 vWorldView;
+varying vec3 vNormalW;
+#else
 varying vec3 vNormalV;
+#endif
+
 varying vec4 vViewPos;
 
-#ifdef POSITION
-varying vec3 vPosition;
+#if defined(POSITION) || defined(BUMP)
+varying vec3 vPositionW;
 #endif
 
 #ifdef VELOCITY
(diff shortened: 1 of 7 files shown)


src/Shaders/geometry.fragment.fx
```diff
@@ -1,40 +1,70 @@
...
+    #ifdef ALPHATEST
 	if (texture2D(diffuseSampler, vUV).a < 0.4)
 		discard;
-#endif
+    #endif
 
     gl_FragData[0] = vec4(vViewPos.z / vViewPos.w, 0.0, 0.0, 1.0);
     //color0 = vec4(vViewPos.z / vViewPos.w, 0.0, 0.0, 1.0);
+
+    #ifdef BUMP
```
julien-moreau: @deltakosh to be discussed, do you agree that the Geometry Buffer Renderer should take bump textures into account when computing the normal pass ?

src/Shaders/geometry.fragment.fx
```diff
@@ -46,4 +76,18 @@ void main() {
...
     gl_FragData[VELOCITY_INDEX] = vec4(velocity, 0.0, 1.0);
     #endif
+
+    #ifdef REFLECTIVITY
+        #ifdef HAS_SPECULAR
+            // Specular
+            vec4 reflectivity = texture2D(reflectivitySampler, vReflectivityUV);
+        #elif HAS_REFLECTIVITY
+            // Reflectivity
+            vec4 reflectivity = vec4(texture2D(reflectivitySampler, vReflectivityUV).rgb, 1.0);
```
julien-moreau: PBR metallic/roughness/etc. not taken into account for now. The renderer simply takes the reflectivity map and doesn't compute any metallic/roughness function at the moment.

src/Shaders/geometry.fragment.fx
deltakosh: I think so

PR #7429, opened 2020-01-12
Add the RenderTargetTexture.getCustomRenderList user function

Popov72:
The final goal is to allow the user to perform per cascade culling in CSM if he ever wants to (Babylon.js does not provide per-cascade culling for the time being).

Changes to achieve that:
* `CascadedShadowGenerator`: compute all the matrices at once before the RTT is starting work. It will allow the user to get the view / project / transform / ... matrices by calling the corresponding methods on `CascadedShadowGenerator` when doing the culling and be sure to have up to date matrices
* `RenderTargetTexture`: added a public `getCustomRenderList` function pointer. This method is called before each rendering so that it is possible for the user to return the list of meshes to use for the rendering of the current texture (when in 2DArray mode, the `layer` index is passed, else `face` index is passed).

The expected usage for per-cascade culling is:
```typescript
let rtt = shadowGenerator.getShadowMap()!;

rtt.getCustomRenderList = (layer: number, renderList: Nullable<Immutable<Array<AbstractMesh>>>) : Nullable<Array<AbstractMesh>> => {
    let meshList = [];
    // do the culling for the cascade with index 'layer' by using the
    // getCascadeViewMatrix(layer), getCSMTransformMatrix(layer), getCascadeMinExtents(layer), etc
    return meshList;
};
```

If it's ok for you, I will update the CSM doc with this expected usage.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -56,6 +56,7 @@
 - Added `CascadedShadowGenerator.autoCalcDepthBounds` to improve the shadow quality rendering ([Popov72](https://github.com/Popov72))
 - Improved cascade blending in CSM shadow technique ([Popov72](https://github.com/Popov72))
 - Speed optimization when cascade blending is not used in CSM shadow technique ([Popov72](https://github.com/Popov72))
+- Added `RenderTargetTexture.getCustomRenderList` to overload the render list at rendering time (and possibly for each layer (2DArray) / face (Cube)) ([Popov72](https://github.com/Popov72))
 
 ### Engine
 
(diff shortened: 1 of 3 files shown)


sebavan:
@deltakosh will be back tomorrow to approve this one, I am ok with it but wonder if we should make it part of the RTT or only in the shadows ?

Popov72:
How would it be possible to do it without modifying `RenderTargetTexture`?

There's currently no way to use a different list of meshes for each rendering when using a 2DArray / Cube.

sebavan:
Agree but the list could be stored on the generator ? I am just wondering :-) to try limiting RTT impacts if possible and if it makes sense.

Popov72:
In the implementation, the list you provide in the return value of `getCustomRenderList` is not stored in the RTT: the list is looped through and the meshes sent to the rendering manager for further processing.

sebavan:
Do not mind me...  The code only moves around :-) Did not notice

Popov72:
Indeed, I have mostly extracted the "prepare rendering manager" code from the `render` method to its own method :)

Issue #7407, opened 2020-01-09
Physics compound body size calculation


JohanFerret:
Physics compound body not touching the ground.

- Bug repro on [playground](https://playground.babylonjs.com):
https://www.babylonjs-playground.com/#492ZK0#27

- Expected result:
Physics compound body touching the ground.

- Current result:
Physics compound body not touching the ground.

RaananW:
(Documentation only) - Seems like a fundamental issue with compounds:

https://www.babylonjs-playground.com/#492ZK0#28

The transformation calculations of the child-impostors seems to be off. Either that or that the rendering of the object is incorrect.

RaananW:
Reopening - Ammo and Oimo and still acting up:

https://playground.babylonjs.com/#1XWKCR#7

RaananW:
The issue was a missing impostor on the parent:

(cannon) https://playground.babylonjs.com/#1XWKCR#9
(ammo) https://playground.babylonjs.com/#1XWKCR#10
(oimo) https://playground.babylonjs.com/#1XWKCR#11

Closing

PR #7431, opened 2020-01-13
Extend size and position should be in local space for the physics engin(s)

RaananW:
Closing #7407 

I checked other physics demos and they dont seem to be influenced by this change, however it needs to be monitored


RaananW:
Please wait with merging

deltakosh:
good?

RaananW:
Still fighting Oimo, I hope to be done very soon

RaananW:
After this commit oimo compounds work. I had to disable the bidirectional transformation transfer between oimo and babylon when using compound meshes. So, this will work:

https://www.babylonjs-playground.com/#492ZK0#32

Setting the position AFTER creating the impostors will not work (!). To update the position use the physics body's position parameter:

`impostor.physicsBody.position.y = 10`:

https://www.babylonjs-playground.com/#492ZK0#33

(Note that the demos will only work after merging those changes to babylon!!)

@deltakosh - good to go now.

PR #7446, opened 2020-01-15
CSM: Make sure all properties are serialized/parsed

Popov72:


sebavan:
could you also check in the babylonFileLoader if CSM are taking into account ? THX

Issue #7323, opened 2019-12-13
Oculus Quest: Raycast laser color constantly reset


bilkusg:
As described in [this forum issue](https://forum.babylonjs.com/t/raycasting-oculus-quest-cant-change-laser-color/7539), using a VRExperienceHelper,  the function

    changeLaserColor

has no effect on the color of the laser beam from the oculus quest controllers.

The following Playground example shows the result-
[Example](https://playground.babylonjs.com/#LZUEAR#1)

If you run this in VR on an Oculus Quest ( and possibly other headsets ) the laser ray should be red, but it remains white.

The reason is that every time round the render loop, the VRExperienceHelper makes as call to _castRayAndSelectObject, which resets the laser pointer color in a hard-coded way. There seems no obvious workaround, as even trying to reset the color in each render loop doesn't have any effect.

bilkusg:
Hmm, sorry to have taken so long, but I've finally been able to get round to testing this. 
What happens now with the oculus quest is that the ray color is indeed set initially, but as soon as the Quest Controller mesh is loaded, something overrides the laser pointer display, and it reverts to being white all the time. So effectively the laser pointer color still won't change.

RaananW:
Hi @bilkusg , does setting the color after the mesh was loaded work?

I'll be able to inspect the issue soon

bilkusg:
No. Even if I try and set it inside the render loop it still does nothing once the quest controller mesh is loaded.

RaananW:
I haven't tested yet. I will reopen and make sure to address this before 4.1 is released

bilkusg:
In case it helps, the problem seems to be in Cameras/VR/webVRCamera.ts round about line 794 where the function activateLightOnSubMeshes is defined. As far as I can tell, the laser pointer with the correct color is overlaid by a hard-coded pointer which is part of the mesh on the Quest controller and gets lit by that function, thus masking out the laser pointer as set elsewhere.

RaananW:
What function are you using to set the color? are you using `setLaserColor` or `updateLaserColor`? 

This will make it "redish":

https://playground.babylonjs.com/#LZUEAR#3

The reaon it is "redish" and not red is the diffuseColor of the laser pointer. I will reduce is to a grayer value so the red will be more prominent. But make sure you use set and not update.

RaananW:
Sorry, i said set and didnt change the function... here:

https://playground.babylonjs.com/#LZUEAR#6

[...]

PR #7473, opened 2020-01-17
Allowing disable lighting on laser pointers

RaananW:
Laser pointers should always be visible. 

In XR laser pointer and selection mesh now have `disableLighting` on per default (configurable), in WebVR, the VR Helper has a new function `setLaserLightingState`.

Note that the `setLaserLightingState` will work only after the controllers were initialized


RaananW:
Closing #7323

deltakosh:
We have a tiny doc check issue

RaananW:
Of course there is... :-)

PR #7497, opened 2020-01-22
Preview Area in NME can now be viewed in a pop up window

belfortk:
Added ability to view the material preview in new window. Edits made in the NME will be reflected in the pop up window. State should of the preview should be maintained across opening and closing the pop up window.


nodeEditor/src/graphEditor.tsx
```diff
@@ -46,6 +61,18 @@ export class GraphEditor extends React.Component<IGraphEditorProps> {
     private _mouseLocationY = 0;
     private _onWidgetKeyUpPointer: any;
 
+    private _PreviewHost: Nullable<HTMLElement>;
```
deltakosh: lower caps on P

PR #7507, opened 2020-01-23
Profiles motion controller and the input repository

RaananW:
This PR integrates the motion controller library available at https://github.com/immersive-web/webxr-input-profiles/

if enabled (default is true), profiles will be loaded from the repository and will be used when a user enters XR and has a supported controller. It includes meshes and movement definitions of the elements of the controller (buttons, triggers, etc').

Also including a small extension of LoadFileAsync so that returned values can be string instead of ArrayBuffer.


deltakosh:
You are on fire ;D

deltakosh:
Still draft?

RaananW:
issue(s) resolved, ready for review

deltakosh:
Do you  mind creating a tiny demo we could share on twitter?

RaananW:
This will work with any XR demo. Every time you load a controller. I have updated my goal keeper trainer to work with it:

https://playground.babylonjs.com/#PPM311#44

RaananW:
But it is not yet available in the playground

PR #7527, opened 2020-01-25
Add missed negate-methods

avin:
add missed `negate` methods


avin:
not sure, a have to add this to `what's new` section :)

deltakosh:
It's ok :)

PR #7532, opened 2020-01-26
Add some hotkeys and titles for toolbar buttons

avin:
Add some hotkeys and titles for **playground** toolbar buttons

Added hotkeys:
- `Alt+Enter` to RUN (like on shadertoy.com sandbox)
- `Ctrl+S` to Save
- `Ctrl+Shift+S` to Download ZIP

### DIFF
--- a/Playground/js/mainWebGPU.js
+++ b/Playground/js/mainWebGPU.js
@@ -262,11 +262,14 @@ class Main {
         this.loadScriptsList(restoreVersionResult);
 
         // -------------------- UI --------------------
+        var handleRun = () => compileAndRun(this.parent, this.fpsLabel);
+        var handleSave = () => this.askForSave();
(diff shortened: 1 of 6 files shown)


Playground/js/mainWebGPU.js
```diff
@@ -410,6 +411,30 @@ class Main {
         this.parent.settingsPG.setScriptLanguage();
         // Check if it's mobile mode. If true, switch to full canvas by default
         this.parent.menuPG.resizeBigCanvas();
+
+        // HotKeys
+        document.onkeydown = function (e) {
```
deltakosh: No conflict with the code editor?

avin: Code editor does not use these combinations. Other editor's combinations work fine

deltakosh: You rock! Thanks a ton!

Issue #7341, opened 2019-12-16
Performance improvements for ScrollViewer


thomlucc:
Need to improve performance of ScrollViewer as it is not usable when there are a high number of elements. Example in https://forum.babylonjs.com/t/fullscreen-ui-with-scrollviewer-performance/7628/7 with 100x50 elements.

salamaashoush:
You can find sample code to test here https://stackblitz.com/edit/typescript-qlkymb

Popov72:
Currently working on this, but is 276000+ controls really sane :) (276041 exactly) ?

salamaashoush:
@Popov72 you can control how many racks created by changing options in `createRooms` function?

Popov72:
I know but I tried with the default settings in the sample, as it is what should be optimized.

Popov72:
See PR #7534

PR #7534, opened 2020-01-26
Scroll Viewer rendering optimization

Popov72:
Addressing #7341 issue.

I tried to not change the codepath when the optimization is disabled, to be sure that if something is broken it's only when enabling the optimization.

So, I added the `ScrollViewer.freezeControls` property.

When set to `true`, the `layout` method is not called anymore for the window child controls, and only the controls intersecting the window are rendered. Instead of calling `layout` for the children, I simply update the top/left properties according to the window translation values.

I have also added the `ScrollViewer.forceHorizontalBar` and `ScrollViewer.forceVerticalBar` properties. When set to `true`, they force the display of the corresponding bars. When you know your scroll viewer will end up with visible bars, you can set those properties to `true` to save some initialization time, as if it is the scroll viewer control that makes a bar visible in the course of the initialization, it will trigger a children layout rebuild, adding more time to the initialization process.

I have also modified `Measure.transformToRef` to get rid of the `new` calls because it led to some major GCs in my testings.

Knowing the number of `layout` / `render` calls came in handy when developing this, so I kept the feature: let me know if I should remove it.

Samples:
* https://popov72.github.io/BabylonDev/showSample.html?sample=scrollviewer2 - 20000 controls. On my computer, it runs around 2 fps when controls are not frozen and 60 fps when they are.
* https://popov72.github.io/BabylonDev/showSample.html?sample=scrollviewer - 276052 controls (it takes some time before displaying something because of the 276000 control creation process). It's the original sample from @salamaashoush in #7341. It runs around 0.5 fps when controls are not frozen and 12 fps when they are. Note that for this one we can see that there are some renderings with only 200/300 rendering calls for a frame whereas it should always be 276000+. I don't understand why, in the previous sample we always have 20000+ renderings for each frame as expected...

Note that I had to change `protected` to `public` for the `Control._children` property because I needed to traverse the control hierarchy in the exact same way than the `layout` and `render` functions, and using the getter `Control.children` was not possible because the grid control (maybe others too) overrides it and returns only the children of the cells and not the cells themselves.

### DIFF
--- a/gui/src/2D/controls/control.ts
+++ b/gui/src/2D/controls/control.ts
@@ -25,6 +25,11 @@ export class Control {
      */
     public static AllowAlphaInheritance = false;
 
+    /** Gets the number of layout calls made the last time the ADT has been rendered */
+    public static numLayoutCalls = 0;
+    /** Gets the number of render calls made the last time the ADT has been rendered */
+    public static numRenderCalls = 0;
+
     private _alpha = 1;
     private _alphaSet = false;
     private _zIndex = 0;
@@ -1356,6 +1361,8 @@ export class Control {
         }
 
         if (this._isDirty || !this._cachedParentMeasure.isEqualsTo(parentMeasure)) {
+            Control.numLayoutCalls++;
+
             this._currentMeasure.transformToRef(this._transformMatrix, this._prevCurrentMeasureTransformedIntoGlobalSpace);
 
             context.save();
@@ -1596,6 +1603,9 @@ export class Control {
             this._isDirty = false;
             return false;
         }
+
+        Control.numRenderCalls++;
+
         context.save();
 
         this._applyStates(context);
--- a/gui/src/2D/controls/control.ts
+++ b/gui/src/2D/controls/control.ts
@@ -105,6 +105,9 @@ export class Control {
     protected _rebuildLayout = false;
 
     /** @hidden */
+    public _customData: any = {};
+
+    /** @hidden */
     public _isClipped = false;
 
     /** @hidden */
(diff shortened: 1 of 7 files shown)


salamaashoush:
Thanks, @Popov72 this makes a huge difference.

Popov72:
Added a new optimization (`setBucketSizes(width, height)`), [sample 2](https://popov72.github.io/BabylonDev/showSample.html?sample=scrollviewer) is now at 60 fps too.

From the doc:

Sets the bucket sizes.
When `freezeControls` is true, setting a non-zero bucket size will improve performances by updating only controls that are visible. The bucket sizes is used to subdivide (internally) the window area to smaller areas into which controls are dispatched. So, the size should be roughly equals to the mean size of all the controls of the window. To disable the usage of buckets, sets either width or height (or both) to 0.
Please note that using this option will raise the memory usage (the higher the bucket sizes, the less memory used), that's why it is not enabled by default.

gui/src/2D/controls/control.ts
```diff
@@ -25,6 +25,11 @@ export class Control {
      */
     public static AllowAlphaInheritance = false;
 
+    /** Gets the number of layout calls made the last time the ADT has been rendered */
+    public static NumLayoutCalls = 0;
```
deltakosh: Should not be Static but per ADT

Popov72: Going to correct that: I think I will need to create a new PR?

deltakosh: yes my dear :)

PR #7536, opened 2020-01-27
Fixed GeometryBufferRenderer to calculate world position when having …

julien-moreau:
…position or bump defined.


julien-moreau:
Haven't updated what's new as the bug was introduced in the PR about screen space reflections.

PR #7541, opened 2020-01-28
[Playground] Add titles on comparison buttons

sailro:
Titles were added on all PG buttons except when comparison mode is active.


sailro:
cc @avin, perhaps there was a reason for not adding titles to all buttons i.e. comparison. (I git-blamed the file and saw you made the initial change for titles)

PR #7547, opened 2020-01-29
Add comments to NME Frames

belfortk:
Added comment field to Frame Property component, Added new comment div to frames.


nodeEditor/src/diagram/graphFrame.ts
```diff
@@ -38,6 +39,7 @@ export class GraphFrame {
     private _ports: NodePort[] = [];
     private _controlledPorts: NodePort[] = [];
     private _id: number;
+    private _comments: string;
```
deltakosh: You also need to serialize/parse it

belfortk: Thanks for catching that!

PR #7551, opened 2020-01-29
Scroll Viewer: Add specific methods to change the horizontal/vertical bar pictures

Popov72:
I did not modify the what's new file as there is already:

- Scroll Viewer extended to include the use of images in the scroll bars([JohnK](https://github.com/BabylonJSGuide/))

### DIFF
--- a/gui/src/2D/controls/scrollViewers/scrollViewer.ts
+++ b/gui/src/2D/controls/scrollViewers/scrollViewer.ts
@@ -27,7 +27,11 @@ export class ScrollViewer extends Rectangle {
     private _barColor: string;
     private _barBackground: string;
     private _barImage: Image;
+    private _horizontalBarImage: Image;
+    private _verticalBarImage: Image;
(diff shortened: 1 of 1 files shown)


lockphase:
Heyas @Popov72 : Thanks for the PR, I just tested a little and it seems that verticalBarImage and verticalThumbImage are rotated by default (see attached pic). I'm not using the horizontal ones. Is it possible to simple switch off the rotation when using the individual hori/vert bar/thumb images ?

sv_bug1

Popov72:
Hi @lockphase,

Indeed, the rotation is automatically performed by `ImageScrollBar` when `isVertical = true`.

Going to see if I can add a property on `ImageScrollBar` to disable the rotation in this case.

Popov72:
I made this PR: https://github.com/BabylonJS/Babylon.js/pull/7578

Popov72:
#7578 is live

PR #7557, opened 2020-01-30
Add/Fix ArcRotateCamera Pinch to Zoom with new NaturalPinchZoom type

rwedoff:
## What Changed
The current pinch to zoom functionality for the ArcRotateCamera is not generic to all models without setting various parameters. Therefore, I added a new pinch type (to support backwards compatibility) that uses the distance of the camera in relation to the pinch distance. As a result, this is a more natural way to zoom in on a model and it generic to any model, without breaking the current pinch to zoom paradigm. 

## Usage
var camera = new BABYLON.ArcRotateCamera(...);
camera.useNaturalPinchZoom = true;
PinchZoom

### DIFF
--- a/src/Cameras/Inputs/arcRotateCameraPointersInput.ts
+++ b/src/Cameras/Inputs/arcRotateCameraPointersInput.ts
@@ -59,9 +59,10 @@ export class ArcRotateCameraPointersInput extends BaseCameraPointersInput {
     public pinchDeltaPercentage = 0;
 
     /**
-     * useNaturalPinchZoom will be used instead of pinchPrecision if set to true.
-     * Pinch natural zoom is a calculation based off the distance of the
-     * pinch and the camera distance.
+     * When useNaturalPinchZoom is true, multi touch zoom will zoom in such
+     * that any object in the plane at the camera's target point will scale
+     * perfectly with finger motion.
+     * Overrides pinchDeltaPercentage and pinchPrecision.
      */
     @serialize()
     public useNaturalPinchZoom: boolean = false;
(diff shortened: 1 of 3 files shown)


src/Cameras/Inputs/arcRotateCameraPointersInput.ts
```diff
@@ -58,6 +58,14 @@ export class ArcRotateCameraPointersInput extends BaseCameraPointersInput {
     @serialize()
     public pinchDeltaPercentage = 0;
 
+    /**
+     * useNaturalPinchZoom will be used instead of pinchPrecision if set to true.
```
AdamMitchell-ms: I'd rephrase this comment (and others in this review) to be something like this:

When useNaturalPinchZoom is true, multitouch zoom will zoom in such that any object in the plane at the camera's target point will scale perfectly with finger motion.  Overrides pinchDeltaPercentage and pinchPrecision.

src/Cameras/Inputs/arcRotateCameraPointersInput.ts
```diff
@@ -139,7 +147,11 @@ export class ArcRotateCameraPointersInput extends BaseCameraPointersInput {
                 this.camera.inertialRadiusOffset +=
                     (pinchSquaredDistance - previousPinchSquaredDistance) * 0.001 *
                     this.camera.radius * this.pinchDeltaPercentage;
-            } else {
+            } else if (this.useNaturalPinchZoom) {
```
AdamMitchell-ms: This should override pinchDeltaPercentage, if set.  So put it as the first IF in this set of blocks.

rwedoff: Good point, will fix.

deltakosh:
Excellent! Thanks a lot!

PR #7560, opened 2020-01-30
Playground tweaks

RaananW:
The playground is very slow. I have never seen a site scoring 1 in the page-speed insights page:

image

Those are 3 tweaks that reduces the loading time (locally, at least) to ca. 80% of what it was before:

- Use CDN for github files (instead of github itself) - This is for the files loaded from Extensions and Tootkit
- Load playground-demo images (there are ca. 130!) only when the demos menu is opened (instead of every page load)
- Avoid using the **definition worker**. This is a bigger issue. from analysis of the worker code, all it does is search for deprecated definitions (an array of strings with currently 9 elements) and then send them back to the main code. To do that it downloads a 7.2 MB file (!). This file is later loaded by the monaco editor as well. (typescript). I think that setting the deprecated array once every release (maybe during build process?) Would be a better way to achieve this.

Other issues:

- Netlify's CDN has a long TTFB - at least in germany (The green part, which is just waiting for the file to load):

image

This is probably because the (sub)domain is not controlled by netlify, but is being CNAME-Forwarded to a single server. If the redirect works on DNS level it will be faster from more locations. (Only an assumption!). 

We could also try using jsdeliver's GitHub CDNs for a week and see which is better.

I will be happy to discuss each one of those (and other ways to make it work faster)

### DIFF
--- a/Playground/js/monacoCreator.js
+++ b/Playground/js/monacoCreator.js
@@ -86,12 +86,25 @@ class MonacoCreator {
     };
 
     setupDefinitionWorker(libContent) {
-        this.definitionWorker = new Worker('js/definitionWorker.js');
-        this.definitionWorker.addEventListener('message', ({ data }) => {
-            this.deprecatedCandidates = data.result;
-            this.analyzeCode();
-        });
-        this.definitionWorker.postMessage({ code: libContent });
+        // What is the reason behind this worker?
+
+        // this.definitionWorker = new Worker('js/definitionWorker.js');
+        // this.definitionWorker.addEventListener('message', ({ data }) => {
+        //     this.deprecatedCandidates = data.result;
+        //     this.analyzeCode();
+        // });
+        // this.definitionWorker.postMessage({ code: libContent });
+        this.deprecatedCandidates = [
+            "FromFloatArray",
+            "FromFloatArrayToRef",
+            "getStrideSize",
+            "getStrideSize",
+            "getOffset",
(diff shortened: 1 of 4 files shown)


deltakosh:
I LOVE IT! I never took the time to check what was wrong but I wanted to do it forever!.. Let's do everything we can to make it even faster.

> We could also try using jsdeliver's GitHub CDNs for a week and see which is better.

GO!

RaananW:
I have some other suggestions to improve the playground, want to talk about it when you have the time?

deltakosh:
sure thing!

Playground/js/monacoCreator.js
```diff
@@ -86,12 +86,25 @@ class MonacoCreator {
     };
 
     setupDefinitionWorker(libContent) {
-        this.definitionWorker = new Worker('js/definitionWorker.js');
-        this.definitionWorker.addEventListener('message', ({ data }) => {
-            this.deprecatedCandidates = data.result;
-            this.analyzeCode();
-        });
-        this.definitionWorker.postMessage({ code: libContent });
+        // What is the reason behind this worker?
```
deltakosh: This is something we need to restore as it is used to analyze the code looking for @deprecated flags

deltakosh: Ok so we chatted offline and actually this worker was dl a 7MB file which gives us a very bad score on Google Search

PR #7562, opened 2020-01-30
Playground - moving possible scripts to the body and to different CDNs

RaananW:
**Moved the editor script** to the body as well, which works very well locally. I want to first check it on netlify before deploying. Is that possible?


deltakosh:
Not anymore as Netlify started to charge us :)

You need to deploy it to your own netlify space

RaananW:
Will do

PR #7572, opened 2020-01-31
Small Playground changes

RaananW:
1) Reinstated the definition worker (until a better solution will be found)
2) It is possible to use query strings instead of hash queries. So this - `?pg=3VMTI9&revision=1` is the same as this `#3VMTI9#1` but indexable.
3) Added meta description and keywords that are being updated when a playground is loaded. we will need to start updating playgrounds with better description, title, and tags
4) Moved ALL javascript files to the body (avoiding render-blocking).

This is the version that is hosted here https://playgroundcdn1.azureedge.net/ and https://playgroundcdn2.azureedge.net/ (different CDN providers). **Please check it as well.**

### DIFF
--- a/Playground/index.html
+++ b/Playground/index.html
@@ -1,410 +1,453 @@
 ﻿<!DOCTYPE html>
 <html>
 
-<head>
-    <title>Babylon.js Playground</title>
-    <meta charset='utf-8' />
-    <meta name="viewport" content="width=device-width, user-scalable=no">
-    <link rel="shortcut icon" href="https://www.babylonjs.com/favicon.ico">
+    <head>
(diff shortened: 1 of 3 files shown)


RaananW:
I will also be running deployment tests from a local branch during the weekend before moving to the new host.

deltakosh:
Getting a 404 for css on the cdn1:
image

RaananW:
I just purged the entire CDN (deployed a new version), did a reload fix it?

RaananW:
image

(Desktop - already much better

deltakosh:
I had to force reload it 3 times but it works now

PR #7578, opened 2020-02-01
Added ImageScrollBar.num90RotationInVerticalMode property

Popov72:
Added `ImageScrollBar.num90RotationInVerticalMode` property to let the user freely rotate the pictures (background / thumb) when in vertical mode.

This property is equal to 1 by default to stay backward compatible.

I noticed the thumb image is rotated -90° whereas it is 90° for the background picture. I left it as is to no break compatibility but was there a reason for this? It may seem awkward for the user to have a different rotation for the background and for the thumb.

Should we have two `num90RotationInVerticalMode` properties, one for the background and one for the thumb image?


deltakosh:
> I noticed the thumb image is rotated -90° whereas it is 90° for the background picture

most likely a bug

> Should we have two num90RotationInVerticalMode properties, one for the background and one for the thumb image?

Let's see if we need it. Let's keep it simple for now

lockphase:
erm, how do I actually use it, can't seem to get it to work.
https://www.babylonjs-playground.com/#4ZC0G4#5

Popov72:
The property is on the bar, not on the picture:

https://www.babylonjs-playground.com/#4ZC0G4#6

lockphase:
Got it! Thanks for the PR!

PR #7583, opened 2020-02-03
Playground changes

RaananW:
New playground changes:

1) Playgrounds can be loaded with a full URL, for example: https://playgroundcdn1.azureedge.net/pg/QY1WYT/revision/0
2) fixed frame.html to fit the new architecture
3) caching and page speed improvements

This is the version deployed to https://playgroundcdn1.azureedge.net/, which is not yet the version on playground.babylonjs.com


RaananW:
URL-Lowercasing is being deployed, might experience a few 404s in the next hour

PR #7602, opened 2020-02-06
Fix for light gizmo changing light transforms on activation.

Drigax:
This PR addresses an issue in the lights gizmo where activating the gizmo causes the light to invert direction.

This appears caused by us not properly setting and recalculating the cached light orientation and direction after creating the gizmo, causing the light's transform to be improperly updated on the next BeforeRender tick, using the stale world matrix of the gizmo. This appears to cause a toggling behavior, as the gizmo is removed and recreated.

My proposed fix forces a world matrix recalculation, and sets the cached value appropriately.


Drigax:
addressing #7603

PR #7607, opened 2020-02-06
XR Headset impostor

RaananW:
It is now possible to use your head when enabling physics for XR.

It is also possible to get the physics impostors of the controllers and the headset, if enabled.


deltakosh:
webgl1 are failing for some reasons

deltakosh:
not your fault..merging

Issue #7377, opened 2019-12-30
Incorrect lights direction when exported to glb 0, 1 or 2 times


noalak:
Primary playground is [#KX53VK#30](https://www.babylonjs-playground.com/#KX53VK#30) testing right handed export to glb.
All bugs could be related, so I grouped them here.

### Black triangles
**Bug repro:** https://www.babylonjs-playground.com/#9DXCHP#1
**Expected result:** Triangles should be colored
**Current result:** When scene is exported and imported once (not twice), the triangles are black (unlighted).
**Comments:**
The light direction is back to front (0,0,-1).
When imported into the sandbox, the triangles are visible due to a default ambient light. The directional light is still back to front and the gizmo toggle has no effect on light direction. Those behaviors in the sandbox are as expected.

### Light direction
**Bug repro:** https://www.babylonjs-playground.com/#KX53VK#30
**Expected result:** The light should be oriented as initially front to back (0,0,1).
**Current result:** When scene is exported and imported twice (unmodified playground), the light direction is reversed: oriented back to front (0,0,-1).
**Comments:**
The gizmo toggle has no effect on light direction (unlike other 2 bugs above).
Triangles are lighted correctly, despite light direction being reversed.

### Linked light dummies
**Bug repro:** https://www.babylonjs-playground.com/#KX53VK#30
**Expected result:** Each _Light_ dummy should be independant from each other
**Current result:** the 2 _Light_ dummies, parent of the actual directional light, are someway linked together. Transform edition of one dummy through Inspector impacts the other. Comparing world matrix using the console.

I'm off the office until Monday 6th.
Adding @Jaskar for any question meanwhile.

deltakosh:
Adding @thomlucc to track them

@noalak @Jaskar can you guys work on it?

Drigax:
Linking
https://github.com/BabylonJS/Exporters/issues/696

As my test process touches on the light gizmo not aligning with the light's actual orientation

Drigax:
Few initial takeaways:
I'll start by figuring out why the gizmo is improperly oriented/modifies the light orientation on activation in a Right-Handed coordinate space. This seems key to being able to properly visually debug the other elements.

I'd assume that we're just improperly flipping the light direction when exporting the scene. This seems similar to our double root issues addressed via #7257 where we're converting our already converted light transform, but I'll have to look closer into the source to confirm.

@noalak for the linked light dummy issue, you linked to the original PG:
[...]

PR #7610, opened 2020-02-06
Fix Right Handed light and shadow orientation regression.

Drigax:
This PR reverts commit c97a6e5e43f3d4727cead22972fdb8215c1d05f8 (created via PR #6712) , which inverted directional light and shadow matrices based on scene handedness. 

This PR also adds a Right handed coordinate system variant of the Soft Shadows test, with associated baseline image, and updated GLTF Serializer with Negative World Matrix to point to playgrounds with proper directional light direction initialization.

Created as part of investigation of #7377.


Popov72:
In the cascaded shadow generator code there is:
```typescript
if (this._scene.useRightHandedSystem) {
    Matrix.OrthoOffCenterRHToRef(this._cascadeMinExtents[cascadeIndex].x, this._cascadeMaxExtents[cascadeIndex].x, this._cascadeMinExtents[cascadeIndex].y, this._cascadeMaxExtents[cascadeIndex].y, minZ, maxZ, this._projectionMatrices[cascadeIndex]);
} else {
    Matrix.OrthoOffCenterLHToRef(this._cascadeMinExtents[cascadeIndex].x, this._cascadeMaxExtents[cascadeIndex].x, this._cascadeMinExtents[cascadeIndex].y, this._cascadeMaxExtents[cascadeIndex].y, minZ, maxZ, this._projectionMatrices[cascadeIndex]);
}
```
to create the orthographic projection matrix of each cascade.

As I understand it (I wrote the code, but I copied what @lockphase did in the first place in this regard, as I don't master all those left/right handedness shenanigans), it was modeled after the same code found in the `DirectionalLight` class, but because this code is now removed, I think we should also update the `CascadedShadowGenerator` code?

Drigax:
Ah @Popov72, I agree that we should update the CascadedShadowGenerator code as well. Do we have a testing scene that I can check? Most likely the CSG would fail right now for our right handed case.

Popov72:
There is the PG linked in the doc: https://www.babylonjs-playground.com/#IIZ9UU#40

I have tried locally to set `scene.useRightHandedSystem = true;` but there are artifacts with both code (current and when only using `OrthoOffCenterLHToRef`)...

Drigax:
Ok, this seems kind of wrong. I reverted to before the lighting change, the shadows look properly rendered:

Left hand:
image
Right hand:
image

And on the latest:
Left hand:
image
Right hand:
image

With the odd long shadow artifact visible depending on the disc location?
Left hand:
image

Right hand:
image

Drigax:
Hmm...as you say changing the shadow matrix calculation to only use the left handed calculation results in some upside down shadow calculations?:

image

Drigax:
Created an issue to discuss this issue specifically. Unfortunately I haven't made much headway other than changing random calculations to not much avail.

Issue #7612, opened 2020-02-07
Member filtering is broken in the playground


sailro:
@RaananW it seems you are working on optimizing the workload around the playground, this is great but please be extra careful as this part is highly technical:

For instance, you recently changed the monaco version used:
https://github.com/BabylonJS/Babylon.js/commit/a5d9190a885ac35b4837c10e865f84ecade15667#diff-68a866b97c7bf8685f1e78573f06eb21

And you removed the hook we made on purpose to filter members, see this PR:
https://github.com/BabylonJS/Babylon.js/pull/7169

We were using the 'dev' version purposely to do something that is not yet possible with the current Monaco API (using monkey patching).

You should review the repo history/PR to make sure you are not breaking something going forward.

I fully agree that this one is a bit tricky and we should add comments to properly explain what we do.

cc @deltakosh 

Thanks!
Seb

RaananW:
I am reverting this right now, not a problem at all.

sailro:
Thanks, I will submit a PR to add better comments for this tricky part.

RaananW:
let me first clean this up.

I did that for two reasons:

1) loading from node_modules is not the cleanest way of loading assets
2) cloudflare's cdn is fast and reliable.

But I understand there is a reason for that.

PR #7613, opened 2020-02-07
Playground changes

RaananW:
- Bringing back dev version of monaco (#7612)
- Avoid CSS imports, loading fontawesome from cloudflare CDN (1 less DNS query)
- Removed HTML comments in debug.html and frame.html

Are indexStable, full, ts and js HTML files being used? At the moment I don't see any use for them, but I am leaving ATM.

### DIFF
--- a/Playground/css/index.css
+++ b/Playground/css/index.css
@@ -1,4 +1,4 @@
-﻿@import url("https://maxcdn.bootstrapcdn.com/font-awesome/4.6.3/css/font-awesome.min.css");
+﻿
 html,
 body {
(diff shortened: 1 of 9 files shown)


deltakosh:
> Are indexStable, full, ts and js HTML files being used? At the moment I don't see any use for them, but I am leaving ATM.

I don't think so (if they are not used to switch between languages, they we can remove them)

RaananW:
I will remove them in a later PR where I am going to delete a few other unused files. worse case we can revert this PR.

PR #7615, opened 2020-02-07
add blend mode options

IIerpos:
I haven't tested it, as I've never built Babylon for myself. But the change is small and simple. Please have a look.

### DIFF
--- a/src/Sprites/spriteManager.ts
+++ b/src/Sprites/spriteManager.ts
@@ -56,7 +56,7 @@ export interface ISpriteManager extends IDisposable {
      */
     intersects(ray: Ray, camera: Camera, predicate?: (sprite: Sprite) => boolean, fastCheck?: boolean): Nullable<PickingInfo>;
 
-        /**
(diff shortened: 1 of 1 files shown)


src/Sprites/spriteManager.ts
```diff
@@ -76,6 +76,19 @@ export interface ISpriteManager extends IDisposable {
  * @see http://doc.babylonjs.com/babylon101/sprites
  */
 export class SpriteManager implements ISpriteManager {
+    /** The sprite's color is added to the destination color without alpha affecting the result */
+    public static BLENDMODE_ONEONE = 0;
```
deltakosh: We should not recreate it but instead use the one from the Constants

deltakosh:
Btw please do not target preview but master :)

src/Sprites/spriteManager.ts
deltakosh: Well I'll do the fix ;)

PR #7618, opened 2020-02-08
Improvements for playground

avin:
Some apps (for ex. `Telegram`) and services convert second hash symbol in Babylon Playground links from `#` to `%23` and these links become unreachable for playground. This PR solves this problem
`#UYS16D%236` --> `#UYS16D#6`

`+` plus make small fix for unreachable selectors error

### DIFF
--- a/Playground/js/main.js
+++ b/Playground/js/main.js
@@ -1073,6 +1073,11 @@ class Main {
             }
         }
         if (pgHash) {
+            var match = pgHash.match(/^(#[A-Z\d]*)(%23)([\d]+)$/);
+            if (match){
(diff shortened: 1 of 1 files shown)


Playground/js/main.js
```diff
@@ -1073,6 +1073,11 @@ class Main {
             }
         }
         if (pgHash) {
+            var match = pgHash.match(/^(#[A-Z\d]*)(%23)([\d]+)$/);
```
RaananW: If you can please add [a-z] to the PG-matcher as well (right next to the [A-Z], just to stay future safe

avin: done

PR #7626, opened 2020-02-11
Minor svg loading bug fix for chrome

lockphase:
Apparently, chrome can fail on getElementById returning null even if the svg viewbox attributes are obtained. So submitting this minor fix. Didn't change what's new because the same msg is there.

PS: tsLint and build all failed due to whitespace errors not from associated files, so this is a blind PR. :(

### DIFF
--- a/gui/src/2D/controls/image.ts
+++ b/gui/src/2D/controls/image.ts
@@ -487,7 +487,7 @@ export class Image extends Control {
                     var docwidth = Number(svgDoc.documentElement.getAttribute("width"));
                     var docheight = Number(svgDoc.documentElement.getAttribute("height"));
                     var elem = <SVGGraphicsElement> <unknown> svgDoc.getElementById(elemid);
-                    if (vb && docwidth && docheight) {
+                    if (elem && vb && docwidth && docheight) {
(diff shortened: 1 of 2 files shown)


gui/src/2D/controls/image.ts
```diff
@@ -486,7 +486,8 @@ export class Image extends Control {
                     var vb = svgDoc.documentElement.getAttribute("viewBox");
                     var docwidth = Number(svgDoc.documentElement.getAttribute("width"));
                     var docheight = Number(svgDoc.documentElement.getAttribute("height"));
-                    if (vb && docwidth && docheight) {
+                    var elem = <SVGGraphicsElement> <unknown> svgDoc.getElementById(elemid);
```
deltakosh:
```suggestion
                    var elem = <SVGGraphicsElement>svgDoc.getElementById(elemid);
```

Should be enough

deltakosh: I'll do the change :)

lockphase: Thanks ! Different time zones =)

deltakosh: My pleasure ;)

PR #7344, opened 2019-12-17
Ignore root nodes when exporting to GLB

noalak:
Ignore the root nodes used for conversion from right-handed to left-handed when exporting to GLB.

Solving issue #6349.

It implements Gary's solution discussed in last PR https://github.com/BabylonJS/Babylon.js/pull/7338.

PR is split in 2 commits:
1) First commit for ignoring **one** root node
2) Second commit to ignore **multiple** root nodes

It might be easier to look at file changed in 1st commit first.

Regarding implementation, the __convertToRightHandedSystemMap_ is initialized with all nodes. You may prefer to have a default value used when Map.get returns undefined.
Right now, it's sometime necessary to cast to boolean in case the node is not found (should never happen): `convertToRightHandedSystemMap.get(babylonTransformNode) as boolean`

@bghgary Thanks for the hints! Could you review the PR ?

PR #7629, opened 2020-02-11
Fix glTF exporting nested __root__ nodes 

Drigax:
This PR addresses an issue where round-tripped glTF scenes are exported with an additional __root__ node encapsulating the scene. This fix addresses this by explicitly avoiding serialization of root objects that have a coordinate space converting transform, named "_root_"

This PR is a cherrypick of @noalak's initial pass via #7344 with enhancements made for right handed round-tripping, and support for edge cases where an overriding side orientation is used.


deltakosh:
Linting is failing

Drigax:
Addressing #6349

Drigax:
This change wall cause CI to fail due to #7377 causing a light encapsulating object to be duplicated, which flips its orientation. Previously this was combined with the world orientation being flipped due to the duplication of world encapsulating nodes, but this fix exposes this behavior:
image

I'll temporarily change the test result image to reflect this, until I fix 7377.

PR #7646, opened 2020-02-13
Add SpriteManager.disableDepthWrite property

Popov72:


CNAME
```diff
@@ -1 +0,0 @@
-babylonjs.com
```
deltakosh: Why this change?

Popov72:
> Only checking for the CNAME

Was testing something on my side, did not see it had been committed, sorry.

PR #7647, opened 2020-02-13
console.log dropped

Pryme8:
Per post:
https://forum.babylonjs.com/t/sprite-blendmodes-not-working-correctly/8534/12?u=pryme8


deltakosh:
Already fixed in the other PR

PR #7658, opened 2020-02-14
Fix glTF serializer light orientation issues

Drigax:
see: https://github.com/BabylonJS/Babylon.js/issues/7377#issuecomment-586203022

This PR is intended to address issues with export and round-tripping of punctual lights via glTF.

A fix was made to the yaw angle calculation of a punctual non-point light node when exported from a right hand scene.

Changes were made to the _applyExtension contract to allow for null nodes to be returned to the extension export caller after being processed via the extension handler function. _applyExtension now ONLY returns the processed node, and can return null.


RaananW:
Linting and documentation fails

deltakosh:
RIGHT on time ;)

PR #7668, opened 2020-02-14
CSM: Fix display bug when scene is right handed

Popov72:
Tested in https://www.babylonjs-playground.com/#IIZ9UU#42 as well as in my custom app with bigger scenes, so the fix should be good.

I didn't update the what's new file as CSM is new in 4.1 anyway.


Drigax:
@Popov72 you are a beast!

Popov72:
Linting problems in `instancedMesh.ts` but I didn't change this file (?)

deltakosh:
I'll fix that part

PR #7685, opened 2020-02-19
Added upward and downward move

Pheater:
Added _forward/backward_ keys and _upward/downward_ keys to retain backwards compatibility to the original incorrect naming scheme, as well as remove confusion.


deltakosh:
Well you should have discussed on the forum first as mentioned on the template. I'm not for adding properties that are duplicate with existing one (which I personally do not find misleading).

deltakosh:
https://github.com/BabylonJS/Babylon.js/blob/master/contributing.md

PR #7689, opened 2020-02-19
Fix order of PBR glTF extensions

bghgary:
Load clearcoat/sheen/specular glTF extension before specGloss.

This fixes an incompatibility when using clearcoat/sheen/specular glTF extension with the specularGlossiness extension.


bghgary:
See here for a sample model: https://github.com/KhronosGroup/glTF-Sample-Models/blob/cd0d44d89b1d333d8033db75dd2b39f4d6355fb6/2.0/SpecGlossVsMetalRough/glTF-Binary/SpecGlossVsMetalRough_clearcoat.glb

PR #7691, opened 2020-02-19
update RGB values for default bg color

belfortk:
Update the default RBG vaules for background color to values that make sense for a Color4.

Before the bg was all white at first load.


belfortk:
Also included a fix for the memory leak introduced in my preview area pop up PR.

It was bothering me.

Issue #7697, opened 2020-02-20
DeviceOrientationCamera properties not displayed in inspector


brian-hay:
### Bug repro
https://www.babylonjs-playground.com/#SRZRWV#3
(activate Inspector then select camera)

### Expected result
DeviceOrientationCamera properties display in inspector like other camera types e.g. ArcRotateCamera.

### Current result
DeviceOrientationCamera properties not displayed in inspector

RaananW:
will soon be deployed

PR #7698, opened 2020-02-20
inspector should show freecamera-based cameras

RaananW:
And WebXR camera was missing the getClassName function


RaananW:
Closing #7697

PR #7729, opened 2020-02-27
Add KHR_mesh_instancing

### LINKED ISSUE #7521: Add glTF loader support for KHR_mesh_instancing
See https://github.com/KhronosGroup/glTF/pull/1691/files for implementation spec.

Drigax:
This pull request implements the KHR_mesh_instancing extension specified via: https://github.com/KhronosGroup/glTF/pull/1691

This implementation implements this extension via instance buffer parsing and instantiation of the individual instances via the provided accessors.

Fixes #7521


loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,109 @@
...
+                            let digitLength = instanceCount.toString().length;
+                            // loop through all the instances and create instances, parent to original transform
+                            for (let i = 0; i < instanceCount; ++i) {
+                                let instance = babylonMeshInstanceNode.clone((babylonMeshInstanceNode.name || babylonMeshInstanceNode.id) + "_" + String(i).padStart(digitLength,'0'), babylonTransformNode, true);
+                                if (instance) {
+                                    instance.position = attributeBuffers["TRANSLATION"] ? Vector3.FromArray(attributeBuffers["TRANSLATION"].buffer, i * 3)
+                                                                                        : Vector3.Zero();
+                                    instance.rotationQuaternion = attributeBuffers["ROTATION"] ? Quaternion.FromEulerAngles(attributeBuffers["ROTATION"].buffer[i * 3], attributeBuffers["ROTATION"].buffer[i * 3 + 1], attributeBuffers["ROTATION"].buffer[i * 3 + 2]) //Quaternion.FromArray(attributeBuffers["ROTATION"].buffer, i * 4) //
```
Drigax: This is improper parsing of the rotation buffer. I originally used https://github.com/KhronosGroup/glTF/blob/b4854e9a86423c601a90c8a3eb64078b750fead6/extensions/2.0/Khronos/KHR_mesh_instancing/samples/teapots_galore_TRS/teapots_galore_TRS.gltf

as a sample scene for implementation, which at time of authoring used an euler angle VEC3 to store the rotation buffer. This is expected to be changed to VEC4 quaternion.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,109 @@
...
+                                                                                                : Quaternion.Identity();
+                                    instance.scaling = attributeBuffers["SCALE"] ? Vector3.FromArray(attributeBuffers["SCALE"].buffer, i * 3)
+                                                                                    : Vector3.One();
+                                    instance.computeWorldMatrix(true);
+                                }
+                            }
+                        }
+                        babylonMeshInstanceNode.setParent(babylonTransformNode);
```
Drigax: Since we are instantiating an additional copy of the source and parent node meshes, I think we'd need some way to properly clean up the scene at export. Currently this implementation adds an additional instance to serve as the instanced copys "source". I'm not sure if there is a cleaner way of handling this case, as we are not guaranteed that the mesh of the node containing this extension is the same as the mesh that will be used for instancing.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+                        if (attributeKeys.length > 0) {
+                            let instanceCount = attributeBuffers[attributeKeys[0]].accessor.count;
+                            let digitLength = instanceCount.toString().length;
+                            for (let i = 0; i < instanceCount; ++i) {
+                                let instance;
+                                if (i === 0){
+                                    instance = babylonMeshInstanceNode;
+                                    instance.setParent(babylonTransformNode);
+                                } else {
+                                    instance = babylonMeshInstanceNode.clone((babylonMeshInstanceNode.name || babylonMeshInstanceNode.id) + "_" + String(i).padStart(digitLength,'0'), babylonTransformNode, true);
```
bghgary: `padStart` is not compatible with IE, so you can't use it :(

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    /** @hidden */
+    public loadNodeAsync(context: string, node: INode, assign: (babylonTransformNode: TransformNode) => void): Nullable<Promise<TransformNode>> {
+        return GLTFLoader.LoadExtensionAsync<IKHRMeshInstancing, TransformNode>(context, node, this.name, (extensionContext, extension) => {
+            const nodeMesh = node.mesh;
+            const promise = this._loader.loadNodeAsync(`#/nodes/${node.index}`, node, (babylonTransformNode) => {
+                babylonTransformNode.parent = this._loader.rootBabylonMesh;
+                const attributeBufferPromises = new Array<Promise<any>>();
+
+                let attributes = extension.attributes;
```
bghgary: use `const` when possible instead of `let`

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    /** @hidden */
+    public dispose() {
+        delete this._loader;
+    }
+
+    /** @hidden */
+    public loadNodeAsync(context: string, node: INode, assign: (babylonTransformNode: TransformNode) => void): Nullable<Promise<TransformNode>> {
+        return GLTFLoader.LoadExtensionAsync<IKHRMeshInstancing, TransformNode>(context, node, this.name, (extensionContext, extension) => {
+            const nodeMesh = node.mesh;
```
bghgary: what is `nodeMesh` used for? There is only one instance inside of the loadNodeAsync continuation. It looks like it's being restored at the end, but no one seems to be overriding `node.mesh`?

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: Removed the parsing hack.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+                let attributes = extension.attributes;
+                let attributeKeys = Object.keys(attributes);
+                const attributeBuffers : { [name: string]: { buffer : Float32Array, accessor : IAccessor } } = {};
+                for (let i = 0; i < attributeKeys.length; ++i) {
+                    let attributeKey = attributeKeys[i];
+                    let accessor = ArrayItem.Get(`${extensionContext}/attributes/${attributeKey}`, this._loader.gltf.accessors, attributes[attributeKey]);
+                    attributeBufferPromises.push(this._loader._loadFloatAccessorAsync(`/accessors/${accessor.bufferView}`, accessor).then((data) => {
+                        attributeBuffers[attributeKey] = { buffer : data, accessor : accessor };
+                    }));
+                }
```
bghgary: I would recommend not looping through all the attributes since we will only be able to understand specific ones and thus loading them will be pointless. I would explicitly load `TRANSLATION`, `ROTATION`, `SCALE`.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: I decided to reuse the initial source mesh created as a result of `_loadmeshAsync` as the first mesh instance.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+
+                const mesh = ArrayItem.Get(`${context}/mesh`, this._loader.gltf.meshes, extension.mesh || nodeMesh);
+                this._loader._loadMeshAsync(`/meshes/${mesh.index}`, node, mesh, (babylonMeshInstanceNode) => {
+                    Promise.all(attributeBufferPromises).then(() => {
```
bghgary: The promise returned by Promise.all is being dropped here. It needs to be captured to get exceptions, ensure that the outer promise completes when all the instances are actually created. Probably what should happen is all of the async work should be collected into the promises and then wait for all of them.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+                                    instance.scaling = attributeBuffers["SCALE"] ? Vector3.FromArray(attributeBuffers["SCALE"].buffer, i * 3)
+                                                                                    : Vector3.One();
+                                    instance.computeWorldMatrix(true);
+                                }
+                            }
+                        }
+                        let babylonMesh = (babylonTransformNode as Mesh);
+                        if (babylonMesh) {
+                            babylonMesh.isVisible = false;
+                        }
```
bghgary: I'm not sure about this. I think perhaps the first one should be a regular mesh and the other ones will be instances of the first one to avoid adding an invisible source mesh? There seems to already be a check for `i == 0`, which I'm also not a big fan of.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+                                    instance = babylonMeshInstanceNode.clone((babylonMeshInstanceNode.name || babylonMeshInstanceNode.id) + "_" + String(i).padStart(digitLength, '0'), babylonTransformNode, true);
+                                }
+
+                                if (instance) {
+                                    instance.position = attributeBuffers["TRANSLATION"] ? Vector3.FromArray(attributeBuffers["TRANSLATION"].buffer, i * 3)
+                                                                                        : Vector3.Zero();
+                                    instance.rotationQuaternion = attributeBuffers["ROTATION"] ? Quaternion.FromArray(attributeBuffers["ROTATION"].buffer, i * 4)
+                                                                                            : Quaternion.Identity();
+                                    instance.scaling = attributeBuffers["SCALE"] ? Vector3.FromArray(attributeBuffers["SCALE"].buffer, i * 3)
+                                                                                    : Vector3.One();
```
bghgary: Should ideally use Vector3.FromArrayToRef on the existing `position` and `scaling` vectors instead of allocating new ones. Vector3.Zero, Quaternion.Identity, Vector3.One also allocate which are also unnecessary here.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -709,7 +709,10 @@ export class GLTFLoader implements IGLTFLoader {
         });
     }
 
-    private _loadMeshAsync(context: string, node: INode, mesh: IMesh, assign: (babylonTransformNode: TransformNode) => void): Promise<TransformNode> {
+    /**
+    * @hidden Define this method to modify the default behavior when loading data for meshes
+    */
```
bghgary:
```suggestion
    /** @hidden */
```

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1561,7 +1564,10 @@ export class GLTFLoader implements IGLTFLoader {
         return accessor._data;
     }
 
-    private _loadFloatAccessorAsync(context: string, accessor: IAccessor): Promise<Float32Array> {
+    /**
+    * @hidden Define this method to modify the default behavior when loading data for meshes
+    */
```
bghgary:
```suggestion
    /** @hidden */
```

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: Ah, my mistake. I created a scoped function to handle the padding.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: I think my first pass was a bit confused. I assumed I wanted to save nodeMesh as our instance fallback mesh in case the extension did not specify a mesh to instance. The spec has evolved since my first pass, and this is no longer needed. we can safely assume that the mesh that contains the extension is the one to be instanced.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: Ok, I created a helper function to handle the asynchronous attribute retrieval, and we only explicitly retrieve `TRANSLATION`, `ROTATION` and `SCALE`

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: Ah, I wasn't sure how to use promises initially, I changed the `loadNodeAsync` promise continuation to return `Promise.all`, which now returns our babylonTransformNode in its continuation.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: I agree, this has been modified to instead iterate through all the mesh primitives, and use the transformNodes returned via `loadNodeAsync` as our source meshes to clone for each primitive instance.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: Ok, I needed to implement `Quaternion.FromArrayToRef` to support this behavior. Previously we did not have a way to set a quaternion's values via indexing into an array.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,98 @@
...
+                for (let i = 0; i < attributes.length; ++i) {
+                    loadAttribute(attributes[i]);
+                }
+
+                return Promise.all(promises).then(() => {
+                    const instanceCount = attributeBuffers[attributes[0]].accessor.count;
+                    const digitLength = instanceCount.toString().length;
+                    const padNumber = function(num: number, length: number) {
+                        var str = String(num);
+                        while (str.length < (length)) { str = "0" + str; }
```
bghgary: nit:
```suggestion
                        while (str.length < length) { str = "0" + str; }
```

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,98 @@
...
+                        return str;
+                    };
```
bghgary: nit: consider putting this in StringTools

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,98 @@
...
+                const loadAttribute = (attribute: string) => {
+                    if (extension.attributes[attribute] == undefined) {
+                        return;
+                    }
+                    const accessor = ArrayItem.Get(`${extensionContext}/attributes/${attribute}`, this._loader.gltf.accessors, extension.attributes[attribute]);
+                    promises.push(this._loader._loadFloatAccessorAsync(`/accessors/${accessor.bufferView}`, accessor).then((data) => {
+                        attributeBuffers[attribute] = { buffer : data, accessor : accessor };
+                    }));
+                };
+                const attributes = ["TRANSLATION", "ROTATION", "SCALE"];
```
bghgary: nit: this allocates an array every time. it's better to call it explicitly instead of looping.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,98 @@
...
+    public dispose() {
+        delete this._loader;
+    }
+
+    /** @hidden */
+    public loadNodeAsync(context: string, node: INode, assign: (babylonTransformNode: TransformNode) => void): Nullable<Promise<TransformNode>> {
+        return GLTFLoader.LoadExtensionAsync<IKHRMeshInstancing, TransformNode>(context, node, this.name, (extensionContext, extension) => {
+            return this._loader.loadNodeAsync(`#/nodes/${node.index}`, node, (babylonTransformNode) => {
+                const promises = new Array<Promise<any>>();
+                const attributeBuffers : { [name: string]: { buffer : Float32Array, accessor : IAccessor } } = {};
```
bghgary: nit: spacing

```suggestion
                const attributeBuffers: { [name: string]: { buffer: Float32Array, accessor: IAccessor } } = {};
```

consider formatting the whole file using auto format

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,98 @@
...
+                        var str = String(num);
+                        while (str.length < (length)) { str = "0" + str; }
+                        return str;
+                    };
+
+                    for (let i = 0; i < instanceCount; ++i) {
+                        if (node._primitiveBabylonMeshes) {
+                            for (let j = 0; j < node._primitiveBabylonMeshes.length; ++j) {
+                                const babylonMeshPrimitive = node._primitiveBabylonMeshes[j];
+                                const instance = babylonMeshPrimitive.clone((babylonMeshPrimitive.name || babylonMeshPrimitive.id) + "_" + padNumber(i, digitLength), babylonTransformNode, true);
```
bghgary: Cloning is not the same as creating a new instance? `Mesh.createInstance`?

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,98 @@
...
+
+                                if (instance) {
+                                    attributeBuffers["TRANSLATION"] ? Vector3.FromArrayToRef(attributeBuffers["TRANSLATION"].buffer, i * 3, instance.position)
+                                                                    : instance.position.set(0, 0, 0);
+                                    attributeBuffers["ROTATION"] ? Quaternion.FromArrayToRef(attributeBuffers["ROTATION"].buffer, i * 4, instance.rotationQuaternion!)
+                                                                 : instance.rotationQuaternion?.set(0, 0, 0, 1);
+                                    attributeBuffers["SCALE"] ? Vector3.FromArrayToRef(attributeBuffers["SCALE"].buffer, i * 3, instance.scaling)
+                                                              : instance.scaling.set(1, 1, 1);
```
bghgary: Once you are using createInstance, there is no need to set the position, rotationQuaternion, scaling if there is no attribute for them since they default to zero, identity, one respectively.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -709,7 +709,8 @@ export class GLTFLoader implements IGLTFLoader {
         });
     }
 
-    private _loadMeshAsync(context: string, node: INode, mesh: IMesh, assign: (babylonTransformNode: TransformNode) => void): Promise<TransformNode> {
+    /** @hidden */
+    public _loadMeshAsync(context: string, node: INode, mesh: IMesh, assign: (babylonTransformNode: TransformNode) => void): Promise<TransformNode> {
```
bghgary: Does this still need to be exposed?

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
bghgary: There are only 3 buffers. Maybe explicitly declaring them instead of a map would make the code simpler.

loaders/src/glTF/2.0/glTFLoader.ts
Drigax: No, I can revert this change.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: I'm not sure if we get much out of removing the dict, since I'll have to change how we assign our loadFloatAccessor result via loadAttribute

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: ah, you're absolutely right here. I'm deep copying the meshes instead of instantiating them.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: Good place to put this, I'm sure its usable elsewhere.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: I'm not too sure if that's guaranteed: https://github.com/BabylonJS/Babylon.js/blob/master/src/Meshes/instancedMesh.ts#L48-L50

In the case of our sample scene, we have two instances of the same mesh, each with their own world matrix. when we import this scene, the second teapot mesh is created as an instance.

When we create an instance of the instance's source mesh, we would have a mesh with world matrix = the source mesh, right? we should still need to set this.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: good idea, I'll run this through the formatter.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: ok, replaced with explicit calls

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: formatted.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
Drigax: I decided to explicitly declare them, as suggested.

loaders/src/glTF/2.0/Extensions/KHR_mesh_instancing.ts
```diff
@@ -0,0 +1,111 @@
...
+                let translationBufferAccessor: Nullable<IAccessor> = null;
+                let rotationBufferAccessor: Nullable<IAccessor> = null;
+                let scaleBufferAccessor: Nullable<IAccessor> = null;
+
+                loadAttribute("TRANSLATION", (data, accessor) => { translationBuffer = data; translationBufferAccessor = accessor; });
+                loadAttribute("ROTATION", (data, accessor) => { rotationBuffer = data; rotationBufferAccessor = accessor; });
+                loadAttribute("SCALE", (data, accessor) => { scaleBuffer = data; scaleBufferAccessor = accessor; });
+
+                return Promise.all(promises).then(() => {
+                    const anyInstanceAccessor = translationBufferAccessor || rotationBufferAccessor || scaleBufferAccessor;
```
bghgary: Chatted offline. We should probably check to make sure the accessor counts are all the same. One option is to do it in the loadAttribute function and save the count so it can be used here.

Drigax: accessor length error checking in loadAttribute.

PR #7737, opened 2020-02-28
Allow logging of shader code when a compilation error occurs

Popov72:
I find it bothersome to find precisely where an error occurs in a shader when there is a compilation error that pops up in the console log:
* the shader code is not dumped, so you have to find manually in the shader list where the code of the shader is (something not easy if you don't have some knowledge of the inner working of Babylon)
* even if you find the file, the line number given in the error message won't be the right one, as shaders are created by concatening / including several files to produce the final code

So, I have updated the `Effect` class to log the (final) shader code, so that you don't have to find them yourself, and now the line number is meaningful. It can be disabled by doing `Effect.LogShaderCodeOnCompilationError = false` but it is enabled by default as I feel it's not a big deal to have additional lines in the output when there are already quite a number when a compilation error occurs, and it's easier for the user as if an error occurs he already has the full log available (compared to changing `LogShaderCodeOnCompilationError ` to `true` and restarting the program, hoping the error will still pop up).

Note that I also try to extract the offending line from the shader code and dump it on a separate log line, to ease debugging. It assumes the error message from the browser contains something like "[FRAGMENT|VERTEX] SHADER ERROR: 0:XXX:" with XXX being the line number. That's how the message is formatted in Chrome. If it's different in another browser, the line won't be extracted but you will still get the full shader code dumped (you can then copy/paste it in an external editor and look for the erroneous line).


src/Materials/effect.ts
```diff
@@ -612,6 +617,26 @@ export class Effect implements IDisposable {
         }
     }
 
+    private _getShaderCodeAndErrorLine(shader: WebGLShader, error: Nullable<string>, isFragment: boolean): [Nullable<string>, Nullable<string>] {
+        const regexp = isFragment ? /FRAGMENT SHADER ERROR: 0:(\d+?):/ : /VERTEX SHADER ERROR: 0:(\d+?):/;
+        const code = this._engine?._gl?.getShaderSource(shader) ?? null;
```
deltakosh: Effect must not see gl (because for native or webgpu there is no gl). So we must come with a different way.

Probably exposing the getShaderSource on engine

Popov72: Ok, looking into that.

Popov72: I have added `_getShaderSource` on `ThinEngine` instead of `Engine` as it is in this class that the shaders are compiled, so it seemed appropriate(?)

src/Materials/effect.ts
```diff
@@ -626,6 +651,30 @@ export class Effect implements IDisposable {
             return " " + attribute;
         }));
         Logger.Error("Defines:\r\n" + this.defines);
+        if (Effect.LogShaderCodeOnCompilationError) {
+            let pcw = this._pipelineContext as WebGLPipelineContext;
```
deltakosh: We cannot type the context as webgl here as well (sorry :( )

It needs to be the root class so effect has no link with webgl

Popov72: As the shaders only exist in a WebGL context, I have added a type guard: let me know if it's ok for you.

I don't know if it disconnects Effect and WebGL enough, but I can see there are other references to WebGL in this class (`WebGLUniformLocation`, `Engine.webGLVersion` for eg).

src/Materials/effect.ts
deltakosh: Yes!

src/Materials/effect.ts
deltakosh: Ideally I would prefer something that all contexts will be force to provide (the uniforms are only used in the webGL context but your code could be executed if we fail compiling a webgpu shader)

Popov72: Ok, I have updated the `IPipelineContext` interface to be able to get the vertex/fragment code. I think it's what you meant?

PR #7748, opened 2020-03-02
Added the roughness and albedoScaling parameters to PBR sheen

Popov72:
I performed the latest changes we spoke about.

I have also corrected the usage of the `REFLECTION` and `ENVIRONMENTBRDF` defines: in a number of places, `REFLECTION` was checked instead of `ENVIRONMENTBRDF`.


loaders/src/glTF/2.0/Extensions/KHR_materials_sheen.ts
```diff
@@ -86,6 +87,12 @@ export class KHR_materials_sheen implements IGLTFLoaderExtension {
             }));
         }
 
+        if (properties.roughnessFactor !== undefined) {
+            babylonMaterial.sheen.roughness = properties.roughnessFactor;
```
sebavan: Should probably add the default value in the else case to prevent reusing the model roughness in gltf by default ?

Popov72: Change performed

sebavan:
Can you ensure sheen is compiling ok without an environment texture ? just to be sure about the define change for REFLECTION ?

sebavan:
Looks ok but just to be sure :-)

Popov72:
Good idea, there was a bug indeed (forgot to replace one check)!

PR #7762, opened 2020-03-04
Add optional predicates to WebXR laser pointer selection

Exolun:
VR implementation had predicates available for filtering ray and mesh hits, so I pulled over pretty much the same thing into the `WebXRControllerPointerSelection` class.

### DIFF
--- a/src/XR/features/WebXRControllerPointerSelection.ts
+++ b/src/XR/features/WebXRControllerPointerSelection.ts
@@ -163,11 +163,6 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
     public selectionMeshPickedColor: Color3 = new Color3(0.3, 0.3, 1.0);
 
     /**
-     * To be optionaly changed by user to define custom selection logic (after ray selection)
-     */
-    public meshSelectionPredicate: (mesh: AbstractMesh) => boolean;
-
-    /**
      * To be optionaly changed by user to define custom ray selection
      */
     public raySelectionPredicate: (mesh: AbstractMesh) => boolean;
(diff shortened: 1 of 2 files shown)


sebavan:
Looks good to me let s had @RaananW for a second pair of eyes :-)

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -162,6 +162,16 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
      */
     public selectionMeshPickedColor: Color3 = new Color3(0.3, 0.3, 1.0);
 
+    /**
+     * To be optionaly changed by user to define custom selection logic (after ray selection)
+     */
+    public meshSelectionPredicate: (mesh: AbstractMesh) => boolean;
```
RaananW: Please make sure a default value is set, as it seems that everything will fail if those two predicates are not initialized.

`() => true` would be great :-)

RaananW:
Been thinking a bit more about this - we already have placeholder for pointer predicates, and that is the scene. You can use the `pointerDownPredicate` or `pointerMovePredicate`, and set them accordingly, if not set (just like the _onPointerDown and _onPointerMove functions in scene.ts do). This way we reuse and don't add new public variables. What do you think?

src/XR/features/WebXRControllerPointerSelection.ts
Exolun: I think I will go ahead and remove it because for my purposes the `meshSelectionPredicate` isn't needed and perhaps a little redundant!

Though I should point out the way it's written now it won't fail because in an `&&` conditional chain if the first value evaluates the to false the evaluation will short circuit.

Updating to remove!

Exolun:
I don't think I like the idea of using `pointerDownPredicate` or `pointerMovePredicate` as opposed to adding a new public field.  The issue is that users of the Babylon API are unlikely to know intuitively that those predicates also apply to VR picking so I think there is benefit to adding an additional predicate more local to the controller API.

If we remove the `meshSelectionPredicate` and leave in the `raySelectionPredicate` it would mean you could still use `pointerMovePredicate` but the `raySelectionPredicate` would overload it if you found the API in `WebXRControllerPointerSelection` instead!

src/XR/features/WebXRControllerPointerSelection.ts
RaananW: sorry, didn't see the first `this.meshSelectionPredicate && `

RaananW:
The entire idea behind the pointer selection feature is using the pointer system (pointer events etc'). The idea is that a single implementation will work in VR and non-VR scenes, without changing any code.

A predicate still needs to be integrated (as you did in the pickWithRay function), the question is which. As you suggested, if you can do this -it would be great - 

keep the ray predicate, but prioritize the pointerMovePredicate of the scene, if defined. It is undefined per default, so the ray predicate will be used most of the time.

Exolun:
Removed the `meshSelectionPredicate`

So for the ray predicate you would like to see something like this?
`this._scene.pickWithRay(ray, this._scene.pointerMovePredicate || this._raySelectionPredicate)`

RaananW:
Would be great

RaananW:
Thanks!

PR #7765, opened 2020-03-05
Fix.sps.pick and multimaterial

jbousquie:
* Fix the picking issue in theSPS when using the MultiMaterials.
* Expose the correct shapeId value in the positionFunction even when using an immutable SPS
* Added the property `props` (typed `any`, default `null`) so TS users can "extend" the SolidParticle object to manage their own logic with no need for modifying the BJS source code.


sebavan:
Thanks a ton

Issue #7678, opened 2020-02-18
Using the inspector to change colors for PBR materials: convert to linear colors?


Popov72:
As the internal workflow for PBR materials is linear, I think the colors chosen from a color picker should be transformed to linear before updating the material property: what do you think?

That would mean updating the `color3LineComponent` / `color4LineComponent` so that they are able to convert to gamma before showing the color picker with the current color, and to convert to linear before updating the property.

deltakosh:
Agree, we should have an optional props for that

Popov72:
PR done: https://github.com/BabylonJS/Babylon.js/pull/7768

Issue #7143, opened 2019-11-12
Discussion of factors in PBR materials


PatrickRyanMS:
We need to have a conversation about default color spaces for base color/emissive/specular factors for PBR materials. Right now we default to linear space for the factors, where textures for these inputs are in sRGB, So if I assign (128, 128, 128) in a base color texture for a material that sits next to a material that has its base color factor assigned to (128, 128, 128) in code they will look different unless the base color factor is converted to linear space.

I know this has been around since 3.3, but it seems to me that defaulting color space for base color factor to linear space is not intuitive to an artist creating the materials.

sebavan:
We actually did remove them a while back as it was confusing to have both and we wanted to align with GLTF.

Maybe it should be an inspector option only ?

PR #7768, opened 2020-03-05
Inspector: Handle PBR colors as colors in linear space

Popov72:
I have updated `Color3LineComponent` and `Color4LineComponent` with a `isLinear` property that indicates that the color is in linear space (if not present or `false`, it is in sRGB/gamma space).

All colors / color manipulation in the class are handled in gamma space, that's why there is in the constructor:
```typescript
        if (props.isLinear) {
            this.state.color.toGammaSpaceToRef(this.state.color);
        }
```
The only exception is that `this.props.target[this.props.propertyName]` is maintained in the right space, meaning linear space if `isLinear` is `true` and gamma space else.

The purpose is to have a coherent GUI: everything displayed in the GUI is in the same space and it is the gamma space, even for the R/G/B input parameters that you can change manually:
image
If you put 0.5 / 0.5 / 0.5 there, you will get 127/127/127 in the color palette:
image

If the color is linear internally, you will get the linear values only in two ways:
* by clicking on the clipboard icon next to the color
* by clicking on `Generate replay code` in the **Tools** pane

as those are the two ways to get the values defined in the inspector.

Maybe we should show in some way that a color in the inspector is a linear color internally and so one should take care with the r/g/b values that are displayed in the GUI (and notably they should not be used "as is")?

I have updated the PBR color properties to use this new `isLinear` property.


sebavan:
Adding @PatrickRyanMS and @bghgary to validate the concept

Popov72:
The initial issue is here: https://github.com/BabylonJS/Babylon.js/issues/7678

bghgary:
Related: #7143

sebavan:
Lets wait on @PatrickRyanMS final say and merge it in :-)

PatrickRyanMS:
So I've been thinking about this since yesterday and I have questions I can't resolve myself. First I agree that seeing a gamma space color picker while choosing linear values causes some confusion. When I pick this color on the slider, am I really picking a gamma color or a linear color? Are they doing the conversion for me when the value gets back to the inspector? But there is a question going the other way as well: If I pick 127, 127, 127 on this gamma picker, why am I getting 0.5, 0.5, 0.5 in a property I know is linear like roughness? Should I expect the value in the inspector to be in linear space particularly since I saw a gamma space color picker? 

But I think the scenario that makes me more nervous is the idea that when I am entering values manually in the inspector in a linear property like roughness, if I enter 0.5, 0.5, 0.5 because that's the linear value I want, are we doing a conversion as though it's a gamma value? When @Popov72 mentions that the only way to get linear values is through the copy button or the replay feature, it implies to me that we aren't communicating that the inspector is using gamma values even though the input may be a linear input. 

Am I thinking this through correctly, or am I missing something obvious?

sebavan:
@Popov72 I let you answer those :-) and I ll merge once all good and clear with @PatrickRyanMS

Popov72:
@PatrickRyanMS You are totally right in what you are saying.

I'm going to try to be clearer:

### Current problem with the inspector and linear color values
----------
https://playground.babylonjs.com/#RGFQ9J

I have set the emissive color to green=0.5. You can see that the sphere is brighter than the color displayed in the inspector / in the color picker. To be sure, I have taken a screenshot and looked for the green value of the sphere and for the small rectangle showing the emissive color:
* g=184 for the sphere -> corresponds roughly to a g=0.5 after gamma correction, meaning: (0.5^(1/2.2))*255 ~186
* g=125 for the small rectangle -> corresponds roughly to 0.5*255

The problem is that this is a linear color, and when displayed on a monitor it should be gamma corrected to match the final output (which apply gamma correction to the final computed pixels).

=> I think we agree about this state of things?

### How to correct the problem
----------
When the inspector must deal with a linear color, it should gamma encode it before displaying it, to counter the gamma decoding performed by the monitor and to show the real color as it should be.

There's no way around that because the user is dealing with colors (using a color picker) through a monitor which does gamma decoding no matter what: the user is doing implicitely a gamma encoding to counter this effect so that the color he is choosing is the one he really wants. To rephrase it: whenever you deal with colors through a color picker, you are working in gamma encoded space.

So, when the final property is a linear color, the inspector must transform it from gamma space to linear space before storing it into the material and from linear to gamma space before displaying it.

If not for the raw r/g/b values and the "Copy to clipboard" feature, then it would be easy: everything is transparent for the user. He is dealing with gamma encoded colors through the color picker, but it doesn't matter as the inspector is doing the necessary conversions in the background, without the user being aware of it, and without the user having a need to know it. When clicking on "Generate replay code", the right values are generated and he can copy/paste the generated code directly in his program.

### Problem with raw r/g/b values and "Copy to clipboard" feature
These are the two problems:
* r/g/b values are displayeds as raw numbers, and raw numbers have no implicit color space. Here we can choose to display linear encoded values or gamma encoded values. I had chosen gamma encoded values to be consistent with the color picker: the values you see in the r/g/b input fields are the same than the ones you see in the color picker widget (save for the 255 multiplier):
image
However, if the user is copying those values instead of the ones generated by "Generate replay code" and pasting them in his program, then the color will be wrong!
* "Copy to clipboard" feature. It's the same problem than above, to be consistent it generates a "#RRGGBB" triplet based on gamma encoded values. So, if using those values in a Babylon program, the color won't be ok.

So, given all this, should we have the r/g/b input fields + "Copy to clipboard" feature generate gamma or linear encoded values?

Maybe it would be safer to generate linear values in case the user would directly copy them? But then he may wonder why there's no adequacy between the r/g/b raw values he is entering and the values displayed by the color picker... That's why I suggested that maybe we could display the property name in another color (or something else), to show it is a linear color and not a regular color, and that would be a hint for the user that something is different here...

PatrickRyanMS:
So @bghgary and I went through this all today and he reminded me that we followed the glTF spec on linear values for emissive and albedo factors, which is opposite of what we expect with textures, which are assumed to be in gamma space. This becomes an issue because if I am working with gamma values and need to assign a factor, I will have gamma data to start with. So basically the best practice would be to declare your factors like this:

`    mat.emissiveColor = new BABYLON.Color3(0,0.5,0).toLinearSpace();
`

That way we can pass gamma values that we will likely be getting from pickers or texture packages. Of course, unless you know you are working with linear values and know what those colors are, but there are very few places you can go to to see linear color without effort, so I believe most people will be working with gamma values.

For the other issue about what color space the values in the inspector copy with the copy value button, I would fall on the side of copying gamma values because we don't know where the user is going to paste the color. For example, the image below shows the pickers from Chrome and Edge (which obviously match) and then what happens when I used the copy value button and pasted the value into a photoshop color picker. It matches what the other color pickers show which is what I think we should do.  

image

Now for the replay code, I would suggest that we use the color3.toLinearSpace() method like the example above rather than converting to the actual linear values which will help the user understand that the values are indeed in gamma space while still passing the correct values to the engine. Does this make sense?

Popov72:
Thanks for your answer!

So, if I understand it well, the current implementation is ok as the raw r/g/b values and the "Copy to clipboard" feature give the color in gamma space?

The only one thing remaining is:

> Now for the replay code, I would suggest that we use the color3.toLinearSpace()

That's what I intended to do today, as I do also think it will make things clearer for the user!

Popov72:
> So, if I understand it well, the current implementation is ok as the raw r/g/b values and the "Copy to clipboard" feature give the color in gamma space?

My mistake: "Copy to clipboard" was in linear space, going to change that as well.

Popov72:
The changes have been performed:
* "Copy to clipboard" now works in gamma space
* "Replay code" adds a ".toLinearSpace()" when the color is linear

To sum up, there are in fact only a few changes to handle a linear color:
* the color from the material property is transformed to gamma space before being handled by the inspector
* the color is tranformed back to linear space before being stored in the property material
* `.toLinearSpace()` is added in the replay code

That's all.

sebavan:
@PatrickRyanMS let me know if it works for you so I merge it before releasing our next npm THX

PatrickRyanMS:
@Popov72, it sounds like your plan is going to work well and thank you for adding the extra effort for the replay code. @sebavan it feels like this is the right path so I would go ahead and merge when it's ready. Thanks!

sebavan:
@Popov72 one tiny conflict on the merge and good to go :-)

Popov72:
Conflict should be resolved now.

PR #7773, opened 2020-03-07
Additive animation

c-morten:
# Additive Animation Overview

Additive animation is a useful feature present in several 3D engines such as Unity and Unreal Engine. It can be used to layer motions on top of each other to build complex performances from smaller re-usable assets. 

By default animation is considered to be ‘override’ - if you play 2 animations on the same rig simultaneously each animation will have 50% influence over the result. The result is always normalized, so the more animations playing at the same time, the smaller amount of influence each individual animation has over the final result. All of the keyframes in override animations are stored relative to the parent. Say for example you have an object with 2 override animations. The first animation has a translation value of [[0, 0, 0]] at frame 0, then it interplates to [[0, 2, 0]] on frame 30, then back to [[0, 0, 0]] on frame 60. The second animation has a translation value of [[0, 0, 0]] at frame 0, interpolates to [[0, 0, 2]] on frame 30, and then back to [[0, 0, 0]] on frame 60. If you played these animations simultaneously at full weight, frame 30 would result in a translation value of [[0, 1, 1]]. Neither the Y or Z axes would ever be able to fully reach a value of 2 with both animations playing.

Additive animation is unique because it does not use that type of normalization logic. You can have N-number of additive animations playing simultaneously and each one will have the exact amount of influence specified. To accomplish this, additive animation values are relative to the object’s value in the previous animation played in a stack, not the parent. So if the animations in the example above were played additively, frame 30 would result in a value of [[0, 2, 2]] because the second animation’s value adds on top of the first.

There are generally 2 aspects to implementing additive animations. The obvious part is runtime playback, different math is used when evaluating animations additively than is used when evaluating override animations. The other part that needs to be addressed is authoring. Because additive animation on an object is relative to its own space, there is no intuitive way to author these animations for objects in hierarchies. Instead a method of conversion needs to be implemented. This involves specifying a frame or pose to ‘subtract’ out of the other keyframes so they are relative to it.

&nbsp;
&nbsp;
## Additive Animation for Babylon.js

The animation system in Babylon.js only supports override animation by default. By making modifications to existing functionality and adding in a conversion method additive animation can be accomplished:

&nbsp;
### Animations/runtimeAnimation.ts

The addition of an **isAdditive** accessor that returns the _host’s **isAdditive** state allows the Animatable to know whether to evaluate the RuntimeAnimation as additive.

&nbsp;
### Animations/animationGroup.ts

An **_isAdditive** property with **isAdditive** accessors was added to specify whether all Animatables contained in the group should be evaluated as additive.

An **isAdditive** parameter was added to the **start** method to specify whether to evaluate the Animatables being started as additive.

The addition of a **MakeAnimationAdditive** class method allows existing Animation objects’ keyframes to be converted to an additive format that are contained in a given AnimationGroup. It uses **Animation.MakeAnimationAdditive** on all of the AnimationGroup’s targeted animations.

&nbsp;
### Bones/skeleton.ts

The addition of a **MakeAnimationAdditive** class method allows existing Animation objects’ keyframes to be converted to an additive format that are contained in a given Skeleton. It uses **Animation.MakeAnimationAdditive** on all of the Animations referenced by Animatables on the Skeleton.

&nbsp;
### Animations/animation.ts

The addition of a **MakeAnimationAdditive** class method allows existing Animation objects’ keyframes to be converted to an additive format. The reference frame supplied determines the value that resulting keyframes will be relative to. For non-quaternion and non-matrix numeric animations, the value of the reference frame is simply subtracted from every keyframe. For ANIMATIONTYPE_QUATERNION animations, the value at the reference frame is normalized and conjugated. It is then pre-multiplied with every keyframe in the track.  For ANIMATIONTYPE_MATRIX animations, the matrix is first decomposed for each frame. The translation of the reference frame is subtracted. The rotation uses the method described previously for quaternions. The scale is divided by the value from the reference frame. Then the keyframe matrix is composed from the trs objects. Non-numeric keyframes are left un-altered since additive does not have real meaning for these types.

&nbsp;
### Animations/animatable.ts

The addition of an **isAdditive** defines whether the Animatable’s Animations will be evaluated as additive.

An **isAdditive** parameter was added to **beginWeightedAnimation, beginAnimation, beginHierarchyAnimation, beginDirectAnimation,** and **beginDirectHierarchyAnimation** to define whether the Animatable being created should be evaluated as additive.

The **_registerTargetForLateAnimationBinding** method was modified to differentiate between additive and non-additive RuntimeAnimations. Non-additive is stored in the holder’s **animations** array and weight is added to the holder’s **totalWeight** value. Additive is stored in the holder’s **additiveAnimations** array and weight is added to the holder’s **totalAdditiveWeight** property.

The **_processLateBindingsForMatrices** method was modified to differentiate between RuntimeAnimations in the holder’s **animations** and **additiveAnimations** arrays. First the animations array is processed. Then the current value of each non-zero weighted Animation in the additiveAnimations array is decomposed. Translation is added to the translation result from the non-additive animations. The quaternion result from the non-additive animations is multiplied with the additive quaternion. Scale is multiplied by the scale result from the non-additive animations. Then the trs objects are composed to the result matrix.

The **_processLateAnimationBindingsForQuaternions** method was modified to differentiate between RuntimeAnimations in the holder’s **animations** and **additiveAnimations** arrays. First the animations array is processed. Then the result quaternion is multiplied by the current value of each non-zero weighted Animation in the additiveAnimations array. 

The **_processLateAnimationBindings** method was modified to differentiate between RuntimeAnimations in the holder’s **animations** and **additiveAnimations** arrays. First the animations array is processed. Then the current value of each non-quaternion non-matrix Animation in the additiveAnimations array is added to the result.


sebavan:
It all looks good to me but I am not the most well versed into our animation system. @deltakosh will come back in a week and be able to make a way deeper analysis of the change.

About the What s new, the idea is that you add a line into the dist/preview release/whats new.md file to highlight your contribution.

This will help consolidating all the next release work :-)

c-morten:
I created a pr for the What's New update:
https://github.com/BabylonJS/Babylon.js/pull/7784

sebavan:
As explained there, it would be better if you could add the change in this PR. It would prevent:
image

c-morten:
Closed #7784 and moved the change to this PR. However the What's New Update check is still failing. Is there a specific format I need to follow for the addition to what's new.md to get that test to pass?

sebavan:
You modified the wrong what s new file, I happens to me a lot :-)

You need to modify the one in dist/preview release/whats new.md and not the one under dist/whats new.md

Basically the first one is for our next release in preview while the second is the one of our latest stable release. Sorry about that I should have made it more clear. 

Could you simply move the comment from file ?

sebavan:
Should be all good now thanks a ton !!!

The final check will be done by @deltakosh as soon as he ll be back next week.

c-morten:
For sure. I’ll be updating the doc next and will include an example as well.

deltakosh:
thanks a ton! it will be live in a hour or so

c-morten:
Awesome! I’m away from home right now but you can expect a documentation update request by EOD Thursday.

deltakosh:
I love your PR...please send more ;D

c-morten:
I just might. Babylon’s animation system was really fun to play around with, really impressive framerate!

Issue #7758, opened 2020-03-04
Allow alpha testing + alpha blending at the same time


Popov72:
See this post in the forum for context: https://forum.babylonjs.com/t/meshes-transparency/8965/9

Currently you can't have alpha testing and alpha blending at the same time for the standard material  (when alpha values come from the diffuse texture) => it is possible with the PBR material.

See this PG: https://playground.babylonjs.com/#68RZKR#3

You have alpha blending on both spheres (left: std material, right: pbr) and not alpha testing. Both are enabled, meaning `mat.needAlphaBlending()` / `mat.needAlphaBlendingForMesh()` and `mat.needAlphaTesting()` both return `true`, but only alpha blending is enabled for the rendering because of this function (defined on `Material`):

```typescript
protected _shouldTurnAlphaTestOn(mesh: AbstractMesh): boolean {
    return (!this.needAlphaBlendingForMesh(mesh) && this.needAlphaTesting());
}
```
Because of this function, when you have both effects on, alpha blending takes precedence and alpha testing is disabled.

We could change this function to:
```typescript
protected _shouldTurnAlphaTestOn(mesh: AbstractMesh): boolean {
    return this.needAlphaTesting();
}
```
to have both effects at the same time: try to uncomment line 38 in the PG and you will see it does work.

HOWEVER, how to have alpha blending without alpha testing with this change? I think it is not possible (at least when we want the alpha values to come from the diffuse texture)...

Maybe I missed something and it is possible to have alpha blending without alpha testing with this change, in which case I can make the PR right away.

But if it's not possible, how to handle that? Should we add a property like `enableBothAlphaBlendingAndTesting` to the material and implement `_shouldTurnAlphaTestOn` as:

```typescript
protected _shouldTurnAlphaTestOn(mesh: AbstractMesh): boolean {
    return (!this.needAlphaBlendingForMesh(mesh) && this.needAlphaTesting() && !this.enableBothAlphaBlendingAndTesting)
            ||
        (this.needAlphaTesting() && this.enableBothAlphaBlendingAndTesting);
}
```
? Bonus: it is not a breaking change.

Or something else?

Note that for PBR it is possible to do whatever we want regarding alpha blending / alpha testing mixing thanks to the `transparencyMode` property (uncomment line 57 in the PG). So whatever we decide for the standard material, I think the PBR material shouldn't be impacted.

sebavan:
[...]

PR #7775, opened 2020-03-07
Add transparencyMode property to StandardMaterial

Popov72:
See issue #7758.

I have added `transparencyMode` to the `Material` base class so that it can also be used by `StandardMaterial` (was only available for `PBRMaterial` until now).

See accompanying sample: https://popov72.github.io/BabylonDev/showSample.html?sample=transparency

I have updated `OptionsLineComponent` to be able to deal with `null` values, as `transparencyMode` can be `null`.

`StandardMaterial` and `PBRMaterial` work about the same when using `transparencyMode`, except:
* When alpha test is enabled (`transparencyMode`is **Alpha Test** or **Alpha Test+Blend**) and **Use alpha from texture** is unchecked.
  * In that case, `PBRMaterial` still uses the alpha from the texture, contrary to `StandardMaterial`. I think `PBRMaterial` is wrong here, but as it is how it works now and changing this would be a breaking change, I kept it => I didn't want to change how `transparencyMode` is behaving for `PBRMaterial` in this PR, but see `strictTransparencyMode` below

Odd behaviour (share by both materials):
* select **Alpha Test** and make sure that **Use alpha from texture** is checked
  * the spheres display correctly:
image

* now change **Mesh visibility** to something different from 1: spheres no longer do alpha testing but alpha blending instead:
image
  * that's because the code switches automatically to alpha blending for the rendering to honor the **mesh visibility** setting. The problem is that now the value for `transparencyMode` does not match what we are seing...

There are other cases like that, for eg. when choosing **Opaque** and changing **Mesh visibility**.

Sometimes I find it hard to understand what's going on depending on your choices for the different properties, that's why I have added a `strictTransparencyMode` which strictly (hence the name) applies the choice of the user for the `transparencyMode` and `useAlphaFromDiffuseTexture` / `useAlphaFromAlbedoTexture` properties.

You can use the [sample](https://popov72.github.io/BabylonDev/showSample.html?sample=transparency) to test this, by checking the **Strict transparency mode** checkbox. I do think that in this mode everything is behaving as expected, depending on your choices.

Let me know what you think about this and how we can going on - I don't mind throwing this to the bin, so don't hesitate to comment :)


sebavan:
I think we should remove the strictTransparencyMode as it is adding extra complexity we do not need so far.

Visibility is meant to force alpha blending as it is the way to slowly fade models.

All the rest looks ok to me.

Popov72:
`strictTransparencyMode` has been removed.

sebavan:
Perfect thanks a lot.

PR #7789, opened 2020-03-10
Recast.js To Ref Updates

MackeyK24:
Recast.js plugin and crowd to ref performance optimizations. Big performance gain when used in game loops. (Does not create new Vector3 instances per frame)


sebavan:
It will go in as soon as you can fix the conflict on the what s new :-)

Thanks a lot

sebavan:
@MackeyK24 could you update your PR to solve the conflict so that I can merge it in ?

MackeyK24:
I don’t get why is conflict. All I did was add a line about recast changes

sebavan:
I fixed it for you, basically what s new was changed on both sides.

MackeyK24:
Thank you so much

PR #7810, opened 2020-03-12
Add an option to optimize collision detection performance.

jsdream:
This brings an option to iterate over a defined list of meshes when testing collision detection of a moving mesh, instead of iteration over all meshes in the scene. That gives a significant performance boost in scenarios when there are many meshes in the scene and it is known which exactly meshes are currently close to the moving mesh (e.g. with what meshes collision is theoretically possible). This improvement has shown really good results in the game with a tile-based world.

More details about use case are available in this forum thread: https://forum.babylonjs.com/t/willing-to-extend-movewithcollisions-method/9127/9

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -46,6 +46,9 @@
 ### WebXR
 - Added optional ray and mesh selection predicates to `WebXRControllerPointerSelection` ([Exolun](https://github.com/Exolun))
 
+### Collisions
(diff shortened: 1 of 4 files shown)


sebavan:
Could you add an entry in the dist/preview release/what s new.md file ??? this will ensure the build is successfull.

Thanks a lot @jsdream

sebavan:
Scratch that, just seen your update :-)

jsdream:
@sebavan Let me know if I have missed anything else as this is my first PR to the BabylonJS codebase ;-)

PR #7811, opened 2020-03-12
Use getAbsolutePosition to ensure world matrix is up-to-date

bghgary:
https://forum.babylonjs.com/t/how-intersectsmesh-works

### DIFF
--- a/src/Meshes/transformNode.ts
+++ b/src/Meshes/transformNode.ts
@@ -1238,7 +1238,7 @@ export class TransformNode extends Node {
             camera = (<Camera>this.getScene().activeCamera);
         }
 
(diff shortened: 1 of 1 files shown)


BabylonJSGuide:
How big a breaking change might this cause?

bghgary:
> How big a breaking change might this cause?

Hmm, fair question. I would hope no one is depending on these functions to use an incorrect world matrix. That seems very unlikely. @sebavan what do you think?

sebavan:
I think none at all as well here, just fixing things :-)

BabylonJSGuide:
OK just wanted to be sure :)

sebavan:
I totally agree it is better to be safe here, you did well to check.

PR #7818, opened 2020-03-13
Add constructor options and equals method to FresnelParameters + tests.

brianzinn:
https://forum.babylonjs.com/t/fresnelparameters-ctor/9108/4


brianzinn:
I didn't think there was a What's New worth mentioning :)

src/Materials/fresnelParameters.ts
```diff
@@ -28,22 +89,32 @@ export class FresnelParameters {
...
-    public bias = 0;
+    public bias: number;
 
     /**
      * Defined the power exponent applied to fresnel term
      */
-    public power = 1;
+    public power: number;
+
+    public constructor(options: IFresnelParametersCreationOptions = {}) {
```
sebavan: You need to add a comment on the constructor as we do in the other classes for the build to be ok :-)

src/Materials/fresnelParameters.ts
```diff
@@ -57,37 +128,47 @@ export class FresnelParameters {
         return newFresnelParameters;
     }
 
+    /**
+     * Determines equality between FresnelParameters objects
+     * @param otherFresnelParameters defines the second operand
+     * @returns true if the power, bias, leftColor, rightColor and isEnabled values are equal to the given ones
+     */
+    public equals(otherFresnelParameters: DeepImmutable<FresnelParameters>): boolean {
```
sebavan: If it is only for tests, should we introduce it here ?

brianzinn: I added `equals(...)` as I use it for diffing out changes when I reconcile changes and need to update the BabylonJS scene.  If you think they are not useful then I can remove - I am okay to check all properties individually when I diff.

PR #7832, opened 2020-03-15
Edit NME Frames Port Labels

belfortk:
• Removed some extraneous console.logs from preview PR
• Added Port Label editing Fields to Frame Property Component
• Added ability to edit port labels


dist/preview release/what's new.md
```diff
@@ -14,6 +14,8 @@
 - Simplified code contributions by fully automating the dev setup with gitpod ([nisarhassan12](https://github.com/nisarhassan12))
 - Add a `CascadedShadowMap.IsSupported` method and log an error instead of throwing an exception when CSM is not supported ([Popov72](https://github.com/Popov72))
 
+- Can now rename NME frames inputs and outputs ([Kyle Belfort](https://github.com/belfortk)
```
deltakosh: We should add a NME chapter and gather all info about NME there

deltakosh:
I think this PR is fine but I would like to merge only when we will be able to serialize/deserialize the names

deltakosh:
Also I think the renaming should be when we click on a node and not through a list of properties in the frame properties

Pinging @PatrickRyanMS and @PirateJC for thoughts.

So far, @belfortk proposal is to have all the port in the property windows:
image

I think we should have specific property for each port instead (Because in the future we will probably rename them but also activate/desactivate them or even add comments)

belfortk:
@deltakosh are you thinking we should or are going to have a nodePortPropertyComponent, similar to the framePropertyComponent?

deltakosh:
yes I think so

PatrickRyanMS:
As I'm looking at the example, it is confusing to have all of the inputs and outputs labeled as "port" in an order that doesn't seem to have a pattern. What happens if I add more nodes into the frame, how are they ordered as inputs/outputs? And what if I want to add an output to be used at a later time (if I were sharing the node or creating an uber node for a project)? Something like this:

image

Where I want to have an out for the color, but also add an output for the lerp node because I want to be able to tap into the output of the lerp for some other reason. How do I add that as an output without needing to wire it up to something, collapse the node, and then break the wire? 

I'm not sure how we have control over the inputs and outputs in this model other than what is wired into and out of the frame or what has an open input or output within the frame. What is the PoR for that scenario? This second part of controlling the available inputs and outputs may be beyond the scope of this change to allow inputs and outputs to have a custom name, but I think how we decide to add inputs and outputs could impact this system as well, i.e. how we rename the inputs/outputs or where that control is stored. Is it on the frame or on the individual node like "visible in inspector".

deltakosh:
- For sorting, I think the Port property window should provide a text box to rename it and two buttons to move the output up or down (easy to integrate from a dev stand point)
- For availability as an output, maybe we can have a list of all outputs and inputs of the frame on the frame property window so we can toggle their visibility (by default only unwired ports are exposed)

PatrickRyanMS:
If we had a list of all inputs/outputs that aren't wired, we could get into a lot of cleanup as well. And a list of all properties to show/hide visibility could get complex without names for every node (if I had two add nodes for example, which output am I picking). 

Could we solve this with the node graph itself by adding a "port" to wire to to expose a node like this:

image

The wire to the input or output would not need to stay, but just be a temporary connection so we can expose a new input/output. It wouldn't even need to be two ports (one for in and out for out) as we would be able to tell which one was wired as this is only exposing the input/output. I think this is similar to something Jason brought up. 

My main concern is that in the sample above, the color splitter and color merger add a lot of potential for unwired inputs/outputs and we need an easy way to mark them for the frame in/out. We should also do a study about how to reveal to the user that an input/output is exposed in the graph. Is there another level of shape behind the node aligned with the input/output to show that it is exposed, again to help with knowing which ones are exposed at a glance? Also what about things like UV in the sample above... can I expose the UV of the texture as a potential input and if it's wired, the UV in the frame gets overridden by the connected frame input? That could open up a pandora's box, but I could see shared nodes needing to have flexibility.

deltakosh:
Above everything my goal is to keep it simple even if they are some "acceptable" flaws. We wont be able to sort out everything (not with a limited team like ours).

That being said, I understand the overall issue (but I do not want to go down the path of specific nodes because it will be a nightmare to support).

Having the two Add input / Add output things will not solve the sorting/ renaming issues. I feel like for this point having a property grid per port exposed by the frame works well.

The only question that remains is how to get that list as taking the non connected ports will not work.

Maybe for ports INSIDE a frame we can add a property for the ports as well (like you said earlier, something like expose outside the frame) ?

PR #7835, opened 2020-03-15
Fix horizon occlusion in PBR materials

Popov72:
Also fixed a bug (material does not compile) when anisotropy is enabled and no bump texture is provided.

Horizon occlusion before fix:
image

Horizon occlusion after fix:
image


sebavan:
looks like some tests are failing, probably some combinations failing.

Popov72:
The test "GLTF Material Double Sided (Back)" fails.

I have pushed a correction when two sided lighting is enabled.

Popov72:
So, the tests passed but I can clearly see that the pictures rendered by "GLTF Material Double Sided (Back)" and "GLTF Material Metallic Roughness (0)" are different from the current renderings, even if not different enough to trigger an error: shouldn't we regenerate the validation screenshots to be on the safe side? I would do it, but I don't really know how to generate the 600x400 pictures expected by the validation process...

sebavan:
You need to run locally : localhost.../tests/validation

Then right click the generated picture and save as :-)

PR #7855, opened 2020-03-18
setTexture and setTextureArray from ShaderMaterial take now a BaseTex…

Popov72:
…ture as input instead of a Texture, allowing to pass a CubeTexture.

See https://forum.babylonjs.com/t/how-can-i-get-each-side-of-cubetexture/2830


epascal:
Thank you for your reactivity.

PR #7863, opened 2020-03-19
Rename and Re-order NME Frame Ports

belfortk:
Implement the ability to rename and re-order NME Frame Ports

• Removed some extraneous console.logs from a previous PR
• Add nodePortPropertyPane component with Port Label editing Fields and re-ordering buttons
• Added ability to edit port labels

I'm not 100% sure that adding two new Observables to the global state to handle reacting to clicking the re-ordering buttons was the best implementation but I couldn't come up with a better way. I also added a GraphFrame._FramePortCounter and the concept of framePortId because I needed a way to keep track of unique framePorts. I thought I could use blockId's, but sometimes multiple ports point to the same block. 

Open to any feedback/suggestions.


nodeEditor/src/diagram/properties/nodePortPropertyComponent.tsx
```diff
@@ -0,0 +1,44 @@
...
+                <img id="logo" src="https://www.babylonjs.com/Assets/logo-babylonjs-social-twitter.png" />
+                <div id="title">
+                    NODE MATERIAL EDITOR
+                </div>
+            </div>
+            <div>
+                <LineContainerComponent title="GENERAL">
+                    <TextInputLineComponent globalState={this.props.globalState} label="Port Label" propertyName="portLabel" target={this.props.nodePort}/>
+                    <ButtonLineComponent label="Move Node Up" onClick={() => {
+                                this.props.globalState.onFramePortMoveUpObserver.notifyObservers(this.props.nodePort);
```
deltakosh: I would recommend having these as observables on the NodePort. They should trigger when the UI changes the nodePort position.

Then the frame UI will register to these observables to update visually

PR #7871, opened 2020-03-20
Add a PBR shader code coverage test

Popov72:
PG used is: https://playground.babylonjs.com/#QI7TL3#3

It is a code coverage, so the aim is to have all code paths executed at least one time when all the spheres are displayed.

The aim is *NOT* to test each combination of PBR parameters and make sure there's no bug! It's too big a task to try to make such a validation test...

The change in `validation.js` is to be sure to see a wrong pixel even when its alpha is very small / zero.


Popov72:
It seems there are two errors in the WebGL2 validation tests:
* the new test coverage fails with:
  * Timeout of 180000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
* Instanced Bones fails with:
  * Firefox 74.0.0 (Ubuntu 0.0.0) Validation Tests Instanced Bones FAILED
	expected false to be true
	AssertionError@node_modules/chai/chai.js:9449:13
	[3]</module.exports/Assertion.prototype.assert@node_modules/chai/chai.js:239:13
	[5]</module.exports/<@node_modules/chai/chai.js:1021:10
	propertyGetter@node_modules/chai/chai.js:7899:29
	proxyGetter@node_modules/chai/chai.js:9247:22
	@tests/validation/integration.js:54:39
	evaluate@http://localhost:3000/tests/validation/validation.js:128:9
	processCurrentScene/</<@http://localhost:3000/tests/validation/validation.js:146:29
	./Engines/engine.ts/Engine</Engine.prototype._renderFrame@http://localhost:3000/dist/preview%20release/babylon.max.js:33930:13
	./Engines/engine.ts/Engine</Engine.prototype._renderLoop@http://localhost:3000/dist/preview%20release/babylon.max.js:33945:26

However, those two tests do work locally for me(?)

deltakosh:
@sebavan can you check what the errors are?

sebavan:
@Popov72, can you merge master back in your branch ??? I wonder if you forked when we had test failures yesterday ?

Instances Bones should not be impacted by your changes so I guess it could be that.

sebavan:
Also you might need to split Tests PBR shader code coverage as it seems to be taken too much time in CPU emulation (our servers do not have gpu for tests.)

sebavan:
Let s see after those 2 actions if there are still issues.

Popov72:
Done, let's see the outcome.

Popov72:
Some errors not related to my code it seems:
* in "Typescript all"
[15:42:54] Error in plugin "webpack-stream"
Message:
    [at-loader] ../../nodeEditor/src/diagram/graphCanvas.tsx:407:27 
    TS2339: Property 'type' does not exist on type '{ x: number; y: number; width: number; height: number; }'.
[at-loader] ../../nodeEditor/src/diagram/graphCanvas.tsx:409:47 
    TS2339: Property 'id' does not exist on type '{ x: number; y: number; width: number; height: number; }'.
[at-loader] ../../nodeEditor/src/diagram/graphCanvas.tsx:420:44 
    TS2339: Property 'id' does not exist on type '{ x: number; y: number; width: number; height: number; }'.
Details:

sebavan:
oh you might need to wait for the next nightly in 15 minutes before merging back in I guess.

Popov72:
Same problem, the 3 tests are in time out and the Instanced Bones test fails in the same way.

Going to test by removing the 3 new tests and see how it comes, at least for Instanced Bones.

Popov72:
No more error with "Instanced Bones" when the 3 tests are removed...

Re-adding them, but adding a new texture to the playground: I was reading a texture from outside and had a warning in Chrome (main site using https but not this picture).

Popov72:
Still does not work...

Are you able to monitor more closely one of the test that times out to see if it is really the rendering that takes 3mn or if it is something else?

And I don't understand why "Instanced bones" fails when those tests fail...

sebavan:
Do not look into instances bones it is a side effect of time out on other tests.

Are you relying on promises which could potentially never exit or things like this in the tests? This could lead to exactly this.

sebavan:
Yup I am seeing lot of whenAllReady. Could you try to replace them with normal assignements ?

sebavan:
Actually it might be as simple as https://playground.babylonjs.com/textures/ParallaxNormal.png ???

image

preventing to end the tests

sebavan:
So looking into it I am not seeing any obvious reasons... :-(

I will try to repro locally

Popov72:
> Actually it might be as simple as https://playground.babylonjs.com/textures/ParallaxNormal.png ???
> 
> image
> 
> preventing to end the tests

Yes, that texture is new with this PR. Before that, I was loading it from an external site, but Chrome was issueing a warning so I thought I would put it locally.

Popov72:
> Yup I am seeing lot of whenAllReady. Could you try to replace them with normal assignements ?

I can't really do that because if I don't wait the texture to be loaded some assignments will fail, because they end up being assigned to Texture._texture, and this property is not defined until the texture is really loaded... There's no 404 in the playground, so all those textures should be loaded in the end and not block in a promise.

For your testing, you can replace `textures/ParallaxNorma.png` by `http://i.imgur.com/zVGaZNi.png`.

sebavan:
Yup I can not explain it for now as it looks all great locally...

sebavan:
Ok so running the tests tasks locally has the same result as the server. You can run gulp to repro, you ll notice "Not a babylon environment map" in the log.

sebavan:
looks like Karma, does not like .env from the playground... still investigating...

sebavan:
@Popov72 it looks like replacing .env by .dds in the 3 playgrounds fixes it. You can make a new push with playground ids:

#QI7TL3#16
#QI7TL3#17
#QI7TL3#18

I am trying to understand why when running in Karma .env are not supported.

tests/validation/config.json
```diff
@@ -677,6 +677,21 @@
             "referenceImage": "pbr_refraction.png"
         },
         {
+            "title": "PBR shader code coverage #1",
+            "playgroundId": "#QI7TL3#12",
```
sebavan:
```suggestion
            "playgroundId": "#QI7TL3#16",
```

tests/validation/config.json
```diff
@@ -677,6 +677,21 @@
...
+            "referenceImage": "pbr_codecoverage1.png"
+        },
+        {
+            "title": "PBR shader code coverage #2",
+            "playgroundId": "#QI7TL3#13",
```
sebavan:
```suggestion
            "playgroundId": "#QI7TL3#17",
```

tests/validation/config.json
```diff
@@ -677,6 +677,21 @@
...
+            "referenceImage": "pbr_codecoverage2.png"
+        },
+        {
+            "title": "PBR shader code coverage #3",
+            "playgroundId": "#QI7TL3#14",
```
sebavan:
```suggestion
            "playgroundId": "#QI7TL3#18",
```

sebavan:
Yup so karma relies internally on https://www.npmjs.com/package/isbinaryfile and we end up in this part of the code: https://github.com/gjtorikian/isBinaryFile/blob/da5d0e2039e7133e4bf8459a90713c9b9e567488/src/index.ts#L124

so as our env files includes json data near the header, the percentage of text is misleading karma to think it is a text so serving it as utf-8 and breaking our parser, so please, do not use local .env in tests and prefer their equivalent .dds files.

This bug drove me nuts :-)

Popov72:
Good catch!

Hoping this time everything will be ok...

Popov72:
Alleluia!

PR #7874, opened 2020-03-20
Add Feedback on Frame Port Selection. Add hiding behavior to move port buttons. 

belfortk:
Implementing Feedback on PR #7863

• Added a white border to Frame Ports on selection
• Moved the onFramePortMove observers to the NodePort class
• Now passing Frame Port Position data to NodePort so that NodePortPropertyComponent knows when to hide move buttons

Given the amount of special behavior/code needed for Frame Ports so far (and presumably so in the future), I think it might be worth creating a new graphFramePort class that extends NodePort. Thoughts? If I could start all over again, I would have given that approach more consideration.


deltakosh:
Please do! we have time and I agree that this approach will be better

Issue #7875, opened 2020-03-20
Updating FPS in immersive XR is wasted effort


jdm:
When entering immersive XR mode on the playground, the FPS counter is continually modified but is not visible. This is wasted effort when attempting to draw immersive frames as fast as possible; in Firefox Reality on HoloLens devices this causes layout to occur which ends up hurting performance.

- Bug repro on [playground](https://playground.babylonjs.com): https://playground.babylonjs.com/#8Y780Y#57
- Expected result: The DOM shouldn't be modified while in immersive mode.
- Current result: [The DOM is modified](https://github.com/BabylonJS/Babylon.js/blob/3cfa4edeba01cf9621fad672e284f1d3a3fb483b/Playground/js/frame.js#L121-L123) while in immersive mode.

PR #7876, opened 2020-03-20
No fps update in playground when in XR

RaananW:
Addressing #7875


Playground/js/main.js
```diff
@@ -164,7 +164,12 @@ compileAndRun = function (parent, fpsLabel) {
                     sceneToRender.render();
                 }
 
-                fpsLabel.innerHTML = engine.getFps().toFixed() + " fps";
+                // Update FPS if camera is not a webxr camera
+                if(!(sceneToRender.activeCamera && 
+                    sceneToRender.activeCamera.getClassName && 
+                    sceneToRender.activeCamera.getClassName() === 'WebXRCamera')) {
```
deltakosh: I would have expected a !== ?

RaananW: this is the condition of an existing active camera, and that it is an xr camera.

If the scene has activeCameras (plural) a `!==` would return false, since I am checking for activaCamera existence

deltakosh: ok just me being dumb (not seeing the ! at the beginning ;))

RaananW:
No idea why it all failed, there are 0 bjs code changes in this PR

paulrouget:
@RaananW when will this make it to the live version of playground?

RaananW:
It is already there. Try a hard refresh (due to new caching policies we have recently added)

paulrouget:
@RaananW is that fix supposed to also work with https://playground.babylonjs.com/frame.html ?

paulrouget:
@RaananW for example: https://playground.babylonjs.com/frame.html#ZNX043#15

RaananW:
Oh, that's the frame.

If you need a full screen mode, you can use full.html instead - https://playground.babylonjs.com/full.html#ZNX043#15 , but I will fix it for the frame as well

PR #7883, opened 2020-03-20
Ensure that TextureAssetTask does not invert textures by default

broederj:


src/Misc/assetsManager.ts
```diff
@@ -509,7 +514,7 @@ export class TextureAssetTask extends AbstractAssetTask implements ITextureAsset
         /**
          * Defines if texture must be inverted on Y axis (default is false)
          */
-        public invertY?: boolean,
+        invertY: boolean = false,
```
deltakosh: well it seems this parameter is not used then

src/Misc/assetsManager.ts
```diff
@@ -520,6 +520,8 @@ export class TextureAssetTask extends AbstractAssetTask implements ITextureAsset
          */
         public samplingMode: number = Texture.TRILINEAR_SAMPLINGMODE) {
         super(name);
+
+        this.invertY = invertY;
```
deltakosh: so why do you not let the public invertY and just set it up to false by default?

broederj: Thanks, yes :D

broederj:
Great! Now let's see how many people will feel this is a breaking change ;-) It could be for some implementations, I guess.

deltakosh:
Does not categorize as breaking changes as it was a bug :)

deltakosh:
Well now that Im thinking more about it, naybe we should set it to true by default and update the doc

So we will not break anything!

I'll do the change

PR #7886, opened 2020-03-21
Allow parenthesis usage in #if expressions in shader code

Popov72:
See https://forum.babylonjs.com/t/bug-when-parsing-some-if-in-shaders/9329

Parenthesis were not handled in `#if` expression parsing.

In the new implemention, the expression is first converted to postfix notation to easily generate the tree of `ShaderDefineExpression` instances corresponding to the input expression.

Note I first replace the `defined(...)` strings to `defined[...]` before converting to postfix as the converting code handles parenthesis as special characters (which is the point of this PR after all).


src/Engines/Processors/Expressions/shaderDefineExpression.ts
```diff
@@ -3,4 +3,96 @@ export class ShaderDefineExpression {
...
     }
+
+    private static OperatorPriority: { [name: string]: number } = {
+        ")": 0,
+        "(": 1,
+        "||": 2,
+        "&&": 3,
+    };
+
+    private static Stack = ['', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', ''];
```
deltakosh:
```suggestion
    private static _Stack = ['', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', ''];

```

Popov72:
Do you want me to perform the Stack to _Stack changes (in another PR as this one is merged)?

deltakosh:
Yes please :)

PR #7887, opened 2020-03-23
Don't update FPS when in Frame and in XR

RaananW:
Just like #7876 but for frame.html

### DIFF
--- a/Playground/js/mainWebGPU.js
+++ b/Playground/js/mainWebGPU.js
@@ -154,9 +154,9 @@ compileAndRun = function(parent, fpsLabel) {
                     }
 
                     // Update FPS if camera is not a webxr camera
-                    if(!(sceneToRender.activeCamera && 
-                        sceneToRender.activeCamera.getClassName && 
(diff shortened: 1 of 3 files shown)


Playground/js/mainWebGPU.js
```diff
@@ -153,7 +153,12 @@ compileAndRun = function(parent, fpsLabel) {
                         scene.render();
                     }
 
-                    fpsLabel.innerHTML = engine.getFps().toFixed() + " fps";
+                    // Update FPS if camera is not a webxr camera
+                    if(!(sceneToRender.activeCamera && 
+                        sceneToRender.activeCamera.getClassName &&
```
Hersir88: What about creating util function to check if camera is WebXRCamera, as I see its used in multiple places ?

RaananW: There is no real need for this complexity. I have added a few defensive checks, which are actually not entirely needed, and are only valid in this specific page(s).
Want to check if it is a WebXR camera? check get getClassName() function, or check instanceof (that would be a bit slower).

PR #7912, opened 2020-03-25
Add Feedback on Frame Port Selection. Add hiding behavior to move port buttons. 

belfortk:
Implementing Feedback on PR #7863 and #7874

• Added a white border to Frame Ports on selection
• Created new FrameNodePort class that extends NodePort
• Moved the onFramePortMove observers to the FrameNodePort class
• Now passing Frame Port Position data to FrameNodePort so that FrameNodePortPropertyComponent knows when to hide move buttons
• Fixed bug where deleting a node from a frame would not remove its ports from the frame


deltakosh:
Thanks buddy! Testing it right now

belfortk:
Looking forward to the feedback

PR #7914, opened 2020-03-25
Fix CustomMaterial and PBRCustomMaterial

Popov72:
In some cases, the uniforms / samplers / attributes are not set correctly, see https://forum.babylonjs.com/t/custommaterial-bug-uniforms-cannot-be-updated-for-meshes-that-have-a-vertex-color-property/9478


Popov72:
WebGL1 visual test "Texture cache" failed but it has nothing to do with my changes...

The custom materials are not involved in this test.

sebavan:
Trying to run it again to see

PR #7920, opened 2020-03-26
Device Input System - Step 1

PolygonalSun:
This PR contains the web implementation for the DeviceInputSystem class.  This class will take input from gamepads, mice, and keyboards and allow someone to access (via pollInput) their input values in the same manner.

While it isn't the full version of this system, it's the first step.  I'd like to get it checked in so that I can get feedback to create a solid foundation for the next level up on this system.

To use:

Create a variable to hold the device input system and use onDeviceConnected and onDeviceDisconnected to provide custom behavior.

Example:

```
var dis = new BABYLON.DeviceInputSystem(canvas); // Canvas object needed to attach pointer
dis.onDeviceConnected((deviceName) => {
     if (deviceName == BABYLON.DeviceInputSystem.POINTER_DEVICE)
     {
          scene.onBeforeRenderObservable.add(() => {
               if (deviceName == BABYLON.DeviceInputSystem.POINTER_DEVICE &&
                    dis.pollInput(deviceName, 2))
                    {
                         console.log("Left Click");
                    }
          });
});
```

Note: This code is not expected to be used by end user but by another system.


src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,223 @@
...
+        });
+
+        this._elementToAttachTo?.addEventListener("pointerdown", (evt) => {
+            if (!this._mouseActive) {
+                this._mouseActive = true;
+                this.registerDevice(DeviceInputSystem.POINTER_DEVICE, 5);
+                this._onDeviceConnected(DeviceInputSystem.POINTER_DEVICE);
+            }
+
+            let mouseButton = this._inputs.get(DeviceInputSystem.POINTER_DEVICE);
```
deltakosh: I would call it pointerBUtton to be aligned with the event name

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,223 @@
...
+            if (kbKey) {
+                kbKey[evt.keyCode] = 0;
+            }
+        });
+    }
+
+    /**
+     * Handle all actions that come from mouse interaction
+     */
+    private handleMouseActions() {
```
deltakosh: same here for mouse instead of Pointer

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,223 @@
...
+
+            this._inputs.set(deviceName, device);
+        }
+    }
+
+    /**
+     * Given a specific device name, remove that device from the device map
+     * @param deviceName Name of device to be removed
+     */
+    private deregisterDevice(deviceName: string) {
```
deltakosh: deRegisterDevice ?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Handle all actions that come from keyboard interaction
+     */
+    private handleKeyActions() {
+        window.addEventListener("keydown", (evt) => {
+            if (!this._keyboardActive) {
+                this._keyboardActive = true;
+                this.registerDevice(DeviceInputSystem.KEYBOARD_DEVICE, 222);
```
deltakosh: why 222 ?

PolygonalSun: I found that Javascript had 222 keycodes for input from the keyboard.  I didn't find a way to get that value programmatically so I used the hard value.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: It should just be one word for the sake of camelcasing though I can change it to _deRegisterDevice_ if it helps with readability.

src/DeviceInput/deviceInputSystem.ts
deltakosh: what about pointer IDs? I would recommend using a negative value to be safe

src/DeviceInput/deviceInputSystem.ts
deltakosh: Lol you know english better than me ;D

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Changed all mouse refs to pointer

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Changed all mouse refs to pointer

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Since we're not using pointer ids, I believe that this isn't applicable.  I have pulled out the numbers into separate variables.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * Handle all actions that come from pointer interaction
+     */
+    private handlePointerActions() {
+        this._elementToAttachTo?.addEventListener("pointermove", (evt) => {
+            if (!this._pointerActive) {
+                this._pointerActive = true;
+                this.registerDevice(DeviceInputSystem.POINTER_DEVICE, this._maxPointerButtons);
```
deltakosh, line 129: I struggle to understand then how this works with multi touch? each pointer has an id to differenciate them and I was expecting the system to handle it

PolygonalSun: Ah...  Now I'm following the pointer id thing.  As far as the current iteration of the code, it wouldn't support multi-touch but at this step that'd be easy enough to add.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+ * This class will take all inputs from Keyboard, Pointer, and
+ * any Gamepads and provide a polling system that all devices
+ * will use.  This class assumes that there will only be one
+ * pointer device and one keyboard.
+ */
+export class DeviceInputSystem {
+    // Private Members
+    private _inputs: Map<string, Array<number>> = new Map();
+    private _onDeviceConnected: (deviceName: string) => void = () => { };
+    private _onDeviceDisconnected: (deviceName: string) => void = () => { };
```
syntheticmagus, line 14: Should these be [`Observable`](https://doc.babylonjs.com/api/classes/babylon.observable)s instead of callbacks?

src/DeviceInput/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./deviceInputSystem";
```
syntheticmagus, line 1: nit: Newline.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+     */
+    private getGamePads(): (Gamepad | null)[] {
+        return navigator.getGamepads();
+    }
+
+    // Static
+    /** POINTER_DEVICE */
+    public static readonly POINTER_DEVICE: string = "Pointer";
+    /** KEYBOARD_DEVICE */
+    public static readonly KEYBOARD_DEVICE: string = "Keyboard";
```
bghgary, line 224: I recommend putting things that are part of the contract of this class together at the top.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * This class will take all inputs from Keyboard, Pointer, and
+ * any Gamepads and provide a polling system that all devices
+ * will use.  This class assumes that there will only be one
+ * pointer device and one keyboard.
+ */
+export class DeviceInputSystem {
+    // Private Members
+    private _inputs: Map<string, Array<number>> = new Map();
```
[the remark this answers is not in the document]
PolygonalSun: Just out of curiosity, what are our minimum browser support requirements?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+                    let axis = this._inputs.get(`${gp.id}-${gp.index}`);
+                    if (axis) {
+                        axis[i + gp.buttons.length] = gp.axes[i].valueOf();
+                    }
+                }
+            }
+        }
+
+        Engine.QueueNewFrame(() => { this.updateDevices(); });
+    }
```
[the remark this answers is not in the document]
PolygonalSun: I could remove the loop part and just call this when pollInput is called.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * getGamePads: returns all gamepads
+     * @returns array with active gamepads
+     */
+    private getGamePads(): (Gamepad | null)[] {
+        return navigator.getGamepads();
+    }
```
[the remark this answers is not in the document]
PolygonalSun: It was an old function from a previous iteration.  I will remove it and just move its code to the updateDevices function

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: I put this at the bottom to match the layout of other BJS files.  Personally, I prefer it at the top so if it's all the same, I'll move this back to the top.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+ * pointer device and one keyboard.
+ */
+export class DeviceInputSystem {
+    // Private Members
+    private _inputs: Map<string, Array<number>> = new Map();
+    private _onDeviceConnected: (deviceName: string) => void = () => { };
+    private _onDeviceDisconnected: (deviceName: string) => void = () => { };
+    private _keyboardActive: boolean = false;
+    private _pointerActive: boolean = false;
+    private _elementToAttachTo: Nullable<HTMLElement>;
```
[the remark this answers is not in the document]
PolygonalSun: It doesn't need to be.  Removed Nullable

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    // Private functions
+    /**
+     * Add device and inputs to device map
+     * @param deviceName Assigned name of device (may be SN)
+     * @param numberOfInputs Number of input entries to create for given device
+     */
+    private registerDevice(deviceName: string, numberOfInputs: number) {
+        if (!this._inputs.has(deviceName)) {
+            var device: Array<number> = [];
```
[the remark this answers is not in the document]
PolygonalSun: Changed declaration

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+    // Public functions
+    /**
+     * Checks for current device input value, given an id and input index
+     * @param deviceName Id of connected device
+     * @param inputIndex Index of device input
+     * @returns Current value of input
+     */
+    public pollInput(deviceName: string, inputIndex: number): number {
+        if (this._inputs.has(deviceName)) {
+            var device = this._inputs.get(deviceName);
```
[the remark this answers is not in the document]
PolygonalSun: Changed to const.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    // Public functions
+    /**
+     * Checks for current device input value, given an id and input index
+     * @param deviceName Id of connected device
+     * @param inputIndex Index of device input
+     * @returns Current value of input
+     */
+    public pollInput(deviceName: string, inputIndex: number): number {
+        if (this._inputs.has(deviceName)) {
```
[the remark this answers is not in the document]
PolygonalSun: Changed to just get

src/DeviceInput/index.ts
PolygonalSun: Added newline

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
belfortk: I've been assuming IE 11+, Firefox 4+, Google Chrome 9+ and Opera 15+

src/DeviceInput/deviceInputSystem.ts
belfortk: Agree with @bghgary, it's weird seeing these at the bottom.

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
deltakosh: Nope you are right

src/DeviceInput/deviceInputSystem.ts
bghgary: The majority of the code puts these statics at the top from what I can see.

src/DeviceInput/deviceInputSystem.ts
bghgary: I'm not sure. Observables are heavier. This contract is not intended for Babylon users. It is a contract for internal use. That said, there still might be a good reason to use Observables even for internal use.

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
bghgary: > Is there something Map gives you that an {} would not provide?

Maps are better for a number of reasons, as noted in the "Objects vs. Maps" [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

The issue is whether is it well-supported in browsers that we support. If you look at the "Browser compatibility" section at the end of that link, you'll see that IE11 has only partial support, which may cause weird issues that are hard to spot.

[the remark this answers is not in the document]
belfortk: Ok, I see that Maps are preferable in certain situations, but I don't see the special properties being used here? There isn't a reference to its size, I don't see any non string or symbol keys, I don't think there are going to be a ton of removals/additions of key-value pairs.

Happy to be be wrong, just trying to learn why why Map was used over {}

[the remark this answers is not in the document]
PolygonalSun: Thanks for the info!  Maps won't work so I'll get that changed over to an associative array.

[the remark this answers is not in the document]
bghgary: > Maps are preferable in certain situations

If we didn't have to worry about IE, then we probably would be using Maps for, well, maps instead of using Objects.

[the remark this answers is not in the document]
belfortk: I think my over-familiarity with JS is coming back to bite me. In my head, maps and objects were pretty much interchangeable terms for the same thing (when used like this). I did not realize there was such a distinction. TIL.

src/DeviceInput/deviceInputSystem.ts
Drigax: @PolygonalSun perhaps we want "unregisterDevice()"? I'm not sure if deregister is proper english... its a slight nit, but something to consider.

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
PolygonalSun: Changed to just use associative array calls

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+    private _pointerActive: boolean = false;
+    private _elementToAttachTo: Nullable<HTMLElement>;
+    private _maxKeyCodes : number = 222;
+    private _maxPointerButtons : number = 5;
+
+    /**
+     * Default Constructor
+     * @param elementToAttachTo - element to attach events to (usually canvas)
+     */
+    constructor(elementToAttachTo: Nullable<HTMLElement>) {
```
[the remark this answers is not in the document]
PolygonalSun: TBH, I'm not sure what would work best here but I can get the object that I need from just using the Scene as a parameter and calling .getEngine().getInputElement().  If you can think of anything that may work better, please let me know.

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
PolygonalSun: Added this back to support use of getInputElement

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
PolygonalSun: TBH, I'm not sure what would work best here but I can get the object that I need from just using the Scene as a parameter and calling .getEngine().getInputElement().  If you can think of anything that may work better, please let me know.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    /**
+     * When a device is connected, perform user specified function
+     * @param callback Callback function to use when a device is connected
+     */
+    public onDeviceConnected(callback: (deviceName: string) => void): void {
+        this._onDeviceConnected = callback;
+    }
```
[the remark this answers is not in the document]
PolygonalSun: Changed to observables

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+                pointerY[1] = evt.clientY;
+            }
+        });
+
+        this._elementToAttachTo?.addEventListener("pointerdown", (evt) => {
+            if (!this._pointerActive) {
+                this._pointerActive = true;
+                this.registerDevice(DeviceInputSystem.POINTER_DEVICE, this._maxPointerButtons);
+                this._onDeviceConnected(DeviceInputSystem.POINTER_DEVICE);
+            }
```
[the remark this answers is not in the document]
PolygonalSun: I didn't move to a separate function but found that I could move the onConnected and onDisconnected calls into the register/deregister calls as they are always paired together

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Changed to unregister because I only saw the use of that word throughout the BJS codebase.

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
bghgary: Let's chat offline.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+        this._onDeviceDisconnected = callback;
+    }
+
+    // Private functions
+    /**
+     * Add device and inputs to device map
+     * @param deviceName Assigned name of device (may be SN)
+     * @param numberOfInputs Number of input entries to create for given device
+     */
+    private registerDevice(deviceName: string, numberOfInputs: number) {
```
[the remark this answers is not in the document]
PolygonalSun: Added underscore to all private functions

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+     */
+    private deregisterDevice(deviceName: string) {
+        this._inputs.delete(deviceName);
+    }
+
+    /**
+     * Handle all actions that come from keyboard interaction
+     */
+    private handleKeyActions() {
+        window.addEventListener("keydown", (evt) => {
```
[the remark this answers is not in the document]
PolygonalSun: Created dispose function

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,268 @@
...
+
+                for (let i = 0; i < gp.axes.length; i++) {
+                    let axis = this._inputs[`${gp.id}-${gp.index}`];
+                    if (axis) {
+                        axis[i + gp.buttons.length] = gp.axes[i].valueOf();
+                    }
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
PolygonalSun: Changed to look for affected gamepad and poll for specific input.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+     * Add device and inputs to device map
+     * @param deviceName Assigned name of device (may be SN)
+     * @param numberOfInputs Number of input entries to create for given device
+     */
+    private registerDevice(deviceName: string, numberOfInputs: number) {
+        if (!this._inputs.has(deviceName)) {
+            var device: Array<number> = [];
+            for (var i = 0; i < numberOfInputs; i++) {
+                device.push(0);
+            }
```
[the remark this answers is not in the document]
PolygonalSun: I'm initializing the values to zero because the pollInput function checks for undefined as a way of making sure that it's staying within the bounds.  The updateDevice function comes after that. 
 Because of this check, any values that start as undefined will remain undefined.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+        this._onDeviceConnected = callback;
+    }
+
+    /**
+     * When a device is disconnected, perform user specified function
+     * @param callback Callback function to use when a device is disconnected
+     */
+    public onDeviceDisconnected(callback: (deviceName: string) => void): void {
+        this._onDeviceDisconnected = callback;
+    }
```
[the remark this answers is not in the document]
PolygonalSun: I've switched these callbacks to Observables.

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
PolygonalSun: I've added a dispose function to handle eventlistener removal

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -1,225 +1,268 @@
...
+        if (device && device![inputIndex] != undefined) {
+            this.updateDevices();
+            return device![inputIndex];
+        }
+        else if (device) {
+            throw `Unable to find input ${inputIndex} on device ${deviceName}`;
         }
         else {
             throw `Unable to find device ${deviceName}`;
         }
```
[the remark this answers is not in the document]
PolygonalSun: I've changed this method to just return undefined if it can't find anything but I have cleaned up the if statement for clarity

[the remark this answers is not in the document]
bghgary: Why the change now? Is there a scenario where you expect consumers to check if a particular input exists?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,225 @@
...
+ */
+export class DeviceInputSystem {
+    // Private Members
+    private _inputs: Map<string, Array<number>> = new Map();
+    private _onDeviceConnected: (deviceName: string) => void = () => { };
+    private _onDeviceDisconnected: (deviceName: string) => void = () => { };
+    private _keyboardActive: boolean = false;
+    private _pointerActive: boolean = false;
+    private _elementToAttachTo: Nullable<HTMLElement>;
+    private _maxKeyCodes : number = 222;
```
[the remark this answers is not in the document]
PolygonalSun: I made them static

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
PolygonalSun: Talked offline, reverted to throw

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,310 @@
...
+            const deviceName = `${evt.gamepad.id}-${evt.gamepad.index}`;
+            this._registerDevice(deviceName, evt.gamepad.buttons.length + evt.gamepad.axes.length);
+        });
+
+        this._gamepadDisconnectedEvent = ((evt: any) => {
+            const deviceName = `${evt.gamepad.id}-${evt.gamepad.index}`;
+            this._unregisterDevice(deviceName);
+        });
+
+        window.addEventListener("gamepadconnected", this._gamepadConnectedEvent);
```
deltakosh, line 284: I would recommend to implement IDisposable and unwire all the events there

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    /**
+     * Handle all actions that come from pointer interaction
+     */
+    private _handlePointerActions() {
+        this._pointerMoveEvent = ((evt) => {
+            if (evt.pointerType == "mouse") {
+                if (!this._mouseActive) {
+                    this._mouseActive = true;
+                    this._registerDevice(DeviceInputSystem.MOUSE_DEVICE, DeviceInputSystem._MAX_MOUSE_INPUTS);
```
deltakosh, line 176: I still struggle to understand why we are not using evt.pointerId instead of DeviceInputSystem._MAX_MOUSE_INPUTS
same for the pointer below.

Keyboard can be at -1  and gamepads from -2 to the infinite (and beyond)

deltakosh: Overall we should not treat the mouse differently than the touch (pointer events are meant to handle that). We should have:
- pointers
- keyboard
- gamepads

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
bghgary: I think this will have subtle bugs if you initialize to 0. If the underlying implementation is event based and the event hasn't fired yet, then polling for a particular input state maybe return 0 when it isn't 0.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: When I tested using pointerId for the mouse, all of the pointerIds were the same, regardless of the mouse click.  I chose to go with the MAX_MOUSE_INPUTS approach because it leveraged the MouseEvent API to be able to distinguish between the different types of clicks (via MouseEvent.button).

deltakosh: "When I tested using pointerId for the mouse, all of the pointerIds were the same, regardless of the mouse click": this is what we want as we need to identify the mouse as an unique entity

src/DeviceInput/deviceInputSystem.ts
bghgary: I had a similar comment earlier and @PolygonalSun added a Dispose method that removes the event listeners. Probably should add the `extends IDiposable` to the class declaration too.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,310 @@
...
+                const device = this._inputs[deviceName];
+
+                if (inputIndex >= gp.buttons.length) {
+                    device[inputIndex] = gp.axes[inputIndex - gp.buttons.length].valueOf();
+                }
+                else {
+                    device[inputIndex] = gp.buttons[inputIndex].value;
+                }
+            }
+        }
```
bghgary, line 308: I still don't like this loop. There are string concatenations for each game pad that may cause allocations which are going to happen every frame. Maybe you can store the gamepads in map so that you don't need to compute the device id.

PolygonalSun: I found an easy way to cache the names by just keeping an array of the device names indexed by their gamepad index

src/DeviceInput/deviceInputSystem.ts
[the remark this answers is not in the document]
PolygonalSun: Originally, I was initializing the values to zero because of how pollInput was checking things.  I found that it would always throw for new values but I've since changed the order of operations.  First, pollInput will check for the device.  If that device exists, it will attempt to update the value for the given index.  Then it'll check if that input is defined.  Finally, it'll return the value.  With this approach, I feel like the second check is kinda unnecessary but it would allow use to not have to initialize values to zero.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
```
syntheticmagus, line 310: nit: Newline.

PolygonalSun: Added newline

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Talked offline, combined mouse/touch into singular pointer handling

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: implemented IDisposable

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -70,16 +67,19 @@ export class DeviceInputSystem {
...
 
         if (!device) {
             throw `Unable to find device ${deviceName}`;
         }
+
+        this._updateDevice(deviceName, inputIndex);
+
         if (device[inputIndex] === undefined) {
             throw `Unable to find input ${inputIndex} on device ${deviceName}`;
+
```
bghgary, line 81: nit: extraneous newline

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -295,7 +251,7 @@ export class DeviceInputSystem {
         // Look for current gamepad and get updated values
         for (let i = 0; i < gamepads.length; i++) {
             const gp = gamepads[i];
```
bghgary: nit:
```suggestion
        for (const gp of gamepads) {
```

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -274,11 +225,16 @@ export class DeviceInputSystem {
         this._gamepadConnectedEvent = ((evt: any) => {
             const deviceName = `${evt.gamepad.id}-${evt.gamepad.index}`;
             this._registerDevice(deviceName, evt.gamepad.buttons.length + evt.gamepad.axes.length);
+            if (!this._gamepads) {
+                this._gamepads = new Array<string>(evt.gamepad.index + 1);
+            }
```
bghgary, line 230: nit:
```suggestion
            this._gamepads = this._gamepads || new Array<string>(evt.gamepad.index + 1);
```

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Removed

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Changed

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Changed

PR #7925, opened 2020-03-27
Check that 'document' exists before trying to unregister events

ryantrem:
The code currently assumes that if `window` exists then `document` does too, but this isn't the case with Babylon Native, so disposing the engine throws. This adds an explicit check that `document` exists before trying to unregister events.
@bghgary


deltakosh:
Thanks!

PR #7936, opened 2020-03-28
fix(assetcontainer): Fixed ImportAnimations

phenry20:
Default targetConverter now converts nodes based on targetProperty


phenry20:
Are there any changes I need to make?

deltakosh:
Nope alllll good :) thanks!

PR #7940, opened 2020-03-29
Allow to use a custom shadow depth material for shadow rendering

Popov72:
So, here's a first draft for this PR, for discussion.

image
(there's a directional and point light in this demo)

One should use the new `customShadowDepthMaterial` property of the `Material` class to set a custom shader used to generate the shadow depth map for the corresponding meshes.

I reused the `customShaderNameResolve` feature to be able to inject new defines/uniforms in the custom shadow depth material (CSDM from now on) provided by the user. That way, I was able to simplify the task for the user when creating the CSDM: simply add 4 or 5 includes in your shader to make it work.

https://playground.babylonjs.com/#6LA5LU#2

Obviously it won't work until the feature is deployed, but you will still be able to see the code.

I have used a `CustomMaterial` to create the CSDM, and apart from the specific code to discard a fragment based on the noise values, it's only a matter of 5 includes:
* `#include<shadowMapVertexDeclaration>`, `#include<shadowMapVertexNormalBias>` and `#include<shadowMapVertexMetric>` in the vertex shader
* `#include<shadowMapFragmentDeclaration>` and `#include<shadowMapFragment>` in the fragment shader

For `CustomMaterial` and `PBRCustomMaterial` we could go even further and inject automatically those includes in the right location, as we know the shaders beforehand (we would need a property like `useAsShadowDepthMaterial = true` on `CustomMaterial` / `PBRCustomMaterial` to know that those includes should be injected).

For the dancing cube on the top left, I used a `ShaderMaterial` for both the regular and shadow depth materials.

Regarding the implementation:
* I have added some `SM` and `SM_` (SM=Shadow Map) to some variables / define names to avoid collision names (for eg, `vPositionW` and `FLOAT` are already used in the standard shaders).
* I needed to inject the defines used by the shadow generator in the collection of defines of the CSDM. The problem is that the authorized properties (names of the allowed defines) on a `MaterialDefine` are the ones statically defined on this class, which are then collected in the internal `_keys` map. So I needed to update this map to add my defines, which I did in an ugly way: `(defines as any)._keys.push(name);`. Let me know if I should make `_keys` public (but hidden from the docs) or if I should handle that in another way
* Please pay attention to my uses of mesh / rendering mesh / effective mesh / ... when dealing with a sub mesh as I'm not confident about that...
* I have split the standard shadow depth shaders in blocks so they can be reused when writing a CSDM. I needed to be able to have an include inside an include (see `shadowMapFragmentDeclaration.fx` that includes `packingFunctions`), that's why I updated `gulp-processShaders.js` because it generated something like `import "./ShadersInclude/ShadersInclude/XXX"` for an include inside an include.

~~Cascaded Shadow Map are not updated yet, waiting for inputs for the current draft first.~~ Done for CSM too.


src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -885,10 +885,23 @@ export class ShadowGenerator implements IShadowGenerator {
             if (this._filter === ShadowGenerator.FILTER_PCF) {
                 engine.setColorWrite(false);
             }
+            if (this._scene.getSceneUniformBuffer().useUbo) {
+                const sceneUBO = this._scene.getSceneUniformBuffer();
```
deltakosh: Why do we need this change?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -1005,6 +1018,11 @@ export class ShadowGenerator implements IShadowGenerator {
     }
 
     protected _bindCustomEffectForRenderSubMeshForShadowMap(subMesh: SubMesh, effect: Effect): void {
+        effect.setMatrix("viewProjection", this.getTransformMatrix());
```
deltakosh: Why do we need this change?

src/Lights/Shadows/shadowGenerator.ts
Popov72: When using a `StandardMaterial` as a base material for the custom depth material as in my sample PG, the view/projection transformations I got in the shader were the ones from the scene, not the ones I bound with:
```typescript
        effect.setMatrix("viewProjection", this.getTransformMatrix());
        effect.setMatrix("view", this._viewMatrix);
        effect.setMatrix("projection", this._projectionMatrix);
```
I had to put the code above to get the right matrices. I don't have this problem when using the ShaderMaterial class. I think standard / PBR materials use UBO by default when available.

src/Lights/Shadows/shadowGenerator.ts
Popov72: I moved those binding in this function because those are overwritten by the CSM class because CSM provides its own implementation of `_bindCustomEffectForRenderSubMeshForShadowMap`. It avoids having those uniforms set by the base class then being immediately overwritten by CSM (double work for nothing).

deltakosh: Ok makes sense!

src/Lights/Shadows/shadowGenerator.ts
deltakosh: oki!

deltakosh:
Just a tiny issue with the doc

Popov72:
> Just a tiny issue with the doc

Yes, I did not document the new class yet.

Also, I need to handle custom block placement in the shader code as well as variable renaming, when the user code is not using the same variable names than the shadow blocks are expecting.

deltakosh:
All good!

deltakosh:
For the what's new, this is definitely a major update that we will promote so please add it in the main chapter :D

PR #7958, opened 2020-04-01
fix array loop from for in to for index

NasimiAsl:


NasimiAsl:
i always wait your ping notify :)

PR #7978, opened 2020-04-03
Fix bug where child NME windows from playground were not closing on page reload

### LINKED ISSUE #7972: NME - Child window for editor does not close when page reloaded
## Bug
The child window opened for the node material editor used to close when the page that invoked the child window was closed but now remains open. I ran into this because I do local development with a local http server and reload the page to update changes. In the past, reloading the page would close down the child window but now the child window is left open and subsequent invokes of the node material editor will load into that same child window. However, subsequent loads into the child window break the interaction with the node graph, i.e. you can't pan the graph or move nodes. 

Testing this with the playground shows that the child window does close when clicking on the run button to re-run a scene, but reloading the playground itself does not close the child window.

belfortk:
FIxes #7972


deltakosh:
Tests are failing

belfortk:
Tests are passing after second run 😇

PR #8011, opened 2020-04-10
Draggable Window of Curve Editor

toledoal:
Hello! This PR opens/closes a draggable window that will contain the Animation Curve Editor.

I still don't know where to place these files correctly (folder structure). Maybe the container window "draggableWindowComponent" should be in another section besides "animations" and can be reused by other components? This also applies to its CSS classes.

Also, what type of content should be put in the "what's new" file? and in which category?

 Pending
- [x] Styles
- [x]  Resizable Window

image


azure-pipelines-cd.yml
```diff
@@ -107,19 +107,19 @@ jobs:
...
       NETLIFY_AUTH_TOKEN: $(babylon.netlify.authToken)
-  - task: AzureFileCopy@2
-    displayName: 'Update Preview Storage'
-    inputs:
-      SourcePath: 'dist/preview release'
-      azureSubscription: 'Visual Studio Enterprise (2bb1afe6-4492-4d3d-9cf8-e658f447d211)'
-      Destination: AzureBlob
-      storage: previewcdn
-      ContainerName: '$web'
+#  - task: AzureFileCopy@2
```
deltakosh, line 110: This file should not have been modified

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditorComponent.tsx
```diff
@@ -0,0 +1,25 @@
...
+    close: Function;
+}
+
+export class CurveEditorComponent extends React.Component<ICurveEditorComponentProps, {isOpen: boolean}> {
+
+    constructor(props: ICurveEditorComponentProps) {
+        super(props);
+    }
+
+    render() {
```
deltakosh, line 16: Does this window can be moved to another screen? If not we need to do what I did for NME (use a regular window)

toledoal: Let me check how the NME works. 👍

azure-pipelines-cd.yml
toledoal: Yes. I don't know what happened here... I'll be more careful about this 😒

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditorComponent.tsx
toledoal: For this one, you mean a de-attachable window? Like 
image

or 

image

deltakosh: Yep

PR #8012, opened 2020-04-10
Make node ports editable

belfortk:
In this PR, I:

• Added a displayName property to NodeMaterialBlockConnectionPoint class
• Moved editing functionality from FrameNodePort to parent NodePort class
• Added NodePortPropertyComponent
• Updated onSelectionChangedObserver logic

After merging this PR, a user will be able to:

• Select node ports
• edit node port names
• If a node port doesn't have a label, the click will select the graph node
• Edited node port names will be reflected in frame ports and vice versa.

Question for @deltakosh, I added the displayName property to nodeMaterialBlockConnectionPoint.ts but it would not show up while dev-ing. When I added it to the babylon.module.d.ts, it would show up. I don't think it was correct for me to add it there though. Should I take it out of the babylon.module.d.ts in this PR? How does it eventually get into the babylon.module.d.ts?


src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
```diff
@@ -133,6 +133,11 @@ export class NodeMaterialConnectionPoint {
      */
     public name: string;
 
+    /**
+     * Gets or sets the connection point name
+     */
+    public displayName: string;
```
deltakosh, line 139: You may want to serialize / deserialize it

belfortk: Added it to NodeMaterialBlockConnectionPoint.serialize()

deltakosh: And what about deserialize???

belfortk: Working on that now. Is there already a deserialize method? I could not find one so I write a new one for NodeMaterialBlockConnectionPoint class.

belfortk: I added _deserializePortDisplayNames() to NodeMaterialBlock. I think it makes the most sense there but happy to discuss if otherwise.

src/Materials/Node/nodeMaterialBlock.ts
```diff
@@ -656,18 +656,43 @@ export class NodeMaterialBlock {
...
 
         serializationObject.inputs = [];
+        serializationObject.outputs = [];
 
         for (var input of this.inputs) {
             serializationObject.inputs.push(input.serialize());
         }
 
+        for (var output of this.outputs) {
+            serializationObject.outputs.push(output.serialize());
```
deltakosh, line 666: This one should only serialize the displayName. So maybe adding a boolean to connectionNodePoint.serialize? We do not want to serialize output connection info

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
```diff
@@ -442,6 +448,19 @@ export class NodeMaterialConnectionPoint {
         return serializationObject;
     }
 
+    // /**
```
deltakosh, line 451: this needs top be removed

belfortk: removed

src/Materials/Node/nodeMaterialBlock.ts
belfortk: updated

PR #8014, opened 2020-04-12
Fix issue when compiling with closure compiler.

ageneau:
Fixes "Invalid delete operand. Only properties can be deleted." when compiling
with the closure compiler.
See https://forum.babylonjs.com/t/possible-bug-when-compiling-babylonjs-with-the-closure-compiler/9923


deltakosh:
Thanks!

PR #8023, opened 2020-04-15
Week2 april

toledoal:
Hi! 

I closed the other PR which had a draggable window. ( 

This PR adds a popup window that will contain the Animation Curve Editor.

The Edit button opens a new window. Also the Curve Editor component can close its own window.

The popupComponent.tsx can be reused to open other windows. 

(Could you tell me what is the format of the What's new update?) I read the docs, but I don't know where the update should be.


what's new.md
```diff
@@ -172,6 +172,11 @@
 
 - Added ```unlit``` mode to lava material ([sebavan](http://www.github.com/sebavan))
 
+
```
deltakosh, line 175: This is not the good what's new file :)

This one is for the stable release (4.1)

The upcoming what's new is here:

toledoal: Ahh. Got it. Thanks

PR #8026, opened 2020-04-15
Make Frame ports optional

belfortk:
• By default, only node ports that are connected to outside graphNodes will be shown on frames
• Node ports on graphNodes in a frame can be toggled to be exposed by clicking checkbox in Node Port Property Pane
• Node's connected to an outside graphNode will have their checkbox disabled
• Added "disabled" optional property to CheckBoxLineComponent

@deltakosh I changed what NodeMaterialBlockConnectionPoint.isConnected() is checking to determine whether or not a node is connected. I hope that is ok. I needed a way to check if nodes were connected to outside nodes like Fragment Output and the original method wasnt catching them.


belfortk:
Ok, I see that modifying NodeMaterialBlockConnectionPoint.isConnected() broke the Firefox tests. Back to the drawing board.

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
```diff
@@ -175,10 +175,18 @@ export class NodeMaterialConnectionPoint {
...
     public get isConnected(): boolean {
+        return this.connectedPoint !== null;
+    }
+
+    /**
+     * Gets a boolean indicating that the current point is connected to anything (another NodeMaterialBlock, a shader, or a Input Block, etc).
+     */
+    public get isConnectedToAnything() {
         return this.connectedPoint !== null || this.isConnectedInFragmentShader || this.isConnectedInVertexShader || this.isConnectedToInputBlock || this._endpoints.length > 0;
+
```
Popov72, line 189: Are your sure for the implementation of `isConnectedToAnything`?

For eg, `isConnectedInFragmentShader` will first return `true` if  `this.target === NodeMaterialBlockTargets.Fragment` but it does not mean that the point is currently connected to something.

Maybe your function is more like a `isConnectedOrConnectableToAnything` thing? But in that case, I guess that any connection point is connectable to something, so it should always return `true`?

sebavan:
@deltakosh will be back on Monday for a more thorough review.

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
belfortk: I am actually very _not_ sure about my implementation of `isConnectedToAnything`! I was hoping to get feedback on it =) 

Basically, I need a way to tell if the graphNode is connected to another graphNode. I noticed that when I was attaching ColorMerger nodes to a FragmentOutput node, `isConnected` would still be returning false. I did not realize `isConnectedInFragmentShader `returning `true` did not always mean that the point is currently connected to something.

Do you think it would be better to just check `NodeMaterialConnectionPoint._endpoints.length`?

nodeEditor/src/diagram/frameNodePort.ts
```diff
@@ -72,7 +72,11 @@ export class FrameNodePort extends NodePort {
         if (!displayManager || displayManager.shouldDisplayPortLabels(block)) {
             let portLabel = root.ownerDocument!.createElement("div");
             portLabel.classList.add("port-label");
-            portLabel.innerHTML = connectionPoint.displayName || connectionPoint.name;        
+            let portName = connectionPoint.displayName || connectionPoint.name;
+            if (portName === "output") {
```
deltakosh, line 76: What if I want to name it "output" ?

nodeEditor/src/diagram/nodePort.ts
```diff
@@ -31,7 +32,11 @@ export class NodePort {
     }
 
     public get portName(){
-        return this.connectionPoint.displayName || this.connectionPoint.name;
+        let portName = this.connectionPoint.displayName || this.connectionPoint.name;
+        if (portName === "output") {
```
deltakosh, line 36: same. We need a better way to detect that. Why not just checking if the ownerBlock.isInputBlock ?

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
Popov72: I think you will need the help of the nme master here!

nodeEditor/src/sharedComponents/checkBoxLineComponent.tsx
```diff
@@ -69,15 +70,22 @@ export class CheckBoxLineComponent extends React.Component<ICheckBoxLineComponen
         this.setState({ isSelected: !this.state.isSelected });
     }
 
+    getDisabled() {
+        if (this.props.disabled) {
+            return this.props.disabled;
```
deltakosh, line 75:
```suggestion
            return !!this.props.disabled;
```
But overall why this function?

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
deltakosh: Not sure to understand that function.

Why not simply using this.isConnected?

nodeEditor/src/diagram/frameNodePort.ts
belfortk: I did not think of that use case. That was silly of me. Updated.

nodeEditor/src/diagram/nodePort.ts
belfortk: Updated

nodeEditor/src/sharedComponents/checkBoxLineComponent.tsx
belfortk: I added an optional property `disabled` to `CheckBoxLineComponent` so that I know when the checkbox should not be responsive to user input.

Specifically, for node ports that are connected to outside nodes, I can still show the checkbox and also communicate that connected nodes are automatically exposed on the frame.

Connected node port (auto-disabled checkbox)
Screen Shot 2020-04-22 at 12 19 56 AM

Not connected node port (checkbox enabled)
Screen Shot 2020-04-22 at 12 20 21 AM

I wrote the function that way because I wanted to make sure that past imports of `CheckBoxLineComponent` wouldn't be affected by the addition of the `disabled` property. But I see now if I do the `!!`, it should accomplish the same thing. Updated.

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
belfortk: I need a way to check if the node is connected to anything in either direction. `this.isConnected` was returning false when the node is connected to a FragmentOutput node. Modifying `this.isConnected` broke the firefox tests so I just wrote another check function.

I am not sure it is the correct implementation.

deltakosh: Well then we need to understand why connecting a node to a FragmentOutput does not flag isConnected ;) This seems to be a bug

deltakosh:
Thanks!

PR #8027, opened 2020-04-15
NME: Support for the PBR material (first round)

Popov72:
Some comments about the code:
* the PBR satellite blocks (reflectivity, reflection, anisotropy, etc) don't implement anything in their `_buildBlock` method, it's the main `PBRMetallicRoughnessBlock` that gets the code it needs and put it in order
* the PBR satellite blocks generate their shader code in the `getCode` method (fragment code) and/or in `handleVertexSide` (vertex code)
* sometimes, the `getCode` method is static because `PBRMetallicRoughnessBlock` needs to get the code even if the satellite block is not connected to it (case for the Ambient Occlusion block)
* the `ReflectionBlock` has much in common with `ReflectionTextureBlock`, so I created a `ReflectionTextureBaseBlock` to factorize the code
* most of the additional properties of the `PBRMaterial` have been added as properties on blocks and are editable in the editor by clicking on the node (example for the PBRMetallicRoughness block):
image
* I have managed to reuse the `pbrDebug` code, as it really comes in handy to diagnose problems during dev:
image
* `PBRMetallicRoughnessBlock`: I could not have two times the same names in input/output like "sheen" and "clearcoat", so I renamed the outputs "sheenDir" and "clearcoatDir" ("Dir" as Direct light)

**Not implemented yet** (but will be):
* ~~direct lighting~~ direct lighting is now in
* ~~sheen block~~ Sheen is now in
* clear coat block
* sub surface block

**Not implemented**:
* `lightmap` / `fog`: it can be added by the user by using existing blocks in the NME
* `vLightingIntensity`: `vec3(1.)` for the time being. It's easy to add support for it, the problem being that it would mean 3 new inputs, one for "direct intensity", one for "environment intensity" and one for "specular intensity", and the PBR block has already quite a number of inputs... Let me know what I should do
* `Reflection`: handles only LOD based cube sampling. Handling systems without lod texture sampler support is cumbersome as it means being able to support 2 more additional textures, for the reflection, clear coat, sheen and sub surface modules. Moreover, those textures must have the same settings (as the reflection mode for eg) than the main texture, so they are somehow linked and can't be edited separately.
* `Irradiance texture`: not supported. It would mean adding yet another texture to the workflow, and most of the time users won't use an irradiance texture, spherical harmonics are just fine quality wise but weight a lot less than a texture
* `metallicF0`: set to `0.04` as a hard value. I don't think we should add a new input for this parameter, as it has only a marginal impact (if any) on the final output
* the micro surface texture is not handled. There's no need to handle this texture as we already have the metallic/roughness texture


sebavan:
Holly Molly :-) what a PR !!!! I ll have a first look today and @deltakosh will have a more thorough one on Monday when he ll be back as my knowledge of nme is not at its best. At least I learned from the PR.

sebavan:
It looks all good to me but my poor knowledge of NME is of no help on this one...

We definitely need to wait on the NME daddy approval ;-)

Popov72:
I'm ok with that :)

Popov72:
Direct lighting are now in.

Popov72:
Note:

In the old code of `ReflectionTextureBlock`, the world normal was (possibly) not normalized before being used in the calculation. The code to get the world normal was:
```javascript
let worldNormal = this.worldNormal.associatedVariableName + ".xyz";
```
and was used "as is" as input for the computations.

So, if the connection point giving the world normal did not provide a normalized normal, the normal was not normalized. I made the changes to normalize the normal before being used in the computation.

I don't know if it was expected that the user would normalize the normal before passing it to the block, but I don't think users are aware of this and most (if not all) of them simply pass the output of the `World normal` block that is created when one drag&drop the `WorldNormal` block to the graph area.

Let me know if I should revert back this change.

I also needed normals being normalized in the PBR blocks, so I normalize the vector myself when appropriate.

The drawback of doing this is that the same vector can be normalized multiple times, each time it is used by a different block that needs it to be normalized... We could maybe improve things by having a "Normalize output" checkbox in the transform block? That way, the blocks needing a normalized vector could check if their input is connected 1/ to a transform block 2/ has this box checked, and don't do normalization if yes.

deltakosh:
Please for universe sake, do small PR next time :)

Popov72:
That was 2801 net new lines, with a lot of comments ;)

PR #8031, opened 2020-04-16
DeviceSourceManager

PolygonalSun:
This PR has the initial code for the DeviceSourceManager, a manager that keeps track of specific devices and provides the user with easier to understand context when getting input.

deviceEnum.ts:
- Added enums for device types and inputs for known gamepads (xbox, ps4, and switch)

deviceInputSystem.ts:
- Increased _MAX_KEYCODES to 255 to account for international keyboard layouts
- Added MOUSE_DEVICE and mouse specific ternary operator to account for chance that a mouse's pointerId is not 1 (spec no longer guarantees this).
- Fixed an issue where touch pointers were no longer being registered after initial one
- Removed a throw from pollInput to allow for new change to input data access
- Added observeInput callback to enable use of event-driven updates for Keyboard, Mouse, and Touch

deviceSourceManager.ts:
- Users can now get input from the input system via the getInput function.  This function is not pollInput as I felt that getInput was less simpler terminology.
- There are 4 different observables (connect-before/after, disconnect-before/after) to allow for better customization of the connection process.  While they have no major use in the DSM, they will provide higher levels with some event control.
- All gamepad detection methods are based off of the device names given by the Gamepad API.  Considerations will need to be made for other platforms.
- With the getInput command, you can call it with the device type and input enum and get the first connected input value or you can specify a slot (only applicable to gamepads) and get a specific devices value.

To use this code, the minimum code that you need is to create an instance of the DeviceSourceManager and have a getInput call somewhere in your game/render loop.

Example:
```
var dsm = new BABYLON.DeviceSourceManager(scene.getEngine());
...
scene.onBeforeRenderObservable.add(() => {
     if (dsm.getInput(BABYLON.DeviceType.Xbox, BABYLON.XboxInputs.A) == 1) {
          // Do something while A is pressed
     }
});
```


src/DeviceInput/deviceInputSystem.ts
```diff
@@ -15,6 +15,8 @@ export class DeviceInputSystem implements IDisposable {
     public static readonly POINTER_DEVICE: string = "Pointer";
     /** KEYBOARD_DEVICE */
     public static readonly KEYBOARD_DEVICE: string = "Keyboard";
+    /** MOUSE_DEVICE */
+    public static readonly MOUSE_DEVICE: string = "Mouse";
```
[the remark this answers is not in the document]
PolygonalSun: I can do that.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -15,6 +15,8 @@ export class DeviceInputSystem implements IDisposable {
     public static readonly POINTER_DEVICE: string = "Pointer";
     /** KEYBOARD_DEVICE */
     public static readonly KEYBOARD_DEVICE: string = "Keyboard";
+    /** MOUSE_DEVICE */
+    public static readonly MOUSE_DEVICE: string = "Mouse";
 
     /**
      * Observable to be triggered when a device is connected
```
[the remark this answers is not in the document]
PolygonalSun: I was talking with @bghgary about this yesterday and the plan is to revert these back to callbacks.

PolygonalSun:
> My understanding is that this API is entirely polling based, specifically to get current input state. It seems like for many situations this may be hard to use. For example:
> 
> 1. When a key is pressed, I want to add an object to a scene - do I need to manually track key state and look for a change across two frames and then react to that?
> 2. As a pointer is moved across the screen, I want to rotate an object - do I need to track pointer position deltas across frames and apply that delta to a rotation?
> 
> I would have though we'd want to include a more event driven model, or at the very least for polling have a way of querying deltas across frames without the consumer having to manage this themselves. Am I missing anything, or are these scenarios not well covered with the current proposal?
> 
> Edit: It's possible I'm not understanding how this is supposed to work. In your example above, you check whether the Xbox input A's state is `1` and your comment says "Do something when A is **hit**". Would that example be doing something only when A transitions from a non-pressed to pressed state, or would it do something every frame as long as A is in a pressed state? I guess if it is the latter, then how would you expect to do the former (perform an action when A transitions from non-pressed to pressed)?

This is a polling based system.  For #1, I'm not 100% sure that I understand the scenario.  With the code as is, you should be able to just listen for input using onBeforeRenderObservable and add your object there.  This may not be what you're asking so I'd like to confirm things with you offline. For #2, the current system just keeps track of absolute position and click status.  For this scenario, you'd need to calculate and track deltas manually.  If you want, I could add additional support for things like movement deltas without major issue.

src/DeviceInput/InputDevices/deviceEnums.ts
```diff
@@ -0,0 +1,189 @@
...
+    GenericDevice = 6
+}
+
+// Device Enums
+/**
+ * Enum for All Pointers (Touch/Mouse)
+ */
+export enum PointerInputs
+{
+    /** Horizontal Axis */
```
Drigax, line 27: What drives this specific pointer input enum?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    // Public Functions
+    /**
+     * Checks for current device input value, given DeviceType, slot, and inputIndex
+     * @param type Enum specifiying device type
+     * @param deviceSlot "Slot" or index that device is referenced in
+     * @param inputIndex Index of device input
+     * @returns Current value of input
+     */
+    public getInput(type: DeviceType, inputIndex: number, deviceSlot: number = this._firstDevice[type]): Nullable<number> {
```
ryantrem, line 69: Correct usage of this function seems not very discoverable or enforceable (e.g. how do you know you are supposed to pass one of the input enum values in for `inputIndex`? You can pass any old number in for `inputIndex` and the code will compile without errors). I think this could be improved with conditional types:
```typescript
type DeviceInputs<T extends DeviceType> =
    T extends DeviceType.Keyboard | DeviceType.GenericDevice ? number :
    T extends DeviceType.Mouse | DeviceType.Touch ? PointerInputs :
    T extends DeviceType.DualShock ? DualShockInputs :
    T extends DeviceType.Xbox ? XboxInputs :
    T extends DeviceType.Switch ? SwitchInputs :
    never;

public getInput<T extends DeviceType>(deviceType: T, inputIndex: DeviceInputs<T>, deviceSlot: number = this._firstDevice[type]): Nullable<number> {
    ...
}
```

This way, when the first parameter (`deviceType`) is known at compile time, then the type of `inputIndex` is both more discoverable and also enforced at compile time. Here is what you see in intellisense when you start typing the function:
image
Since `deviceType` has not been specified, `inputIndex` is unconstrained, and intellisense shows exactly the same thing as what it does with your existing implementation. Here is what you see when you've specified `DeviceType.Mouse` for the `deviceType` parameter:
image
Notice that the `inputIndex` parameter has now been constrained to be of type `PointerInputs`, which is much more discoverable to someone trying to use this API. Further, once `deviceType` has been specified as `DeviceType.Mouse`, then providing anything *other* than `PointerInputs` for the `inputIndex` parameter results in a compile error:
image

ryantrem:
> This is a polling based system. For #1, I'm not 100% sure that I understand the scenario. With the code as is, you should be able to just listen for input using onBeforeRenderObservable and add your object there. This may not be what you're asking so I'd like to confirm things with you offline. For #2, the current system just keeps track of absolute position and click status. For this scenario, you'd need to calculate and track deltas manually. If you want, I could add additional support for things like movement deltas without major issue.
> 
> For the XboxInputs.A example, the comment may be a bit misleading. It should perform whatever action is in the if statement block, each frame, as long as the A button remains pressed. For the alternative, the user would have to add additional code to their query to handle single button presses (eg. add a bool to check that a button was pressed and the execute an action if the button set the bool to true and is currently 0).

For my first example above, I just meant that when the `A` keyboard key (for example) goes from up state to down state, I want to add an object to the scene. I think ultimately all my examples are examples of wanting frame-to-frame deltas. For things like an axis (e.g. pointer position), the delta is really obvious (it moved x pixels). I think we can apply the same thing to buttons though (the delta would be 0 if the button has the same state, either down or up, between frames; the delta would be 1 if it went from not pressed to pressed; the delta would be -1 if it went from pressed to not pressed). I think to make the programming model easy, we should do one of two things (or both):
1. In addition to being able to query the current input state, also be able to query the input state delta since the last frame (for any device type + input index), *and* be able to get a list of input indexes (per device type) that have changed since the last frame (this would make it easy to see that on a device with many input indices, like a keyboard, that specifically the `A` key has changed since the last frame).
2. Expose events for when an input index value changes.

cc @bghgary

src/DeviceInput/InputDevices/deviceEnums.ts
```diff
@@ -0,0 +1,189 @@
...
+    /** Generic Device */
+    GenericDevice = 6
+}
+
+// Device Enums
+/**
+ * Enum for All Pointers (Touch/Mouse)
+ */
+export enum PointerInputs
+{
```
ryantrem, line 26: Nit: Move opening curly brace to same line as enum declaration.

PolygonalSun:
@ryantrem @bghgary I believe all of the requested changes should be included now.

PolygonalSun:
Here's a sample script that should demonstrate some of the new features:
```

    // Set to false to disable events (false is the default)
    var test = new BABYLON.DeviceSourceManager(scene.getEngine(), true);

    test.onBeforeDeviceConnectedObservable.add((device) => {
        const type = device["deviceType"];
        const slot = device["deviceSlot"];
        switch(type)
        {
            case BABYLON.DeviceType.Keyboard:
                console.log("Keyboard connected.");
                break;
            case BABYLON.DeviceType.Mouse:
                console.log("Mouse connected.");
                break;
            case BABYLON.DeviceType.Touch:
                console.log("Touch detected with id of " + slot);
                break;
            case BABYLON.DeviceType.Xbox:
                console.log("Xbox controller connected in Slot " + slot);
                break;
            case BABYLON.DeviceType.DualShock:
                console.log("Dualshock controller connected in Slot " + slot);
                break;
            case BABYLON.DeviceType.Switch:
                console.log("Switch controller connected in Slot " + slot);
                break;
            case BABYLON.DeviceType.Generic:
            default:
                console.log("Unknown device connected in Slot " + slot);
                break;
        }
    });

    // Test for getDeviceSource
    test.onAfterDeviceConnectedObservable.add((device) => {
        if (device["deviceType"] == BABYLON.DeviceType.Keyboard) {
            test.getDeviceSource(BABYLON.DeviceType.Keyboard).onInputChangedObservable.add((source) => {
                console.log("Input KeyCode: " + source["inputIndex"]);
            });
        }
    });

    scene.onBeforeRenderObservable.add(() => {
        if (test.getInput(BABYLON.DeviceType.Xbox, BABYLON.XboxInput.A, 0) == 1)
        {
            console.log("Player 1: A");
        }
        else if (test.getInput(BABYLON.DeviceType.Xbox, BABYLON.XboxInput.A) == 1)
        {
            console.log("First available: A");
        }
    });
    return scene;
};
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I've implemented the DeviceInputs type to enforce these conditional types

src/DeviceInput/InputDevices/deviceEnums.ts
PolygonalSun: Fixed

src/DeviceInput/InputDevices/deviceEnums.ts
PolygonalSun: This enum is used with any Touch or Mouse input.  Touch inputs use the first three values while Mouse inputs use all of them.

RaananW:
Would it be possible to add an observable when input actually changed and have the user listen to this, instead of running it on the beforeRender loop?
If I understand @ryantrem correctly, this is what he expected the API to look like (in the first comment) - more event driven. 
I know it is much simplified, but the WebXR motion controller API is also polling based. I added a few ways for the user to monitor changes, which you might be able to use here. And again, I understand this is a much broader API. 

And just for me to understand - I noticed you removed observables and switched to callback-based API. Is there a specific reason for that?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -71,24 +92,22 @@ export class DeviceInputSystem implements IDisposable {
         const device = this._inputs[deviceName];
 
         if (!device) {
-            throw `Unable to find device ${deviceName}`;
+            return null;
```
bghgary, line 95: It's not clear to me why this needs to change.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -169,43 +194,56 @@ export class DeviceInputSystem implements IDisposable {
...
 
             const pointer = this._inputs[deviceName];
             if (pointer) {
+                if (this.observeInput) {
+                    this.observeInput(deviceName, evt.button + 2, pointer[evt.button + 2], 0);
+                }
                 pointer[evt.button + 2] = 0;
             }
-            if (evt.pointerId != 1) // Don't unregister the mouse
+            if (evt.pointerType != "mouse") // Don't unregister the mouse
```
bghgary, line 246: nit: Comments that say exactly what the code does is meanless IMO. Maybe explain why you don't unregister the mouse.

src/DeviceInput/InputDevices/deviceEnums.ts
```diff
@@ -0,0 +1,199 @@
...
+    /** RStickXAxis */
+    RStickXAxis = 20,
+    /** RStickYAxis */
+    RStickYAxis = 21
+}
+
+/**
+ * Type to handle enforcement of inputs
+ */
+export type DeviceInputs<T extends DeviceType> =
```
bghgary, line 193: nit: this isn't an enum, but it is in a file called deviceEnums

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    // Public Functions
+    /**
+     * Checks for current device input value, given DeviceType, slot, and inputIndex
+     * @param type Enum specifiying device type
+     * @param inputIndex Index of device input
+     * @param deviceSlot "Slot" or index that device is referenced in
+     * @returns Current value of input
+     */
+    public getInput<T extends DeviceType>(type: T, inputIndex: DeviceInputs<T>, deviceSlot: number = this._firstDevice[type]): Nullable<number> {
```
bghgary, line 153: Is this function still useful? It is duplicate functionality with getDeviceSource and then calling getInput?

src/DeviceInput/InputDevices/deviceEnums.ts
```diff
@@ -0,0 +1,199 @@
...
+    /** Touch Pointers */
+    Touch = 2,
+    /** PS4 Dual Shock */
+    DualShock = 3,
+    /** Xbox */
+    Xbox = 4,
+    /** Switch Controller */
+    Switch = 5,
+    /** Generic */
+    Generic = 6
```
ryantrem, line 18: If we are going to keep `Generic`, can we set its value to `Number.MAX_VALUE` or something so that when we add additional device types in the future, `Generic` can remain at the very end of the enum?

src/DeviceInput/InputDevices/deviceEnums.ts
ryantrem, line 193: If all the enums are not plural, and this type resolves to one of those enums, then it seems like it should also not be plural. That is: `DeviceInput<T extends DeviceType>`

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+import { Observable } from '../../Misc/observable';
+
+/**
+ * Class that handles all input for a specific device
+ */
+export class DeviceSource<T extends DeviceType> {
+    /**
+     * Observable to handle device input changes per device
+     */
+    public onInputChangedObservable = new Observable<{ inputIndex: number, previousState: Nullable<number>, currentState: Nullable<number> }>();
```
ryantrem, line 15:
```suggestion
    public onInputChangedObservable = new Observable<{ inputIndex: DeviceInput<T>, previousState: Nullable<number>, currentState: Nullable<number> }>();
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+    private _deviceInputSystem: DeviceInputSystem;
```
ryantrem, line 16: `readonly`?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+     * @param deviceInputSystem Reference to DeviceInputSystem
+     * @param deviceName Name of device to be used by DeviceInputSystem
+     * @param deviceType Type of device
+     * @param deviceSlot "Slot" or index that device is referenced in
+     */
+    constructor(deviceInputSystem: DeviceInputSystem,
+        /** Name of device to be used by DeviceInputSystem */
+        public deviceName: string,
+        /** Type of device */
+        public deviceType: DeviceInputs<T>,
```
ryantrem, line 30: The type of `deviceType` is `DeviceType`, not `DeviceInputs<T>`. `DeviceInputs<T>` resolves to one of the input enums that corresponds to the `DeviceType`.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+            this._touchPoints = new Array<string>();
+        }
+    }
+
+    /**
+     * Get input for specific input
+     * @param inputIndex index of specific input on device
+     * @returns Input value from DeviceInputSystem
+     */
+    public getInput(inputIndex: T): Nullable<number> {
```
ryantrem, line 45: `inputIndex` should be of type `DeviceInputs<T>` since it is the enum corresponding to the `DeviceType`.
```suggestion
    public getInput(inputIndex: DeviceInputs<T>): Nullable<number> {
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+    /**
+     * Remove specific point from array of touch points (DeviceType.Touch only)
+     * @param name Name of Specific Touch Point
+     */
+    public removeTouchPoints(name: string) {
+        if (this.deviceType == DeviceType.Touch) {
+            const touchIndex = this._touchPoints.indexOf(name);
+            this._touchPoints.splice(touchIndex, 1);
+        }
+    }
```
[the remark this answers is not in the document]
ryantrem: I think these are only publicly exposed so that `DeviceSourceManager` can set them. This is a bit different from what I had imagined. I thought all the state would still be managed by `DeviceSourceManager`, and `DeviceSource` would just provide a typesafe way to interact with a specific device managed by `DeviceSourceManager`. I think doing that would make a bit more sense, and also probably making `DeviceSource` a nested class of `DeviceSourceManager` so that `DeviceSource` can call private methods of `DeviceSourceManager`. This way the public contract will be very clear and not provide access to private state.

src/DeviceInput/InputDevices/deviceSourceManager.ts
ryantrem: Yea I find this rather confusing. I would vote to remove it to simplify the API.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+        return this._devices[type][deviceSlot].getInput(inputIndex);
+    }
+
+    /**
+     * Gets a DeviceSource, given a type and slot
+     * @param deviceType Enum specifiying device type
+     * @param deviceSlot "Slot" or index that device is referenced in
+     * @returns DeviceSource object
+     */
+    public getDeviceSource(deviceType: DeviceType, deviceSlot: number = this._firstDevice[deviceType]): Nullable<DeviceSource<DeviceType>> {
```
ryantrem, line 167: This will not actually get us the type safety we are aiming for. We need a generic type parameter for the conditional type to work:
```suggestion
    public getDeviceSource<T extends DeviceType>(deviceType: T, deviceSlot: number = this._firstDevice[deviceType]): Nullable<DeviceSource<T>> {
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+
+        return this._devices[deviceType][deviceSlot];
+    }
+
+    /**
+     * Gets an array of DeviceSource objects for a given device type
+     * @param deviceType Enum specifiying device type
+     * @returns Array of DeviceSource objects
+     */
+    public getDeviceSources(deviceType: DeviceType): ReadonlyArray<DeviceSource<DeviceType>> {
```
ryantrem, line 180:
```suggestion
    public getDeviceSources<T extends DeviceType>(deviceType: T): ReadonlyArray<DeviceSource<T>> {
```

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -1,30 +1,51 @@
-import { Observable } from "../Misc/observable";
 import { Engine } from '../Engines/engine';
 import { IDisposable } from '../scene';
 import { Nullable } from '../types';
 
+/**
+ * Pointer device string
+ * @hidden
+ */
+const POINTER_DEVICE: string = "Pointer";
```
ryantrem, line 9: Any reason you define these on one line, and export them on a different line? I can't see any reason not to just do:
```suggestion
export const POINTER_DEVICE: string = "Pointer";
```
And remove the export statements on lines 21-26.

src/DeviceInput/index.ts
```diff
@@ -1 +1,3 @@
 export * from "./deviceInputSystem";
```
ryantrem, line 1: Do we need have some imports here before the exports to ensure these new types don't get stripped out? cc @bghgary

bghgary: Probably. If types don't get used, they won't make it into the js without an additional import.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+            this._removeDevice(deviceName);
+            this.onAfterDeviceDisconnectedObservable.notifyObservers({ deviceType, deviceSlot });
+        };
+
+        if (!this._deviceInputSystem.observeInput && enableObserveEvents) {
+            this._deviceInputSystem.observeInput = (deviceName, inputIndex, previousState, currentState) => {
+                const deviceType = this._getDeviceTypeFromName(deviceName);
+                const deviceSlot = this._getDeviceSlot(deviceName);
+
+                if (deviceType == DeviceType.Keyboard || deviceType == DeviceType.Mouse || deviceType == DeviceType.Touch) {
```
ryantrem, line 138: Why is this restricted to a specific set of device types? Should't any of these have the option of raising events?

src/DeviceInput/InputDevices/deviceEnums.ts
PolygonalSun: I can pull this out into its own file called deviceTypes.

PolygonalSun: I'll change this to DeviceInput

src/DeviceInput/InputDevices/deviceSourceManager.ts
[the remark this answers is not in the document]
PolygonalSun: Added

PolygonalSun: Changed

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Changed

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+     * @param deviceType Type of device
+     * @param deviceSlot "Slot" or index that device is referenced in
+     */
+    constructor(deviceInputSystem: DeviceInputSystem,
+        /** Name of device to be used by DeviceInputSystem */
+        public deviceName: string,
+        /** Type of device */
+        public deviceType: DeviceInputs<T>,
+        /** "Slot" or index that device is referenced in */
+        public deviceSlot: number = 0) {
```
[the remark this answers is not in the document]
PolygonalSun: Added readonly

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Changed

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I originally wanted to leave this to keep with the initial idea of simply asking the DSM for an input but using the getDeviceSource function is functionally the same thing.  Removed.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Changed

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+                let first = -1;
+
+                while (first < 0 && i < this._devices[type].length) {
+                    if (this._devices[type][i]) {
+                        first = i;
+                    }
+                    i++;
+                }
+
+                this._firstDevice[type] = first;
```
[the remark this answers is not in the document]
PolygonalSun: Changed.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+     * @returns DeviceType enum value
+     */
+    private _getDeviceTypeFromName(deviceName: string): DeviceType {
+        if (deviceName == KEYBOARD_DEVICE) {
+            return DeviceType.Keyboard;
+        }
+        else if (deviceName == MOUSE_DEVICE) {
+            return DeviceType.Mouse;
+        }
+        else if (deviceName.search(POINTER_DEVICE) !== -1) {
```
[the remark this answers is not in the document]
PolygonalSun: Changed.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: I originally did this because of a validation issue in the build process but never reverted them after just using ```@hidden```.  Changed.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -1,30 +1,51 @@
...
      */
-    public onDeviceConnectedObservable = new Observable<string>();
+    public onDeviceDisconnected: (deviceName: string) => void;
 
     /**
-     * Observable to be triggered when a device is disconnected
+     * Callback to be triggered when event driven input is updated
      */
-    public onDeviceDisconnectedObservable = new Observable<string>();
+    public observeInput: (deviceName: string, inputIndex: number, previousState: Nullable<number>, currentState: Nullable<number>) => void;
```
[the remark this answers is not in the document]
PolygonalSun: Changed.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Changed comment to explain why we don't want to remove the mouse.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: So I was testing the removal of this and it breaks my original ideal of single line poll-and-forget.  Before, you could just keep asking and either get nothing or get input.  Now, it will fail if the device you're listening to hasn't been added yet.  This can be worked around by just checking for the device before polling.  I do want to revisit this but for now I can leave it out.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Added.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+
+    /**
+     * Get slot for a given input
+     * @param deviceName Name of Device
+     * @returns Slot Number
+     */
+    private _getDeviceSlot(deviceName: string): number {
+        const splitName = deviceName.split("-");
```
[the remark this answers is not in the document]
PolygonalSun: I've converted everything to just use type and slot so string concatenation shouldn't be an issue anymore.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+                break;
+        }
+    }
+
+    /**
+     * Gets DeviceType from the device name
+     * @param deviceName Name of Device from DeviceInputSystem
+     * @returns DeviceType enum value
+     */
+    private _getDeviceTypeFromName(deviceName: string): DeviceType {
```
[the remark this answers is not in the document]
PolygonalSun: Moved to DeviceInputSystem and renamed to _getGamepadDeviceType.  This is because I removed the unnecessary if statements for Keyboard and Pointers.  The only thing that's left is gamepad stuff.  Additionally, the _getDeviceSlot function no longer exists.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+    // Public Functions
+    /**
+     * Checks for current device input value, given DeviceType, slot, and inputIndex
+     * @param type Enum specifiying device type
+     * @param inputIndex Index of device input
+     * @param deviceSlot "Slot" or index that device is referenced in
+     * @returns Current value of input
+     */
+    public getInput<T extends DeviceType>(type: T, inputIndex: DeviceInputs<T>, deviceSlot: number = this._firstDevice[type]): Nullable<number> {
+        if (!this._devices[type] || this._firstDevice[type] === undefined || this._devices[type][deviceSlot] === undefined) {
```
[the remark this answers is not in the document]
PolygonalSun: There are times when this will be undefined, such as when it's checking for a gamepad that isn't connected.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    /**
+     * Gets a DeviceSource, given a type and slot
+     * @param deviceType Enum specifiying device type
+     * @param deviceSlot "Slot" or index that device is referenced in
+     * @returns DeviceSource object
+     */
+    public getDeviceSource(deviceType: DeviceType, deviceSlot: number = this._firstDevice[deviceType]): Nullable<DeviceSource<DeviceType>> {
+        if (!this._devices[deviceType] || this._firstDevice[deviceType] === undefined || this._devices[deviceType][deviceSlot] === undefined) {
```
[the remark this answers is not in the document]
PolygonalSun: See previous comment related to this.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Originally, this was changed to account for checking for input before a device exists but there are checks outside of this so I put the throw back as it just never throw under *normal* circumstances.

src/DeviceInput/InputDevices/deviceSourceManager.ts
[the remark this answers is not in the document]
PolygonalSun: My initial idea was to just keep track of how many touch inputs there were and keep them in a manageable set of slots so that the user doesn't need to keep up with a bunch of changing pointer ids.  For the time being, I've ripped out all of that code and will be testing things without it.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: These strings have actually been removed.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I had originally added these checks because we don't have event based support for gamepads.  In the code, we don't call the onInputChanged function for gamepads so removing this if statement won't affect gamepads.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Added

src/DeviceInput/index.ts
PolygonalSun: Added import

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -35,21 +13,21 @@ export class DeviceInputSystem implements IDisposable {
     /**
      * Callback to be triggered when a device is connected
      */
-    public onDeviceConnected: (deviceName: string) => void;
+    public onDeviceConnected: (deviceType: DeviceType, deviceSlot: number) => void = () => {};
```
ryantrem, line 16: I guess this means the native implementation will need to either duplicate the enum, or just use numbers directly, or consume the actual enum via napi. This seems like it complicates things a bit more for alternative implementations of `DeviceInputSystem`. Any thoughts on this @bghgary?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    // Private Members
+    private _devices: Array<Array<DeviceSource<DeviceType>>>;
+    private _firstDevice: Array<number>;
+    private _deviceInputSystem: DeviceInputSystem;
+
+    /**
+     * Default Constructor
+     * @param engine engine to pull input element from
+     * @param enableObserveEvents boolean to enable use of observe events
```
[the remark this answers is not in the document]
PolygonalSun: I didn't notice any performance issues when I was running this but I've been testing on a few reasonably powerful devices so I wasn't sure if my results were all that reliable for the average user.  In any case, I've removed the check and I'll revisit it later if need be.

src/DeviceInput/InputDevices/deviceEnums.ts
PolygonalSun: With the way that we've implemented keeping track of the devices, I feel like setting this to max wouldn't be the best course of action.  We use the number of enums in DeviceType to create and allocate an array and then copy the data into their corresponding areas using the associated enum's assigned number.

bghgary: Maybe make Generic = 0 and shift everything else?

src/DeviceInput/deviceInputSystem.ts
bghgary: Meh, I doubt this enum is going to change much, so it's probably fine to duplicate it.

src/DeviceInput/InputDevices/deviceSourceManager.ts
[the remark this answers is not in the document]
bghgary: We should at least capture a few profiles with these changes to make sure nothing is eating up too much resources.

src/DeviceInput/InputDevices/deviceSourceManager.ts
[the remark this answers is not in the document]
bghgary: Does that means deviceSlot can be `undefined`?

src/DeviceInput/InputDevices/deviceEnums.ts
PolygonalSun: I'd be fine with that.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: If we're okay with the current implementation.  I'll just mark this as resolved.

PolygonalSun:
> Would it be possible to add an observable when input actually changed and have the user listen to this, instead of running it on the beforeRender loop?
> If I understand @ryantrem correctly, this is what he expected the API to look like (in the first comment) - more event driven.
> I know it is much simplified, but the WebXR motion controller API is also polling based. I added a few ways for the user to monitor changes, which you might be able to use here. And again, I understand this is a much broader API.
> 
> And just for me to understand - I noticed you removed observables and switched to callback-based API. Is there a specific reason for that?

To answer your questions, I've added a callback and observables to allow the user to receive and handle events that will fire off when input sources are updated.  The only caveat is that these will not be able to be used with Gamepads because they are not event based.  I may revisit this later on.  For the change back to callback from Observables, this was done to make the implementation of the Native version of the DeviceInputSystem easier.

src/DeviceInput/InputDevices/deviceSourceManager.ts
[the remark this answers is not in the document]
PolygonalSun: Yes.  If all devices for a given DeviceType are removed, deviceSlot could be undefined.

src/DeviceInput/InputDevices/deviceSourceManager.ts
bghgary: I'm not sure I understand. You can use the other function to do it all in one line no?

PolygonalSun: You can, if a DeviceSource exists for it.  Otherwise, you're trying to get input from a null object.

bghgary: With newer versions of TypeScript, it is easy to do this in one-line:
```javascript
deviceSourceManager.getDeviceSource(...)?.getInput(...)
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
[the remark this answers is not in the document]
bghgary: Okay, so deviceSlot currently has a type of number but it can also be undefined. Maybe you should default deviceSlot to undefined and set the deviceSlot manually after checking the _firstDevice.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: That works for TypeScript but it looks like it isn't available yet for JavaScript (proposal for it is finished though so it could be coming soon.).

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -0,0 +1,186 @@
...
+     * @param deviceSlot "Slot" or index that device is referenced in
+     * @returns DeviceSource object
+     */
+    public getDeviceSource<T extends DeviceType>(deviceType: T, deviceSlot: number = this._firstDevice[deviceType]): Nullable<DeviceSource<T>> {
+        if (!this._devices[deviceType] || this._firstDevice[deviceType] === undefined || this._devices[deviceType][deviceSlot] === undefined) {
+            return null;
+        }
+
+        return this._devices[deviceType][deviceSlot];
+    }
```
bghgary, line 117:
```suggestion
    public getDeviceSource<T extends DeviceType>(deviceType: T, deviceSlot?: number): Nullable<DeviceSource<T>> {
        if (deviceSlot === undefined) {
            if (this._firstDevice[deviceType] === undefined) {
                return null;
            }

            deviceSlot = this._firstDevice[deviceType];
        }

        if (!this._devices[deviceType] || this._devices[deviceType][deviceSlot] === undefined) {
            return null;
        }

        return this._devices[deviceType][deviceSlot];
    }

```

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -67,28 +66,34 @@ export class DeviceInputSystem implements IDisposable {
...
         return device[inputIndex];
     }
 
     /**
      * Dispose of all the eventlisteners and clears the observables
      */
     public dispose() {
-        this.onDeviceConnectedObservable.clear();
-        this.onDeviceDisconnectedObservable.clear();
-
```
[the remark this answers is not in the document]
PolygonalSun: I converted those observables back to callbacks.  I'll update the comment.

PR #8037, opened 2020-04-19
Correct query params in pg

RaananW:
When an external website adds query params to our pg links the PG didn't work.

It now parses the query only under specific terms (pathname should be empty/base)

### DIFF
--- a/Playground/js/main.js
+++ b/Playground/js/main.js
@@ -566,7 +566,7 @@ class Main {
                         exampleList.appendChild(noResultContainer);
                     }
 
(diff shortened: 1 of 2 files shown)


Popov72:
Note that:
```
if (!location.hash && restoreVersionResult == false && (location.pathname === '/' 
       || !location.pathname)) {
```
makes my local access to the playground not work anymore. It does work when reverting to:
```
if (!location.hash && restoreVersionResult == false) {
```
that's because `location.pathname` = `/Playground/index-local.html`

RaananW:
Will fix that ASAP

PR #8038, opened 2020-04-19
Fix bug on resizing + glow or highlight layer when multiple scenes

Popov72:
Follow up to #8029, it's the same problem for the glow and highlight layers.


src/Layers/glowLayer.ts
```diff
@@ -316,13 +316,14 @@ export class GlowLayer extends EffectLayer {
...
-                if (internalTexture) {
+                let internalTexture2 = this._blurTexture2.getInternalTexture();
+                if (internalTexture2) {
                     this._scene.postProcessManager.directRender(
                         this._postProcesses2,
-                        internalTexture,
+                        internalTexture2,
                         true);
                 }
+                this._engine.unBindFramebuffer(internalTexture2 ?? internalTexture, true);
```
RaananW, line 326: Just wondering (and by all means not asking for a change) why use the ?? instead of ||? Is there a chance that a texture will be falsy but still have a value?

Popov72: Now that `??` is supported in typescript, I tend to use it as much as possible over `||`, which may fail in some circumstances. Also, I think `??` is clearer, because it is really a check for `null` that I want here, whereas a logical or is more cryptic for the matter.

PR #8040, opened 2020-04-20
Serializing / parsing missing properties for mesh instances.

julien-moreau:


src/Meshes/mesh.ts
```diff
@@ -3022,6 +3024,24 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
             } else if (instance.rotation) {
                 serializationInstance.rotation = instance.rotation.asArray();
             }
+
+            // Physics
+            //TODO implement correct serialization for physics impostors.
```
julien-moreau, line 3029: Just copy/pasted the code above to discuss about the comment :)
Should we keep this copy/paste until the correct serialization is done and just call impostor.serialize() ? Or wait to merge this PR until we get the correct serialization ?

deltakosh: fine with it ;)

julien-moreau: You rox :)

PR #8056, opened 2020-04-22
Added support for CSG Vertex Color

PirateJC:
https://forum.babylonjs.com/t/csg-inheritance-vertexcolor/10057


sebavan:
@PirateJC the tests are failing on the previous CSG test.

image

You can launch it locally with : http://localhost:1338/tests/validation?CSG

I guess you need to take into account the case where the meshes do not have vertex colors :-)

PR #8057, opened 2020-04-23
fix: Web Workers have no window

brianzinn:
This is to allow Web Workers to perform things like `SceneLoader.ImportMesh` without having to switch engine to offline (`enableOfflineSupport = false`).


brianzinn:
Just a minor fix, so nothing "new".

deltakosh:
Thanks!

PR #8064, opened 2020-04-24
Fixes to update UI state when frames and ports are connected

belfortk:
Fixes for 3 bugs found:

• Node outports connected to another node inside the same frame no longer have their "Expose on Frame" toggle disabled by default 

• Deleting a link from an in-frame un-exposed node outport to an outside node now removes the automatic frame port

• Creating links from nodes ports inside the frame now update the node port property panel


nodeEditor/src/diagram/graphFrame.ts
```diff
@@ -191,12 +194,9 @@ export class GraphFrame {
...
             return;
         }
-        this.ports.forEach((framePort:FrameNodePort) => {
-            framePort.dispose()
-        });
-
-        this._createFramePorts();
-        this.ports.forEach((framePort: FrameNodePort) => framePort.node._refreshLinks());
+        this.isCollapsed = !this.isCollapsed
+        this.isCollapsed = !this.isCollapsed
```
deltakosh, line 198: Suspicious

belfortk: Agreed! I was running into issues with some node links not being drawn correctly after other unrelated ones were deleted and this "hack" fixed them.

I wanted to come up with a better way before officially submitting this pr.

PR #8067, opened 2020-04-24
buffers: fix crash & improve creation

ottoville:
Creating a zero sized buffer would crash the tab and give "status access violation" error in chromium based browser. Now it throws instead.
Also improved the buffer creation, if possible the data is uploaded right away without temporal buffers.


sebavan:
Thanks :-)

PR #8070, opened 2020-04-24
SPS feature added : vertex color and uv update

jbousquie:
In the SPS, now particle vertex colors and uvs can be updated as well as positions in the call to `updateParticleVertex()` when this mode is enabled.
Example : 
```
  SPS.updateParticleVertex = function(particle, vertex, i) {
      if (i > 2 && i < 6) {
          vertex.position.x -= 0.5;
          vertex.color.r = 1;
          vertex.color.g = 1;
          vertex.color.b = 1;
          vertex.uv.x *= Math.sin(k);
          vertex.uv.y *= Math.sin(k);
      }
  }
```
Note that the former way to set the vertex position still work also :  
```
vertex.x += 0.2;
vertex.y += 0.2;
```
So no change for the existing user code ;-)

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -105,6 +105,7 @@
 ### Particles
 
 - Added local space support for GPU particles ([CraigFeldpsar](https://github.com/craigfeldspar))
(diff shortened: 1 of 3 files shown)


deltakosh:
Love you for taking in account back compat :)

Hope you are doing well my friend!

PR #8074, opened 2020-04-24
Added support for vert color inheritance for CSG objects

PirateJC:


sebavan:
One last issue :-)

image

You have the line number and the problem.

just fix and push, that will all update for you.

PR #8075, opened 2020-04-25
Further fix of updateOptions method of Sound class

randle-mcmurphy:
The `updateOptions` method of `Sound` class still has another bug (even after latest fix). If `offset` and/or `length` properties are not set in the `ISoundOptions` parameter, it assigns the private members `_offset` and/or `_length` to undefined. So if they were set in the constructor, their values are lost.

This playground will demonstrate the problem: https://www.babylonjs-playground.com/#DIJA7S#2

Notice that after calling `updateOptions`, the whole track is played, instead of only playing the part defined by `offset` and `length`.

My PR should fix the issue.

### DIFF
--- a/src/Audio/sound.ts
+++ b/src/Audio/sound.ts
@@ -459,8 +459,8 @@ export class Sound {
             this.refDistance = options.refDistance ?? this.refDistance;
             this.distanceModel = options.distanceModel ?? this.distanceModel;
             this._playbackRate = options.playbackRate ?? this._playbackRate;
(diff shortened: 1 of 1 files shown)


randle-mcmurphy:
Sorry for the mess 🤦

RaananW:
I'm totally fine with this change, but please submit it to masterand not to preview

randle-mcmurphy:
@RaananW yes, I realized my mistake after I submitted. Please feel free to close the PR.

PR #8078, opened 2020-04-26
Path3D _distances / length computation fix

Poolminer:
The `_distances` array is supposed to contain the distance from the first point to the i​-th point along the path, but in `_compute`, if the i-​th point is distance 0 away from the previous point, it will add the distance of an earlier point instead (`prev = this._getLastNonNullVector(i)`).

The length of [0.0, 0.5, 0.7, 0.7, 1.0] should be 1.0, but with `Path3D` it will say 1.2 (the second 0.7 will be compared to the 0.5):

https://playground.babylonjs.com/#J4DF1M

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -2,14 +2,14 @@
 
 ## Major updates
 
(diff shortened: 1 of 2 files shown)


deltakosh:
Thanks!

PR #8091, opened 2020-04-30
Add Support to Auto-Update Inspector when Switching Scenes

belfortk:
Previously, the Inspector would not update when switching scenes. You would have to close it and re-open it.

With this PR, you can call `debugLayer.setAsActiveScene()` to smoothly update the scene in the inspector.


inspector/src/components/sceneExplorer/sceneExplorerComponent.tsx
```diff
@@ -79,6 +80,12 @@ export class SceneExplorerComponent extends React.Component<ISceneExplorerCompon
         this.sceneMutationFunc = this.processMutation.bind(this);
 
         this.sceneExplorerRef = React.createRef();
+        this._onNewSceneObserver = this.props.globalState.onNewSceneObservable.add((scene: Scene) => {
+            this.setState({
+                ...this.state,
```
deltakosh, line 85: this line is not required

belfortk: updated!

PR #8094, opened 2020-05-01
Light gizmo now uses Utility Layer Scene instead of Original Scene fo…

julien-moreau:
…r its material and transform node.


julien-moreau:
@TrevorDev do you confirm that it should be in utility layer scene instead of original scene?
Thanks :)

CedricGuillemet:
@julien-moreau I'm not sure. What's the issue you are trying to fix?

julien-moreau:
> @julien-moreau I'm not sure. What's the issue you are trying to fix?

My use case is that I’m using light gizmo in a scene editor that exports all materials, meshes, etc of the scene being edited (using the scene serializer) and I think we don’t want that material and transform node being exported each time

PR #8095, opened 2020-05-01
Adding filtering for HDR maps

### LINKED ISSUE #7867: Support RealTime filtering of cube maps for PBR
This could help with real time probes, support for unprefiltered environments. We could inspire ourself from :

https://developer.nvidia.com/gpugems/gpugems3/part-iii-rendering/chapter-20-gpu-based-importance-sampling

Or newer techniques.

CraigFeldspar:
Closes #7867
.hdr environment files that are loaded will now give accurate PBR reflections.

## Description
Added a scene component to filter HDR cube maps on load. Filtering is achieved by importance-sampling a GGX distribution and applying it on different mip levels of a render target, corresponding to different roughness values.

Realtime filtering capacities were also added on PBR material, though it's not really efficient compared to the above baking-on-load method. It should be used only for dynamic reflection textures like probes.

Implementation was based on IBLBaker shaders and the following article : 
http://blog.selfshadow.com/publications/s2013-shading-course/karis/s2013_pbs_epic_notes_v2.pdf

## Performance impact
**Prefiltering of the cube map is activated by default on HDRCubeTexture**, thus resulting in a slight impact in the loading time of the HDR (offscreen renderings must be performed to filter the map).  

## Additionnal notes
HDRCubeTexture could now be a child class from CubeTexture, or be merged with CubeTexture to simplify the API and avoid code duplication.

## Example PG
http://localhost:1338/Playground/index-local.html#P2NE2S#2

Top sphere is a prefiltered DDS environment
Bottom left sphere is realtime prefiltered environment with 16 samples
Bottom right sphere is a .hdr texture prefiltered on load with 128 samples

Slider adjusts roughness of all 3 spheres.


sebavan:
Ok I reviewed more in depth so we ll take it offline in meeting next week, I do not think it would be lots of changes but mainly in the used formula, precomputed values and such. I ll try to share the formulas before end of day Monday in includes so  that we can all pick and choose the same :-)

CraigFeldspar:
All changes as discussed offline have been applied

New PG to try with probe filtering, looks pretty good :)
http://localhost:1338/Playground/index-local.html#P2NE2S#3

I'm still wondering if we should set `reflectionTexture.realtimeFiltering` to true by default on probes (with low quality), since : 
 + enabling this flag will have no effect on standardMaterial having this reflectionTexture
 + on PBR, with this flag disabled, material BDRF becomes really wrong since roughness is not taken into account

sebavan:
> I'm still wondering if we should set reflectionTexture.realtimeFiltering to true by default on probes (with low quality)

Yes pretty please :-)

sebavan:
Actually no :-) cause it couples as in the PR hdrFiltering and baseTexture meaning any ppl who are not using it still have it in their bundles.

I think all is good in the PR except this part :-) It is looking great actually. 

I ll make a thorough review in a bit Thanks a lot !!!

src/Materials/Textures/baseTexture.ts
```diff
@@ -9,6 +9,7 @@ import { Constants } from "../../Engines/constants";
 import { IAnimatable } from '../../Animations/animatable.interface';
 import { GUID } from '../../Misc/guid';
 import { ISize, Size } from '../../Maths/math.size';
+import { HDRFiltering } from "./Filtering/hdrFiltering";
```
sebavan, line 12: Actually this could be placed in Engine.Constants and reuse from constants here to prevent the coupling and bundle size if not in use :-)

src/Shaders/ShadersInclude/helperFunctions.fx
```diff
@@ -73,6 +73,27 @@ float square(float value)
     return value * value;
 }
 
+// https://learnopengl.com/PBR/IBL/Specular-IBL
+// Hammersley
+float radicalInverse_VdC(uint bits)
```
sebavan, line 78: This should be enclosed in #ifdef WEBGL2 and else use the fallback as defined here :

https://learnopengl.com/PBR/IBL/Specular-IBL

in the section Hammersley sequence without bit operator support

```

float VanDerCorpus(uint n, uint base)
{
    float invBase = 1.0 / float(base);
    float denom   = 1.0;
    float result  = 0.0;

    for(uint i = 0u; i < 32u; ++i)
    {
        if(n > 0u)
        {
            denom   = mod(float(n), 2.0);
            result += denom * invBase;
            invBase = invBase / 2.0;
            n       = uint(float(n) / 2.0);
        }
    }

    return result;
}
// ----------------------------------------------------------------------------
vec2 HammersleyNoBitOps(uint i, uint N)
{
    return vec2(float(i)/float(N), VanDerCorpus(i, 2u));
}
```

sebavan:
I ll merge yours before my gltf extensions one as soon as you have done the tiny changes so I ll do the merge :-)

CraigFeldspar:
Decoupled for lighter bundles and enabled by default on probes ! 

new PG : http://localhost:1338/Playground/index-local.html#P2NE2S#4

PR #8097, opened 2020-05-01
NME: Create Post Processes

Popov72:
Added this entry:
image
(two times "Mode", maybe not the best...)

When switched in "PostProcess" mode, the default shader code is:
image

We said that we would not allow vertex code editing in the first implementation, but that was easier for me to have the vertex code editable, and at some point we would have allowed the user to edit the vertex code anyway, so...

There's nothing in the `CurrentScreen` texture, so the preview area is black at start. Maybe we can put something has a default texture, but which one? Where to load it from?

Also, in "PostProcess" mode, the **Animation** and **Mesh** menus are removed from the left menu, and a **PostProcess** entry is added:
image

All the buttons around the preview area are removed too, except the one that opens the preview in a new window:
image

On coding side, when you have built your material, you can create the `PostProcess` corresponding to the material by calling `nodeMaterial.createPostProcess(...);`. I don't call this method automatically after a build because it needs some parameters, notably the camera and/or the engine.


deltakosh:
Excellent!!

Yeah for the mode, actually you were right, having it as first entry in general would be better.

Let's use the one you captured and store it https://github.com/BabylonJS/Babylon.js/tree/master/assets in a new nme folder?

nodeEditor/src/components/preview/previewAreaComponent.tsx
```diff
@@ -26,78 +27,90 @@ export class PreviewAreaComponent extends React.Component<IPreviewAreaComponentP
         this._onIsLoadingChangedObserver = this.props.globalState.onIsLoadingChanged.add((state) => this.setState({isLoading: state}));
     }
 
+    componentDidMount() {
+        this.eventNMERefresh = this.eventNMERefresh.bind(this);
```
deltakosh, line 31: why that?

Popov72: Because I have not been able to make the `nodeListComponent` refresh by other means...

When switching to/from "PostProcess" / "Material" modes, the `nodeListComponent` must refresh because the list of blocks depend on the mode.

I have reused the same method to have the preview area refresh too, as I don't really know how to make it refresh by other means. I guess there's always the possibility to add some observables on the global state to make this communication, but it felt wrong to do that just for a single call for refresh...

deltakosh:
Yep we need to avoid that as this could be problematic for bundler.

Try doing like that: https://github.com/BabylonJS/Babylon.js/blob/master/src/Meshes/mesh.ts#L39

Popov72:
> Try doing like that: https://github.com/BabylonJS/Babylon.js/blob/master/src/Meshes/mesh.ts#L39

Yes I already tried that but it does not work, because I need to access properties from `NodeMaterial` instances when in `CurrentScreenBlock`, and I need to create an instance/access property of `CurrentScreenBlock` from `NodeMaterial`...

Well, in `CurrentScreenBlock` I only need to access `NodeMaterial.IgnoreTexturesAtLoadTime`, I guess I can remove this

Popov72:
Circular dependency solved, `Mode` moved to the **General** section and a default texture added to the `CurrentScreenBlock`.

nodeEditor/src/components/preview/previewAreaComponent.tsx
Popov72: Replaced by observables.

PR #8099, opened 2020-05-02
check zero buffer update, reformat pipeline layout, parallel shader loading

ottoville:
- Fix unnecessary mapped buffer on buffer creation in certain cases
- Do not kill tab in chrome if updating buffer with 0 size data
- Use getBindGroupLayout to create a layout if necessary.
- Load fragment shader while vertex shader is being processing.

### DIFF
--- a/src/Engines/webgpuEngine.ts
+++ b/src/Engines/webgpuEngine.ts
@@ -298,9 +298,6 @@ export class WebGPUEngine extends Engine {
         return Promise.reject("gslang is not available.");
     }
 
-    public getDevice(): GPUDevice {
-        return this._device;
-    }
     private _initializeLimits(): void {
         // Init caps
         // TODO WEBGPU Real Capability check once limits will be working.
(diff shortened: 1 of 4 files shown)


src/Engines/webgpuEngine.ts
```diff
@@ -298,6 +298,9 @@ export class WebGPUEngine extends Engine {
         return Promise.reject("gslang is not available.");
     }
 
+    public getDevice(): GPUDevice {
```
sebavan, line 301: I would prefer to not expose any WebGPU specific types out of engine and keep all specific work enclosed inside. 

Is that possible ?

src/Engines/webgpuEngine.ts
```diff
@@ -569,23 +579,30 @@ export class WebGPUEngine extends Engine {
...
+                new Uint8Array(uploadMapping).set(new Uint8Array(src.buffer, srcByteOffset + offset, uploadCount));
+                uploadBuffer.unmap();
+                commandEncoder.copyBufferToBuffer(
+                    uploadBuffer, 0,
+                    buffer, dstByteOffset + offset,
+                    uploadCount);
+            }
+            this._device.defaultQueue.submit([commandEncoder.finish()]);
+        } catch (e) {
+            Logger.Error(e);
```
sebavan, line 602: would be coool if you could prefix with a specific message as well

src/Materials/effect.ts
```diff
@@ -669,6 +683,65 @@ export class Effect implements IDisposable {
         }
     }
 
+    public getPipelineLayout(): GPUPipelineLayout {
```
sebavan, line 686: This should not be moved in effect as the goal is keep any WebGPU specifics in WebGPUEngine only. People using WebGL, Native or Null engine should not have this code in their distrib.

I guess it would also prevent the GpuDevice exposed ?

ottoville:
I restored the _getPipelineLayout method into engine but also modified the _getBindGroupsToRender method to create a layout with getBindGroupLayout if layout has not been created.
The functionality has been described here https://github.com/gpuweb/gpuweb/issues/446
This will make _getPipelineLayout method totally unnecessary, it can be deleted and still engine works. I did not deleted it as I dont know how this will affect on performance.

sebavan:
Love it !!! yup do not delete for now but will do in a bit I guess :-)

PR #8103, opened 2020-05-05
Add Sample Animation & Keyframe Dragging

toledoal:
In this PR the user can drag&drop the animation keyframes. 
The svgDraggableArea would be the basis for the graph section.
keyframeSvgPoint is the SVG representation of a keyframe, it may have two control points to modify the curve.

Also, there is a way to add new simple animations. 
 
Limitations:

1. Users can only add ANIMATIONTYPE_FLOAT animations. (working to expand that of course)
2. Users cannot add new keyframes in the editor.
3. Users cannot drag/modify control points.


deltakosh:
I'm getting an error when moving the last key to the left on this PG:
https://www.babylonjs-playground.com/#QYFDDP#1

toledoal:
Thanks! I'll check it

inspector/src/components/actionTabs/tabs/propertyGrids/animations/svgDraggableArea.tsx
```diff
@@ -0,0 +1,132 @@
...
+            } else {
+                return undefined;
+            }
+
+        } else {
+            return undefined;
+        }
+       
+      }
+
```
deltakosh, line 98: Nit: Please keep only one line return between blocks

toledoal:
So, after I update the keys with setKeys(), and the animation continues to render, the code breaks in animation.ts when the key gets incremented beyond length of the keys. Maybe since we are reducing the frames for the animation we get into this point. I'll keep exploring. This bug happens if the animation is currently playing. If we stop it, then drag the keyframe and resume playing it doesn't happen. A possible solution is to stop the animation when users redefine the last keyframe.

image

deltakosh:
Yes I think it is fair to pause the animation when manipulating the curve

toledoal:
I don't know exactly why the validation test failed. It is something on my end?

PR #8118, opened 2020-05-07
StandardMaterial - RGBD Support

MackeyK24:
Added RGBD ReflectionTexture, RefractionTexture and LightmapTexture support

### DIFF
--- a/src/Shaders/default.fragment.fx
+++ b/src/Shaders/default.fragment.fx
@@ -255,6 +255,9 @@ void main(void) {
 
 #ifdef LIGHTMAP
 	vec3 lightmapColor = texture2D(lightmapSampler, vLightmapUV + uvOffset).rgb * vLightmapInfos.y;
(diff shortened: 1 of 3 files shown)


src/Shaders/default.fragment.fx
```diff
@@ -255,6 +255,9 @@ void main(void) {
 
 #ifdef LIGHTMAP
 	vec3 lightmapColor = texture2D(lightmapSampler, vLightmapUV + uvOffset).rgb * vLightmapInfos.y;
```
Popov72, line 257: The `* vLightmapInfos.y` multiplication should be done after the `#ifdef RGBDLIGHTMAP` block.

deltakosh: good catch.. can you PR pretty please?

Popov72: @MackeyK24 beat me to it!

PR #8123, opened 2020-05-07
Graph canvas

toledoal:
This PR expands on the curve editor frame and value axes. Idea is to set scroll. The main purpose of this functionality is to allow users to "navigate" through their animation.

canvas


deltakosh:
Pinging @PatrickRyanMS : Do we want a scroll or do we always want the graph to be fully represented?

PatrickRyanMS:
We definitely need to be able to zoom and scroll (if the zoom level cannot represent the whole curve). When animating, you need to be able to see only a section of the curve for some tasks, while being able to see larger portions other times. 

Ideally, if we select a couple of keys, we have the ability to zoom to fit those parts of the curve to the panel. This can either be from a button or key shortcut depending on what is possible.

PR #8128, opened 2020-05-08
GeometryBufferRenderer: Fix problem with alpha tested objects

Popov72:
See also https://forum.babylonjs.com/t/alpha-from-albedo-works-in-sandbox-but-not-in-my-scne/10646/6


julien-moreau:
Nice catch !!!

PR #8131, opened 2020-05-08
NME: Add particle fragment editing support

Popov72:
I wanted to support as much as possible all the features of the particle system, so I had to perform some changes in `ParticleSystem` and `Engine.createEffectForParticles`:
* several different custom effects instead of just one (based on the blend mode)
* refactoring of some code to be able to use it from the outside (`fillDefines`, `fillUniformsAttributesAndSamplerNames`)
* `onBeforeDrawParticlesObservable` to be able to update the effect uniforms
* adding support for custom effects to the gpu particle systems so that shaders created in the nme can also be used for them

With those changes, all features are supported in the node material (except for the clip planes).

However, one should note that everything is still driven by the parameters of the particle system instance! For eg, if the current particle system displayed in the preview area does not use "ramp gradients", the `RampGradientBlock` does nothing, it's just a pass-through. Same thing for the `BlendMultiply` block. The exception is the `ParticleTexture` block: if you provide a texture for this block it will be used as the particle texture in the preview area, else the texture defined in the current particle system will be used.

For information, the default particle shader is:
image

It handles all the features of the particle system except for `textureMask`, but it can be added thanks to the `ParticleTextureMask` block. The full particle fragment with all features is:
image


src/Materials/Node/nodeMaterial.ts
```diff
@@ -1267,6 +1378,48 @@ export class NodeMaterial extends PushMaterial {
...
+        const uv = new InputBlock("uv");
+        uv.setAsAttribute("particle_uv");
+
+        const texture = new ParticleTextureBlock("ParticleTexture");
+        uv.connectTo(texture);
+
+        const color = new InputBlock("Color");
+        color.setAsAttribute("particle_color");
+
+        const multiply = new MultiplyBlock("texture * color");
```
deltakosh, line 1399: let me be picky :)

```suggestion
        const multiply = new MultiplyBlock("Texture * Color");

```

deltakosh: I'll fix it :D

Popov72: Arf!

PR #8133, opened 2020-05-09
Add Functionality to retrieve all Meshes affected by a highlight layer 

erik-pilgrim:
Rather than have to keep a list of highlighted elements, we can expose the private property _meshes via a get function to retrieve all affected Meshes. This assists in the selection and de-selection of multiple items within a scene.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -23,7 +23,7 @@
 - Added support preserving vert colors for CSG objects ([PirateJC](https://github.com/PirateJC))
 - Added support in `ShadowGenerator` for fast fake soft transparent shadows ([Popov72](https://github.com/Popov72))
 - Added `boundingBoxRenderer.onBeforeBoxRenderingObservable` and `boundingBoxRenderer.onAfterBoxRenderingObservable` ([Deltakosh](https://github.com/deltakosh))
-
+- Added support to return all affected Meshes by a HighlightLayer ([erik-pilgrim](https://github.com/erik-pilgrim))
 ### Engine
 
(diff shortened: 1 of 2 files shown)


sebavan:
The main issue is that it is easy to break anything if the list is not being used in readonly mode. I am wondering if it should be exposed, @deltakosh ?

deltakosh:
well we can return a copy?

sebavan:
Yup agree, would be better, could you change the PR for that ?

erik-pilgrim:
@sebavan sure- updated the PR

deltakosh:
Linting is failing :)

deltakosh:
Well there is something wrong with the PR as all tests are failing:)

deltakosh:
Ping?

erik-pilgrim:
Hey @deltakosh attempting to use the gitpod to build and find what the problem is, but after 40 min of build time, i cant delete or start over. I'll give this a bit more time before starting it over- but this is a styling issue, not a functional issue.

sebavan:
It is actually not functional only:

image

erik-pilgrim:
Guys, I’ve tried literally all day to use your gitpod system in an attempt to build this and it will not load. We can reject this pr and I’ll reopen when I can get it working. I don’t have the space on my local box to pull and build the repo this minute but will shortly.
Also, any reasoning on why it wouldn’t be at least castable to an array type?

________________________________

sebavan:
no problem I ll close for now.

You can not put it in an array because it looks like this:
{
  "1": null,
  "2":  Mesh,
  "3":  Mesh
}

PR #8135, opened 2020-05-10
Added support of metadata in Sound class.

julien-moreau:


RaananW:
What is the exact usecase for this? What is the difference between this and tagging it?

deltakosh:
Well at least it is uniform with the rest of the code :)

julien-moreau:
@RaananW right, was just to keep It uniform
I’m using metadatas in my project and it was very useful to have it on sounds as well

PR #8136, opened 2020-05-10
TransformNode as spatial sound source fix

Poolminer:
https://forum.babylonjs.com/t/spatial-sound-sound-does-not-follow-a-transformnode-but-a-mesh/10717/

### DIFF
--- a/src/Audio/sound.ts
+++ b/src/Audio/sound.ts
@@ -937,17 +937,18 @@ export class Sound {
     }
 
     private _onRegisterAfterWorldMatrixUpdate(node: TransformNode): void {
-        if (!(<any>node).getBoundingInfo) {
-            return;
-        }
-        let mesh = node as AbstractMesh;
         if (this._positionInEmitterSpace) {
-            mesh.worldMatrixFromCache.invertToRef(TmpVectors.Matrix[0]);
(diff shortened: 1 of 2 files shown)


src/Audio/sound.ts
```diff
@@ -937,17 +937,18 @@ export class Sound {
...
         if (this._positionInEmitterSpace) {
-            mesh.worldMatrixFromCache.invertToRef(TmpVectors.Matrix[0]);
+            node.worldMatrixFromCache.invertToRef(TmpVectors.Matrix[0]);
             this.setPosition(TmpVectors.Matrix[0].getTranslation());
         }
         else {
-            let boundingInfo = mesh.getBoundingInfo();
-            this.setPosition(boundingInfo.boundingSphere.centerWorld);
+            if (!(<any>node).getBoundingInfo) {
+                this.setPosition(node.position);
```
RaananW, line 946: node has an absolute position, it will be the right position to use since the sound is not attached to its parent

deltakosh:
OIk all good after the fix requested by Raanan

src/Audio/sound.ts
Poolminer: > 
> 
> OIk all good after the fix requested by Raanan

Thanks, the fix is in

deltakosh:
You rock! thanks

PR #8154, opened 2020-05-12
Draft: inline functions in shader code

Popov72:


src/Engines/Processors/shaderProcessor.ts
```diff
@@ -23,7 +24,10 @@ export class ShaderProcessor {
     public static Process(sourceCode: string, options: ProcessingOptions, callback: (migratedCode: string) => void) {
         this._ProcessIncludes(sourceCode, options, (codeWithIncludes) => {
             let migratedCode = this._ProcessShaderConversion(codeWithIncludes, options);
-            callback(migratedCode);
+            let sci = new ShaderCodeInliner(migratedCode);
+            sci.debug = true;
```
deltakosh, line 28: do we want to keep it in debug?

sebavan: we wont even keep it in GL :-), nw I flag as resolve as this file will not be in the final PR.

src/Engines/Processors/shaderProcessor.ts
```diff
@@ -9,6 +9,7 @@ import { ShaderDefineExpression } from './Expressions/shaderDefineExpression';
 import { ShaderDefineArithmeticOperator } from './Expressions/Operators/shaderDefineArithmeticOperator';
```
sebavan, line 9: could you simply revert this file ??? and we ll merge it. It works beautifully well so far in WebGPU, so it should be the same in native , thanks a ton.

@syntheticmagus @CedricGuillemet @bghgary to keep on your side the integration snippet for native :-)

Popov72:
Rollback done!

src/Engines/Processors/shaderProcessor.ts
Popov72: Done

PR #8163, opened 2020-05-13
DX and UX improvements for Gizmos

edzis:
Done:
- Show a different material while gizmo is being dragged
- Expose gizmo materials for color changes
- Optionally apply the same materials  to custom meshes based on state changes
- Less code through generalization
- Do not apply hover material to other meshes while one is being dragged
- What's new

TODO:
- Show Pointer cursor on hover - not sure how
- Apply also to BoundingBoxGizmo - current setup too complex
- Usage examples in documentation

Tested via https://gist.github.com/edzis/461dd1f6036509ea942088dbb1a43130

### DIFF
--- a/src/Gizmos/axisDragGizmo.ts
+++ b/src/Gizmos/axisDragGizmo.ts
@@ -16,7 +16,7 @@ export class AxisDragGizmo extends DraggableGizmo {
     private _arrow: TransformNode;
 
     /** @hidden */
-    public static _CreateArrow(scene: Scene, material?: StandardMaterial): TransformNode {
+    public static _CreateArrow(scene: Scene, material: Nullable<StandardMaterial> = null): TransformNode {
         var arrow = new TransformNode("arrow", scene);
         var cylinder = CylinderBuilder.CreateCylinder("cylinder", { diameterTop: 0, height: 0.075, diameterBottom: 0.0375, tessellation: 96 }, scene);
         var line = CylinderBuilder.CreateCylinder("cylinder", { diameterTop: 0.005, height: 0.275, diameterBottom: 0.005, tessellation: 96 }, scene);
@@ -24,7 +24,7 @@ export class AxisDragGizmo extends DraggableGizmo {
         line.parent = arrow;
 
         // Position arrow pointing in its drag axis
-        cylinder.material = material ?? null;
+        cylinder.material = material;
         cylinder.rotation.x = Math.PI / 2;
         cylinder.position.z += 0.3;
         line.position.z += 0.275 / 2;
--- a/src/Gizmos/axisDragGizmo.ts
+++ b/src/Gizmos/axisDragGizmo.ts
@@ -44,9 +44,9 @@ export class AxisDragGizmo extends DraggableGizmo {
 
     /**
      * Creates an AxisDragGizmo
-     * @param gizmoLayer The utility layer the gizmo will be added to
      * @param dragAxis The axis which the gizmo will be able to drag on
      * @param color The color of the gizmo
+     * @param gizmoLayer The utility layer the gizmo will be added to
      */
     constructor(dragAxis: Vector3, color: Color3 = Color3.Gray(), gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer, parent: Nullable<PositionGizmo> = null) {
         super({ dragAxis }, color, gizmoLayer, parent);
--- a/src/Gizmos/axisDragGizmo.ts
+++ b/src/Gizmos/axisDragGizmo.ts
@@ -1,43 +1,19 @@
-import { PointerDragBehavior } from "../Behaviors/Meshes/pointerDragBehavior";
-import { PointerInfo } from "../Events/pointerEvents";
 import { Color3 } from "../Maths/math.color";
 import { Matrix, Vector3 } from "../Maths/math.vector";
-import { AbstractMesh } from "../Meshes/abstractMesh";
 import { CylinderBuilder } from "../Meshes/Builders/cylinderBuilder";
 import { Mesh } from "../Meshes/mesh";
 import { TransformNode } from "../Meshes/transformNode";
-import { Observable, Observer } from "../Misc/observable";
 import { UtilityLayerRenderer } from "../Rendering/utilityLayerRenderer";
 import { Scene } from "../scene";
 import { Nullable } from "../types";
-import { Gizmo } from "./gizmo";
-import { GizmoMaterialSwitcher } from "./gizmoMaterialSwitcher";
+import { DraggableGizmo } from "./draggableGizmo";
 import { PositionGizmo } from "./positionGizmo";
 import { StandardMaterial } from "../Materials/standardMaterial";
 /**
  * Single axis drag gizmo
  */
-export class AxisDragGizmo extends Gizmo {
-    /**
-     * Drag behavior responsible for the gizmos dragging interactions
-     */
-    public dragBehavior: PointerDragBehavior;
-    private _pointerObserver: Nullable<Observer<PointerInfo>> = null;
-    /**
-     * Drag distance in babylon units that the gizmo will snap to when dragged (Default: 0)
-     */
-    public snapDistance = 0;
-    /**
-     * Event that fires each time the gizmo snaps to a new location.
-     * * snapDistance is the the change in distance
-     */
-    public onSnapObservable = new Observable<{ snapDistance: number }>();
-
-    private _isEnabled: boolean = true;
-    private _parent: Nullable<PositionGizmo> = null;
-
+export class AxisDragGizmo extends DraggableGizmo {
     private _arrow: TransformNode;
-    private _materialSwitcher: GizmoMaterialSwitcher;
 
     /** @hidden */
     public static _CreateArrow(scene: Scene, material?: StandardMaterial): TransformNode {
@@ -73,8 +49,7 @@ export class AxisDragGizmo extends Gizmo {
      * @param color The color of the gizmo
      */
     constructor(dragAxis: Vector3, color: Color3 = Color3.Gray(), gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer, parent: Nullable<PositionGizmo> = null) {
-        super(gizmoLayer);
-        this._parent = parent;
+        super({ dragAxis }, color, gizmoLayer, parent);
 
         // Build mesh on root node
         this._arrow = AxisDragGizmo._CreateArrow(gizmoLayer.utilityLayerScene);
@@ -83,18 +58,6 @@ export class AxisDragGizmo extends Gizmo {
         this._arrow.scaling.scaleInPlace(1 / 3);
         this._arrow.parent = this._rootMesh;
 
-        // Add drag behavior to handle events when the gizmo is dragged
-        this.dragBehavior = new PointerDragBehavior({ dragAxis: dragAxis });
-        this.dragBehavior.moveAttached = false;
-        this._rootMesh.addBehavior(this.dragBehavior);
-
-        // Create Material Switcher
-        this._materialSwitcher = new GizmoMaterialSwitcher(
-            color,
-            this.dragBehavior,
-            gizmoLayer._getSharedGizmoLight(),
-            gizmoLayer.utilityLayerScene
-        );
         this._materialSwitcher.registerMeshes(
             this._rootMesh.getChildMeshes(false)
         );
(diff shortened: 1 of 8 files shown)


src/Gizmos/axisDragGizmo.ts
```diff
@@ -1,56 +1,51 @@
...
+        material?: StandardMaterial
+    ): TransformNode {
         var arrow = new TransformNode("arrow", scene);
-        var cylinder = CylinderBuilder.CreateCylinder("cylinder", { diameterTop: 0, height: 0.075, diameterBottom: 0.0375, tessellation: 96 }, scene);
-        var line = CylinderBuilder.CreateCylinder("cylinder", { diameterTop: 0.005, height: 0.275, diameterBottom: 0.005, tessellation: 96 }, scene);
-        line.material = material;
+        var cylinder = CylinderBuilder.CreateCylinder(
+            "cylinder",
+            {
+                diameterTop: 0,
```
CedricGuillemet, line 27: It would be interesting to expose those values with a factor for scale.

src/Gizmos/axisDragGizmo.ts
```diff
@@ -74,41 +72,44 @@ export class AxisDragGizmo extends Gizmo {
...
-        this._rootMesh.addBehavior(this.dragBehavior);
 
         var localDelta = new Vector3();
         var tmpMatrix = new Matrix();
         this.dragBehavior.onDragObservable.add((event) => {
             if (this.attachedMesh) {
                 // Convert delta to local translation if it has a parent
                 if (this.attachedMesh.parent) {
-                    this.attachedMesh.parent.computeWorldMatrix().invertToRef(tmpMatrix);
+                    this.attachedMesh.parent
```
CedricGuillemet, line 104: Keep current code formatting

src/Gizmos/axisDragGizmo.ts
```diff
@@ -1,56 +1,51 @@
...
-    /**
-     * Event that fires each time the gizmo snaps to a new location.
-     * * snapDistance is the the change in distance
-     */
-    public onSnapObservable = new Observable<{ snapDistance: number }>();
-
-    private _isEnabled: boolean = true;
-    private _parent: Nullable<PositionGizmo> = null;
-
+export class AxisDragGizmo extends DraggableGizmo {
```
CedricGuillemet, line 15: I'm wondering if adding this new inherance will break compatibility with existing user code/PG

RaananW: Only an issue if the signature of the functions changed (and if someone used private functions/parameters).

RaananW:
As @CedricGuillemet mentioned, it will be great to expose all meshes (or at least the mesh-creation parameter) so that the users will be able to modify them as they see fit

src/Gizmos/axisDragGizmo.ts
edzis: The values are too interconnected between meshes and their position and scale. I don't believe in the ROI to do that.

I would rather make it easy to copy-paste the current mesh generation code and make adjustments. But why I did not do that is because there is an inexplicable scale factor outside the mesh creation code and in some cases the meshes seem to be explicitly hidden from the user through a transform node. Can you, please, explain that or give the green light to get rid of those restrictions?

src/Gizmos/axisDragGizmo.ts
edzis: I have explicitly kept the signature unchanged but can not judge the spread and expected stability of people using private fields. I have to defer on your call.
I have isolated the new feature from the refactor in separate commits. I believe the refactor adds value to the codebase (by removing code), but I am ok not to include that commit.

edzis:
@RaananW @CedricGuillemet Can you help me gather insights on 2 of the TODOS?

---

> Do not apply hover material to other meshes while one is being dragged - need to introduce scene-bound state

Are you ok with that idea? Are there some examples in the codebase I should follow for symmetry?

---

> Show Pointer cursor on hover - not sure how

Can you give some breadcrumbs on how to achieve that?

src/Gizmos/axisDragGizmo.ts
edzis: What formatting setup are you using not to defer to doing that manually?
The changes were introduced by VS Code auto-formatting fit Prettier with `"prettier.tabWidth": 4`.
If I disable Prettier, I am not getting much help from formatter.

RaananW:
> > Show Pointer cursor on hover - not sure how
> 
> Can you give some breadcrumbs on how to achieve that?

Sure - the CSS pointer cursor property is your friend - https://developer.mozilla.org/en-US/docs/Web/CSS/cursor



> > Do not apply hover material to other meshes while one is being dragged - need to introduce scene-bound state
> 
> Are you ok with that idea? Are there some examples in the codebase I should follow for symmetry?

It will be best to keep the framework with as little global states. I see a few ways to solve this (which does not mean they are the best ;-) ):

1) Introduce onDragging-Started and onDragingEnded observables, and make sure that all materials with hover material react to them
2) iterate through `scene.meshes` on every drag and "disable" all materials marked with hover. The issue here would be that you need to know what "disable" means so you will have to keep the original state
3) Find a central point (that is not the scene) that holds all references to "hoverable" objects, and make sure it controls the material instead of the gizmo itself.

Is there a cerntal gizmo holder?  would the gizmo manager be a good candidate for that?

src/Gizmos/axisDragGizmo.ts
RaananW: I am perfectly fine with that, I don't think it will be an issue at all. We don't back-compat private methods, only public facing APIs, and they stay the same. so all good.

src/Gizmos/axisDragGizmo.ts
CedricGuillemet: You can do a lint pass
```
gulp tsLint
```

src/Gizmos/axisDragGizmo.ts
CedricGuillemet: What about adding a single factor value in addition to the cylinders radii?

src/Gizmos/axisDragGizmo.ts
edzis: OK, then keeping as is.

src/Gizmos/axisDragGizmo.ts
edzis: That does not revert back to the previous formatting so I'll do that manually and avoid using prettier while coding.

edzis:
@RaananW 
> the CSS pointer cursor property is your friend

I am aware of that but I was hoping for some leads on how to use https://doc.babylonjs.com/api/classes/babylon.actionmanager#hovercursor or some of the other places where `hoverCursor` is exposed as that seems to be build exactly for this scenario - I am just failing to find insights about that.

src/Gizmos/axisDragGizmo.ts
edzis: Done

src/Gizmos/axisDragGizmo.ts
edzis: I would like not to include that in this PR - my changes do not influence anything regarding the way to achieve this.

RaananW:
> @RaananW 
> > the CSS pointer cursor property is your friend
> 
> I am aware of that but I was hoping for some leads on how to use https://doc.babylonjs.com/api/classes/babylon.actionmanager#hovercursor or some of the other places where `hoverCursor` is exposed as that seems to be build exactly for this scenario - I am just failing to find insights about that.

I assume you managed by now (sorry for a late reply), but if not you can see how it's done here https://github.com/BabylonJS/Babylon.js/blob/master/src/Inputs/scene.inputManager.ts#L185

src/Gizmos/axisDragGizmo.ts
```diff
@@ -1,56 +1,30 @@
...
     private _arrow: TransformNode;
-    private _coloredMaterial: StandardMaterial;
-    private _hoverMaterial: StandardMaterial;
 
     /** @hidden */
-    public static _CreateArrow(scene: Scene, material: StandardMaterial): TransformNode {
+    public static _CreateArrow(scene: Scene, material?: StandardMaterial): TransformNode {
```
RaananW, line 19: can be `Nullable<StandardMaterial> = null`. Then you don't need the `?? null` afterwards

src/Gizmos/draggableGizmo.ts
```diff
@@ -0,0 +1,125 @@
...
+    public onSnapObservable = new Observable<{ snapDistance: number }>();
+
+    protected _pointerObserver: Nullable<Observer<PointerInfo>> = null;
+    protected _materialSwitcher: GizmoMaterialSwitcher;
+
+    protected _isEnabled: boolean = true;
+    protected _parent: Nullable<Gizmo> = null;
+
+    /**
+     * Creates an InteractableGizmo
```
RaananW, line 37: incorrect documentation

src/Gizmos/draggableGizmo.ts
```diff
@@ -0,0 +1,125 @@
...
+    protected _attachedMeshChanged(value: Nullable<AbstractMesh>) {
+        if (this.dragBehavior) {
+            this.dragBehavior.enabled = value ? true : false;
+        }
+    }
+
+    /**
+     * If the gizmo is enabled
+     */
+    public set isEnabled(value: boolean) {
```
RaananW, line 73: maybe just "enabled"? also - documentation missing the parameter

src/Gizmos/gizmoMaterialSwitcher.ts
```diff
@@ -0,0 +1,197 @@
...
+        dragBehavior: PointerDragBehavior,
+        sharedGizmoLight: HemisphericLight,
+        scene: Scene
+    ) {
+        this._dragBehavior = dragBehavior;
+        this._sharedGizmoLight = sharedGizmoLight;
+        this._scene = scene;
+
+        // Create Materials
+        this._defaultMaterial = new StandardMaterial("_defaultMaterial", scene);
```
RaananW, line 52: is it possible to override these values? I noticed that the default material is exposed but maybe we can add an optional default material in the constructor options?

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -1,93 +1,59 @@
...
-     */
-    public onSnapObservable = new Observable<{ snapDistance: number }>();
-
-    private _isEnabled: boolean = true;
-    private _parent: Nullable<RotationGizmo> = null;
-
+export class PlaneRotationGizmo extends DraggableGizmo {
     /**
      * Creates a PlaneRotationGizmo
      * @param gizmoLayer The utility layer the gizmo will be added to
```
RaananW, line 17: nit - order of the params is wrong

src/Gizmos/draggableGizmo.ts
edzis: Here I am copy-pasting what existed before in each of the subclasses to keep the api and documentation unchanged.

src/Gizmos/gizmoMaterialSwitcher.ts
edzis: I am also a bit concerned restricting this to just StandardMaterial. But is there a likely need for users to have an easy way to use a different material class instead of just changing properties of these StandardMaterials? Would this even work in the utility layer with that single light?

If that is a realistic usecase then I would prefer not changing the signature but expose a static MaterialConstructor with the default implementation returning these standardMaterials. But then I would likely need to refactor for those to materials to be kept together in a single data structure.

Also, is there a real need for 3 materials instead altering the color of a single material. I see that the default material has specularity color but besides color otherwise they are identical.

src/Gizmos/draggableGizmo.ts
RaananW: Sounds good!

src/Gizmos/draggableGizmo.ts
edzis: Fixed

src/Gizmos/planeRotationGizmo.ts
edzis: Fixed

src/Gizmos/axisDragGizmo.ts
edzis: Fixed

src/Gizmos/gizmoMaterialSwitcher.ts
RaananW: due to backwards compatibility we cannot remove any public members. If you think however that it can be the same material, go right ahead and set those properties to return a single material. 

I don't think it needs to be more than a standard material, but it should be accessible for the user that's what I meant with overriding the values. As simple as possible, nothing too complex.

deltakosh:
Up?

RaananW:
@edzis Any updates on this one?

deltakosh:
Last call before I close it :)

PR #8165, opened 2020-05-13
Various fixes for realtime texture filtering

CraigFeldspar:
+ WebGL 1 compatibility
+ Realtime Filter uniforms now use UBO
+ Moving realtime filter properties from `BaseTexture` to `PBRBaseMaterial`
+ Enabling realtime filtering for clearcoat, refraction and translucency
+ Fixing cube seams when realtime filtering was activated
+ Fixing wrong irradiance direction when using probes (INV_CUBIC projection is now properly taken into account for irradiance direction)

Test playground : 
http://localhost:1338/Playground/index-local.html#P2NE2S#7


CraigFeldspar:
Still seems like in webgl1, prefiltering to RTT doesn't work.
When rendering to different mip levels, API states that when using `framebufferTexture2D`, lod level should be 0. It seems outdated though since we can chose lod level in WebGL2 and this isn't mentionned anywhere : 
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/framebufferTexture2D

Still looking into it, but we may have to disable prefiltering in webgl 1, keeping only realtime filtering.

CraigFeldspar:
Still seems like in webgl1, prefiltering to RTT doesn't work.
When rendering to different mip levels, API states that when using framebufferTexture2D, lod level should be 0. It seems outdated though since we can chose lod level in WebGL2 and this isn't mentionned anywhere :
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/framebufferTexture2D

Still looking into it, but we may have to disable prefiltering in webgl 1, keeping only realtime filtering.

sebavan:
Totally prefiltering is expected to not work in WebGL1. 

Thanks a lot I leave you to the DOC :-) and the sub surface and will try all the changes on my side but it looks pretty good.

Issue #6443, opened 2019-06-04
WebGPU Support


sebavan:
Here is the work which will happen until the release regarding our the WebGPU support:

- [x] Shader Auto Upgrade:
    - [x] Create Processor
    - [x] Auto attribs upgrade
    - [x] Auto varyings upgrade
    - [x] Auto textures upgrade
    - [x] Auto uniforms upgrade
    - [x] Auto UBOs detection
    - [x] Binding Upgrades
    - [x] Manage UBO in vertex + fragment
    - [x] Manage Array in UBOs for bones
    - [x] Manage Array in UBOs for morph targets
- [x] Manage Texture Sampling Modes
- [x] Manage Blending
- [x] Handle Canvas Resizing
- [x] Update WebGPU.d.ts
- [x] GLB support
- [x] Fix Full Build
- [x] Playground support
  - [x] Merge in Master
- [x] Dynamic Materials -> Defines Changes
- [x] Texture Support without WebGL back-end
- [x] Mapped Buffers
- [x] Render target support
- [x] Post Process support
- [x] First Cleanup and TODO WEBGPU pass
- [x] Raw Textures
- [x] Dynamic Texture Sampler Setting -> Filtering and Wrapping
- [ ] Safari Preview Support (if shader choice has been made #6695)
- [x] **Compute Shader support**
- [x] Support all the left over states (clear, stencil, depth, culling...)
- [x] Transform feedback and Occlusion Queries
- [ ] Optimize uniform buffers and groups (dynamic buffer index)
- [ ] **Big Cleanup, tests, and docs**
- [X] Command buffer "Reuse-Like" mechanism
- [ ] Spector.js Support
- [x] Last Optimization phase
- [X] Optimize UBOs (like IBL/Scene and so on)
- [ ] ** Release \o/**

If you are interested to contribute one of them, no problem, I am all about sharing ;-)

furcatomasz:
is it (list) already up to date?

sebavan:
Yep, I am almost ready to post the shader conversion which will enable most of our webgl shaders to be compatible. As soon as it is done I will finally be able to create a playground for WebGPU.

sebavan:
It is definitely getting there despite a still growing list of todos:
![image](https://user-images.githubusercontent.com/16655165/62504008-f1ea1b00-b7f6-11e9-867b-580f9b90f2d4.png)

sebavan:
[...]

PR #8173, opened 2020-05-13
Remove LookBehind from ShaderCodeInliner

syntheticmagus:
Alternative regex to remove lookbehind syntax from regex in shaderCodeInliner.ts.


sebavan:
Should help with #6443 about pg not working in FF Thanks a ton !!!

PR #8183, opened 2020-05-14
Add hook to update shader code before compilation

Popov72:
Should help for https://forum.babylonjs.com/t/recent-material-pipeline-changes/6885/12


Popov72:
I have tested this by inlining all pbr functions, it worked saved for a small error:
```
outParams.ambientOcclusionColor = mix(ambientOcclusionColor, aoStoreInMetalMap, vReflectivityInfos.z);
```
`ambientOcclusionColor` is a **in** parameter but also appears in `outParams.ambientOcclusionColor`, so making the latest wrong after the replace. I have just updated the name of the in parameter to `ambientOcclusionColorIn`.

PR #8208, opened 2020-05-18
Mesh: add thin instances

Popov72:
I took model on the regular instance implementation for the thin instance implementation:
* added a property `_thinInstanceDataStorage` of type `_ThinInstanceDataStorage` to hold the main data (much like `_instanceDataStorage` / `_InstanceDataStorage`).
* created an additional `_userThinInstanceBuffersStorage` to hold data for additional custom attributes other than the matrices (much like `_userInstancedBuffersStorage`). This way, if no custom attributes are added, no penalty paid

I have prefixed all methods by `thinInstance`, I can change that (notably, `addThinInstance` is `thinInstanceAdd` instead).

By default, the bounding box is recomputed to encompass all thin instances, but it can be disabled by setting `doNotSyncBoundingInfo` to `true` (already existing property).

As discussed, thin instances are always displayed with shader code assuming non uniform scaling, but with an optimized code (shearing not supported).


Popov72:
Hum, shouldn't I add an optional parameter to the `thinInstanceSetBuffer` method to create a static buffer instead of a dynamic one, in case the user know they won't update the buffer?

sebavan:
sounds totally fine to me :-)

sebavan:
Also should try to ensure we use the types that are not creating dupplicate in the engine updateBuffer ???

Popov72:
I also wonder if we shouldn't also let the user be able to set the current number of instances (but always < matrixBuffer.length / 16): that way, they would be able to allocate a bigger buffer at start in case they need to have more instances during the course of the program, they would not have to reallocate / copy a buffer each time they have to add an, instance.

Popov72:
Visual test added

PR #8212, opened 2020-05-18
Add shader source editor

atg:
Adds a [shader source editor](https://forum.babylonjs.com/t/shader-source-editor-in-debug-inspector-for-shadermaterial/10665?u=atg) to babylon.

Screen Shot 2020-05-18 at 20 21 58

image

You can edit both the fragment and vertex shaders, and it updates the scene in real time. The strip in the middle is the history bar which lets you roll over snapshots of your edits to compare or revert to different versions.

Not ready to merge yet, there's still some issues with the way things are packaged.


src/Materials/shaderMaterial.ts
```diff
@@ -862,6 +864,40 @@ export class ShaderMaterial extends Material {
         return false;
     }
 
+    private BJSNODEMATERIALEDITOR = NodeMaterial._getGlobalNodeMaterialEditor();
```
deltakosh, line 867: I'm REALLY not a big fan on introducing this dependency :(
If I'm playing with treeshacking, I do not want to get all the NME when only shadermaterial is required

nodeEditor/src/editDistance.ts
```diff
@@ -0,0 +1,261 @@
+// Derived from jsdiff by Kevin Decker et al
```
deltakosh, line 1: That is a lot of code and honestly I'm not sure we want it in the Shader material editor. Why do we need to get diffs?

nodeEditor/src/index.ts
```diff
@@ -1 +1,2 @@
-export * from "./nodeEditor";
\ No newline at end of file
+export * from "./nodeEditor";
+export * from "./sourceEditor/shaderSourceEditor";
```
deltakosh, line 2: This should not be here (I would like to avoid having dependencies between both editors)

atg:
@deltakosh I do know Monaco. The problem is, Monaco doesn't support GLSL and it's a huge multi-MB dependency.

I want something lightweight because the shader editor has to be dynamically loaded when the user clicks the "Edit Source" button. If I used Monaco then the user would suffer increased load times without actually deriving any benefit from it. CodeMirror does everything I need it to, it's 10x smaller, loads almost instantly and supports GLSL out of the box.

nodeEditor/src/index.ts
atg: You mean move it to project root? I just put it in there to save you the trouble of maintaining a separate npm module for it, currently it would share `babylonjs-node-editor`. It can be moved to project root and separated from the NME if desired.

deltakosh: No I meant, just move it into the inspector. we have other editor there. THe only one which is separated is NME because it is independant. No need to have a module for the Shader Material Editor

sebavan:
Looks better to me but I ll leave it open for @deltakosh to decide on the amount of code being added to the inspector. I am quite pleased it does not impact the core lib though :-)

@deltakosh is off for a couple days but should be able to look into it on Tuesday.

atg:
Yeah I've still got a few more things on my todo list but I've cleaned it up a lot.

src/Materials/Node/nodeMaterial.ts
```diff
@@ -151,10 +151,10 @@ export class NodeMaterial extends PushMaterial {
     /** Gets or sets a boolean indicating that node materials should not deserialize textures from json / snippet content */
     public static IgnoreTexturesAtLoadTime = false;
 
-    private BJSNODEMATERIALEDITOR = this._getGlobalNodeMaterialEditor();
+    private BJSNODEMATERIALEDITOR = NodeMaterial._getGlobalNodeMaterialEditor();
```
deltakosh, line 154: Can you revert your changes on NodeMaterial?

src/Materials/Node/nodeMaterial.ts
```diff
@@ -151,10 +151,10 @@ export class NodeMaterial extends PushMaterial {
     /** Gets or sets a boolean indicating that node materials should not deserialize textures from json / snippet content */
     public static IgnoreTexturesAtLoadTime = false;
 
-    private BJSNODEMATERIALEDITOR = this._getGlobalNodeMaterialEditor();
+    private BJSNODEMATERIALEDITOR = NodeMaterial._getGlobalNodeMaterialEditor();
 
     /** Get the inspector from bundle or global */
-    private _getGlobalNodeMaterialEditor(): any {
+    static _getGlobalNodeMaterialEditor(): any {
```
deltakosh, line 157: can you make it private static and then rename it to _**G**etGlobalNodeMaterialEditor ?

inspector/src/components/actionTabs/tabs/propertyGrids/materials/shaderMaterialPropertyGridComponent.tsx
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    render() {
+        const material = this.props.material;
+
+        return (
+            <div className="pane">
+                <CommonMaterialPropertyGridComponent globalState={this.props.globalState} lockObject={this.props.lockObject} material={material} onPropertyChangedObservable={this.props.onPropertyChangedObservable} />
+                <LineContainerComponent globalState={this.props.globalState} title="CONFIGURATION">
+                <ButtonLineComponent label="Edit Source" onClick={() => this.edit()} />
```
deltakosh, line 38: nit: tabulation

inspector/src/components/sourceEditor/shaderSourceEditor.ts
```diff
@@ -0,0 +1,139 @@
+import { ShaderMaterial } from "babylonjs/Materials/shaderMaterial";
+import { renderSSE, ShaderSourceRecompileCallback, HistoryDataSource } from './shaderSourceEditorUI';
+
+class Popup {
```
deltakosh, line 4: We should have it already as we have a popup feature for the AnimationCurve editor

inspector/src/components/sourceEditor/shaderSourceEditor.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    static render(hostElement: HTMLElement, options: IShaderSourceEditorOptions) {
+        const material = options.shaderMaterial;
+        const effect = material.getEffect();
+        const fragSource = (effect as any)?.originalFragmentSourceCode || '';
+        const vertexSource = (effect as any)?.originalVertexSourceCode || '';
+        const dataSource = new HistoryDataSource(this.keyForMaterial(material));
+        const delegate = {
+            recompile(fragSource: string, vertexSource: string, callback: ShaderSourceRecompileCallback) {
+                effect?._rebuildProgram(vertexSource, fragSource, (piplineContext) => {
```
deltakosh, line 125: We cannot use _ function

@sebavan : Can you check if there is a way to rebuild or maybe just expose that feature?

inspector/src/components/actionTabs/tabs/propertyGrids/materials/shaderMaterialPropertyGridComponent.tsx
deltakosh: buttonLineComponent should be more on the right

deltakosh:
You should also credit yourself in the what's new file in the dist/preview release folder (please mark it as major feature)

sebavan:
@atg, I am looking into the rebuild as I am not sure we could use it for shaderMaterial. using RebuildProgram is meant to change output code not shader source. So it won t handle webgl1 to webgl2 or defines and so on.

Here the functionnality is identical to spector and I am not sure it is intented. It should allow to modify the sources as they are so we can save back the original shader not the output after processing ???

atg:
@sebavan Yes that's the intention. I'm adding a `rebuild` method to Effect that does the preprocessing and then calls `_rebuildProgram`.

sebavan:
In this case, maybe the easiest and less intrusive would be to add a update method to the shaderMaterial itself to recreate the effect ?

The effect is shared with all the framework and keeping it as thin as possible is always good if possible.

deltakosh:
+1. This must be on a shaderMaterial.updateShadeCoder() function

sebavan:
This should be possible without modifying effect at all ? I would really prefer this.

deltakosh:
Agree with @sebavan: Effect should remain unchanged

On a side note, I will add API on the ShaderMaterial to serialize it to files and to the snippet server so we will be able to load materials the same way we load NME shaders

https://github.com/BabylonJS/Babylon.js/issues/8267

deltakosh:
To prepare it please add similar code: https://github.com/BabylonJS/Babylon.js/blob/master/inspector/src/components/actionTabs/tabs/propertyGrids/sprites/spriteManagerPropertyGridComponent.tsx#L71

Only difference will be that to save to the snippet Server, the payload will be on .shaderMaterial instead of .spriteManager: https://github.com/BabylonJS/Babylon.js/blob/master/inspector/src/components/actionTabs/tabs/propertyGrids/sprites/spriteManagerPropertyGridComponent.tsx#L148

deltakosh:
I added these API for you in this PR: https://github.com/BabylonJS/Babylon.js/commit/f9a5f6371a2a68f7748d1a8be06a191a4c823d95

deltakosh:
Up?

deltakosh:
up?

sebavan:
@atg any update on this one ?

deltakosh:
Last call before I close it

deltakosh:
Closing it for being idle for too long. Please feel free to reopen when you can

Issue #8149, opened 2020-05-12
Incorrect rendering using EXT_mesh_gpu_instancing


zeux:
The attached scene has a variant of SimpleMeshes from glTF-Sample-Models converted with instancing support using gltfpack.

Bear in mind that support for instancing in gltfpack is experimental, but it seems like this is hitting some bugs in Babylon.JS loader.

The two variants of the model are:
- simplemeshes_noq.glb - encoded using raw floating point attributes
- simplemeshes.glb - encoded using KHR_mesh_quantization (each instance embeds a dequant matrix with the scale 1/16383)

When I open these files in the sanbox, they both look incorrect:

- `simplemeshes_noq.glb` shows three triangles, rendering the base mesh in addition to instanced data. My understanding is that the middle triangle should not be rendered here.

![image](https://user-images.githubusercontent.com/1106629/81652098-0c859c00-93e8-11ea-9d61-02132d0f27b5.png)

- `simplemeshes.glb` shows just one triangle - it looks like it's the triangle for the base mesh. I'm not quite sure what happened to the instanced triangles here.

![image](https://user-images.githubusercontent.com/1106629/81651663-b0227c80-93e7-11ea-8dff-6af25caa13ba.png)

[instancing.zip](https://github.com/BabylonJS/Babylon.js/files/4614025/instancing.zip)

bghgary:
> simplemeshes_noq.glb shows three triangles, rendering the base mesh in addition to instanced data. My understanding is that the middle triangle should not be rendered here.

This isn't super clear in the extension spec, but I agree the source mesh shouldn't show. I've fixed this in #8156.

> simplemeshes.glb shows just one triangle - it looks like it's the triangle for the base mesh. I'm not quite sure what happened to the instanced triangles here.

I don't see anything wrong with the code here. It looks like the data in the glb may be wrong? The scale is super tiny 0.00006 and the positions are suspicious. Are the positions supposed to be normalized perhaps?

bghgary:
The PR fixes both problems now. Thanks for the report!

zeux:
> EDIT2: Never mind. This is because the camera is taking the source mesh into account. I'll see if I can fix it.

Ah, right, and there's no dequant matrix in the node so it's super big. Makes sense, thanks for the fixes!

PR #8213, opened 2020-05-18
EXT_mesh_gpu_instancing: Use thin instances when possible

Popov72:
If the node we must create instances for is an `InstancedMesh`, thin instances can't be used because the features are mutually exclusive (and `InstancedMesh` subclass `AbstractMesh` whereas thin instances pertain to `Mesh`, anyway).

That's why there are several checks `if (!(babylonMesh as Mesh).thinInstanceSetBuffer) ...` in the code, to deal with both cases.

### DIFF
--- a/loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
+++ b/loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
@@ -8,6 +8,7 @@ import { GLTFLoader, ArrayItem } from "../glTFLoader";
 import { IGLTFLoaderExtension } from "../glTFLoaderExtension";
 import { INode } from "../glTFLoaderInterfaces";
 import { AbstractMesh } from 'babylonjs/Meshes/abstractMesh';
+import { TmpVectors } from 'babylonjs/Maths/math.vector';
 
 const NAME = "EXT_mesh_gpu_instancing";
 
@@ -15,11 +16,6 @@ interface IEXTMeshGpuInstancing {
     mesh?: number;
     attributes: { [name: string]: number };
 }
-
-const T = new Vector3(0, 0, 0);
-const R = new Quaternion(0, 0, 0, 1);
-const S = new Vector3(1, 1, 1);
-const M = new Matrix();
 
 /**
  * [Proposed Specification](https://github.com/KhronosGroup/glTF/pull/1691)
@@ -124,18 +120,18 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
                     const matrices = canUseThinInstances ? new Float32Array(instanceCount * 16) : null;
 
                     if (matrices) {
-                        T.copyFromFloats(0, 0, 0);
-                        R.copyFromFloats(0, 0, 0, 1);
-                        S.copyFromFloats(1, 1, 1);
+                        TmpVectors.Vector3[0].copyFromFloats(0, 0, 0); // translation
+                        TmpVectors.Quaternion[0].copyFromFloats(0, 0, 0, 1); // rotation
+                        TmpVectors.Vector3[1].copyFromFloats(1, 1, 1); // scale
 
                         for (let i = 0; i < instanceCount; ++i) {
-                            translationBuffer && Vector3.FromArrayToRef(translationBuffer, i * 3, T);
-                            rotationBuffer && Quaternion.FromArrayToRef(rotationBuffer, i * 4, R);
-                            scaleBuffer && Vector3.FromArrayToRef(scaleBuffer, i * 3, S);
+                            translationBuffer && Vector3.FromArrayToRef(translationBuffer, i * 3, TmpVectors.Vector3[0]);
+                            rotationBuffer && Quaternion.FromArrayToRef(rotationBuffer, i * 4, TmpVectors.Quaternion[0]);
+                            scaleBuffer && Vector3.FromArrayToRef(scaleBuffer, i * 3, TmpVectors.Vector3[1]);
 
-                            Matrix.ComposeToRef(S, R, T, M);
+                            Matrix.ComposeToRef(TmpVectors.Vector3[1], TmpVectors.Quaternion[0], TmpVectors.Vector3[0], TmpVectors.Matrix[0]);
 
-                            M.copyToArray(matrices, i * 16);
+                            TmpVectors.Matrix[0].copyToArray(matrices, i * 16);
                         }
                     }
 
--- a/loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
+++ b/loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
@@ -1,13 +1,10 @@
 import { Vector3, Quaternion, Matrix } from 'babylonjs/Maths/math.vector';
-import { InstancedMesh } from 'babylonjs/Meshes/instancedMesh';
 import { Mesh } from 'babylonjs/Meshes/mesh';
 import { TransformNode } from "babylonjs/Meshes/transformNode";
-import { StringTools } from 'babylonjs/Misc/stringTools';
 import { Nullable } from "babylonjs/types";
 import { GLTFLoader, ArrayItem } from "../glTFLoader";
 import { IGLTFLoaderExtension } from "../glTFLoaderExtension";
 import { INode } from "../glTFLoaderInterfaces";
-import { AbstractMesh } from 'babylonjs/Meshes/abstractMesh';
 import { TmpVectors } from 'babylonjs/Maths/math.vector';
 
 const NAME = "EXT_mesh_gpu_instancing";
@@ -39,6 +36,9 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
     constructor(loader: GLTFLoader) {
         this._loader = loader;
         this.enabled = this._loader.isExtensionUsed(NAME);
+        if (this.enabled) {
+            loader._disableInstancedMesh = true;
+        }
     }
 
     /** @hidden */
@@ -53,21 +53,6 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
 
             if (!node._primitiveBabylonMeshes) {
                 return promise;
-            }
-
-            let useThinInstancesForAllMeshes = true;
-            let canUseThinInstances = false;
-
-            // Hide the source meshes.
-            for (const babylonMesh of node._primitiveBabylonMeshes) {
-                if (!(babylonMesh as Mesh).thinInstanceSetBuffer) {
-                    babylonMesh.isVisible = false;
-                    useThinInstancesForAllMeshes = false;
-                } else {
-                    canUseThinInstances = true;
-                    (babylonMesh as Mesh)._thinInstanceDataStorage.instancesCount = 1;  // make sure mesh.hasThinInstances returns true from now on (else async loading of the thin instance data will lead to problems
-                                                                                        // as the mesh won't be considered as having thin instances until thinInstanceSetBuffer is called)
-                }
             }
 
             const promises = new Array<Promise<Nullable<Float32Array>>>();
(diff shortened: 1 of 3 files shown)


loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
```diff
@@ -83,28 +98,60 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
             loadAttribute("SCALE");
 
             if (instanceCount == 0) {
+                for (const babylonMesh of node._primitiveBabylonMeshes) {
+                    if ((babylonMesh as any).thinInstanceSetBuffer) {
+                        (babylonMesh as Mesh)._thinInstanceDataStorage.instancesCount = 0;
```
deltakosh, line 103: why do we need to use private member?

Popov72: Because setting the `thinInstanceCount` property won't work (that's what I did first) because there's no thin instance matrix buffer yet, so you can't set a `thinInstanceCount` property greater than 0.

This is a kind of hack to make sure that `hasThinInstances` will return `true` in line #784 of **glTFLoader.ts**, because this line can be executed before we get to the point where we set the matrix buffer in EXT_mesh_gpu_instancing class. At which point `thinInstanceCount` will be set to the right value.

deltakosh: maybe we can add a prepareThinInstanceBuffer() or something

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
```diff
@@ -16,6 +16,11 @@ interface IEXTMeshGpuInstancing {
     attributes: { [name: string]: number };
 }
 
+const T = new Vector3(0, 0, 0);
+const R = new Quaternion(0, 0, 0, 1);
+const S = new Vector3(1, 1, 1);
+const M = new Matrix();
```
bghgary, line 22: Perhaps use `TmpVectors`?

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
```diff
@@ -54,9 +59,19 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
                 return promise;
             }
 
+            let useThinInstancesForAllMeshes = true;
+            let canUseThinInstances = false;
+
             // Hide the source meshes.
             for (const babylonMesh of node._primitiveBabylonMeshes) {
-                babylonMesh.isVisible = false;
+                if (!(babylonMesh as Mesh).thinInstanceSetBuffer) {
```
bghgary, line 67: Chatted with @sebavan. We shouldn't check for existence of this function since if someone is loading this glTF extension, we want to use thin instances. Import the thinInstanceMesh at the top of this file to load it and then you shouldn't have to branch for this. It should simplify all the changes in this PR quite a bit?

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
```diff
@@ -83,28 +98,60 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
...
+                            const babylonInstancedMeshes = babylonMesh.getChildMeshes(true, (node) => (node as AbstractMesh).isAnInstance);
+                            for (let i = 0; i < instanceCount; ++i) {
+                                const babylonInstancedMesh = babylonInstancedMeshes[i];
+                                translationBuffer && Vector3.FromArrayToRef(translationBuffer, i * 3, babylonInstancedMesh.position);
+                                rotationBuffer && Quaternion.FromArrayToRef(rotationBuffer, i * 4, babylonInstancedMesh.rotationQuaternion!);
+                                scaleBuffer && Vector3.FromArrayToRef(scaleBuffer, i * 3, babylonInstancedMesh.scaling);
+                                babylonInstancedMesh.refreshBoundingInfo();
+                            }
+                        } else {
+                            (babylonMesh as Mesh).refreshBoundingInfo();
```
bghgary, line 153: You should be able to rely on the core loader to do the refreshBoundingInfo of all the source meshes (since this code is a continuation of the core loader) and not have logic in the core loader switching between hasThinInstances.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -702,7 +702,11 @@ export class GLTFLoader implements IGLTFLoader {
 
         return Promise.all(promises).then(() => {
             this._forEachPrimitive(node, (babylonMesh) => {
-                babylonMesh.refreshBoundingInfo(true);
+                if ((babylonMesh as Mesh).hasThinInstances) {
+                    // refresh already done
+                } else {
+                    babylonMesh.refreshBoundingInfo(true);
+                }
```
bghgary, line 709: See comment in the extension. I don't think you should switch here.

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: So, that would be something like `thinInstancePrepareBuffer(kind: string, numInstances: number = 32, stride: number = 16)` that would allocate a matrix buffer big enough for the number of instances passed in parameter? Or should it be used only to pre-allocate matrix storage (so we could get rid of kind and stride)?

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -777,7 +781,7 @@ export class GLTFLoader implements IGLTFLoader {
 
         this.logOpen(`${context}`);
 
-        const shouldInstance = this._parent.createInstances && (node.skin == undefined && !mesh.primitives[0].targets);
+        const shouldInstance = this._parent.createInstances && (node.skin == undefined && !mesh.primitives[0].targets) && (!primitive._instanceData || !primitive._instanceData.babylonSourceMesh.hasThinInstances);
```
bghgary, line 784: I don't understand why we need additional conditions here? This check is for instanced meshes which have nothing to do with thin instances? Perhaps I'm missing something.

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: The problem comes from the fact that as an optimization, the glTF loader creates an `InstancedMesh` instead of a `Mesh` when possible (see line 784 in **glTFLoader**). But if this mesh has "gpu instances" (as per the extension), creating thin instances for it won't work because `InstancedMesh` don't support thin instances.

The easiest way to deal with that would be to disable this optimization completely when **EXT_mesh_gpu_instancing** is used, but we don't know in advance if it will be used or not...

So that would require pre-parsing the file to look for the extension usage. Or is there a faster way to know? I think a glTF file can be flagged to say "this extension XXX is used/required", maybe we can rely on that?

In that case, I can indeed remove all my ugly code :)

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: Ok

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: Ok, good to know the `refreshBoundInfo` call is always executed before running the extension code, will do the change.

loaders/src/glTF/2.0/glTFLoader.ts
Popov72: See above:
> The problem comes from the fact that as an optimization, the glTF loader creates an InstancedMesh instead of a Mesh when possible (see line 784 in glTFLoader). But if this mesh has "gpu instances" (as per the extension), creating thin instances for it won't work because InstancedMesh don't support thin instances.

In the **teapots_galore.glb** file you have two nodes, each using the "gpu instancing" extension. The two nodes point to the same mesh definition. So, when processing the second node, the code you pointed to create an `InstancedMesh` instead of a `Mesh` and so thin instances can't be used anymore for this mesh.

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: We may not need this if I can remove all my ugly code and disable `InstancedMesh` optimization when the gpu_instancing extension is used. Waiting for @bghgary answers.

loaders/src/glTF/2.0/glTFLoader.ts
Popov72: Ok, removed.

Popov72:
I performed some changes, let me know if those make sense.

Popov72:
Also of note: I create static buffers for thin instances for performance sake, not dynamic ones. I don't know if it's the right thing to do (?)

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
bghgary: I see. I suppose it is possible for a glTF to have both, but yeah, I agree if this extension exists then it is unlikely someone wanted both mesh instancing and instancing with buffers. I think ideally we should be able to disable calling createInstance if this extension is in use for this node. I just saw that you added a flag to disable creation of instances. Instead of doing that globally, maybe you can do it per mesh?

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
```diff
@@ -54,11 +55,6 @@ export class EXT_mesh_gpu_instancing implements IGLTFLoaderExtension {
                 return promise;
             }
 
-            // Hide the source meshes.
-            for (const babylonMesh of node._primitiveBabylonMeshes) {
-                babylonMesh.isVisible = false;
-            }
-
```
bghgary, line 61: This is still necessary right?

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: > I just saw that you added a flag to disable creation of instances. Instead of doing that globally, maybe you can do it per node?

How would that work?

Say I have two nodes, each pointing to the same mesh, and only the second node has the gpu instancing extension.

So, when handling the second node, an `InstancedMesh` will be created because the mesh from the first node was not flagged as "don't allow instancing". We would have to check for the gpu instancing extension right in the glTF loader, when deciding to create an `InstancedMesh` or not, but I don't think we want to do that?

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: No, we should not do it for thin instances.

By default, the mesh itself is no put the in the matrix buffer, so it won't be rendered. On the contrary, the mesh must have `visible` = `true` so that the frustum check returns the right answer.

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
bghgary: This extension is on a glTF node, which is before the mesh is loaded. I think you can change your _disableInstancedMesh flag into a ref count and increment the ref count right before calling the original loadNode: https://github.com/BabylonJS/Babylon.js/pull/8213/files#diff-3dddcf4f449da13644db39f2e82479f4R52. This should cause the core loader to not create an instanced mesh. Then decrement the ref count when the extension is done loading.

bghgary: Hmm, though this may cause an issue with child nodes that should create an instanced mesh. I wonder if the spec should allow a node with this extension to have child nodes. 🤔

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
bghgary: Okay, I added this because createDefaultCamera was taking the source mesh into account when it shouldn't. Please make sure the assets in https://github.com/BabylonJS/Babylon.js/issues/8149 still work properly in the sandbox.

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: It seems something like this work:
```typescript
public loadNodeAsync(context: string, node: INode, assign: (babylonTransformNode: TransformNode) => void): Nullable<Promise<TransformNode>> {
    return GLTFLoader.LoadExtensionAsync<IEXTMeshGpuInstancing, TransformNode>(context, node, this.name, (extensionContext, extension) => {
        this._loader._disableInstancedMesh = true;

        const promise = this._loader.loadNodeAsync(`#/nodes/${node.index}`, node, assign);

        this._loader._disableInstancedMesh = false;
```
instead of doing a global disabling.

Would that work (must say I don't understand how the ref count would make the core loader not to use instanced meshes without making changes to the loader itself)?

bghgary: Yes, this should work if there are no children. I was suggesting ref counting since there maybe a recursive load because loadNodeAsync traverses the nodes.

Popov72: Something like:
```typescript
public loadNodeAsync(context: string, node: INode, assign: (babylonTransformNode: TransformNode) => void): Nullable<Promise<TransformNode>> {
    return GLTFLoader.LoadExtensionAsync<IEXTMeshGpuInstancing, TransformNode>(context, node, this.name, (extensionContext, extension) => {
        this._loader._disableInstancedMesh++;

        const promise = this._loader.loadNodeAsync(`#/nodes/${node.index}`, node, assign);

        this._loader._disableInstancedMesh--;
```
and testing for `this._loader._disableInstancedMesh === 0` on the loader side?

Sorry for all those back and forth, but I have little clue how everything works under the hood in the gltf loader (and that does not seem easy to grasp with all the async jobs going on).

bghgary: Yup, exactly.

> Sorry for all those back and forth

No need to be. Thanks for all the work!

loaders/src/glTF/2.0/Extensions/EXT_mesh_gpu_instancing.ts
Popov72: Yes, those two assets do display correctly.

Popov72: Hum, there are a little translated on the right compared to the online sandbox...

Popov72: Ok, the shifting occurs only when not setting `scene.useRightHandedSystem = true`. Looking for why...

Popov72: Fixed. The world matrix of the mesh has to be taken into account to calculate the global bounding info.

Popov72:
The "GLTF Extension EXT_mesh_gpu_instancing" test fails, but it does work on my computer...

deltakosh:
This is what was rendered: 
image

Popov72:
Yep.

Maybe it is linked with the problem that the code is not put in the webpack packages.

Trying to push a change, but I have problems with github...

PR #8215, opened 2020-05-18
Fix SSAO 2 detection.

sebavan:


src/PostProcesses/RenderPipeline/Pipelines/ssao2RenderingPipeline.ts
```diff
@@ -155,7 +155,7 @@ export class SSAO2RenderingPipeline extends PostProcessRenderPipeline {
         if (!engine) {
             return false;
         }
-        return engine.getCaps().drawBuffersExtension;
+        return engine.webGLVersion >= 2;
```
deltakosh, line 158: This does not exist in webgl1?

sebavan: Nope the shaders are not supported and use dedicated syntax from glsl 3.0. I ll check in a bit if it is possible to fix easily.

sebavan: Basically relying on arrays initialization, modulo and others operators.

deltakosh: not a problem though. No need to fix the shaders honestly

PR #8216, opened 2020-05-19
Fix OBJ serializing mirrored

tdurand:
Hi

See post on the forum about this: https://forum.babylonjs.com/t/obj-export-mirrored/10835/10

- change serializer to OBJ to invert the X vertices by default as OBJ is right handed
- add flag to make this configurable in case the babylon scene is rightHanded

Sorry if something is wrong, first contribution to Babylon.js 😁


RaananW:
Thanks so much for the PR :-)

There are 2 linting problems (you can see them when checking the PR checklist's failed tasks): 

```
Failed to lint: /home/vsts/work/1/s/serializers/src/OBJ/objSerializer.ts [19, 146]: trailing whitespace, /home/vsts/work/1/s/serializers/src/OBJ/objSerializer.ts [73, 19]: missing whitespace.
```

serializers/src/OBJ/objSerializer.ts
```diff
@@ -16,9 +16,10 @@ export class OBJExport {
      * @param materials defines if materials should be exported
      * @param matlibname defines the name of the associated mtl file
      * @param globalposition defines if the exported positions are globals or local to the exported mesh
+     * @param sceneUsesRightHandedSystem defines if the babylon scene is right handed, in this case the exported obj X vertices won't be inverted 
      * @returns the OBJ content
      */
-    public static OBJ(mesh: Mesh[], materials?: boolean, matlibname?: string, globalposition?: boolean): string {
+    public static OBJ(mesh: Mesh[], materials?: boolean, matlibname?: string, globalposition?: boolean, sceneUsesRightHandedSystem?: boolean): string {
```
RaananW, line 22: `useRightHandedSystem` would be great. This is what we use throughout the framework

serializers/src/OBJ/objSerializer.ts
```diff
@@ -67,7 +68,13 @@ export class OBJExport {
             }
 
             for (var i = 0; i < trunkVerts.length; i += 3) {
-                output.push("v " + trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
+                // Babylon default is left handed, while OBJ default is right handed
+                // By default we serialize in OBJ inverting the X vertices
+                if(sceneUsesRightHandedSystem) {
+                    output.push("v " + trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
+                } else {
+                    output.push("v " + -trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
```
RaananW, line 76: Got two questions here:

1) You are changing the default behavior (which just serialized all of the verts). Is that the right behavior for left-handed systems? Were we wrong all this time?
2) Usually we invert the 'z' axis. I might not know 100% how OBJ works - is the axis order different?

dist/preview release/what's new.md
```diff
@@ -203,6 +203,7 @@
 - Fixed `Sound` not accepting a `TransformNode` as a source for spatial sound ([Poolminer](https://github.com/Poolminer))
 - Fixed an issue with transformation set after physics body was created using cannon.js ([#7928](https://github.com/BabylonJS/Babylon.js/issues/7928)) ([RaananW](https://github.com/RaananW))
 - Fix bug when using `ShadowOnlyMaterial` with Cascaded Shadow Map and `autoCalcDepthBounds` is `true` ([Popov72](https://github.com/Popov72))
+- Fix OBJ serializer default scene scene handedness causing [OBJ Mirror export](https://forum.babylonjs.com/t/obj-export-mirrored/10835/10)
```
RaananW, line 206: please add your name and github account link. Unless you don't want, of course :-)

serializers/src/OBJ/objSerializer.ts
tdurand: I had named it `useRightHandedSystem` firstly but then changed as I think this can cause confusion as it could mean that we want the export to use a rightHandedSystem.

It is not the case, by default we want to convert left handed to right handed and the flag means : "hey I use rightHandedSystem in my scene, please keep the export as it is rightHanded"

Not sure if it makes sense 😉

serializers/src/OBJ/objSerializer.ts
tdurand: 1. Seems so, but I'm not the best on that topic ... maybe @Drigax can jump in here to validate, which suggested this in the first place

2. Also can't answer, I just know that this changes made my scenes exported correctly @Drigax could you help here ?  https://forum.babylonjs.com/t/obj-export-mirrored/10835/10

dist/preview release/what's new.md
tdurand: I think the forum link is better for future reference to this, in the end this fix is mostly made by @Drigax and not me .. I'm just trying to help with his guidance !

serializers/src/OBJ/objSerializer.ts
RaananW: yeah, i get it. Naming is always a fun part of development.

I am fine with your current naming :-). As long as the documentation is clear, it is perfectly fine.

tdurand:
regarding the lint issue.. currently cloning Babylon.js to try to fix it.. it did the PR with Gitpod and it doesn't seem to warn me about anything

RaananW:
> regarding the lint issue.. currently cloning Babylon.js to try to fix it.. it did the PR with Gitpod and it doesn't seem to warn me about anything

Oh! you can still use it. It is just a missing whitespace and an extra whitespace. Probably an extra space somewhere.  The line numbers are written in the comment I pasted

serializers/src/OBJ/objSerializer.ts
RaananW: I am totally fine with the changes if it is also a bug fix. will wait for @Drigax for a final approval

serializers/src/OBJ/objSerializer.ts
```diff
@@ -67,7 +68,13 @@ export class OBJExport {
             }
 
             for (var i = 0; i < trunkVerts.length; i += 3) {
-                output.push("v " + trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
+                // Babylon default is left handed, while OBJ default is right handed
+                // By default we serialize in OBJ inverting the X vertices
+                if (sceneUsesRightHandedSystem) {
```
deltakosh, line 73: I would recommend just asking the scene. The scene can be retrieved from meshes[0].getScene

We do not need an extra parameter

tdurand: nice suggestion, just changed that to make everything much simpler !

serializers/src/OBJ/objSerializer.ts
Drigax: I agree with @RaananW's suggestion, since Babylon's axes are setup as +Z Forward, +X right and +Y up, we should invert the Z axis. currently, we were inverting both Z and X with your proposed fix, imported into blender with default settings.

serializers/src/OBJ/objSerializer.ts
```diff
@@ -67,7 +67,13 @@ export class OBJExport {
             }
 
             for (var i = 0; i < trunkVerts.length; i += 3) {
-                output.push("v " + trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
+                // Babylon.js default is left handed, while OBJ default is right handed
+                // Need to invert X vertices unless Babylon is set to use a right handed system
+                if (mesh[0].getScene().useRightHandedSystem) {
+                    output.push("v " + trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
+                } else {
+                    output.push("v " + -trunkVerts[i] + " " + trunkVerts[i + 1] + " " + trunkVerts[i + 2]);
```
Drigax, line 75:
```suggestion
                    output.push("v " + trunkVerts[i] + " " + trunkVerts[i + 1] + " " + -trunkVerts[i + 2]);
```

tdurand:
> Thanks for the fix, just a small change to be consistent with our axes

just applied your fix 👍

tdurand:
> The code documentation says "x" while we actually change the "z" axis, but this is nitpicking at another level. Approved.

True, just fixed !

RaananW:
you're awesome 😁

tdurand:
truth is @Drigax did all the work, just helping with the logistics 🙌 

Many thanks and keep rocking with Babylon.js !

PR #8220, opened 2020-05-19
Fix mesh instance disappearing when shadow caster

Popov72:
See https://forum.babylonjs.com/t/difference-between-instanced-non-instanced-with-alpha-materials-and-shadows/10949/5


Popov72:
2 problems in the visual tests, but I don't understand:
* first one, the texture of the plane is the "missing" texture, so it seems it could not be read?
* second one, it seems ok, the picture seems to mach the model

Both tests are working on my local computer.

deltakosh:
there is a crash with webgl2

PR #8223, opened 2020-05-19
Add loader for KHR_materials_variants extension

MiiBond:


loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,131 @@
...
+            let tag_mapping = this._tagsToMap.get(tag);
+            if (tag_mapping === undefined) {
+                tag_mapping = [];
+            }
+            const material = ArrayItem.Get(`#/materials/`, this._loader.gltf.materials, mapping.material);
+            let meshEntry: VariantMapping = {
+              mesh: babylonMesh,
+              materialPromise: Promise.resolve(null)
+            };
+            if (babylonMesh instanceof Mesh) {
```
bghgary, line 107: What happens with instancing? The loader currently uses instancing when multiple nodes point to the same mesh. Does the spec allow material variants on this?

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,131 @@
...
+          });
+        });
+      };
+      return this._loader._loadMeshPrimitiveAsync(context, name, node, mesh, primitive, assignMesh);
+    });
+
+    if (loadExtensions) {
+        return loadExtensions;
+    }
+    return this._loader._loadMeshPrimitiveAsync(context, name, node, mesh, primitive, assign);
```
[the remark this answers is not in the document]
MiiBond: That's what I originally did but I was getting a TS error because `GLTFLoader.LoadExtensionAsync`'s return value is nullable and `_loadMeshPrimitiveAsync` isn't.
```'Nullable<Promise<AbstractMesh>>' is not assignable to type 'Promise<AbstractMesh>'```

File: loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts [hunk 3]
```diff
@@ -0,0 +1,131 @@
...
+interface VariantMapping {
+  mesh: AbstractMesh;
+  materialPromise: Promise<Nullable<Material>>;
+  material?: Material;
+}
+
+/**
+ * [Specification](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_variants)
+ */
+export class KHR_materials_variants implements IGLTFLoaderExtension {
```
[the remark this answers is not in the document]
MiiBond: Done

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,131 @@
...
+   * The name of this extension.
+   */
+  public readonly name = NAME;
+
+  private _loader: GLTFLoader;
+
+  public defaultVariant: string | undefined;
+  public availableVariants: string[];
+
+  private _tagsToMap: Map<string, VariantMapping[]> = new Map();
```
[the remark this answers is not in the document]
MiiBond: Done

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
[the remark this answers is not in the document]
MiiBond: Fixed base _loadMeshPrimitiveAsync function to be nullable.

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
MiiBond: Disabled instancing for these meshes. Please check.

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,138 @@
...
+            const assignMesh = (babylonMesh: AbstractMesh) => {
+                assign(babylonMesh);
+                const babylonDrawMode = (GLTFLoader as any)._GetDrawMode(context, primitive.mode);
+                // For each mapping, look at the tags and make a new entry for them
+                extension.mapping.forEach((mapping: IKHRMaterialVariantsMapping) => {
+                    mapping.tags.forEach((tag: string, index: number) => {
+                        let tagMapping = this._tagsToMap[tag];
+                        if (tagMapping === undefined) {
+                            tagMapping = [];
+                        }
```
bghgary, line 113:
```suggestion
                        const tagMapping = this._tagsToMap[tag] || [];
```

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,138 @@
...
+                            mesh: babylonMesh,
+                            materialPromise: Promise.resolve(null)
+                        };
+                        if (babylonMesh instanceof Mesh) {
+                            const matPromise = this._loader._loadMaterialAsync(`#/materials/${mapping.material}`, material, babylonMesh, babylonDrawMode, (material) => {
+                                meshEntry.material = material;
+                            });
+                            meshEntry.materialPromise = matPromise;
+                        }
+                        tagMapping.push(meshEntry);
```
bghgary, line 125:
```suggestion
                        const meshEntry: VariantMapping = {
                            mesh: babylonMesh,
                            materialPromise: Promise.resolve(null)
                        };
                        meshEntry.materialPromise = this._loader._loadMaterialAsync(`#/materials/${mapping.material}`, material, babylonMesh!, babylonDrawMode, (material) => {
                            meshEntry.material = material;
                        });
                        tagMapping.push(meshEntry);
```

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,138 @@
...
+        if (extensions && extensions[this.name]) {
+            const extension = extensions[this.name] as IKHRMaterialVariantsTop;
+            this.defaultVariant = extension.default;
+        }
+    }
+
+    public _loadMeshPrimitiveAsync(context: string, name: string, node: INode, mesh: IMesh, primitive: IMeshPrimitive, assign: (babylonMesh: AbstractMesh) => void): Nullable<Promise<AbstractMesh>> {
+        return GLTFLoader.LoadExtensionAsync<IKHRMaterialVariants, AbstractMesh>(context, primitive, this.name, (extensionContext, extension) => {
+
+            const assignMesh = (babylonMesh: AbstractMesh) => {
```
bghgary, line 104:
```suggestion
        return GLTFLoader.LoadExtensionAsync<IKHRMaterialVariants, AbstractMesh>(context, primitive, this.name, (extensionContext, extension) => {
            const assignMesh = (babylonMesh: AbstractMesh) => {
```

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,138 @@
...
+    /** @hidden */
+    public onLoading(): void {
+        const extensions = this._loader.gltf.extensions;
+        if (extensions && extensions[this.name]) {
+            const extension = extensions[this.name] as IKHRMaterialVariantsTop;
+            this.defaultVariant = extension.default;
+        }
+    }
+
+    public _loadMeshPrimitiveAsync(context: string, name: string, node: INode, mesh: IMesh, primitive: IMeshPrimitive, assign: (babylonMesh: AbstractMesh) => void): Nullable<Promise<AbstractMesh>> {
```
bghgary, line 101:
```suggestion
    /** @hidden */
    public _loadMeshPrimitiveAsync(context: string, name: string, node: INode, mesh: IMesh, primitive: IMeshPrimitive, assign: (babylonMesh: AbstractMesh) => void): Nullable<Promise<AbstractMesh>> {
```

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,138 @@
...
+        this._loader = loader;
+        this.enabled = this._loader.isExtensionUsed(NAME);
+    }
+
+    /** @hidden */
+    public dispose() {
+        delete this._loader;
+    }
+
+    public getVariants(): string[] {
```
bghgary, line 63: need typedoc comments on all of these public functions.

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
```diff
@@ -0,0 +1,132 @@
...
+        return Object.keys(this._tagsToMap);
+    }
+
+    public selectVariant(variantName: string | string[]) {
```
edzis, line 67: This is method and data is shared and accumulated by all models loaded. That would be ok for most of the cases of 3D Commerce where typically only one model is loaded during a session. But I believe it would be worth striving for support of multiple models loaded and allow managing each separately.

At the same time I am not sure about the appropriate mechanism:

1. Can we add metadata to the root that exposes the data and methods?
2. Can we return a custom wrapper that expose that?
3. Can we get an instance of the returned AssetContainer and use a WeakMap to link the data with that?
4. Can we just store the mappings in mesh metadata and allow users to instantiate a Manager for an AssetContainer that gathers the data and exposes the functionality?
5. Something else?

I am leaning towards 4 but I wonder what others think.

bghgary: > This is method and data is shared and accumulated by all models loaded.

This isn't true. Each time a glTF is loaded, new loader extensions are created. So, this is per glTF asset. @MiiBond and I are still discussing this offline on whether this is the right way to select variants though.

loaders/src/glTF/2.0/Extensions/KHR_materials_variants.ts
MiiBond: Done

sebavan:
@MiiBond, it looks like this has some compilation issues:

image

MiiBond:
> @MiiBond, it looks like this has some compilation issues:
Yes, sorry, that was the result of a "fix" yesterday. It's corrected now.

sebavan:
IT is getting pretty close, just missing a few typedoc comments on some of the new types:

image

PR #8225, opened 2020-05-19
Add a DeviceInputSystem factory function

ryantrem:
This change replaces the `DeviceInputSystem` public constructor with a public static factory function that either creates a native `DeviceInputSystem` or a JS `DeviceInputSystem`. I tried a lot of different ways of handling this, and had hoped we could do it without making any changes to Babylon.js, but it turns out this is hard to do with ES6+modules. This is about the least intrusive change I think I can make in Babylon.js to allow us to transparently swap out a native input system with the DOM based one when running in the context of Babylon Native. I talked at length with @bghgary and @sebavan about this as well, and there are some additional ideas around this, but they are much larger changes. FYI @PolygonalSun 

For the associated Babylon Native change, see https://github.com/BabylonJS/BabylonNative/pull/268

### DIFF
--- a/src/DeviceInput/deviceInputSystem.ts
+++ b/src/DeviceInput/deviceInputSystem.ts
@@ -47,10 +47,6 @@ export class DeviceInputSystem implements IDisposable {
     private static _MAX_KEYCODES: number = 255;
     private static _MAX_POINTER_INPUTS: number = 7;
 
-    /**
-     * Default Constructor
-     * @param engine - engine to pull input element from
-     */
     private constructor(engine: Engine) {
         const inputElement = engine.getInputElement();
         if (inputElement) {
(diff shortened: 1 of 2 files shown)


src/DeviceInput/deviceInputSystem.ts
```diff
@@ -3,6 +3,8 @@ import { IDisposable } from '../scene';
 import { Nullable } from '../types';
 import { DeviceType } from './InputDevices/deviceEnums';
 
+declare const _native: any;
```
bghgary, line 6:
```suggestion
/** @hidden */
declare const _native: any;
```

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -59,6 +61,15 @@ export class DeviceInputSystem implements IDisposable {
         }
     }
 
+    public static Create(engine: Engine): DeviceInputSystem {
```
bghgary, line 64: needs typedoc comment

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -59,6 +58,19 @@ export class DeviceInputSystem implements IDisposable {
         }
     }
 
+    /**
+     * Creates a new DeviceInputSystem instance
+     * @param engine - engine to pull input element from
+     */
```
bghgary, line 64:
```suggestion
    /**
     * Creates a new DeviceInputSystem instance
     * @param engine Engine to pull input element from
     * @returns The new instance
     */
```

PR #8252, opened 2020-05-23
Fix #8250

### LINKED ISSUE #8250: Camera's view isn't persisted if rotationQuaternion is animated
# Bugs
Repro playground: https://www.babylonjs-playground.com/#CGUX9W

Once you load the above playground, the camera will animate to look at a cube in the scene. After the animation finishes, try to pan the camera. It will snap back to its initial view instead of smoothly panning near where you just animated it.

This feels like a bug. I'd expect animations to correctly persist the camera's view.

sebavan:


src/Cameras/targetCamera.ts
```diff
@@ -312,11 +312,19 @@ export class TargetCamera extends Camera {
...
         if (needToRotate) {
+            //rotate, if quaternion is set and rotation was used
+            if (this.rotationQuaternion) {
+                this.rotationQuaternion.toEulerAnglesToRef(this.rotation);
+            }
+
             this.rotation.x += this.cameraRotation.x * directionMultiplier;
             this.rotation.y += this.cameraRotation.y * directionMultiplier;
 
+            console.log(this.cameraRotation);
```
Popov72, line 323: You probably want to remove the `console.log` (2 occurrences)

sebavan: Mouahahahahah... Yup pushing in a sec, thanks a ton.

PR #8257, opened 2020-05-24
Fix nme pbr following latest changes for reflectance

Popov72:
Also add support for realtime filtering + a validation test for pbr node material


Popov72:
The new validation test fails, but I don't understand why...

The rendered picture is quite similar to the ref image, but still with 10000 different pixels.

On my local computer everything is ok, on both Chrome and Firefox.

This is the rendered image:
image

And the ref image:
node-material-pbr-1

You can see the ref image is more "crisp", notably for the plane.

sebavan:
This  is weird I ll have a look tomorrow :-(

Popov72:
The sphere on the left is rendered with the PBRMaterial and the sphere on the right with the node material: as you can see, they are identical in each picture.

So, I don't think the problem comes from the shader code itself, maybe it comes from the .dds file which would be different from the one in my local host (myne is the one from the github repository (Playground/textures), unmodified of course)?

sebavan:
I am looking into it now and I ll let you know if I can figure smthg out but I wonder if it could be related to mipmapping or smthg like this.

sebavan:
I am pretty sure anisotropie is not supported in xvfb mode :-) Could you try to use https://playground.babylonjs.com/#D8AK3Z#5 instead and regenerate the ref picture ???

sebavan:
I ll merge it as soon as done

Popov72:
Done.

What's strange is that we didn't have to do that for other validation tests that do work...

sebavan:
Yup, maybe aniso was not such a problem there ? let s see.

sebavan:
All good ;-)

Issue #6539, opened 2019-06-27
Instanced Meshes can't use the same edgeRenderer


Ian19:
If a scene contains instances of a mesh that all use an edgeRenderer then **multiple identical lineShader materials are created** for each instance. It might be efficient if mesh instances could all use one instance of lineShader or edgeRenderer.

[forum](https://forum.babylonjs.com/t/instancing-question-can-multiple-meshes-use-the-same-edgesrenderer/4153)

Maybe code like this:

box2.edgesRenderer = box1.edgesRenderer;

[Playground](https://www.babylonjs-playground.com/#TAZ2CB#51)

Thanks in advance.

sebavan:
This would only be possible for instances where only one draw call could potentially render all the edges.

Popov72:
I'm not sure it can be done for instances...

Everything instance related is done at the sub mesh level (notably the list of visible instances), whereas the edge renderer works at the mesh geometry level: it is using the vertex buffer to create the lines, irrelevant of the sub meshes structure of the mesh.

So, that would mean creating in the edge renderer a set of lines for each different sub mesh, and in the render method we should loop over the sub meshes and issue a draw call for each. The problem is that we don't know which sub meshes are visible or not in this frame. Also, we don't have the world0/.../world3 vertex buffers set correctly (the current values are the ones set to draw the last drawn sub mesh), so we have to update them before each draw call.

Should we still do it? If yes, it still remain the problem to know which sub-meshes of the mesh are visible in the current frame.

Note however that thin instances can be supported easily in the edge renderer (I have already performed the changes in my local repo).

Popov72:
Regarding the op problem:
> multiple identical lineShader materials are created for each instance

We could create the `lineShader` a single time and keep it as a static property of the class.

Popov72:
Easier to discuss with some code: #8265

sebavan:
Love the idea and understand the issue, go for it, it will already saves the state change associated to the program which costs a lot :-)

We ll close this as soon as the PR goes in. Thanks a lot.

sebavan:
Thanks to @Popov72 all done !!!

Ian19:
Thanks Popov72!

PR #8265, opened 2020-05-26
Edge Renderer: support regular and thin instances

Popov72:
Also, create a single instance of the line shader for all edge renderers.

### DIFF
--- a/src/Rendering/edgesRenderer.ts
+++ b/src/Rendering/edgesRenderer.ts
@@ -164,6 +164,9 @@ export class EdgesRenderer implements IEdgesRenderer {
     /** Gets or sets a boolean indicating if the edgesRenderer is active */
     public isEnabled = true;
 
+    /**
+     * List of instances to render in case the source mesh has instances
+     */
     public customInstances = new SmartArray<Matrix>(32);
 
     private static GetShader(scene: Scene): ShaderMaterial {
--- a/src/Rendering/edgesRenderer.ts
+++ b/src/Rendering/edgesRenderer.ts
@@ -151,21 +151,30 @@ export class EdgesRenderer implements IEdgesRenderer {
     /** Gets or sets a boolean indicating if the edgesRenderer is active */
     public isEnabled = true;
 
-    private static _LineShader: ShaderMaterial;
+    private static _LineShaders: Array<ShaderMaterial> = [];
+    private static _LineShaderScenes: Array<Scene> = [];
 
     private static GetShader(scene: Scene): ShaderMaterial {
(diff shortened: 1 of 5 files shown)


Popov72:
Regarding the line shader material, maybe we should create a material for each different scene, and use the scene as a lookup to get the right material? Else, using edge renderers in different scenes won't work with the current code change...

src/Rendering/edgesRenderer.ts
```diff
@@ -149,6 +151,32 @@ export class EdgesRenderer implements IEdgesRenderer {
     /** Gets or sets a boolean indicating if the edgesRenderer is active */
     public isEnabled = true;
 
+    private static _LineShaders: Array<ShaderMaterial> = [];
+    private static _LineShaderScenes: Array<Scene> = [];
```
sebavan, line 155: It might be simpler to keep track of it on the scene  directly ???

Popov72: Done.

sebavan:
I guess we can go out of draft :-)

src/Rendering/edgesRenderer.ts
```diff
@@ -447,7 +469,7 @@ export class EdgesRenderer implements IEdgesRenderer {
      * @return true if ready, otherwise false.
      */
     public isReady(): boolean {
-        return this._lineShader.isReady();
+        return this._lineShader.isReady(this._source, this._source.hasThinInstances);
```
deltakosh, line 472: Are we sure it is hasTHIN and not regular instances instead?

Popov72: The PR does not support regular  instances, see my comments in #6539

deltakosh: oh ok I did not read that. And how will it looks like from the user standpoint? what should I do to get instancing for my edge renderers?

Popov72: ~~You mean, "what should I do to get edge renderer for my instances"?~~

~~Nothing changed, except that a shader material is not created for each edge renderer, only a single one is created by scene.~~

Ok, I may have found a way, as I can see the current implementation for edge rendering for instances does not account for visible sub meshes, anyway.

Going to commit some changes.

Popov72: Committed.

I have added a new `edgesShareWithInstances` property on the mesh, to let the user tell if they want to use the edge renderer of the source mesh for all instances or not.

sebavan:
Loooks goood !!!!

PR #8266, opened 2020-05-26
[Enhancement] Observable-based timer

RaananW:
When implementing timed functions in XR i noticed I used this pattern quite a lot - set time-to-end, measure time using the onXRFrameObservable, and clear the observer after the timer has ended.

I then thought - this might be needed globally, and can actually be used with any observable. So - introducing the Simple and Advanced timers.

The timers count time on each call of the observable attached to them. meaning that they can also be used as a form of "debounce" for pointer events. 

The simplest example - notify me in 2 seconds:

```javascript
SetAndStartTimer({
        time: 2000,
        countingObservable: scene.onBeforeRenderObservable, 
        onEnded: () => { alert("Huzza!!!"); }
    });
```

Advantage here is that we are not using setTimeout. the onEnded function is guaranteed to be triggered inside the render loop (better yet - inside the beforeRender loop)

hold pointer down for 3 seconds:

```javascript
let blackMaterial = ..... // create a material with black diffuse color
let pointerDown = false;
scene.onPointerDown = () => {
    pointerDown = true;
    SetAndStartTimer({
        time: 3000,
        countingObservable: scene.onBeforeRenderObservable, 
        onEnded: () => { alert("Huzza!!!"); },
        breakCondition: () => !pointerDown,
        onTick: (data) => {
            // turn the material to green for visualization
            blackMaterial.diffuseColor.set(0, data.completeRate, 0);
        }
    });
}
scene.onPointerUp = () => {
    pointerDown = false;
}
```

Any observable can be used, if it makes sense. For example - draw on a mesh for 5 seconds:

```javascript
scene.onPointerDown = () => {
    pointerDown = true;
    SetAndStartTimer({
        time: 3000,
        countingObservable: scene.onPointerObservable,  // use the onPointer observable for the ticks
        observableParameters: { mask: PointerEventTypes.POINTERMOVE } // filter to pointer move only
        onEnded: () => { commitPainting() },
        onTick: (data) => {
            if(data.payload.pickInfo.pickedPoint) { /* paint on the mesh with the pick info */ }
        }
    });
}
```

The advanced timer has a lot more options (like the options to manually stop and restart the timer), a state machine and a lot of available observables, when needed. But essentially, it works the same.

Disadvantage is that it is never 100% exact. Even when using the beforeRenderObservable, there will always be a few extra milliseconds on the timer.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -27,6 +27,7 @@
 - Added support in `ShadowGenerator` for fast fake soft transparent shadows ([Popov72](https://github.com/Popov72))
 - Added `boundingBoxRenderer.onBeforeBoxRenderingObservable` and `boundingBoxRenderer.onAfterBoxRenderingObservable` ([Deltakosh](https://github.com/deltakosh))
 - Added initial code for user facing DeviceSourceManager ([PolygonalSun](https://github.com/PolygonalSun))
+- Added a Simple and advanced timer, based on observables ([RaananW](https://github.com/RaananW))
 
 ### Engine
 
--- a/src/Misc/timer.ts
+++ b/src/Misc/timer.ts
@@ -69,17 +69,26 @@ export interface ITimerData<T> {
  * The current state of the timer
  */
 export enum TimerState {
+    /**
+     * Timer initialized, not yet started
+     */
     INIT,
+    /**
+     * Timer started and counting
+     */
     STARTED,
+    /**
+     * Timer ended (whether aborted or time reached)
+     */
     ENDED
 }
 
 /**
  * A simple version of the timer. Will take options and start the timer immediately after calling it
  *
- * @param options options to initialize this timer
+ * @param options options with which to initialize this timer
  */
-export const SetAndStartTimer = (options: ITimerOptions<any>): Nullable<Observer<any>> => {
+export function SetAndStartTimer(options: ITimerOptions<any>): Nullable<Observer<any>> {
     let timer = 0;
     const startTime = Date.now();
     options.observableParameters = options.observableParameters ?? {};
@@ -103,7 +112,7 @@ export const SetAndStartTimer = (options: ITimerOptions<any>): Nullable<Observer
         }
     }, options.observableParameters.mask, options.observableParameters.insertFirst, options.observableParameters.scope);
     return observer;
-};
+}
 
 /**
  * An advanced implementation of a timer class
--- a/src/Misc/timer.ts
+++ b/src/Misc/timer.ts
@@ -9,13 +9,13 @@ export interface ITimerOptions<T> {
     /**
      * Time-to-end
      */
-    time: number;
+    timeout: number;
     /**
-     * What is the counting observable. Will usually be OnBeforeRenderObservable.
-     * Time calculation is done ONLY when the observable is notifying, meaning that if
+     * The context observable is used to calculate time deltas and provides the context of the timer's callbacks. Will usually be OnBeforeRenderObservable.
+     * Countdown calculation is done ONLY when the observable is notifying its observers, meaning that if
      * you choose an observable that doesn't trigger too often, the wait time might extend further than the requested max time
      */
-    countingObservable: Observable<T>;
+    contextObservable: Observable<T>;
     /**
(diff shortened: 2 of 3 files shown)


Popov72:
Nice, could come in handy!

As it is essentially a count down (from `time` to 0), maybe something like `setCountdown` would be more appropriate as a name (timer is a very generic name)?

Also, I would name `countingObservable` simply `observable` (or `context` or `observableContext`), as it does not only serve to count, one can also execute some code at each tick.

Lastly, I would move:
```typescript
const data: ITimerData<any> = {
    startTime,
    currentTime: now,
    deltaTime: timer,
    completeRate: timer / options.time,
    payload
};
```
outside the function, to avoid creating a new object at each tick:
```typescript
const data: ITimerData<any> = {
    startTime,
    currentTime: startTime,
    deltaTime: 0,
    completeRate: 0,
    payload: undefined,
};
const observer = options.countingObservable.add((payload: any) => {
    const now = Date.now();
    timer = now - startTime;
    data.currentTime = now;
    data.deltaTime = timer;
    data.completeRate = timer / options.time;
    data.payload = payload;
    [...]
```

RaananW:
> As it is essentially a count down (from `time` to 0), maybe something like `setCountdown` would be more appropriate as a name (timer is a very generic name)?

Thanks :-) Sounds good!

Same goes to the observable. Will rename it. 



> Lastly, I would move:
> 
> ```ts
> const data: ITimerData<any> = {
>     startTime,
>     currentTime: now,
>     deltaTime: timer,
>     completeRate: timer / options.time,
>     payload
> };
> ```
> 
> outside the function, to avoid creating a new object at each tick:

I thought of that. My only issue with this is that the last frame's data object will be overwritten, as it is essentially the same object. As a user I would expect the payload of last frame to still stay the same in the next frame, so I can compare. Just like event payloads in (native) callbacks.

Popov72:
> I thought of that. My only issue with this is that the last frame's data object will be overwritten, as it is essentially the same object. As a user I would expect the payload of last frame to still stay the same in the next frame, so I can compare. Just like event payloads in (native) callbacks.

Note sure I understand this... Do you have an example of what we could achieve/not achieve with the current code / modified code?

deltakosh:
Well this is a rad idea! I like it a lot

Do you mind updating the Observable documentation?

RaananW:
> Note sure I understand this... Do you have an example of what we could achieve/not achieve with the current code / modified code?

if you store the last data object, it will be changed as well:

```javascript
let lastDataSent;
SetAndStartTimer({
        time: 2000,
        countingObservable: scene.onBeforeRenderObservable, 
        onEnded: () => { alert("Huzza!!!"); }
        onTick: (data) => {
            if(lastDataSent) { 
                // will always be 0 if we reuse the object
                console.log(lastDataSent.deltaTime - data.deltaTime) 
            }
            lastDataSent = data;
        }
    });
```

This is of course a naive example, as there is no reason to do that, but if we want to reuse the payload in some way and we don't store it directly, it will change in the next tick. Not sure how often this will be used, but I am trying to prevent the confusion of storing an object and having it change as a side effect.

Popov72:
> This is of course a naive example, as there is no reason to do that, but if we want to reuse the payload in some way and we don't store it directly, it will change in the next tick. 

Indeed, but if the user wants to keep track of some data from the payload they can do it on their side (in your example, either by cloning the whole data object or only saving the last deltaTime), whereas implementing the feature with a new object creation at each tick can't be modified.

Note, you could allow both, by passing a flag at creation time (something like `dontCreateNewPayloadEachTick`).

RaananW:
I am all for object (re)usability, und totally understand the need for it. I think however that in this case it will only add unneeded overhead. Just like onPointerObservable that creates a new pointer info object (which is the expected behavior)

Popov72:
> Just like onPointerObservable that creates a new pointer info object (which is the expected behavior)

Ok, did not know that (I never had a look at the event system implemented in Babylon).

One of the difference, however, is the frequency of calls for each case: generally 60 times per second if you use `onBeforeRenderObservable` (or the like) against when an event occur (generally much less often - apart maybe for `onMouseMove` but I'm not sure it is used often (?)).

Anyway, the core team is fine with it, so let's close it :)

PR #8270, opened 2020-05-26
Add Support for importing and export frames

belfortk:
Add support to export a frame and import it into another Node Material.
This is done by serializing current Node Material, adding the frame and
blocks data from the frame source to the serialization object, then
deserializing. Basically follows the same pattern as loading a new Node
Material.


belfortk:
This PR accomplishes the goal of importing and exporting frames but I don't think it does it gracefully. When importing, I am basically deserializing current material and then re-loading a new Node Material with the frame data grafted on. This approach makes it difficult to add nice quality of life features like making the frames collapsed or auto-selected on import. I also worry about this solution scaling as a user adds more and more frames or works with increasingly complex materials.

I think it would be worth it to take a timeboxed day or 2 to investigate a way to add a frame to the Material without reloading the whole thing.

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
```diff
@@ -146,6 +146,11 @@ export class NodeMaterialConnectionPoint {
      */
     public isOptional: boolean;
 
+    /**
+     * Gets or sets a boolean indicating that this connection point is exposed on a
```
deltakosh, line 150: Missing word in the comments

belfortk: Updated! thanks for catching that

nodeEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -163,6 +170,40 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
         }, undefined, true);
     }
 
+    
+    loadFrame(file: File) {
+        Tools.ReadFile(file, (data) => {
+            // get Frame Data from file
+            let decoder = new TextDecoder("utf-8");
+            const frameData = JSON.parse(decoder.decode(data));
```
sebavan, line 178: Should not serialize deserialize but rather append to the current Node Material and editor data.

dist/preview release/babylon.module.d.ts
```diff
@@ -307,23 +307,23 @@ declare module "babylonjs/Misc/logger" {
         /**
```
sebavan, line 307: Should not be part of The PR.

sebavan:
Looks way better, let s sync offline today on the id subject ;-) and this should be all good

src/Materials/Node/nodeMaterial.ts
```diff
@@ -1606,13 +1606,52 @@ export class NodeMaterial extends PushMaterial {
      * Clear the current graph and load a new one from a serialization object
      * @param source defines the JSON representation of the material
      * @param rootUrl defines the root URL to use to load textures and relative dependencies
+     * @param hasFrames defines whether or not source is Frame or standard NodeMaterial
      */
-    public loadFromSerialization(source: any, rootUrl: string = "") {
+    public loadFromSerialization(source: any, rootUrl: string = "", hasFrames: boolean = false) {
```
deltakosh, line 1611: This is where I'm a bit stressed.
THe NodeMaterial should NOT know about frames

All the editor and UI should not live here..

The editorData field was meant to store data that the NodeEDitor should not care about.

src/Materials/Node/nodeMaterialBlockConnectionPoint.ts
```diff
@@ -146,6 +146,11 @@ export class NodeMaterialConnectionPoint {
      */
     public isOptional: boolean;
 
+    /**
+     * Gets or sets a boolean indicating that this connection point is exposed on a frame
+     */
+    public isExposedOnFrame: boolean;
```
deltakosh, line 152: This is the tiny exception we did to simplify the notion of exposing a node to a frame. But all others changes in NodeMaterial that are directly linked to the editor should be challenged

deltakosh:
Let's do it in multiple steps
- First @belfortk can you do a PR with everything linked to isExposedOnFrame only? This should be a tiny one that we can easily test
- Rework the exportation to make sure we only export the selected frame (I feel like if there is more than one frame they are all exported but with 0 node inside). There are also information about blocks which do not exist in the json file (eg. all the editor data is parsed)
This is a file I exported from a frame with 2 nodes:
image
- Once the exportation is solved. We can focus on loading it. First let's do a PR that will load the frame without the links. Just the frame and the nodes inside
- Then we will work on adding the links back from the file

PR #8272, opened 2020-05-26
First support of KHR_materials_transmission

MiiBond:


src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
```diff
@@ -270,6 +270,11 @@ struct subSurfaceOutParams
             refractionTransmittance *= cocaLambert(volumeAlbedo, vThicknessParam.y);
         #endif
 
+        #ifdef SS_ALBEDOFORREFRACTIONTINT
```
sebavan, line 273: You still have this part running:

// Compute tint from min distance only.
            vec3 volumeAlbedo = computeColorAtDistanceInMedia(vTintColor.rgb, vTintColor.w);
            refractionTransmittance *= cocaLambert(volumeAlbedo, vThicknessParam.y);
        
So volume albedo is physically computed but then override just for one part so I guess you want a #elif from the previous cases ???

sebavan: Also environmentRefraction is multiplied by refractionTransmittance a bit later which would then be equivalent to multiply twice by the albedo or its derived color ?

MiiBond: Being outside the `#elif` is intensional. This extension affects the surface only so it's possible to still have a different interior colour. e.g. volumeAlbedo could be green and the surfaceAlbedo could be red but they should both be applied to transmitted light.

MiiBond: I modified `environmentRefraction` instead of `refractionTransmittance` only because that's what the `SS_LINKREFRACTIONTOTRANSPARENCY` block did.

sebavan: Ok cool so last question is should it be before or after:
// Decrease Albedo Contribution
        outParams.surfaceAlbedo = surfaceAlbedo * (1. - refractionIntensity);


Sorry for all the questions :-)

MiiBond: I think it should be before, like the volume absorption logic. It would essentially look the same as a constant thickness volume absorption.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -0,0 +1,95 @@
...
+        if (!(babylonMaterial instanceof PBRMaterial)) {
+            throw new Error(`${context}: Material type not supported`);
+        }
+        let pbrMaterial = babylonMaterial as PBRMaterial;
+
+        // Enables "refraction" texture which represents transmitted light.
+        pbrMaterial.subSurface.isRefractionEnabled = true;
+
+        // Since this extension models thin-surface transmission only, we must make IOR = 1.0
+        pbrMaterial.subSurface.indexOfRefraction = 1.0;
```
sebavan, line 72: Another quick question. This is impacting F0 as IOR is used to compute the default F0. Is that OK ?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Defines whether this extension is enabled.
+     */
+    public enabled: boolean;
+
+    /**
+     * Defines a number that determines the order the extensions are applied.
+     */
+    public order = 190;
```
sebavan, line 34: order should probably be below 180 or this would cancel the setup of KHR_materials_ior done with order 180

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: This extension should not affect F0 for reflections. The purpose of changing the subsurface IOR was only to remove refraction.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: Very, very good point :)

MiiBond: Done. I made it 175. How's that?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
sebavan: So you might need a special case for this here: https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/PBR/pbrBaseMaterial.ts#L1857

We are trying for consistency to keep all in sync so I guess you will need to by pass our shared value here.

sebavan: looks all good to me now :-)

sebavan:
Thanks @MiiBond one more lint issue and you ll be good to go !!!

image

deltakosh:
Thanks!!

Issue #5630, opened 2018-12-11
Multi-line support for InputBox


hariettem:
Hitting enter (or shift-enter?) can add a new line to the text field.

aWeirdo:
Perhaps a prototype instead ? 
allows for easy access (combined with keyboard observerable) to define your own way(s) of creating new lines
e.g. input.addLine() 🤔

thomlucc:
Closing as no activity. Feel free to re-open.

gabrieljbaker:
Just noting @thomlucc  that there are still people out there who think this would be really cool. :)

thomlucc:
All-right, re-opening the issue in this case.

thomlucc:
#11710

thomlucc:
Clsoing as it is fixed in #11710

PR #8275, opened 2020-05-27
Added prototype of InputTextArea

fneitzel:
As per request I send this proposal of an InputTextArea. So one can have a multiple line InputTextField. Here is a link to the forum discussion: https://forum.babylonjs.com/t/gui-textinput-with-multiple-lines/9927

The are a quite a few things left and since I got no time in the moment, maybe others can push this forward.

What's missing:

    include margin
    text highlighting (copy/paste/delete)
    backspace deletion

I don't know how to reolve this issue:
```
 [at-loader] ../../gui/src/2D/controls/inputTextArea.ts:906:55 
    TS2339: Property 'text' does not exist on type 'object'.
```


deltakosh:
Hello and thanks a ton! Before going in depth I think we should consider having the textarea inheriting from textinput as they share a lot of code (to avoid duplications and maintenance hurdles)

sebavan:
Related to #5630

sebavan:
It is getting there nicely, text highlighting and the rest would be kickass :-)

sebavan:
@fneitzel do you think you will address the PR ?

deltakosh:
Closing for being stalled for too long

Please feel free to reopen when possible

PR #8278, opened 2020-05-27
Animation list component

toledoal:
Hello! 

This PR has several fixes and improvements:

1. Linear Interpolation toggle button.
2. Play, play backward, and pause animation within the Timeline.
3. Dragging/updating keyframes in the Timeline.
4. Delete an animation and messages for errors
5. Tangent Broken mode toggle button


inspector/src/components/actionTabs/tabs/propertyGrids/animations/svgDraggableArea.tsx
```diff
@@ -71,7 +71,7 @@ export class SvgDraggableArea extends React.Component<ISvgDraggableAreaProps>{
             if (coord !== undefined) {
 
                 var newPoints = [...this.props.keyframeSvgPoints];
-
+                // Check for NaN values here.
```
RaananW, line 74: Is that a TODO?

toledoal: Yes 😄 but I haven't been able to work on a bug that I'm guessing happens here. And also I don't have the exact steps to reproduce. So I added the comment here to work on later.

PR #8279, opened 2020-05-27
Vector Math MoveTowards Support

MackeyK24:
Added vector math MoveTowards and MoveTowardsToRef functions

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -510,6 +510,44 @@ export class Vector2 {
     }
 
     /**
+    * Moves a value current towards target.
+    *
+    * This is essentially the same as Vector2.Lerp but instead the function will ensure that the speed never exceeds maxDelta.
+    * Negative values of maxDelta pushes the value away from target.
+    * @param current current value
+    * @param target target value
+    * @param maxDelta max distance to move
+    * @returns resulting value
+    */
+    public static MoveTowards(current: Vector2, target: Vector2, maxDistanceDelta: number): Vector2 {
+        var result: Vector2 = new Vector2(0.0, 0.0);
+        Vector2.MoveTowardsToRef(current, target, maxDistanceDelta, result);
+        return result;
+    }
+
+    /**
+    * Moves a value current towards target.
+    *
+    * This is essentially the same as Vector2.Lerp but instead the function will ensure that the speed never exceeds maxDelta.
+    * Negative values of maxDelta pushes the value away from target.
+    * @param current current value
+    * @param target target value
+    * @param maxDelta max distance to move
+    * @param result vector to set resulting value
+    */
+   public static MoveTowardsToRef(current: Vector2, target: Vector2, maxDistanceDelta: number, result: Vector2): void {
+        var toVector_x: number = target.x - current.x;
+        var toVector_y: number = target.y - current.y;
+        var sqDist: number = toVector_x * toVector_x + toVector_y * toVector_y;
+        if (sqDist == 0 || (maxDistanceDelta >= 0 && sqDist <= maxDistanceDelta * maxDistanceDelta)) {
+            result.set(target.x, target.y);
+        } else {
+            var dist: number = Math.sqrt(sqDist);
+            result.set((current.x + toVector_x / dist * maxDistanceDelta), (current.y + toVector_y / dist * maxDistanceDelta));
+        }
+    }
+
+    /**
      * Returns a new Vector2 located for "amount" (float) on the linear interpolation between the vector "start" adn the vector "end".
      * @param start defines the start vector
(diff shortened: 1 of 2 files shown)


src/Maths/math.vector.ts
```diff
@@ -2237,6 +2314,46 @@ export class Vector4 {
...
+    * Moves a value current towards target.
+    *
+    * This is essentially the same as Vector4.Lerp but instead the function will ensure that the speed never exceeds maxDelta.
+    * Negative values of maxDelta pushes the value away from target.
+    * @param current current value
+    * @param target target value
+    * @param maxDelta max distance to move
+    * @param result vector to set resulting value
+    */
+    public static MoveTowardsToRef(current: Vector4, target: Vector4, maxDelta: number, result: Vector4): void {
```
deltakosh, line 2343: Well do we want that inside the framework? It seems really niche and Math is imported everywhere so I'm very conservative with its size

cc @sebavan  for a second thought

MackeyK24: I come across this function alot when porting Unity style game mechanics. Just thought it would be helper for general vector math mechanics.

Hersir88: Maybe could be separate module like movement ?

src/Maths/math.vector.ts
```diff
@@ -509,6 +509,44 @@ export class Vector2 {
         return new Vector2(x, y);
     }
 
+    /**
+    * Moves a value current towards target.
+    *
+    * This is essentially the same as Vector2.Lerp but instead the function will ensure that the speed never exceeds maxDelta.
```
RaananW, line 515: I believe the documentation is confusing - maxDistance is not speed, it is, as the name suggests, the maximum distance the result will be from the origin

src/Maths/math.vector.ts
```diff
@@ -509,6 +509,44 @@ export class Vector2 {
...
+    public static MoveTowards(current: Vector2, target: Vector2, maxDistanceDelta: number): Vector2 {
+        var result: Vector2 = new Vector2(0.0, 0.0);
+        Vector2.MoveTowardsToRef(current, target, maxDistanceDelta, result);
+        return result;
+    }
+
+    /**
+    * Moves a value current towards target.
+    *
+    * This is essentially the same as Vector2.Lerp but instead the function will ensure that the speed never exceeds maxDelta.
```
RaananW, line 531: If it is essentially Lerp, why not extend Lerp with maxDistance?

MackeyK24:
MoveTowards is basically a port of Vector.MoveTowards from Unity... I basically did a verbatim port from unity vector classes

MackeyK24:
If we don’t want MoveTowards in our Babylon vector classes, that’s kool too... I can always add them in my toolkit utilities. I just come across a lot using different unity mechanics for third person movement. Thought others might like access to such vector movement function

MackeyK24:
Also the maxDelta distance  description is a copy of Scaler.MoveTowards param description

RaananW:
> Also the maxDelta distance description is a copy of Scaler.MoveTowards param description

This is the Unity documentation - https://docs.unity3d.com/ScriptReference/Vector3.MoveTowards.html . Speed is being used in the examples to calculate the step, but the variable itself is not speed

> If we don’t want MoveTowards in our Babylon vector classes, that’s kool too...

Totally not what we mean. I hope it didn't come out this way. What we are saying is this - vector.math is imported everywhere, let's keep it as small as possible. 

My general question was this - why not expand the lerp function with maxDistance? Defining maxDistance as MAX_NUMBER if not provided? 

Personally I don't think MoveTowards is a perfect name for this function. Move indicates an action on an object. A mesh, a node. This is the mathematical computation of a future movement. But naming is always personal

MackeyK24:
I definitely see your point. And keeping it small is a priority. I just copied MoveTowards as is from unity so it can be used the exact same way you would unity... so you could literally follow one the unity third person controller examples... verbatim.

With that being said, I’m game for renaming , or moving somewhere that is not everywhere.

Sorry if it sounds like I’m bitching bout it. Not my intent 😊

RaananW:
> I just copied MoveTowards as is from unity

I am not sure their open source license allows you to do that.

MackeyK24:
Well ported... created Babylon js code using Babylon js classes that do the same thing

RaananW:
Just went over the code again - what is the difference between this and 

```javascript
const distance = Vector3.Distance(start, end);
Vector3.Lerp(start, end, Math.min(distance, maxDistance));
```

MackeyK24:
So does that mean no go on this pull request ?

deltakosh:
Oh sorry I thought we closed on that :) (check last answer from Raanan)

Issue #5967, opened 2019-02-25
Add support for Secondary Normal Map / Detail Normal Map


GreyWorks:
Feature request as requested in the forum discussion.

- Link to [forum discussion](https://forum.babylonjs.com/t/secondary-maps-detail-maps-detail-normal-possible/1449) about the feature
- Feature description
Add support for a secondary normal map with possibly different scaling which gets blended with the main normal map.
This is a feature in popular game engines such as Unity or Unreal.

Quote from the Unity documentation
> Secondary Maps (or Detail maps) allow you to overlay a second set of textures on top of the main textures listed above. You can apply a second Albedo colour map, and a second Normal map. Typically, these would be mapped on a much smaller scale repeated many times across the object’s surface, compared with the main Albedo and Detail maps.
>
>The reason for this is to allow the material to have sharp detail when viewed up close, while also having a normal level of detail when viewed from further away, without having to use a single extremely high texture map to achieve both goals.
>
>Typical uses for detail textures would be: - Adding skin detail, such as pores and hairs, to a character’s skin - Adding tiny cracks and lichen growth to a brick wall - adding small scratches and scuffs to a large metal container

- Additional links that could help implementing the feature:
[Unity documentation for the feature](https://docs.unity3d.com/Manual/StandardShaderMaterialParameterDetail.html)
[Unreal documentation for the feature](https://docs.unrealengine.com/en-us/Engine/Renderin/Materials/HowTo/DetailTexturing)
[Article about normal blending](https://blog.selfshadow.com/publications/blending-in-detail/)

As mentioned by PatrickRyan in the forum discussion this might be a feature for 4.1.
Thank you for looking into it.

GreyWorks:
I finally came up with a method to do it!
https://github.com/GreyWorks/Babylonjs-tests/tree/custom-pbr-material

I also made a post in the original thread about it
https://forum.babylonjs.com/t/secondary-maps-detail-maps-detail-normal-possible/1449/9

Sadly I don't have the time to fully flesh it out and make a pull request out of it.

deltakosh:
Well we should be able to do it for 4.2 :) (or 4.1 if we have time cc @sebavan)

Pryme8:
This sounds interesting, is anyone on it from @GreyWorks example?

GreyWorks:
I still work with the code mentioned above.
[...]

PR #8284, opened 2020-05-30
Add detail map support to the standard and PBR materials

Popov72:
My doc sources:
* Unity: https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@9.0/manual/Mask-Map-and-Detail-Map.html and https://docs.unity3d.com/Manual/StandardShaderMaterialParameterDetail.html
* Unreal: https://docs.unrealengine.com/en-US/Engine/Rendering/Materials/HowTo/DetailTexturing/index.html
* Cryengine: https://docs.cryengine.com/display/SDKDOC2/Detail+Maps

As Unity and Cryengine both define a detail map as being a composite of a normal + diffuse + gloss channels, I used that too. I picked the Unity mapping of the channels:
* r: grey scale diffuse
* g: normal y component
* b: roughness channel (instead of Unity's glossiness) => used only for PBR material
* a: normal x component

The normal z component is rebuilt in the shader.

To blend the bump + detail normals, I used 3 different methods (linear, whiteout, RNM), explained here: https://blog.selfshadow.com/publications/blending-in-detail/ (`detailNormalBlendMethod` property in the standard/PBR materials).

We don't really see the difference between whiteout and RNM in my sample, but it seems RNM is really better, and as whiteout and RNM are both supported by Unity, I think we should support them too. However, should we keep the linear method?

Sample: http://localhost:1338/Playground/index-local.html#5NS7A2

Without detail mapping:
image

With detail mapping:
image

The small black dots you can see in the 2nd picture come from the diffuse channel of the detail map (as this channel does not wrap, we can see some cuts in the picture).

We can make the detail diffuse channel more or less visible with the `detailDiffuseBlendLevel` property:
* 0: the channel is not blended with the regular diffuse texture at all
* 1: the channel is fully blended with the regular diffuse texture
* the formula is: `baseColor = baseColor * (saturate(vec3(detailDiffuseChannel) + 1 - detailDiffuseBlendLevel));`

The strength of the detail bump channel is controled by the `detailBumpLevel` property: 0=no bump from the detail map, 1=full bump from the detail map.

For PBR materials, the `detailRoughnessBlendLevel` property allows to blend the detail roughness channel with the roughness of the material. However, I did not really know how to blend... I ended up using: `metallicRoughness.g = saturate(metallicRoughness.g + detailColor.b * vDetailInfos.w);`, `vDetailInfos.w` being the value of the `detailRoughnessBlendLevel` property. Multiplying the two roughnesses seem wrong to me so I have added them instead... Not sure if a lerp is adequate either...


sebavan:
Couple questions :-)

What does substance outputs for the detail maps if anything ? @PatrickRyanMS  would you mind having a look cause we need to map to the output of at least one of the authoring tools to simplify the workflow.

About linear if not widely supported I am in favor of dropping it to simplify the integration ?

Also is there a way to disable the effect when seen from far away to save both perf and prevent aliasing ? Kind of blend level being a function of camera distance from mesh ? and disable if 0 or smthg of the sort ?

Other than that it looks pretty good :-) Nailed it again @Popov72

Popov72:
> About linear if not widely supported?

No idea about that, I could not find a lot of information about detail map implementation.

I kept it only in case it would be useful for low perf devices (as mobiles), even if the saving is not huge:
* linear: 1 add and 1 mul
* whiteout: 1 add and 2 mul

So only one mul saved... In the end, there's also a `normalize` in both cases, which probably accounts for nearly all the computation time anyway...

> Also is there a way to disable the effect when seen from far away to save both perf and prevent aliasing ? Kind of blend level being a function of camera distance from mesh ? and disable if 0 or smthg of the sort ?

What I can do is adding a `disableDetailMap` flag on the material to allow switching off the effect without having to remove the texture itself. The user can use the `detailDiffuseBlendLevel `, `detailBumpLevel` and `detailRoughnessBlendLevel ` properties to adjust the effect based on custom logic.

Having an automatic way to do it would probably be a nice feature to have, but we would need to come up with a way to do it properly: based only on mesh distance? Or should we take into account the texture size / mesh size too? What formula to use?

Also, I wonder if it is the responsibility of the engine to handle that? For eg, we could imagine this lod mechanism would also be useful for regular bump mapping, but the engine currently does not provide it. However, we do have a `disableBumpMap` property (at least for the pbr material), so I'm going to add a `disableDetailMap` as well.

Popov72:
Maybe I can do as for the PBR material (see `PBRClearCoatConfiguration`, `PBRSheenConfiguration` etc classes), a separate `DetailMap` class that handles everything so that the standard / PBR materials only have to call methods from this class?

sebavan:
> No idea about that, I could not find a lot of information about detail map implementation.

In this case please, remove linear it is not worth saving one Mul :-)

> What I can do is adding a disableDetailMap flag on the material to allow switching off the effect without having to remove the texture itself. The user can use the detailDiffuseBlendLevel , detailBumpLevel and detailRoughnessBlendLevel properties to adjust the effect based on custom logic.

> Having an automatic way to do it would probably be a nice feature to have, but we would need to come up with a way to do it properly: based only on mesh distance? Or should we take into account the texture size / mesh size too? What formula to use?

> Also, I wonder if it is the responsibility of the engine to handle that? For eg, we could imagine this lod mechanism would also be useful for regular bump mapping, but the engine currently does not provide it. However, we do have a disableBumpMap property (at least for the pbr material), so I'm going to add a disableDetailMap as well.

I am thinking distance changing the intensity of the effect might be a good start but maybe in a first step we simply document it with a playground sample and wait to see if use often before introducing more complexity.

I am all in for the separate config file.

Popov72:
Regarding fading out the detail map, Unity is using a setting called "Fade Out Mipmaps" (https://docs.unity3d.com/540/Documentation/Manual/class-TextureImporter.html):

`
Enable this to make the mipmaps fade to gray as the mip levels progress. This is used for detail maps. The left most scroll is the first mip level to begin fading out at. The rightmost scroll defines the mip level where the texture is completely grayed out
`

image

They fade to gray (0.5,0.5,0.5,0.5) because, with those values, the detail albedo/normal/roughness channels don't contribute anymore thanks to the formula used to blend the detail map channels with the other maps:
* albedo. The blending is done as `saturate(albedoColor * detailAlbedo * 2)`, so when `detailAlbedo = 0.5` we get the `albedoColor` untouched.
* normal. When the x and y values of the detail normal are equal to `0.5`, the unpacked values are `2 * 0.5 - 1 = 0`, meaning the reconstructed value for z is `sqrt(1 - 0^2 - 0^2) = 1`, which means an unperturbed (0,0,1) normal.
* roughness. I don't know how the blending is done, but we can probably use the same blending than for albedo: `saturate(roughness * detailRoughness * 2)`

Maybe we should do the same thing?

sebavan:
Looks kind of perfect :-)

PatrickRyanMS:
From a creation standpoint, detail maps are not a mainstream output because there is a discrepancy in the engines. Unity and Cryengine are using a channel packed texture, but Unreal uses independent textures. From that standpoint, you would author your diffuse, roughness, and normal maps as you would typically do so if going to Unreal. For the others, you need to set up the export path manually. Even in the Unity doc, they mention manually combining your texture channels in Photoshop. Substance Painter has one preset for the Unity HDRP, but it is only the detail mask in the in-box export settings. The detail mask output is highlighted below, and you can see on the right all of the options that can be put into the output maps, 

image

I had to create a custom output pack to match for the detail map requirements of the Unity HDRP detail map:

image

For Designer, there is no concept of the Unity HDRP maps when you set up a new project. They keep things higher level like PBR-MR or PBR-SG. With outputs, they are just terminal nodes that you can set up however you like. There are some presets based on which template you start with, but they are not necessary and you can delete all of them and put in anything you like as I did below. This is a different material I was working on but it is also a channel pack, and you just need to drop an RGBA Merge node in and manually wire up your textures. 

image

The trickiest part of all of it will be remembering how to pack your texture. I have set up custom ORM packs for glTF in Painter which are similar to the detail mask texture for Unity HDRP, but the packing is in a different order. If I need to pack in Designer for ORM, I need to do the same thing as above.

Even if you only had GIMP available, you should be able to push specific layers into specific channels, but doing this in any image editor is a very tedious and manual process. For example, if I have a multi-layer document in Photoshop, I can open the blending options for each layer and check the boxes for the channels that layer will render in like the image below. However, you have to manually open the options for every layer and set them individually, clicking off two checkboxes for each layer. If you have your texture built in multiple layers... let's say I have multiple bakes of my AO map to prevent penetration bake errors and I assemble them in layers in photoshop (since I have to merge them anyway) I would need to go into each one and set the channels manually. 

image

And for a normal map, I would need to pull the individual channels into different layers in the first place to be able to assign the R channel to Alpha by making it a mask on a group for the whole document. This all gets very messy in an image editor.  The process would be copying the red channel of the normal layers as displayed into an alpha channel (which isn't really an alpha channel in Photoshop as saving as a PNG ignores alpha channels because they were originally just used as masks) then I have to make a selection using the alpha channel I just created and then group all layers in my document so I can apply a layer mask on that group. This will make the document partially transparent since I am now masking everything with one channel of the normal. But when I save, I now get an alpha channel with the data from the normal R channel in it. TEDIOUS!

Not sure if this answers your question, but I also run into the same issues when I want to create a sheen, anisotropy, or clear coat texture. Since the shaders in the tools don't work the same as ours, there's a lot of guesswork and prep before export and I need to view in engine to make sure everything is rendering correctly rather than being able to view in the authoring tool.

Popov72:
Thanks for the detailed explanation.

A solution would be to have some helpers in Babylon that would take regular rgb/rgba maps for the different components (albedo, roughness, normal, ...) and generate a map with channels filled with the right information, as expected by the engine.

So, for the detail map, we could have something like:
```typescript
ImageHelper.createDetailTexture(albedo: Texture, roughness: Texture, normal: Texture, options: any): Texture
```
`options` would help to customize the inputs (for eg, `albedoIsGrayscale: true` would make the function take `albedo.r` for the detail albedo channel instead of doing `luminance(albedo)`)

sebavan:
Yup let s follow the unity format and we could do a kind of packedTexture which takes 4 textures in inputs, one per channel so it could be used as we want if we see that it becomes a pain point ?

Popov72:
I have created a separate `DetailMap` class.

Should I update the what's new file in the `Major updates` section?

sebavan:
I would say engine update but not major for this one ?

Popov72:
Ok, done.

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1,4 +1,4 @@
-import { serialize, serializeAsImageProcessingConfiguration, expandToProperty } from "../../Misc/decorators";
+import { serialize, serializeAsImageProcessingConfiguration, expandToProperty  } from "../../Misc/decorators";
```
sebavan, line 1: [nit] not sure of the extra space

src/Materials/material.detailMap.ts
```diff
@@ -0,0 +1,262 @@
...
+    DETAIL_NORMALBLENDMETHOD: number;
+
+    /** @hidden */
+    _areTexturesDirty: boolean;
+}
+
+/**
+ * Define the code related to the detail map parameters of a material
+ */
+export class DetailMap {
```
sebavan, line 27: Should it be DetailMapConfiguration for the class name ?

src/Materials/material.detailMap.ts
```diff
@@ -0,0 +1,262 @@
...
+    DETAIL: boolean;
+    DETAILDIRECTUV : number;
+    DETAIL_NORMALBLENDMETHOD: number;
+
+    /** @hidden */
+    _areTexturesDirty: boolean;
+}
+
+/**
+ * Define the code related to the detail map parameters of a material
```
sebavan, line 25: Could we add the sources of inspiration ? and as well in the shader for future references ?

Popov72:
All requested changes performed.

sebavan:
w000000000000t so fassssstttttt :-) you can change the state from draft to ready to review and @deltakosh can do a quick review pass, thanks a lot

sebavan:
Closes #5967

deltakosh:
Congrats fella! always good and efficient!

PR #8300, opened 2020-06-02
adds types to dependencies instead of devDeps

RaananW:
Fixing this - https://forum.babylonjs.com/t/babylon4-1-0-inspector-package-typescript-compilation-error-cannot-find-module-react-cannot-find-namespace-jsx/11303/9

If there is no objection, keep types in deps and not devDeps - @deltakosh , @sebavan


sebavan:
no objections at all as it should have been the case as we can see in loaders :-)

image

PR #8303, opened 2020-06-02
Handle scales in camera matrices

Popov72:
See thorough discussion here: https://forum.babylonjs.com/t/camera-minz-and-maxz-do-not-work-properly-when-camera-parent-is-scaled/11267/13

Here's another PG: https://playground.babylonjs.com/#XTYH95

In this one, the small box is a child of the big box, and the camera is a child of the small box.

When modifying the scale of the small box, you will see that at some time the box disappears: that's when the scale is too low and the box is clipped.

I think it may be strange for a user to have their objects disappear based on their scale, so this PR tries to take into account the scales of the parent matrix to build the camera matrix.

What I did is doing as for transform nodes, meaning creating the camera matrix as being the camera local matrix multiplied by the parent matrix - before, it was constructed by mean of the transformed position/target/up vectors.

The code:
```typescript
this._globalPosition.copyFrom(position);
this._globalCurrentTarget.copyFrom(target);
this._globalCurrentUpVector.copyFrom(up);

if (this.getScene().useRightHandedSystem) {
    Matrix.LookAtRHToRef(this._globalPosition, this._globalCurrentTarget, this._globalCurrentUpVector, this._viewMatrix);
} else {
    Matrix.LookAtLHToRef(this._globalPosition, this._globalCurrentTarget, this._globalCurrentUpVector, this._viewMatrix);
}

if (this.parent) {
    const parentWorldMatrix = this.parent.getWorldMatrix();
    this._viewMatrix.invert();
    this._viewMatrix.multiplyToRef(parentWorldMatrix, this._viewMatrix);
    this._viewMatrix.getTranslationToRef(this._globalPosition);
    this._viewMatrix.invert();
    this._markSyncedWithParent();
}
```
The first `invert` is because the (local) `_viewMatrix` is constructed (by `LookAtLHToRef` / `LookAtRHToRef`) so as to transform world coordinates to view coordinates but we need it the other way around.

After we multiply by the parent matrix, we `invert` it again because `_viewMatrix` is expected to transform world coordinates to view coordinates, not the other way around.

I'm no expert in all those matrix computations and transformations (I did a bit by "try and see if it looks good"), so additional reviews / testing is more than welcome (I did check that the PG from the forum post and the PG linked above do work with the changes)!

### DIFF
--- a/src/Cameras/targetCamera.ts
+++ b/src/Cameras/targetCamera.ts
@@ -424,22 +424,23 @@ export class TargetCamera extends Camera {
     }
 
     protected _computeViewMatrix(position: Vector3, target: Vector3, up: Vector3): void {
-        if (this.parent) {
-            const parentWorldMatrix = this.parent.getWorldMatrix();
-            Vector3.TransformCoordinatesToRef(position, parentWorldMatrix, this._globalPosition);
-            Vector3.TransformCoordinatesToRef(target, parentWorldMatrix, this._globalCurrentTarget);
-            Vector3.TransformNormalToRef(up, parentWorldMatrix, this._globalCurrentUpVector);
-            this._markSyncedWithParent();
-        } else {
-            this._globalPosition.copyFrom(position);
-            this._globalCurrentTarget.copyFrom(target);
-            this._globalCurrentUpVector.copyFrom(up);
-        }
+        this._globalPosition.copyFrom(position);
+        this._globalCurrentTarget.copyFrom(target);
+        this._globalCurrentUpVector.copyFrom(up);
 
         if (this.getScene().useRightHandedSystem) {
             Matrix.LookAtRHToRef(this._globalPosition, this._globalCurrentTarget, this._globalCurrentUpVector, this._viewMatrix);
         } else {
             Matrix.LookAtLHToRef(this._globalPosition, this._globalCurrentTarget, this._globalCurrentUpVector, this._viewMatrix);
+        }
+
+        if (this.parent) {
+            const parentWorldMatrix = this.parent.getWorldMatrix();
+            this._viewMatrix.invert();
+            this._viewMatrix.multiplyToRef(parentWorldMatrix, this._viewMatrix);
+            this._viewMatrix.getTranslationToRef(this._globalPosition);
+            this._viewMatrix.invert();
+            this._markSyncedWithParent();
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Cameras/targetCamera.ts
```diff
@@ -424,23 +424,24 @@ export class TargetCamera extends Camera {
...
-            Vector3.TransformCoordinatesToRef(position, parentWorldMatrix, this._globalPosition);
-            Vector3.TransformCoordinatesToRef(target, parentWorldMatrix, this._globalCurrentTarget);
-            Vector3.TransformNormalToRef(up, parentWorldMatrix, this._globalCurrentUpVector);
-            this._markSyncedWithParent();
-        } else {
-            this._globalPosition.copyFrom(position);
-            this._globalCurrentTarget.copyFrom(target);
-            this._globalCurrentUpVector.copyFrom(up);
-        }
+        this._globalPosition.copyFrom(position);
```
deltakosh, line 427: this values will be false then if there is a parent as we do not take parent into account

src/Cameras/targetCamera.ts
```diff
@@ -424,23 +424,24 @@ export class TargetCamera extends Camera {
...
             Matrix.LookAtLHToRef(this._globalPosition, this._globalCurrentTarget, this._globalCurrentUpVector, this._viewMatrix);
         }
+
+        if (this.parent) {
+            const parentWorldMatrix = this.parent.getWorldMatrix();
+            this._viewMatrix.invert();
+            this._viewMatrix.multiplyToRef(parentWorldMatrix, this._viewMatrix);
+            this._viewMatrix.getTranslationToRef(this._globalPosition);
+            this._viewMatrix.invert();
+            this._markSyncedWithParent();
```
deltakosh, line 443: I like it! it is much cleaner!

src/Cameras/targetCamera.ts
Popov72: I don't understand.

The `_globalXXX` variables are first filled with values in local space (position, target, up).

Then, `_globalPosition` is updated with the world position when there's a parent (when no parent, the local position is the world position).

`_globalCurrentTarget` and `_globalCurrentUpVector` variables are only used in this function for the initial `LookAtLHToRef` computation, so there's no need to update them to world values. Maybe I should remove them altogether and directly use `target` and `up` in the `LookAtLHToRef` call?

deltakosh: Yeah I fill like we can safely remove them

PR #8306, opened 2020-06-03
Fix detail map when no bump mapping in standard material

Popov72:


deltakosh:
same on other PR, seems like something was broken before

PR #8308, opened 2020-06-03
Add support to import frames (frame nodes unconnected)

belfortk:
This PR implement loading the frame without the links. Just the frame and the nodes inside. Any node ports that are exposed are also kept across import/export.

### DIFF
--- a/nodeEditor/src/diagram/graphCanvas.tsx
+++ b/nodeEditor/src/diagram/graphCanvas.tsx
@@ -857,7 +857,7 @@ export class GraphCanvasComponent extends React.Component<IGraphCanvasComponentP
         // Frames
         if (editorData.frames) {
             if (isImportingAFrame) {
-                var frame = GraphFrame.Parse(editorData.frames[editorData.frames.length - 1], this, editorData.map);
+                const frame = GraphFrame.Parse(editorData.frames[editorData.frames.length - 1], this, editorData.map);
                 this._frames.push(frame);
                 this.globalState.onSelectionChangedObservable.notifyObservers(frame);
             } else {
--- a/nodeEditor/src/diagram/graphCanvas.tsx
+++ b/nodeEditor/src/diagram/graphCanvas.tsx
@@ -842,30 +842,30 @@ export class GraphCanvasComponent extends React.Component<IGraphCanvasComponentP
     }
 
     processEditorData(editorData: IEditorData, isImportingAFrame = false) {
-        if (!isImportingAFrame) {
-            const frames = this._frames.splice(0);
-            for (var frame of frames) {
-                frame.dispose();
-            }
-
-            this._frames = [];
-            this.x = editorData.x || 0;
-            this.y = editorData.y || 0;
-            this.zoom = editorData.zoom || 1;
+        const frames = this._frames.splice(0);
+        for (var frame of frames) {
+            frame.dispose();
         }
+
+        this._frames = [];
+        this.x = editorData.x || 0;
+        this.y = editorData.y || 0;
+        this.zoom = editorData.zoom || 1;
 
         // Frames
         if (editorData.frames) {
-            if (isImportingAFrame) {
-                const frame = GraphFrame.Parse(editorData.frames[editorData.frames.length - 1], this, editorData.map);
+            for (var frameData of editorData.frames) {
+                var frame = GraphFrame.Parse(frameData, this, editorData.map);
                 this._frames.push(frame);
-                this.globalState.onSelectionChangedObservable.notifyObservers(frame);
-            } else {
-                for (var frameData of editorData.frames) {
-                    var frame = GraphFrame.Parse(frameData, this, editorData.map);
-                    this._frames.push(frame);
-                }
-            }
+            }  
+        }
+    }
+
+    addFrame(editorData: IEditorData) {
+        if (editorData.frames) {
+            const frame = GraphFrame.Parse(editorData.frames[editorData.frames.length - 1], this, editorData.map);
+            this._frames.push(frame);
+            this.globalState.onSelectionChangedObservable.notifyObservers(frame);
         }
     }
  
--- a/nodeEditor/src/diagram/graphCanvas.tsx
+++ b/nodeEditor/src/diagram/graphCanvas.tsx
@@ -841,23 +841,30 @@ export class GraphCanvasComponent extends React.Component<IGraphCanvasComponentP
         this.props.globalState.onRebuildRequiredObservable.notifyObservers();
     }
 
-    processEditorData(editorData: IEditorData) {
-        const frames = this._frames.splice(0);
-        for (var frame of frames) {
-            frame.dispose();
+    processEditorData(editorData: IEditorData, isImportingAFrame = false) {
+        if (!isImportingAFrame) {
+            const frames = this._frames.splice(0);
+            for (var frame of frames) {
+                frame.dispose();
+            }
+
+            this._frames = [];
+            this.x = editorData.x || 0;
+            this.y = editorData.y || 0;
+            this.zoom = editorData.zoom || 1;
         }
-
-        this._frames = [];
-
-        this.x = editorData.x || 0;
-        this.y = editorData.y || 0;
-        this.zoom = editorData.zoom || 1;
 
         // Frames
         if (editorData.frames) {
-            for (var frameData of editorData.frames) {
-                var frame = GraphFrame.Parse(frameData, this, editorData.map);
+            if (isImportingAFrame) {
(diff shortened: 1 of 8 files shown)


nodeEditor/src/diagram/graphCanvas.tsx
```diff
@@ -841,23 +841,30 @@ export class GraphCanvasComponent extends React.Component<IGraphCanvasComponentP
...
+            this.y = editorData.y || 0;
+            this.zoom = editorData.zoom || 1;
+        }
 
         // Frames
         if (editorData.frames) {
-            for (var frameData of editorData.frames) {
-                var frame = GraphFrame.Parse(frameData, this, editorData.map);
+            if (isImportingAFrame) {
+                const frame = GraphFrame.Parse(editorData.frames[editorData.frames.length - 1], this, editorData.map);
```
deltakosh, line 860: we can assume it is editorData.frames[0] right?

deltakosh: or even better do we really need to test isImportingAFrame?

nodeEditor/src/graphEditor.tsx
```diff
@@ -157,6 +157,11 @@ export class GraphEditor extends React.Component<IGraphEditorProps, IGraphEditor
             }
         });
 
+        this.props.globalState.onImportFrameObservable.add(() => {
+            this.loadGraph();
```
deltakosh, line 161: why do we need to call this.loadGraph() ?

src/Materials/Node/nodeMaterial.ts
```diff
@@ -1677,6 +1677,53 @@ export class NodeMaterial extends PushMaterial {
         this._mode = source.mode ?? NodeMaterialModes.Material;
     }
 
+    /**
+     * Load new blocks and frame from a frame serialization object
+     * @param source defines the JSON representation of the frame
+     * @param rootUrl defines the root URL to use to load textures and relative dependencies
+     */
+    public addFrameNodesFromeSerialization(source: any, rootUrl: string = "") {
```
deltakosh, line 1685: I would like to let the NOdeMaterial unchanged

deltakosh: No code  should be added on bjs core

nodeEditor/src/diagram/graphCanvas.tsx
belfortk: I refactored this to be more clear after talking with @sebavan .

Because I am appending new frames to editorData.frames, when I import multiple frames in a row, the new frame is not editorData.frames[0], it is at the end of the array.

nodeEditor/src/graphEditor.tsx
belfortk: I was using it to draw the nodes. refactored so now i am drawing nodes in GraphEditor.addFrameFromSerialization()

src/Materials/Node/nodeMaterial.ts
belfortk: moved to graphEditor

nodeEditor/src/graphEditor.tsx
deltakosh: you should be bale to ask for a repaint with globalData.refreshRequired if my memory serves

nodeEditor/src/graphEditor.tsx
```diff
@@ -157,6 +158,18 @@ export class GraphEditor extends React.Component<IGraphEditorProps, IGraphEditor
             }
         });
 
+        this.props.globalState.onImportFrameObservable.add((source: any) => {
+            let editorData = this.props.globalState.nodeMaterial.editorData;        
+    
+            if (editorData instanceof Array) {
```
deltakosh, line 164: is it required?

nodeEditor/src/graphEditor.tsx
```diff
@@ -475,6 +495,47 @@ export class GraphEditor extends React.Component<IGraphEditorProps, IGraphEditor
...
+            if (blockType) {
+                let block: NodeMaterialBlock = new blockType();
+                block._deserialize(parsedBlock, this.props.globalState.nodeMaterial.getScene(), rootUrl);
+                map[parsedBlock.id] = block;
+
+                this.createNodeFromObject(block);
+            }
+        }
+
+        if (source.locations || source.editorData && source.editorData.locations) {
```
deltakosh, line 513: I feel like we should not duplicate that code but instead reuse the one we already have
We could update loadFromSerialization to have a flag that let the function knows we do not want to clear the current graph (so skip the this.clear) and instead add loaded data to current data

My ask was to not mention anything frame related in bjs core as the concept of frame does not exist there but it is ok to update the existing functions to make them more flexible

nodeEditor/src/graphEditor.tsx
```diff
@@ -475,6 +495,47 @@ export class GraphEditor extends React.Component<IGraphEditorProps, IGraphEditor
...
+            }[] = source.locations || source.editorData.locations;
+
+            for (var location of locations) {
+                if (map[location.blockId]) {
+                    location.blockId = map[location.blockId].uniqueId;
+                }
+            }
+
+            editorData.locations = editorData.locations.concat(locations);
+            editorData.frames = editorData.frames ? editorData.frames.concat(source.editorData.frames) : source.editorData.frames;
```
deltakosh, line 527: this will stay here of course

nodeEditor/src/diagram/graphCanvas.tsx
```diff
@@ -858,7 +857,15 @@ export class GraphCanvasComponent extends React.Component<IGraphCanvasComponentP
...
                 var frame = GraphFrame.Parse(frameData, this, editorData.map);
                 this._frames.push(frame);
-            }
+            }  
+        }
+    }
+
+    addFrame(editorData: IEditorData) {
+        if (editorData.frames) {
+            const frame = GraphFrame.Parse(editorData.frames[editorData.frames.length - 1], this, editorData.map);
```
deltakosh, line 866: why not editorData.frames[0] ?

nodeEditor/src/diagram/graphCanvas.tsx
```diff
@@ -841,14 +841,13 @@ export class GraphCanvasComponent extends React.Component<IGraphCanvasComponentP
         this.props.globalState.onRebuildRequiredObservable.notifyObservers();
     }
 
-    processEditorData(editorData: IEditorData) {
+    processEditorData(editorData: IEditorData, isImportingAFrame = false) {
```
deltakosh, line 844: 2nd parameter does not seem to be used anymore

nodeEditor/src/diagram/graphCanvas.tsx
deltakosh: That does not seem to be the case in your code now. I see a call to it once only. Where does the import multiple frames in a row happening?

belfortk: If a user imports a frame, GraphEditor.addFrameFromSerialization() will concat that frame to editorData.frames. If the user imports a second frame in the same session, the second frame is concat-ed to editorData.frame. So the frame I need to parse and add is at the end of editorData.frame.

nodeEditor/src/graphEditor.tsx
belfortk: I no longer need the repaint since I am calling createNodeFromObject() in GraphEditor.addFrameFromSerialization()

nodeEditor/src/graphEditor.tsx
belfortk: I am not 100% sure. I added it because it was in GraphEditor.build() and since nodeMaterial.editorData is of type: any I figured it was better safe than sorry to add the check. I can remove it if you think it is not needed.

nodeEditor/src/diagram/graphCanvas.tsx
belfortk: because if the user imports multiple frames in a row, the frame that needs to be added to the editor is at the end of editorData.frames.

nodeEditor/src/diagram/graphCanvas.tsx
belfortk: will remove!

nodeEditor/src/diagram/graphCanvas.tsx
deltakosh: Ok maybe I do not understand it correctly.: if an user imports a frame the editorData of the file will only have one frame in it right?

nodeEditor/src/graphEditor.tsx
deltakosh: Yeah I think it is safe to remove it

deltakosh:
All good thanks!!!

PR #8337, opened 2020-06-08
added virtualJoystick customizations

Rockwell15:


src/Misc/virtualJoystick.ts
```diff
@@ -73,7 +91,30 @@ export class VirtualJoystick {
...
+        var {
+            puckSize = 40,
+            containerSize = 60,
+            color = "cyan",
+            puckImage = undefined,
+            containerImage = undefined,
+            position = undefined,
+            alwaysVisible = false,
+            limitToContainer = false,
+        } = customizations || {};
```
deltakosh, line 116: we can use this kind of construct to ensure that customizations is always correctly defined:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Helpers/environmentHelper.ts#L343

src/Misc/virtualJoystick.ts
```diff
@@ -284,6 +378,70 @@ export class VirtualJoystick {
...
+
+        delete this._joystickPosition;
+    }
+
+    /**
+    * Sets whether or not joystick container is always visible.
+    * Can only be always visible if `_joystickPosition` is set
+    * @param alwaysVisible defines whether or not the joystick container is always visible
+    */
+    public setAlwaysVisible(alwaysVisible: boolean) {
```
deltakosh, line 415: Maybe a property?

src/Misc/virtualJoystick.ts
```diff
@@ -284,6 +378,70 @@ export class VirtualJoystick {
...
+        this._joystickPuckSize = newSize;
+        this._clearPuckSize = ~~(this._joystickPuckSize * 2.1);
+        this._clearPuckSizeOffset = ~~(this._clearPuckSize / 2);
+    }
+
+    /**
+    * Sets the size of the container
+    * @param newSize container size
+    */
+    public setContainerSize(newSize: number) {
```
deltakosh, line 395: Same as other, I would recommend to use property instead of setXXX

deltakosh:
Also there is an issue with the documentation check

src/Misc/virtualJoystick.ts
Rockwell15: @deltakosh 

I made a method for this since it didn't make sense to me to have a joystick always visible if it doesn't have a set position, so modifying this behavior via a method allows for a check that position is set.

Do you think `alwaysVisible` should be allowed whether or not a fixed position is set?

src/Misc/virtualJoystick.ts
Rockwell15: @deltakosh 

Just to be clear, are you suggesting I should make an options property on virtualJoystick to hold all the customization data? Or were you just pointing out the use of an interface?

deltakosh: I'm pointing the use of... :)

src/Misc/virtualJoystick.ts
```diff
@@ -43,12 +82,30 @@ export class VirtualJoystick {
...
+    public limitToContainer: boolean;
+
     // Used to draw the virtual joystick inside a 2D canvas on top of the WebGL rendering canvas
     private static _globalJoystickIndex: number = 0;
+    private static _alwaysVisibleSticks: number = 0;
     private static vjCanvasContext: CanvasRenderingContext2D;
     private static vjCanvasWidth: number;
     private static vjCanvasHeight: number;
     private static halfWidth: number;
+    private static _getDefaultOptions(): VirtualJoystickCustomizations {
```
deltakosh, line 97:
```suggestion
    private static _GetDefaultOptions(): VirtualJoystickCustomizations {

```

src/Misc/virtualJoystick.ts
```diff
@@ -277,13 +411,82 @@ export class VirtualJoystick {
...
+        this.alwaysVisible = false;
+
+        delete this._joystickPosition;
+    }
+
+    /**
+     * Defines whether or not the joystick container is always visible
+     */
+    public set alwaysVisible(value: boolean) {
+        if (this._alwaysVisible == value) {
```
deltakosh, line 458: ===

Rockwell15:
@deltakosh 

I see you've approved the commit but mentioned 2 changes. Do I make another commit & pull request w/ the 2 changes?

deltakosh:
Nope you can add to this PR:) I will squash everything when merging :)

Thanks again!

deltakosh:
Good job!! Do you want to update the doc page now?

Rockwell15:
Thanks!

Yeah, do I need to add a "whats's new" to that too? Or should just updating the VirtualJoystick "how to" suffice?

deltakosh:
nope jsut  updating the vjs page would be perfect :)

PR #8354, opened 2020-06-11
react types as peer dependency

RaananW:
i used the initial version defined as the base for the dependency semver. Anything higher than this (we are already using the 16.9.x branch) will be accepted.


sebavan:
@RaananW could you check if the deploy scripts in the publisher is well redirecting peers in the package.json ? I can not remember it :-)

RaananW:
> @RaananW could you check if the deploy scripts in the publisher is well redirecting peers in the package.json ? I can not remember it :-)

Good question. checking

RaananW:
image
It does seem like it, but I am running npm locally just to be sure.

RaananW:
@sebavan - yes it does.

PR #8375, opened 2020-06-12
NME: Add rgba and a to the outputs of ReflectionTexture block

Popov72:
See https://forum.babylonjs.com/t/reflection-only-material/11621/7


Popov72:
Error:
```
[18:05:11] 'buildMin' errored after 31 s
[18:05:11] Error in plugin "webpack-stream"
Message:
    [at-loader] ../../inspector/src/components/actionTabs/tabs/propertyGrids/variantsPropertyGridComponent.tsx:41:52 
    TS2339: Property 'Loader' does not exist on type 'typeof GLTF2'.
Details:
    domainEmitter: [object Object]
    domainThrown: false
```

deltakosh:
pinging @sebavan who is probably already aware

sebavan:
yup coming from a quick workaround. should be properly fixed in about an hour or two. Else the inspector was all broken on the pg :-(

sebavan:
you can take back master in

PR #8385, opened 2020-06-15
Make glTF default sheen weight equal to one

MiiBond:


MiiBond:
We're discussing this on a call right now. I think we're deciding to default extensions to `off`, just like clear coat. So, the spec should change for sheen rather than the loader implementation. I'll close this.

sebavan:
Ok no problem I ll update the spec today.

PR #8388, opened 2020-06-15
Render transparent meshes with the same stencil state than the opaque…

Popov72:
… ones

In `RenderingGroup.render`, there's a `engine.setStencilBuffer(false);` issued before rendering sprites / particles / transparent meshes.

Regarding sprites and particles I can understand we want to control the rendering and not be bothered with the stencil buffer (albeit now that you can have custom shaders - at least for particles - it may be questionable), but for transparent meshes it's a bit strange to reset the stencil buffer to `false`.

I had been working with the stencil buffer in a PG and everything was working well until I added some transparency to my material, making it switch from the opaque to the transparent queue, making my PG suddenly fail.

This PR simply sets the stencil buffer state to the same value than the one used to render the opaque/alpha test meshes.

As it is possibly a breaking change, I have updated the corresponding section of the What's new file.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -235,3 +235,4 @@
 - HDRCubeTexture default rotation is now similar to the industry one. You might need to add a rotation on y of 90 degrees if you scene changes ([Sebavan](https://github.com/sebavan/))
 - PBRMaterial index of refraction is now defined as index of refraction and not the inverse of it ([Sebavan](https://github.com/sebavan/))
 - `SceneLoaderProgress` class is now `ISceneLoaderProgress` interface ([bghgary](https://github.com/bghgary))
+- Rendering of transparent meshes: stencil state is now set to the value registered in the engine instead of being set to `false` unconditionally ([Popov72](https://github.com/Popov72))
--- a/src/Layers/highlightLayer.ts
+++ b/src/Layers/highlightLayer.ts
@@ -513,6 +513,17 @@ export class HighlightLayer extends EffectLayer {
     }
 
     /**
+     * Returns true if the mesh can be rendered, otherwise false.
+     * @param mesh The mesh to render
+     * @param material The material used on the mesh
+     * @returns true if it can be rendered otherwise false
+     */
+    protected _canRenderMesh(mesh: AbstractMesh, material: Material): boolean {
+        // all meshes can be rendered in the highlight layer, even transparent ones
+        return true;
+    }
+
+    /**
      * Adds specific effects defines.
      * @param defines The defines to add specifics to.
      */
--- a/src/Rendering/outlineRenderer.ts
+++ b/src/Rendering/outlineRenderer.ts
@@ -299,7 +299,7 @@ export class OutlineRenderer implements ISceneComponent {
         this._savedDepthWrite = this._engine.getDepthWrite();
         if (mesh.renderOutline) {
             var material = subMesh.getMaterial();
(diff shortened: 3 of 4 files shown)


Popov72:
In fact the need for the stencil buffer for transparent meshes is more to read from it than to write.

What I was doing is simulating a Xray:
image
The dragon is drawn is the stencil buffer. Then, where the plane is drawn, the stencil is incremented by one. So, we end up with 2 in the stencil where the dragon and the plane intersects.

I draw the dragon skeleton with a transparent material (and depth test = false) where stencil = 2. This last pass does not write to the stencil, it is using (reading from) it to draw only the skeleton where it should.

There's a `onBeforeTransparentRendering` callback in the `RenderingGroup` class, but to access it means we need to retrieve a pointer to the rendering group used by the renderer (don't know if it is possible / easy).

Apart from that, it seems wrong to me that the stencil is disabled when rendering transparent objects. I understand that transparent objects are particular in that they are sorted / rendered last / don't write to the zbuffer, but the stencil is not really part of this equation. As said above, there are valid usages for the stencil for transparent objects when using it only as a mask to write/not write in the color buffer (and not to write to the stencil).

I don't understand this one:
> This otherwise could break all the current code relying on stencil. Any thoughts ?

The stencil state is only set just before rendering the transparent objects, and it is set to `false` by the existing code just after. So, the only possible breaking change is if people enable the stencil buffer and have transparent objects to render.

sebavan:
Currently highlights and glow are relying on stencil and expect to not be impacted if transparent meshes are in the scene.

Yup accessing the rendering group might not be as convenient, so lets test first what the impact is on highlights layer ? and the other functionnalities relying on stencil:

- HighlightLayer
- OutlineRenderer
- GlowLayer

Popov72:
As I can see it, the glow layer does not use the stencil buffer. The code used for the rendering is:
```typescript
var engine = this._engine;
var previousStencilBuffer = engine.getStencilBuffer();
engine.setStencilBuffer(false);
engine.drawElementsType(Material.TriangleFillMode, 0, 6);
engine.setStencilBuffer(previousStencilBuffer);
```

Regarding the highlight layer, this code change makes it fail if adding a transparent mesh to the highlight layer.

**Left sphere not added to the layer**:
image
**Left sphere added to the layer**:
image
With the not modified code, we end up with picture 1 even if adding the sphere to the layer.

However, if adding this to the `HighlightLayer` class:
```typescript
protected _canRenderMesh(mesh: AbstractMesh, material: Material): boolean {
    return true;
}
```
It does make work the highlight layer for transparent meshes:
image
It seems an improvement to me :)

Regarding the outline renderer, I performed several tests with and without transparent objects and the change from the PR didn't change anything.

sebavan:
It does nothing by design cause overlapping is going wrong with transparency in a couple cases.

Try overlapping a bunch of them with highlights for testing :-)

Popov72:
Like this:

**Without the highlight layer**:
image
**With the layer**:
image

?

I don't think the transparency property of a mesh is a problem to render in the highlight layer because the alpha/visibility/opacity texture values are not involved in the special renderings done by the layer: everything is done with some specific shaders that does not use the alpha/visibility/opacity texture properties of the mesh/material.

sebavan:
yup looks like it is ok (the issue was before related to the depth not being written) so ship it :-) and if that breaks, you can adapt it back I guess.

Popov72:
I have also performed a small correction in the outline renderer.

PR #8396, opened 2020-06-16
Fix error handling and KTX/Basis loading issues

### LINKED ISSUE #8394: KTX2 with KHR_image_basisu fails to load when using .glb/.bin embedding
The attached file contains several variants of a model, exported with gltfpack using KHR_image_basisu with KTX2 containers.

The "separate" version loads fine in Babylon JS using latest sandbox; this is the version with separate files.
Both "embedded" (textures inside .bin) and glb (single file) versions fail to load with a 2 MB error message along the lines of

> Uncaught (in promise) Error: /textures/1: Error while trying to load image: 
> 171,75,84,88,32,50,48,187,13,10,26,10,0,0,0,0,1,0,0,0,0,8,0,0,0,8,0,0,0,0,0,0,0,0,0,0,1,0,0,0,12,0,0,0,1,0,0,0,112,1,0,0,76,0,0,0,188,1,0,0,24,0,0,0,216,1,0,0,0,0,0,0,199,36,0,0,0,0,0,0,51,149,2,0,0,0,0,0,130,97,5,0,0,0,0,0,0,0,0,0,0,0,0,0,187,221,0,0,0,0,0,0,120,183,1,0,0,0,0,0,0,0,0,0,0,0,0,0,32,89,0,0,0,0,0,0,155,132,0,0,0,0,0,0,0,0,0,0,0...
> - Fallback texture was used

It would be nice to emit shorter error messages :) and also indicate the reason for the error. I'm reasonably sure that the file in question is correct, and it's a bug in Babylon.JS. Unsure what the problem is precisely, as the file header looks correct based on the error message (starts with "\xAB KTX 20"); maybe the MIME type isn't recognized correctly?

[rubyrose.zip](https://github.com/BabylonJS/Babylon.js/files/4783758/rubyrose.zip)

bghgary:
Fixes #8394

* Update Basis flag to disallow loose cubemaps as it's not supported in the texture loader right now.
* Fix issue with error handling such that errors show up in the sandbox again.
* Truncate the error message if it's too long.
* Use mime type in internal texture loaders.

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -356,7 +356,7 @@ export class GLTFLoader implements IGLTFLoader {
             });
 
             return resultPromise;
-        }, (error) => {
+        }).catch((error) => {
             if (!this._disposed) {
                 this._parent.onErrorObservable.notifyObservers(error);
                 this._parent.onErrorObservable.clear();
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -2797,7 +2797,7 @@ export class ThinEngine {
         let loader: Nullable<IInternalTextureLoader> = null;
 
         for (const availableLoader of ThinEngine._TextureLoaders) {
-            if (availableLoader.canLoad(extension)) {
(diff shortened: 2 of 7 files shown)


deltakosh:
One doc issue but else LGTM

PR #8397, opened 2020-06-17
Adding frame playback

toledoal:
Hello 😄 !

In this PR I continue integrating the design redlines. It also adds the dragging behavior of the play-head and bug fixes for adding animations.


inspector/src/components/actionTabs/tabs/propertyGrids/animations/playhead.tsx
```diff
@@ -1,27 +1,95 @@
...
+    e.preventDefault();
+    this._direction = 0;
+    this._active = false;
+  }
 
+  calculateMove() {
+    return `calc(${this.props.frame * this.props.offset}px - 13px)`;
+  }
+
+  moveFrameTo(e: any) {
```
deltakosh, line 54: why any? we should avoid every any

toledoal: good catch,. Thanks! I missed that

deltakosh:
Seems something is wrong in the tests

toledoal:
let me check it here as well,

toledoal:
Yeah I missed to remove one unused import

PR #8412, opened 2020-06-19
GLTF: use min/max values from position accessor to set the bounding box of meshes

Popov72:
See #4102 

As an aside, how can I get rid of the first 3 commits that are displayed in this PR?

They are not related to the PR, I don't know why they are here... I guess it's because I have done them on the master branch of my local repo(?)


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -686,8 +687,30 @@ export class GLTFLoader implements IGLTFLoader {
...
+                        max.copyFromFloats(...mmax);
+                        babylonMesh.getBoundingInfo().reConstruct(min, max);
+                        if (babylonMesh.subMeshes) {
+                            for (let index = 0; index < babylonMesh.subMeshes.length; index++) {
+                                babylonMesh.subMeshes[index].getBoundingInfo().reConstruct(min, max);
+                            }
+                        }
+                        babylonMesh._updateBoundingInfo();
+                    }
+                }
```
bghgary, line 708: Is it possible to do this as soon as the min/max is read to avoid using metadata?

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -686,8 +687,30 @@ export class GLTFLoader implements IGLTFLoader {
...
+                let recomputeBoundingInfo = true;
+                if (!this.parent.alwaysComputeBoundingBox && !babylonMesh.skeleton) {
+                    const tmp = GLTFLoader.GetTmpMetadata(babylonMesh);
+                    const mmin: [number, number, number] = tmp.positionMin, mmax: [number, number, number] = tmp.positionMax;
+                    if (mmin !== undefined && mmax !== undefined) {
+                        recomputeBoundingInfo = false;
+                        min.copyFromFloats(...mmin);
+                        max.copyFromFloats(...mmax);
+                        babylonMesh.getBoundingInfo().reConstruct(min, max);
+                        if (babylonMesh.subMeshes) {
```
bghgary, line 701: The sub meshes have the same min/max as the parent? That seems a bit weird.

Popov72: You told me that no sub meshes were created by the GLTF loader, so I think there is a single sub mesh and this sub mesh == main mesh?

loaders/src/glTF/2.0/glTFLoader.ts
Popov72: To avoid using metadata, I would need to add this code directly in the `loadAttribute` function.

But, can I be sure that the `babylonMesh.skeleton` property is up to date at that time, as I need it in the code above?

Popov72: Another problem is that `babylonGeometry.applyToMesh(babylonMesh)` is done later on (in relation to `loadAttribute`), and calling `applyToMesh` leads to creating the bounding info based on the current vertex data buffer, so overwrite the bouding box I create in `loadAttribute`.

Popov72: I have tried something, let me know what you think about the changes.

Note: it seems the `skeleton` property is indeed up to date, at least in my testing.

loaders/src/glTF/2.0/glTFLoader.ts
Popov72: I could remove the code updating the sub meshes altogether as if there is a single sub mesh == main mesh, it will end up updating the same mesh twice...

Instead, I could add a comment saying that because GLTF does not create additional sub meshes, there's no need to update sub meshes bounding info?

sebavan:
Ping @bghgary for second pass review :-) I ll merge it right after.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -864,6 +866,27 @@ export class GLTFLoader implements IGLTFLoader {
             const accessor = ArrayItem.Get(`${context}/attributes/${attribute}`, this._gltf.accessors, attributes[attribute]);
             promises.push(this._loadVertexAccessorAsync(`/accessors/${accessor.index}`, accessor, kind).then((babylonVertexBuffer) => {
                 babylonGeometry.setVerticesBuffer(babylonVertexBuffer, accessor.count);
+                if (babylonVertexBuffer.getKind() === VertexBuffer.PositionKind) {
+                    postProcesses.push(() => {
```
bghgary, line 870: I don't think this is a good idea. Some day, we will remove the underscore from the _loadVertexDataAsync function so that it will be part of the contract for glTF loader extension (currently because of the underscore, it is considered experimental). I don't think adding this kind of mechanism should be part of this contract.

bghgary: It seems like Geometry has bounding info. I wonder if it makes sense to update the geometry's bounding box info with the info from min/max and then make sure the code is taking advantage of it?

I also see things like the mesh taking the world matrix into account for the bounding info. Does this code work when there is a non-identity transform on the mesh?

Popov72: Regarding the bounding info of the `Geometry class`, it is in fact not used in the class.

So, I have added the possibility to use this bounding info as the bounding info applied to meshes sharing the geometry and have updated the GLTF loader to use this new capability.

Let me know if it's ok for you.

Note: I have checked that using non identity transforms does work with the new code.

bghgary: Nice. I like this much better. Is the flag on the geometry necessary? Can we use the presence of the boundingInfo? @sebavan ?

Popov72: I have updated the code. Let's wait for @sebavan inputs.

Note that `Geometry._boundingInfo` is never set by the class, and the only one location it is read from is here (before the code update):
```
        if (this.isReady()) {
            this._applyToMesh(mesh);
        }
        else {
            mesh._boundingInfo = this._boundingInfo;
        }
```
It is at the end of `applyToMesh`.

src/Meshes/geometry.ts
```diff
@@ -275,9 +275,12 @@ export class Geometry implements IGetSetVerticesData {
             var meshes = this._meshes;
             var numOfMeshes = meshes.length;
 
+            var minimum = this._boundingInfo?.minimum ?? this._extend.minimum;
```
sebavan, line 278: Would not this be a breaking change preventing to update bounding info if they were already present e.g. update positions as we often do in pg ?

sebavan: Also as we are not bypassing the _updateExtend we will not save anything ?

Popov72: So I think adding back the `useBoundingInfoFromGeometry` property is the safest thing to do, as the new code will only be executed if it is `true` and so no breaking change possible?

I can update `_updateExtend` like this:
```typescript
private _updateExtend(data: Nullable<FloatArray> = null) {
    if (this.useBoundingInfoFromGeometry && this._boundingInfo) {
        this._extend = {
            minimum: this._boundingInfo.minimum.clone(),
            maximum: this._boundingInfo.maximum.clone()
        }
    } else {
        if (!data) {
            data = this.getVerticesData(VertexBuffer.PositionKind)!;
        }

        this._extend = extractMinAndMax(data, 0, this._totalVertices, this.boundingBias, 3);
    }
}
```

sebavan:
Looks all good to me let s remove the draft flag and I ll merge it in :-)

Thanks a lot !!!

PR #8425, opened 2020-06-23
Fix bug in WebXRHit test due to directly casting to Babylon types

Alex-MSFT:
This is fixing a regression in the WebXR Hit Test implementation caused by a change to the Babylon implementation of Vector3 and Quaternion.  This code was incorrectly casting the a WebXR Transform's position and orientation properties directly to Babylon types, which do not fulfill the same class contract as the underlying member variables are named differently.  This used to work as the variables had identical naming. 

To fix this I'm simply changing the call from "copyFrom" to "set" and referencing the transform properties directly without casting.  I have tested this in the BabylonNative project and verified the fix works.  @RaananW If you could help me test this out in the web playground either by giving me some pointers for how to do this properly, or validating for me that would be great!

### DIFF
--- a/src/XR/features/WebXRHitTest.ts
+++ b/src/XR/features/WebXRHitTest.ts
@@ -226,8 +226,10 @@ export class WebXRHitTest extends WebXRAbstractFeature implements IWebXRHitTestF
             if (!pose) {
                 return;
             }
-            this._tmpPos.copyFrom(pose.transform.position as unknown as Vector3);
-            this._tmpQuat.copyFrom(pose.transform.orientation as unknown as Quaternion);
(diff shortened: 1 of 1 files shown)


RaananW:
Great! Missed this one. This is the right way to go - I'll run a few tests tomorrow

sebavan:
@RaananW I let you merge when your tests are done and I ll run a nightly just after.

RaananW:
I'm merging, this is the right fix. No worries about the what's new check failing

sebavan:
Perfect I ll do a nightly in the next hour, thanks for the fix !!!

PR #8428, opened 2020-06-24
Fix crash with CSG when no uvs defined

Popov72:
See https://forum.babylonjs.com/t/export-the-boolean-generated-new-mesh-to-a-glb-file-load-the-glb-file-into-the-scene-and-then-boolean-the-error/11958

### DIFF
--- a/src/Meshes/csg.ts
+++ b/src/Meshes/csg.ts
@@ -42,7 +42,7 @@ class Vertex {
         /**
          * The texture coordinate of the vertex
          */
-        public uv: Vector2,
+        public uv?: Vector2,
         /**
          * The texture coordinate of the vertex
          */
@@ -54,7 +54,7 @@ class Vertex {
      * @returns A new Vertex
      */
     public clone(): Vertex {
-        return new Vertex(this.pos.clone(), this.normal.clone(), this.uv.clone(), this.vertColor?.clone());
+        return new Vertex(this.pos.clone(), this.normal.clone(), this.uv?.clone(), this.vertColor?.clone());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Meshes/csg.ts
```diff
@@ -75,7 +75,7 @@ class Vertex {
     public interpolate(other: Vertex, t: number): Vertex {
         return new Vertex(Vector3.Lerp(this.pos, other.pos, t),
             Vector3.Lerp(this.normal, other.normal, t),
-            Vector2.Lerp(this.uv, other.uv, t),
+            this.uv && other.uv ? Vector2.Lerp(this.uv, other.uv, t) : undefined,
```
RaananW, line 78: I know the next line is doing the same, but this way of single-line assignment seems unreadable to me. A simple `if()` would be better IMO

Popov72: Not sure what you mean? May you had something like this in mind: (a && b) && (c=d) (which is indeed not very readable in my opinion)?

Here it is doing:
```javascript
new Vertex(
    Vector3.Lerp(this.pos, other.pos, t),
    Vector3.Lerp(this.normal, other.normal, t),
    this.uv && other.uv ? Vector2.Lerp(this.uv, other.uv, t) : undefined,
    this.vertColor && other.vertColor ? Color4.Lerp(this.vertColor, other.vertColor, t) : undefined
);
```
I think it is readable(?)

The alternate way with `if` would be:
```typescript
public interpolate(other: Vertex, t: number): Vertex {
    let uv = undefined;
    if (this.uv && other.uv) {
        uv = Vector2.Lerp(this.uv, other.uv, t);
    }
    let vertColor = undefined;
    if (this.vertColor && other.vertColor) {
        vertColor = Color4.Lerp(this.vertColor, other.vertColor, t)
    }
    return new Vertex(
        Vector3.Lerp(this.pos, other.pos, t),
        Vector3.Lerp(this.normal, other.normal, t),
        uv,
        vertColor);
}
```

RaananW: It is all good. Personally if feels like too much for a constructor (thou i probably do those myself from time to time). I approved it and will merge it very soon.

Issue #4103, opened 2018-04-09
Expose glTF extension interfaces so that export can use them


sbtron:
Needed for #4098

bghgary:
Just take all of the interfaces in the loader extensions and put them in a new file so that serializers can use them.

Popov72:
Why not just exporting the interfaces instead of moving them in a new file?

bghgary:
That's a possibility. We weren't using modules when this issue was created.

sebavan:
Yup from the same file will do !!!

PR #8429, opened 2020-06-24
GLTF: export all interfaces

Popov72:
See #4103


sebavan:
Now that they are public, they all need more docs :-)

https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=6371&view=logs&j=182f86de-37f5-5af4-c789-e92a7c8fa3dc&t=14c2bff8-fe23-5e9c-07aa-9b09ac155f63

Popov72:
Ah, fantastic...

bghgary:
Since it's so easy to export, I'm wondering maybe we don't need to do this until we need them in the exporter. Also, even if we do, the documentation for this seems unnecessary. I'm not sure if it's useful for people to see this in the documentation. Maybe just use `@hidden` for these?

sebavan:
Ok all good for me so let s just close this and the associated issue @bghgary ??? would prevent @Popov72 to write thousands of missing comments and I feel totally aligned on doing it later based on need.

bghgary:
Do we already have any duplicate interface definitions between the loader and exporter? Maybe just fix those if there are any.

Popov72:
I have added the @hidden tags as it was a quick change.

How can I import some types from the loaders into the serializers? Should the webpack config be changed to reference the loader package?

The 3 extensions supported by the serializer are redefining the interfaces found in the loader extensions.

sebavan:
Maybe we should place them in the gltf definition d.ts instead as it is already accessible by both ? as it feels weird to have the extra npm dependency ???

Popov72:
You are speaking of babylonjs-gltf2interface?

Let me know what I should do.

sebavan:
Yup speaking about this one. @bghgary any thoughts ? I would really prefer prevent the dependency for typings only if we can.

bghgary:
I'm okay with putting it in gltf2-interface.

sebavan:
Perfect sounds like a plan to me !!! :-)

sebavan:
LGTM will merge as soon as the tests are ok :-)

bghgary:
Since gltf2-interface is public, it might be a good idea to make the names of the interfaces more consistent. Maybe we should come up with a naming scheme?

dist/preview release/glTF2Interface/babylon.glTF2Interface.d.ts
```diff
@@ -905,4 +905,286 @@ declare module BABYLON.GLTF2 {
...
+    }
+
+    /**
+     * Interfaces from the ExtrasAsMetadata extension
+     */
+
+    /** @hidden */
+    interface ObjectWithMetadata {
+        metadata: any;
+    }
```
bghgary, line 949: This should not be exposed. Not all interfaces are useful externally.

bghgary:
I also would recommend not exposing any interfaces for glTF extensions that are not ratified yet since they are subject to change.

sebavan:
> Since gltf2-interface is public, it might be a good idea to make the names of the interfaces more consistent. Maybe we should come up with a naming scheme?

They look good to me but I am not the best at names, so do not hesitate if you want some renames

> I also would recommend not exposing any interfaces for glTF extensions that are not ratified yet since they are subject to change. Or at least, we should mark them as such.

A comment would be enough in typedoc as exposing them is our path forward between exporters and loaders.

Popov72:
Comments added and metadata interface removed.

I also tried to rationalize the interface names, let me know what do you think.

sebavan:
It looks all perfect to me :-)

Popov72:
The unit tests of the serializers fail with something like:
image
The faulty js code in `postExportNodeAsync` is:
```typescript
const lightType = (
    babylonLight.getTypeID() == Light.LIGHTTYPEID_POINTLIGHT ? IKHRLightsPunctual_LightType.POINT : (
        babylonLight.getTypeID() == Light.LIGHTTYPEID_DIRECTIONALLIGHT ? IKHRLightsPunctual_LightType.DIRECTIONAL : (
            babylonLight.getTypeID() == Light.LIGHTTYPEID_SPOTLIGHT ? IKHRLightsPunctual_LightType.SPOT : null
        )));
```
It seems `IKHRLightsPunctual_LightType` is undefined (it is coming from `import { IKHRLightsPunctual_LightType, IKHRLightsPunctual_LightReference, IKHRLightsPunctual_Light, IKHRLightsPunctual } from "babylonjs-gltf2interface";`)...

When I execute the unit test code in a PG it does work, I don't get this error.

Any idea why it fails? Maybe it's a problem with the setup of the unit tests that does not make "babylonjs-gltf2interface" available to the test code?

RaananW:
> Any idea why it fails? Maybe it's a problem with the setup of the unit tests that does not make "babylonjs-gltf2interface" available to the test code?

This is just a guess (untested) but usually d.ts files are not compiled to javascript. you can use the values during compile time, but not during runtime, as they are not actually there.
Enums with values should be part of a compiled .ts file to be included in the build

dist/preview release/glTF2Interface/babylon.glTF2Interface.d.ts
```diff
@@ -905,4 +905,286 @@ declare module BABYLON.GLTF2 {
...
+        bufferView: number;
+        attributes: { [name: string]: number };
+    }
+
+    /**
+     * Interfaces from the KHR_lights_punctual extension
+     */
+
+    /** @hidden */
+    enum IKHRLightsPunctual_LightType {
```
bghgary, line 958: Try putting `const enum` here.

Popov72: Thanks, trying it!

sebavan: This should fix the tests... I guess :-)

RaananW: Or that :-)

sebavan:
Looks like it passed, I ll merge as soon as the rest of the pipeline is green and create the nightly :-)

Issue #5579, opened 2018-11-30
Add support for loading glb from a base64-encoded string or arraybuffer


bghgary:
See www.html5gamedevs.com/topic/41216-load-glb-model-into-scene-directly-from-content-string

major-mayer:
Is it correct, that it is only possible to load GLB files from a Base64 string?
I tried to do this with a base64 encoded stl file (using the FileReader API) and it didn't work...?

sebavan:
Ping @Popov72 which will be the best placed to answer :-)

Popov72:
The PR was only for glb.

You should post to the forum and provide a small PG so that we can have look for stl files.

bghgary:
@Popov72 Just noticed we only added support for base64. Are we adding array buffer support too?

sebavan:
@bghgary do we need array buffers ? Are createObjectURL not enough ? just wondering out loud :-)

bghgary:
I guess so. I can't remember why I put array buffer in the title.

sebavan:
Let s add it if it comes back then

keverw:
I was thinking array buffers would be useful over base64 due to the 33% overhead. I know on browser there is blob but if doing serverside stuff with nullEngine beign able to use array buffers would be better I think. Plus since you need the XHR polyfill in Node, this way could write my own code to read from the file system and load the mesh that way.

br-matt:
Array buffer would be much appreciated. Currently trying to figure how to convert array buffer glb response to base64 and then use `Tools.DecodeBase64` without much luck yet.

deltakosh:
pinging @bghgary

bghgary:
Are `File` objects available in these situations? If so, that's another workaround to be able to load from a buffer.

BABYLON.SceneLoader.LoadAsync("<root url>", new File([glbArrayBuffer], "Box.glb"));

If not, can you file a feature request separate from this one? We can track it and try to figure out a way to do this.

br-matt:
Thanks for the quick response ! It is available and I gave it a shot. 

![Screen Shot 2021-04-05 at 1 40 51 PM](https://user-images.githubusercontent.com/66705855/113605998-6b15c300-9615-11eb-9c2b-9ff0a7b5a493.png)

I got this after. I am not too familiar with glb headers. The file can be loaded normally via .glb and also array buffer -> base64. I will go ahead and open that soon unless you have ideas about the headers here

bghgary:
[...]

PR #8435, opened 2020-06-24
Added support for .glb file loading through a base64 encoded filename

Popov72:
See #5579 

It will make this PG work: https://playground.babylonjs.com/#7F6S08#15

I have checked that the `directLoad` signature change (it now returns a `Promise<any>` instead of a `any`) only impacts the glTF loader (not the .obj / .babylon ones.

### DIFF
--- a/loaders/src/glTF/glTFFileLoader.ts
+++ b/loaders/src/glTF/glTFFileLoader.ts
@@ -20,6 +20,7 @@ import { GLTFValidation } from './glTFValidation';
 import { Light } from 'babylonjs/Lights/light';
 import { TransformNode } from 'babylonjs/Meshes/transformNode';
 import { RequestFileError } from 'babylonjs/Misc/fileTools';
+import { StringTools } from 'babylonjs/Misc/stringTools';
 
 interface IFileRequestInfo extends IFileRequest {
     _lengthComputable?: boolean;
@@ -456,6 +457,8 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
     private _progressCallback?: (event: ISceneLoaderProgressEvent) => void;
     private _requests = new Array<IFileRequestInfo>();
 
+    private static magicBase64Encoded = "Z2xURg"; // "glTF" base64 encoded (without the quotes!)
+
     /**
      * Name of the loader ("gltf")
      */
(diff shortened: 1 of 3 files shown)


sebavan:
Looks like a nice start !!!! adding @bghgary for review

loaders/src/glTF/glTFFileLoader.ts
```diff
@@ -634,13 +637,24 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
 
     /** @hidden */
     public canDirectLoad(data: string): boolean {
-        return data.indexOf("asset") !== -1 && data.indexOf("version") !== -1;
+        return (data.indexOf("asset") !== -1 && data.indexOf("version") !== -1) ||
+                StringTools.StartsWith(data, "data:base64," + GLTFFileLoader.magicBase64Encoded);
```
bghgary, line 641: There may also be a mime type in for a [data URI](https://en.wikipedia.org/wiki/Data_URI_scheme#Syntax)? Do we need to handle them?

src/Loading/sceneLoader.ts
```diff
@@ -123,7 +123,7 @@ export interface ISceneLoaderPluginBase {
      * @param data string containing the data
      * @returns data to pass to the plugin
      */
-    directLoad?(scene: Scene, data: string): any;
+    directLoad?(scene: Scene, data: string): Promise<any>;
```
bghgary, line 126:
```suggestion
    directLoadAsync?(scene: Scene, data: string): Promise<any>;
```

I'm not sure this is okay to change, but typically, our async functions that return a Promise have a suffix of `Async`.

src/Loading/sceneLoader.ts
```diff
@@ -417,7 +417,13 @@ export class SceneLoader {
         SceneLoader.OnPluginActivatedObservable.notifyObservers(plugin);
 
         if (directLoad) {
-            onSuccess(plugin, plugin.directLoad ? plugin.directLoad(scene, directLoad) : directLoad);
+            if (plugin.directLoad) {
+                plugin.directLoad(scene, directLoad).then((data) => {
```
bghgary, line 421: You should also handle errors and call the onError callback.

loaders/src/glTF/glTFFileLoader.ts
Popov72: After decoding, the base64 string should be binary data: so, should we support something other than "application/octet-stream"?

If not, I can add the check for it, else I would need some more info on what should be supported (and possibly how).

src/Loading/sceneLoader.ts
Popov72: @sebavan ?

src/Loading/sceneLoader.ts
Popov72: Done.

loaders/src/glTF/glTFFileLoader.ts
bghgary: GLB has its own mime type, see https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#mime-type. I think we should support this and `application/octet-stream`?

One other random thought. Should we allow a glTF JSON encoded as base64 here? I say no, but just wanted to call it out.

Popov72: Check of mime types added.

I don't think we should allow glTF JSON encoded base64 here either, @sebavan to decide.

src/Loading/sceneLoader.ts
sebavan: I would prefer to not change the signature if possible to prevent any regression

loaders/src/glTF/glTFFileLoader.ts
sebavan: As you both vote for no. I ll definitely go with you :-) I have no real preferences either.

Popov72:
Let me know if I need to do something else regarding this PR.

src/Loading/sceneLoader.ts
bghgary: It already is changing the signature, so there isn't really a difference.

PR #8438, opened 2020-06-24
Timeline improvements

toledoal:
Hello! 😃 

This PR adds improvements to the timeline and continues with design redlines implementation.

1. Adds the frame per second for the current animation (Visual clue adds in viewer)
2. Sets loop mode of selected animation


inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -211,9 +214,9 @@ export class AnimationCurveEditorComponent extends React.Component<
     }
 
     let valueLines = Math.round((this.state.scale * this._heightScale) / 10);
-    console.log(highestFrame);
+    console.log(length);
```
sebavan, line 217: not sure we should keep the log

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -1382,6 +1391,14 @@ export class AnimationCurveEditorComponent extends React.Component<
     }
   }
 
+  getCurrentFrame(frame: number) {
```
sebavan, line 1394: maybe should be renamed to isCurrentFrame ? as it does not return the current frame but only if we are on it

inspector/src/components/actionTabs/tabs/propertyGrids/animations/graphActionsBar.tsx
```diff
@@ -56,7 +56,7 @@ export class GraphActionsBar extends React.Component<IGraphActionsBarProps> {
             onClick={this.props.addKeyframe}
           />
           <IconButtonLineComponent
-            tooltip={'Remove Keyframe'}
+            tooltip={'Frame'}
```
sebavan, line 59: Wonder if we should change the tooltip as it looks less descriptive.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
```diff
@@ -159,27 +246,133 @@ export class Timeline extends React.Component<
     this.setState({ activeKeyframe: null });
   }
 
+  private active: string = '';
```
sebavan, line 249: Could we replace with _active

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
```diff
@@ -159,27 +246,133 @@ export class Timeline extends React.Component<
...
 
+  private active: string = '';
+
   scrollDragStart(e: React.TouchEvent<HTMLDivElement>): void;
   scrollDragStart(e: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
   scrollDragStart(e: any) {
     e.preventDefault();
-    if ((e.target.class = 'scrollbar') && this._scrollbarHandle.current) {
-      this._scrolling = true;
+    if (e.target.className === 'scrollbar') {
```
sebavan, line 255: The class check looks redundant

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: good point. I'll rename it

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: yes, that make sense. I'll remove it

inspector/src/components/actionTabs/tabs/propertyGrids/animations/graphActionsBar.tsx
toledoal: Yes, the idea for this is to "frame" the selected keyframes... I guess we could have "Frame selected keyframes"

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
toledoal: yes, makes sense... all privates should have underscore... _ :)

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
toledoal: let me check again. But I remember adding this because I'm using a container to set dragging events but I'm actually moving a child...

toledoal: yes, that was redundant 👍 , I removed it , thanks for the review.

toledoal:
About the refactoring. Yes,  I need to to a bunch of things about it. Mainly for the main component. I just want to create specific PR for this.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -197,7 +200,7 @@ export class AnimationCurveEditorComponent extends React.Component<
 
   setAxesLength() {
     let length = 20;
-    let newlength = Math.round(this._canvasLength * this.state.scale); // Check Undefined, or NaN
```
RaananW, line 200: Was this comment addressed in any way?

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -1382,6 +1391,14 @@ export class AnimationCurveEditorComponent extends React.Component<
     }
   }
 
+  isCurrentFrame(frame: number) {
+    if (this.state.currentFrame === frame) {
```
RaananW, line 1395: (nit) `return this.state.currentFrame === frame` will be cleaner

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
```diff
@@ -7,31 +7,87 @@ interface ITimelineProps {
...
+        const scrollFrameLength = end - start;
+        const widthPercentage = (scrollFrameLength * 100) / scrollFrameLimit;
+        const scrollPixelWidth = (widthPercentage * containerWidth) / 100;
+        // check 0 values here... to avoid Infinity
+        if (scrollPixelWidth === Infinity) {
+          console.log('error infinity');
+        }
+        return scrollPixelWidth;
+      } else {
+        return undefined;
```
RaananW, line 86: not needed (same as the next return undefined)

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
```diff
@@ -163,23 +251,125 @@ export class Timeline extends React.Component<
...
+      this._active = 'rightDraggable';
       this._shiftX =
         e.clientX - this._scrollbarHandle.current.getBoundingClientRect().left;
-      this._scrollbarHandle.current.style.left = e.pageX - this._shiftX + 'px';
     }
   }
 
   scrollDrag(e: React.TouchEvent<HTMLDivElement>): void;
   scrollDrag(e: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
   scrollDrag(e: any) {
```
RaananW, line 286: Turned out to be quite a function :-)

Is there a way to separate it to smaller functions just to keep the function a bit smaller? Just if you see a nice and clean way to do that.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: ah, yes. I need to remove it. The problem was actually happening because I wasn't rounding the value to int. So the array would break. The Math.rounds resolved the issue. Thanks for noticing this :)

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: gotcha!

inspector/src/components/actionTabs/tabs/propertyGrids/animations/timeline.tsx
toledoal: sure I can. I can split it in three. 1. Handle scrollbar dragging 2. Handle left resize 3.Handle right resize

sebavan:
Looks Better to me @RaananW  ???

RaananW:
> Looks Better to me @RaananW ???

LGTM

PR #8455, opened 2020-06-27
Prepass for ssao

CraigFeldspar:
This is a work in progress

Uses the newly implemented prepass renderer to fasten the rendering of SSAO for WebGL 2.
Basically uses multi render target to avoid rendering the scene geometry twice.

For now the code is very ugly and hacky, but it's a proof that we can quickly switch ssao2 to prepass renderer with the same visual quality.
Performance : when forcing integrated graphics on my laptop, I get a 16 FPS difference (from 26 FPS to 42 FPS) on the following playground : 
http://localhost:1338/Playground/index-local.html#PQKG24


sebavan:
It is a pretty nice start !!! and I like the perf gain we can reach !!!

sebavan:
LGTM let s do a thorough review and Then Road to Deferred !!!!

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -679,7 +679,8 @@ export abstract class PBRBaseMaterial extends PushMaterial {
      * Should this material render to several textures at once
      */
     public get shouldRenderToMRT() {
-        return this.subSurface.isScatteringEnabled;
+        return this.subSurface.isScatteringEnabled ||
```
sebavan, line 682: Do you think materialShouldRenderGeometry could be enough so the test is simpler with SSS ?

src/Materials/material.ts
```diff
@@ -613,7 +613,7 @@ export class Material implements IAnimatable {
     /**
      * Stores a reference to the scene
      */
-    private _scene: Scene;
+    protected _scene: Scene;
```
sebavan, line 616: not sure we should change it ? Everywhere else we rely on material.getScene()

src/PostProcesses/RenderPipeline/Pipelines/ssao2RenderingPipeline.ts
```diff
@@ -388,8 +423,12 @@ export class SSAO2RenderingPipeline extends PostProcessRenderPipeline {
             effect.setFloat("yViewport", Math.tan(this._scene.activeCamera.fov / 2));
             effect.setMatrix("projection", this._scene.getProjectionMatrix());
 
-            effect.setTexture("textureSampler", this._depthTexture);
-            effect.setTexture("normalSampler", this._normalTexture);
+            if (this._forceGeometryBuffer) {
+                effect.setTexture("depthSampler", this._scene.enableGeometryBufferRenderer()!.getGBuffer().textures[0]);
+                effect.setTexture("normalSampler", this._scene.enableGeometryBufferRenderer()!.getGBuffer().textures[1]);
+            } else {
+                effect.setTexture("depthNormalSampler", this._prePassRenderer.prePassRT.textures[2]);
```
sebavan, line 430: I wonder if we could add constants for the indices in the MRT as part of BABYLON.Constants ??? those gets inlined during build ;-)

src/PostProcesses/RenderPipeline/Pipelines/ssao2RenderingPipeline.ts
```diff
@@ -292,7 +306,11 @@ export class SSAO2RenderingPipeline extends PostProcessRenderPipeline {
             effect.setFloat("near", this._scene.activeCamera.minZ);
             effect.setFloat("far", this._scene.activeCamera.maxZ);
             effect.setFloat("radius", this.radius);
-            effect.setTexture("depthSampler", this._depthTexture);
+            if (this._forceGeometryBuffer) {
+                effect.setTexture("depthNormalSampler", this._scene.enableGeometryBufferRenderer()!.getGBuffer().textures[0]);
+            } else {
+                effect.setTexture("depthNormalSampler", this._prePassRenderer.prePassRT.textures[2]);
```
sebavan, line 312: I wonder if we could add constants for the indices in the MRT as part of BABYLON.Constants ??? those gets inlined during build ;-)

src/PostProcesses/RenderPipeline/Pipelines/ssao2RenderingPipeline.ts
```diff
@@ -278,11 +288,15 @@ export class SSAO2RenderingPipeline extends PostProcessRenderPipeline {
...
```
sebavan, line 294: I wonder if we could add constants for the indices in the MRT as part of BABYLON.Constants ??? those gets inlined during build ;-)

src/Materials/PBR/pbrBaseMaterial.ts
CraigFeldspar: actually with subsurface we need irradiance, and with ssao we need geometry information
Your comment is still valid though, for harmonization I will replace this by a boolean named `materialsShouldRenderIrradiance`

src/Engines/constants.ts
```diff
@@ -484,4 +484,24 @@ export class Constants {
...
+     */
+    public static readonly PREPASS_IRRADIANCE_INDEX = 1;
+
+    /**
+     * Prepass texture index for depth + normal
+     */
+    public static readonly PREPASS_DEPTHNORMAL_INDEX = 2;
+
+    /**
+     * Prepass texture index for depth + normal
```
sebavan, line 504: depth + normal should be albedo ???

sebavan:
Could you revert the protected _scene: Scene; in material and rely on material.getScene() ?

sebavan:
Then it is a go !!!

src/Engines/constants.ts
CraigFeldspar: my bad :)

CraigFeldspar:
> Could you revert the protected _scene: Scene; in material and rely on material.getScene() ?

If we are talking about the same thing, it was already reverted in 3f7e31f :)
Or I may have missed something else ?

sebavan:
Just added @deltakosh for final review, thanks a ton @CraigFeldspar

PR #8464, opened 2020-06-30
Panning Fixes

toledoal:
This draft PR continues to solve major issues of the Curve Editor.

I'll be adding more commits to this draft PR.

Will edit this message when PR is ready to submit 😄


sebavan:
@toledoal I just tried this on top of the latest changes locally and It does not look like working better :-( I am actually not sure of what it fixes.

Let me summarize the issues to address by priorities (some are new and probably regressions):

I would suggest to address them one at a time on separate PRs as currently the entire editor is not usable.

1. the bottom slider does not move/scale the window but the map which should stay full scale to always see all the keys. It also flickers a lot on resize.

2. play head is not dragable and acting weird. It looks like I can only move it by clicking the frame I want.

3. dragging does not work if you exit the dragable zone while dragging on all dragable areas

4. dragging keys does not change the shape of the curves

5. manually editing values is impossible

6. by trying to delete points, I obtained this shape:
image

7. by trying to add some keys, they always add on 0 1 2 3... despite choosing key position 0 7 15 17:
image

8. Jumping from key frames (with forward backward) does not change the current view position so we kind of lost track of where we are.

9. the curve shape controls are not working and I was not able to go from tangents to the other possibilities.

10. there is no cancel button when trying to add a new animation

11. in the new animation window, the target is always empty

12. it is possible to end up with pretty weird curves:
image

13. I am still seeing a lot of errors in the console while using the tool:
image

I guess these are all the major ones I have seen so far, but I might see more when I can start using the editor.

On the bright part the visual looks great :-) and I hope they are all related to the same underlying issue or to my editor beeing in a weird state.

toledoal:
Sure, makes sense to address these ones one by one. I'll be working on the first one 👍

PR #8470, opened 2020-07-02
Playhead drag bug

toledoal:
Hello! This PR implements the Playhead drag /left/right

Next PR will focus on moving the canvas alongside the Playhead/current frame. 
Right now we lose focus on the Playhead if the current frame is beyond showing canvas.


sebavan:
Good to go, NEXT !!!! :-)

PR #8473, opened 2020-07-02
WIP Texture editor

DarraghBurkeMS:
Creates the foundation of the texture editor. Supports opening a pop-out window in the inspector for viewing int-type textures (no float support). Supports pan and zoom with middle mouse button and mouse wheel.

Still need to:
* Configure build properly
* clean up code

### DIFF
--- a/Tools/Config/config.json
+++ b/Tools/Config/config.json
@@ -47,7 +47,8 @@
         "serializers",
         "gui",
         "inspector",
-        "nodeEditor"
+        "nodeEditor",
+        "textureEditor"
     ],
     "es6modules": [
         "core",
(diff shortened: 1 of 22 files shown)


sebavan:
That is a pretty nice start !!!! my only concern is that I do not think it should be a separate package. This will not be a standalone app so The code could directly be developed within the inspector I guess like we do for curve editor.

@PatrickRyanMS  Can you confirm ?

PatrickRyanMS:
Yes, @sebavan, this should be invoked from the inspector like the curve editor or node editor so that we can retain a consistent UX with the editors.

DarraghBurkeMS:
Refactored to live inside inspector package. Closing PR.

Issue #8478, opened 2020-07-03
Animation Curve Editor List of Bugs/Features


toledoal:
Hello everybody! 😃 

#This is the current list of bugs/features that will be addressing for the Animation Curve Editor

- [x]  ~Bug (Flickering Scrollbar on resize)~
- [x] ~fps input doesn't update the frames per second property in animation~
- [x] ~when open animation to edit, the scale of animation is compressed~
- [x] ~when opening an animation the animation is not stopped~
- [x]  ~The Playhead is not draggable~
- [x]  ~The Playhead behavior is strange, the draggable area appears to be small~
- [x] ~Dragging keyframes should have limits to left/right.~
- [x] ~Dragging keyframes not working as expected~
- [x] ~Adding/Deleting keyframes produces bug on curve's path~
- [x] ~Stop animation after clicking Playhead.~
- [x] ~Stop animation on Loop button click~
- [x] ~Remove Size animation type on create~
- [x] ~Graph action controls hide sometimes. Should be visible all times~
- [x] ~Value box should be empty and disabled if no key frames are selected~
- [x] ~Frame Input should not reposition Playhead.~
- [x] ~Value box is not working (see https://github.com/BabylonJS/Babylon.js/pull/8515#issuecomment-656225459).~
- [x] ~Connect value and frame boxes to Keyframe dragging.~
- [x] ~Value scale is wrong and cannot change scale for types (int, float, color)~
- [x] ~Test font loading in Edge doesn't load sometimes~
- [x] ~Play animation in editor behaves strangely (probably fps related)~
- [x] ~Key frame should be able to be dragged back to 0 if it is the first one~
- [x] ~When adding a new keyframe, add value between two points in a curve~
- [x] ~You cannot delete a key with the 'delete' key, only the 'backspace' key. It should probably be both?~
- [x] ~When moving the tangent handles, they seem to be limited and cannot be raised or lowered past a certain point? I expected to be able to move a tangent to any height and it not be limited.~
- [x]  ~Broke Mode on/off. When moving control point the other distorts~

**Items left for initial Scope**
- [x] ~Dragging the playhead appears to have no connection to the scene? If I scrub the playhead in the curve editor, shouldn't the object being animated update in the playground?~
- [x]  ~Dragging the canvas around feels sluggish and stuttery~
[...]

PR #8479, opened 2020-07-03
Curve Editor Bugs

toledoal:
#8478 

Hello! This PR provides a Fix 

1. open animation to edit, the scale of animation is compressed
2. when opening an animation the animation is not stopped
3. The Fps input doesn't update the frames per second property in animation


sebavan:
I am waiting only on the build step before merging

PR #8490, opened 2020-07-06
fix bug

toledoal:
Hello! 😄  This PR addresses the bugs

1. The Playhead is not draggable.
2. The Playhead behavior is strange, the draggable area appears to be small.

 In issue #8478


sebavan:
Thanks I ll do a nightly now so we can test :-)

PR #8498, opened 2020-07-07
Optimize mesh.forceSharedVertices for speed

Popov72:
See https://forum.babylonjs.com/t/smooth-a-large-mesh/12284 and https://playground.babylonjs.com/#JXIJQM

It takes 2mn30s on my computer. After the changes, it takes less than 0.5s.


Popov72:
It's more moving to a o(n) search for each vertex to o(1) (array.indexOf() replaced by a lookup in a map).

PR #8513, opened 2020-07-08
Implement glTF serialization of babylon skeletons

Drigax:
This PR makes several changes to the glTF serialization logic to allow for export of Babylon skeletons and mesh skinning information.

1)Add references of ISkin to in-memory representation of gltf scene to be exported
2)Refactored mesh attribute export logic in `writeAttributeData` via adding support for serialization of non-float mesh attributes.
3)Added `MatricesIndices` and `MatricesWeights` vertex buffer attribute export logic.
4)refactoring of `createBufferViewKind` to allow for proper calculation of buffer view size given attribute type.

Partial fix of #5100


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -729,14 +738,57 @@ export class _Exporter {
...
+                    vertexAttributes.push(vertexData.asArray());
+                }
+                break;
+            }
+            case VertexBuffer.MatricesWeightsKind:
+            case VertexBuffer.MatricesWeightsExtraKind: {
+                for (let k = 0, length = meshAttributeArray.length / stride; k < length; ++k) {
+                    index = k * stride;
+                    const vertexData = Vector4.FromArray(meshAttributeArray, index);
+                    vertexData.normalize();
```
Drigax, line 755: Current outstanding issue is that currently exporting test scene: https://playground.babylonjs.com/#T087A8#12

Exports with errors related to non-normalized skinning weights...
I added this `normalize()` call @L755 to address this, but the error persists. I'll continue to investigate....

Drigax: Duh. I'm currently normalizing a vec4 using euclidean distance... we just want all the members to add up to 1. I should have a fix shortly...

Drigax: fixed.

Drigax:
Sure, I'll try to add a test that just does skinning, probably roundtripping dude, or a mixamo rig

serializers/src/glTF/2.0/glTFUtilities.ts
```diff
@@ -193,4 +193,22 @@ export class _GLTFUtilities {
...
     }
+
+    public static _NormalizeMatricesWeightsFromRef(matricesWeights: Vector4) {
+        const sum = matricesWeights.x + matricesWeights.y + matricesWeights.z + matricesWeights.w;
+        if (sum > 0) {
+            matricesWeights.scaleInPlace(1 / sum);
+        }
+    }
+
+    public static _GetRightHandedMatrixFromRef(matrix: Matrix) {
```
Drigax, line 204: I'm not too sure how to derive this change, I came to this result visually after noticing that `Matrix.toggleModelMatrixHandInPlace`
https://github.com/BabylonJS/Babylon.js/blob/a11d765c5dfa6d6db2e3945ac84389a22d2ca9d1/src/Maths/math.vector.ts#L4327-L4338

inverted m[9] and m[14] of my output inverse bind matrix, so I created this custom function to handle the matrix handedness converstion, but I'm not sure if this is 100% correct...

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -729,14 +738,57 @@ export class _Exporter {
...
+            case VertexBuffer.MatricesIndicesExtraKind: {
+                for (let k = 0, length = meshAttributeArray.length / stride; k < length; ++k) {
+                    index = k * stride;
+                    const vertexData = Vector4.FromArray(meshAttributeArray, index);
+                    vertexAttributes.push(vertexData.asArray());
+                }
+                break;
+            }
+            case VertexBuffer.MatricesWeightsKind:
+            case VertexBuffer.MatricesWeightsExtraKind: {
```
Drigax, line 751: Upon further thought, I realize that this structure won't quite work for export of matrix weights, since we're dependent on both the MatricesWeights and MatricesWeightsExtra vertex attributes inorder to make sure that the joint weightings sum to 1. I think I'll have to handle joint weights as a special case, outside of the current attribute export structure...

Drigax:
Added a visual test, going to look into an elegant way to handle 5+ bone skinning export

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1562,6 +1652,59 @@ export class _Exporter {
...
+     * @param nodes Babylon transform nodes
+     * @param binaryWriter Buffer to write binary data to
+     * @returns Node mapping of unique id to index
+     */
+    private createSkinsAsync(babylonScene: Scene, nodeMap: { [key: number]: number }, binaryWriter: _BinaryWriter): Promise<{ [key: number]: number }> {
+        let promiseChain = Promise.resolve();
+        const skinMap: { [key: number]: number } = {};
+        for (let skeleton of babylonScene.skeletons) {
+            // create skin
+            const skin: ISkin = { joints: []};
```
bghgary, line 1667: nit: there are some minor spacing issues, might want to auto format the file.

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1562,6 +1652,59 @@ export class _Exporter {
...
+                        if (this._convertToRightHandedSystem) {
+                            _GLTFUtilities._GetRightHandedMatrixFromRef(boneMatrix);
+                        }
+                        inverseBindMatrices.push(boneMatrix);
+                        skin.joints.push(nodeMap[transformNode.uniqueId]);
+                    }
+                }
+            }
+            // create buffer view for inverse bind matrices
+            let byteStride = 64; // 4 x 4 matrix of 32 bit float
```
bghgary, line 1685: nit: use `const` instead of `let` when possible

PR #8515, opened 2020-07-09
Stop animation and other bugs

toledoal:
Hi! 😃 

This PR addresses the following bugs in issue #8478 

1. Frame Input should not reposition Playhead. (The input now indicates the frame to put a keyframe or edit the value)
2. Connect the value box to Playhead change. (When value changes in value input the keyframe updates its value)

The following even though they stop the animation because the bug on playing the animation still feels weird. So I think the best option is to fix the play animation bug in the next PR.

3. Stop animation on Loop button click (The animation stops on clicking the loop button or changing frames per second. 
The weird play animation bug is still there.
4. Stop animation after clicking Playhead. (The weird play animation bug is still there)

These we low hanging fruits to solve.

5. Remove Size animation type on "create a new animation". (Also reorder the list alphabetically)
6. Graph action controls hide sometimes. Should be visible all times (Used to be hidden, now only disabled if there is not animation selected)

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -9,7 +9,7 @@
 - Reflection probes can now be used to give accurate shading with PBR ([CraigFeldpsar](https://github.com/craigfeldspar) and ([Sebavan](https://github.com/sebavan/)))
 - Added SubSurfaceScattering on PBR materials ([CraigFeldpsar](https://github.com/craigfeldspar) and ([Sebavan](https://github.com/sebavan/)))
 - Added editing of PBR materials, Post processes and Particle fragment shaders in the node material editor ([Popov72](https://github.com/Popov72))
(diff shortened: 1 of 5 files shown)


sebavan:
@toledoal The value input still does not work... Basically in edit putting a value of 

We can not delete the content to put another value like going from 1 to 2 ->
image

It does not display the current value (currently 1 where curve shows 5) ->
image

Changing the value does not change the value to the requested one ->
image

I ll put this back into the list. You should test those behavior before the next PR in edit as well.

toledoal:
Ok. Thanks. I'll verify this. The value scale is currently wrong. I'll work on that bug as well. The problem with this particular bug is it actually connected to other issues.

sebavan:
Makes much more sense so the main issue becomes the deletion and the fact that it still tries to change value on playhead position at start

PR #8518, opened 2020-07-09
Alternate option to create the edges in the edge renderer

Popov72:
While trying to improve the rendering of https://forum.babylonjs.com/t/drawing-outside-edges-of-a-mesh-created-from-csg/11926/9, I ended up writing a new function to create the edges of a mesh that happens to be a bit faster than the existing one.

Here are some testings:
| Asset | Existing func | New func | Speed up|
|-------|---------------|------------|-----------|
| dude.babylon | 0.2s | 0.06s | x3 |
| SSAOcat.babylon | 0.46s | 0.14s | x3 |
| StanfordBunny.obj | 0.2s | 0.015s | x13 |
| ufo.glb | 1.4s | 0.1s | x14 |
| skull.babylon | 120s | 0.25s | x480 |

The display is exactly the same with the new and the existing algorithm in all those tests.

Using the new algorithm is an option, however, as there may be some cases which would not lead to the same rendering...

I have also added some options (that don't exist with the legacy algorithm) that allow to improve the creation in some cases, notably when using CSG to create meshes, which can produces cases like this:
image
There are two triangles, ABC and CDE.

AC is not shared by another triangle, so the renderer considers it to be an edge and draws it.

For this to be rendered correctly by the edges renderer, the triangles EBC and ABE should be created and ABC removed.

The `applyTessellation` option allows to detect such cases and create additional triangles to deal with that. Activating this option is quite CPU intensive, so it is off by default (also, in normal circumstances, you don't need this option).

Using this option will make this PG to render correctly: https://playground.babylonjs.com/#0HA29N

With the existing algorithm:
image

With the new algorithm + `applyTessellation = true` + `useFastVertexMerger = false` (see the doc for this one):
image


sebavan:
Should we even keep the old one ? what are the downsides ?

Popov72:
I don't know...

As the algorithm is not the same, it may not produce the exact same result in all cases, that's why I didn't want to remove it completely.

src/Rendering/edgesRenderer.ts
```diff
@@ -348,6 +395,308 @@ export class EdgesRenderer implements IEdgesRenderer {
...
+                let p2Index = remapVertexIndices[indices[index + (i + 2) % 3]];
+
+                if (p0Index === p1Index) { continue; }
+
+                TmpVectors.Vector3[0].copyFromFloats(positions[p0Index * 3 + 0], positions[p0Index * 3 + 1], positions[p0Index * 3 + 2]);
+                TmpVectors.Vector3[1].copyFromFloats(positions[p1Index * 3 + 0], positions[p1Index * 3 + 1], positions[p1Index * 3 + 2]);
+                TmpVectors.Vector3[2].copyFromFloats(positions[p2Index * 3 + 0], positions[p2Index * 3 + 1], positions[p2Index * 3 + 2]);
+
+                if (!faceNormal) {
+                    faceNormal = Vector3.Cross(TmpVectors.Vector3[1].subtract(TmpVectors.Vector3[0]), TmpVectors.Vector3[2].subtract(TmpVectors.Vector3[1]));
```
sebavan, line 641: substractToRef ?

deltakosh:
LGTM. let's keep the old one but use the new one by default.

If there is no issue in the next couple of months we removed it

Deel96:
Hello,
I'm the poster of the thread mentioned above.
I'm impressed how fast the process from question asking / "bug" reporting to a pull request is.

Is it possible to give a timeline when this features gets released outside of the alpha version?
Big thank you

deltakosh:
probably in less than 2hours :D

PR #8523, opened 2020-07-10
Comment clarification

BabylonJSGuide:
The comment, as it stands, referring to _between the two given vectors_ does not describe correctly the purpose of the method. It has, and does, cause confusion, for example https://www.babylonjs-playground.com/#0RYA8I#4 from https://forum.babylonjs.com/t/rotation-of-connected-object/12043/11?u=johnk

After discussion that led to me clearing up my own confusion, https://forum.babylonjs.com/t/angle-object-anglebetweentwopoints-is-wrong/12348/9?u=johnk I suggest the PR to change the comment only.

### DIFF
--- a/src/Maths/math.path.ts
+++ b/src/Maths/math.path.ts
@@ -84,9 +84,9 @@ export class Angle {
     }
 
     /**
-     * Gets a new Angle object valued with the angle value in radians between the two given vectors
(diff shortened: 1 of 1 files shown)


deltakosh:
Totally fine for me

Thanks!

PR #8529, opened 2020-07-10
Texture inspector

DarraghBurkeMS:
This PR adds the ability to view individual RGBA channels, and fixes the bug where textures would be flipped vertically in the editor. All tooling code has been removed so it is purely inspecting features.

~~Adds editing capabilities to the texture editor. Tools can be loaded from external Javascript files and are able to add their own keyboard/pointer observables, as well as draw on the texture using the 2D canvas API (no WebGL context is provided yet).~~

~~I have created a couple of tools here:~~

~~1. https://darraghburkems.github.io/BJSTools/Floodfill.js~~
~~2. https://darraghburkems.github.io/BJSTools/Paintbrush.js~~

~~Looking for feedback on the tool loading mechanism. My main questions:~~

~~1. Does it make sense to put the tool state in the React component? I felt that this was cleaner. Ideally I'd like to move as much state as possible into the component and away from the manager (which is imperative rather than declarative).~~
~~2. Is using eval a smart way to add tools at runtime? It does bring security and performance concerns, but I'm not sure if there is a superior alternative.~~


deltakosh:
What is the rationale being loading tools from url? inspector can be embed in places where there is no external network so I would recommend to have all the tools available out of the box.

Unless you consider them as plugins? that community can add on their own and we do not distribute them?

deltakosh:
Also one additional comments: let's finish the first version of that PR, merged it so community can test it and then we can add more features

DarraghBurkeMS:
> What is the rationale being loading tools from url? inspector can be embed in places where there is no external network so I would recommend to have all the tools available out of the box.
> 
> Unless you consider them as plugins? that community can add on their own and we do not distribute them?

The idea that we had in mind was that the community would be free to develop and share their own tools. Since it is hard for us to envision everything that users may want to do with the editor, we thought it would be better to give people the ability to create their own. I was also hoping to put together a tutorial later on for tool creation.

We thought we might provide a set of base tools that would be loaded every time, and then give users the flexibility to load in whichever other tools they may need.

deltakosh:
So here is my recommendation:
- Connect with @PatrickRyanMS to do a design review
- Remove all the external tools (for now)
- Make sure we can merge that PR for patrick to test
- Implement the design
- Enjoy :)

deltakosh:
> > What is the rationale being loading tools from url? inspector can be embed in places where there is no external network so I would recommend to have all the tools available out of the box.
> > Unless you consider them as plugins? that community can add on their own and we do not distribute them?
> 
> The idea that we had in mind was that the community would be free to develop and share their own tools. Since it is hard for us to envision everything that users may want to do with the editor, we thought it would be better to give people the ability to create their own. I was also hoping to put together a tutorial later on for tool creation.
> 
> We thought we might provide a set of base tools that would be loaded every time, and then give users the flexibility to load in whichever other tools they may need.

It makes sense but I wonder if we want it to be registered from code or from the UI. From a security standpoint it stresses me a bit

sebavan:
Actually we can not have eval in the code at all due to use in some apps that will fail there security checks. Registering by code would be better.

deltakosh:
yeah that was my fear.. I remember having to remove eval from the code ;)

deltakosh:
all good to merge
do you mind to remove the draft mode?

PR #8531, opened 2020-07-13
Value curve

toledoal:
Hi! 😃 

This PR has fixes for the following bugs in #8478 

1. Value box should be empty and disabled if no key frames are selected
2. The value box is not working. 
3. Connect value and frame boxes to Keyframe dragging. *When dragging the selected keyframe we update the values in frame and value input boxes. *We still have bug "Key frame should be able to be dragged back to 0 if it is the first one"

3. Frame Input should not reposition Playhead. *We need to decide what to do when: a) what is the default Playhead frame, currently starts at frame 0. We don't reposition Playhead on keyframe value/frame change.


inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -38,8 +38,8 @@ interface ICanvasAxis {
 }
 
 export interface IActionableKeyFrame {
-  frame: number;
-  value: any;
+  frame?: number | string;
```
sebavan, line 41: Why would a frame be a string here ?

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -38,8 +38,8 @@ interface ICanvasAxis {
 }
 
 export interface IActionableKeyFrame {
-  frame: number;
-  value: any;
+  frame?: number | string;
+  value?: any;
```
sebavan, line 42: This should be a number

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -296,24 +296,50 @@ export class AnimationCurveEditorComponent extends React.Component<
     }
   }
 
+  getKeyframeValueFromAnimation(id: string) {
+    const animation = this.state.selected as Animation;
+    const index = parseInt(id.split('_')[3]);
```
sebavan, line 301: Those constructs look error prone ?

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: These properties are the ones connected to the input box. So this is for the case of an empty/disabled input box, so we can have an empty value, if not we force the input to always have a 0.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: I have an SVG curve id created based on the coordinate, the key index, and the name. This is to have a way to match the Keyframe in the animation and the SVG representation. Let me think of another way to sync those...

sebavan: If not possible it is fine if you create it as well and this is not a client input. but in this case, maybe adding a helper encode/decode would make it more obvious and less subject to maintenance issues.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
sebavan: this is already number | undefined with the question mark. So I think you might not need the string if undefined is used in case of empty or invalid string ?

toledoal: There is an update I'll do to this after a review today.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
toledoal: Ah. Great idea! I'm using this in a couple of places. So, yes, this is a benefit. I'll create that method/function

toledoal:
I think I'm going to need more time fixing this 😃

toledoal:
Hello! So, I finally was able to add the correct behavior for these inputs. Only upon the 'enter' key we now update the values. If the user goes away from the inputs we disabled them. When updating the value on each frame or value and enter key, we update the value on thus the selected keyframe.

sebavan:
Could you also do the value change on the blur event ???

toledoal:
Ah, good option. Yes. I think I need to verify if the value is not empty. 😃

sebavan:
LGTM I ll merge into

PR #8540, opened 2020-07-15
Gizmo Node support

CedricGuillemet:
Gizmo support for Node (TransformNode, Bone, Mesh,...)
The principle is to work on the world matrix in the gizmo, then, put it in parent space (if any parent) and update local matrix for bones or position/rotation/scaling for TransformNode. This is done in ```_matrixChanged``` function.

I didn't update boundingBoxGizmo nor gizmoManager as I don't think it makes sens outside of meshes. Let me know what you think.

Interface is the same with added property called ```attachedNode```


deltakosh:
Some tests are failing Burt else LGTM

src/Gizmos/axisDragGizmo.ts
```diff
@@ -3,7 +3,7 @@ import { Nullable } from "../types";
 import { PointerInfo } from "../Events/pointerEvents";
 import { Vector3, Matrix } from "../Maths/math.vector";
 import { TransformNode } from "../Meshes/transformNode";
-import { AbstractMesh } from "../Meshes/abstractMesh";
+import { Node } from "..";
```
sebavan, line 6: Should probably need the ../node

src/Gizmos/positionGizmo.ts
```diff
@@ -4,6 +4,7 @@ import { Nullable } from "../types";
 import { Vector3 } from "../Maths/math.vector";
 import { Color3 } from '../Maths/math.color';
 import { AbstractMesh } from "../Meshes/abstractMesh";
+import { Node } from "..";
```
sebavan, line 7: same here

PR #8542, opened 2020-07-16
Added support for the alpha component to the SSR post-process

Popov72:
See https://forum.babylonjs.com/t/ssr-does-not-support-transparent-background/12432


julien-moreau:
You rox !!

PR #8554, opened 2020-07-20
Value Axis

toledoal:
Hello! 😃  
This PR has a fix for the correct values on the Value axis bug "Value scale is wrong and cannot change scale for types (int, float, color)." in #8478. 

I still need to work on the scale types. Currently, it only works for the float type. So this PR only addresses that part.


inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -226,17 +226,19 @@ export class AnimationCurveEditorComponent extends React.Component<
   }
 
   setValueLines() {
-    const initialValues = new Array(10).fill(0).map((s, i) => {
-      return { value: i * 10, label: -i };
+    const initialValues = new Array(10).fill(0).map((_, i) => {
+      return { value: i * 10, label: 2 * ((10 - i) / 10) };
```
RaananW, line 230: Is there a reason why this one is 2 * (10-i) / 10 and not `(10-i) / 5` ? It will be more consistent with the next calculation, which is `((i + 10) * -1) / 5`.

sebavan: Agree with Raanan and could you put the magic formulas in dedicated functions so that it helps with maintenance ???

toledoal: Yes, so the number 2, right now, is because by default we show the user values from 0 to 2 (for float type) on the screen. And the middle value is 1. The next calculation creates the values on the drag up or down of the screen. Yes, my plan is to separate (specify) this magic formula for other scales. 😄  So we eventually have (hopefully) one function and just send the proper ranges to recreate this one.

sebavan: Yup so would be cool to start creating it within this PR to enhance readibility.

sebavan: THX

deltakosh:
up?

toledoal:
This one still needs more work. Because I just realized the axis values are printed correctly on the graph, but the actual values on the keyframe are not right.

sebavan:
Let me know once ok to merge on your side and I ll do it with the nightly :-) The style changes make it difficult to review :-)

toledoal:
Sorry, my prettier config went crazy. Let me try to set it back so it is not difficult to read.

toledoal:
Ok, finally I got it working 😃

sebavan:
Great I ll make a nightly and @PirateJC will be able to start testing :-)

PR #8556, opened 2020-07-20
Gizmo axis thickness

CedricGuillemet:
optional parameter to set axis thickness 

https://forum.babylonjs.com/t/enlarge-the-thickness-of-the-axes-in-gizmos/10277


RaananW:
Just need to solve the what's new conflicts

PR #8557, opened 2020-07-20
Fix docs Matrix.RotationYawPitchRoll and Matrix.RotationYawPitchRollT…

VSerain:
In the documentation “pitch” and “roll” are marked for axis X, but “roll” should be for the Z axis !

https://forum.babylonjs.com/t/error-in-matrix-documentation/12669


deltakosh:
Thanks!

PR #8559, opened 2020-07-20
Boundingbox gizmo jitter & fix point/directional light update

CedricGuillemet:
2 fixes for the price of 1.
first:
https://forum.babylonjs.com/t/max-gizmo-size/12374/3
In pointer drag behavior there was a gimbal lock when the axis for the plane is colinear to the camera direction. This results in pseudo random directed planes and jittering in scales for bounding box gizmo. The same behavior applies to other features that use this code.
second:
missing check for matrix update for light gizmos 
https://forum.babylonjs.com/t/unable-to-move-rotate-lights-using-scene-explorer-position-rotation-modes/12673/2


deltakosh:
There is a linting issue :)

PR #8561, opened 2020-07-20
Texture Inspector Design Refresh

DarraghBurkeMS:
Implements many of the UI elements designed by @PatrickRyanMS, including:
- New layout with properties on the top, tools on the left, and channels on the right
- 21 SVG icons for all buttons

Also implements several other features:
- Saving texture to local file
- Resetting texture to original state
- Channel mixing - now you can mix and match channels rather than just isolating single channels
- Face switching for cubemaps
- Grab pixel data under the cursor

The code for external tool loading is in now (using loadScript rather than eval), but there is no UI for adding tools at the moment. This means it is also impossible to modify the texture in the present state.


deltakosh:
Overall really good!. Some feedback (using this scene: https://www.babylonjs-playground.com/#AWQ0HM)

- no tooltip on buttons
- clicking on texture named "no name" crashes
- clicking on Edit on a new texture does not open a new window or replace the current texture in the texture inspector
- not sure what the dark background means when you click on the letter:
image
- clickable buttons must have a pointer: cursor (including the channels on the right)
-

DarraghBurkeMS:
> Overall really good!. Some feedback (using this scene: https://www.babylonjs-playground.com/#AWQ0HM)
> 
> * no tooltip on buttons
> * clickable buttons must have a pointer: cursor (including the channels on the right)

Fixed, thanks!

> * clicking on texture named "no name" crashes

Wasn't able to reproduce this one using the same playground on the latest version of the code. Did you open a different texture first maybe?

> * clicking on Edit on a new texture does not open a new window or replace the current texture in the texture inspector

Looks like this issue exists for the curve editor as well; it is related to way we are handling pop ups. I am investigating right now. Hopefully we can get multiple pop out windows open at the same time - inspecting multiple textures could be a useful workflow.

> * not sure what the dark background means when you click on the letter:
>   image

The idea is that you can "lock" a channel so that any edits won't be applied to that channel. So if you only want to edit the alpha channel, you could lock R, G & B. That feature isn't implemented yet though

deltakosh:
> Hopefully we can get multiple pop out windows open at the same time - inspecting multiple textures could be a useful workflow.
Agree!

> That feature isn't implemented yet though
So let's not enable it for now as this could be misleading

DarraghBurkeMS:
@deltakosh Fixed the issue with editing a different texture. When you click "Edit" with a window open:
- If it's the same texture, nothing will happen
- If it's a different texture, the original window will close and a new one will open

I also fixed a couple other bugs in the inspector that were causing it to crash.

inspector/src/components/actionTabs/lines/textureLineComponent.tsx
```diff
@@ -49,28 +49,28 @@ export class TextureLineComponent extends React.Component<ITextureLineComponentP
         this.updatePreview();
     }
 
-    updatePreview() {
+   async updatePreview() {
```
DarraghBurkeMS, line 52: Hope using async/await is okay here...I thought it was cleaner than Promises

inspector/src/components/actionTabs/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx
```diff
@@ -37,17 +37,24 @@ interface ITexturePropertyGridComponentProps {
     onPropertyChangedObservable?: Observable<PropertyChangedEvent>
 }
 
-export class TexturePropertyGridComponent extends React.Component<ITexturePropertyGridComponentProps> {
+interface ITexturePropertyGridComponentState {
```
DarraghBurkeMS, line 40: I migrated this to state because we want to trigger a re-render every time we open or close the texture editor popout

inspector/src/components/actionTabs/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx
```diff
@@ -230,7 +247,7 @@ export class TexturePropertyGridComponent extends React.Component<ITextureProper
                     <AnimationGridComponent globalState={this.props.globalState} animatable={texture} scene={texture.getScene()!} lockObject={this.props.lockObject} />
                 }
                 {
-                    (texture as any).rootContainer &&
+                    (texture as any).rootContainer && this._adtInstrumentation &&
```
DarraghBurkeMS, line 250: If this is undefined it can cause the inspector to crash

inspector/src/components/actionTabs/lines/textureLineComponent.tsx
deltakosh: it is no problem at all (we do not support IE for the inspector)

deltakosh:
Perfect! LGTM!

PR #8570, opened 2020-07-21
Remove console.log when using KHR_materials_transmission

zeux:
Also pass extensionContext instead of context when loading extension data.
(I believe this part is not essential here but it makes implementation match that
for other extensions)

### DIFF
--- a/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
+++ b/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
@@ -80,7 +80,7 @@ export class KHR_materials_transmission implements IGLTFLoaderExtension {
         }
 
         if (extension.transmissionTexture) {
-            return this._loader.loadTextureInfoAsync(context, extension.transmissionTexture)
+            return this._loader.loadTextureInfoAsync(`${context}/transmissionTexture`, extension.transmissionTexture)
                 .then((texture: BaseTexture) => {
                     pbrMaterial.subSurface.thicknessTexture = texture;
                     pbrMaterial.subSurface.useMaskFromThicknessTexture = true;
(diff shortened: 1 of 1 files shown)


loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -48,11 +48,10 @@ export class KHR_materials_transmission implements IGLTFLoaderExtension {
     /** @hidden */
     public loadMaterialPropertiesAsync(context: string, material: IMaterial, babylonMaterial: Material): Nullable<Promise<void>> {
         return GLTFLoader.LoadExtensionAsync<IKHRMaterialsTransmission>(context, material, this.name, (extensionContext, extension) => {
-            console.log(extensionContext);
             const promises = new Array<Promise<any>>();
             promises.push(this._loader.loadMaterialBasePropertiesAsync(context, material, babylonMaterial));
             promises.push(this._loader.loadMaterialPropertiesAsync(context, material, babylonMaterial));
-            promises.push(this._loadTransparentPropertiesAsync(context, material, babylonMaterial, extension));
+            promises.push(this._loadTransparentPropertiesAsync(extensionContext, material, babylonMaterial, extension));
```
RaananW, line 54: Is this change intentional? The 2 lines above still use `context`

zeux: The context is informational, and I believe it's used solely for error logging. The other two calls here load base material properties, and the third call loads extended properties - this way for example when loading the transmission texture the context for it will contain the extension, but other textures won't because they weren't loaded from the extension.

deltakosh:
@bghgary is out so please bear with us as he is OOF for the week

PR #8572, opened 2020-07-21
Bone: fix bug introduced by latest changes

Popov72:
See https://forum.babylonjs.com/t/skeletal-animation-potential-regression-due-to-yesterdays-nightly/12695/3

It seems the `_invertedAbsoluteTransform` is a "one shot" matrix generated at start and should not be modified, even when recomputing `_absoluteTransform`...


Drigax:
Yes, my understanding is that _invertedAbsoluteTransform represents the bone's absolute transform in its rest position

PR #8593, opened 2020-07-23
Font display

toledoal:
Hi! 😃 , This PR has a workaround the problem for loading fonts in the Curve Editor. 

This is a bug "Test font loading in Edge doesn't load sometimes" #8478 

It appears that the problem is caused by trying to copy the styles. We get a CSS load error. 

image

```
DOMException: Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules
    at Function../inspector.ts.Inspector._CopyStyles (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:77426:32)
    at Function../inspector.ts.Inspector._CreatePopup (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:77632:14)
    at PopupComponent../components/popupComponent.tsx.PopupComponent.openPopup (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:75089:80)
    at PopupComponent../components/popupComponent.tsx.PopupComponent.componentDidMount (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:75082:14)
    at commitLifeCycles (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:47995:22)
    at commitLayoutEffects (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:50984:7)
    at HTMLUnknownElement.callCallback (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:28369:14)
    at Object.invokeGuardedCallbackDev (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:28418:16)
    at invokeGuardedCallback (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:28473:31)
    at commitRootImpl (http://localhost:1338/.temp/localDevUMD/inspector/babylon.inspector.bundle.js:50722:9) {stack: <accessor>, code: 18, name: 'SecurityError', message: 'Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules'}
```

By adding the fonts on the Curve Editor stylesheet we are able to load the fonts correctly.


deltakosh:
This is weird because we do not have the issue with NME for instance (which is also pop out) or the texture inspector

toledoal:
I'll look into NME implementation, to see what is the difference there... A few weeks ago the curve editor was able to load the fonts correctly though...

toledoal:
I think I found what to do. In NME there is this fix for the fonts (link) and late binding. This is not in the Inspector.ts file, so I might add a Popup.ts class. 
image

deltakosh:
deal!

inspector/src/components/popup.ts
```diff
@@ -0,0 +1,79 @@
...
+            // need this for late bindings
+            this._CopyStyles(window.document, parentDocument);
+        }, 0);
+
+        (this as any)[windowVariableName] = popupWindow;
+
+        return parentControl;
+    }
+
+    private static _CopyStyles(sourceDoc: HTMLDocument, targetDoc: HTMLDocument) {
```
deltakosh, line 52: I'm fine with that change but then we need to rewire the inspector.tsx code to avoid duplications

toledoal: Sure, makes sense. Is it ok to create another PR for that particular update on inspector.tsx ?

toledoal: Another option is to refactor what we currently have in Inspector, instead of adding the Popup.ts class... I don't any wrong effects for the other tabs... what do you think...

deltakosh: I like that (2nd) idea :)

toledoal: ok, So I'll just refactor this and send a new commit for this PR 😃

inspector/src/inspector.ts
```diff
@@ -300,6 +316,11 @@ export class Inspector {
 
         this._CopyStyles(window.document, parentDocument);
 
+        setTimeout(() => {
```
deltakosh, line 319: I would recommend to have the timeout on demand to avoid popping effect

toledoal: sorry, I don't know understand the concept. For this, you mean to set it like this?
```
const timeOut = setTimeout(() => {
this._CopyStyles(window.document, parentDocument);
}, 0);

```

deltakosh: I mean either call the copy style directly or with a timeout based on a parameter of the function so no changes for current users of that function

toledoal: ah, got it.

PR #8602, opened 2020-07-24
Fix Play animation

toledoal:
Hi! 😄  This PR has a fix for the #8478 Animation play bug.

The main problem was that I was confusing the speedRatio with the animation FPS... 

When the FPS inputs changed, I was also changing the speedRatio...


inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx
```diff
@@ -1380,12 +1380,10 @@ export class AnimationCurveEditorComponent extends React.Component<
                 let firstFrame = keys[0].frame;
                 let LastFrame = this.state.selected.getHighestFrame();
                 if (direction === 1) {
-                    this._mainAnimatable = this.props.scene.beginAnimation(target, firstFrame, LastFrame, this.state.isLooping, this.state.fps);
-                    console.log(this._mainAnimatable);
-                    console.log(`${target}, ${firstFrame}, ${LastFrame}, ${this.state.isLooping}, ${this.state.fps}`);
+                    this._mainAnimatable = this.props.scene.beginAnimation(target, firstFrame, LastFrame, this.state.isLooping);
```
sebavan, line 1383: If both cases are the same why not removing the if ???

sebavan: Scratch that, got it :-)

PR #8608, opened 2020-07-24
Fix Effect Layer Orientation

sebavan:
https://forum.babylonjs.com/t/inverted-glow-layer-for-gltf-imported-mesh-after-baking-transformations-to-vertices/12823/9


Popov72:
Note sure my comment went in, so here it is again:

Line 678, `0` is a valid value for renderingMesh.overrideMaterialSideOrientation, so you should do `renderingMesh.overrideMaterialSideOrientation ?? material.sideOrientation;` instead.

PR #8613, opened 2020-07-26
P8 skeleton viewer try2

Pryme8:
I had to go back and make a new branch to get things to start working again.

Also, there is a noticeable hang when the inspector makes the skeleton widget vs making it with its constructor in code.

Really wondering what is going on with it!

Ill be pushing the rest of the code for it tomorrow, along with the fallback methods.

I also updated the old line viewer to use observables instead of the functional binding.

Everything so far should be backward compatible. 

Gonna do a build right now and commit the linting and build fixes.

### DIFF
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -15,7 +15,8 @@ import { VertexBuffer } from '../Meshes/buffer';
 import { ISkeletonViewerOptions } from './ISkeletonViewer';
 import { Observer } from '../Misc/observable';
 
-import '../Meshes/Builders/sphereBuilder';
+import { SphereBuilder } from '../Meshes/Builders/sphereBuilder';
+import { ShapeBuilder } from '../Meshes/Builders/shapeBuilder';
 
 /**
  * Class used to render a debug view of a given skeleton
@@ -316,7 +317,7 @@ export class SkeletonViewer {
 
                                     let up0 = up.scale(midStep);
 
-                                    let spur = MeshBuilder.ExtrudeShapeCustom(bc.name + ':spur',
+                                    let spur = ShapeBuilder.ExtrudeShapeCustom(bc.name + ':spur',
                                     {
                                         shape:  [
                                                     new Vector3(1, -1,  0),
@@ -358,7 +359,7 @@ export class SkeletonViewer {
 
                                 let sphereBaseSize = displayOptions.sphereBaseSize || 0.2;
 
-                                let sphere = MeshBuilder.CreateSphere(bone.name + ':sphere', {
+                                let sphere = SphereBuilder.CreateSphere(bone.name + ':sphere', {
                                     segments: 6,
                                     diameter: sphereBaseSize,
                                     updatable: true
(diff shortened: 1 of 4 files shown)


deltakosh:
There is a doc error. You can check all these checks locally with "npm run build" in the gulp/tools folder

src/Debug/skeletonViewer.ts
```diff
@@ -5,31 +5,44 @@ import { Nullable } from "../types";
...
+
+    /** If SkeletonViewer is enabled. */
     private _isEnabled = false;
-    private _renderFunction: () => void;
-    private _utilityLayer: Nullable<UtilityLayerRenderer>;
 
-    /**
-     * Returns the mesh used to render the bones
-     */
-    public get debugMesh(): Nullable<LinesMesh> {
```
deltakosh, line 30: This is a breaking change

src/Debug/skeletonViewer.ts
```diff
@@ -152,8 +218,205 @@ export class SkeletonViewer {
...
+                            bone._index = i;
+                        }
+                        if (bone._index === -1) {
+                            continue;
+                        }
+
+                        let boneAbsoluteRestTransform = new Matrix();
+                        getAbsoluteRestPose(bone, boneAbsoluteRestTransform);
+
+                        let anchorPoint = new Vector3(); //bone.getPosition(BABYLON.Space.WORLD, this.mesh)
```
deltakosh, line 280: comments to remove

src/Debug/skeletonViewer.ts
```diff
@@ -152,8 +218,205 @@ export class SkeletonViewer {
...
+                        boneAbsoluteRestTransform.decompose(undefined, undefined, anchorPoint);
+
+                            bone.children.forEach((bc, i) => {
+                                let childAbsoluteRestTransform : Matrix = new Matrix();
+                                bc.getRestPose().multiplyToRef(boneAbsoluteRestTransform, childAbsoluteRestTransform);
+                                let childPoint = new Vector3(); //bc.getPosition(BABYLON.Space.WORLD, this.mesh)
```
deltakosh, line 286: same

src/Debug/skeletonViewer.ts
Pryme8: can you expand on that please?

Pryme8: I moved the get debugMesh to under the constructor.

deltakosh: Actually I realized they are at the end of the file can you please move the properties and the static at the beginning?

Pryme8: I just moved the statics, move the getters and setters to the beginning as well?

deltakosh: Check other files. Static first then properties then constructor then functions

Pryme8: will fix that now then mark this resloved

deltakosh:
You added new parameter to the constructor (options and _scene), you need to document them:
image

by the way parameters of a constructor should not start with _, the _scene should become scene and then you can  copy the scene parameter to this._scene in the constructor body

PR #8623, opened 2020-07-27
Add transmission prerender pass when using KHR_materials_transmission

MiiBond:
To better support transmission as defined in the glTF `KHR_materials_transmission` extension, here's an example of a Helper that can setup the rendering of the opaque scene behind the transmissive object(s).
To enable it, add this line:
`var transHelper = new BABYLON.TransmissionHelper({renderSize: 1024}, scene);`

Currently, transmissive surfaces just show the IBL through them like shown here:
Without helper

With the helper running, transmissive surfaces will show the opaque scene behind instead:
With helper

Looking for feedback. Is this even the right way to handle this issue?


sebavan:
I am also wondering at what extend we could rely on our realtime filtering feature to ensure an accurate roughness ?

sebavan:
I also wonder if it should be derived into ScreenSpaceRefraction as we are doing for reflection as it can be extended to standard mat ?

I know it is a lot of questions but I love the idea very much !!!

MiiBond:
I've added `dispose` and removed `getRenderTargetTextures = null` (I think it was necessary in an earlier version of this).
I didn't know that there was an option for realtime-filtering a texture. That sounds great but I'll have to dive into how it works. It looks like it's a flag on a material and not a texture right now and when I enable it, I get shader compile errors.

sebavan:
yup it is only at the material level as we did not split refraction from reflection in this mode which is smthg I could add later.

I bet the error comes from texture 2d not being supported at the moment (only cube) but it would be so awesome to introduce this part :-)

MiiBond:
It would be great to have this for 2D refraction textures, for sure.
I have a version of this helper for our viewer's transparency code which is far more complicated (and horribly messy) to handle multiple layers of transmissive objects but, ultimately, it still generates a 2D refraction texture to be used by the material.

deltakosh:
We have some checks issue (mostly doc and linting)

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -6,6 +6,258 @@ import { IMaterial } from "../glTFLoaderInterfaces";
...
+    /**
+     * The size of the render buffers
+     */
+    renderSize: number;
+}
+
+/**
+ *
+ */
+export class TransmissionHelper {
```
bghgary, line 26: Does this need to be exported?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -6,6 +6,258 @@ import { IMaterial } from "../glTFLoaderInterfaces";
...
+     */
+    private readonly _scene: Scene;
+    private _options: ITransmissionHelperOptions;
+    private _opaqueRTIndex: number = -1;
+
+    private _opaqueRenderTarget: Nullable<RenderTargetTexture> = null;
+    private _opaqueMeshesCache: Mesh[] = [];
+    private _transparentMeshesCache: Mesh[] = [];
+
+    public disabled: boolean = false;
```
bghgary, line 48: nit: Double negative for properties are confusing. Consider using `enabled` instead.

bghgary: I'm also a bit confused by this property being public. Are you expecting users of this class to be able to set this?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -6,6 +6,258 @@ import { IMaterial } from "../glTFLoaderInterfaces";
...
+        this._scene.onNewMeshAddedObservable.add(this._addMesh.bind(this));
+        // Listen for when a mesh is removed from to the scene and remove it from our cache lists.
+        this._scene.onMeshRemovedObservable.add(this._removeMesh.bind(this));
+    }
+
+    // When one of the meshes in the scene has its material changed, make sure that it's in the correct cache list.
+    private onMeshMaterialChanged(mesh: AbstractMesh) {
+        if (mesh instanceof Mesh) {
+            let transparent_idx = this._transparentMeshesCache.indexOf(mesh);
+            let opaque_idx = this._opaqueMeshesCache.indexOf(mesh);
```
bghgary, line 164: nit: snake case isn't used anywhere in this class except here. seems a bit odd.
nit: let --> const

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
bghgary: Actually, is this property even necessary? It seems to be only used in updateOptions. Is that expected?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -6,6 +6,258 @@ import { IMaterial } from "../glTFLoaderInterfaces";
...
+            renderSize: 512
+        };
+    }
+
+    /**
+     * Stores the creation options.
+     */
+    private readonly _scene: Scene;
+    private _options: ITransmissionHelperOptions;
+    private _opaqueRTIndex: number = -1;
```
bghgary, line 42: Does this need to be a member? It's only used in one function.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -6,6 +6,258 @@ import { IMaterial } from "../glTFLoaderInterfaces";
...
+        // Remove any layers rendering to the opaque scene.
+        if (this._scene.layers) {
+            this._scene.layers.forEach((layer) => {
+                if (this._opaqueRenderTarget) {
+                    let idx = layer.renderTargetTextures.indexOf(this._opaqueRenderTarget);
+                    if (idx >= 0) {
+                        layer.renderTargetTextures.splice(idx, 1);
+                    }
+                }
+            });
```
bghgary, line 209: The if check can be outside of the for loop. Also, consider using for...of loops instead of forEach everywhere since it's a bit cheaper.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -73,6 +327,9 @@ export class KHR_materials_transmission implements IGLTFLoaderExtension {
 
         if (extension.transmissionFactor !== undefined) {
             pbrMaterial.subSurface.refractionIntensity = extension.transmissionFactor;
+            if (pbrMaterial.subSurface.refractionIntensity && this._transmissionHelper == undefined) {
+                this._transmissionHelper = new TransmissionHelper({}, pbrMaterial.getScene());
```
bghgary, line 331: I'm not sure what would keep this object alive. This loader extension should, in theory, get garbage collected once the load is complete. Once that happens, who is holding this object? Is that an issue?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: Yeah, sorry. This is left over from the much more complex version that I have in the Adobe viewer. I'll try to clean this up in this version.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: I think the reason that it's inside the loop is that tslint complains when I put it outside. This is because `this._opaqueRenderTarget` is nullable and `layer.renderTargetTextures.indexOf` doesn't take `null` as a valid type. Is there a nice TS solution to this?

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: Yes, this is an issue with doing things this way. The helper should ideally stick around so that it can handle material changes and mesh additions and removals. Also, we shouldn't be creating new helpers for each transmissive glTF loaded... that would be bad.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
Popov72: Just pass `this._opaqueRenderTarget!` instead of `this._opaqueRenderTarget` if you know that `this._opaqueRenderTarget` can't be null but TS doesn't.

bghgary: If you use for...of, it won't be an issue anymore :)

bghgary: That's one of the down sides to using forEach, not to mention possible perf issues.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: Okay, finally had a chance to change this.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: Thank you for the tips.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
bghgary: Not sure if you meant to remove the property or still keep it?

MiiBond: Sorry, I should remove it. Thanks.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: It's gone now.

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
MiiBond: Nope.

sebavan (approved):

sebavan:
Actually, @MiiBond, what is calling dispose on the transmission helper ? I am wondering if it would not leak ?

MiiBond:
@sebavan the helper has global scope and only one should ever be created. We don't want it to be cleaned up when the loader is and we don't want another one to be created when another loader is.

sebavan:
@MiiBond  I was thinking on scene.dispose, should we add a scene.onDisposeObservable to call the helper dispose function ?

MiiBond:
@sebavan Quite right. I'll open another PR.

PR #8626, opened 2020-07-27
Fix

toledoal:
This PR has improvements for the following in issue #8478 

-You cannot delete a key with the 'delete' key, only the 'backspace' key. It should probably be both?
- (Feature) When adding a new keyframe, add value between two points in the curve
- (Feature) Flat tangents (only for selected keyframes)

### DIFF
--- a/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor.scss
+++ b/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor.scss
@@ -508,14 +508,15 @@ body {
     align-items: flex-start;
     justify-content: flex-start;
     flex-direction: column;
-    height: 465px;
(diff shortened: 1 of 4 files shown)


inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor.scss
```diff
@@ -660,7 +661,7 @@ body {
         }
 
         .timeline-wrapper {
-          margin-top: -45px;
+          margin-top: -31px;
```
RaananW, line 664: Where is this value coming from? 31px?

toledoal: This change is just to allow a better alignment here 
image

PR #8637, opened 2020-07-29
(hopefully) Fix skeleton returnToRest

Popov72:
See https://forum.babylonjs.com/t/return-to-rest-not-behaving-as-expected/12570/68


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1125,7 +1125,6 @@ export class GLTFLoader implements IGLTFLoader {
                 baseMatrix.multiplyToRef(babylonParentBone.getInvertedAbsoluteTransform(), baseMatrix);
             }
 
-            babylonBone.setRestPose(baseMatrix);
```
Drigax, line 1128: Bone._restPose is no longer set if you remove this.

Popov72: Yes, it is set by the constructor, with the local matrix passed in: the local matrix is copied to the rest matrix.

After having spoken with @Deltakosh, it appears the rest pose is the initial matrix state, when no animation or anything else is applied. So, it is the local matrix passed when calling the constructor. That's why there was no setter for the rest pose in the first place.

To me, the `baseMatrix` as calculated in this code does not correspond to the Babylonjs rest matrix.

Now, that may not be what some users need, but it's another question.

With this change, all models from the thread linked above do work: we can call `returnToRest` whenever we want, the meshes come back to their initial state correctly (which may not be the T-pose).

deltakosh:
Tests are failing :D

loaders/src/glTF/2.0/glTFLoader.ts
Drigax: > Yes, it is set by the bone constructor, with the local matrix passed in: the local matrix is copied to the rest matrix.

I don't think that that's a correct assumption to make.

Take a scenario where a model is exported with posed joint orientation. (imagine someone modelled a non-animated statue, or a posed toy, or something else skeletally rigged, but not necessarily animated) 

If we don't use the inverseBindMatrix that defines our actual bindPose supplied by the glTF, and instead use the node orientation provided that the bone's node is exported as, we are now using an arbitrary pose for the skeleton's rest pose, but still binding it using its actual rest pose.

Drigax: Yeah I think I'm wrong here, my mistake, our bindPose is what we want to use to transform our mesh vertices, while our rest pose is what we want to consider as "home" for a given joint, I'd assume. where our bind pose is what its modelled as, while rest may be the toy in its exported posed position

Popov72:
> Tests are failing :D

Not my fault :)

```
[at-loader] ../../inspector/src/components/actionTabs/tabs/propertyGrids/animations/animationCurveEditorComponent.tsx:1606:65 
    TS2322: Type 'number' is not assignable to type 'string | undefined'.
```

deltakosh:
Master is building so it may be a difference on your branch

sebavan:
So weird indeed... or a new version of a npm dependency just falled in I ll fix it on master if your branch fails again

sebavan:
Actually after the merge, it looks all better.

PR #8639, opened 2020-07-29
Fix #8638

### LINKED ISSUE #8638: Implementing thin instance picking support
See https://forum.babylonjs.com/t/picking-instances-using-ext-mesh-gpu-instancing/12399/7

deltakosh:
There is a REALLY small perf hit as we need to use a smartarray but I was not able to find any difference even with thousands of meshes

### DIFF
--- a/src/Culling/ray.ts
+++ b/src/Culling/ray.ts
@@ -692,29 +692,31 @@ Scene.prototype._internalPick = function (rayFunction: (world: Matrix) => Ray, p
             continue;
         }
 
-        let world = mesh.skeleton && mesh.skeleton.overrideMesh ? mesh.skeleton.overrideMesh.getWorldMatrix() : mesh.getWorldMatrix();
-
-        let result = this._internalPickForMesh(pickingInfo, rayFunction, mesh, world, fastCheck, onlyBoundingInfo, trianglePredicate);
-
-        if (result) {
-            pickingInfo = result;
-
-            if (fastCheck) {
-                return pickingInfo;
-            }
-        }
-
         if (mesh.hasThinInstances && (mesh as Mesh).thinInstanceEnablePicking) {
             let thinMatrices = (mesh as Mesh).thinInstanceGetWorldMatrices();
-            for (let world of thinMatrices) {
+            for (let index = 0; index < thinMatrices.length; index++) {
+                let world = thinMatrices[index];
                 let result = this._internalPickForMesh(pickingInfo, rayFunction, mesh, world, fastCheck, onlyBoundingInfo, trianglePredicate);
 
                 if (result) {
                     pickingInfo = result;
+                    pickingInfo.thinInstanceIndex = index;
 
                     if (fastCheck) {
                         return pickingInfo;
                     }
+                }
+            }
+        } else {
+            let world = mesh.skeleton && mesh.skeleton.overrideMesh ? mesh.skeleton.overrideMesh.getWorldMatrix() : mesh.getWorldMatrix();
+
+            let result = this._internalPickForMesh(pickingInfo, rayFunction, mesh, world, fastCheck, onlyBoundingInfo, trianglePredicate);
+
+            if (result) {
+                pickingInfo = result;
+
+                if (fastCheck) {
+                    return pickingInfo;
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


deltakosh:
PG to test: https://www.babylonjs-playground.com/#RC2IAH

src/Culling/ray.ts
```diff
@@ -671,22 +692,31 @@ Scene.prototype._internalPick = function (rayFunction: (world: Matrix) => Ray, p
...
 
-        if (!fastCheck && pickingInfo != null && result.distance >= pickingInfo.distance) {
-            continue;
+        if (result) {
+            pickingInfo = result;
+
+            if (fastCheck) {
+                return pickingInfo;
+            }
         }
```
Popov72, line 706: If the mesh has thin instances (and picking is enabled), we should not check for intersection with the mesh world matrix but only with the thin instance matrices. The mesh itself is not rendered when it has thin instances: if one wants to render the mesh, `thinInstanceAddSelf` should be called (and then a matrix will be added to the matrix buffer).

src/Culling/ray.ts
```diff
@@ -710,15 +740,24 @@ Scene.prototype._internalMultiPick = function (rayFunction: (world: Matrix) => R
...
+        let world = mesh.skeleton && mesh.skeleton.overrideMesh ? mesh.skeleton.overrideMesh.getWorldMatrix() : mesh.getWorldMatrix();
 
-        var result = mesh.intersects(ray, false, trianglePredicate);
-        if (!result || !result.hit) {
-            continue;
+        let result = this._internalPickForMesh(null, rayFunction, mesh, world, false, false, trianglePredicate);
+
+        if (result) {
+            pickingInfos.push(result);
         }
```
Popov72, line 750: Same thing here than above.

src/Meshes/thinInstanceMesh.ts
```diff
@@ -228,12 +238,34 @@ Mesh.prototype.thinInstanceBufferUpdated = function(kind: string): void {
...
 };
 
+Mesh.prototype.thinInstanceGetWorldMatrices = function(): Matrix[] {
+    if (!this._thinInstanceDataStorage.matrixData || !this._thinInstanceDataStorage.matrixBuffer) {
+        return [];
+    }
+    const matrixData = this._thinInstanceDataStorage.matrixData;
+
+    if (!this._thinInstanceDataStorage.worldMatrices || this._thinInstanceDataStorage.worldMatrices.length !== this._thinInstanceDataStorage.instancesCount) {
+        this._thinInstanceDataStorage.worldMatrices = new Array<Matrix>();
```
Popov72, line 255: ~~We could have removed / added a thin instance in sequence, so the check `this._thinInstanceDataStorage.worldMatrices.length !== this._thinInstanceDataStorage.instancesCount` could be false but we would still need to update the `worldMatrices` array. Also, the user can use `thinInstanceSetMatrixAt` that would change a matrix without changing the array size.~~ see next comment

I think the easiest way to handle the "dirtyness" of `worldMatrices` is to reset it to null in: `thinInstanceAdd`, `thinInstanceAddSelf` and `thinInstanceSetMatrixAt`. That way all cases will be handled and you only need the `if (!this._thinInstanceDataStorage.worldMatrices)` part of the check.

src/Meshes/thinInstanceMesh.ts
```diff
@@ -228,12 +238,34 @@ Mesh.prototype.thinInstanceBufferUpdated = function(kind: string): void {
...
+
+        for (let i = 0; i < this._thinInstanceDataStorage.instancesCount; ++i) {
+            this._thinInstanceDataStorage.worldMatrices[i] = Matrix.Zero();
+        }
+    }
+
+    for (let i = 0; i < this._thinInstanceDataStorage.instancesCount; ++i) {
+        Matrix.FromArrayToRef(matrixData, i * 16, this._thinInstanceDataStorage.worldMatrices[i]);
+    }
+
```
Popov72, line 265: My bad in the comment above: you copy the matrices each time the function is called, so no sync problems with the thin instance matrices!

However, with the proposed changes on dirtyness, this could be added in the `if` block above and not done each time we call the function (with tens of thousands thin instances, it may make a difference (?)). But to be efficient, the previous array of world matrices should be kept somewhere before nulling it to avoid recreating the array + empty matrices in case the current thin instance matrix buffer length is the same than the length of the saved matrix array.

src/Culling/ray.ts
```diff
@@ -653,15 +659,30 @@ Scene.prototype.createPickingRayInCameraSpaceToRef = function (x: number, y: num
     return this;
 };
 
+Scene.prototype._internalPickForMesh = function (pickingInfo: Nullable<PickingInfo>, rayFunction: (world: Matrix) => Ray, mesh: AbstractMesh, world: Matrix, fastCheck?: boolean, onlyBoundingInfo?: boolean, trianglePredicate?: TrianglePickingPredicate) {
+    let ray = rayFunction(world);
```
Popov72, line 663: I think this function should take a thinInstanceIndex (-1 by default) that it copies into the returned `PickingInfo` structure, so that the user can know which thin instance is intersected.

Popov72:
Also, don't forget to update the doc: https://doc.babylonjs.com/how_to/how_to_use_thininstances ("Support" section) 😃

src/Culling/ray.ts
deltakosh: Good point!

src/Meshes/thinInstanceMesh.ts
deltakosh: Yeah It thought about it 

> Also, don't forget to update the doc: https://doc.babylonjs.com/how_to/how_to_use_thininstances ("Support" section) 😃

Yes as soon as it will be merged, I will add more doc for it

deltakosh:
> Also, don't forget to update the doc: https://doc.babylonjs.com/how_to/how_to_use_thininstances ("Support" section) 😃

Yes as soon as it will be merged, I will add more doc for it

src/Culling/ray.ts
deltakosh: yeah I debated with myself a lot about that...but you right I'll add it

src/Meshes/thinInstanceMesh.ts
deltakosh: I went to a live version which is FAR more efficient. Please let me know your thought (If I missed something obvious)

Popov72: Indeed, that's a lot better!

The only thing is that the worldMatrices array will never shrink... Not sure if it is a problem, though.

deltakosh: Yeah not a biggy! I'll merge then
Thanks a ton!

PR #8640, opened 2020-07-30
gizmos now support cameras

CedricGuillemet:
follow up on this thread : https://forum.babylonjs.com/t/gizmo-support-for-cameras/12888/9


src/Gizmos/gizmo.ts
```diff
@@ -161,7 +163,33 @@ export class Gizmo implements IDisposable {
...
+                camera.parent.getWorldMatrix().invertToRef(parentInv);
+                this._attachedNode._worldMatrix.multiplyToRef(parentInv, localMat);
+                localMat.decompose(scaling, transformQuaternion, position);
+            } else {
+                this._attachedNode._worldMatrix.decompose(scaling, transformQuaternion, position);
+            }
+            camera.position = position;
+            if (inheritsTargetCamera) {
+                var targetCamera = this._attachedNode as TargetCamera;
+                var angles = transformQuaternion.toEulerAngles();
```
RaananW, line 189: a target camera can have a quaternion as rotation. If it is set, copy from the transforrmQuaternion

src/Gizmos/gizmo.ts
```diff
@@ -161,7 +163,33 @@ export class Gizmo implements IDisposable {
         if (!this._attachedNode) {
             return;
         }
-        if (this._attachedNode.getClassName() === "Mesh" || this._attachedNode.getClassName() === "AbstractMesh" || this._attachedNode.getClassName() === "TransformNode") {
+        var inheritsTargetCamera = this._attachedNode.getClassName() === "FreeCamera"
```
RaananW, line 166: Just a thought - I am not sure how often this function is executed, but - if it is executed every frame, maybe first check if classname contains camera, to avoid all of those comparisons (and skip in case of a mesh)

src/Gizmos/gizmo.ts
```diff
@@ -161,7 +163,33 @@ export class Gizmo implements IDisposable {
...
-        if (this._attachedNode.getClassName() === "Mesh" || this._attachedNode.getClassName() === "AbstractMesh" || this._attachedNode.getClassName() === "TransformNode") {
+        var inheritsTargetCamera = this._attachedNode.getClassName() === "FreeCamera"
+                        || this._attachedNode.getClassName() === "FlyCamera"
+                        || this._attachedNode.getClassName() === "ArcFollowCamera"
+                        || this._attachedNode.getClassName() === "TargetCamera"
+                        || this._attachedNode.getClassName() === "TouchCamera"
+                        || this._attachedNode.getClassName() === "UniversalCamera";
+        if  (this._attachedNode.getClassName() === "ArcRotateCamera" || inheritsTargetCamera) {
+            var camera = this._attachedNode as Camera;
+            var transformQuaternion = new Quaternion(0, 0, 0, 1);
```
RaananW, line 174: Same as above - if this is constantly called, you can keep the GC from over-working if you use the predefined Temp vectors/quaternion/matrix object(s)

src/Gizmos/gizmo.ts
Popov72: On the `Camera` class you have a hidden public read only property `_isCamera = true`: maybe just test for it?

src/Gizmos/gizmo.ts
```diff
@@ -161,7 +163,33 @@ export class Gizmo implements IDisposable {
...
+            var position = new Vector3(0, 0, 0);
+            var scaling = new Vector3(1, 1, 1);
+            if (camera.parent) {
+                var parentInv = new Matrix();
+                var localMat = new Matrix();
+                camera.parent.getWorldMatrix().invertToRef(parentInv);
+                this._attachedNode._worldMatrix.multiplyToRef(parentInv, localMat);
```
Popov72, line 181: Why not calling `getWorldMatrix()` instead of accessing directly `_worldMatrix` (it could be out of date)?

PR #8651, opened 2020-07-31
NME: Import frame

### LINKED ISSUE #8387: Finish frame import / export work
Add Option to save a portion of a graph (and an option to import it as well): Almost done, import does not yet restore links

Popov72:
Fix #8387

### DIFF
--- a/src/Materials/Node/nodeMaterial.ts
+++ b/src/Materials/Node/nodeMaterial.ts
@@ -1627,20 +1627,19 @@ export class NodeMaterial extends PushMaterial {
             }
         }
 
-        // Connections
-        if (!merge) {
-            // Starts with input blocks only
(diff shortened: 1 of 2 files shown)


src/Materials/Node/nodeMaterial.ts
```diff
@@ -1659,6 +1659,10 @@ export class NodeMaterial extends PushMaterial {
                 }
             }
 
+            if (merge && this.editorData && this.editorData.locations) {
+                locations.concat(this.editorData.locations);
+            }
+
```
Popov72, line 1665: I feel this was missing from the previous PR so I added it: is that ok?

deltakosh: yep

deltakosh:
Ok to merge! thanks buddy!

PR #8659, opened 2020-08-01
Improve buffer

ottoville:
This is my second proposal for buffer creation and updating. It uses the "getMappedRange" method for chunks, avoiding of waiting any promise caused by buffer mapping.


sebavan:
Does not this result in exactly the same behaviour ? I am wondering what it improves in the end ?

Write buffer is probably doing the same internally natively so as long as we do not do the full rewrite by reusing a single large buffer with async up and down I would prefer to not change it to this except if there is any advantages.

It actually seems more heavy by relying on intermediate buffers and so on. 

The only annoying part we still have is the loop to upload due to a left over bug in chrome windows. Actually, there is no max size defined in the spec.

deltakosh:
Up?

sebavan:
@ottoville ?

deltakosh:
Closing for now. Please reopen at will :)

ottoville:
You are right. One way to make it more performt would be to create a constant staging buffer and feed modified data throug the staging buffer. The constant staging buffer is suitable for uniform buffers that are always same size and do change every frame, such as the scene ubo.

PR #8666, opened 2020-08-03
fix gizmo update for some meshes like GroundMesh

CedricGuillemet:
Quick fix for https://forum.babylonjs.com/t/getclassname-in-gizmos-code/13071/3


Popov72:
All good to me!

PR #8669, opened 2020-08-03
Canvas Drag Issues

toledoal:
Hi! 😃 This PR contains fixes for the following bugs in issue #8478

- [ ] Dragging the canvas around feels sluggish and stuttery -Improvements on dragging (svgDraggable.tsx)
- [ ] Moving the play to next/previous keyframes with the UI does not reposition the canvas (controls.tsx)
- [ ] Animation Begin/End buttons do not go to the first/last frame. They are currently advancing the frame by one, which is incorrect behavior. (controls.tsx)
- [ ] The play control buttons do not reposition canvas (this one comprehends the two previous bugs) (Controls.tsx)
- [ ] Flat tangents (only for selected keyframes)
- [ ] When changing animation frame size, change the length of available frames (timeline.tsx)

Temporarily removed Vector3 and other animations from creation (changes in addAnimation.tsx) and also removed the initial keyframe when adding new animation.  

We can explore the Canvas drag, and explore if there is a need to polish it.


toledoal:
Yes, actually I'm working on the resize window and zoom bugs, and I'm going to change most of those. About the style, I don't think now it will change, because I was not using the prettier file defined in the root directory (4 spaces). And now that I'm using it that why it got so different.  :(

sebavan:
Ok so let s wait on the resize before merging so at least there would be less confusion reading the magic numbers. what do you think ?

toledoal:
The resize bug is a big one... and I think I'm going to work on it for a few days... if you can try to test this one would be better. Let me try to fix the magic numbers for this PR only.

sebavan:
Perfect let s do that !!! I let you fix the magic numbers and then I ll merge.

toledoal:
Please note that I will refactor the SVG elements (basically for styling and magic numbers) in the next PR...

Issue #8646, opened 2020-07-30
Delete Object with Delete Key in Scene Viewer 


msDestiny14:
Delete and mesh, particle etc with delete key in scene viewer if it is being focused/selected.

PR #8675, opened 2020-08-03
Delete Objects in Scene Explorer with the Delete Key

msDestiny14:
Be able to delete an object in the scene explorer with the delete key when it is focused. Also fixed a bug relating to the key input not registering in the inspector in playground.

https://github.com/BabylonJS/Babylon.js/issues/8646#issue-669201642


deltakosh:
Excellent!

PR #8677, opened 2020-08-03
Expose feature points as an experimental WebXR feature

Alex-MSFT:
This PR introduces feature points as an experimental feature when running on top of BabylonNative.  Feature Points are the base understanding of the world geometry for AR Devices which are the foundation of planes and scene meshes.  The high level overview of the feature can be seen in [FeaturePoints.md](https://github.com/Alex-MSFT/Babylon.js/blob/feature_points/src/LibDeclarations/NativeExtensions/FeaturePoints.md).

This PR adds a new webxr.nativeextensions.d.ts file which allows us to extend the WebXR spec beyond what is currently supported in the browser today, which allows us to create extended support for native functionality on BabylonNative.  

This PR also adds a new WebXRFeaturePointSystem to track feature points the common flow for using this system is as follows

- Enable the feature on the XRSession using the new feature key: WebXRFeatureName.FEATURE_POINTS
- Register an observer for: onFeaturePointsUpdatedObservable, which returns the current frame's updated feature points as a list of feature point IDs
- On callback look up the updated feature point in the module's exposed featurePointCloud member and perform any app side logic based on the updated feature point values.
- Alternatively you can process the entire featurePointCloud on demand which should remain durable across frames.


src/LibDeclarations/NativeExtensions/FeaturePoints.md
```diff
@@ -0,0 +1,61 @@
+# Feature point cloud
```
Alex-MSFT, line 1: Is there a better place for this file to live?

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Observers registered here will be executed whenever new feature points are available (on XRFrame while the session is tracking).
+     * Will notify the observers about which feature points have been updated.
+     */
+    public onFeaturePointsUpdatedObservable: Observable<number[]> = new Observable();
+    /**
+     * The currrent feature point cloud maintained across frames.
+     */
+    private _featurePointCloud: Array<IWebXRFeaturePoint> = [];
+    public readonly featurePointCloud: Array<IWebXRFeaturePoint> = this._featurePointCloud;
```
Alex-MSFT, line 52: Should this be a dictionary of id -> feature point instead?  Right now this is dependent on the IDs being strictly ordered.  Since we handle this on the BabylonNative side this is okay, but maybe it would be safer not to make this assumption.

src/LibDeclarations/NativeExtensions/FeaturePoints.md
```diff
@@ -0,0 +1,61 @@
+# Feature point cloud
+## Introduction
+This document proposes an API for exposing Feature Point Clouds for WebXR.
+
+## Overview
+The API is synchronous, and on demand.  When served a new XRFrame the consuming app may query for the current feature point cloud for a given frame. As feature points comprise a large set of raw data it is best not to expose this to the JS application unless needed for app processing, so feature points must be requested handled in an on-demand fashion in order to reduce the amount of data presented to the consuming app, which must set its own frequency for updating its world understanding.
```
hendryems, line 6: [nit] comprise **of**

src/LibDeclarations/NativeExtensions/FeaturePoints.md
```diff
@@ -0,0 +1,61 @@
+# Feature point cloud
+## Introduction
+This document proposes an API for exposing Feature Point Clouds for WebXR.
+
+## Overview
+The API is synchronous, and on demand.  When served a new XRFrame the consuming app may query for the current feature point cloud for a given frame. As feature points comprise a large set of raw data it is best not to expose this to the JS application unless needed for app processing, so feature points must be requested handled in an on-demand fashion in order to reduce the amount of data presented to the consuming app, which must set its own frequency for updating its world understanding.
+
+The returned array will contain the feature point cloud tracked in the current frame, including points that may not be currently visible.  Feature points will be presented in a flat array of numbers where each point is presented as a set of 5 entries representing the X, Y, Z position in world space, the confidence value, and an ID that is durable across frames.
```
hendryems, line 8:
```Feature points will be presented in a flat array of numbers where each point is presented as a set of 5 entries representing the X, Y, Z position in world space, the confidence value, and an ID that is durable across frames.
```

The above is slightly confusing -- I guess part of it is my lack of deep knowledge between `XRFrame` vs. `WebXRFeaturePointSystem`. I understand that client would be using `WebXRFeaturePointSystem` and will be working with array of IWebXRFeaturePoint. Is XFrame just implementation details then?

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,156 @@
...
+     * The (Babylon) version of this module.
+     * This is an integer representing the implementation version.
+     * This number does not correspond to the WebXR specs version
+     */
+    public static readonly Version = 1;
+    /**
+     * Observers registered here will be executed whenever new feature points are available (on XRFrame while the session is tracking).
+     * Will notify the observers about which feature points have been updated.
+     */
+    public onFeaturePointsUpdatedObservable: Observable<number[]> = new Observable();
```
hendryems, line 49: [API design] We'll want to surface up the following changes notification types:
- Feature Points Added
- Feature Points Updated
- Feature Points Removed

So I think this should be 3 number[].

src/XR/features/WebXRFeaturePointSystem.ts
hendryems: JS noob here but from an API design perspective, we should make it safe and intuitive to use. Because otherwise, I would expect the expected usage here is that I have to iterate through array to find the ID of feature point that actually got added/updated making this a On2 operation.

Making a dictionary of <id, feature point> would make the use case of simply iteration more cumbersome though. Are there any practical implications in JS here?

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,156 @@
...
+     */
+    protected _onXRFrame(frame: XRFrame) {
+        if (!this.attached || !this._enabled || !frame) {
+            return;
+        }
+        let featurePointRawData : number[] | undefined = frame.featurePointCloud;
+        if (!featurePointRawData || featurePointRawData.length == 0) {
+            return;
+        } else {
+            let numberOfFeaturePoints : number = featurePointRawData.length / 5;
```
hendryems, line 106:
```
featurePointRawData.length / 5
```

Is there anything on JS to 'assert' length is multiples of 5?

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,156 @@
...
+            let updatedFeaturePoints = new Array(numberOfFeaturePoints);
+            for (var i = 0; i < numberOfFeaturePoints; i++) {
+                let rawIndex : number = i * 5;
+                let id = featurePointRawData[rawIndex + 4];
+                updatedFeaturePoints[i] = id;
+
+                // IDs should be durable across frames and strictly increasing from 0 up, so use them as indexing into the feature point array.
+                if (id == this._featurePointCloud.length) {
+                    this._featurePointCloud.push({
```
hendryems, line 115: I don't see where feature points get removed -- is that out of scope for this particular change?

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,156 @@
...
+            return;
+        }
+        let featurePointRawData : number[] | undefined = frame.featurePointCloud;
+        if (!featurePointRawData || featurePointRawData.length == 0) {
+            return;
+        } else {
+            let numberOfFeaturePoints : number = featurePointRawData.length / 5;
+            let updatedFeaturePoints = new Array(numberOfFeaturePoints);
+            for (var i = 0; i < numberOfFeaturePoints; i++) {
+                let rawIndex : number = i * 5;
```
ryantrem, line 109:
```suggestion
                const rawIndex: number = i * 5;

```

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,156 @@
...
+                let id = featurePointRawData[rawIndex + 4];
```
ryantrem, line 110:
```suggestion
                const id = featurePointRawData[rawIndex + 4];

```

src/XR/webXRFeaturesManager.ts
```diff
@@ -77,6 +77,10 @@ export class WebXRFeatureName {
      * The name of the teleportation feature
      */
     public static TELEPORTATION = "xr-controller-teleportation";
+    /**
+     * The name of the feature points feature.
+     */
+    public static FEATURE_POINTS = "xr-feature-points";
```
ryantrem, line 83: Presumably this (and others) should be `readonly`?

src/XR/features/WebXRFeaturePointSystem.ts
Alex-MSFT: Yes for now we don't remove stale feature points (this is not something ARCore does) we'll need to come up with our own notion of a stale/removed feature point.

src/XR/features/WebXRFeaturePointSystem.ts
Alex-MSFT: I think from a practical implementation it's about the same.  You're correct iterating through feature points themselves would be more cumbersome, and I think there's a little bit of an efficiency overhead we save by using arrays, but I'm not too sure myself.  I'll leave this question open for now for more opinions.

src/LibDeclarations/NativeExtensions/FeaturePoints.md
Alex-MSFT: From the point of view of the consumer of BabylonJS yes the XRFrame is an implementation detail.  Basically the entire WebXR surface area is abstracted by Babylonjs and presented through various modules that the user attaches and observes.

In this case the WebXRFeaturePointSystem is an abstraction on the XRFrame's newly exposed featurePointCloud API.

src/XR/features/WebXRFeaturePointSystem.ts
Alex-MSFT: We can definitely split this into added, and updated though right now we don't actually handle removed as this is not something surfaced by ARCore or ARKit as far as I can tell we just stop receiving updates for that point.  We would need to create our own definition of a "removed" feature point.

src/XR/webXRFeaturesManager.ts
Alex-MSFT: I think that's probably true I was following what was already in place here @RaananW thoughts?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -0,0 +1,11 @@
+// This file contains native only extensions for WebXR  These APIs are not supported in the browser yet.
+// They are intended for use with either Babylon Native https://github.com/BabylonJS/BabylonNative or
+// Babylon React Native: https://github.com/BabylonJS/BabylonReactNative
+
+interface XRSession {
```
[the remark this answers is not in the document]
Alex-MSFT: These just get auto-merged with the interfaces from webxr.d.ts, so I don't think we need an intersection type.

chrisfromwork:
At a high level, should this functionality live in Babylon.JS/NativeXr or should this extension be added to a custom repo/native plugin? It seems like the choice of adding this feature here may be done out of convenience.

Thus far, I have not seen any examples of Babylon.js logic for BabylonNative that has not had a parallel mapping to web/webxr. It seems like this sets a new precedent and may detract from the developer assurance that Babylon.js code will work on the web. Would it make more sense if these javascript files were included in the BabylonNative repo, that way they were only referenceable if someone was actually building a native experience?

Alex-MSFT:
> At a high level, should this functionality live in Babylon.JS/NativeXr or should this extension be added to a custom repo/native plugin? It seems like the choice of adding this feature here may be done out of convenience.
> 
> Thus far, I have not seen any examples of Babylon.js logic for BabylonNative that has not had a parallel mapping to web/webxr. It seems like this sets a new precedent and may detract from the developer assurance that Babylon.js code will work on the web. Would it make more sense if these javascript files were included in the BabylonNative repo, that way they were only referenceable if someone was actually building a native experience?

I think this is definitely open for debate.  Right now there are various features that are supported on only on specific web browsers which we have as part of BabylonJS.  In general I wanted this feature to serve as a first step towards giving us a fast track to unlocking native only extensions without having to go through the full WebXR proposal process, so I definitely want this to be part of the conversation.

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * On receiving a new XR frame if this feature is attached notify observers new feature point data is available.
+     */
+    protected _onXRFrame(frame: XRFrame) {
+        if (!this.attached || !this._enabled || !frame) {
+            return;
+        }
+        let featurePointRawData : number[] | undefined = frame.featurePointCloud;
```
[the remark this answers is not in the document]
Alex-MSFT: Going to leave this one as is, since Nullable is a Babylon defined type, and the interface is technically outside of the Babylon namespace.

chrisfromwork:
> > At a high level, should this functionality live in Babylon.JS/NativeXr or should this extension be added to a custom repo/native plugin? It seems like the choice of adding this feature here may be done out of convenience.
> > Thus far, I have not seen any examples of Babylon.js logic for BabylonNative that has not had a parallel mapping to web/webxr. It seems like this sets a new precedent and may detract from the developer assurance that Babylon.js code will work on the web. Would it make more sense if these javascript files were included in the BabylonNative repo, that way they were only referenceable if someone was actually building a native experience?
> 
> I think this is definitely open for debate. Right now there are various features that are supported on only on specific web browsers which we have as part of BabylonJS. In general I wanted this feature to serve as a first step towards giving us a fast track to unlocking native only extensions without having to go through the full WebXR proposal process, so I definitely want this to be part of the conversation.

It would be great to get a feel for whether or not we could add this to the BabylonNative codebase and have it consumed by BabylonNative applications. I feel like we can better ensure a high quality Babylon.js experience by having this extensions javascript file only picked up by BabylonNative builds. 

A bit unrealted to the code here, but I am also curious whether best practices suggest not registering type names when a WebXR/NativeXr feature is unavailable. Some examples for hand tracking that @RaananW  shared checked whether XRHand was undefined to gage whether or a feature was available. I think right now the NativeXr plugin registers all class definitions for all platforms whether or not they are supported.

src/XR/features/WebXRFeaturePointSystem.ts
Alex-MSFT: I added an explicit check and error state.

bghgary:
Converting this to draft for now to avoid accidental merge. Waiting for some discussions to happen internally.

src/XR/webXRFeaturesManager.ts
RaananW: `readonly` makes sense, of course. You can change the rest to readonly as well.

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+                const id = featurePointRawData[rawIndex + 4];
+                // IDs should be durable across frames and strictly increasing from 0 up, so use them as indexing into the feature point array.
+                if (id == this._featurePointCloud.length) {
+                    this._featurePointCloud.push({
+                        position: new Vector3(
+                             featurePointRawData[rawIndex],
+                             featurePointRawData[rawIndex + 1],
+                             featurePointRawData[rawIndex + 2]),
+                        confidenceValue: featurePointRawData[rawIndex + 3],
+                        id: id
```
RaananW, line 131: If we are talking about hundred-thousands of those points, we should probably try to be very "cheap" on space used. is the ID here absolutely needed? A developer accessing it will surely have a way to get the ID in a different way.

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+
+            const numberOfFeaturePoints : number = featurePointRawData.length / 5;
+            let updatedFeaturePoints = new Array();
+            let addedFeaturePoints = new Array();
+            for (var i = 0; i < numberOfFeaturePoints; i++) {
+                const rawIndex: number = i * 5;
+
+                const id = featurePointRawData[rawIndex + 4];
+                // IDs should be durable across frames and strictly increasing from 0 up, so use them as indexing into the feature point array.
+                if (id == this._featurePointCloud.length) {
```
RaananW, line 124: You could do something like this and save the ID comparison:

```javascript
// create a new point, setting the ID and 0-values for the rest (will be updated right after)
this._featurePointCloud[id] = this._featurePointCloud[id] || { id, confidenceValue: 0, position: new Vector() };
this._featurePointCloud[id].position.x = featurePointRawData[rawIndex];
this._featurePointCloud[id].position.y = featurePointRawData[rawIndex + 1];
this._featurePointCloud[id].position.z = featurePointRawData[rawIndex + 2];
this._featurePointCloud[id].confidenceValue = featurePointRawData[rawIndex + 3];
updatedFeaturePoints.push(id);
```

src/LibDeclarations/NativeExtensions/FeaturePoints.md
```diff
@@ -0,0 +1,69 @@
...
+The returned array will contain the feature point cloud tracked in the current frame, including points that may not be currently visible. Feature points will be presented in a flat array of numbers where each point is presented as a set of 5 entries representing the X, Y, Z position in world space, the confidence value, and an ID that is durable across frames.
+
+## Feature Points - API Changes
+The below details the changes to the existing API to allow querying for feature points.
+
+XRFrame will expose a new attribute featurePointCloud which returns a set that includes all feature points present in the given frame:
+```typescript
+interface XRSession {
+  ...
+  trySetFeaturePointCloudEnabled(enabled: boolean): boolean;
```
RaananW, line 17: should that return a promise? are there no async-porcesses when enabling it?

src/XR/features/WebXRFeaturePointSystem.ts
Alex-MSFT: Makes sense, we can remove ID here since we are getting it by id reference in the first place.

src/LibDeclarations/NativeExtensions/FeaturePoints.md
Alex-MSFT: There is not, but I don't think that's necessary.  From the BabylonNative perspective we are just fulfilling the extended WebXR contract which might differ between supported devices.  From a feature perspective this doesn't really differ from other features on the BabylonNative side in any meaningful way.

src/LibDeclarations/NativeExtensions/FeaturePoints.md
Alex-MSFT: Right now it does not do anything async it just tells the session to start collecting them similar to the updateWorldTrackingState call for plane detection.

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     * Dispose this feature and all of the resources attached
+     */
+    public dispose(): void {
+        super.dispose();
+
+        this._featurePointCloud.length = 0;
+        this.onFeaturePointsUpdatedObservable.clear();
```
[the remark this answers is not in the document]
Alex-MSFT: Thanks fixed!

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+     */
+    id : number;
+}
+
+/**
+ * The feature point system is used to detect feature points from real world geometry.
+ * This feature is currently experimental and only supported on BabylonNative, and should not be used in the browser.
+ * The newly introduced API can be seen in webxr.nativeextensions.d.ts and described in FeaturePoints.md.
+ */
+export class WebXRFeaturePointSystem extends WebXRAbstractFeature {
```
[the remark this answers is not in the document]
Alex-MSFT: good idea, added

src/XR/features/WebXRFeaturePointSystem.ts
Alex-MSFT: I need to be able to separate them into added an updated feature points so I still need that first safety check, but I do like this pattern of instantiating it and not having two different sets of code populating the values.

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * On receiving a new XR frame if this feature is attached notify observers new feature point data is available.
+     */
+    protected _onXRFrame(frame: XRFrame) {
+        if (!this.attached || !this._enabled || !frame) {
+            return;
+        }
+
+        const featurePointRawData: number[] | undefined = frame.featurePointCloud;
+        if (!featurePointRawData || featurePointRawData.length == 0) {
```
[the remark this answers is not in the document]
Alex-MSFT: Thanks.  My fingers are too used to doing == I keep forgetting to do ===.

src/XR/features/WebXRFeaturePointSystem.ts
```diff
@@ -0,0 +1,176 @@
...
+    public readonly onFeaturePointsAddedObservable: Observable<number[]> = new Observable();
+    /**
+     * Observers registered here will be executed whenever a feature point has been updated (on XRFrame while the session is tracking).
+     * Will notify the observers about which feature points have been updated.
+     */
+    public readonly onFeaturePointsUpdatedObservable: Observable<number[]> = new Observable();
+    /**
+     * The currrent feature point cloud maintained across frames.
+     */
+    public readonly featurePointCloud: Array<IWebXRFeaturePoint> = this._featurePointCloud;
```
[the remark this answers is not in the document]
Alex-MSFT: Done

deltakosh:
@RaananW  ok to merge?

RaananW:
Ok to merge :-)

PR #8678, opened 2020-08-03
Light particles

deltakosh:


src/Particles/IParticleSystem.ts
```diff
@@ -12,6 +10,8 @@ import { Observable } from "../Misc/observable";
 
 declare type Animation = import("../Animations/animation").Animation;
 
+import { AbstractMesh } from '../Meshes/abstractMesh';
```
sebavan, line 13: would the declare type be enough ???

PR #8680, opened 2020-08-04
gLTF loader: Added the loadAllMaterials property

### LINKED ISSUE #6818: Load standalone materials from GLTF
# Feature request

Forum Link: https://forum.babylonjs.com/t/standalone-gltf-materials/5481/2
- Load standalone materials from a gltf file.

Example GLTF file:
{
“asset”: {
“generator”: “COLLADA2GLTF”,
“version”: “2.0”
},
“materials”: [
{
“pbrMetallicRoughness”: {
“baseColorFactor”: [
0.800000011920929,
0.0,
0.0,
1.0
],
“metallicFactor”: 0.0
},
“name”: “Red”
}
]
}

Popov72:
Fix #6818

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -309,6 +309,39 @@ export class GLTFLoader implements IGLTFLoader {
                 promises.push(this.loadSceneAsync(`/scenes/${scene.index}`, scene));
             }
 
+            if (this.parent.loadAllMaterials && this._gltf.materials) {
+                for (let m = 0; m < this._gltf.materials.length; ++m) {
+                    const material = this._gltf.materials[m];
+                    const context = "/materials/" + m;
+                    const babylonDrawMode = Material.TriangleFillMode;
+                    let babylonData = material._data ? material._data[babylonDrawMode] : null;
+
+                    if (babylonData) {
+                        continue;
+                    }
+
+                    this.logOpen(`${context} ${material.name || ""}`);
+
+                    const babylonMaterial = this.createMaterial(context, material, babylonDrawMode);
+
+                    babylonData = {
+                        babylonMaterial: babylonMaterial,
+                        babylonMeshes: [],
+                        promise: this.loadMaterialPropertiesAsync(context, material, babylonMaterial)
+                    };
+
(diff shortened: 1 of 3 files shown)


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -309,6 +309,39 @@ export class GLTFLoader implements IGLTFLoader {
...
+
+                    material._data = material._data ?? {};
+                    material._data[babylonDrawMode] = babylonData;
+
+                    GLTFLoader.AddPointerMetadata(babylonMaterial, context);
+                    this._parent.onMaterialLoadedObservable.notifyObservers(babylonMaterial);
+
+                    this.logClose();
+                }
+            }
```
bghgary, line 343: I understand why you did this, but it does duplicate a lot of code. Might be good to factor out somehow or find a different way. This change might have some weird interactions with other flags (e.g. `compileMaterials`) and extensions like MSFT_lod.

Popov72: I can factor code indeed.

Regarding interactions with `compileMaterials` it's ok, this function is iterating over `babylonData.babylonMeshes` to compile the material for a given mesh and this array is empty in the code above.

For interactions with other extensions I'm afraid I'm not knowledgeable enough so what do you recommend?

bghgary: > Regarding interactions with `compileMaterials` it's ok, this function is iterating over `babylonData.babylonMeshes` to compile the material for a given mesh and this array is empty in the code above.

That is actually my problem with it. The `compileMaterials` flag is supposed to make sure the materials are ready for use before completing the load. Since we don't know what materials these will be assigned to, it might be good to disallow both of these flags to be on at the same time. I'm not sure though.

> For interactions with other extensions I'm afraid I'm not knowledgeable enough so what do you recommend?

At a minimum, we are no longer calling `loadExtensions` for the materials, so none of the materials from this code path will load material extensions. Once that's fixed, if this material has, say, a `MSFT_lod` extension, I have no idea what that will do.

Popov72: See #8704

PR #8684, opened 2020-08-04
First pass of KHR_materials_volume loader

MiiBond:
This PR includes absorption and thickness logic for the KHR_materials_volume extension.


deltakosh:
Do we have any here?

RaananW:
@MiiBond - any update to this PR?

sebavan:
@RaananW this will wait after 4.2 :-) but in progress.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,131 @@
...
+            pbrMaterial.subSurface.tintColor = new Color3(1.0, 1.0, 1.0);
+            pbrMaterial.subSurface.tintColor.copyFromFloats(extension.attenuationColor[0], extension.attenuationColor[1], extension.attenuationColor[2]);
+        } else {
+            pbrMaterial.subSurface.tintColor = new Color3(1.0, 1.0, 1.0);
+        }
+        if (extension.subsurfaceColor !== undefined && extension.subsurfaceColor.length == 3) {
+            const sssColour = new Color3(0.0, 0.0, 0.0);
+            sssColour.copyFromFloats(extension.subsurfaceColor[0], extension.subsurfaceColor[1], extension.subsurfaceColor[2]);
+            if (sssColour != Color3.Black()) {
+                // This logic may be hard to follow here:
```
MiiBond, line 91: This logic seems a little odd so it's worth reviewing. The current intent in the volume extension is that we can use the existence of `KHR_materials_translucency` as a hint that the renderer can use the diffusion approximation for SSS.

src/Materials/Node/Blocks/PBR/subSurfaceBlock.ts
```diff
@@ -146,7 +146,7 @@ export class SubSurfaceBlock extends NodeMaterialBlock {
 
         #ifdef SUBSURFACE
             vec2 vThicknessParam = vec2(0., ${thickness});
-            vec4 vTintColor = vec4(${tintColor}, ${refractionTintAtDistance});
+            vec4 vTintColor = vec4(toLinearSpace(${tintColor}), ${refractionTintAtDistance});
```
MiiBond, line 149: The tint colour needs to be linearized but my change here doesn't seem to accomplish that. @sebavan am I doing something silly here?

src/Shaders/ShadersInclude/pbrBlockFinalLitComponents.fx
```diff
@@ -34,7 +34,7 @@
     #endif
 
     #if defined(SS_TRANSLUCENCY)
-        finalIrradiance *= (1.0 - subSurfaceOut.transmittance);
+        finalIrradiance *= (1.0 - vSubSurfaceIntensity.y);
```
MiiBond, line 37: The previous code results in all non-transmitted light to be reflected diffusely but this isn't really what we want, I don't think. If translucency is fully on, all non-transmitted light should be absorbed, not reflected. Therefore, I think the right thing to do here is multiply by `1.0 - translucencyIntensity`

src/Materials/Node/Blocks/PBR/subSurfaceBlock.ts
Popov72: The `tintColor` must not be linearized here: it's the responsibility of the user to linearize it in the NME (there is a switch on the texture block for that, as well as a switch on the color3/color4 block, so it's easy to do).

Popov72: Also, the NME sub surface block will need to be updated to handle the new `surfaceAlbedo` parameter passed to the `subSurfaceBlock(...)` shader call.

Also, (optionally?) the `SS_ALBEDOFORTRANSLUCENCYTINT` could be handled as a new switch on the SubSurface block.

MiiBond: Thanks, @Popov72 .

src/Shaders/ShadersInclude/pbrBlockFinalLitComponents.fx
Popov72: Don't know for this one, will let @sebavan answer.

However, I would think that `translucency is fully on` means the material is fully transparent (https://en.wikipedia.org/wiki/Transparency_and_translucency), so transmittance=1 and then there's no light reflected diffusely => `1.0 - subSurfaceOut.transmittance` would be the right factor in that case.

sebavan: Not sure as well,

but like @Popov72  1 - transmittance sounds like better here. I agree that some part should be absorbed though but not sure what amount. I do not think that we should anyway use intensity here cause it would prevent any parts like ears or nostrils where both are heavily at play.

sebavan: Maybe we need a special value like transmittance e.g. absorbance and multiply as well by 1 - absorbance in case of SSS ???

MiiBond: The issue is that `transmittance` takes absorption into account so `1.0 - transmittance` is the amount of light not absorbed. This means that that light is reflected instead and a 100% translucent material reflects a lot of light.
`translucencyIntensity` is the percentage of light that passes into the surface so I think it's the right value to use here.

Here is an example of a 100% translucent material with a red tintColor:
With `1.0 - transmittance`:
Screen Shot 2020-11-10 at 11 55 08 AM

With `1.0 - translucentIntensity`:
Screen Shot 2020-11-10 at 11 53 18 AM

The second one is what I would expect.

sebavan: Could you share how it would look on the head you used previously ? just for comparison purpose ?

sebavan: I actually think you might be right @MiiBond Let s see on the head what it loooks and use this into the release if it looks ok ?

sebavan: @Popov72 , @MiiBond we have only tomorrow left for the change we put in 4.2 would be great to decide which one to put in to prevent adding an extra back compat flag ? I propose @MiiBond you check on the previous head but my guess is it will be similar to before ? actually, at least it should preserve energy :-)

And then we can push ?

Popov72: I will let you judge as I don't know what a 100% translucency material should do (on my mind it should be fully transparent but obviously it is not in the screenshots) :)

sebavan: Ok after a few test the translucency intensity looks better, tried on a pure sphere and the other method looks a bit weird. We ll revisit if needed in 5.0, but I ll make a PR for 4.2 doing what Mike proposed here but reusing our computed intensity from the maps. PR should be here shortly.

sebavan: https://github.com/BabylonJS/Babylon.js/pull/9390

MiiBond: Awesome! Thanks.

emackey:
KHR_materials_volume has been updated to remove subsurfaceColor.  At some future point there will be subsurface scattering that involves the volume extension along with future SSS extension(s).  But for the near-term release, volume will be: thickness, attenuationDistance, and attenuationColor, used only for refraction (not scattering).

MiiBond:
@bghgary @sebavan Now that the volume extension is dropping scattering, we should look at finally merging this.

Popov72:
@MiiBond Do you have some assets you can share that you use to test your PR?

emackey:
I'm not set up to build a branch of Babylon, but here's Sketchfab's Mosquito in Amber, hand-edited to use the new PBR Next extensions.

Ideally, if it works, the mosquito should distort inside the amber.  Bonus points if it looks like it does on Sketchfab.  I may have picked a different IOR value than what Sketchfab is using:  I set it at 1.55 which is what the Internet thinks the IOR of amber is.

[MosquitoInAmber_withRefraction.zip](https://github.com/BabylonJS/Babylon.js/files/5889898/MosquitoInAmber_withRefraction.zip)

bghgary:
@Popov72 I got some asset from @MiiBond a few days ago. I'll share them with you.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    /** @hidden */
+    public dispose() {
+        (this._loader as any) = null;
+    }
+
+    /** @hidden */
+    public loadMaterialPropertiesAsync(context: string, material: IMaterial, babylonMaterial: Material): Nullable<Promise<void>> {
+        return GLTFLoader.LoadExtensionAsync<IMaterialsTransmission>(context, material, this.name, (extensionContext, extension) => {
+            console.log(extensionContext);
```
Popov72, line 55: nit: remove console.log

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+        let pbrMaterial = babylonMaterial as PBRMaterial;
+
+        // If transparency isn't enabled already, this extension shouldn't do anything.
+        // i.e. it requires either the KHR_materials_transmission or KHR_materials_translucency extensions.
+        if (!pbrMaterial.subSurface.isRefractionEnabled && !pbrMaterial.subSurface.isTranslucencyEnabled || !extension.thicknessFactor) {
+            return Promise.resolve();
+        }
+
+        // Since this extension models thin-surface transmission only, we must make IOR = 1.0
```
Popov72, line 76: Wrong comment (coming from KHR_materials_transmission)

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,103 @@
...
+        pbrMaterial.subSurface.volumeIndexOfRefraction = pbrMaterial.indexOfRefraction;
+        const attenuationDistance = extension.attenuationDistance !== undefined ? extension.attenuationDistance : Number.MAX_VALUE;
+        pbrMaterial.subSurface.tintColorAtDistance = attenuationDistance;
+        if (extension.attenuationColor !== undefined && extension.attenuationColor.length == 3) {
+            pbrMaterial.subSurface.tintColor = new Color3(1.0, 1.0, 1.0);
```
Popov72, line 81: nit: no need for this line as the `tintColor` is always initialized with a `Color3.White()` call.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,103 @@
...
+            pbrMaterial.subSurface.tintColor.copyFromFloats(extension.attenuationColor[0], extension.attenuationColor[1], extension.attenuationColor[2]);
+            pbrMaterial.subSurface.tintColor = pbrMaterial.subSurface.tintColor.toLinearSpace();
+        } else {
+            pbrMaterial.subSurface.tintColor = new Color3(1.0, 1.0, 1.0);
```
Popov72, line 85: nit: as above, you can just use `set` or `copyFromLoats` instead of creating a new `Color3`.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        pbrMaterial.subSurface.minimumThickness = 0.0;
+        pbrMaterial.subSurface.maximumThickness = extension.thicknessFactor !== undefined ? extension.thicknessFactor : 1;
```
Popov72, line 89: nit: `pbrMaterial.subSurface.maximumThickness = extension.thicknessFactor ?? 1;` takes less characters ;)

src/Materials/Node/Blocks/PBR/subSurfaceBlock.ts
```diff
@@ -227,6 +229,8 @@ export class SubSurfaceBlock extends NodeMaterialBlock {
                 #ifdef SS_LINKREFRACTIONTOTRANSPARENCY
                     alpha = subSurfaceOut.alpha;
                 #endif
+            #elif defined(SS_TRANSLUCENCY)
+                surfaceAlbedo = subSurfaceOut.surfaceAlbedo;
```
Popov72, line 233: You don't need to add those two lines because the TRANSLUCENCY path in the subsurface block does not update the albedo color (whereas the REFRACTION path does).

Popov72 (dismissed):
LGTM, only nits!

I think we will also need to add a validation test for this: probably with the KHR_volume_testing.glb file?

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
emackey: If the `thicknessFactor` is undefined in `KHR_materials_volume`, the default is zero, which is a special value that indicates the material is thin-walled (not volumetric).  It should behave as if the extension were not present.  (Someday, this will be a core material property, and you won't be able to test for the extension's presence).

MiiBond: I'll just change this to `pbrMaterial.subSurface.maximumThickness = extension.thicknessFactor`
Looks like I'm already doing an early-out when `thicknessFactor` is undefined or 0.

src/Materials/Node/Blocks/PBR/subSurfaceBlock.ts
MiiBond: Oops. That was left over from some of the scattering approximation stuff. Thanks!

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
MiiBond: Done. Thanks.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
MiiBond: Done. Thanks.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
MiiBond: Fixed. Thanks.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
MiiBond: Done. Thanks.

Popov72:
Ok for me, only missing:

> I think we will also need to add a validation test for this: probably with the KHR_volume_testing.glb file?

MiiBond:
@Popov72, how should I go about doing that? Should the volume test GLB be added to the BabylonJS/Assets repo and then create a Playground that loads from there?

Popov72:
Yes, that would be the best thing to do: add the .glb to BabylonAssets/meshes/ and create a playground to be used in the validation test suite.

Thanks!

MiiBond:
@Popov72 I opened a PR for a test asset this morning.
https://github.com/BabylonJS/Assets/pull/27

Popov72:
Pinging @deltakosh, I don't know how to publish this.

deltakosh:
Merged! (for the asset I mean)

src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
```diff
@@ -195,6 +195,10 @@ struct subSurfaceOutParams
             float refractionLOD = getLodFromAlphaG(vRefractionMicrosurfaceInfos.x, alphaG);
         #endif
 
+        // Scale refractionLOD with IOR so that an IOR of 1.0 results in no microfacet refraction and
+        // an IOR of 1.5 results in the default amount of microfacet refraction.
+        refractionLOD = mix(refractionLOD, 0.0, clamp(vRefractionInfos.y * 3.0 - 2.0, 0.0, 1.0));
```
Popov72, line 200: I don't understand this correction.

`refractionLOD` is the mipmap level to use to sample the `refractionSampler`, why would it be related to IOR? In my understanding, `refractionLOD` should only be related to the alphaG/roughness of the material and IOR is used to compute the refraction vector.

MiiBond: The blurred light transmission is caused by micro-facet refractions that we model as roughness. As the IOR gets close to 1, the amount of refraction decreases to nothing (assuming the surrounding medium is air). A piece of rough diamond "blurs" the light transmitted through it more than a piece of glass of the same roughness.
Now, that said, my approximation is a simple linear scale which may not be as correct as it can be. The previous micro-refraction amount was already very approximate, of course.

Popov72: Ok, so you make a link between blurriness and IOR: the higher the IOR, the higher the blurriness. Is that really true? Shouldn't we instead use a new material property to model this blurriness? It seems in Arnold they name that "Extra roughness": https://docs.arnoldrenderer.com/display/A5AFMUG/Transmission#Transmission-ExtraRoughness (they have also other parameters that seem to impact the bluriness, like "dispersion abbe" or "transmission scatter").

Also, I'm concerned that it is a breaking change, so we should be careful and make sure it was a bug and/or add a switch to enable/disable this feature. @sebavan thoughts?

emackey: It's really true, and the effect is reproduced in various path tracers including Enterprise PBR tracers and Blender Cycles.  The blurriness comes from not just the intensity of the microfacet roughness, but also the amount by which those microfacets are capable of distorting the light.  In the case of IOR 1.0, the light is able to cross the boundary at any angle and not change its vector at all, so the microfacets have no capability to produce any blurriness at all in the transmission, only on the specular reflection.

I made a test model:  [TransmissionRoughnessTest.zip](https://github.com/BabylonJS/Babylon.js/files/5958739/TransmissionRoughnessTest.zip)

If you import this into Blender though, you'll have to manually reassign the IOR values for about 5 materials.

Popov72: Ok, thanks for the explanation, will let @sebavan answer about this one.

sebavan: Yup love it I d consider it a fix but the linear approach at the LOD level sounds weird knowing our LODs are not stored linearly. I wondered if it should be applied to alphaG instead or better at the perceptual roughness level which is the only linear value ?

emackey: @MiiBond Do you know?

MiiBond: Yes, that might be more correct. I'll look into this more this week.

MiiBond: Okay, I updated my logic to scale the roughness or alpha depending on the IOR. Results look closer to what Dimension renders for high roughness with low IOR. Of course, the screen space refraction is kind of hard to compare to a pathtracer :)

MiiBond: @sebavan when you guys get a chance, this could be checked again.

MiiBond:
For reference, here's the new glTF sample model that @emackey made, running in Babylon with this PR.
Screen Shot 2021-03-08 at 10 09 42 AM
The model can be found here: 
https://github.com/KhronosGroup/glTF-Sample-Models/tree/transmission-roughness/2.0/TransmissionRoughnessTest

src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
sebavan: Looks pretty good to me, @Popov72 could you do a last check so that we can merge it ?

@bghgary any issues you foresee with merging ?

Popov72: LGTM

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,98 @@
...
+        pbrMaterial.subSurface.tintColorAtDistance = attenuationDistance;
+        if (extension.attenuationColor !== undefined && extension.attenuationColor.length == 3) {
+            pbrMaterial.subSurface.tintColor.copyFromFloats(extension.attenuationColor[0], extension.attenuationColor[1], extension.attenuationColor[2]);
+            pbrMaterial.subSurface.tintColor = pbrMaterial.subSurface.tintColor.toLinearSpace();
+        }
+
+        pbrMaterial.subSurface.minimumThickness = 0.0;
+        pbrMaterial.subSurface.maximumThickness = extension.thicknessFactor;
+        if (extension.thicknessTexture) {
+            return this._loader.loadTextureInfoAsync(context, extension.thicknessTexture)
```
bghgary, line 86: The context here should be the extension context plus the property name for the thickness texture.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,98 @@
...
+            promises.push(this._loadTransparentPropertiesAsync(context, material, babylonMaterial, extension));
+            return Promise.all(promises).then(() => { });
+        });
+    }
+
+    private _loadTransparentPropertiesAsync(context: string, material: IMaterial, babylonMaterial: Material, extension: IMaterialsTransmission): Promise<void> {
+        if (!(babylonMaterial instanceof PBRMaterial)) {
+            throw new Error(`${context}: Material type not supported`);
+        }
+        let pbrMaterial = babylonMaterial as PBRMaterial;
```
bghgary, line 66: nit: you can just use babylonMaterial directly since TypeScript will automatically figure out it is a PBRMaterial based on the if check above.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -478,13 +478,16 @@ export class DeviceInputSystem implements IDisposable {
...
                 for (let i = 0; i < pointer.length; i++) {
-                    if (pointer[i][PointerInput.LeftClick] === 1) {
+                    if (pointer[i]?.[PointerInput.LeftClick] === 1) {
                         pointer[i][PointerInput.LeftClick] = 0;
 
                         if (this.onInputChanged) {
                             this.onInputChanged(DeviceType.Touch, i, PointerInput.LeftClick, 1, pointer[i][PointerInput.LeftClick]);
                         }
+
+                        this._unregisterDevice(DeviceType.Touch, i);
```
bghgary, line 490: Is this a merge error?

bghgary (commented):
small comments. looks good otherwise.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
MiiBond: Done

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
MiiBond: Done

src/DeviceInput/deviceInputSystem.ts
MiiBond: Must be. Though I didn't intentionally change that file so I'm not sure how this happened...

MiiBond: Fixed.

loaders/src/glTF/2.0/Extensions/KHR_materials_volume.ts
```diff
@@ -0,0 +1,97 @@
...
+        babylonMaterial.subSurface.tintColorAtDistance = attenuationDistance;
+        if (extension.attenuationColor !== undefined && extension.attenuationColor.length == 3) {
+            babylonMaterial.subSurface.tintColor.copyFromFloats(extension.attenuationColor[0], extension.attenuationColor[1], extension.attenuationColor[2]);
+            babylonMaterial.subSurface.tintColor = babylonMaterial.subSurface.tintColor.toLinearSpace();
+        }
+
+        babylonMaterial.subSurface.minimumThickness = 0.0;
+        babylonMaterial.subSurface.maximumThickness = extension.thicknessFactor;
+        if (extension.thicknessTexture) {
+            return this._loader.loadTextureInfoAsync(`${context}/thicknessTexture`, extension.thicknessTexture)
```
bghgary, line 85: Sorry, I wasn't very clear before. This is still the context of the material. You need the `extensionContext` from the parent function.

MiiBond: Gotcha. Done.

bghgary (approved):
Looks good to me from the glTF side.

sebavan (approved):

Popov72:
This PG (which is the validation test `PBR shader code coverage 2`) does not work: https://playground.babylonjs.com/#QI7TL3#17

That's because it's `roughness` and not `alphaG` which is passed to the sub surface block function:
```javascript
        #ifdef SS_LINEARSPECULARREFRACTION
            const in float roughness,
        #else
            const in float alphaG,
        #endif
```
As you now need `alphaG` in all cases, I think you can move it outside the #if block.

Popov72:
It still does not work because you have to account for the change in other files (`pbr.fragment.fx` and `subSurfaceBlock.ts`).

You can check locally that the test is ok with this url: http://localhost:1338/Playground/index-local.html#QI7TL3#17

Popov72:
LGTM!

PR #8688, opened 2020-08-05
Camera Gizmo with Frustum display

CedricGuillemet:
cameraGizmo with Inspector integration.
image
I've tested as much as I could but a second tester is hugely welcome. There are many code paths changed in this PR and I might have missed something. I'm also not a huge fan of merging PR the day before going on vacation.
Anyway. cameraGizmo supports perspective/ortho view. Frustum display is optional and fully dynamic (try to change fov or resize the window while the gizmo is visible)
I also did some changes on the gizmomanager to support nodes instead of just meshes. I kept the attachmesh function to preserve back compatibility.
I'm not sure I did the right thing in sceneTreeItemComponent.tsx  for 'Record movement for generating replay code'. Let me know.

The PG I used for testing : https://www.babylonjs-playground.com/#IR7LVT


deltakosh:
Nice!

deltakosh:
There is a conflict

deltakosh:
A few feedback:
- The clipping in the layer scene is too small:
image

- margin is not big enough between camera and eye icon:
image

deltakosh:
(btw this can wait after your vacations ;))

deltakosh:
Will do it;)

PR #8694, opened 2020-08-05
Texture inspector: allows tools to use post processes

DarraghBurkeMS:
This PR enables tools to use post processes. It also introduces a new tool, the contrast and exposure tool, which demonstrates this functionality using the ImageProcessing post process.

Img

It also fixes several other issues:
- Tools may now register their own UI. The inspector also provides a draggable window which tools can add their own UI to, making it very easy to add UI elements.
- Individual channel viewing is now implemented using a shader rather than CPU processing. This should make editing large textures significantly faster. The snippet server dependency has also been removed, so the texture inspector should work even without network connection.
- If you have a texture inspector window open, and you try to inspect another texture, the original window will close and a new window will open with the new texture.
- The brush tool now uses the canvas lineTo() function as opposed to just placing circles at the cursor. This produces a much smoother painting effect.
- The toolbar has been disabled for cubemaps since editing cubemaps is not supported.
- The tools API now provides access to the BABYLON namespace, allowing external tools to reference Babylon objects. (For instance, if you want to lookup values of PointerEventTypes, you can do that now.)


inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/defaultTools/Eyedropper.ts
```diff
@@ -1,60 +1,59 @@
...
-    MS4yOGwtLjgxLDMuMjYtLjU2LjU2YTEuMTYsMS4xNiwwLDAsMCwwLDEuNjVsLjQxLjQxYTEuMTcsMS4xNywwLDAsMCwxLjY1LDBsLjU2LS41NiwzLjI2LS44MWEyLjc1LDIuNzUsMCwwLDAsMS4yOC0uNzNsNi4xNC02LjE0YTEuODcsMS44NywwLDAsMCwuODQuMjEs
-    MS44MywxLjgzLDAsMCwwLDEuMzMtLjU1bC45LS45YTEuODcsMS44NywwLDAsMCwuMDgtMi41NywxMS41NCwxMS41NCwwLDAsMCwyLjMyLTEuNjZDMzAuMzIsMTQuNzQsMzEsMTIuMzUsMjkuMzIsMTAuNjhaTTE2LjE1LDI2Ljc5YTEuMjEsMS4yMSwwLDAsMS0uNTgu
-    MzNMMTIsMjhsLjktMy41OWExLjIxLDEuMjEsMCwwLDEsLjMzLS41OGw2LjA3LTYuMDcsMi45NCwyLjk0Wm05LjIxLTcuMzgtLjkuOWMtLjE5LjItLjM0LjItLjU0LDBsLTQuNC00LjRhLjQuNCwwLDAsMSwwLS41NGwuOS0uOWEuNDMuNDMsMCwwLDEsLjI3LS4xMS4z
-    OS4zOSwwLDAsMSwuMjcuMTFsNC40LDQuNEEuMzguMzgsMCwwLDEsMjUuMzYsMTkuNDFabTMuMzgtNS45M2EzLjcsMy43LDAsMCwxLTEsMS43LDExLjY3LDExLjY3LDAsMCwxLTIuMzUsMS42MkwyMy4yLDE0LjU5YTExLjY3LDExLjY3LDAsMCwxLDEuNjItMi4zNSwz
-    LjcsMy43LDAsMCwxLDEuNy0xLDEuODMsMS44MywwLDAsMSwyLjIyLDIuMjJaIiBzdHlsZT0iZmlsbDojZmZmIi8+PC9zdmc+`
+import { ToolParameters, ToolData } from '../textureEditorComponent';
```
deltakosh, line 1: nit picking but all file should start with a lower case

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/propertiesBar.tsx
```diff
@@ -1,144 +1,149 @@
...
-                        <img title='Save' className='icon button' src={saveButton} onClick={() => this.props.saveTexture()}/>
-                    </div>
-                </div>
-        </div>;
-    }
+import * as React from 'react';
+import { BaseTexture } from 'babylonjs/Materials/Textures/baseTexture';
+import { PixelData } from './textureCanvasManager';
+
+interface PropertiesBarProps {
```
deltakosh, line 5: Interface should start with I

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/propertiesBar.tsx
```diff
@@ -1,144 +1,149 @@
...
+const faces = [
+    posX,
+    negX,
+    posY,
+    negY,
+    posZ,
+    negZ
+]
+
+function PixelData(props : PixelDataProps) {
```
deltakosh, line 52: should start with _ if private (not exported)

deltakosh: And I'm not a big fan of "flying" functions, so maybe just move them as private functions inside the propertiesBar?

DarraghBurkeMS: So the PixelData component should instead be _pixelData and should be a private function of the class?

And faces/the icons should be private member variables?

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/defaultTools/Eyedropper.ts
DarraghBurkeMS: Not nitpicking at all, totally makes sense! Fixed

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/propertiesBar.tsx
DarraghBurkeMS: Done

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/propertiesBar.tsx
DarraghBurkeMS: @deltakosh I moved everything inside the class, does that look better?

DarraghBurkeMS:
@deltakosh No, thank you for the super fast review ❤️

PR #8696, opened 2020-08-05
P8 skeleton viewer try2

Pryme8:
DONE - Attach Gizmos to Bones through Inspector
DONE - Expose Sphere/Spur Sizing Parameters in Debug Menu

*IMPORTANT*
Also Id like to note a bug with the gizmo attachments.  If you do a build of this branch and drop into any scene with a skeleton with mesh is not at Zero the gizmo does not go into the correct place.  Don't think its anything that I would have done and is behavior that will need to be fixed on the gizmo manager level?
TODO - Solve noticeable fps drop when constructing or swapping to sphere/spheres & spur mode.


deltakosh:
There is one file in conflict

Pryme8:
Checking it out now.

Im trying to figure out how to get this in, and then merged with Popovs PR.

deltakosh:
you can cancel it, merge from master when I'll merge popov's one and do a new PR

Pryme8:
Sounds good

PR #8697, opened 2020-08-06
reworked bindbuffers to use a vertex array object

CedricGuillemet:
fix for https://github.com/BabylonJS/BabylonNative/issues/358
instead of trying to bind vertex stream individually with bindBuffers, a vertex array object is created and used for rendering.

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -42,7 +42,6 @@ interface INativeEngine {
     createVertexBuffer(data: ArrayBufferView, dynamic: boolean): any;
     deleteVertexBuffer(buffer: any): void;
     recordVertexBuffer(vertexArray: any, buffer: any, location: number, byteOffset: number, byteStride: number, numElements: number, type: number, normalized: boolean): void;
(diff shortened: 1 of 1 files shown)


src/Engines/nativeEngine.ts
```diff
@@ -411,7 +395,20 @@ export class NativeEngine extends Engine {
...
         }
+    }
 
+    public bindBuffers(vertexBuffers: { [key: string]: VertexBuffer; }, indexBuffer: Nullable<NativeDataBuffer>, effect: Effect): void {
+        if (this._boundBuffersVertexArray) {
+            this._native.deleteVertexArray(this._boundBuffersVertexArray);
+        }
+        this._boundBuffersVertexArray = this._native.createVertexArray();
+        this._recordVertexArrayObject(this._boundBuffersVertexArray, vertexBuffers, indexBuffer, effect);
+        this._native.bindVertexArray(this._boundBuffersVertexArray);
```
bghgary, line 406: I put a comment on the native side also. I think we probably should split this into bindVertexArray and bindIndexBuffer so that it's not as confusing. WebGL's VAO only has vertex buffers. We can fix this later though.

PR #8698, opened 2020-08-06
Fix skeleton viewer

Popov72:
See https://forum.babylonjs.com/t/return-to-rest-not-behaving-as-expected/12570/72

I have checked everything was working for all of the 11 meshes of this PG: https://playground.babylonjs.com/#EBBE0Z

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -1158,7 +1158,7 @@ export class GLTFLoader implements IGLTFLoader {
                 baseMatrix.multiplyToRef(babylonParentBone.getInvertedAbsoluteTransform(), baseMatrix);
             }
 
-            babylonBone.setTPose(baseMatrix);
+            babylonBone.setBindPose(baseMatrix);
 
             babylonBone.updateMatrix(baseMatrix, false, false);
             babylonBone._updateDifferenceMatrix(undefined, false);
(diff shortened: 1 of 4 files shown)


Pryme8:
How do I merge in this stuff to my PR? -asking for a friend.

Also is the isn't technically the tPose matrix technically doing what the restPose is supposed to be?

Just wanna make sure I understand the changes.

deltakosh:
Not sure as well what is the difference with restpose?

Pryme8:
I am hesitant to call it tPose because there is nothing that guarantees the model is in a tPose for initial rigging (it might be A pose it might be an action pose).

Id lean tword riggingPose instead of tPose if it is in fact different then the restPose.

deltakosh:
I would also want to make sure we need another property

Popov72:
I think the rest pose is the default pose the model was in when exported from the DCC tool.

Here's a comment from @Drigax for a previous PR:
```
Yeah I think I'm wrong here, my mistake, our bindPose is what we want to use to transform our mesh vertices, while our rest pose is what we want to consider as "home" for a given joint, I'd assume. where our bind pose is what its modelled and rigged as, while rest may be the toy in its exported posed position, or A pose, or whatever arbitrary pose. sounds good.
```
So, maybe TPose should be renamed `bindPose`? Or do you want it to be `riggedPose`?

It is different from the rest pose because it takes into account the inverse bind matrices.

Pryme8:
Id for sure lean tword bindPose or rigPose then so we dont have people assuming the model should be in a tPose when that data might not be there.

Popov72:
I have renamed `TPose` into `bindPose`.

deltakosh:
Ok that makes sense now! Thanks

deltakosh:
just a tiny doc issue

PR #8702, opened 2020-08-07
P8 skeleton viewer try2

Pryme8:
Merge and update with the changes that have been happening.

Still need to resolve why the gizmos are not going to the right spot on the bones, but I will have to take a look into the gizmo manager.

Also was wondering if there are any suggestions to prevent the noticeable lag when construction the SkeletonViewer in one of the newer view modes?


Popov72:
> Also was wondering if there are any suggestions to prevent the noticeable lag when construction the SkeletonViewer in one of the newer view modes?

You could create a single sphere and a single spur outside of the loop and clone + scale them when needed to avoid some vertex buffer creations (notably for the sphere, where a single geometry will be shared by all clones).

Pryme8:
I dont know if its so much the creation of the meshes (and I could not do it for the spurs as they are all different sizes and lengths) as it is when I bind the vertex weights I think. 

Ill give it a shot with the sphere, but I think having to set the weights adds to the calculations overhead.  Maybe introduce a constructor argument that will let me set the weights with an array on initialization of the spheres?

the spurs are like 16 facets and I doubt would hang the system that hard, but I might be mistaken.

Pryme8:
Would it be weird to recursively build it?  Just commit one sphere and spur to build at a time and have the function then non block the thread.

I mean you would see a cascade of the meshes appearing then, but at least it would not lock up the IO

deltakosh:
Error is a transient one..not problem

PR #8703, opened 2020-08-07
Update RayHelper to Mesh in Physics loop and make renderLine not pickable

regnaio:
RayHelper is currently updated to Mesh in scene.registerBeforeRender but not in the physics loop, e.g. onAfterStepObservable. This causes the RayHelper to lag behind the mesh in the physics loop, even if computeWorldMatrix(true) is called on the mesh in the physics loop.

The _renderLine was also made not pickable. Currently, if the RayHelper is showing, its visible line is pickable. In the case of rapid firing bullets with RayHelpers, pick hits will be registered on the RayHelper visible line.

We may also consider adding this line as well:
this._attachedToMesh.getScene().onAfterPhysicsObservable.add(this._updateToMeshFunction);

### DIFF
--- a/src/Debug/rayHelper.ts
+++ b/src/Debug/rayHelper.ts
@@ -72,6 +72,7 @@ export class RayHelper {
             this._scene = scene;
             this._renderPoints = [ray.origin, ray.origin.add(ray.direction.scale(ray.length))];
             this._renderLine = Mesh.CreateLines("ray", this._renderPoints, scene, true);
+            this._renderLine.isPickable = false;
 
             if (this._renderFunction) {
                 this._scene.registerBeforeRender(this._renderFunction);
@@ -171,6 +172,7 @@ export class RayHelper {
         if (!this._updateToMeshFunction) {
             this._updateToMeshFunction = (<() => void>this._updateToMesh.bind(this));
             this._attachedToMesh.getScene().registerBeforeRender(this._updateToMeshFunction);
+            this._attachedToMesh.getScene().onAfterStepObservable.add(this._updateToMeshFunction);
         }
 
         // force world matrix computation before the first ray helper computation


src/Debug/rayHelper.ts
```diff
@@ -171,6 +172,7 @@ export class RayHelper {
         if (!this._updateToMeshFunction) {
             this._updateToMeshFunction = (<() => void>this._updateToMesh.bind(this));
             this._attachedToMesh.getScene().registerBeforeRender(this._updateToMeshFunction);
+            this._attachedToMesh.getScene().onAfterStepObservable.add(this._updateToMeshFunction);
```
deltakosh, line 175: This is weird because technically onBeforeRender is called right after onAfterStep so this should not be required

deltakosh: Can you repro the issue in the PG? I do not want to run code multiple times if not required

regnaio: Sure! Will create a PG by early afternoon

regnaio: Please find a link to a Playground below:
https://playground.babylonjs.com/#R5CSPZ#2

Description:
Every second, a cyan bullet with a RayHelper attached is fired at a wall. If the wall is hit, it turns red and logs a message to the console.

Simultaneously, many box meshes are being rapidly added to deliberately lower the FPS. Before the FPS starts the lower, every bullet successfully registers a hit with the wall and logs a console message. However, as the FPS lowers due to the box meshes, bullets sometimes register no hits or even multiple hits. The changes in this PR resolves this.

The engine has `deterministicLockstep` enabled, which means physics and anything in `onAfterStepObservable` should yield the same results regardless of FPS.

Please find a video detailing the issue below:
https://www.youtube.com/watch?v=0YntF1lAOkA&feature=youtu.be

The first half of the video uses https://preview.babylonjs.com/babylon.js and shows the bug. The second half uses the new changes in `lib/babylon.js` and shows the bug resolved.

Thank you for your time and help, @deltakosh!

regnaio: For your reference, here is a workaround without the RayHelper changes: https://playground.babylonjs.com/#R5CSPZ#3

Added the lines:
```
bullet.computeWorldMatrix(true);
ray.origin = bullet.getAbsolutePosition();
```

This compensates for the RayHelper's position lagging behind the Mesh's position

deltakosh: Ok so I understand. My recommendation would then be to call computeWorldMatrix() (I don't think we need to force it) inside RayHelper before setting ray.origin

regnaio: You're right. The critical line is `ray.origin = bullet.getAbsolutePosition()`. If raycast checks were inside `registerBeforeRender` rather than `onAfterStepObservable`, we would not need to update the ray's origin and can omit this line, since there is `this._attachedToMesh.getScene().registerBeforeRender(this._updateToMeshFunction)`.

`this._attachedToMesh.getScene().onAfterStepObservable.add(this._updateToMeshFunction)` allows the RayHelper to accurately follow the mesh inside the physics loop _without_ the line `ray.origin = bullet.getAbsolutePosition()`.

If this change is not accepted, `ray.origin = bullet.getAbsolutePosition()` is not a difficult workaround. However, it becomes more difficult when the bullet also rotates in the physics loop. Not only would we have to manually update the ray's origin, but we also have to manually calculate its direction in the physics loop. This is automatically handled for us in `registerBeforeRender` but not in `onAfterStepObservable`.

`RayHelper` is amazing in that it automatically handles origin and direction updates for `Rays` attached to `Meshes`. In the current _preview.babylonjs.com/babylon.js_, we cannot leverage the power of this feature in the physics loop. Without this change, we have to calculate the origin and direction ourselves in every physics step. Thus, there would be little reason to use a `RayHelper` instead of only a `Ray` in a physics loop.

deltakosh: Sorry I was unclear, my ask was to replace this PR with a single change by adding `this._attachedToMesh.computeWorldMatrix()` there: https://github.com/BabylonJS/Babylon.js/blob/master/src/Debug/rayHelper.ts#L210

This way the ray will always be up to date

regnaio: Ah, my bad for misunderstanding.

I just made the changes you recommended:
- Deleted `this._attachedToMesh.getScene().onAfterStepObservable.add(this._updateToMeshFunction);`
- Added `this._attachedToMesh.computeWorldMatrix();` before `this._attachedToMesh.getDirectionToRef(this._meshSpaceDirection, ray.direction);`

These changes did not resolve the issue in the Playground example, as bullets are still sometimes not registering one hit per second or registering multiple hits per second.

These changes did not achieve the desired result as `this._attachedToMesh.getScene().onAfterStepObservable.add(this._updateToMeshFunction);` did. Perhaps we need to add something?

Thank you for your help!

deltakosh: Does it work if you call computeWorldMatrix(true)?

deltakosh: Wait no I understand now :)

I did not pay enough attention to your code. Your fix is the right one :)

Merging now

PR #8707, opened 2020-08-07
Check Event Type Touch input Camera

CedricGuillemet:
This is a fix for https://forum.babylonjs.com/t/move-forward-in-universal-camera-not-working-in-chrome/13121/6
On Safari with a Universal Camera, mouse event are treated like touch input. So when you click and drag, instead of rotating the camera, il also translates the camera.
I fixed it by checking the instance type of the event. It worked for me but this is typically the kind of fix that breaks multiple things down the line.


deltakosh:
Should be fine

_Deltaksh famous last words_

PR #8713, opened 2020-08-08
Fix the bounding box of instances

Popov72:
See https://forum.babylonjs.com/t/shared-skeleton-between-glb-import-instances-incorrectly-culls-instances/13209/6


shasaur:
Amazing work! Will this work beyond a skeleton's initial pose (regarding what @deltakosh mentioned)?

shasaur:
Okay that's fine, so meshes with animations will still be fine as long as the animation doesn't cause any of the vertices to move outside the original bounding box, in which case I suppose you might get some edge case where a mesh might disappear too soon.

deltakosh:
Yes and you will have either the option to flag the mesh as always active or call mesh.refreshBoundingInfo(true);

Popov72:
Another option is to simply set an initial bounding box a little bigger / bigger enough to encompass all frames of the animation.

PR #8721, opened 2020-08-11
Remove Nullable<number> from DeviceSourceManager contracts

Drigax:
This PR follows up on my observations in: 
https://forum.babylonjs.com/t/devicesource-initializes-inputs-to-null/12961/2

In short, for event-driven input devices such as a keyboard, we initialize the inputs of this device as `null` This behavior doesnt seem to make sense, if we consider an input as either expressed or not, `null` would suggest an invalid value.

This PR proposes to remove null from our contract, requesting an input value should now always return a `number` value

### DIFF
--- a/src/DeviceInput/deviceInputSystem.ts
+++ b/src/DeviceInput/deviceInputSystem.ts
@@ -29,7 +29,7 @@ export class DeviceInputSystem implements IDisposable {
     public onInputChanged: (deviceType: DeviceType, deviceSlot: number, inputIndex: number, previousState: Nullable<number>, currentState: Nullable<number>) => void;
 
     // Private Members
-    private _inputs: Array<Array<Array<Nullable<number>>>> = [];
+    private _inputs: Array<Array<Array<number>>> = [];
     private _gamepads: Array<DeviceType>;
     private _keyboardActive: boolean = false;
     private _pointerActive: boolean = false;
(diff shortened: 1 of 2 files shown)


src/DeviceInput/deviceInputSystem.ts
```diff
@@ -138,10 +138,10 @@ export class DeviceInputSystem implements IDisposable {
         }
 
         if (!this._inputs[deviceType][deviceSlot]) {
-            const device = new Array<Nullable<number>>(numberOfInputs);
+            const device = new Array<number>(numberOfInputs);
 
             for (let i = 0; i < numberOfInputs; i++) {
-                device[i] = null;
+                device[i] = 0; /* set device input as unpressed */
```
PolygonalSun, line 144: So the original reason why we kept this as null was that there wasn't a single initialization value that worked for every device.  While gamepads, keyboards, and mouse buttons makes sense to have a init value of 0, mouse/touch x and y don't always start at 0.  This would mean that there would be a brief window where the pointer's location would be incorrect but still be a valid value (as opposed to null, which could be checked and ignored).  If we go this route, it might also be a good idea to immediately update the values without triggering an onInputChanged event.

PolygonalSun: This could be easily done by adding
```
pointer[0] = evt.clientX;
pointer[1] = evt.clientY;
```
to just after the _registerDevice calls for both _pointerMoveEvent and_pointerDownEvent.

Drigax: I can dig it. I'll add this into this PR as well.

bghgary:
Babylon Native input implementation needs to be updated once this is merged. https://github.com/BabylonJS/BabylonNative/blob/master/Plugins/NativeInput/Source/DeviceInputSystem.cpp

cc @ryantrem @Alex-MSFT

PR #8728, opened 2020-08-11
Deactive sliders when canvas loses focus

### LINKED ISSUE #8719: GUI Sliders become permanently active if window loses focus while sliding
**Repro**
- Bug repro on [playground](https://www.babylonjs-playground.com/#U9AC0N#1)
- Expected result: If you begin dragging the slider, and the canvas loses focus (for example, if the user changes tabs), the slider should also lose focus and stop sliding.
- Current result: The slider becomes permanently "active", so it will continue to slide with the mouse even after the mouse button has been released. There doesn't seem to be a way to get it to stop sliding.

Note: this behavior also happens if you trigger a context menu by right clicking, although this is disabled in the playground.

 - OS: Windows 10
 - Browser: Chrome
 - Version: 83.0.4103.116

DarraghBurkeMS:
Adds an _onCanvasBlur() function that GUI controls can override, which handles canvas blurring events. For all controls, it sets downCount to 0, since no control should think the pointer is still down if the canvas has lost focus. Fixes #8719 by deactivating a slider when the canvas is blurred.

### DIFF
--- a/gui/src/2D/controls/control.ts
+++ b/gui/src/2D/controls/control.ts
@@ -1854,14 +1854,7 @@ export class Control {
     }
 
     /** @hidden */
-    public _onCanvasBlur(): void {
-        if (!this._isEnabled) {
-            return;
-        }
-        this._downCount = 0;
-
-        this._downPointerIds = {};
-    }
+    public _onCanvasBlur(): void {}
 
     /** @hidden */
     public _processObservables(type: number, x: number, y: number, pointerId: number, buttonIndex: number, deltaX?: number, deltaY?: number): boolean {
--- a/gui/src/2D/controls/control.ts
+++ b/gui/src/2D/controls/control.ts
@@ -1854,6 +1854,16 @@ export class Control {
     }
 
     /** @hidden */
+    public _onCanvasBlur(): void {
+        if (!this._isEnabled) {
+            return;
+        }
+        this._downCount = 0;
+
+        this._downPointerIds = {};
+    }
+
+    /** @hidden */
     public _processObservables(type: number, x: number, y: number, pointerId: number, buttonIndex: number, deltaX?: number, deltaY?: number): boolean {
         if (!this._isEnabled) {
             return false;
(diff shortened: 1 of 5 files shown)


gui/src/2D/controls/control.ts
```diff
@@ -1853,6 +1853,16 @@ export class Control {
         if (canNotify && this.parent != null) { this.parent._onWheelScroll(deltaX, deltaY); }
     }
 
+    /** @hidden */
+    public _onCanvasBlur(): void {
```
deltakosh, line 1857: What happens if I click a button, keep the mouse button clicked, move outside of the window and get back to the window?

deltakosh: https://www.babylonjs-playground.com/#HH8Q22

Try with the color picker for instance. Currently I can pick a color, maintain the mouse button and go outside of the window

DarraghBurkeMS: Fair point, with this solution we would have to implement _onCanvasBlur for every control which uses pointerUp to handle going outside the window as well.

The easy solution would be to just trigger onPointerUp for those controls, but the problem with that is _onPointerUp expects coordinates, which a canvasBlur event wouldn't have.

Not sure what the best approach is...I can definitely implement _onCanvasBlur for all the controls which need it if we think that's the right solution.

deltakosh: I'm more on the fence to do nothing for all controls and start by having it  do something only for slider for now

DarraghBurkeMS: I would think maybe we should fix it for the color picker as well since it is possible to get the color picker "stuck" by releasing the mouse outside the window, just like the slider. I don't think we need to fix it for the button neccessarily.

deltakosh: Did you manage to get it stuck?

DarraghBurkeMS: The color picker currently gets stuck if I click on it, switch to a new tab, release the mouse button, and then come back. We could fix this by just calling _forcePointerUp I think.

deltakosh: You are right! Let's do it :)

DarraghBurkeMS: Done! Let me know what you think

PR #8745, opened 2020-08-17
Fix changing the loop value for a sound already started did not work

Popov72:
See https://forum.babylonjs.com/t/cant-use-updateoption-inside-an-actionmanger/13395


Popov72:
Closing as already and better fixed by https://github.com/BabylonJS/Babylon.js/pull/8746

Issue #8708, opened 2020-08-07
Merge PrePass and GeometryBufferRenderer


deltakosh:
- GeometryBuffer should be an empty shell that uses PrePass internally for back compat
- PrePass will then check as part of its pull process what the geometryBuffer can need
- Prepass is pure: SubSurface and everything not related to the pure rendering to textures must be moved as independent scene components
- We need to extract all postprocesses from PrePass and make them independent (and used by the new Scene Components)
- Composition is done through postprocesses when required like for subSurface composition (as part of the final image processing)
- Std and PBR materials will expose a new property: isPrePassCapable. This will let the PrePass known that they can produce expected outcome. We need to add support for velocity and reflectivity.
- If a material is NOT isPrepassCapable then we use the current geometry shader to get some of the expected outputs (subsurface will not work for instance)
- PrePass must work with webgl1 + draw buffer extension

PR #8747, opened 2020-08-17
Prepass adjustments

CraigFeldspar:
Related to #8708

+ Clears prepass of any reference to subsurface scattering
+ Adding composition of textures in the prepass MRT (prepass automatically determines what the mrt needs to render, nad manages texture indices)


src/Rendering/prePassRenderer.ts
```diff
@@ -21,25 +23,89 @@ export class PrePassRenderer {
         throw _DevTools.WarnImport("PrePassRendererSceneComponent");
     }
 
+    /**
+     * Constant used to retrieve the irradiance texture index in the textures array
+     * using getIndex(PrePassRenderer.IRRADIANCE_TEXTURE_TYPE)
+     */
+    public static readonly IRRADIANCE_TEXTURE_TYPE = 0;
```
deltakosh, line 30: good idea! So we can add more like the velocity

deltakosh:
still failing :(

CraigFeldspar:
there is a bug with multisampling

PR #8752, opened 2020-08-18
Texture Inspector: Selection, Channel Locking, Mip Previews

DarraghBurkeMS:
- Adds the ability to select a rectangular area and only paint to that area
- Allows you to lock individual channels, meaning those channels won't be affected by any tools that you use
- Allows you to view different mips levels, from 0 to 12
- Changes the way alpha works. Rather than representing a blend factor, you will now paint the exact RGBA value that you specify onto the canvas. This gives the artist more control over what they are painting
- Optimized painting by staggering texture updates and using drawImage() rather than other, slower canvas methods
- Adds a checkerboard background to the preview pane, making alpha in textures more readable and giving a more consistent UI across the board


deltakosh:
Checks are not going through:)

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/defaultTools/paintbrush.ts
```diff
@@ -85,13 +135,19 @@ export const Paintbrush : IToolData = {
         cleanup () {
             Object.entries(this.GUI).forEach(([key, value]) => value.dispose());
             this.isPainting = false;
-            if (this.pointerObservable) {
-                this.getParameters().scene.onPointerObservable.remove(this.pointerObservable);
+            if (this.pointerObserver) {
+                this.getParameters().scene.onPointerObservable.remove(this.pointerObserver);
             }
         }
+        hexToRgb(hex : string) {
```
deltakosh, line 142: we should use Color3.FromHexString

DarraghBurkeMS: Just what I was looking for :) Thanks!

Issue #8754, opened 2020-08-18
[XR] Allow styling the overlay to which XR buttons are added


RaananW:
See for example https://forum.babylonjs.com/t/webxr-change-default-enter-exit-icons/13427/3

PR #8755, opened 2020-08-18
[XR] Expose HTML overlay and allow styling

RaananW:
Closing #8754 

It is now possible to configure the overlay of the XR buttons. The default styling is:

```
"z-index:11;
position: absolute;
right: 20px;bottom: 50px;"
```

The button has the class `xr-button-overlay` added to it, and is exposed if the dev wants to configure using javascript


deltakosh:
Tiny conflict due to previous merge

PR #8802, opened 2020-08-25
Fix for .babylon exported gltf meshes having improper deformation

### LINKED ISSUE #7244: Sandbox export to .babylon file breaks animations
Forum post with context and example:

https://forum.babylonjs.com/t/about-glb-convert-to-babylon/7324

Drigax:
This PR addresses an issue in the babylon scene serialization and deserialization logic for gltf imported meshes. On scene round trip via .babylon format, meshes appear small, improperly oriented and no longer animated by node animation
Issue is caused by helper members `Skeleton.overrideMesh` and `Bone.linkedTransformNode` not being serialized and modified bone serialization indexing logic to match schema used by glTF loader, where glTF coordinate space transform nodes and their associated bones are ignored by the skeleton hierarchy when resolving world transform.

Fixes #7244

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -3450,12 +3450,12 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         }
 
         // Morph targets
-        if (parsedMesh.morphTargetManagerId) {
+        if (parsedMesh.morphTargetManagerId > -1) {
             mesh.morphTargetManager = scene.getMorphTargetManagerById(parsedMesh.morphTargetManagerId);
         }
 
         // Skeleton
-        if (parsedMesh.skeletonId !== undefined ) {
+        if (parsedMesh.skeletonId !== undefined && parsedMesh.skeletonId !== null) {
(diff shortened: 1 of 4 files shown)


src/Meshes/mesh.ts
```diff
@@ -3449,12 +3450,12 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         }
 
         // Morph targets
-        if (parsedMesh.morphTargetManagerId > -1) {
```
deltakosh, line 3452: why this one? what if id is === 0?

Drigax: ah, thanks for catching that, I missed copying over my change @L3458

Drigax: I'll also add a `null` check as well

Drigax: I didn't realize that `MorphTargetManagerId` can only be int :), reverting that change

tskarthikus:
I am using a glb with animation and trying to export it as such without any alteration.

image

But, the export is having deformations on bones.
It is reprocible in Babylon Sandbox.

image

PR #8816, opened 2020-08-26
Fix invalid bone index when parsing mesh data

Popov72:
See https://forum.babylonjs.com/t/always-get-a-deformation-spike-from-3ds-max/13535/17


deltakosh:
OMG this is a major flow we had since forever! So glad you found it! It must be there since v1 :D

Issue #8797, opened 2020-08-25
NME - Wire sometimes disconnecting from trailing node to connect with 0,0 when collapsing frames


PatrickRyanMS:
**Repro**
- Launch nme and create a small node tree in a frame. This happens with one out as shown below or with multiple wires leaving the frame. 
- Collapse frame. 
- Expected result: The wires still connect between the ports on the edge of the frame and the surrounding nodes in the same wiring path.
- Current result: Note that sometimes you get into a state where a wire will disconnect from one port and connect with 0,0 on the canvas. When you get into this state, opening and closing the frame will yield a 100% repro.

**Screenshots**

![image](https://user-images.githubusercontent.com/31326855/91109206-23669c00-e62f-11ea-8110-d213c4174879.png)

PR #8818, opened 2020-08-26
NME: Bug Fix for Incorrect Drawn Node Links

msDestiny14:
Delay the disposing port until it is confirmed another one will not be added.

https://github.com/BabylonJS/Babylon.js/issues/8797


Popov72 (approved):
LGTM ;)

Issue #8514, opened 2020-07-09
GLTF Exporter incorrect rendering


luruke:
Hello,
I'm experiencing issues with the glTF Exporter (https://doc.babylonjs.com/extensions/gltfexporter).

My current project is running 3.3.0,
here's how the scene looks like vs the .glb file exported (viewed via babylon sandbox)

![Screenshot 2020-07-09 at 12 27 35](https://user-images.githubusercontent.com/61326/87028947-aa76c680-c1df-11ea-8138-e2ce9253261c.png)

The surrounding box that is `isVisible = false` in the code doesn't seems to match the gltf
![Screenshot 2020-07-09 at 12 29 06](https://user-images.githubusercontent.com/61326/87029037-d5f9b100-c1df-11ea-90b8-1e16cc9063f2.png)

Even after i hide the box the model doesn't looks correct. I assume UV/Texture match are wrong?
![Screenshot 2020-07-09 at 12 29 15](https://user-images.githubusercontent.com/61326/87029140-ffb2d800-c1df-11ea-895e-48863b0ef103.png)

To run a test I upgraded my whole project to 4.1.0. The glb file exported is smaller compared to 3.3.0 (21mb vs 36mb) and apparently it's missing a texture, 
![Screenshot 2020-07-09 at 12 31 38](https://user-images.githubusercontent.com/61326/87029255-2b35c280-c1e0-11ea-8e76-3f2cebb2d751.png)

If I ignore this and open on blender, i can still see that uv/textures are wrong:
![Screenshot 2020-07-09 at 12 35 55](https://user-images.githubusercontent.com/61326/87029695-c75fc980-c1e0-11ea-8fdf-e524da4f3b4d.png)

Unfortunately the project is a bit complex to provide a small contained jsfiddle to replicate this, but you can see the project live at https://www.mobitec.be/fr/configurateur/chaise/pure-c08?department=&upholstery=simili&data=%7B%221%22%3A694,%222%22%3A48%7D

Is this is a known bug, any advice/pitfall or something to avoid when using glTF exporter?

Drigax:
Thanks for the report,

Currently visibility is not supported in glTF, we currently export all nodes in scene regardless of whether they are enabled/disabled. Adding @bghgary, do you think this is correct behavior, or should we rather omit exporting disabled nodes?

[...]

PR #8824, opened 2020-08-27
Fix glb/gltf export: texture not tiled / invalid format for pictures

Popov72:
See https://github.com/BabylonJS/Babylon.js/issues/8514


serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -515,7 +515,15 @@ export class _GLTFMaterialExporter {
...
-                const canvas = engine.getRenderingCanvas();
+                const canvas0 = engine.getRenderingCanvas();
+
+                let canvas: Nullable<HTMLCanvasElement> = document.createElement("canvas");
+
+                canvas.width = canvas0?.width ?? 0;
+                canvas.height = canvas0?.height ?? 0;
+
+                var destCtx = canvas.getContext('2d');
+                destCtx!.drawImage(canvas0!, 0, 0);
```
Popov72, line 526: I think the block of code above that copies the canvas is needed, else it's possible we render a picture whereas the previous one is still not processed by the code below (which converts the canvas to a blob then to a base64 string), overwritting the previous data in the canvas.

Or are we sure all the promises involved in the process will be completed right away, so in effect all the conversion code is synchronous (in which case no need to copy the canvas)? In effect, it's only `fileReader.readAsDataURL` that will trigger `fileReader.onLoad` that is/can be asynchronous. If we are sure it won't be done asynchronously, I can remove my code.

I tried to have a look at the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL) and the [spec](https://w3c.github.io/FileAPI/#readAsDataURL), but I can't see anything saying that `readAsDataURL` would be synchronous...

Popov72:
A bit of explanation regarding the "invalid format" for pictures:

The problem was that the JSON `images` node is filled by looping through `this.images` (line 953 in `glTFExporter.ts`) whereas the binary data of the pictures were written to the glb file in the order of the key of `this._imageData` because the loop was `for (key in this._imageData)`.

I think that by default (at least in Chrome - I don't think it is a requirement of the spec) the keys are looped in the order they were created (else I think we would have seen the problem earlier). So, if insertions are made in `_images` in the same order than in `_imageData` then everything is ok. However, because of asynchronous operations, it's possible we perform the insertions out of order, meaning `_images[0]` does not correspond to the first object inserted in `_imageData`.

This one made me sweat quite a bit, as the bug did not appear all the time because of the asynchronous dimension and because for a long time I was looking at the wrong place (as some pictures were invalid, I was looking for problems in the texture management/conversion code...).

serializers/src/glTF/2.0/glTFMaterialExporter.ts
deltakosh: I think this is fine to be secure

deltakosh:
Good job..this kind of race condition are terrible to fix
Thanks!

Issue #8825, opened 2020-08-27
Inspector: Launch tool from Icon


msDestiny14:
Be able to launch tools (nme) from an icon next to the object name like shown below. 

![GetImage](https://user-images.githubusercontent.com/67929198/91475295-3c9a6300-e850-11ea-8aab-6f294edc51d7.png)

PR #8827, opened 2020-08-27
Inspector: Launch tool from Icon

msDestiny14:
Capture

https://github.com/BabylonJS/Babylon.js/issues/8825

Also fixed a warning I noticed coming from NME.

### DIFF
--- a/inspector/src/components/sceneExplorer/entities/materialTreeItemComponent.tsx
+++ b/inspector/src/components/sceneExplorer/entities/materialTreeItemComponent.tsx
@@ -1,13 +1,14 @@
 import { IExplorerExtensibilityGroup } from "babylonjs/Debug/debugLayer";
 import { Material } from "babylonjs/Materials/material";
-
-import { faBrush } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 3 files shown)


inspector/src/components/sceneExplorer/entities/materialTreeItemComponent.tsx
```diff
@@ -18,6 +19,21 @@ export class MaterialTreeItemComponent extends React.Component<IMaterialTreeItem
...
+
+        if(this.props.material.getClassName() === "NodeMaterial") { 
+        
+            return (
+                <div className="materialTools">
+                    <TreeItemLabelComponent label={this.props.material.name} onClick={() => this.props.onClick()} icon={faBrush} color="orange" />
+                    {
+                        <ExtensionsComponent target={this.props.material} extensibilityGroups={this.props.extensibilityGroups} />
+                    }
+                    <div className="icon" onClick={() => {(this.props.material as NodeMaterial).edit()}} title="Node Material Editor" color="white">
```
deltakosh, line 31: can you change it so we do not need to set the main part twice? (we should just add the icon if it is a nodematerial

PR #8828, opened 2020-08-28
Allow changing color of thumb in sliders

Patbox:
Small change to sliders, that allows changing color of thumbs.
image


deltakosh:
This is great thanks!

PR #8833, opened 2020-08-28
Remove Nullable<Number> from DeviceSourceManager.getInput, add observable cleanup to dispose()

Drigax:
This PR finalizes changes to the deviceSourceManager contract by via #8721 by removing Nullable<number> from `DeviceSourceManager.getInput`, this also improves the cleanup of DeviceSourceManager by properly cleaning up observables on `dispose()`"

Also, a change was made to `DeviceSourceManager.getDeviceSources()` that filters out null device entries.


Drigax:
@deltakosh  I'm adding one more fix to the `DeviceSourceManager.getDeviceSources`  entrypoint, that only returns non-null devices, courtesy of @ryantrem. This should allow us to avoid situations where we have a bunch of null devices returned in an array due to touch events causing connection/disconnection of new devices

PR #8844, opened 2020-09-01
[gizmo] fix camera LH/RH coord + fix instanced meshes update

### LINKED ISSUE #8784: Using Gizmos With Instanced Mesh
I'm sure it's a bug, so I'll not use the forum first :)

Using gizmos with InstancedMesh class has some problems: the world matrix of the instance is well updated according to the gizmo's manipulation but its transformation properties (position, rotation / rotation quaternion and scaling) are not updated.

Using the given playground: move using plane gizmos. Then, use one of the axis gizmos. You'll see that the sphere has its transformation reset to the original one.

According to this line, the "_matrixChanged" method doesn't take care of InstancedMesh class: https://github.com/BabylonJS/Babylon.js/blob/master/src/Gizmos/gizmo.ts#L197

- Bug repro on [playground](https://www.babylonjs-playground.com/#31M2AP#119): https://www.babylonjs-playground.com/#31M2AP#119
- Expected result: The instanced mesh has its transformation properties updated.
- Current result: The instanced mesh has its position reset before it is moved again and transformation properties are not updated.

CedricGuillemet:
fixes #8784 and this issue : https://forum.babylonjs.com/t/attaching-a-gizmo-to-a-camera-with-userighthandedsystem/13532
For instanced meshes, simple check in the update matrix code.
for LH/RH:
camera worldMatrix is computed as the inverse of the viewMatrix. So, when doing a loop with gizmos (worldmatrix editing-> viewMatrix computation -> worldMatrix as the inverse), a rotation on the Y axis happens each frame giving a nice flipping effect.
Now, when decomposing translation/scaling/rotation, I compute rotation on the Y axis as if the continuity of the worldmatrix is preserved.
Also fixed the camera gizmo display mesh that didn't take RightHand coordinates into account.

### DIFF
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -115,7 +115,7 @@ export class Gizmo implements IDisposable {
     private _tempMatrix1 = new Matrix();
     private _tempMatrix2 = new Matrix();
     private _rightHandtoLeftHandMatrix = Matrix.RotationY(Math.PI);
-    
+
     /**
      * Creates a gizmo
      * @param gizmoLayer The utility layer the gizmo will be added to
@@ -195,7 +195,7 @@ export class Gizmo implements IDisposable {
             if (camera.parent) {
                 var parentInv = this._tempMatrix2;
                 camera.parent._worldMatrix.invertToRef(parentInv);
-                this._attachedNode.getWorldMatrix().multiplyToRef(parentInv, this._tempMatrix1);
+                this._attachedNode._worldMatrix.multiplyToRef(parentInv, this._tempMatrix1);
                 worldMatrix = this._tempMatrix1;
             } else {
                 worldMatrix = this._attachedNode._worldMatrix;
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -227,7 +227,7 @@ export class Gizmo implements IDisposable {
             }
 
             camera.position.copyFrom(this._tempVector);
-        } else if ((<Mesh>this._attachedNode)._isMesh || this._attachedNode.getClassName() === "AbstractMesh" || this._attachedNode.getClassName() === "TransformNode"|| this._attachedNode.getClassName() === "InstancedMesh") {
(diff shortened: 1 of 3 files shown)


RaananW (approved):
LGTM! just added a small suggestion

src/Gizmos/gizmo.ts
```diff
@@ -187,16 +189,27 @@ export class Gizmo implements IDisposable {
...
+                this._attachedNode.getWorldMatrix().multiplyToRef(parentInv, this._tempMatrix1);
+                worldMatrix = this._tempMatrix1;
             } else {
-                this._attachedNode.getWorldMatrix().decompose(this._tempVector2, this._tempQuaternion, this._tempVector);
+                worldMatrix = this._attachedNode._worldMatrix;
             }
 
+            if (camera.getScene().useRightHandedSystem) {
+                // avoid desync with RH matrix computation. Otherwise, rotation of PI around Y axis happens each frame resulting in axis flipped because worldMatrix is computed as inverse of viewMatrix.
+                Matrix.RotationY(Math.PI).multiplyToRef(worldMatrix, this._tempMatrix2);
```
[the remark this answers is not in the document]
CedricGuillemet: That's a great idea! Thanks!

src/Gizmos/gizmo.ts
```diff
@@ -187,16 +189,27 @@ export class Gizmo implements IDisposable {
...
+            var worldMatrixUC;
             if (camera.parent) {
-                var parentInv = new Matrix();
-                var localMat = new Matrix();
-                camera.parent.getWorldMatrix().invertToRef(parentInv);
-                this._attachedNode.getWorldMatrix().multiplyToRef(parentInv, localMat);
-                localMat.decompose(this._tempVector2, this._tempQuaternion, this._tempVector);
+                var parentInv = this._tempMatrix2;
+                camera.parent._worldMatrix.invertToRef(parentInv);
+                this._attachedNode.getWorldMatrix().multiplyToRef(parentInv, this._tempMatrix1);
```
Popov72, line 197: Don't you want to also use `this._attachedNode._worldMatrix` here?

CedricGuillemet: Good catch!

PR #8847, opened 2020-09-01
Fix issue with joint index export of glTF Skins

Drigax:
Context is available here: https://forum.babylonjs.com/t/gltf-export-bug-skinning-has-errors/13744/13

In short, we had an issue with glTF skin export, where we improperly assumed that the Babylon `Skeleton.bones` order was identical to the bone indices used by our shader. This change properly fixes the `glTFSkin` export to now use the `Bone._index` to order the joint indices exported in the skin.

### DIFF
--- a/serializers/src/glTF/2.0/glTFExporter.ts
+++ b/serializers/src/glTF/2.0/glTFExporter.ts
@@ -13,6 +13,7 @@ import { Mesh } from "babylonjs/Meshes/mesh";
 import { MorphTarget } from "babylonjs/Morph/morphTarget";
 import { LinesMesh } from "babylonjs/Meshes/linesMesh";
 import { InstancedMesh } from "babylonjs/Meshes/instancedMesh";
+import { Bone } from "babylonjs/Bones/bone";
 import { BaseTexture } from "babylonjs/Materials/Textures/baseTexture";
 import { Texture } from "babylonjs/Materials/Textures/texture";
 import { Material } from "babylonjs/Materials/material";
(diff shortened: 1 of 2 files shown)


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1889,17 +1890,26 @@ export class _Exporter {
             let inverseBindMatrices: Matrix[] = [];
             let skeletonMesh = babylonScene.meshes.find((mesh) => { mesh.skeleton === skeleton; });
             skin.skeleton = skeleton.overrideMesh === null ? (skeletonMesh ? nodeMap[skeletonMesh.uniqueId] : undefined) : nodeMap[skeleton.overrideMesh.uniqueId];
+            let boneIndexMap: Map<number, Bone> = new Map();
```
bghgary, line 1893: Careful with using Map as they are not fully supported in legacy browsers. Might consider using a normal object for this.

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1889,17 +1890,26 @@ export class _Exporter {
...
-                        skin.joints.push(nodeMap[transformNode.uniqueId]);
+                boneIndex = bone.getIndex();
+                if (boneIndex > -1) {
+                    boneIndexMap.set(boneIndex, bone);
+                }
+                boneIndexMax = Math.max(boneIndexMax, boneIndex);
+            }
+            for (let i = 0; i <= boneIndexMax; ++i) {
+                let bone = boneIndexMap.get(i)!;
+                let transformNode = bone.getTransformNode();
```
bghgary, line 1905: nit: generally, use `const` if possible instead of `let`

serializers/src/glTF/2.0/glTFExporter.ts
Drigax: thanks, replaced with a generic object instead

serializers/src/glTF/2.0/glTFExporter.ts
Drigax: Thanks, replaced let with const for `createSkinsAsync`

PR #8852, opened 2020-09-02
Dispose transmission helper when scene disposes

MiiBond:


loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -62,6 +62,9 @@ class TransmissionHelper {
         };
         this._scene = scene;
         this.onErrorObservable = new Observable();
+        this._scene.onDisposeObservable.addOnce((scene) => {
+            this.dispose();
```
RaananW, line 66: Is this class being used outside of the extension itself (`KHR_materials_transmission`)?

It can be disposed when the extension is being disposed, this way if for some reason the loader and its extension is being disposed before the scene is the `TransmissionHelper ` will still be correctly disposed.

sebavan: nope not used outside so I guess it is fine as is ?

RaananW: Wouldn't it be better to dispose this in the dispose function of the extension?

sebavan: @bghgary I guess the extension are disposed right after loading per loading session isnt it ?

sebavan: if not and this is per scene, totally agree @RaananW

bghgary: glTF loader extensions are disposed immediately after loading is complete.

sebavan: so actually we need to store _transmissionHelper on the scene and dispose it on scene dispose to prevent conflicts in multiscene. so the dispose is ok and @MiiBond could we move the _transmissionHelper on the scene to avoid conflicts in multiscene scenarios ?

sebavan: I am gonna merge this one and do the update locally in a couple ours, will be fixed in the next nightly later on today.

sebavan:
Closed by  #8864

PR #8853, opened 2020-09-02
clickable light/camera gizmos

### LINKED ISSUE #8737: Picking on CameraGizmo and LightGizmo
Hello,

It would be interesting to be able to click on a gizmoLight and cameraGizmo to display the GizmoManager once clicked.

**Why ?**
This is for a publisher. The user should be able to select the GizmoCamera and GizmoLight on the stage to display the GizmoManager and move the camera or light.

It would also be useful to retrieve the light attached to the GizmoLight and retrieve the camera attached to the GizmoCamera. So that the user can modify their properties. A bit like when we select a mesh on the scene, we know that it is a mesh selected, we can retrieve the properties (position, name ...) and we can display the GizmoManager when this mesh is selected.

Would be useful if we could select (picked) the LightGizmo and CameraGizmo

Discution forum : 
https://forum.babylonjs.com/t/attachcontrol-and-cameragizmo-with-frustum/13307/5?u=dad72

I must have tried something on this PG, but it's not good when you have multiple cameraGizmo on stage. I have not been able to find or choose the correct camera, but still the last camera added to the scene. But the idea is there.
https://www.babylonjs-playground.com/#Z4ZGAY#24

Thank you in advance.

CedricGuillemet:
fixes #8737 
Clickable camera and light gizmos.

Example of use:
```
var gizmoManager = new BABYLON.GizmoManager(scene, 3);
gizmoManager.positionGizmoEnabled = true;

var lightGizmo = new BABYLON.LightGizmo();
lightGizmo.light = light;

lightGizmo.onClickedObservable.add((_light) => {
            manager.attachToNode(_light);              
});
```

And in a PG : https://playground.babylonjs.com/#M48IGY

### DIFF
--- a/src/Gizmos/cameraGizmo.ts
+++ b/src/Gizmos/cameraGizmo.ts
@@ -39,18 +39,16 @@ export class CameraGizmo extends Gizmo {
         this._material.diffuseColor = new Color3(0.5, 0.5, 0.5);
         this._material.specularColor = new Color3(0.1, 0.1, 0.1);
 
-        if (gizmoLayer) {
-            this._pointerObserver = gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
-                if (!this._camera) {
-                    return;
-                }
+        this._pointerObserver = gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
+            if (!this._camera) {
+                return;
+            }
 
-                var isHovered = pointerInfo.pickInfo && (this._rootMesh.getChildMeshes().indexOf(<Mesh>pointerInfo.pickInfo.pickedMesh) != -1);
-                if (isHovered && pointerInfo.type === PointerEventTypes.POINTERDOWN && pointerInfo.event.button === 0) {
-                    this.onClickedObservable.notifyObservers(this._camera);
-                }
-            });
-        }
+            var isHovered = pointerInfo.pickInfo && (this._rootMesh.getChildMeshes().indexOf(<Mesh>pointerInfo.pickInfo.pickedMesh) != -1);
+            if (isHovered && pointerInfo.event.button === 0) {
+                this.onClickedObservable.notifyObservers(this._camera);
+            }
+        }, PointerEventTypes.POINTERDOWN);
     }
     private _camera: Nullable<Camera> = null;
 
@@ -141,6 +139,7 @@ export class CameraGizmo extends Gizmo {
      * Disposes of the camera gizmo
      */
     public dispose() {
+        this.onClickedObservable.clear();
         this.gizmoLayer.utilityLayerScene.onPointerObservable.remove(this._pointerObserver);
         if (this._cameraMesh) {
             this._cameraMesh.dispose();
--- a/src/Gizmos/cameraGizmo.ts
+++ b/src/Gizmos/cameraGizmo.ts
@@ -11,6 +11,8 @@ import { BoxBuilder } from "../Meshes/Builders/boxBuilder";
 import { CylinderBuilder } from '../Meshes/Builders/cylinderBuilder';
 import { Matrix } from '../Maths/math';
 import { LinesBuilder } from "../Meshes/Builders/linesBuilder";
+import { PointerEventTypes, PointerInfo } from '../Events/pointerEvents';
+import { Observer, Observable } from "../Misc/observable";
 
 /**
  * Gizmo that enables viewing a camera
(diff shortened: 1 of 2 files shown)


src/Gizmos/lightGizmo.ts
```diff
@@ -39,6 +47,19 @@ export class LightGizmo extends Gizmo {
         this._material = new StandardMaterial("light", this.gizmoLayer.utilityLayerScene);
         this._material.diffuseColor = new Color3(0.5, 0.5, 0.5);
         this._material.specularColor = new Color3(0.1, 0.1, 0.1);
+
+        if (gizmoLayer) {
```
RaananW, line 51: not needed if it is no longer optional

src/Gizmos/lightGizmo.ts
```diff
@@ -39,6 +47,19 @@ export class LightGizmo extends Gizmo {
...
         this._material.specularColor = new Color3(0.1, 0.1, 0.1);
+
+        if (gizmoLayer) {
+            this._pointerObserver = gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
+                if (!this._light) {
+                    return;
+                }
+
+                var isHovered = pointerInfo.pickInfo && (this._rootMesh.getChildMeshes().indexOf(<Mesh>pointerInfo.pickInfo.pickedMesh) != -1);
+                if (isHovered && pointerInfo.type === PointerEventTypes.POINTERDOWN && pointerInfo.event.button === 0) {
```
RaananW, line 58: If this should be executed only on pointer down you can prefilter using the second variable of the `add` function:

```javascript
gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
    /* guaranteed to be a pointer down event*/
}, PointerEventType.POINTERDOWN);
```

src/Gizmos/lightGizmo.ts
```diff
@@ -25,12 +27,18 @@ export class LightGizmo extends Gizmo {
     private _cachedPosition = new Vector3();
     private _cachedForward = new Vector3(0, 0, 1);
     private _attachedMeshParent: TransformNode;
+    private _pointerObserver: Nullable<Observer<PointerInfo>> = null;
+
+    /**
+     * Event that fires each time the gizmo is clicked
+     */
+    public onClickedObservable = new Observable<Light>();
```
RaananW, line 35: Don't forget to clear this on dispose()

src/Gizmos/cameraGizmo.ts
```diff
@@ -19,17 +21,36 @@ export class CameraGizmo extends Gizmo {
     private _cameraMesh: Mesh;
     private _cameraLinesMesh: Mesh;
     private _material: StandardMaterial;
+    private _pointerObserver: Nullable<Observer<PointerInfo>> = null;
+
+    /**
+     * Event that fires each time the gizmo is clicked
+     */
+    public onClickedObservable = new Observable<Camera>();
```
RaananW, line 29: clear on dispose

src/Gizmos/cameraGizmo.ts
```diff
@@ -19,17 +21,36 @@ export class CameraGizmo extends Gizmo {
...
-    constructor(gizmoLayer?: UtilityLayerRenderer) {
+    constructor(gizmoLayer: UtilityLayerRenderer = UtilityLayerRenderer.DefaultUtilityLayer) {
         super(gizmoLayer);
 
         this._material = new StandardMaterial("cameraGizmoMaterial", this.gizmoLayer.utilityLayerScene);
         this._material.diffuseColor = new Color3(0.5, 0.5, 0.5);
         this._material.specularColor = new Color3(0.1, 0.1, 0.1);
+
+        if (gizmoLayer) {
+            this._pointerObserver = gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
```
RaananW, line 43: same prefilter as in lightGizmo can be applied here

RaananW (changes requested):
the only real change is clearing the observable(s)  when disposing. the rest are just suggestions :-)

src/Gizmos/cameraGizmo.ts
CedricGuillemet: Fixed

src/Gizmos/cameraGizmo.ts
CedricGuillemet: Fixed

src/Gizmos/lightGizmo.ts
CedricGuillemet: Fixed

src/Gizmos/lightGizmo.ts
CedricGuillemet: Fixed

src/Gizmos/lightGizmo.ts
CedricGuillemet: Fixed

RaananW (approved):

PR #8860, opened 2020-09-02
Expose Node Gradient In the Inspector

msDestiny14:
Allowing the gradient node from NME to be visible in the inspector.

https://github.com/BabylonJS/Babylon.js/issues/8787


src/Materials/Node/nodeMaterialBlock.ts
```diff
@@ -706,6 +706,7 @@ export class NodeMaterialBlock {
         serializationObject.id = this.uniqueId;
         serializationObject.name = this.name;
         serializationObject.comments = this.comments;
+        serializationObject.visibleInInspector = this.visibleInInspector;
```
Popov72, line 709: You should also update the `InputBlock._dumpPropertiesCode` method to remove the handling of `visibleInInspector` and move it to `NodeMaterialBlock._dumpPropertiesCode` instead.

Popov72:
LGTM!

sebavan (approved):
LGTM as well, adding @RaananW as an extra reviewer

src/Materials/Node/Blocks/gradientBlock.ts
```diff
@@ -6,25 +6,57 @@ import { NodeMaterialBlockTargets } from '../Enums/nodeMaterialBlockTargets';
...
+     * Gets the color associated with this step
+     */
+    public get color(): Color3 {
+        return this._color;
+    }
+
+    /**
+     * Sets the color associated with this step
+     */
+    public set color(val: Color3) {
```
RaananW, line 45: React loves immutable objects. It keeps the state auto-updatable and make your state management a bit easier. If you use these objects in a react state (as it seems to me, but please correct me if I am wrong!!) don't add a setter. instead create a new object.

Of course - unless it is impossible due to the architecture (i saw that the parent is being notified)

sebavan:
@msDestiny14 I ll merge it as soon as the comments have been addressed :-)

src/Materials/Node/Blocks/gradientBlock.ts
msDestiny14: will get resolved in a second PR.

RaananW (approved):
Discussed offline, approved :-)

PR #8867, opened 2020-09-03
Texture Inspector Mipmap Improvements

DarraghBurkeMS:
Improves the way mipmaps are handled. Lower mip levels are now displayed using nearest sampling, so they appear sharp rather than muddy. You are no longer able to edit lower level mipmaps (this behavior was broken anyway); instead, attempting to edit will take you back to level 0. Mipmaps are correctly regenerated when the texture is edited.

Also adds custom cursors to some of the tools, making it more clear what action they are performing.


inspector/src/lod.ts
```diff
@@ -2,12 +2,17 @@ import { Effect } from "babylonjs/Materials/effect";
 
 let name = 'lodPixelShader';
 let shader = `
+#ifdef GL_ES
```
sebavan, line 5: this is not usefull in webgl GL_ES is always true :-)

inspector/src/lod.ts
```diff
@@ -2,12 +2,17 @@ import { Effect } from "babylonjs/Materials/effect";
...
+#endif
+
 varying vec2 vUV;
 uniform sampler2D textureSampler;
 uniform float lod;
+uniform vec2 texSize;
 void main(void)
 {
-gl_FragColor=textureLod(textureSampler,vUV,lod);
+    gl_FragColor = textureLod(textureSampler,vUV,lod);
```
sebavan, line 15: this requires an extension in Webgl1 with smthg like #extension GL_EXT_shader_texture_lod : enable

sebavan: and if texture lod are not supported, you could disable the mip map tool :-)

inspector/src/lodCube.ts
```diff
@@ -2,30 +2,34 @@ import { Effect } from "babylonjs/Materials/effect";
...
-#endif
-#ifdef POSITIVEZ
-gl_FragColor=textureCube(textureSampler,vec3(uv,1.001),lod);
-#endif
-#ifdef NEGATIVEZ
-gl_FragColor=textureCube(textureSampler,vec3(uv,-1.001),lod);
-#endif
+    vec2 uv=vUV*2.0-1.0;
+    #ifdef POSITIVEX
+    gl_FragColor=textureCube(textureSampler,vec3(1.001,uv.y,uv.x),lod);
```
sebavan, line 16: why the 0.001 ? I am wondering where it impacts.

sebavan (changes requested):
LGTM, just a couple comments.

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/textureCanvasManager.ts
```diff
@@ -104,9 +104,13 @@ export class TextureCanvasManager {
     private static RESET_KEY = 'KeyR';
     private static DESELECT_KEY = 'Escape'
 
+    /** The number of milliseconds between texture updates */
+    private static PUSH_FREQUENCY = 32;
```
Popov72, line 108: I think variable names in all upper case is for constants, meaning either `const` or `readonly` variables?

For static variables that can be modified it would be something like `PushFrequency`.

inspector/src/lod.ts
DarraghBurkeMS: Ah, do we need to update the example from this doc then? https://doc.babylonjs.com/how_to/how_to_use_postprocesses#custom-postprocesses

inspector/src/lodCube.ts
DarraghBurkeMS: I'm not sure, I based this code on the cube pass postprocess: https://github.com/BabylonJS/Babylon.js/blob/master/src/Shaders/passCube.fragment.fx. @deltakosh do you know why this is 1.001 rather than 1.0?

inspector/src/lod.ts
sebavan: would be great

DarraghBurkeMS: PR created

inspector/src/components/actionTabs/tabs/propertyGrids/materials/textures/textureCanvasManager.ts
DarraghBurkeMS: Resolved! Thanks

inspector/src/lod.ts
DarraghBurkeMS: Added a check for that! Thanks :)

sebavan:
You should not have any BABYLON. in your code as this breaks the build :-)

DarraghBurkeMS:
> You should not have any BABYLON. in your code as this breaks the build :-)

Fixed! Sometimes it's hard to context switch between the playground and the engine. Always having to remember whether or not to type the BABYLON

PR #8892, opened 2020-09-07
Add validation test for the KTX2 decoder

Popov72:
Let's see if it works on Azure server...


sebavan (approved):

sebavan:
Here is what we see in the failing tests if that helps:

image

Popov72:
I guess it means it does not work: I have removed the test from the integration run.

PR #8896, opened 2020-09-08
Gizmo parenting

CedricGuillemet:
This PR is a followup on this thread : https://forum.babylonjs.com/t/how-to-gizmo-on-empty-node/13723/6
Basically, mouse delta translation was transformed from world space coordinates to local space and used with the worldtransform of the node.
When the node had no parent, it was fine. But when parented, things went wrong. Now, plane and axis drag gizmo are working on world coordinates and local transform is computed in _matrixChanged function when needed.


sebavan (approved):
LGTM, will wait on @deltakosh approval later today in case I miss smthg

PR #8908, opened 2020-09-09
collision response flag 

CedricGuillemet:
Follow up on this thread : https://forum.babylonjs.com/t/mesh-checkcollision-but-pass-thru-other-mesh/13988
Basically, a flag that still triggers onCollide observable but disable collision response on position/velocity

Simple PG to test it when it's merged : https://www.babylonjs-playground.com/#B9NNK2


deltakosh (approved):
excellent!

PR #8909, opened 2020-09-09
Make motion blur + screen space reflections use prepass

CraigFeldspar:


src/Materials/PBR/pbrAdditionnalPrePassConfiguration.ts
```diff
@@ -0,0 +1,74 @@
+import { UniformBuffer } from "../../Materials/uniformBuffer";
+import { Matrix } from "../../Maths/math.vector";
+import { Scene } from "../../scene";
+import { PrePassRenderer } from "../../Rendering/prePassRenderer";
+
+export class PBRAdditionnalPrePassConfiguration {
+    public previousWorld: Matrix;
+    public previousViewProjection: Matrix;
+
+    constructor() {
```
deltakosh, line 10: No need right?

src/Materials/PBR/pbrPrePassConfiguration.ts
```diff
@@ -0,0 +1,74 @@
+import { UniformBuffer } from "../../Materials/uniformBuffer";
+import { Matrix } from "../../Maths/math.vector";
+import { Scene } from "../../scene";
+import { PrePassRenderer } from "../../Rendering/prePassRenderer";
+
+export class PBRAdditionnalPrePassConfiguration {
+    public previousWorld: Matrix;
+    public previousViewProjection: Matrix;
+
+    constructor() {
```
deltakosh, line 10: same

src/Rendering/ssao2Configuration.ts
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public readonly texturesRequired: number[] = [
+        PrePassRenderer.DEPTHNORMAL_TEXTURE_TYPE
+    ];
+
+    /**
+     * Builds a ssao2 configuration object
+     * @param scene The scene
+     */
+    constructor() {
```
deltakosh, line 31: we can remove empty functions

julien-moreau:
That’s awesome, no excuse for me to implement full metallic/roughness workflow for SSR!!

CraigFeldspar:
playground I used for tests : http://localhost:1338/Playground/index-local.html#PIZ1GK#169

src/Shaders/pbr.vertex.fx
```diff
@@ -173,13 +170,56 @@ void main(void) {
...
     vPositionW = vec3(worldPos);
-#ifdef PREPASS
-    vViewPos = (view * worldPos).rgb;
-#endif
+
+    #ifdef PREPASS_DEPTHNORMAL
+        vViewPos = (view * worldPos).rgb;
+    #endif
+
+    #if defined(PREPASS_VELOCITY) && defined(BONES_VELOCITY_ENABLED)
```
deltakosh, line 189: we should do it with an include to share the code

src/Materials/prePassConfiguration.ts
```diff
@@ -0,0 +1,70 @@
...
+     * Add the required samplers to the current list.
+     * @param samplers defines the current sampler list.
+     */
+    public static AddSamplers(samplers: string[]): void {
+        // pass
+    }
+
+    /**
+     * Binds the material data.
+     * @param effect defines theeffect to update
```
Popov72, line 45: nit: missing space

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1745,6 +1762,7 @@ export abstract class PBRBaseMaterial extends PushMaterial {
         // Matrices
         if (!defines.INSTANCES || defines.THIN_INSTANCES) {
             this.bindOnlyWorldMatrix(world);
+            this.prePassConfiguration.bindForSubMesh(this._activeEffect, scene, mesh, world, this.isFrozen);
```
Popov72, line 1765: Are you sure this should not be done for instances?

We don't pass the world matrix for instances because it's `world0/world1/world2/world3` which is used instead, but I do think the prepass config should be done for instances, as the velocity shader code is also applied in the instance case.

src/Materials/materialHelper.ts
```diff
@@ -303,37 +310,63 @@ export class MaterialHelper {
...
+        {
+            type: Constants.PREPASS_ALBEDO_TEXTURE_TYPE,
+            define: "PREPASS_ALBEDO",
+            index: "PREPASS_ALBEDO_INDEX",
+        },
+        {
+            type: Constants.PREPASS_DEPTHNORMAL_TEXTURE_TYPE,
+            define: "PREPASS_DEPTHNORMAL",
+            index: "PREPASS_DEPTHNORMAL_INDEX",
+        }];
```
Popov72, line 349: Is it intended that `PREPASS_COLOR_TEXTURE_TYPE` is not in the texture list?

src/Materials/materialHelper.ts
```diff
@@ -823,11 +857,27 @@ export class MaterialHelper {
...
         }
     }
 
+    // Copies the bones transformation matrices into the target array and returns the target's reference
+    private static _CopyBonesTransformationMatrices(source: Float32Array, target: Float32Array): Float32Array {
+        for (let i = 0; i < source.length; i++) {
+            target[i] = source[i];
+        }
+
+        return target;
```
Popov72, line 878: `target.set(source)` should probably be faster as the browser should be able to make something like a memcpy behind the hood.

src/Materials/standardMaterial.ts
```diff
@@ -1360,6 +1377,7 @@ export class StandardMaterial extends PushMaterial {
         // Matrices
         if (!defines.INSTANCES || defines.THIN_INSTANCES) {
             this.bindOnlyWorldMatrix(world);
+            this.prePassConfiguration.bindForSubMesh(this._activeEffect, scene, mesh, world, this.isFrozen);
```
Popov72, line 1380: Same comment than above, I think this line should be out of the `if` statement.

src/PostProcesses/motionBlurPostProcess.ts
```diff
@@ -78,19 +83,29 @@ export class MotionBlurPostProcess extends PostProcess {
      * @param textureType Type of textures used when performing the post process. (default: 0)
      * @param blockCompilation If compilation of the shader should not be done in the constructor. The updateEffect method can be used to compile the shader at a later time. (default: false)
      */
-    constructor(name: string, scene: Scene, options: number | PostProcessOptions, camera: Nullable<Camera>, samplingMode?: number, engine?: Engine, reusable?: boolean, textureType: number = Constants.TEXTURETYPE_UNSIGNED_INT, blockCompilation = false) {
+    constructor(name: string, scene: Scene, options: number | PostProcessOptions, camera: Nullable<Camera>, samplingMode?: number, engine?: Engine, reusable?: boolean, textureType: number = Constants.TEXTURETYPE_UNSIGNED_INT, blockCompilation = false, forceGeometryBuffer = false) {
```
Popov72, line 86: nit: missing the `@param forceGeometryBuffer ` doc.

src/PostProcesses/postProcess.ts
```diff
@@ -186,6 +188,12 @@ export class PostProcess {
     private _texelSize = Vector2.Zero();
     private _forcedOutputTexture: Nullable<InternalTexture>;
 
+    /**
+    * Prepass configuration in case this post process needs a texture from prepass
+    * @hidden
+    */
+    public prePassEffectConfiguration: PrePassEffectConfiguration;
```
Popov72, line 195: If it's hidden, maybe that's because this property is not meant to be used by the end user?

In that case, I think you should prefix it with an underscore.

src/PostProcesses/screenSpaceReflectionPostProcess.ts
```diff
@@ -68,7 +72,7 @@ export class ScreenSpaceReflectionPostProcess extends PostProcess {
      * @param textureType Type of textures used when performing the post process. (default: 0)
      * @param blockCompilation If compilation of the shader should not be done in the constructor. The updateEffect method can be used to compile the shader at a later time. (default: false)
      */
-    constructor(name: string, scene: Scene, options: number | PostProcessOptions, camera: Nullable<Camera>, samplingMode?: number, engine?: Engine, reusable?: boolean, textureType: number = Constants.TEXTURETYPE_UNSIGNED_INT, blockCompilation = false) {
+    constructor(name: string, scene: Scene, options: number | PostProcessOptions, camera: Nullable<Camera>, samplingMode?: number, engine?: Engine, reusable?: boolean, textureType: number = Constants.TEXTURETYPE_UNSIGNED_INT, blockCompilation = false, forceGeometryBuffer = false) {
```
Popov72, line 75: As above, missing the param doc for `forceGeometryBuffer`.

src/Materials/materialHelper.ts
CraigFeldspar: Yes, since color index is always 0, to keep consistency with non-mrt materials

src/Materials/PBR/pbrBaseMaterial.ts
CraigFeldspar: you are definitely right on that, but as of today, velocity (and therefore motion blur) is broken with instances, since we only calculate velocity from the main world matrix.  
This motion blur + instances issue should be fixed in another PR.

Anyway, in the meantime, I have put this line out of the if statement

PR #8916, opened 2020-09-09
Fix paddingBottom of children not taken into account in ScrollViewer

Popov72:
See https://forum.babylonjs.com/t/textblock-padding-doesnt-work-as-expected-inside-a-scrollviewer/13838/5


deltakosh (approved):

gui/src/2D/controls/scrollViewers/scrollViewerWindow.ts
```diff
@@ -301,8 +301,8 @@ export class _ScrollViewerWindow extends Container {
                 child._offsetTop(this._currentMeasure.top - child._currentMeasure.top);
             }
 
-            maxWidth = Math.max(maxWidth, child._currentMeasure.left - this._currentMeasure.left + child._currentMeasure.width);
-            maxHeight = Math.max(maxHeight, child._currentMeasure.top - this._currentMeasure.top + child._currentMeasure.height);
+            maxWidth = Math.max(maxWidth, child._currentMeasure.left - this._currentMeasure.left + child._currentMeasure.width + child.paddingRightInPixels);
+            maxHeight = Math.max(maxHeight, child._currentMeasure.top - this._currentMeasure.top + child._currentMeasure.height + child.paddingBottomInPixels);
```
Popov72, line 305: It seems `child._currentMeasure.left - this._currentMeasure.left` takes care of the `paddingLeft` and `child._currentMeasure.top - this._currentMeasure.top` of the `paddingTop`, so I just added the correction for `paddingRight` and `paddingBottom`.

I wonder if `child._currentMeasure.left - this._currentMeasure.left` should be replaced by `child.paddingLeftInPixels` and `child._currentMeasure.top - this._currentMeasure.top` by `child.paddingTopInPixels` but my understanding of all those measures is not strong enough to answer...

deltakosh: yeah this is correct. this is the problem with left and width.

Your fix is fine (we may want to review the overall process one day though)

deltakosh: wait... this should have been taken in account here:
https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/2D/controls/container.ts#L343

Edit: wait (2): unless the control is clipped then yeah we are missing it. But we will take it twice if !clipped

Popov72: I'm not sure there is a way to know that a control is clipped?

`Control._isClipped` let you know if a control is completely clipped from its parent, meaning not visible, but I guess you meant partially clipped?

deltakosh: well my point is that if !clipped we do not seem to integrate padding (right and bottom):
https://github.com/BabylonJS/Babylon.js/blob/master/gui/src/2D/controls/container.ts#L327

Popov72: They are integrated when `adaptWidthToChildren` / `adaptHeightToChildren` is true.

Note that:
> But we will take it twice if !clipped

Is not what I experienced with my correction on the sample PG: I needed to add the paddingBottom of the children to the children height to have a proper display in the scroll viewer (but I think it's specific to the scroll viewer).

Note that the scroll viewer (ScrollViewerWindow in fact) is computing its own width/height based on its children, it does not rely on the Container code you linked above.

deltakosh: That works for me then!

deltakosh:
Just a linting issue

Popov72:
It's a problem with the CapsuleBuilder:

```
/home/vsts/work/1/s/src/Meshes/Builders/capsuleBuilder.ts:207:1
ERROR: 207:1  no-trailing-whitespace  trailing whitespace
```
@Pryme8 FYI

deltakosh:
I fixed it theorically. can you sync with master?

Pryme8:
> It's a problem with the CapsuleBuilder:
> 
> ```
> /home/vsts/work/1/s/src/Meshes/Builders/capsuleBuilder.ts:207:1
> ERROR: 207:1  no-trailing-whitespace  trailing whitespace
> ```
> 
> @Pryme8 FYI

Did the sync fix it?

Pryme8:
> It's a problem with the CapsuleBuilder:
> 
> ```
> /home/vsts/work/1/s/src/Meshes/Builders/capsuleBuilder.ts:207:1
> ERROR: 207:1  no-trailing-whitespace  trailing whitespace
> ```
> 
> @Pryme8 FYI

Did the sync fix it?

Popov72:
Only WebGL validation tests remain to finish, to it seems it will work!

PR #8923, opened 2020-09-10
Add setCurrentPoseAsRest

Pryme8:
Just waiting to see if the build is successful.


deltakosh:
Failures are my bad
please sync with master and it should be fine

Pryme8:
Ok cool, I was trying to figure that out!

Also it seems like there is something up with bone.getLocalMatrix() it seems to be modifying the matrix?

Is this related?

deltakosh:
Nope not related. I did not change that

Popov72:
> Also it seems like there is something up with bone.getLocalMatrix() it seems to be modifying the matrix?

`getLocalMatrix` is modifying the matrix if there were some pending changes not integrated in the matrix yet so that you get an up to date matrix.

Pryme8:
Should we add this?
https://forum.babylonjs.com/t/bones-freeze-transformations/14036/22?u=pryme8

Popov72:
Waiting for inputs from @deltakosh or @bghgary, but if doing it I think you can add it as an option (true by default), in case we miss a use case where the user wouldn't want us to modify the textures...

PR #8949, opened 2020-09-13
SkeletonViewer: Added option to show local rotation axis on each bone.

reimund:
Added support for new displayOptions: `showLocalAxes` and `localAxesSize`.

The new option `showLocalAxes` defaults to false but when enabled it lets you debug the rotations of a skeleton:

Screenshot 2020-09-13 at 23 05 31

### DIFF
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -743,14 +743,11 @@ export class SkeletonViewer {
     }
 
     private _buildLocalAxes(): void {
-        if (this._localAxes) {
-            for (let axisMesh of this._localAxes) {
-                axisMesh.dispose();
-            }
-
-            this._localAxes = [];
+        for (let axisMesh of this._localAxes) {
+            axisMesh.dispose();
         }
 
+        this._localAxes = [];
         let displayOptions = this.options.displayOptions || {};
 
         if (!displayOptions.showLocalAxes) {
@@ -761,26 +758,24 @@ export class SkeletonViewer {
             const targetScene = this._utilityLayer.utilityLayerScene;
             const size = displayOptions.localAxesSize || 0.075;
 
-            if (targetScene) {
-                for (let b of this.skeleton.bones) {
-                    if (b._index === -1 || (!this._boneIndices.has(b.getIndex()) && !this.options.useAllBones)) {
-                        continue;
-                    }
-                
-                    let axisX = Mesh.CreateLines('axisX', [Vector3.Zero(), new Vector3(size, 0, 0)], targetScene, true);
-                    let axisY = Mesh.CreateLines('axisY', [Vector3.Zero(), new Vector3(0, size, 0)], targetScene, true);
-                    let axisZ = Mesh.CreateLines('axisZ', [Vector3.Zero(), new Vector3(0, 0, size)], targetScene, true);
-
-                    axisX.color = new Color3(1, 0, 0);
-                    axisY.color = new Color3(0, 1, 0);
-                    axisZ.color = new Color3(0, 0, 1);
-
-                    axisX.parent = axisY.parent = axisZ.parent = b;
-                    axisX.renderingGroupId = axisY.renderingGroupId = axisZ.renderingGroupId = this.renderingGroupId;
-
-                    this._localAxes.push(axisX, axisY, axisZ);
+            for (let b of this.skeleton.bones) {
+                if (b._index === -1 || (!this._boneIndices.has(b.getIndex()) && !this.options.useAllBones)) {
+                    continue;
                 }
-            }    
+            
+                let axisX = Mesh.CreateLines('axisX', [Vector3.Zero(), new Vector3(size, 0, 0)], targetScene, true);
+                let axisY = Mesh.CreateLines('axisY', [Vector3.Zero(), new Vector3(0, size, 0)], targetScene, true);
+                let axisZ = Mesh.CreateLines('axisZ', [Vector3.Zero(), new Vector3(0, 0, size)], targetScene, true);
+
+                axisX.color = new Color3(1, 0, 0);
+                axisY.color = new Color3(0, 1, 0);
+                axisZ.color = new Color3(0, 0, 1);
+
+                axisX.parent = axisY.parent = axisZ.parent = b;
+                axisX.renderingGroupId = axisY.renderingGroupId = axisZ.renderingGroupId = this.renderingGroupId;
+
+                this._localAxes.push(axisX, axisY, axisZ);
+            }	
         }
     }
 
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -299,7 +299,7 @@ export class SkeletonViewer {
 
     /** The SkeletonViewers Mesh. */
     private _debugMesh: Nullable<LinesMesh>;
-    
+
     /** The local axes Meshes. */
     private _localAxes: LinesMesh[] = [];
 
@@ -753,7 +753,7 @@ export class SkeletonViewer {
         if (!displayOptions.showLocalAxes) {
             return;
         }
-        
+
         const targetScene = this._utilityLayer!.utilityLayerScene;
         const size = displayOptions.localAxesSize || 0.075;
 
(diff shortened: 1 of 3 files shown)


src/Debug/skeletonViewer.ts
```diff
@@ -735,6 +742,48 @@ export class SkeletonViewer {
...
+        if (!displayOptions.showLocalAxes) {
+            return;
+        }
+        
+        if (this._utilityLayer) {
+            const targetScene = this._utilityLayer.utilityLayerScene;
+            const size = displayOptions.localAxesSize || 0.075;
+
+            if (targetScene) {
+                for (let b of this.skeleton.bones) {
```
Popov72, line 765: You can avoid some testings and indentation levels in this function because `this._localAxes`, `this.options.displayOptions`, `this._utilityLayer` and `targetScene` will always be non null and defined.

Popov72:
Only a nit, LGTM!

reimund:
@Popov72, I removed some of the checks, but my IDE was complaining about risk of _utilityLayer being null so I left that one in there.

reimund:
I noticed that the local axes only show if the skeleton is assigned to a mesh (ie `mesh.skeleton = mySkeleton`). If the `displayMode ` is set to `DISPLAY_SPHERE_AND_SPURS` this is always true, but for `DISPLAY_LINES` it might not be.

Adding the following to `_buildLocalAxes` works, but feels a bit awkward. Do you think it's an acceptable solution?

```javascript
if (this.displayMode === SkeletonViewer.DISPLAY_LINES) {
	axisX.skeleton = this.skeleton;
}
```

Popov72:
> @Popov72, I removed some of the checks, but my IDE was complaining about risk of _utilityLayer being null so I left that one in there.

You can use `this._utilityLayer!.utilityLayerScene` as you know that `_utilityLayer` won't be null/undefined.

Popov72:
> I noticed that the local axes only show if the skeleton is assigned to a mesh (ie mesh.skeleton = mySkeleton). If the displayMode is set to DISPLAY_SPHERE_AND_SPURS this is always true, but for DISPLAY_LINES it might not be.

I think that's because the world matrices of the bones are not recalculated if the skeleton is not applied to at least one mesh of the scene.

I think setting the `skeleton` property of the axis is ok, as the skeleton will be processed only a single time even if set to multiple meshes anyway, and it avoid to put the burden on the user in this case. Maybe add a comment to explain why this line is needed.

reimund:
Alright, thanks!

deltakosh:
Linting is reporting an issue

Pryme8:
I would argue that these should be in a single line system.  Otherwise that is ton of drawcalls.

reimund:
I tried merging them to one mesh but I failed. Then I figured the use case is to debug one or a few skeletons, not win a benchmark competition. Any pointers on how to merge them is appreciated though.

reimund:
@Pryme8 I figured out at least how to combine them to one mesh per bone...

src/Debug/skeletonViewer.ts
```diff
@@ -743,11 +744,11 @@ export class SkeletonViewer {
...
 
     private _buildLocalAxes(): void {
-        for (let axisMesh of this._localAxes) {
-            axisMesh.dispose();
+        if (this._localAxes) {
+            this._localAxes.dispose();
         }
-
-        this._localAxes = [];
+
```
Popov72, line 750: Why removing the array of `LinesMesh` as it was before?

You will only dispose the latest mesh created, meaning all other meshes will still be displayed even if not needed anymore.

reimund: You are right. That was a stupid mistake.

But now I started trying to make it into single mesh again. I got all the lines created and all, but I can't get the skinning to work. Is there any limitation stopping us from skinning a LinesMesh?

Popov72: It seems skinning is supported on a LineMesh, at least the vertex shader has the required bit of code.

Have you set the `skeleton` property of the LineMesh?

src/Debug/skeletonViewer.ts
```diff
@@ -772,17 +772,18 @@ export class SkeletonViewer {
...
-            }
+            this._localAxes.push(axes);
+        }
 
-            this._localAxes.parent = b;
-            this._localAxes.renderingGroupId = this.renderingGroupId;
+        if (this.displayMode === SkeletonViewer.DISPLAY_LINES && !this._localAxes[0].skeleton) {
+            // The local axes needs a mesh. Since the world matrix of a bone is not recalculated unless its
+            // skeleton is applied to at least one mesh. For the other display mode we don't have this issue.
+            this._localAxes[0].skeleton = this.skeleton;
```
Popov72, line 785: I would have applied the skeleton to all X axis because if `_localAxes[0]` leaves the frustum for whatever reason (animation / user moving the mesh), then the skeleton won't be processed anymore and the axis won't be displayed correctly.

reimund: I see. Thanks.

reimund: Question: Would it be good practice to include the pull request ID in each related commit?

src/Debug/skeletonViewer.ts
reimund: This is my current code.

I've realized I need to apply some rotation to the points passed to `CreateLineSystem` as well, since it's not parented like before.

As you can see I do call `this._localAxes.skeleton = this.skeleton`.

```javascript
        let mwk: number[] = [];
        let mik: number[] = [];
        const vertsPerBone = 6;

        for (let i in this.skeleton.bones) {
            let bone = this.skeleton.bones[i];
            
            if (bone._index === -1 || (!this._boneIndices.has(bone.getIndex()) && !this.options.useAllBones)) {
                continue;
            }

            let boneAbsoluteRestTransform = new Matrix();
            let anchorPoint = new Vector3();

            this.getAbsoluteRestPose(bone, boneAbsoluteRestTransform);
            boneAbsoluteRestTransform.decompose(undefined, undefined, anchorPoint);

            let linePoints = [
                [anchorPoint, new Vector3(anchorPoint.x + size, anchorPoint.y, anchorPoint.z)],
                [anchorPoint, new Vector3(anchorPoint.x, anchorPoint.y + size, anchorPoint.z)],
                [anchorPoint, new Vector3(anchorPoint.x, anchorPoint.y, anchorPoint.z + size)]
            ];
            let lineColors = [[red, red], [green, green], [blue, blue]];

            lines.push(...linePoints);
            colors.push(...lineColors);
            
            for (let j = 0; j < vertsPerBone; j++) {
                mwk.push(1, 0, 0, 0);
                mik.push(bone.getIndex(), 0, 0, 0);
            }
        }

        this._localAxes = LinesBuilder.CreateLineSystem('localAxes', { lines: lines, colors: colors, updatable: true }, targetScene);
        this._localAxes.setVerticesData(VertexBuffer.MatricesWeightsKind, mwk, false);
        this._localAxes.setVerticesData(VertexBuffer.MatricesIndicesKind, mik, false);
        this._localAxes.skeleton = this.skeleton;
        this._localAxes.renderingGroupId = this.renderingGroupId;
        this._localAxes.parent = this.mesh;
        this._localAxes.computeBonesUsingShaders = this.options.computeBonesUsingShaders ?? true;
```

Popov72: Hard to tell...

Have you tried to get the rotation matrix (quaternion) from `boneAbsoluteRestTransform` instead?

Also, you say it does not work, but do you still see something moving? Meaning, does applying a skeleton actually work on the LineMesh?

src/Debug/skeletonViewer.ts
Popov72: I don't think so, as nobody does this :)

src/Debug/skeletonViewer.ts
reimund: Yes, I tried using `boneAbsoluteRestTransform` but I didn't think it looked right. It accumulates rotations up the bone hierarchy, which I'm guessing is not desired in this case.

But perhaps I'm wrong. I guess it depends on how you interpret the local matrix passed to the bind pose (in the constructor that is). Is the rotation component of that matrix relative to the parent bone, or is it in world space?

Once this PR is working and correct, this can be made much more clear I think.

Popov72: > Yes, I tried using boneAbsoluteRestTransform but I didn't think it looked right. It accumulates rotations up the bone hierarchy, which I'm guessing is not desired in this case.

Yes, I think too.

So it would seem that `bone.getBindPose().getRotationMatrix();` is the right way to do it, don't know why it does not work...

reimund: It does work, with your fix to hardware skinning. So thank you!

reimund:
Now it's good to go!! Thanks to @Popov72 and @Pryme8. I owe you beers!

deltakosh:
Seems good to me. Only one ask: is that exposed in the Inspector?

deltakosh:
Ok maybe with another PR: https://github.com/BabylonJS/Babylon.js/blob/master/inspector/src/components/actionTabs/tabs/propertyGrids/meshes/skeletonPropertyGridComponent.tsx#L47

Pryme8:
I'm gonna message you about this too @reimund, as I am writing the SkeletonViewer docs as we speak and would like to include this in it.

reimund:
> I'm gonna message you about this too @reimund, as I am writing the SkeletonViewer docs as we speak and would like to include this in it.

Sure. I actually added all the other options to the table as well but figured it should not be part of this PR.

Slightly off topic, but I wonder if the `displayOptions` should be part of an enum or be static constants instead of magic strings?

PR #8953, opened 2020-09-14
Bones: fix returnToRest + updateMatrix

Popov72:
See https://forum.babylonjs.com/t/bones-freeze-transformations/14036/21

It does fix the problem with `returnToRest` in the PG, but as I don't really understand everything (anything) that's going on with the difference matrix, maybe it's not the right fix...

The fix to `upateMatrix` however is a required fix.

### DIFF
--- a/src/Bones/bone.ts
+++ b/src/Bones/bone.ts
@@ -244,7 +244,7 @@ export class Bone extends Node {
         if (this._skeleton._numBonesWithLinkedTransformNode > 0) {
             this.updateMatrix(this._restPose.clone(), false, false);
         } else {
-            this.updateMatrix(this._restPose.clone());
+            this.updateMatrix(this._restPose.clone(), false, true);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/Bones/bone.ts
```diff
@@ -244,7 +244,7 @@ export class Bone extends Node {
         if (this._skeleton._numBonesWithLinkedTransformNode > 0) {
             this.updateMatrix(this._restPose.clone(), false, false);
         } else {
-            this.updateMatrix(this._restPose.clone());
+            this.updateMatrix(this._restPose.clone(), false, true);
```
RaananW, line 247: I know this was already like that, but since you are already changing it - why use clone, if updateMatrix copies the content and doesn't change the matrix provided? This generates an unneeded object that is being GC'ed right after

Popov72: True, I have performed the change.

RaananW (approved):

PR #8957, opened 2020-09-14
Update WebXRHandTracking.ts

RaananW:
Temporarily remove XRHand compatibility test to avoid problems with compilation in certain environments.

Will find a solution at a later time


rickfromwork:
Does line 107 need to be updated as well?
"private generateHandPartsDefinition(hand: XRHand) {"

rickfromwork:
As it turns out, I was doing something wrong with my implementation in babylonNative, and mistakenly thought the problem was elsewhere. No need to complete this, unless other areas had difficulties as well.

RaananW:
Thanks for letting me know. PR closed, let me know if you still need any changes

PR #8961, opened 2020-09-15
Fixed issue with Connected Gamepads and DeviceSourceManager

PolygonalSun:
This PR fixes an issue where the DeviceInputSystem and the DeviceSourceManager do not properly detect, register, and handle gamepads that are already connected when a script using the DSM is run.


src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -109,6 +110,10 @@ export class DeviceSourceManager implements IDisposable {
      * @returns DeviceSource object
      */
     public getDeviceSource<T extends DeviceType>(deviceType: T, deviceSlot?: number): Nullable<DeviceSource<T>> {
+        if (!this._checkConnectedGamepads) {
```
Drigax, line 113: Why do we want to put this behavior behind a flag (defaulted false)?

PolygonalSun: The reason why I set this behind a flag is that I only want this to be run once and only when the first GetDeviceSource call happens.  All that this will do is just catch already connected gamepads that would otherwise miss being registered because they never throw a gamepadconnected event.

Drigax: If we only want this to happen once, and to catch already connected devices before our DSM was created, then why not do this at construction? Also, I'm not sure if I see any logic that ensures that this is only called once, and only on the first time this function is called...

Also, maybe invert the condition logic? it seems like you're checking for the opposite of what your variable name `DeviceSourceManager._checkConnectedGamepads` implies.

PolygonalSun: I can fix the variable naming.  As for why I have it here, if I run it in the constructor and the user adds any functions to the Connected/Disconnected observables, those custom observable functions would never run because all of the registration would take place before the user had a chance to add them to the observables.  By putting them in the getDeviceSource, I will get already connected gamepads to be registered and activate those functions.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -109,6 +110,11 @@ export class DeviceSourceManager implements IDisposable {
      * @returns DeviceSource object
      */
     public getDeviceSource<T extends DeviceType>(deviceType: T, deviceSlot?: number): Nullable<DeviceSource<T>> {
+        if (this._checkForConnectedGamepads) {
```
Drigax, line 113: I definitely think this wouldbe more at home in the DeviceSourceManager constructor since:
1) you're using a private class member for `this._checkForConnectedGamepads`
2) it shouldn't really affect the logic of `getDeviceSource` device source manager, since we can now  have an additional precondition guaranteeing that any existing devices will be enumerated if we construct one with this flag as a parameter.
3) it avoids awkward situations where we may want to disable this without needing to fiddle with class members (just pass it into the constructor) and it prevents us from overloading getDeviceSource to also enumerate devices as well...

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -103,6 +103,24 @@ export class DeviceInputSystem implements IDisposable {
         return device[inputIndex];
     }
 
+    /**
+     * Checks for existing connections to gamepads and registers them, if necessary
+     */
+    public checkForConnectedGamepads() {
```
Drigax, line 109: This should be ok to keep public, since we only call `deviceInputSystem.onDeviceConnected` once when we encounter a new device.

Drigax (approved):
Other than some general architectural concerns, looks great.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Just an update, after talking with Gary and testing some things, I'll be putting this call in the constructor.  The original idea for my choice was that I felt like custom callbacks should happen before devices are added to the DSM.  I wanted the users to be able to handle controller context switching when a device is connected (eg. showing Xbox controls when an xbox contorller is connected) but looking at it now, I don't think that this would properly support that scenario.  TBH, we should probably leave that to the users and keep things simple.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Just an update, after talking with Gary and testing some things, I'll be putting this call in the constructor. The original idea for my choice was that I felt like custom callbacks should happen before devices are added to the DSM. I wanted the users to be able to handle controller context switching when a device is connected (eg. showing Xbox controls when an xbox contorller is connected) but looking at it now, I don't think that this would properly support that scenario. TBH, we should probably leave that to the users and keep things simple.

PolygonalSun: tl;dr - You're right and I just needed to sleep on it.

deltakosh:
up.. can we merge?

PolygonalSun:
> up.. can we merge?

Not yet, I have to push up some additional changes

deltakosh:
Thanks for the draft :)

PolygonalSun:
> Thanks for the draft :)

So my changes are complete but I removed What's New update because it was suggested that it wasn't necessary.  Should I add it back or can it be merged without it?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -103,6 +103,19 @@ export class DeviceInputSystem implements IDisposable {
...
+     * Checks for existing connections to devices and register them, if necessary
+     */
+    public checkForConnectedDevices() {
+        const gamepads = navigator.getGamepads();
+
+        for (let i = 0; i < gamepads.length; i++) {
+            if (gamepads[i]) {
+                this._addGamePad(gamepads[i]);
+            }
+        }
```
bghgary, line 116: nit: this avoids indexing into the array twice
```suggestion
        for (const gamepad of gamepads) {
            if (gamepad) {
                this._addGamePad(gamepad);
            }
        }
```

bghgary (commented):
Did you figure out what caused this or was it already broken?

PolygonalSun:
> Did you figure out what caused this or was it already broken?

So I tried this with my test script and it looks like this was always failing.

deltakosh:
Can I merge?

PolygonalSun:
> Can I merge?
There's still one that thing that I need to do with this so not quite yet.

deltakosh:
Where are we with this PR?

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: I've made this private and the function will only be executed when a callback is assigned to onDeviceConnected.  This should still only execute once.

PolygonalSun:
> Where are we with this PR?

I've finished with my latest changes.  the connection observables/callback have been changed to handle gamepads that were already connected.  They will be executed to incorporate those devices into the Input System and Source Manager.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -49,19 +49,33 @@ export class DeviceSource<T extends DeviceType> {
 export class DeviceSourceManager implements IDisposable {
     // Public Members
     /**
-     * Observable to be triggered when before a device is connected
+     * Observable to be triggered when before a device is connected, any new observers added will be triggered against already connected devices
      */
-    public readonly onBeforeDeviceConnectedObservable = new Observable<{ deviceType: DeviceType, deviceSlot: number }>();
+    public readonly onBeforeDeviceConnectedObservable = new Observable<{ deviceType: DeviceType, deviceSlot: number }>((observer) => {
```
sebavan, line 54: Could be extracted as a type or interface for ease of use ?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -133,6 +147,24 @@ export class DeviceSourceManager implements IDisposable {
         return this._devices[deviceType].filter((source) => { return !!source; });
     }
 
+    /**
+     * Returns a read-only list of all available devices
+     * @returns Read-only array with active devices
+     */
+    public getListofDevices(): ReadonlyArray<{ deviceType: DeviceType, deviceSlot: number }> {
+        let deviceArray: Array<{ deviceType: DeviceType, deviceSlot: number }> = new Array<{ deviceType: DeviceType, deviceSlot: number }>();
```
sebavan, line 155: const ???

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -133,6 +147,24 @@ export class DeviceSourceManager implements IDisposable {
...
+     * @returns Read-only array with active devices
+     */
+    public getListofDevices(): ReadonlyArray<{ deviceType: DeviceType, deviceSlot: number }> {
+        let deviceArray: Array<{ deviceType: DeviceType, deviceSlot: number }> = new Array<{ deviceType: DeviceType, deviceSlot: number }>();
+        this._devices.forEach((deviceSet) => {
+            deviceSet.forEach((device) => {
+                const deviceType: DeviceType = device.deviceType;
+                const deviceSlot: number = device.deviceSlot;
+
+                deviceArray.push({ deviceType, deviceSlot });
```
sebavan, line 161: Should we return a copy or should we keep it internally as it is readonly ?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -13,10 +13,22 @@ declare const _native: any;
  * pointer device and one keyboard.
  */
 export class DeviceInputSystem implements IDisposable {
+
+    // Setter/Getter for onDeviceConnected property
```
sebavan, line 17: Can be removed

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -133,6 +147,24 @@ export class DeviceSourceManager implements IDisposable {
         return this._devices[deviceType].filter((source) => { return !!source; });
     }
 
+    /**
+     * Returns a read-only list of all available devices
+     * @returns Read-only array with active devices
+     */
+    public getListofDevices(): ReadonlyArray<{ deviceType: DeviceType, deviceSlot: number }> {
```
bghgary, line 154: The `o` in `of` is not capitalized? Maybe just `getDevices`?

src/DeviceInput/InputDevices/deviceSourceManager.ts
bghgary: `: Array<{ deviceType: DeviceType, deviceSlot: number }>` is extraneous

src/DeviceInput/InputDevices/deviceSourceManager.ts
bghgary: How about returning DeviceSource?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -49,19 +49,33 @@ export class DeviceSource<T extends DeviceType> {
...
-     * Observable to be triggered when before a device is connected
+     * Observable to be triggered when before a device is connected, any new observers added will be triggered against already connected devices
      */
-    public readonly onBeforeDeviceConnectedObservable = new Observable<{ deviceType: DeviceType, deviceSlot: number }>();
+    public readonly onBeforeDeviceConnectedObservable = new Observable<{ deviceType: DeviceType, deviceSlot: number }>((observer) => {
+        this.getListofDevices().forEach((device) => {
+            const deviceType: DeviceType = device.deviceType;
+            const deviceSlot: number = device.deviceSlot;
+
+            this.onBeforeDeviceConnectedObservable.notifyObserver(observer, { deviceType, deviceSlot });
```
bghgary, line 59: There is a small ordering issue with this. If someone adds to onAfterDeviceConnected before onBeforeDeviceConnected, then it will fire onAfterDeviceConnected first. Not sure if this will be an issue though.

I'm still not exactly sure why we have on before/after device connected events.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I can change this to getDevices

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: removed

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Changing this to DeviceSource

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Removing observables as discussed offline

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Fixing based on recommendations above

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: With the switch to DeviceSource, I'll just be returning a list of refs to existing DeviceSources

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -14,9 +14,19 @@ declare const _native: any;
...
     /**
-     * Callback to be triggered when a device is connected
+     * Returns onDeviceConnected callback property
+     * @returns Callback with function to execute when a device is connected
      */
-    public onDeviceConnected: (deviceType: DeviceType, deviceSlot: number) => void = () => { };
+    public get onDeviceConnected() { return this._onDeviceConnected; }
+
+    /**
+     * Sets callback function when a device is connected
```
bghgary, line 23: Do you need a similar comment here for if devices are already connected?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -14,9 +14,19 @@ declare const _native: any;
...
+     * @param callback Function to execute when a device is connected
+     */
+    public set onDeviceConnected(callback) {
+        this._onDeviceConnected = callback;
+        this._checkForConnectedDevices();
```
bghgary, line 28: What happens if the gamepad connection events fire between the constructor and someone setting this callback?

What about keyboard and mouse? Do these fire device connected?

PolygonalSun: If a gamepadconnected event fires before the callback is set, what _should_ happen is that the device will still be added to the DeviceInputSystem by the eventlistener body.  This is the same for the keyboard though the keyboard and mouse are initially added to the DeviceInputSystem at first input.

I did find an interesting bug though.  If a mouse button is pressed but not moving when the DSM is initialized, it won't register it properly.  I've changed a bunch of the logic around and will be treating the Mouse and Touch (if applicable) as already connected devices.  Since the keyboard will constantly fire keydown events when a button is held, I don't feel like it's necessary to do the same thing with the Keyboard.

bghgary: If you are going to treat keyboard/mouse as automatically connected, a comment would be good.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -125,7 +152,54 @@ export class DeviceInputSystem implements IDisposable {
         window.removeEventListener("gamepaddisconnected", this._gamepadDisconnectedEvent);
     }
 
+    /**
+     * Checks for existing connections to devices and register them, if necessary
+     * Currently handles gamepads and mouse
+     */
```
bghgary, line 158: Sorry I wasn't clear. I meant add comments to the contract so that people using the callbacks know that onConnected events for input manager / device source manager will not fire for keyboards.

PolygonalSun: I'm not sure I understand what you mean.  With how the code is set up, it will still fire the connected callback/observables with the first input of a keyboard (and any changes to the connected obserables or DIS callback).  Did you, by any chance, mean the onDisconnected events?  The only difference between keyboards/mice and other input methods is that they don't ever trigger a disconnect scenario.

bghgary: We chatted offline and straightened things out.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -125,7 +152,54 @@ export class DeviceInputSystem implements IDisposable {
...
+
+        for (const gamepad of gamepads) {
+            if (gamepad) {
+                this._addGamePad(gamepad);
+            }
+        }
+
+        // If the device in use has mouse capabilities, pre-register mouse
+        if (matchMedia('(pointer:fine)').matches) {
+            this._addPointerDevice(DeviceType.Mouse, 0, 0, 0);
```
bghgary, line 170: Does starting the mouse at 0,0 cause any issues?

bghgary (approved):

Issue #8777, opened 2020-08-20
NME - Update UX for consistency when exposing inputs/outputs to frame


PatrickRyanMS:
## Update UX for Consistency 
The UX around exposing an input/output from a node in the parent frame is not consistent when interacting with a constant input node like float, Vector2, Color3, Time, etc. When you click on an output of node that can be exposed to the frame, the slot on the node transitions to the highlight state but the node itself stays unselected. The inspector in NME displays the parameters to enter a custom name and a toggle to expose on the frame:

![image.png](https://images.zenhubusercontent.com/5dc2034af3f3b50001abf7aa/616c2a76-becd-48a2-9a1e-ebb768fdf622)

When you click on an output of a constant input node, which cannot be exposed on the frame for optimization reasons, both the slot and the node enter the highlighted state which is inconsistent. Also, the parameters displayed in the inspector are the top level node material parameters, not anything that has to do with the node. The user also needs to click away from the node in question and back onto the node to see the parameters for the node:

![image.png](https://images.zenhubusercontent.com/5dc2034af3f3b50001abf7aa/1cb0db8d-8309-49ce-b5b3-8c4715d88c96)

For consistency, if a user clicks on the output of a constant input node, we should show a message in the inspector that says "This node is a constant input node and cannot be exposed to the frame." We should also follow the same highlight states where the output is the only selected element and the user is able to click between the node and the output and update the inspector with the parameters/message for both.

PR #8962, opened 2020-09-15
NME - Update UX for consistency when exposing inputs/outputs to frame

msDestiny14:
UX

https://github.com/BabylonJS/Babylon.js/issues/8777


deltakosh (approved):
perfect thanks!

PR #8965, opened 2020-09-15
ishovered check for gizmos

CedricGuillemet:
followup on this thread: https://forum.babylonjs.com/t/how-to-prevent-scene-pick-on-axisdraggizmo-drag-start/14182/3
Store the isHovered boolean to be able to check later for picking

PG for testing : https://playground.babylonjs.com/#MT4TYN#24

### DIFF
--- a/src/Gizmos/gizmoManager.ts
+++ b/src/Gizmos/gizmoManager.ts
@@ -72,6 +72,20 @@ export class GizmoManager implements IDisposable {
     }
 
     /**
+     * True when the mouse pointer is hovering a gizmo mesh
+     */
+    public get isHovered() {
+        var hovered = false;
+        for (var key in this.gizmos) {
+            var gizmo = <Nullable<Gizmo>>((<any>this.gizmos)[key]);
+            if (gizmo) {
+                hovered = hovered || gizmo.isHovered;
+            }
+        }
+        return hovered;
+    }
+
+    /**
      * Instatiates a gizmo manager
      * @param scene the scene to overlay the gizmos on top of
      * @param thickness display gizmo axis thickness
--- a/src/Gizmos/gizmoManager.ts
+++ b/src/Gizmos/gizmoManager.ts
@@ -78,8 +78,9 @@ export class GizmoManager implements IDisposable {
         var hovered = false;
         for (var key in this.gizmos) {
             var gizmo = <Nullable<Gizmo>>((<any>this.gizmos)[key]);
-            if (gizmo) {
-                hovered = hovered || gizmo.isHovered;
+            if (gizmo && gizmo.isHovered) {
+                hovered = true;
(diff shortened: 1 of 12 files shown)


src/Gizmos/gizmoManager.ts
```diff
@@ -71,6 +71,20 @@ export class GizmoManager implements IDisposable {
...
 
+    /**
+     * True when the mouse pointer is hovering a gizmo mesh
+     */
+    public get isHovered() {
+        var hovered = false;
+        for (var key in this.gizmos) {
+            var gizmo = <Nullable<Gizmo>>((<any>this.gizmos)[key]);
+            if (gizmo) {
+                hovered = hovered || gizmo.isHovered;
```
Popov72, line 82: Nit: you could break here if you want!

src/Gizmos/positionGizmo.ts
```diff
@@ -87,6 +87,18 @@ export class PositionGizmo extends Gizmo {
...
         });
     }
+
+    /**
+     * True when the mouse pointer is hovering a gizmo mesh
+     */
+    public get isHovered() {
+        var hovered = false;
+        [this.xGizmo, this.yGizmo, this.zGizmo, this.xPlaneGizmo, this.yPlaneGizmo, this.zPlaneGizmo].forEach((gizmo) => {
+            hovered = hovered || gizmo.isHovered;
```
Popov72, line 97: This creates an array which is immediately GC each time `isHovered` is called (same thing below for the other gizmo classes).

I don't know how bad it is given the very small size of the array(?)

Popov72:
LGTM!

src/Gizmos/gizmoManager.ts
CedricGuillemet: I get 'Jump target cannot cross function boundary.' when trying to break. Do you have some code suggestion on the proper way to do it?

deltakosh:
Waiting for answers on the two comments and we can merge

src/Gizmos/positionGizmo.ts
CedricGuillemet: I will revisit that in the future.

src/Gizmos/gizmoManager.ts
deltakosh: pinging @Popov72

Popov72: I don't know, something like that works for me:
```typescript
        var toto: { [name: string] : number } = { a: 1, b: 2, c: 3 };
        for (var k in toto) {
            if (toto[k] === 2) {
                break;
            }
        }
```

CedricGuillemet: Fixed. Thank you @Popov72

PR #8969, opened 2020-09-15
Update decorators.ts

aWeirdo:
Ignore uniqueId on Serialize & Parse to keep them "unique per session".
https://doc.babylonjs.com/api/classes/babylon.uniqueidgenerator


deltakosh:
Thanks!

aWeirdo:
Didn't even mess up the linting, progress 😄 haha

yuqianma:
I think this commit breaks `SceneRecorder`. It relies on the `uniqueId` to diff.
Maybe we can ignore it only when parsing?

deltakosh:
rats! good point! 
I will check that

PR #8975, opened 2020-09-16
Fix typo: is ignore for now -> is ignored for now

Neob91:
JSDoc: Fix a typo, and remove a non-existent param

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -3113,7 +3113,7 @@ export class Quaternion {
 
     /**
      * Returns a new Vector3 set with the Euler angles translated from the current quaternion
-     * @param order is a reserved parameter and is ignore for now
+     * @param order is a reserved parameter and is ignored for now
      * @returns a new Vector3 containing the Euler angles
      */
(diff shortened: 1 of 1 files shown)


src/Maths/math.vector.ts
```diff
@@ -3125,7 +3125,7 @@ export class Quaternion {
     /**
      * Sets the given vector3 "result" with the Euler angles translated from the current quaternion
      * @param result defines the vector which will be filled with the Euler angles
-     * @param order is a reserved parameter and is ignore for now
+     * @param order is a reserved parameter and is ignored for now
```
Neob91, line 3128: @deltakosh Just noticed this function doesn't even accept `order` at all. Can I just drop this param in the JSDoc then? ;)

deltakosh: Yes! thanks for that

Neob91: Done
3fbcbc4b5a00ffe3c3054d07fabe891397adc5db

deltakosh:
Thanks buddy!

Issue #8981, opened 2020-09-16
Saving to Snippet Server should not auto expand, collapsed frames


PirateJC:
When saving to the snippet server, collapsed frames are auto expanded. They should not be. Collapsed frames should be saved in their collapsed form.

PR #8983, opened 2020-09-16
Fix to auto expanding frames

msDestiny14:
Oops!

https://github.com/BabylonJS/Babylon.js/issues/8981


PirateJC (changes requested):
It seems like we should remove line 1269 instead of commenting it out.

PirateJC (approved):

Issue #8979, opened 2020-09-16
Deleting a collapsed frame should also delete all nodes inside of it


PirateJC:
Currently when you have a frame in expanded mode, select the frame and hit delete, the frame itself is deleted by the nodes stay. This is expected behavior.

When a frame is in collapsed mode it acts the exact same way and should NOT. When a collapsed frame is selected and deleted, all nodes inside of it should also be deleted.

PR #8984, opened 2020-09-16
NME - Deleting collapsed node

msDestiny14:
Deleting a collapsed frame will now delete all the nodes inside.

https://github.com/BabylonJS/Babylon.js/issues/8979


PirateJC (approved):
Looks good to me, @deltakosh I have no idea why this line is there?

this.isCollapsed = false;

Why do we need to expand the frame when deleting it?

deltakosh:
i don't think so (maybe it was required earlier)

PR #8992, opened 2020-09-18
NME - Frame Port Ordering

msDestiny14:
Part 1 of 2.

Frame port order now maintained between opening and closing. Next step will be to export and reimport this setting.

### DIFF
--- a/nodeEditor/src/diagram/graphFrame.ts
+++ b/nodeEditor/src/diagram/graphFrame.ts
@@ -59,6 +59,8 @@ export class GraphFrame {
     private _frameInPorts: FrameNodePort[] = [];
     private _frameOutPorts: FrameNodePort[] = [];
     private _controlledPorts: NodePort[] = []; // Ports on Nodes that are shown on outside of frame
+    private _exposedInPorts: NodePort[] = [];
+    private _exposedOutPorts: NodePort[] = [];
     private _id: number;
     private _comments: string;
     private _frameIsResizing: boolean;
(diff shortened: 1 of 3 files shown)


nodeEditor/src/diagram/graphFrame.ts
```diff
@@ -127,69 +134,125 @@ export class GraphFrame {
     }
 
     private _createFramePorts() {
+
         for (var node of this._nodes) {
             node.isVisible = false;
+        }
+
+        for (var port of this._exposedOutPorts) { // Output
+            if(port.node == null || port.node.enclosingFrameId != this.id) {
```
RaananW, line 143: nit - try as much as possible to use `===` instead of `==` . type safe and faster to execute

nodeEditor/src/diagram/graphFrame.ts
```diff
@@ -677,12 +740,17 @@ export class GraphFrame {
                 return;
             }
             elementsArray = Array.from(this._inputPortContainer.childNodes);
+            const indexInContainer = this._frameInPorts.findIndex(framePort => framePort === nodePort);
```
RaananW, line 743: (same goes to the other findIndex calles) - is it guarantied that the element is inside the array? Or do you need to check that the index is 0 and up?

RaananW (approved):
Great! :-)

With two small suggestions/questions

deltakosh (approved):

nodeEditor/src/diagram/graphFrame.ts
msDestiny14: This is guarantied. In fact the function call below this line also does a similar search with no bounds checking.

nodeEditor/src/diagram/graphFrame.ts
deltakosh: Did you resolve it? I still see the ==

deltakosh: yes you did 👍

deltakosh:
Thanks you rock! And you are faster every day! I love that

PR #8994, opened 2020-09-18
Bone: New method to freeze rotation 

reimund:
This new method allows you to update the zero rotation state for a given bone.

A bone with `rotationQuaternion = new BABYLON.Quaternion(0, 0, 0, Math.PI / 4)`.
Then set to zero with `rotationQuaternion = Quaternion.Zero()`.

The same bone with `rotationQuaternion = new BABYLON.Quaternion(0, 0, 0, Math.PI / 4)` followed by call to `freezeRotation()`.
Then set to zero with `rotationQuaternion = Quaternion.Zero()`.

### DIFF
--- a/src/Bones/bone.ts
+++ b/src/Bones/bone.ts
@@ -1150,8 +1150,8 @@ export class Bone extends Node {
 	}
 
 	/**
-     * Freezes the rotations of the bone. The result is a bone in the same pose but
-	 * with a zeroed local rotation.
+     * Freezes the rotations of the bone by updating the base matrix. The result is a
+	 * bone in the same pose but with a zeroed local rotation.
      * @param updateRestPose defines if the rest pose should be updated
      */
 	public freezeRotation(updateRestPose: boolean = false) {
--- a/src/Bones/bone.ts
+++ b/src/Bones/bone.ts
@@ -191,7 +191,7 @@ export class Bone extends Node {
     }
 
     /**
-     * Gets the base matrix (initial matrix which remains unchanged)
+     * Gets the base matrix.
      * @returns a matrix
      */
     public getBaseMatrix(): Matrix {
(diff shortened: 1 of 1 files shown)


RaananW (commented):
Thanks for the PR!

Not commenting about functionality (will let @deltakosh do that), please try as much as possible to reuse objects instead of creating new ones. You are creating 10 objects on each call of this function.
I know this doesn't happen on each frame, but in many places (for example setting `this.rotationQuaternion` to be a new Zero quaternion) can be solved by using the current object instead of triggering the GC after creating a new object. 
If needed you can use the global temp math obejcts (quats, vectors etc'), or even create private temporary objects in the class itself.
Oh, and make sure linting is correct (you can see the failed linting task and what exactly is the problem in the build log)

reimund:
Thanks @RaananW!

I tend to always forget linting. >_<
It all looks so nice in my Visual Studio Code window :)

I will fix it.

reimund:
@RaananW, does this look better? I've never used the TmpVectors before so I hope I used them correctly.

RaananW:
> @RaananW, does this look better? I've never used the TmpVectors before so I hope I used them correctly.

Sure! This is a valid way to use them. Just wanted to add - use multiplyToRef and Matrix.ComposeToRef and you are not creating any object in this entire function.

Popov72:
I have one comment: did you test this code with a skeleton mesh loaded from a gltf/glb file?

Those are not using the same code path/functionalities from the bone/skeleton classes, so it would be worth to check it's also working in that case.

And I think you could also add this method to the `skeleton` class so we can update all bones at once (so I lied, I had two comments!).

RaananW:
> so I lied, I had two comments!.
😆

deltakosh:
I struggle to understand the need for this feature. When do you want to do that? Why? (just need convincing ;))

reimund:
> I struggle to understand the need for this feature. When do you want to do that? Why? (just need convincing ;))

I'm sure there are other use cases, but in my specific case it's to line up my Maya imported rig with motion capture data.

There will be no post-processing of the mocap data and the T-Pose in the data can be facing any direction. So I want to rotate my rig to match the T-Pose but I also need to freeze the rotations before applying the animation or the skinning will blow up.

I started a thread about it on the forum and this became my solution.
https://forum.babylonjs.com/t/bones-freeze-transformations/14036/24

PR #8995, opened 2020-09-18
Renamed `getAbsoluteRestPose` -> `getAbsoluteBindPoseToRef` + related vars.

reimund:
I figured since this private method uses `getBindPose` that should be reflected in the name. Also seems like `ToRef` is the naming convention that should be used.

Also renamed related variables to follow this change.

### DIFF
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -560,13 +560,14 @@ export class SkeletonViewer {
         }
     }
 
-    private getAbsoluteRestPose(bone: Nullable<Bone>, matrix: Matrix) {
+    /** function to get the absolute bind pose of a bone by accumulating transformations up the bone hierarchy. */
+    private getAbsoluteBindPoseToRef(bone: Nullable<Bone>, matrix: Matrix) {
         if (bone === null || bone._index === -1) {
             matrix.copyFrom(Matrix.Identity());
             return;
         }
 
-        this.getAbsoluteRestPose(bone.getParent(), matrix);
+        this.getAbsoluteBindPoseToRef(bone.getParent(), matrix);
         bone.getBindPose().multiplyToRef(matrix, matrix);
         return;
     }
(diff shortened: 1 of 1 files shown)


reimund:
Damnit, I have my IDE set up to use the same indentation as the file, but it never seems to work.

src/Debug/skeletonViewer.ts
```diff
@@ -560,13 +560,14 @@ export class SkeletonViewer {
         }
     }
 
-    private getAbsoluteRestPose(bone: Nullable<Bone>, matrix: Matrix) {
+    /** function to get the absolute bind pose of a bone by accumulating transformations up the bone hierarchy. */
+    private getAbsoluteBindPoseToRef(bone: Nullable<Bone>, matrix: Matrix) {
```
Popov72, line 564: nit: private functions should have a leading underscore in their name (I think we missed this one in the previous PR).

Popov72:
Apart from a nit, LGTM.

src/Debug/skeletonViewer.ts
```diff
@@ -611,18 +612,18 @@ export class SkeletonViewer {
...
                 let anchorPoint = new Vector3();
 
-                boneAbsoluteRestTransform.decompose(undefined, undefined, anchorPoint);
+                boneAbsoluteBindPoseTransform.decompose(undefined, undefined, anchorPoint);
 
                 bone.children.forEach((bc, i) => {
-                    let childAbsoluteRestTransform : Matrix = new Matrix();
-                    bc.getRestPose().multiplyToRef(boneAbsoluteRestTransform, childAbsoluteRestTransform);
+                    let childAbsoluteRestPoseTransform : Matrix = new Matrix();
+                    bc.getRestPose().multiplyToRef(boneAbsoluteBindPoseTransform, childAbsoluteRestPoseTransform);
```
reimund, line 624: Is it correct to use `getRestPose` here? In the helper function we use `getBindPose`. Is there a reason for not using the the same when figuring out the child bone position?

deltakosh: @Pryme8  / @Popov72  / @CedricGuillemet

Pryme8:
I am not sure about this part.  I think @popov72

src/Debug/skeletonViewer.ts
Popov72: Yes, as it is computing an absolute bind pose matrix it should start by using the local bind matrix, not the rest pose.

src/Debug/skeletonViewer.ts
Popov72: Yes, as it is computing an absolute bind pose matrix it should start by using the local bind matrix, not the rest pose.

reimund: Are you saying that it should be changed to `getBindPose`?

Popov72: yes

Pryme8:
This is not breaking any backwards compatibility is it?

Popov72:
I guess some testing should be performed to check it is ok.

Note that there's no really backward compatibility problem as if I'm not mistaken the part of the skeleton viewer which is using this code is new in 4.2.

Pryme8:
> I guess some testing should be performed to check it is ok.
> 
> Note that there's no really backward compatibility problem as if I'm not mistaken the part of the skeleton viewer which is using this code is new in 4.2.

I think that is for the lines view isn't it?   I'm pretty sure I never used it in the new stuff.

deltakosh:
Are we good to merge then?

reimund:
> Are we good to merge then?

I think so!

PR #8998, opened 2020-09-18
adds the ability to loadscript for babylon native

CedricGuillemet:
fixes https://github.com/BabylonJS/BabylonNative/issues/219

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -273,6 +273,20 @@ export class NativeEngine extends Engine {
 
         Tools.Log("Babylon Native (v" + Engine.Version + ") launched");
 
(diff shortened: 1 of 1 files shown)


src/Engines/nativeEngine.ts
```diff
@@ -273,6 +273,20 @@ export class NativeEngine extends Engine {
 
         Tools.Log("Babylon Native (v" + Engine.Version + ") launched");
 
+        Tools.LoadScript = function (scriptUrl, onSuccess, onError, scriptId) {
+            Tools.LoadFile(scriptUrl, (data) => {
+                Function(data as string).apply(null);
```
RaananW, line 278: That's a great wai to parse a script string and execute it :-)

You can ask for an optional context (`this`) and replace it with the null, if provided.

RaananW (approved):

PR #8999, opened 2020-09-18
missing function mandatory for some validation tests

CedricGuillemet:
Needed to put more tests in validation


src/Engines/nativeEngine.ts
```diff
@@ -951,6 +951,22 @@ export class NativeEngine extends Engine {
...
+     * @param format defines the format of the data
+     * @param forceBindTexture if the texture should be forced to be bound eg. after a graphics context loss (Default: false)
+     */
+    public updateDynamicTexture(texture: Nullable<InternalTexture>, canvas: HTMLCanvasElement, invertY: boolean, premulAlpha: boolean = false, format?: number): void {
+        if (texture) {
+            texture.isReady = true;
+        }
+        // TODO: Stub! This function is needed for some GLTF validation tests.
+    }
+
```
bghgary, line 969: I'm not sure I understand why we are adding a stub function. Can you explain? If this function is being called, doesn't that break the test that is running?

CedricGuillemet: Newly added validation tests (GLTF tests) call that function to get a texture with text from canvas. That text texture is rendered to get some hints on the validation rendered picture. That worked great with d3d: invalid texture bound = empty transparent texture. But on Opengl, I got quads textured with previously bound texture. So, I had to update the texture with a transparent .png to get a proper texture. Can discuss that offline.

deltakosh:
Can I merge?

CedricGuillemet:
Nop, putting it back to draft as I want to push some more functions.

CedricGuillemet:
Ready for review/merge!

src/Engines/nativeEngine.ts
CedricGuillemet: new issue for this function : https://github.com/BabylonJS/BabylonNative/issues/423

bghgary (approved):

Issue #8980, opened 2020-09-16
Custom frame port order not preserved when exporting


PirateJC:
When creating a frame with multiple input and output ports...you can specify the order you want the ports to display in, when that frame is in collapsed mode.

Exporting the frame does NOT preserve the sort order of input OR export ports. The sort order always exports as alphabetical but should preserve the user's sort order.

PR #9002, opened 2020-09-19
Custom Frame Port Order Preserved When Exporting

msDestiny14:
Part 2 of 2

https://github.com/BabylonJS/Babylon.js/issues/8980


nodeEditor/src/diagram/graphFrame.ts
```diff
@@ -1415,9 +1426,35 @@ export class GraphFrame {
         } else {
             newFrame.refresh();
         }
+        
+        
+        for (var node of newFrame.nodes) {
+            for (var port of node.outputPorts) { // Output
+                if(port.exposedOnFrame) {
+                    port.isExposed = true;
```
deltakosh, line 1434: why do we have isExposed and exposedonFrame?

msDestiny14: The difference is one is weather it actually is exposed or is the user desires it to be exposed. In theory it would make sense to change the variable name now but the json export name can't change cause of backwards compatibility.

deltakosh: ok makes sense! thanks

PR #9004, opened 2020-09-19
Update PBR clear coat and sheen to latest spec

### LINKED ISSUE #9001: Update PBR sheen and clearcoat to the latest Khronos Spec
Sheen:
https://github.com/KhronosGroup/glTF/pull/1688

Clearcoat:
https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_clearcoat

Popov72:
Fix #9001.


sebavan (approved):

sebavan:
It looks like there might be a compilation issue in Webgl1

Popov72:
Fixed

bghgary:
I guess we don't have any validation tests for clearcoat or sheen?

bghgary:
I'll work on updating the [sample model](https://github.com/KhronosGroup/glTF-Sample-Models/pull/266).

bghgary:
It's updated here: https://github.com/KhronosGroup/glTF-Sample-Models/pull/266

Popov72:
> I guess we don't have any validation tests for clearcoat or sheen?

We have:
* Node material PBR 1: https://playground.babylonjs.com/#D8AK3Z#7
* PBR shader code coverage:
  * 1: https://playground.babylonjs.com/#QI7TL3#16
  * 2: https://playground.babylonjs.com/#QI7TL3#17
  * 3: https://playground.babylonjs.com/#QI7TL3#18

Those PGs are activating clearcoat and sheen.

Notably, the last 3 are testing different settings of the clearcoat and sheen modules.

Do you want me to add a validation test with the sheen asset?

loaders/src/glTF/2.0/Extensions/KHR_materials_sheen.ts
```diff
@@ -84,7 +84,15 @@ export class KHR_materials_sheen implements IGLTFLoaderExtension {
...
 
+        if (properties.sheenRoughnessTexture) {
+            promises.push(this._loader.loadTextureInfoAsync(`${context}/sheenRoughnessTexture`, properties.sheenRoughnessTexture, (texture) => {
+                texture.name = `${babylonMaterial.name} (Sheen Roughness)`;
+                babylonMaterial.sheen.textureRoughness = texture;
+            }));
+        }
+
         babylonMaterial.sheen.albedoScaling = true;
+        babylonMaterial.sheen.useRoughnessFromMainTexture = false;
```
bghgary, line 95: Should we check to see if the two textures can be combined and set this to true? Might have to check for KHR_texture_transform which will make this a bit complicated. Ditto for sheen.

bghgary:
> Do you want me to add a validation test with the sheen asset?

Are the PGs you mentioned not already in the validation tests? If not, then yes, I think we should add one.

loaders/src/glTF/2.0/Extensions/KHR_materials_sheen.ts
Popov72: The check is done by the PBR code (sheen and clearcoat doing the same thing here): if both (internal) textures are identical, only one is passed to the shader and is used for both cases.

Popov72:
Yes, the PGs I have linked are already in the validation test list.

loaders/src/glTF/2.0/Extensions/KHR_materials_sheen.ts
bghgary: Hrm, I looked for it and didn't see it. I must have missed it.

Popov72: Look for CLEARCOAT_TEXTURE_ROUGHNESS_IDENTICAL  and SHEEN_TEXTURE_ROUGHNESS_IDENTICAL

src/Materials/PBR/pbrClearCoatConfiguration.ts
```diff
@@ -186,6 +215,8 @@ export class PBRClearCoatConfiguration {
     public prepareDefines(defines: IMaterialClearCoatDefines, scene: Scene): void {
         if (this._isEnabled) {
             defines.CLEARCOAT = true;
+            defines.CLEARCOAT_USE_ROUGHNESS_FROM_MAINTEXTURE = this._useRoughnessFromMainTexture;
+            defines.CLEARCOAT_TEXTURE_ROUGHNESS_IDENTICAL = this._texture !== null && this._texture._texture === this._textureRoughness?._texture;
```
bghgary, line 219: Is this good enough? What if the texture transform or sampling mode is different?

bghgary:
> Yes, the PGs I have linked are already in the validation test list.

I mean glTF models with clearcoat and sheen, not just clearcoat and sheen directly.

Popov72:
> > Yes, the PGs I have linked are already in the validation test list.
> 
> I mean glTF models with clearcoat and sheen, not just clearcoat and sheen directly.

Ok, then no there isn't. Have you a model for the clearcoat validation test?

bghgary:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/ClearCoatTest would be good.

src/Materials/PBR/pbrClearCoatConfiguration.ts
Popov72: The sampling mode is a property of `InternalTexture`, so that one is ok.

I had simplified the implementation to use the matrix transformation of the main texture if there is a main texture and the matrix transformation of the roughness texture if there is no main texture but a roughness texture.

Going to perform the changes to take into account the matrices from both textures.

bghgary: Sounds good

Popov72:
Code is updated, waiting for https://github.com/BabylonJS/Assets/pull/16 to be merged before adding validation tests.

deltakosh:
Ok Gary has a feedback on the assets so waiting to merge it

deltakosh:
we do have some conflicts to fix (I merged the assets)

Popov72:
Fixed and validation tests added.

deltakosh:
it fails :(

Popov72:
There is a 3mn timeout on the clear coat test, but it's exactly the same than for sheen which works... Only the asset is different, but it does work here: https://playground.babylonjs.com/#YG3BBF#1

deltakosh:
@sebavan can we increase the timeout?

Popov72:
In fact the asset is quite small for clear coat: only 233 ko...

And it is bigger for sheen: 4.86Mo in all.

clear coat asset is glb whereas it is gltf for sheen...

Popov72:
There was a conflict with the clear coat file, let's see how it goes now.

Popov72:
Still does not work, trying another PG...

Popov72:
Well, it seems the dev server does not like the `ClearCoatTest.glb` asset, I can't make it to work...

deltakosh:
Make sure there is no upper caps / lower caps issues ?

Popov72:
It is working when using it in the PG: https://playground.babylonjs.com/#YG3BBF#1

deltakosh:
can we try with the gltf version?

Popov72:
Waiting for https://github.com/BabylonJS/Assets/pull/17 to be in before changing the test.

Popov72:
Same problem, timeout...

deltakosh:
There is something wrong here adding @sebavan  to have his thougts

deltakosh:
up?

sebavan:
Still need to look into the tests should be able to do it tomorrow night.

Popov72:
I have excluded the clearcoat unit test from automatic testing so that the PR can be merged: @sebavan is currently looking at the problem in another PR.

PR #9011, opened 2020-09-21
KTX2 loader: Allow configuring the urls used by the decoder

Popov72:
See https://forum.babylonjs.com/t/ktx2-transcoder-support-progress/13960/12


bghgary (approved):

deltakosh:
some checks are failing

PR #9012, opened 2020-09-21
Use geometry buffer as a fallback for prepass

CraigFeldspar:
This PR should be merged after #8909 

Test playground : http://localhost:1338/Playground/index-local.html#PIZ1GK#177

Important note : 
+ As discussed offline, materials which are not "image processing ready" will have their color converted to gamma space. I converted `normalMaterial` to "image processing ready" as a proposition of implementation. This is done after the model of implementation for fog. If the proposition is accepted, I will then convert all materials from the material library.


materialsLibrary/src/normal/normal.fragment.fx
```diff
@@ -97,4 +97,8 @@ void main(void) {
 #include<fogFragment>
 
 	gl_FragColor = color;
+
+#ifdef IMAGEPROCESSINGPOSTPROCESS
```
CraigFeldspar, line 101: this should be replaced with a `#include <linearTransform.fx>`, but when doing this, a weird 301 redirection to an invalid URL happens in the playground. Any idea why ?

deltakosh: discussed offline

Popov72:
In `imageProcessingCompatibility.fx`, shouldn't you also check that the pre pass renderer is used before converting to linear?

CraigFeldspar:
@deltakosh  @sebavan ready for review 👍 

@Popov72 about `imageProcessingCompatibility.fx`, actually no because we have the exact same problem when imageprocessing is used as a post process, regardless of prepass

src/Rendering/geometryBufferRenderer.ts
```diff
@@ -31,6 +33,11 @@ interface ISavedTransformationMatrix {
  * This renderer is helpfull to fill one of the render target with a geometry buffer.
  */
 export class GeometryBufferRenderer {
+    /**
+     * Constant used to retrieve the depth + normal texture index in the G-Buffer textures array
+     * using getIndex(GeometryBufferRenderer.DEPTHNORMAL_TYPE)
+     */
+    public static readonly DEPTHNORMAL_TEXTURE_TYPE = 0;
```
Popov72, line 40: nit: DEPTHNORMAL_TYPE should read DEPTHNORMAL_TEXTURE_TYPE in the doc

src/Rendering/geometryBufferRenderer.ts
```diff
@@ -79,14 +86,103 @@ export class GeometryBufferRenderer {
...
 
     protected _effect: Effect;
     protected _cachedDefines: string;
 
+    /**
+     * @hidden
+     * Sets up internal structures to share outputs with PrePassRenderer
+     * This method should only be called by the PrePassRenderer itself
+     */
+    public linkPrePassRenderer(prePassRenderer: PrePassRenderer) {
```
Popov72, line 103: nit: the name of a hidden public method should start with an underscore I think because it is not meant to be used by the end user: @deltakosh can you confirm?

src/Rendering/geometryBufferRenderer.ts
```diff
@@ -79,14 +86,103 @@ export class GeometryBufferRenderer {
...
+        this._enableVelocity = false;
+        this._attachments = [];
+    }
+
+    /**
+     * @hidden
+     * Replaces a texture in the geometry buffer renderer
+     * Useful when linking textures of the prepass renderer
+     */
+    public replaceTexture(texture: Texture, geometryBufferType: number, index: number) {
```
Popov72, line 142: The texture parameter is not used in the function?

src/Rendering/prePassRenderer.ts
```diff
@@ -257,6 +317,48 @@ export class PrePassRenderer {
         }
     }
 
+    private _updateGeometryBufferLayout() {
+        if (this._geometryBuffer) {
+            this._geometryBuffer.resetLayout();
+
+            const attachments = this._defaultAttachments.slice();
+            const gl = this._scene.getEngine()._gl;
+            attachments[0] = gl.NONE;
```
Popov72, line 326: I'm not sure it's allowed to access the gl layer in such high level classes... @sebavan ?

Maybe we need a function in the engine to retrieve this value?

src/Rendering/prePassRenderer.ts
```diff
@@ -257,6 +317,48 @@ export class PrePassRenderer {
...
+                    geometryBufferConstant: GeometryBufferRenderer.VELOCITY_TEXTURE_TYPE,
+                }
+            ];
+
+            // replace textures in the geometryBuffer RT
+            for (let i = 0; i < matches.length; i++) {
+                const index = this._mrtLayout.indexOf(matches[i].prePassConstant);
+                if (index !== -1) {
+                    this._geometryBuffer.replaceTexture(this.prePassRT.textures[index], matches[i].geometryBufferConstant, index);
+                    attachments[index] = (<any>gl)["COLOR_ATTACHMENT" + index];
```
Popov72, line 354: Same thing here, it is some webgl specific code.

src/Rendering/prePassRenderer.ts
```diff
@@ -418,7 +521,8 @@ export class PrePassRenderer {
 
         if (!this.enabled) {
             // Prepass disabled, we render only on 1 color attachment
-            this._engine.bindAttachments([this._engine._gl.COLOR_ATTACHMENT0]);
+            this._engine.restoreDefaultFramebuffer();
+            this._engine.bindAttachments([this._engine._gl.BACK]);
```
Popov72, line 525: Here too.

Popov72:
LGTM !

deltakosh (approved):

sebavan (commented):
LGTM, I wonder if we should auto inject imageProcessingCompatibility.fx  if it is not in the file for shaderMaterials or nodeMaterials for instance ???

CraigFeldspar:
> LGTM, I wonder if we should auto inject imageProcessingCompatibility.fx if it is not in the file for shaderMaterials or nodeMaterials for instance ???

@sebavan you mean by processing the shader code ?

sebavan:
Yup just wondering if it is worth it or if a good documentation would be better ???

@deltakosh, what would you prefer ?

deltakosh:
If we can avoid the doc I think it is better

CraigFeldspar:
actually, I thought about injecting this way, but it turns out it could go bad :

for instance : 
```
if (...) {
 gl_Fragcolor = ...
 return;
}
```

Injection at the end of `main`  would have no effect when there is branching

sebavan:
So I guess it could only be doc :-( but at least in the Node Material we could probably add it ???

sebavan:
It is in so perfect !!! and we only need it for the shaderMaterial in the doc then ?

sebavan (approved):

CraigFeldspar:
> It is in so perfect !!! and we only need it for the shaderMaterial in the doc then ?

yep, @deltakosh I'll make you some sweet doc for that :)

PR #9014, opened 2020-09-22
SkeletonViewer: Added new display option: spurFollowsChild.

reimund:
When enabling this option (currently disabled by default) the skinning of the spur is altered so that the far end of the spur (the one that's supposed to meet the child bone) is skinned to the child bone rather than the current bone.

In my retargeting use case where bone positions are set explicitly the current behaviour leads to spurs that float around unattached. This solution will not always show a perfectly shaped spur since it can be deformed by the child bone, but it has the advantage of being very fast while still maintaining a debug view of the skeleton that's not blown up.

Related [Playground](https://www.babylonjs-playground.com/#7GM6J3#4).

`spurFollowsChild` disabled (default):
Screenshot 2020-09-22 at 10 05 32

`spurFollowsChild` enabled:
Screenshot 2020-09-22 at 10 05 56

### DIFF
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -673,6 +673,8 @@ export class SkeletonViewer {
                     for (let i = 0; i < numVertices; i++) {
                         mwk.push(1, 0, 0, 0);
 
+                        // Select verts at end of spur (ie vert 10 to 14) and bind to child
+                        // bone if spurFollowsChild is enabled.
                         if (displayOptions.spurFollowsChild && i > 9) {
                             mik.push(bc.getIndex(), 0, 0, 0);
                         }
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -389,6 +389,7 @@ export class SkeletonViewer {
         options.displayOptions.sphereBaseSize = options.displayOptions.sphereBaseSize ?? 0.15;
         options.displayOptions.sphereScaleUnit = options.displayOptions.sphereScaleUnit ?? 2;
         options.displayOptions.sphereFactor = options.displayOptions.sphereFactor ?? 0.865;
+        options.displayOptions.spurFollowsChild = options.displayOptions.spurFollowsChild ?? false;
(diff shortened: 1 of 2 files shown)


src/Debug/skeletonViewer.ts
```diff
@@ -666,19 +667,25 @@ export class SkeletonViewer {
...
-                    spur.convertToFlatShadedMesh();
-
                     let numVertices = spur.getTotalVertices();
                     let mwk: number[] = [], mik: number[] = [];
 
                     for (let i = 0; i < numVertices; i++) {
                         mwk.push(1, 0, 0, 0);
-                        mik.push(bone.getIndex(), 0, 0, 0);
+
+                        if (displayOptions.spurFollowsChild && i > 9) {
```
Popov72, line 676: I think `i > 9` was for debugging purpose?

Popov72 (commented):
LGTM save for the small debug thing to remove.

src/Debug/skeletonViewer.ts
reimund: No, that's the logic that selects the vertices to be bound to the child bone.

reimund: The spur consists of an extruded square of five vertices. It is extruded in three steps. The first square (vert 0 to 4) are on the bone, the second square (vert 5 to 9) are on the `midStep`. The last square end up on the child bone (vert 10 to 14).

This how I interpret the code.

I moved down the call to `convertToFlatShadedMesh` since it modifies the mesh and makes it almost impossible to select the vertices that end up on the child bone. This is also a tiny optimization I think, since less weights are needed if we add them before converting to flat shading. `convertToFlatShadedMesh` then automatically updates all vertex indices.

reimund: Perhaps a code comment is warranted.

deltakosh: Yes just a code comment to avoid future questions :D

deltakosh:
Something is wrong with the tests :)

PR #9017, opened 2020-09-22
Force worldmatrix to be updated before baking vertices

CedricGuillemet:
Followup on https://forum.babylonjs.com/t/navmesh-parenting/14329/3

World matrix was not properly updated when creating a navmesh. Forcing its update before baking vertices.


Popov72:
LGTM :)

Popov72 (approved):
LGTM!

PR #9018, opened 2020-09-22
Fix clearcoat F0 that should not be remapped for gltf files

Popov72:
See https://github.com/KhronosGroup/glTF-Sample-Models/pull/268#issuecomment-696448391


sebavan (approved):

src/Materials/PBR/pbrClearCoatConfiguration.ts
```diff
@@ -21,6 +21,7 @@ export interface IMaterialClearCoatDefines {
     CLEARCOAT_TEXTUREDIRECTUV: number;
     CLEARCOAT_BUMP: boolean;
     CLEARCOAT_BUMPDIRECTUV: number;
+    CLEARCOAT_REMAPP_F0: boolean;
```
bghgary, line 24:
```suggestion
    CLEARCOAT_REMAP_F0: boolean;

```

deltakosh: I'll do the change

PR #9021, opened 2020-09-22
Fix glTF serializer, Add color space conversion for mesh vertex colors with StandardMaterial applied.

Drigax:
This PR addresses:
https://forum.babylonjs.com/t/option-to-include-textures-in-babylon-files/14070/15

Where a glTF-roundtripped asset has preceivably incorrect vertex color. The fix involves changing the colorspace of the exported mesh from Babylon.js's StandardMaterial sRGB vertex color space to glTF's expected linear color space.


deltakosh (approved):
I like it!

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -728,9 +729,28 @@ export class _Exporter {
...
                 for (let k = 0, length = meshAttributeArray.length / stride; k < length; ++k) {
                     index = k * stride;
-                    const vertexData = stride === 3 ? Vector3.FromArray(meshAttributeArray, index) : Vector4.FromArray(meshAttributeArray, index);
+                    let vertexData;
+                    if (stride === 3) {
+                        vertexData = Vector3.FromArray(meshAttributeArray, index);
+                        if (convertToLinear) {
+                            Color3.FromArrayToRef(meshAttributeArray, index, tmpColor3);
+                            tmpColor3.toLinearSpace();
+                            vertexData.copyFromFloats(tmpColor3.r, tmpColor3.g, tmpColor3.b);
```
Popov72, line 744: Instead of the 3 lines above, I think you could simply do:
```typescript
vertexData.toLinearSpaceToRef(vertexData);
```
and get rid of the temporary colors.

Drigax: We don't always want to convert to linear space, if the input color is already expressed in linear, we'll convert again, and get the wrong color.

Drigax: Unfortunately we can't cast Vector3 to Color3...

Drigax: Unfortunately we can't cast Vector3 to Color3 with how we currently define the types.

Popov72: Can't you do:
```
vertexData = Color3.FromArray(meshAttributeArray, index);
```
instead of:
```
vertexData = Vector3.FromArray(meshAttributeArray, index);
```
to only work with colors?

Drigax: Thanks, I see what you mean now, I was able to avoid having to do an additional copy/conversion by following your suggestion.

deltakosh:
tests are failing

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -2,7 +2,7 @@ import { AccessorType, IBufferView, IAccessor, INode, IScene, IMesh, IMaterial,
 
 import { FloatArray, Nullable, IndicesArray } from "babylonjs/types";
 import { Vector2, Vector3, Vector4, Quaternion, Matrix } from "babylonjs/Maths/math.vector";
-import { Color3 } from "babylonjs/Maths/math.color";
+import { Color3, Color4, TmpColors } from "babylonjs/Maths/math.color";
```
aWeirdo, line 5: TmpColors no longer used 👍

serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -226,7 +226,7 @@ export class _GLTFMaterialExporter {
             return _cubicBezierCurve(t, P0.y, P1.y, P2.y, P3.y);
         }
 
-        let diffuse = babylonStandardMaterial.diffuseColor.toLinearSpace().scale(0.5);
```
bghgary, line 229: What's the reason for this change?

Drigax: Currently we're halving the base color of the material, when our specular->roughness conversion does not use it at all. 

It feels like an "eyeball" factor, and in our current case  leads to poor conversion between simple materials. We removed this from the Autodesk exporters a while back, and I think we should do the same here.

bghgary: I agree it's an eyeball factor, but there is no correct conversion. The fact that Autodesk exporters do it differently is not necessarily a good argument. I don't think their shader model is the same as ours. If I recall correctly, there were a number of models that we converted that looked much too bright without this factor.

aWeirdo: Currently, while being halfed, most results i've got are too dark. (to the extend where it's highly visible) , i think i tested around  .66 - .75 with reasonably results

deltakosh:
Waiting for the reviewers to confirm before merging

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Sounds like this problem isn't as trivial as I was expecting. Lets keep the focus of this PR to just the vertex color changes, and we can revisit the material base color conversion in another PR after I do some more investigation. Ideally something similar to https://kcoley.github.io/glTF/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/examples/convert-between-workflows-bjs/ using our current conversion math should allow us to get a better idea of what we're seeing.

bghgary: Is that link for conversion between standard and PBR or the original spec/gloss vs metal/rough? I thought we had already created a site that did standard to PBR somewhere.

Drigax: That one was the old spec/gloss to metal rough, I'm not sure where we hosted a standard to PBR conversion sample...

Drigax:
We should be good to merge now, rolled back the controversial changes. Checking in...

serializers/src/glTF/2.0/glTFMaterialExporter.ts
bghgary: Hmm, I'm pretty sure we added one or there would be no reason to do so since I'm already hosting it here: https://bghgary.github.io/glTF/convert-between-workflows-bjs/ which is the original one. I think it's this one: https://github.com/kcoley/glTF/tree/kcoley/standardMatToSpecGloss

PR #9023, opened 2020-09-23
Native constants

syntheticmagus:
PR to change the way constant values are communicated between Babylon.js and Babylon Native.

This PR contains changes required by https://github.com/BabylonJS/BabylonNative/pull/416.


src/Engines/nativeEngine.ts
```diff
@@ -1551,4 +1501,95 @@ export class NativeEngine extends Engine {
...
+            case Constants.TEXTURE_LINEAR_LINEAR:
+                return this._native.SAMPLER_TRILINEAR;
+            case Constants.TEXTURE_LINEAR_NEAREST:
+                return this._native.SAMPLER_MINPOINT_MAGLINEAR_MIPLINEAR;
+            case Constants.TEXTURE_NEAREST_NEAREST_MIPLINEAR:
+                return this._native.SAMPLER_MINPOINT_MAGPOINT_MIPLINEAR;
+            case Constants.TEXTURE_LINEAR_LINEAR_MIPNEAREST:
+                return this._native.SAMPLER_TRILINEAR;
+            default:
+                throw new Error("Unexpected sampling mode: " + samplingMode + ".");
```
bghgary, line 1538: nit: Looks like you are following existing pattern, but adding a period seems a bit overkill to me :)

src/Engines/nativeEngine.ts
```diff
@@ -1551,4 +1501,95 @@ export class NativeEngine extends Engine {
...
+            case Constants.TEXTURE_LINEAR_NEAREST_MIPLINEAR:
+                return this._native.SAMPLER_MINPOINT_MAGLINEAR_MIPLINEAR;
+            case Constants.TEXTURE_LINEAR_LINEAR:
+                return this._native.SAMPLER_TRILINEAR;
+            case Constants.TEXTURE_LINEAR_NEAREST:
+                return this._native.SAMPLER_MINPOINT_MAGLINEAR_MIPLINEAR;
+            case Constants.TEXTURE_NEAREST_NEAREST_MIPLINEAR:
+                return this._native.SAMPLER_MINPOINT_MAGPOINT_MIPLINEAR;
+            case Constants.TEXTURE_LINEAR_LINEAR_MIPNEAREST:
+                return this._native.SAMPLER_TRILINEAR;
```
bghgary, line 1536: I didn't check all of them, but at least this one seems to be a non-exact mapping. Are there more? Maybe a comment is warranted?

src/Engines/nativeEngine.ts
```diff
@@ -1551,4 +1501,95 @@ export class NativeEngine extends Engine {
...
+
+    private _getNativeTextureFormat(format: number, type: number): number {
+        if (format == Constants.TEXTUREFORMAT_RGBA && type == Constants.TEXTURETYPE_UNSIGNED_INT) {
+            return this._native.TEXTURE_FORMAT_RGBA8;
+        }
+        else if (format == Constants.TEXTUREFORMAT_RGBA && type == Constants.TEXTURETYPE_FLOAT) {
+            return this._native.TEXTURE_FORMAT_RGBA32F;
+        }
+        else {
+            throw new Error("Unexpected texture format or type: format " + format + ", type " + type + ".");
```
bghgary, line 1550: Is this because we don't support them yet? If so, maybe "Unsupported" instead of "Unexpected"

src/Engines/nativeEngine.ts
```diff
@@ -1551,4 +1501,95 @@ export class NativeEngine extends Engine {
...
+    private _getNativeAttribType(type: number): number {
+        switch (type) {
+            case VertexBuffer.UNSIGNED_BYTE:
+                return this._native.ATTRIB_TYPE_UINT8;
+            case VertexBuffer.SHORT:
+                return this._native.ATTRIB_TYPE_INT16;
+            case VertexBuffer.FLOAT:
+                return this._native.ATTRIB_TYPE_FLOAT;
+            default:
+                throw new Error("Unexpected attribute type: " + type + ".");
```
bghgary, line 1592: ditto

src/Engines/nativeEngine.ts
rgerd: Just out of curiosity, if we have something like this would it be preferable to use string interpolation like `Unexpected sampling mode: ${samplingMode}.`? 

Feels like it would make the "Unexpected texture format..." line more manageable as well.

bghgary: Yup, string interpolation would be good. It ends up being the same once transpiled to JavaScript with the settings we use for Babylon.js for compatibility reasons.

src/Engines/nativeEngine.ts
syntheticmagus: Tracked in https://github.com/BabylonJS/BabylonNative/issues/417.

bghgary (approved):

PR #9029, opened 2020-09-23
Change use of mousemove to pointermove for Fly and Free camera mouse inputs

PolygonalSun:
This PR will change the events used for mouse movements from mousemove to pointermove.  This change came about because of an issue with Firefox and pointerlock not producing mousemove events.


deltakosh (changes requested):
We need to move everything inside the pointer observer

deltakosh (approved):
Thanks!

PR #9033, opened 2020-09-25
Native texture sampling constants rename

syntheticmagus:
Renamed and re-sorted native texture sampling constants to match changes to Babylon Native.

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -1531,14 +1531,6 @@ export class NativeEngine extends Engine {
                 return this._native.TEXTURE_LINEAR_LINEAR_MIPNEAREST;
             case Constants.TEXTURE_LINEAR_NEAREST:
                 return this._native.TEXTURE_LINEAR_NEAREST;
-
-            // Legacy constants
-            case Constants.TEXTURE_NEAREST_SAMPLINGMODE:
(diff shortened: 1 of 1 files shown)


src/Engines/nativeEngine.ts
```diff
@@ -1506,34 +1507,38 @@ export class NativeEngine extends Engine {
...
+            case Constants.TEXTURE_LINEAR_NEAREST:
+                return this._native.TEXTURE_LINEAR_NEAREST;
+
+            // Legacy constants
+            case Constants.TEXTURE_NEAREST_SAMPLINGMODE:
+                return this._native.TEXTURE_NEAREST_NEAREST;
+            case Constants.TEXTURE_BILINEAR_SAMPLINGMODE:
+                return this._native.TEXTURE_LINEAR_LINEAR;
+            case Constants.TEXTURE_TRILINEAR_SAMPLINGMODE:
+                return this._native.TEXTURE_LINEAR_LINEAR_MIPLINEAR;
```
bghgary, line 1541: Is this necessary? They are the same values as the other constants?

syntheticmagus: Ha! No, good catch. I keep forgetting they're just numbers and thinking of them like enums.

bghgary (approved):

Issue #9028, opened 2020-09-23
Exporting and Importing custom frames does not show the output port


PirateJC:
When creating a custom frame, if the output port has the "expose to frame" setting as set to true, when you export and import that frame, the output port does is not visible on the collapsed frame.

Issue #9027, opened 2020-09-23
Exporting/Importing frames does not show all of the input ports.


PirateJC:
When you create a custom frame, export it and then reimport it into your project, the frame does not show all of the input ports that have the "expose to frame" option set to true.

PR #9034, opened 2020-09-25
Bug Fixes For Exposed Frame Ports

msDestiny14:
Changes include:

- Removing isExposed variable. This is now tracked through the port position only now. 
- Corrected mistake on port position when a link was deleted.
- Corrected mistake on port position 0 not being added.
- Edges cases where ports were being removed by mistake.

Testing: For this I ran several manual tests using large frames with many nodes, removing and adding nodes with links, and reexporting and importing multiple times. (More testing may be required to find further edge cases).

https://github.com/BabylonJS/Babylon.js/issues/9027
https://github.com/BabylonJS/Babylon.js/issues/9028

### DIFF
--- a/nodeEditor/src/diagram/graphFrame.ts
+++ b/nodeEditor/src/diagram/graphFrame.ts
@@ -140,51 +140,43 @@ export class GraphFrame {
             node.isVisible = false;
         }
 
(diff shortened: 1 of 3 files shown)


PirateJC (approved):
Looks good!!!!

PR #9039, opened 2020-09-26
Picking faceless mesh

CedricGuillemet:
Follow up on email discussion regarding 'picking a unique model (point cloud shape)'

In the mesh picking check that at least 1 submesh can be picked by ray-triangle otherwise falls back to bbox intersection.

### DIFF
--- a/src/Meshes/abstractMesh.ts
+++ b/src/Meshes/abstractMesh.ts
@@ -1545,7 +1545,10 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
             if (!material) {
                 continue;
             }
-            if (this.getIndices()?.length && (material.fillMode == Constants.MATERIAL_TriangleStripDrawMode)) {
+            if (this.getIndices()?.length && (material.fillMode == Constants.MATERIAL_TriangleStripDrawMode ||
+                    material.fillMode == Constants.MATERIAL_TriangleFillMode ||
+                    material.fillMode == Constants.MATERIAL_WireFrameFillMode ||
+                    material.fillMode == Constants.MATERIAL_PointFillMode )) {
                 anySubmeshSupportInterect = true;
                 break;
             }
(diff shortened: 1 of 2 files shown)


src/Meshes/abstractMesh.ts
```diff
@@ -1535,6 +1535,32 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
+                break;
+            }
+        }
+
+        // no sub mesh support intersection, fallback to BBox that has already be done
+        if (!anySubmeshSupportInterect) {
+            pickingInfo.hit = true;
+            pickingInfo.pickedMesh = this;
+            pickingInfo.distance = Vector3.Distance(ray.origin, boundingInfo.boundingSphere.center);
+            pickingInfo.subMeshId = 0;
```
Popov72, line 1562: Maybe setting `subMeshId` to -1 would be better as there's really no submesh intersecting the raw?

src/Meshes/abstractMesh.ts
```diff
@@ -1535,6 +1535,32 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
 
         var subMeshes = this._scene.getIntersectingSubMeshCandidates(this, ray);
         var len: number = subMeshes.length;
+
+        // Check if all submeshes are using a material that don't allow picking (point/lines rendering)
+        // if no submesh can be picked that way, then fallback to BBox picking
+        var anySubmeshSupportInterect = false;
```
Popov72, line 1541: nit: typo `anySubmeshSupportIntersect`

sebavan:
Love it, just wonder if there are any ways to prevent the double loop with caching or by doing some checks as part of the former one as I am afraid of the perf impact ???

src/Meshes/abstractMesh.ts
CedricGuillemet: Good catch!

CedricGuillemet:
@sebavan unfortunately, I have to loop on submeshes :( Is `getMaterial()` from submesh that costly?

deltakosh:
I think it is ok as the loop will break  immediatly most of the time

PR #9040, opened 2020-09-26
Add muted setting to VideoTexture, fix autoplay in Chrome

simonihmig:
Auto-playing a VideoTexture in Chrome failed to work. According to their [autoplay policy](https://developers.google.com/web/updates/2017/09/autoplay-policy-changes) muted autoplay is allowed, so this adds support for a `muted` setting.

But this still seems to be not enough for Chrome, as a video element detached from DOM seems still to not autoplay. See this reproduction: https://jsfiddle.net/6105nLzr/4/ - when line 13 is commented out, the video does *not* start to play (see console output). So calling `play()` programmatically which fixed this for me.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -164,6 +164,7 @@
 - Added the possibility to update the shader code before being compiled ([Popov72](https://github.com/Popov72))
 - Added the `shadowOnly` property to the `BackgroundMaterial` class ([Popov72](https://github.com/Popov72))
 - Added support for lightmaps in unlit PBR materials ([Popov72](https://github.com/Popov72))
+- Added `muted` setting to `VideoTexture`, fix autoplay in Chrome ([simonihmig](https://github.com/simonihmig))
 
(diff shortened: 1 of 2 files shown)


RaananW:
Just for me to understand - auto-play will only work if muted and play is called, right?

deltakosh:
yes this is correct. Thanks for the fix

simonihmig:
> Just for me to understand - auto-play will only work if muted and play is called, right?

Not necessarily *only* if muted, but that's one sufficient criteria to be eligible for autoplaying. See this from the link above:

> Muted autoplay is always allowed.
> Autoplay with sound is allowed if:
> User has interacted with the domain (click, tap, etc.).
> On desktop, the user's Media Engagement Index threshold has been crossed, meaning the user has previously played video with sound.
> The user has added the site to their home screen on mobile or installed the PWA on desktop.
> Top frames can delegate autoplay permission to their iframes to allow autoplay with sound.

PR #9041, opened 2020-09-28
Vec3 * Quat

Pryme8:
I don't think we have a method for this yet.

If we already do this can be skipped if not then here we go.

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -1091,6 +1091,60 @@ export class Vector3 {
     }
 
     /**
+     * Returns this Vector3 after being multiplied by a quaternion.
+     * @param quat the quaternion to multiply the vector by
+     * @returns this Vector3
+     */
+    public multiplyByQuat(quat: Quaternion): Vector3 {
+        let num = quat.x * 2,
+        num2 = quat.y * 2,
+        num3 = quat.z * 2,
+        num4 = quat.x * num,
+        num5 = quat.y * num2,
+        num6 = quat.z * num3,
+        num7 = quat.x * num2,
+        num8 = quat.x * num3,
+        num9 = quat.y * num3,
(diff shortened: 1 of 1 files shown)


src/Maths/math.vector.ts
```diff
@@ -1090,6 +1090,60 @@ export class Vector3 {
...
+        num10 = quat.w * num,
+        num11 = quat.w * num2,
+        num12 = quat.w * num3;
+
+        let x = (((1 - (num5 + num6)) * this.x) + ((num7 - num12) * this.y)) + ((num8 + num11) * this.z),
+        y = (((num7 + num12) * this.x) + ((1 - (num4 + num6)) * this.y)) + ((num9 - num10) * this.z),
+        z = (((num8 - num11) * this.x) + ((num9 + num10) * this.y)) + ((1 - (num4 + num5)) * this.z);
+
+        this.set(x, y, z);
+
```
Hersir88, line 1117: this multiplyByQuatToRef(quat, this), no need to duplicate code

Hersir88: @deltakosh would know better but would assume that this should copy vector and return that, to follow other methods in the, file, as there are 3 types of them standard returns fresh copy, `*inPlace` modifies current vector and `*ToRef` passed one

RaananW:
Is this the same as `Vector3.rotateByQuaternionToRef` ?

deltakosh:
I agree this is duplicate code

Pryme8:
> Is this the same as `Vector3.rotateByQuaternionToRef` ?

Now that I look at it yes it is.  My bad, for some reason last night I thought they were different then this morning I was gonna close it anyways because I kind of realized that as well.

*sigh* the things we think of when we are tired and just getting stuff done.

Thanks for taking a look though!

PR #9050, opened 2020-09-29
Texture: Add a property to avoid deformations when rotating the texture and us…

Popov72:
…ing non-uniform scaling.

See https://forum.babylonjs.com/t/texture-rotation-not-working-for-non-square-images/14482

PG: https://www.babylonjs-playground.com/#T32V41#22

### DIFF
--- a/src/Materials/Textures/texture.ts
+++ b/src/Materials/Textures/texture.ts
@@ -1,7 +1,7 @@
 import { serialize, SerializationHelper } from "../../Misc/decorators";
 import { Observable } from "../../Misc/observable";
 import { Nullable } from "../../types";
-import { Matrix, Vector3 } from "../../Maths/math.vector";
+import { Matrix, TmpVectors, Vector3 } from "../../Maths/math.vector";
 import { BaseTexture } from "../../Materials/Textures/baseTexture";
 import { Constants } from "../../Engines/constants";
 import { _TypeStore } from '../../Misc/typeStore';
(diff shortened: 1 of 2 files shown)


src/Materials/Textures/texture.ts
```diff
@@ -456,7 +466,11 @@ export class Texture extends BaseTexture {
             this.vScale === this._cachedVScale &&
             this.uAng === this._cachedUAng &&
             this.vAng === this._cachedVAng &&
-            this.wAng === this._cachedWAng) {
+            this.wAng === this._cachedWAng &&
+            this.uRotationCenter === this._cachedURotationCenter &&
+            this.vRotationCenter === this._cachedVRotationCenter &&
+            this.wRotationCenter === this._cachedWRotationCenter &&
+            this.homogeneousRotationInUVTransform === this._cachedHomogeneousRotationInUVTransform) {
```
deltakosh, line 473: I would vote for testing the rotationCenter only if homogeanous is true (that could save 3 tests for each texture matrix per frame)

Popov72: Those variables are also used in the "legacy" case and to me it appears it was a bug in the previous code: updating any of these values would not update the matrix whereas it should do.

Popov72:
Regarding the build failing, it fails in the "npm install" step, I don't know what it means...

deltakosh:
That makes sense!
The build crashing is weird..

wtesler:
Hi, this is hellowill89 from the forum thread. 

Very nice work @Popov72 !

I roughly added your changes in this playground and confirmed the changes do produce the desired result:

https://www.babylonjs-playground.com/#T32V41#23

Popov72:
It is showing some errors with linting and "build and unit tests" but when clicking to see the detail there's no error...

deltakosh:
Ok let me merge :)
I'll run a nightly

wtesler:
Thanks so much everyone. Really appreciated the quick turnaround.

deltakosh:
always :)

Issue #8972, opened 2020-09-15
Add noMipmap option to photoDome config


Exolun:
My team is interested in using mipMaps on certain photoDomes.  Any chance we could get a `noMipmap` flag added to the config object to enable initializing with mipmaps?

Thanks!

deltakosh:
Sure thing!

RaananW:
generateMipMaps flag added to options

PR #9051, opened 2020-09-29
No mip map option added to photo dome

RaananW:
Fixing #8972 

Was not added to the video texture, as video texture mipmaps can be very expensive.

### DIFF
--- a/src/Helpers/photoDome.ts
+++ b/src/Helpers/photoDome.ts
@@ -55,7 +55,7 @@ export class PhotoDome extends TextureDome<Texture> {
     }
 
     protected _initTexture(urlsOrElement: string, scene: Scene, options: any): Texture {
-        return new Texture(urlsOrElement, scene, true, !this._useDirectMapping, undefined, undefined, (message, exception) => {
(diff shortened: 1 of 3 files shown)


src/Helpers/photoDome.ts
```diff
@@ -55,7 +55,7 @@ export class PhotoDome extends TextureDome<Texture> {
     }
 
     protected _initTexture(urlsOrElement: string, scene: Scene, options: any): Texture {
-        return new Texture(urlsOrElement, scene, true, !this._useDirectMapping, undefined, undefined, (message, exception) => {
+        return new Texture(urlsOrElement, scene, options.noMipMap, !this._useDirectMapping, undefined, undefined, (message, exception) => {
```
deltakosh, line 58: if it was true before we should set it to true by default right?

RaananW: yes! will do.

RaananW: Converted to generateMipMaps, and with that kept the default behavior for both video and photo dome

PR #9055, opened 2020-09-29
Fixed TriPlanarMaterial and WaterMaterial

julien-moreau:
Related to https://forum.babylonjs.com/t/problems-getting-triplanarmaterial-working-without-bump-texture-see-playground-link/14441
And https://forum.babylonjs.com/t/water-material-with-smaller-size-water-mesh/13843/14

### DIFF
--- a/materialsLibrary/src/water/waterMaterial.ts
+++ b/materialsLibrary/src/water/waterMaterial.ts
@@ -104,27 +104,27 @@ export class WaterMaterial extends PushMaterial {
     public maxSimultaneousLights: number;
 
     /**
-    * @param {number}: Represents the wind force
-    */
+     * Defines the wind force.
+     */
     @serialize()
     public windForce: number = 6;
     /**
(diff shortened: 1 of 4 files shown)


materialsLibrary/src/water/waterMaterial.ts
```diff
@@ -178,6 +178,12 @@ export class WaterMaterial extends PushMaterial {
     */
     @serialize()
     public waveSpeed: number = 1.0;
+
+    /**
+     * @param {number}: Defines the number of times waves are repeated. This is typically used to adjust waves count according to the ground's size where the material is applied on.
```
deltakosh, line 183: you don't need the type or even the param keyword for a field

julien-moreau: Ahah yes I have hesitated, I just followed the way it was commented before. Updating comments right now :)

deltakosh (approved):

PR #9056, opened 2020-09-29
Mouse wheel control for FreeCamera and derivatives.

mrdunk:
Considering the UniversalCamera is the default camera now i was surprised by the lack of functionality when it comes to controlling it.
I loosely based this new implementation on `followCameraMouseWheelInput.ts` which i wrote ~2 years ago but have tried to address some shortcomings in that implementation:
1) Better scaling between browser implementations; Most notably Firefox which unlike most other browsers uses text lines as a scroll unit, not pixels like nearly all the other browsers..
2) Make use of multiple scroll wheels; Most laptop touch pads have 2 axis: X&Y although the spec allows 3. I'm not sure if there are actually any out there that support all 3....

It is split into a Base class `BaseCameraMouseWheelInput.ts` that deals with events
and a FreeCamera implementation `freeCameraMouseWheelInput.ts` which applies mouse wheel axis to camera properties.
I think there would be merit in porting other camera types to use `BaseCameraMouseWheelInput.ts` but that is outside the scope of this submission.

Some sensible defaults are applied as standard:
Move camera forward/back is controlled by the mouse wheel Y axis.
Pan left/right is controlled by the mouse wheel X axis.
Building the changes and running a default playground will let the user see these in action with no additional code/configuration required.

I have manually tested `freeCameraMouseWheelInput` on recent versions on the following browsers:
Chromium (Version 85.0.4183.83 (Official Build) Built on Ubuntu , running on Ubuntu 18.04 (64-bit))
FireFox (81.0 (64-bit) , running on Ubuntu 18.04 (64-bit))

It would be good to manually test on some older browsers and anything on Mac or Windows before merging this.
The browsers listed above are all i have access to.

It currently passes the automated tests. (npm run build)
I haven't really given much thought to unit tests for this yet.
I remember writing some for `BaseCameraPointersInput.ts` a while back. I will look at those again for inspiration if what i have so far meets with approval.

I'm not convinced i provide a good API for configuring the FreeCamera implementation. Specifically, choosing which camera property and axis to affect with each scroll wheel axis.
Opinions (or better, prior art examples in the codebase) welcome.

There are still some small TODOs still in the code but thought i'd get this in front of potential reviewers before i do too much polishing.

Also there is almost definitely some work to do on scaling mouse wheels across different browser / hardware / OS versions so a developer using this code can be confident a mouse wheel will have a stable sensitivity across different platforms.
This would be outside the scope of what i can do, having limited platforms available. I suspect we could leverage forum help with this? Publish a survey on the forum where users run a playground that logs mouse wheel ticks and have users drag the wheel a set distance.

Anyway,
i'm reasonably confident the code in this pull request is better than nothing.
Please let me know what you think; I have thick skin so don't worry about sugar coating any review comments.

dunk.


deltakosh:
it is definitely a good idea.
Some questions though:
- Can we factor what is in followCameraMouseWheelInput.ts ?
- I'm wondering if adding it by default could be seen as a breaking change ?

Pinging @PolygonalSun as he is the gatekeeper for everything input related

mrdunk:
It could definitely be disabled by default if there are concerns.
Simply removing the `addMouseWheel()` from `src/Cameras/freeCamera.ts` would make this "opt in".
It may be wise to do this anyway for the first release while we have some forum users test it.

As for a breaking change,
that depends on your projects definition of "breaking".

It is entirely new functionality for the FreeCamera and derivatives.
In my opinion, this is a feature addition; Something that i was surprised did not already exist and work by default.

However i can also see the argument that adding a new feature may be unexpected and unwelcome to some projects importing Babylon.
Even if this is the winning argument then i think making this feature available but "opt in" is still perfectly acceptable.

deltakosh:
I feel like it is not a big deal to turn it on but later. So yeah let's stick with off by default for now

mrdunk:
Off by default then.
Done.

So to enable, you just need to
`camera.inputs.addMouseWheel();`.
That will automatically give you the default options.

Now i think about it, another way this could be disabled is per mouse wheel axis.
`camera.inputs.attached["mousewheel"].wheelYMoveRelative = null;`.
I think what i just submitted is cleaner while we evaluate this though.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,402 @@
...
+    /**
+     * Z axis.
+     */
+    Z
+}
+
+/**
+ * An interface for user definable callback to be called on mouse wheel movement.
+ */
+export interface IFreeCameraMouseWheelCustomCallback {
```
mrdunk, line 29: I just discovered Babylon's Observer pattern.
This section would be better implemented using that.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,402 @@
...
+        return this._wheelZActionProperty;
+    }
+
+    /**
+     * A user configurable callback to be called on mouse wheel movement.
+     * To be used whenever the default functionality of this class does not
+     * change the required camera parameter by default.
+     */
+    @serialize()
+    public customCallback: Nullable<IFreeCameraMouseWheelCustomCallback> = null;
```
mrdunk, line 307: Will refactor to use the Observer pattern instead.
Also, no reason the Observable couldn't be in the base class so any future implementations using the Base get it.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,373 @@
...
+/**
+ * Defines the potential axis to be altered in a transform operation.
+ */
+export enum AXIS {
+    /**
+     * X axis.
+     */
+    X,
+    /**
+     * Y axis.
```
PolygonalSun, line 17: I feel like the naming of this may be a bit confusing as there is already an "Axis" class that has the same options (X, Y, Z) and returns a vector with corresponding values.

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,202 @@
...
+    public getClassName(): string {
+        return "BaseCameraMouseWheelInput";
+    }
+
+    /**
+     * Get the friendly name associated with the input class.
+     * @returns the input friendly name
+     */
+    public getSimpleName(): string {
+        return "mousewheel";
```
PolygonalSun, line 166: Just out of curiosity, what scenarios would this be used in?  Also, does this have any relation to the string used for mouse wheel events?

mrdunk: This is part of the input loading infrastructure.
If you look at the contents of `camera.inputs.attached` in any playground, this value is the key in that collection relating to the `ICameraInput` instance.
See https://doc.babylonjs.com/how_to/customizing_camera_inputs for more detail.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
mrdunk: Glad you brought this up.
There is already a `JOYSTICKAXIS`. See: https://doc.babylonjs.com/api/enums/babylon.joystickaxis
I considered creating a `MOUSEEWHEELAXIS` in a similar vein.... but was a bit concerned about namespace pollution.

I wondered if a generic `INPUTAXIS` might be more appropriate; It could then be reused for other uses. We'd probably want to put it in it's own file that we could include from here.

But such an enum would be potentially useful elsewhere (outside inputs) too.
But like you say, we'd then need to think of a better name for it.

mrdunk: Actually,
I was confusing this with some other code i wrote recently.
Looking back at this PR,
the axis in question here isn't the mouse-wheel axis but rather the axis of the property the mouse wheel will affect.

Similar arguments could be made about re-usability vs polluting the namespace though.
With Babylon's commitment to backwards compatibility, the number of names in use is only going to grow...

Whatever we use has to be `export`-ed as it's used to configure the feature.

Can we come up with something that is reusable?
Do you have an official style-guide? This is the closest i found:
https://doc.babylonjs.com/how_to/contribute_to_api#format-of-comments
It implies enum names should actually PascalCase.
So what about:
`AxisEnum` ?
`EAxis` ?
`ChooseAxis` ?
`SelectAxis` ?

I'm not loving any of those.
Am i overthinking this?

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,373 @@
...
+    private _wheelYActionProperty: Nullable<AXIS> = AXIS.Z;
+    private _wheelZAction: Nullable<Vector3> = null;
+    private _wheelZActionProperty: Nullable<AXIS> = null;
+
+    /**
+     * Called once per mouse wheel axis. Will update the camera according to any
+     * configured properties for that axis.
+     */
+    private _updateCamera(
+        value: number, action: Nullable<Vector3>, property: Nullable<number>): void {
```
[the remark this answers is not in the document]
mrdunk: Ack.
fixed.

deltakosh:
Please bear with us :) @PolygonalSun  is dealing with IRL issues but will get back to you asap :)

deltakosh:
Adding @RaananW  to help review as well

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
RaananW: Just wondering (I'll review this better tomorrow) why not use the Axis enum we already have? It's rather genetic and has x,y and z

mrdunk: There's a Class `Axis` here: `src/Maths/math.axis.ts`
and an Enum `JojstickAxis` here: `src/Misc/virtualJoystick.ts`.
I don't think i'm missing a plain Enum.

What about creating an Enum `Coordinate` in `src/Misc`?
It makes lexical sense and isn't too long.
`Cartesian` is another contender but it does not imply a single element but rather all 3 (X, Y and Z).

Naming is hard.

mrdunk:
Don't worry how long this takes; You folks are busy doing the day job. I'm just hacking around for fun. Whenever you get to it is fine.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
RaananW: Sorry, you are totally right.

The best would be to have an Axis enum, but... we already have an axis object. Naming... Well, Coordinate enum sounds good :-)

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,202 @@
...
+                this._wheelDeltaZ +=
+                    this.wheelPrecisionZ * platformScale *
+                    (<any>event).wheelDeltaZ / this._normalize;
+            } else if ((<any>event).wheelDelta) {
+                // IE >= v9   (Has WebGL >= v11)
+                // Maybe others?
+                this._wheelDeltaY -=
+                    this.wheelPrecisionY * (<any>event).wheelDelta / this._normalize;
+            } else if (event.detail) {
+                // Firefox < v17  (Has WebGL >= v4)
```
RaananW, line 96: Firefox 17 was released on november 20th, 2012. It has long been updated :-) I don't think we need to support it. 

IE11 is a different story, as this browser doesn't get any more updates.

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,202 @@
...
+            this._wheel,
+            PointerEventTypes.POINTERWHEEL);
+    }
+
+    /**
+     * Detach the current controls from the specified dom element.
+     * @param element Defines the element to stop listening the inputs from
+     */
+    public detachControl(element: Nullable<HTMLElement>): void {
+        if (this._observer && element) {
```
RaananW, line 127: why are you checking for element here?

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Attach the input controls to a specific dom element to get the input from.
+     * @param element Defines the element the controls should be listened from
+     * @param noPreventDefault Defines whether event caught by the controls
+     *   should call preventdefault().
+     *   (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
+     */
+    public attachControl(element: HTMLElement, noPreventDefault?: boolean): void {
+        this._wheel = (pointer, _) => {
```
RaananW, line 57: `, _` can be removed, i guess?

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,373 @@
...
+     * Called for each rendered frame.
+     */
+    public checkInputs(): void {
+        if (this._wheelDeltaX === 0 &&
+                this._wheelDeltaY === 0 &&
+                this._wheelDeltaZ == 0) {
+            return;
+        }
+
+        this._moveRelative.setAll(0);
```
RaananW, line 298: All in all I find the way you deal with the vector references a bit confusing (referencing the vectors that will be updated).

Those 3 lines are hard to get until you read the entire code. because you are resetting them, and then, technically, never set their values again. 

Maybe setting some form of an ActionType enum that will be checked during updateCamera is enough?

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
```diff
@@ -0,0 +1,373 @@
...
+        this._rotateRelative.setAll(0);
+        this._moveScene.setAll(0);
+
+        this._updateCamera(this._wheelDeltaX, this._wheelXAction, this._wheelXActionProperty);
```
RaananW, line 302: could be a single line, updating all Axes. But that's just nitpicking :-)

RaananW (commented):
I think this is great! Mouse wheel in all directions would be a nice addition.

As I wrote in one of the comments, I think some parts are a bit confusing to read. But other than that - LGTM!

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
mrdunk: huh. I presumed you'd need all the parameters from the prototype but i guess not; Seems to compile error free and work in playground without.
It surprises me TypeScript doesn't complain about this when we come to add this to the observable.

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
mrdunk: OK, lets loose the last `else if`.
I agree it's unlikely to catch anything useful.

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
mrdunk: no idea. looks like a fossil.
Will remove.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
mrdunk: Sold.
I've put it in `src/Maths/math.axis.ts` as it seems to fit there.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
mrdunk: Yes,
I originally had things the other way around; Each vector axis (`_moveRelative`, `_rotateRalative`, `_moveScene`) referenced the touch-wheel axis that should control it.
While this sounds more sane,
It actually lead to much more code: You could easily map a single touch-wheel axis to multiple vector axis. Checks to insure that an error was raised when that happened needed getters and setters for /every/ vector axis.

I think it's very important to get this part right though. If it needs more code, lets do that.
This is how the user interacts with this code so it has to be logical.
I think whatever API we pick should be applicable to other Inputs too. eg: Mouse/Touch input.
I rant about this part here:
https://forum.babylonjs.com/t/universal-camera-with-mouse-scroll-wheel-and-touch-pinch/14139/5

mrdunk: If we do decide to leave the API as have it, I think we can fix the ambiguity in these lines with inline comments explaining what's going on.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
mrdunk: sure.

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
mrdunk: > Maybe setting some form of an ActionType enum that will be checked during updateCamera is enough?

I actually missed your point here on the first read. (Was too busy going off on my API rant.)
Yes, that will make things easier to understand. I'll do that.
(I'd still like a unified API for all Inputs though....)

src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
RaananW: Function overloading works a bit differently in typescript :-)

src/Cameras/Inputs/freeCameraMouseWheelInput.ts
RaananW: Hey, civilized rants are the main reason things change. :-)

I have no issue regarding the reason you are adding those three "spaces" which the user can control with the input. it was just the matter of not seeing them update in code. First thing I would do is search for `_moveRelative.set` or `_moveRelative = ???` , but the only thing i could find (outside of setters for other properties) was the `setAll(0)`. Not an issue if you look further into the code, but we are all lazy developers...

I totally get what you are saying. I have the forum page open and will answer when I get the time between PRs and refactors. 4.2 is coming up and we are all very busy trying to make the best out of this release.

Thanks for the changes! I will merge this very soon.

RaananW (approved):
LGTM.

thanks!!

RaananW:
Oh, we'll need doc update on this one, whenever you can. Quick note - We are currently working on updating our doc site, so any changes you will make will be manually merged by us in the background.

tdbe:
How do you change the properties of the mouse wheel that you add via camera.inputs.addMouseWheel() to a touch or universal camera? I just want to change the scrolling sensitivity. The only thing that comes up on the internet is the arcRotateCamera's "wheelPrecision" parameter. This parameter does not exist anywhere else. I hate these inconsistencies with babylon. It should be immediately self-obvious what the properties of a control are and how to access them.

sebavan:
Please ask the question on the forum forum.babylonjs.com as all the community is there to help :-)

And this is how you could do it: https://playground.babylonjs.com/#MZQQFX

tdbe:
> Please ask the question on the forum forum.babylonjs.com as all the community is there to help :-)
> 
> And this is how you could do it: https://playground.babylonjs.com/#MZQQFX

I must have spent 5 hours on this.

- wheelPrecision is only only on the arc camera
- camera.inputs.attached.mousewheel does NOT have any wheel precision properties on neither Touch nor Universal cameras.
- `Property 'wheelPrecisionX' does not exist on type 'ICameraInput<FreeCamera>'`
So no. But digging through the code, this "works":
```
if(camera.inputs._mouseWheelInput){
    camera.inputs._mouseWheelInput.wheelPrecisionX = 0.01;
    camera.inputs._mouseWheelInput.wheelPrecisionY = 0.01;
     camera.inputs._mouseWheelInput.wheelPrecisionZ = 0.01;
}
```
However, scrolling in and out, thecamera was slowly going under the ground because it was under a pivot and god knows what the internal zooming babylon code uses as forward and how it's broken by its own inertia angle lerps.

So I did my own camera zooming, around my own calculated direction vector. Digging through PainterEvents and Wheel Event I figured this out and this gives me a positive or negative number and this works and that's really the end of the time I have learning abot the hidden mysteries of wheels in babylon :+1: 
Hope this helps someone navigate this space until you upgrade to a more transparent and consistent and data oriented component design like most game engines.

```
const pointerWheel = function (wheelEvt:WheelEvent){
      pointerWheelVal = (wheelEvt.deltaX + wheelEvt.deltaY + wheelEvt.deltaZ)/300;
      //alert(pointerWheelVal);
    };
    scene.current.onPointerObservable.add((d, s) => {
      pointerWheel(d.event);
    },PointerEventTypes.POINTERWHEEL)
```

PR #9060, opened 2020-09-30
Copy Button to the Gradient Node

msDestiny14:
copy


deltakosh:
I would have move the copy button on the left of the delete button
Besides that this is perfect and LGTM

deltakosh (approved):

msDestiny14:
> I would have move the copy button on the left of the delete button
> Besides that this is perfect and LGTM

oh yes! Because that way while you're sliding if you misclick you don't delete it. I will switch.

PR #9062, opened 2020-10-01
Disable gizmo rotation when in billboard mode

CedricGuillemet:
follow up on this thread https://forum.babylonjs.com/t/problem-with-gizmos-and-billboard-modes/14547

### DIFF
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -252,10 +252,12 @@ export class Gizmo implements IDisposable {
             } else {
                 this._attachedNode._worldMatrix.decompose(transform.scaling, this._tempQuaternion, transform.position);
             }
-            if (transform.rotationQuaternion) {
-                transform.rotationQuaternion.copyFrom(this._tempQuaternion);
-            } else {
-                transform.rotation = this._tempQuaternion.toEulerAngles();
+            if (!transform.billboardMode) {
+                if (transform.rotationQuaternion) {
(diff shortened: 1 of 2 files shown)


RaananW (approved):

Popov72:
Shouldn't we simply document that the rotation gizmo should not be used with billboard mode instead of blocking it by code? If one displays the rotation gizmo in this case, they may wonder why dragging the circles do nothing. Whereas if it does something (wrong), they may ask in the forum or read the doc that will explain it is expected.

RaananW:
> Shouldn't we simply document that the rotation gizmo should not be used with billboard mode instead of blocking it by code? If one displays the rotation gizmo in this case, they may wonder why dragging the circles do nothing. Whereas if it does something (wrong), they may ask in the forum or read the doc that will explain it is expected.

I agree with both - document it is not working in billboard mode, and limit that in code to stay defensive, as not so many developers read the docs in full. 

Will wait before merging for some more voices

CedricGuillemet:
I can log in console with trying to attach a billboarded transform to a rotationGizmo.

julien-moreau:
+1 for the log. But as @RaananW says, developers don't necessary read the full doc and they also don't necessary take care of the logs ahah

deltakosh:
Yes let's log it but I like the defensive stance as well in the code

deltakosh (approved):

RaananW:
Linting failed. Will merge right after it's fixed

PR #9077, opened 2020-10-02
Upgrade Gizmo Meshes

davesauce14:
As discussed in the forum below, here is my PR for Unity inspired gizmo meshes:

> Demo here: https://proud-pond-0c7430d0f.azurestaticapps.net/#/rotation
> Forum Entry: https://forum.babylonjs.com/t/unity-inspired-gizmo-meshes/14554

Increased usability of the following of the Babylon Gizmos.

- Position Gizmo
- Scaling Gizmo
- Rotation Gizmo

Features include:

- Increased mouse collision threshold
- Enabled Hover State
- Grey out other axis to emphasize current action
- Scale gizmo mesh moves on drag
- Universal scalar increases/decreases gizmo mesh of all other axis
- Rotation gizmo instantiates "rotation circle" re-enforcing rotation percentage for user

Screenshots:

Rotation
image

Scale
image

Position
image


davesauce14:
Hi all, I just wanted to provide some details / ideas I had along with my PR:

In order to achieve the responsive gizmo features you see in the demo I needed to maintain state between all 3 axis.
For this I have added a pointerObserver in the parent gizmos.

> ├── `PositionGizmo` _// Parent Gizmo_
> │   ├── xGizmo    _// Axis Gizmo_
> │   ├── yGizmo     _// Axis Gizmo_
> │   └── zGizmo     _// Axis Gizmo_
> ├── `RotationGizmo` _// Parent Gizmo_
> │   ├── xGizmo     _// Axis Gizmo_
> │   ├── yGizmo     _// Axis Gizmo_
> │   └── zGizmo     _// Axis Gizmo_
> ├── `ScalingGizmo` _// Parent Gizmo_
> │   ├── xGizmo     _// Axis Gizmo_
> │   ├── yGizmo     _// Axis Gizmo_
> │   ├── yGizmo     _// Axis Gizmo_
> │   └── uniform    _// Axis Gizmo_    ... Also Planar, etc...

Axis Gizmos add themselves to the `gizmoAxisCache` which is owned by the parent.
And when a user interacts with the gizmo (hover, mouseDown, mouseUp) the parent handles the logic of graying our other axis-mesh, etc.

As this is my first contribution with you guys, I will likely have a few iterations of comments, but there was a few I had in mind off the bat which I have added inline the source.

aWeirdo:
Awesome!
One thought on the scale gizmo, 
it would be cool if it followed the mouse better, 
(both scaling of the mesh and the gizmo itself)

Maybe it's just me nitpicking 😅

davesauce14:
Hi @aWeirdo, 

Thank you for finding that bug, I have resolved it in my latest commit!
You will not see the changes reflected in the demo above since that uses my custom classes before contributing to Babylon :)

For your idea of moving the scaling gizmo with the user's mouse movement, I agree that would be cool and that is how Unity's gizmo works as well, however that would require a little more work as we would have to change the underlying behavior of how the attached mesh scaling behavior works as-well. 

The scalar gizmo must move proportional to the scale being applied to the mesh, and we are scaling proportional to the original object size, in contrast to Unity which is scaling 1:1 with users mouse, which is why you see the difference in gizmo mesh behavior. (both gizmo mesh and object mesh need to be in sync, or it feels weird)

- Unity's behavior: Follows user's pointer, scaling is applied proportionate to drag percentage, allows for negative scaling.
- Babylon's behavior: Scaling is proportionate to the object being scaled, doesn't allow for negative scaling.

[Babylon's Scale Logic](https://github.com/BabylonJS/Babylon.js/blob/master/src/Gizmos/axisScaleGizmo.ts#L100)

I wouldn't want to change drag behavior as part of this PR, but we can take it up after if we come to an agreement

src/Gizmos/axisDragGizmo.ts
```diff
@@ -78,20 +87,32 @@ export class AxisDragGizmo extends Gizmo {
...
 
-        // Build mesh on root node
-        this._arrow = AxisDragGizmo._CreateArrow(gizmoLayer.utilityLayerScene, this._coloredMaterial, thickness);
+        // Build Mesh + Collider
+        const arrow = AxisDragGizmo._CreateArrow(gizmoLayer.utilityLayerScene, this._coloredMaterial, thickness);
+        const collider = AxisDragGizmo._CreateArrow(gizmoLayer.utilityLayerScene, this._coloredMaterial, thickness * 4, true);
+
+        // Add to Root Node
+        this._gizmoMesh = new Mesh("", gizmoLayer.utilityLayerScene);
+        [...arrow.getChildMeshes(), ...collider.getChildMeshes()].forEach((mesh) => {
```
RaananW, line 109: Wouldn't is be enough to add the root mesh of the arrow and collider instead of all of its children? It is the parent of all meshes created in CreateArrow

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -201,6 +269,100 @@ export class PlaneRotationGizmo extends Gizmo {
         }
     }
 
+    private angleBetween3DCoords(origin: Vector3, coord1: Vector3, coord2: Vector3): number {
```
RaananW, line 272: Don't we have anything that can cover this in our math lib? (`GetAngleBetweenVectors`?)

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -201,6 +269,100 @@ export class PlaneRotationGizmo extends Gizmo {
...
+        angle = Math.round(angle * 1000) / 1000;
+        angle = Angle.FromDegrees(angle).radians();
+
+        // Flip if its cross has negitive y orientation
+        if (cross.y < 0) { angle = -angle; }
+
+        return angle;
+    }
+
+    private setupRotationCircle(paths: any[], parentMesh: AbstractMesh): Mesh {
```
RaananW, line 300: I assume the any[] array is actually a Vector3[]?

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -201,6 +269,100 @@ export class PlaneRotationGizmo extends Gizmo {
...
+        mat.backFaceCulling = false;
+        const mesh = Mesh.CreateRibbon("ignore", paths, false, false, 0, this.gizmoLayer.utilityLayerScene, true);
+        mesh.material = mat;
+        mesh.material.alpha = .25;
+        mesh.rotation.x = Math.PI / 2;
+        parentMesh.addChild(mesh);
+        return mesh;
+    }
+
+    private updateRotationPath(pathArr: any[], newFill: number): void {
```
RaananW, line 330: Same as above - please try as much as possible not to use any

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -201,6 +269,100 @@ export class PlaneRotationGizmo extends Gizmo {
...
+        for (let p = -Math.PI / 2; p < Math.PI / 2 - 1.5; p += step / 2) {
+            const path = pathArr[tessellationCounter];
+            if (path) {
+                let radianCounter = 0;
+                for (let i = 0; i < this.circleConstants.pi2; i += step) {
+                    if (path[radianCounter]) {
+                        if (i < Math.abs(newFill)) {
+                            const eie = (newFill > 0) ? i : i * -1;
+                            const pea = (newFill > 0) ? p : p * -1;
+                            path[radianCounter].x = this.circleConstants.radius * Math.sin(eie) * Math.cos(pea);
```
RaananW, line 343: using the vector's `set` function will be a bit quicker, as `x`, `y`, and `z` are actually getters/setters and not class members

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -201,6 +269,100 @@ export class PlaneRotationGizmo extends Gizmo {
...
+                }
+            }
+
+            tessellationCounter ++;
+        }
+    }
+
+    private updateRotationCircle(mesh: Mesh, paths: any[], newFill: number, dragPlanePoint: Vector3): void {
+        this.updateRotationPath(paths, newFill);
+        mesh = Mesh.CreateRibbon("ribbon", paths, false, false, 0, this.gizmoLayer.utilityLayerScene, undefined, undefined, mesh);
```
RaananW, line 363: Will it be possible to update the mesh itself and not create a new one? The old mesh is passed here as a reference - I find changing the reference is very confusing for a developer looking at/using this function.
Also - if not possible - return the mesh, and make sure to dispose the old one.

src/Gizmos/scaleGizmo.ts
```diff
@@ -38,6 +41,15 @@ export class ScaleGizmo extends Gizmo {
     private _uniformScalingMesh: Mesh;
     private _octahedron: Mesh;
     private _sensitivity: number = 1;
+    private _coloredMaterial: StandardMaterial;
+    private _hoverMaterial: StandardMaterial;
+    private _disableMaterial: StandardMaterial;
+    private _observables: Nullable<Observer<PointerInfo>>[] = [];
```
RaananW, line 47: if you are already setting it, there is no need to set it Nullable

src/Gizmos/scaleGizmo.ts
```diff
@@ -124,6 +123,55 @@ export class ScaleGizmo extends Gizmo {
...
+        this._octahedron = PolyhedronBuilder.CreatePolyhedron("", { type: 1 }, uniformScaleGizmo.gizmoLayer.utilityLayerScene);
+        this._octahedron.scaling.scaleInPlace(0.007);
+        this._uniformScalingMesh.addChild(this._octahedron);
+        uniformScaleGizmo.setCustomMesh(this._uniformScalingMesh, true);
+        var light = this.gizmoLayer._getSharedGizmoLight();
+        light.includedOnlyMeshes = light.includedOnlyMeshes.concat(this._octahedron);
+
+        // Drag Event Listeners
+        uniformScaleGizmo.dragBehavior.onDragObservable.add((e) => {
+            document.dispatchEvent(new CustomEvent('universalGizmoDrag', {
```
RaananW, line 156: were these events already defined? or did you just add them? I am actually asking about the naming convention

src/Gizmos/scaleGizmo.ts
```diff
@@ -188,6 +236,81 @@ export class ScaleGizmo extends Gizmo {
...
+     * Subscribes to pointer up, down, and hover events. Used for responsive gizmos.
+     */
+    public subscribeToPointerObserver(): void {
+        const pointerObserver = this.gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
+            if (pointerInfo.pickInfo) {
+                // On Hover Logic
+                if (pointerInfo.type === PointerEventTypes.POINTERMOVE) {
+                    if (this.dragging) { return; }
+                    this.gizmoAxisCache.forEach((statusMap, parentMesh) => {
+                        const isHovered = pointerInfo.pickInfo && (parentMesh.getChildMeshes().indexOf((pointerInfo.pickInfo.pickedMesh as Mesh)) != -1);
```
RaananW, line 258: getChildMeshes will recreate the children array on each frame. Please make sure to cache this call to avoid constant garbage collection

RaananW, line 258: Also - no need to test for pickInfo again. if typescript gives you a hard time, use the `!` character after pickInfo (of course only in this case, as you are chekcing it before)

src/Gizmos/scaleGizmo.ts
```diff
@@ -188,6 +236,81 @@ export class ScaleGizmo extends Gizmo {
...
+
+                // On Mouse Down
+                if (pointerInfo.type === PointerEventTypes.POINTERDOWN) {
+                    // If user Clicked Gizmo
+                    if (this.gizmoAxisCache.has(pointerInfo.pickInfo.pickedMesh?.parent as Mesh)) {
+                        this.dragging = true;
+                        const statusMap = this.gizmoAxisCache.get(pointerInfo.pickInfo.pickedMesh?.parent as Mesh);
+                        statusMap!.active = true;
+                        this.gizmoAxisCache.forEach((statusMap, parentMesh) => {
+                            const isHovered = pointerInfo.pickInfo && (parentMesh.getChildMeshes().indexOf((pointerInfo.pickInfo.pickedMesh as Mesh)) != -1);
```
RaananW, line 279: same as above (`getChildMeshes`)

src/Gizmos/axisScaleGizmo.ts
```diff
@@ -65,26 +67,44 @@ export class AxisScaleGizmo extends Gizmo {
...
+            const newMeshPosition = arrowMesh.position.z + ((newScale - originalScale) / 4);
+            if (newMeshPosition >= 0) {
+                arrowMesh.position.z = newMeshPosition;
+                arrowTail.scaling.y = newScale;
+                arrowTail.position.z = arrowMesh.position.z / 2;
+            }
+        };
+
+        const resetGizmoMesh = () => {
+            arrowMesh.position = new Vector3(nodePosition.x, nodePosition.y, nodePosition.z);
```
RaananW, line 104: please use the setters here and try avoiding creating new objects

src/Gizmos/axisScaleGizmo.ts
```diff
@@ -139,34 +159,81 @@ export class AxisScaleGizmo extends Gizmo {
                 this._matrixChanged();
             }
         });
+        // On Drag Listener: to move gizmo mesh with user action
+        this.dragBehavior.onDragObservable.add((e) => increaseGizmoMesh(e.dragDistance));
+        this.dragBehavior.onDragEndObservable.add(resetGizmoMesh);
+
+        // Listeners for Universal Scalar
+        document.addEventListener('universalGizmoDrag', (e) => increaseGizmoMesh((e as any).detail));
```
RaananW, line 167: would it be possible to use observables intead of event registration here? especially since you are the one creating and triggering them? Reason is that with observables we are sure when they are executed (inside the render loop)

src/Gizmos/axisScaleGizmo.ts
```diff
@@ -139,34 +159,81 @@ export class AxisScaleGizmo extends Gizmo {
...
-            var material = this._isHovered ? this._hoverMaterial : this._coloredMaterial;
-            this._rootMesh.getChildMeshes().forEach((m) => {
-                m.material = material;
-                if ((<LinesMesh>m).color) {
-                    (<LinesMesh>m).color = material.diffuseColor;
-                }
-            });
+            if (!this._parent) {
+                var material = this._isHovered ? this._hoverMaterial : this._coloredMaterial;
+                this._rootMesh.getChildMeshes().forEach((m) => {
```
RaananW, line 187: if this happen often, try avoiding calling getChildMeshes and cache this call for future reference. Otherwise a new array is created each time this function is called

src/Gizmos/positionGizmo.ts
```diff
@@ -192,6 +201,81 @@ export class PositionGizmo extends Gizmo {
...
+        const pointerObserver = this.gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
+            if (pointerInfo.pickInfo) {
+                // On Hover Logic
+                if (pointerInfo.type === PointerEventTypes.POINTERMOVE) {
+                    if (this.dragging) { return; }
+                    this.gizmoAxisCache.forEach((statusMap, parentMesh) => {
+                        const isHovered = pointerInfo.pickInfo && (parentMesh.getChildMeshes().indexOf((pointerInfo.pickInfo.pickedMesh as Mesh)) != -1);
+                        const material = isHovered || statusMap.active ? statusMap.hoverMaterial : statusMap.material;
+                        parentMesh.getChildMeshes().forEach((m) => {
+                            if (m.name !== 'ignore') {
```
RaananW, line 226: is there a better way than comparing strings to check if this mesh needs to be ignored?
If you can't find any, please set this string to be a static member of one of the main classes, and reuse the variable, instead of having `"ignore"`  everywhere.

RaananW (changes requested):
Hey!

Thanks for this amazing PR. Huge!!

I have a few change requests, I hope I don't frustrate you too much. Apart from that, there are a few issues that keep on surface:

1) Reuse objects as much as you can, instead of recreating them. If a vector is already there, use its set function instead of creating a new one. Same goes to any other object. Meshes are updatable (when set this way), materials and colors as well.
2) getChildMeshes creates a new array every time it is called. This will be heavy on the garbage collector, if the function is being executed on each frame (and even if it isn't ). Wherever possible, cache this call to avoid this
3) Naming convention - Please do you best to fit your new functions to this - https://doc.babylonjs.com/how_to/approved_naming_conventions

src/Gizmos/positionGizmo.ts
CedricGuillemet: I agree with Raanan. Can you build up a list of ignored meshes. Or separate hierarchy of ignore mesh under 2 different parents.

davesauce14:
@RaananW not a problem at all, this is a great learning opportunity for me. Just give me some time, I need to fit this in between my other work.

davesauce14:
Hi, @RaananW @CedricGuillemet,

I have updated the PR with your feedback above and responded to additional comments. 
Please review when you get the chance!

src/Gizmos/planeRotationGizmo.ts
RaananW: Also - 

1) this function, called on each frame, is creating 5 Vectors on each call. This must be avoided. The Temp Vectors are great for such a task.
2) Naming convention

src/Gizmos/scaleGizmo.ts
RaananW: Unless I am missing some null call, this hasn't been addressed, no?

RaananW (commented):
Thanks for the update!!

Apart from small naming convention issues (would be great to address them), and the few open question(s) about the `angleBetween3DCoords`, I have no technical(!) issues with this PR.

I will let @CedricGuillemet comment about functionality.

BTW - all tests are failing. Please go through the test results to see what went wrong and how you can fix it.

Thanks again! You're awesome :-)

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -51,48 +62,76 @@ export class PlaneRotationGizmo extends Gizmo {
...
+
+                // Remove Rotation Circle from parent mesh before drag interaction
+                this._gizmoMesh.removeChild(this._rotationCircle);
+
+                lastDragPosition.copyFrom(e.dragPlanePoint);
+                dragPlanePoint = e.dragPlanePoint;
+                const origin = this._rotationCircle.getAbsolutePosition().clone();
+                const originalRotationVector = this._rotationCircle.getAbsolutePosition().clone().addInPlace(direction);
+                const dragStartVector = e.dragPlanePoint;
+                let angle = this.angleBetween3DCoords(origin, originalRotationVector, dragStartVector);
```
CedricGuillemet, line 120: this can be replaced with a call to `GetAngleBetweenVectors` with (originalRotationVector.subtract(origin), dragStartVector.subtract(origin), this._rotationCircle.up). also, originalRotationVector and dragStartVector are misleading. they are points, not vectors.

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -51,48 +62,76 @@ export class PlaneRotationGizmo extends Gizmo {
...
+
+                if (Vector3.Dot(this._rotationCircle.up, Vector3.Down()) > 0) {
```
CedricGuillemet, line 122: Also can be removed.

src/Gizmos/planeRotationGizmo.ts
```diff
@@ -118,11 +157,13 @@ export class PlaneRotationGizmo extends Gizmo {
                     localPlaneNormalTowardsCamera = Vector3.TransformCoordinates(planeNormalTowardsCamera, rotationMatrix);
                 }
                 // Flip up vector depending on which side the camera is on
+                let cameraFlipped = false;
```
CedricGuillemet, line 160: Why do you need this flip? Is it because of back face culling?

davesauce14:
Thank you @RaananW @CedricGuillemet for the prompt review.
I have addressed the items in your comments. 
Awaiting review.

CedricGuillemet:
Few minor things:

- when using a gizmomanager with position and rotation gizmos enabled, using rotation gizmo doesn't grey the position gizmo.
- with this PG : https://playground.babylonjs.com/#8GY6J8#48 the scale axis doesn't stay yellow when interacting and mouse pointer is not on the gizmo

deltakosh:
@RaananW @CedricGuillemet  can you gentlement do the last check?

davesauce14:
Thanks @deltakosh, 
Pending one more enhancement, will finish in ~1hr :)
Thanks for keeping an eye out, will respond when the deed is done!

davesauce14:
Hi guys, revisions are out, lets get it in!

CedricGuillemet:
I just did another pass of tests and it's all working for me. @RaananW can you approve this PR once you are ok ?

RaananW (approved):
I have no technical issues. If @CedricGuillemet is fine with the final result, I'm fine with it as well.

Great work 😊 Thanks a lot!

davesauce14 (commented):
Finished

davesauce14:
@CedricGuillemet @RaananW Thank you guys!

src/Gizmos/planeRotationGizmo.ts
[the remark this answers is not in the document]
RaananW: Is this still valid? I see the function is no longer in the code

src/Gizmos/rotationGizmo.ts
```diff
@@ -152,6 +160,82 @@ export class RotationGizmo extends Gizmo {
...
+      @param cache display gizmo axis thickness
+     */
+    public addToAxisCache(mesh: Mesh, cache: any) {
+        this.gizmoAxisCache.set(mesh, cache);
+    }
+
+    /**
+     * Subscribes to pointer up, down, and hover events. Used for responsive gizmos.
+     */
+    public subscribeToPointerObserver(): void {
```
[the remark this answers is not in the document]
RaananW: Was that resolved?

RaananW:
@CedricGuillemet  - any comments on this one, or can we merge?

src/Gizmos/planeRotationGizmo.ts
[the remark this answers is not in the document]
davesauce14: Function no longer used.

src/Gizmos/rotationGizmo.ts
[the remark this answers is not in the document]
davesauce14: Yes

dad72:
@RaananW A little higher Cedric said it was ok for him, that you could merge if it was ok for you.
I guess you can merge it.

PR #9080, opened 2020-10-04
BaseCameraPointersInput.ts to use checkInputs()

mrdunk:
I'm planning on using `BaseCameraPointersInput.ts` while playing with better `UniversalCamera` mouse inputs.
Unfortunately `BaseCameraPointersInput.ts` needs a bit of work.
I think what i have here is worth consideration even without any future `UniversalCamera` work. The tests indicate it's completely backwards compatible and i think it will improve the `ArkRotateCamera` and `FollowCamera` that use it.

When i originally wrote `BaseCameraPointersInput.ts` i missed the significance of the `checkInputs()` method.
Refactoring to add it now (without tearing too much of the original out) requires a few compromises.

The biggest problem i encountered was my paranoia about keeping events in sequence; If pointer events are being gathered separately from the render code where `checkInputs()` is called from we need to be very careful to apply events in the same order they were gathered.
eg: Let's pretend someone is drawing some huge geometry which is reducing performance to 0.1FPS. They are merrily clicking and dragging pointer events. 2 mouse clicks need to fire `onMouseDown()`, `onMouseUp()`, `onMouseDown()`, `onMouseUp()` in that exact order.
While it would be simpler to keep separate collections for all `onMouseDown` events and another for `onMouseUp`, this would be problematic when it comes to correctly ordering them later.

I've gone with a single `_allEvents` collection.
I don't really like how i've had to abuse type checking to get them all in the same collection.
Advice welcome here. (Very limited TS experience.)

I've also added some Observables.
I've still to write a few simple unit tests for those.


src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -89,36 +89,43 @@ export abstract class BaseCameraPointersInput implements ICameraInput<Camera> {
...
-                    this.pointB = {x: evt.clientX,
+                } else if (this._pointB === null) {
+                    this._pointB = {x: evt.clientX,
                               y: evt.clientY,
                               pointerId: evt.pointerId,
                               type: evt.pointerType };
                 }
 
-                this.onButtonDown(evt);
+                const event: _Event = [_EventType.ButtonDownEvent, <_ButtonDownEvent>evt];
```
deltakosh, line 118: we are creating a new array perf frame which seems not good for GC

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -89,36 +89,43 @@ export abstract class BaseCameraPointersInput implements ICameraInput<Camera> {
                               evt.msMovementY ||
                               0;
 
-                this.onTouch(null, offsetX, offsetY);
-                this.pointA = null;
-                this.pointB = null;
+                // TODO: Do we want to coalesce these events so only one is
+                // added per render frame? Seek advice from reviewer on how
+                // engine.isPointerLock is used.
+                let event: _Event = [_EventType.TouchEvent, [null, offsetX, offsetY]];
```
deltakosh, line 95: we should not create new array per event. That will be bad for GC

deltakosh (commented):
I'm afraid we are too close to release to have this kind of change
Several question thougth:
- How will this impact subclasses? I do not see that change in your PR
- What about backward compatibility (what about people using pointA and pointB before?) The backward compat is one of our core philosophy and cannot be compromised.

I guess I need to see the impact of sub classes to understand the overall goal of that PR because so far I struggle to get a grasph on why we want it

mrdunk:
> I'm afraid we are too close to release to have this kind of change

I presume that's a transient state?
As in you aren't accepting this level of submission now but might in the future?

> * How will this impact subclasses? I do not see that change in your PR

This change is invisible to the subclasses.
No changes would be needed there. I guess i should have stressed this more in the first post.

> * What about backward compatibility (what about people using pointA and pointB before?) The backward compat is one of our core philosophy and cannot be compromised.

This change is invisible to the subclasses. Backwards compatibility was the primary goal.

`pointA` and `pointB` were always private, just not named according to your style guide. I merely cleaned up the naming of these.
Some minor changes were also needed to the tests in the way `pointA` and `pointB` were used but this was due to the asynchronous nature of the change; We could get away with abusing them in the old (synchronous) implementation but not with my refactored (asynchronous) version.

> so far I struggle to get a grasp on why we want it

So my interpretation of the `ICameraInput.checkInputs()` method is that it is intended to decouple input events from the render loop; Many input events can be coalesced into less input events which are applied during `checkInputs()`

When manual testing I certainly see several `PointerEventTypes.POINTERMOVE` events per frame when dragging the mouse around.

Funny thing: I went to all this effort to decouple everything and haven't actually coalesced the events yet.
I'll hold of on doing that pending your opinion.
That part won't be complicated.

You know Babylon much better than me.
You also know frontend stuff better than me. (i'm usually /much/ lower on the stack.)
So if this all looks pointless to you, then let's scrap this whole thing; Apologies for the noise, etc.
It was a fun afternoon for me.

Thinking about this again now i've implemented it,
i'm not entirely convinced it's worth the extra complexity; The full implications of serializing the events hadn't really occurred to me until i tied it.

deltakosh:
> As in you aren't accepting this level of submission now but might in the future?

yes totally but this could still make it (release is on nov the 12th) if I can understand the change

> pointA and pointB were always private, just not named according to your style guide. I merely cleaned up the naming of these.

Gotcha! good catch!

> i'm not entirely convinced it's worth the extra complexity

Yeah this is where I am but I would love to get another though from @RaananW

src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -344,9 +459,51 @@ export abstract class BaseCameraPointersInput implements ICameraInput<Camera> {
...
-    private pointB: Nullable<PointerTouch>;
+    private _pointA: Nullable<PointerTouch>;
+    private _pointB: Nullable<PointerTouch>;
+}
+
+/**
+ * The following types are for various event types in the queue that feeds
+ * BaseCameraPointersInput.checkInputs().
+ */
+type _TouchEvent = [Nullable<PointerTouch>, number, number];
```
RaananW, line 487: I am not sure about the structure of those events. Using an array instead of an objects makes it feel very unstructured.
setting this to be a typescript interface will make it much cleaner:

```javascript
export interface ICameraInputTouchEvent {
  pointerTouch?: PointerTouch;
  name1: number;
  name2: number;
}
```

I know this means restructuring the way the events are stored and the way they work, but it can be unreadable for future devs maintaining this.

Same thing goes to the rest of the events defined (notice that you are not actually using them when creating them, only casting them afterwards)

RaananW (commented):
Hey, thanks for the PR!

I totally get the reason behind this - the event order might be important in certain cases. Whether it is your paranoia or someone else's, someone might need to deal with it :-)

I don't think it's too complex, but i think the implementation could be a bit more structured/readable. Setting the events inside arrays that keep on being recreated on each frame (casted to an abstract (not literally :-) ) `_Event` type seems to me very hard to read and understand. I think renaming the types (or interfaces IMO) differently will also be a bit clearer.

It's very well contained, but maybe restructuring this will do better - since we are talking about 5 different events, creating empty events, update them if and when they are triggered, and then add them to the allEvents array, which is being iterated in the checkInput. This way, you reuse objects and never create a new one. 

Bottom line is - I don't see why we can't have this integrated - it makes sense and shouldn't hurt performance if we follow the basic guidelines of back-compat and object-reusability.  We can discuss this internally as well

src/Cameras/Inputs/BaseCameraPointersInput.ts
mrdunk: Ack. makes sense.
Typescript is new to me so still trying to get a feeling for what patterns work.

And readability trumps everything IMO.

mrdunk:
> since we are talking about 5 different events, creating empty events, update them if and when they are triggered, and then add them to the allEvents array, which is being iterated in the checkInput. This way, you reuse objects and never create a new one.

Ah but that's the problem: A single event to be re-used per type isn't enough.
There can be multiple events of the same type per frame.
We don't know in advance how many, although we can make some sensible estimates and create more if they are ever needed.
Only create, never return is probably safe here although timers could be added to return excess if it was deemed necessary.

So yes, that can be done but it will need some sort of basic event factory.
Before i write something like that, is there an existing library i should be using?

As might be apparent by now i've no experience with garbage collected languages. This is interesting.

mrdunk:
Also,
in TS, is there a way to determine an object's type at run time? I searched online but didn't find a pattern  that fits.
When i iterate the `_allEvents` collection, it would be nice if i didn't need that ugly enum labeling what type of event it is.

RaananW:
> is there a way to determine an object's type at run time?

I'll start with this - there is typescript in runtime :-)  TypeScript transpiles to javascript, so at run time, it's pure javascript. The enum type switch (using numbers, eventually), is the way to go here (without overcomplicating things or using `instanceof`, which is a very expensive operation) 



> A single event to be re-used per type isn't enough

Gotcha! Yeah, makes sense. I am thinking with you now, so there might be a better suggestion, but - how about a growing array per type - events are added there to be reused. So:

1) an empty array for each event type
2) when an event is triggered find if there is an event that can be reused according to a specified index
3) reuse the event and push it to allEvents, OR push a new event to the right index and use it
4) shift events out of the allEvents array, but never from the event-based array, which is just used for reusable event objects.

Having said that - this feels very complicated for no reason. I assume that most of the times, a single event per frame will be triggered. This is an un-researched assumption, so i might be completely wrong. So maybe we could have those 5 reusable events only, but create new ones when they do happen more than once per frame.

> is there an existing library i should be using?

We try not to rely on external libraries (apart from big ones like physics engines), so we write everything we need. And we try to make it as simple as possible, to keep it maintainable and readable.

mrdunk:
> I'll start with this - there is no typescript in runtime :-)

Yup. that makes sense looking at this from compiling back to vanilla JS perspective.

> how about a growing array per type - events are added there to be reused. 

Yes. this is where i was going too.
If we pass the type-array and index into _allEvents we could use the type-array to determine the event type.

> Having said that - this feels very complicated for no reason.

Yes. I am considering whether we should scrap this whole thing.
The original code called the `onSOMETHING(...)` callbacks from within `this._pointerInput(...)`.
While this isn't perfect, (you can't coalesce multiple events into less events), i suspect it's not /that/ bad either; It depends on the complexity of the  `onSOMETHING(...)` callbacks i guess.
The existing clients (followCameraPointerInput` and `arkRotateCameraPointerInput`) don't do anything particularly time consuming in the callbacks.
Certainly the old code was simpler, therefore easier to maintain, therefore will become more efficient at some point in the future, even if it is slower right now.

I'm not qualified to make the call; I'm a back end guy. Don't know how the browser event loops interact with each other.
At a lower level i'd write tests to prove which approach was more efficient but i've no idea how you'd even start testing this sort of thing this close to the UI...

> I assume that most of the times, a single event per frame will be triggered.

So in my limited manual tests,
with a lightly loaded render stage,
i was seeing ~3-4 touch events if i rapidly zipped the mouse about.
Presumably heavily loaded systems could get worse. Do you have ways of testing this? My hunch is you could justify dropping events that exceed this by an order of magnitude; If the system is that heavily loaded, the user has bigger problems.

In theory, we can coalesce drag events into a single event but
It's actually more complicated than that;
Imagine the following events all coming in on between 2 frames:
1. Mouse down 1
2. Drag 1
3. Drag 2
4. Mouse up 1
5. Mouse down 2
6. Drag 3
7. Mouse up 2

We can happily coalesce 
Drag 1 and Drag 2 into a single event but after Mouse up 1 we need to start a new event for Drag 3 as the user probably cares that these are 2 separate click-drag things.

This is all fun to think about but i suspect the performance gains are minimal compared to the complexity.

RaananW:
I have to say I love the idea, and it does make sense. 

But! let's keep it for after 4.2 release (coming in a month). We don't want to introduce any complex changes to already-existing components at the moment. Thank you so much for all of the time invested in this! We can keep this open (maybe as draft?) or reopen this when we are done with preparations for the release, if it is ok with you!

mrdunk:
> I have to say I love the idea, and it does make sense.
Ok. let's do it.

Although it will be a while before this merges, i'll get this cleaned up over the next day or two; The chances are i will have a different project demanding my attention in a month and the closer this one is to being ready to ship the more chance there is of me coming back to it.

deltakosh:
Closing the PR for now. Let's recreate one after release :D

PR #9081, opened 2020-10-04
freeCameraTouchInput bug: Uneven panning on touch.

mrdunk:
Well this is fascinating.
Please close if it's a known issue.
I'm not sure words will explain it very well so here's a playground:

https://playground.babylonjs.com/#WPLZCD#5
I've disabled mouse and forced on touch so you can see the issue on your workstation.

Click and drag left and right.
Notice the step change in pan speed at a repeatable but arbitrary point.

it's caused by a surprising interaction between the `camera.cameraRotation` values and `this.cameraRotation.scaleInPlace(this.inertia);` in `targetCamera.ts`.

The workaround is to set `camera.inertia = 0;`.
If you decide not to merge this change, this badly needs documenting. 

When i was playing with Babylon the last time around i presumed this was annoying but intended behavior.
After looking at the code i'm convinced it's a bug.

My diff:
Simply assigning `camera.cameraRotation.y = this._offsetX / this.touchAngularSensibility` instead of adding fixes this.
On my setup the pan speed after the fix is similar to the average of the 2 speeds before.
This does change the behavior but IMO, the current implementation is verging on unusable.

My diff does a small bit of extra tidy-up too. Blame my OCD:
`if (this._offsetX && this._offsetY) {` does not take into account the case where one of the axis equal zero but the other has a valid value.

### DIFF
--- a/src/Cameras/Inputs/freeCameraTouchInput.ts
+++ b/src/Cameras/Inputs/freeCameraTouchInput.ts
@@ -163,19 +163,24 @@ export class FreeCameraTouchInput implements ICameraInput<FreeCamera> {
      * This is a dynamically created lambda to avoid the performance penalty of looping for inputs in the render loop.
      */
     public checkInputs(): void {
-        if (this._offsetX && this._offsetY) {
-            var camera = this.camera;
-            camera.cameraRotation.y += this._offsetX / this.touchAngularSensibility;
+        if (this._offsetX === null || this._offsetY === null) {
+            return;
+        }
+        if (this._offsetX === 0 && this._offsetY === 0) {
+            return;
+        }
 
-            if (this._pointerPressed.length > 1) {
-                camera.cameraRotation.x += -this._offsetY / this.touchAngularSensibility;
-            } else {
-                var speed = camera._computeLocalCameraSpeed();
-                var direction = new Vector3(0, 0, speed * this._offsetY / this.touchMoveSensibility);
+        var camera = this.camera;
+        camera.cameraRotation.y = this._offsetX / this.touchAngularSensibility;
 
-                Matrix.RotationYawPitchRollToRef(camera.rotation.y, camera.rotation.x, 0, camera._cameraRotationMatrix);
-                camera.cameraDirection.addInPlace(Vector3.TransformCoordinates(direction, camera._cameraRotationMatrix));
-            }
+        if (this._pointerPressed.length > 1) {
+            camera.cameraRotation.x = -this._offsetY / this.touchAngularSensibility;
+        } else {
+            var speed = camera._computeLocalCameraSpeed();
+            var direction = new Vector3(0, 0, speed * this._offsetY / this.touchMoveSensibility);
+
+            Matrix.RotationYawPitchRollToRef(camera.rotation.y, camera.rotation.x, 0, camera._cameraRotationMatrix);
+            camera.cameraDirection.addInPlace(Vector3.TransformCoordinates(direction, camera._cameraRotationMatrix));
         }
     }
 


deltakosh:
I see the problem but I do not understand the fix :D We are simply getting rid of the inertia which is not what we want :)

There must be another issue or at least a fix that will keep the inertia in place. 

Asking @Popov72  to have a look :)

mrdunk:
> I see the problem but I do not understand the fix :D We are simply getting rid of the inertia which is not what we want :)

So i'm only suggesting disabling inertia as a configuration only workaround.
This PR leaves inertia working as expected.

To understand problem, 
`camera.cameraRotation.y` is not the camera's current rotation but rather the rate of change.
`inertia` reduces `camera.cameraRotation` every frame.

Try un-commenting the line where `camera.cameraRotation.y` is dumped to console every frame.
You'll see inertia reducing it to 0 when touch events stop.

The existing code adds to `camera.cameraRotation.y` every frame.
This makes no sense. It is a rate of change, not an absolute position.

RaananW:
`camera.cameraRotation`, thou maybe not-so-precisely named, is used all over to define the camera's current rotation delta on the x and y axes. It is an important part of the `_checkInputs` function, called on `TargetCamera` , and is used all over the place.

> You'll see inertia reducing it to 0 when touch events stop.

That's the intended usage of this vector. So it behaves correctly. 

I do understand what you are saying - the touch movement feels off. You expect it to work the way the ArcRotateCamera is working, right? When I move the finger it should move with me and stop with me.

Popov72:
> When I move the finger it should move with me and stop with me.

If we want this behaviour, simply set inertia to 0, no?

I don't know the input system much, but the PG seems ok to me:
* if I press left mouse button and move a small distance, the rotation is slow and stops when I release LMB
* if I press left mouse button and move a larger distance, the rotation is fast and continues a bit when I release LMB due to inertia and being past the angular sensibility

By changing the touchMoveSensibility / touchAngularSensibility / inertia values one should be able to achieve any desired behaviour?

mrdunk:
> I do understand what you are saying - the touch movement feels off. You expect it to work the way the ArcRotateCamera is working, right? When I move the finger it should move with me and stop with me.

No. You misunderstand.
I get that the camera should move/rotate relative to the amount you have dragged.
What is not working correctly is that the rate at which the camera rotates is not linear.
There is a step change when you drag more than a certain amount.

Try this:
1. Click in the PG.
2. Creep the mouse rightwards (or leftwards if you prefer...). Note that the rate of rotation speeds up as you increase the length of your drag (as expected).
3. Keep creeping the mouse until the value being displayed in console changes from `false` to `true`. Note that there is a step increase in rotation speed when this happens.

I encourage you to try actually using this interface on a mobile platform; It is not good.

mrdunk:
> camera.cameraRotation, thou maybe not-so-precisely named, is used all over to define the camera's current rotation delta on the x and y axes.

Yes. exactly.
It is rotation /delta/.

The way the current code does it:
Increasing the delta every frame by the touch-drag length, then applying the inertia modifier is getting clipped in `TargetCamera._checkInputs`.
```
if (needToRotate) {
            if (Math.abs(this.cameraRotation.x) < this.speed * Epsilon) {
                this.cameraRotation.x = 0;
            }

            if (Math.abs(this.cameraRotation.y) < this.speed * Epsilon) {
                this.cameraRotation.y = 0;
            }
            this.cameraRotation.scaleInPlace(this.inertia);
        }
```
Since `this.rotation` gets set before this happens, and `this.cameraRotation` is non-linear, `this.rotation` is also non linear; You see the step change where it is clipped.

Instead /setting/ the delta every frame a touch-drag is active, then allowing it to decay back to 0 when the touch-drag is canceled gives much more deterministic results.
Because every frame starts with a constant `this.cameraRotation`, inertia only has an effect when the drag event stops.

mrdunk:
Words are hard.
let me know if you want to set up a video call where i can demonstrate.

deltakosh:
After reading this I tend to agree that inertia is not required for touch.
Waiting for @Popov72  and @RaananW  confirmation

mrdunk:
> After reading this I tend to agree that inertia is not required for touch.
But it can easily work.
There is nothing wrong with the concept in the current code. There's just a wrinkle in the details.

I'll hack this PR into a Playground this evening to demonstrate.

deltakosh:
Sorry I meant: we do not want inertia WHILE moving. So I'm ok with this PR ;)

deltakosh (approved):

mrdunk:
Ack.
thanks!
(Words are hard.)

deltakosh:
Thanks a lot for your help

PR #9082, opened 2020-10-05
Fix alphatest transparency so it no-longer requires a diffuse/albedo texture if an opacityTexture is present

metaobjects:


Popov72:
The problem is that it is a breaking change...

We would need a new parameter to trigger the new code path, so I will let @deltakosh  / @sebavan answer about this.

If we add a new code path, I would be in favor of simply testing `this._transparencyMode === Material.MATERIAL_ALPHATEST || this._transparencyMode === Material.MATERIAL_ALPHATESTANDBLEND` which is the easiest and more straightforward way to do it. We could call that path the `strict transparency mode enforcement` path, where we only use the value of `transparencyMode` to enable alpha testing / alpha blending. I think that could simplify a lot dealing with transparency, as the different combinations we have today can make it difficult to achieve the desired result (see https://doc.babylonjs.com/resources/transparency_and_how_meshes_are_rendered).

metaobjects:
I couldn't think of any situations where you would have an opacity texture and set alphatest mode and not expect the opacity texture to be used, especially as cutout works without any diffuse texture at all. What are the kind of situations this would break?

metaobjects:
Unless it's because of the default "this._transparencyMode == null || " clause people may have opacity textures assigned without setting an alpha mode and this changes the behavior in those cases? I'm happy to expand the if statement so if transparencyMode is null it only checks the diffuse/albedo texture if that would fix it?

sebavan (approved):

PR #9093, opened 2020-10-05
Fix black skeleton viewer meshes in sandbox

Drigax:
Fix for issue with skeleton viewer lighting. 
Changed skeleton viewer to use utility layer for mesh creation. 
Added utility layer lighting creation, adjusted light intensity to show surface detail.

Before:
image

After:
image

### DIFF
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -747,7 +747,6 @@ export class SkeletonViewer {
             }
 
             const light = this.utilityLayer!._getSharedGizmoLight();
-            light.includedOnlyMeshes = light.includedOnlyMeshes.concat(this.debugMesh!);
             light.intensity = 0.7;
 
             this._revert(animationState);
--- a/src/Debug/skeletonViewer.ts
+++ b/src/Debug/skeletonViewer.ts
@@ -558,6 +558,7 @@ export class SkeletonViewer {
     private _revert(animationState: boolean): void {
         if (this.options.pauseAnimations) {
             this.scene.animationsEnabled = animationState;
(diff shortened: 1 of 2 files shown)


deltakosh (approved):
Thanks!

sebavan (approved):
Love it !!!!

deltakosh:
Some checks are failing....

src/Debug/skeletonViewer.ts
```diff
@@ -744,6 +746,10 @@ export class SkeletonViewer {
                 this.debugMesh.alwaysSelectAsActiveMesh = true;
             }
 
+            const light = this.utilityLayer!._getSharedGizmoLight();
+            light.includedOnlyMeshes = light.includedOnlyMeshes.concat(this.debugMesh!);
```
Popov72, line 750: This will add in `includedOnlyMeshes ` the (same) mesh each time `_buildSpheresAndSpurs` is called.

You should check if the mesh is not already in the array before adding it.

Also, why not simply using `light.includedOnlyMeshes.push(this.debugMesh!)`?

deltakosh: Lol I totally missed it

Drigax: Is it the same mesh? Each time we call this function we dispose and rebuild a new mesh, so I'm not sure if checking for the mesh in the included mesh list is going to be a useful check...

We have a few preconditions however:
1) each SkeletonViewer has its own unique UtilityLayer scene and light
2) each time we recreate the debug mesh we dispose the old one.

It might be best to even just set the light we use to just light the newly created debug mesh?

Drigax: I'll just remove this line, I don't think we're getting much out of it...

PR #9096, opened 2020-10-05
Shadows: Fix wrong shadows when using lod meshes + too many draw calls

Popov72:
See https://forum.babylonjs.com/t/how-to-cast-shadows-of-instances-with-level-of-detail/13829/7

This PG should work when the PR is applied: https://playground.babylonjs.com/#F7KZ7C#1


Popov72:
"Texture cache" validation test is failing (in WebGL2) but it is working in my local repo.

Issue #8888, opened 2020-09-07
Second example of "Create Compound Bodies" is broken


fabienrohrer:
From the page ["Create Compound Bodies"](https://doc.babylonjs.com/how_to/compounds), open [the second playground example](https://www.babylonjs-playground.com/#492ZK0#12). The skull is not affected by physics and remains still in the air much above the ground.

CedricGuillemet:
I'm on it!

CedricGuillemet:
Reassigning to @RaananW as it's related to previous changes.

PR #9097, opened 2020-10-06
Physics scaling issue

RaananW:
Fixing #8888 

There is still an issue with oimo under specific conditions.

Playground that will work after merging:

https://playground.babylonjs.com/#492ZK0#120


CedricGuillemet:
Rotations do not seem to be taken into account. Will this work with compound of oriented objects?

RaananW:
The issue here was scaling, I changed nothing else.

sub-rotation is working well in ammo:

image

It was always an issue with oimo (Would be good to re-write the entire oimo sync procedures).  Looking into cannon now

RaananW:
And now in cannon as well:

image

CedricGuillemet (approved):
LGTM!

RaananW:
I am closing for now, I want to test every example we have before finalizing it.

RaananW:
Apart from oimo acting up on compound of the mesh itself (due to an issue with the scaling), all demos work as expected.

deltakosh:
Let's merge?

PR #9107, opened 2020-10-07
fix linkWithMesh signature to support TransformNode

beppemarazzi:
with 41333fe067eadfa3096a6a249b4b7a1b00e32812 and f8d2898fc42910ec79af62889f62179481ed5fe6 the behavior of `linkWithMesh` was improved to support a plain `TransformNode` instead of `AbstractMesh`, but the signature was not changed so the user may be confused. This simple PR fixes this.


deltakosh:
Thanks!

PR #9113, opened 2020-10-07
KHR_texture_basisu: Use RGBA format for KTX2 files that are not color data…

### LINKED ISSUE #8982: KHR_texture_basisu should transcode UASTC textures with non-color data to uncompressed RGBA when appropriate
See https://github.com/BabylonJS/Babylon.js/pull/8775#discussion_r483266575

The spec now has [a note](https://github.com/KhronosGroup/glTF/blob/0ac09684d67ee8feea417832da3085ac25daeb91/extensions/2.0/Khronos/KHR_texture_basisu/README.md#using-ktx-v2-images-with-basis-universal-supercompression-for-material-textures) on this:
> When the high-quality GPU formats (ASTC or BC7) are not available at runtime, engines should transcode UASTC textures with non-color data to uncompressed RGBA to avoid artifacts.

Popov72:
… if ASTC/BC7 not available when transcoding an UASTC source picture.

Close #8982.


deltakosh (approved):

deltakosh:
we have some issues :D

lexaknyazev:
JSYK, there's a lite transcoder module for UASTC to RGBA32 conversion.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -56,6 +56,29 @@ interface IRegisteredExtension {
...
+    Glossiness,
+    /** specular map */
+    Specular,
+    /** transmission map (thickness) */
+    Transmission,
+    /** metallic/roughness data */
+    MetallicRoughness,
+    /** occlusion data */
+    Occlusion
+}
```
bghgary, line 80: I'm not sure why this enum is necessary. The code only needs to know whether the texture is color data or not. Can we just use a boolean?

loaders/src/glTF/2.0/glTFLoaderExtension.ts
```diff
@@ -107,19 +109,21 @@ export interface IGLTFLoaderExtension extends IGLTFBaseLoaderExtension, IDisposa
      * @param context The context when loading the asset
      * @param textureInfo The glTF texture info property
      * @param assign A function called synchronously after parsing the glTF properties
+     * @param textureDataType type of data held by the texture
      * @returns A promise that resolves with the loaded Babylon texture when the load is complete or null if not handled
      */
-    loadTextureInfoAsync?(context: string, textureInfo: ITextureInfo, assign: (babylonTexture: BaseTexture) => void): Nullable<Promise<BaseTexture>>;
+    loadTextureInfoAsync?(context: string, textureInfo: ITextureInfo, assign: (babylonTexture: BaseTexture) => void, textureDataType: TextureDataType): Nullable<Promise<BaseTexture>>;
```
bghgary, line 115: It think this is a breaking change? Existing glTF loader extensions will not compile right? We should discuss this.

bghgary (commented):
Sorry for the late review.

bghgary:
@lexaknyazev 
> JSYK, there's a lite transcoder module for UASTC to RGBA32 conversion.

@Popov72 Are we using this already?

loaders/src/glTF/2.0/glTFLoader.ts
Popov72: I had done this instead of the bool in case we would need a more precise information in the future.

I have reverted to use a bool.

loaders/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -4,15 +4,14 @@ import { Material } from "babylonjs/Materials/material";
...
 import { IGLTFLoaderExtension } from "../glTFLoaderExtension";
-import { GLTFLoader } from "../glTFLoader";
+import { GLTFLoader, TextureDataType } from "../glTFLoader";
 import { IKHRMaterialsClearcoat } from 'babylonjs-gltf2interface';
 
 const NAME = "KHR_materials_clearcoat";
 
 /**
- * [Proposed Specification](https://github.com/KhronosGroup/glTF/pull/1677)
+ * [Specification](https://github.com/KhronosGroup/glTF/pull/1677)
```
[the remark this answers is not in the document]
Popov72: Done

loaders/src/glTF/2.0/glTFLoaderExtension.ts
Popov72: I think it's not.

If I do this in KHR_texture_transform (just for testing purpose):
```typescript
public loadTextureInfoAsync(context: string, textureInfo: ITextureInfo,
  assign: (babylonTexture: BaseTexture) => void/*, isColorData = true*/): Nullable<Promise<BaseTexture>> {
    [...]
```
it does work.

Popov72:
> @lexaknyazev
> 
> > JSYK, there's a lite transcoder module for UASTC to RGBA32 conversion.
> 
> @Popov72 Are we using this already?

Not yet, will add it in the new PR.

loaders/src/glTF/2.0/glTFLoader.ts
bghgary: We don't know what kind of textures there will be in the future. This enum includes entries from extensions which works for uilt-in extensions but can't be updated in custom extensions. Anyways, you've already updated to a bool, so we should be good.

loaders/src/glTF/2.0/glTFLoaderExtension.ts
bghgary: I'm not sure adding a parameter is the right fix for this. It might be better to add a property to the ITextureInfo class instead.

PR #9114, opened 2020-10-07
Add a property to the EffectLayer class to render the bounding boxes unaffected by the effect

Popov72:
See https://forum.babylonjs.com/t/strange-bloom-effect-on-bounding-boxes-when-a-highlight-layer-is-added/14737

Property added: `disableBoundingBoxesFromEffectLayer`


deltakosh (approved):
good one! I'm also working on adding support for BB for isntances :)

Issue #9122, opened 2020-10-08
Array.from errors when running in IE11


bghgary:
Array.from is not supported in IE11. There are a bunch of usages in the code that should be replaced. Some of them can be replaced with Tools.Slice, but some of them cannot.

This [PR](#9119) was an attempt at fixing using Tools.Slice, but it doesn't compile for some cases since it was converting a typed array to a regular array.

syntheticmagus:
#9123 addresses this for at least the one easy case.

PR #9127, opened 2020-10-08
Remove Array.from references from Core

RaananW:
A new function was introduced - SliceToArray that will force the return of an array instead of a typed array.

Adressing #9122


Popov72:
Great, LGTM!

sebavan (approved):

RaananW:
(Replaces #9119)

Issue #9120, opened 2020-10-07
Loading logo is misplaced when running in IE11


bghgary:
![image](https://user-images.githubusercontent.com/19158740/95399916-299e9800-08be-11eb-93a9-acdcd3488000.png)

Here is the src folder in my localDev folder for the repro:
[src.zip](https://github.com/BabylonJS/Babylon.js/files/5344586/src.zip)

Note that you will run into #9119 which needs to be merged before this repro is possible.

Issue #8941, opened 2020-09-11
Loading spinner is scaled up based on window size instead of rendering at specified scale


PatrickRyanMS:
**Repro**
- Load a glb into the sandbox or load a PG that renders the loading spinner
- Expected result: loading spinner is rendered at a consistent scale set by design
- Current result: loading spinner scales based on the size of the browser window to fill the window

**Screenshots**
![image](https://user-images.githubusercontent.com/31326855/92970450-8cbe1b80-f433-11ea-8ef4-549098e21a34.png)

PR #9128, opened 2020-10-08
Fixed loading screen in IE

RaananW:
Tested on IE11, chrome, edge and mobile chrome.

Closing #9120


Popov72:
Does #8941 still ok after the fix?

I'm just asking because David did some changes for #8941 that are touching the same blocks of CSS.

sebavan (approved):

RaananW:
> Does #8941 still ok after the fix?

As far as my tests go, yes. It is just an addition on top of the work I did there to keep things correctly sized and aligned.

PR #9132, opened 2020-10-08
Avoid GC when updating _positions array

RaananW:
`geometry._positions` resets every frame (when bounding box updates) and recreates when (for example) `mesh.intesects` is called.
With large meshes this can be very costly, as vectors are constantly created and GCed (on each frame).

This PR keeps a cache of the positions, while keeping the current state as it is. _positions is still null or full, _positionsCache is updated when needed, reusing the vectors if possible.

Important lines (due to formatting) starts at 875

### DIFF
--- a/src/Meshes/geometry.ts
+++ b/src/Meshes/geometry.ts
@@ -872,18 +872,16 @@ export class Geometry implements IGetSetVerticesData {
             return false;
         }
 
+        for (let index = this._positionsCache.length * 3, arrayIdx = this._positionsCache.length; index < data.length; index += 3, ++arrayIdx) {
+            this._positionsCache[arrayIdx] = Vector3.FromArray(data, index);
+        }
+
         for (let index = 0, arrayIdx = 0; index < data.length; index += 3, ++arrayIdx) {
-            if (this._positionsCache[arrayIdx]) {
-                this._positionsCache[arrayIdx].set(data[0 + index], data[1 + index], data[2 + index]);
-            } else {
-                this._positionsCache[arrayIdx] = Vector3.FromArray(data, index);
-            }
+            this._positionsCache[arrayIdx].set(data[0 + index], data[1 + index], data[2 + index]);
         }
 
         // just in case the number of positions was reduced, splice the array
-        if (this._positionsCache.length > data.length / 3) {
-            this._positionsCache.splice(data.length / 3);
-        }
+        this._positionsCache.length = data.length / 3;
 
         this._positions = this._positionsCache;
 
--- a/src/Meshes/geometry.ts
+++ b/src/Meshes/geometry.ts
@@ -1,7 +1,7 @@
 import { Nullable, FloatArray, DataArray, IndicesArray } from "../types";
 import { Scene } from "../scene";
 import { Vector3, Vector2 } from "../Maths/math.vector";
-import { Color4 } from '../Maths/math.color';
+import { Color4 } from "../Maths/math.color";
 import { Engine } from "../Engines/engine";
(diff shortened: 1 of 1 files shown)


src/Meshes/geometry.ts
```diff
@@ -872,12 +872,21 @@ export class Geometry implements IGetSetVerticesData {
...
+            } else {
+                this._positionsCache[arrayIdx] = Vector3.FromArray(data, index);
+            }
+        }
 
-        for (var index = 0; index < data.length; index += 3) {
-            this._positions.push(Vector3.FromArray(data, index));
+        // just in case the number of positions was reduced, splice the array
+        if (this._positionsCache.length > data.length / 3) {
+            this._positionsCache.splice(data.length / 3);
```
Popov72, line 885: If the user keeps changing the number of vertices each frame / often (which is one use case for this PR I think), you may call `splice` a great number of times (I don't know how bad it is to splice but there must be at least one array creation, the one returned by the function).

I would not splice the array and would keep the larger size the array has ever grown, I'm not sure it is a problem in regular working mode => we can always provide a `resetPositionsCache()` if really needed.

Also, it's a bit worrying to have a `if/else` check for each element of the array. I would remove this check (and keep only the `.set` path) and instead add something like this before the loop:
```typescript
for (let index = data.length, arrayIdx = this._positionsCache.length; index < data.length; index += 3, ++arrayIdx) {
    this._positionsCache[arrayIdx] = Vector3.FromArray(data, index);
}
```
As this loop already fills the `_positionsCache` array beginning at `data.length / 3` you could "optimize" the second loop limit to be `Math.min(savedPositionsCacheLength * 3, data.length)` instead of `data.length`, `savedPositionsCacheLength` being the length of `_positionsCache` before the first loop (not sure we need to do that, though, as `_positionsCache` won't grow indefinitely, so at some time `_positionsCache.length * 3` will always be less then `data.length`).

Disclaimer, all of this is untested so could well be wrong :)

RaananW: The usecase here is actually constant update of existing vertices and not adding/removing vertices. This usecase should be covered thou.

Not quite sure what the problem with the if/else inside the loop but i like the approach of updating the size prior to updating the actually vectors. I'll use this.

Regarding deleting - I have thought long about this. My initial approach was to force both arrays to have the same length (a simple a1.length = data.length/3). I then decided to use aplice, but when I think about it now, it is not needed. i should set the length and let the browser take care of the rest. This is just to stay defensive. as you say - it shouldn't matter. Unless the user iterates this array, and then it does. If the user is actually using the indices, it doesn't matter what the size of this array really is.

Popov72: > Not quite sure what the problem with the if/else inside the loop but i like the approach of updating the size prior to updating the actually vectors. I'll use this.

I may be too sensitive about if/then/else because in the old days back when doing demos, it was very bad to branch, the processor cache was trashed in the process. Maybe it's not so bad today (also, we were coding in assembly language whereas today we have a number of layers between the processor and the javascript engine that probably trash the cache several times before executing actual js code...).

deltakosh (approved):

src/Meshes/geometry.ts
RaananW: I remember seeing quite a few perf tests in javascript for if/else and switch. I can't find them right now, but performance was well optimized.

Anyhow - changed :-)

deltakosh: It is ok to branch, we did several test in the past and as long as the objects are not on slow path (eg. more than 256 properties) there is no slow down

PR #9133, opened 2020-10-08
KTX2 decoder: add support for the UASTC->RGBA universal transcoder

Popov72:
I have also updated the existing transcoders to their latest version.

I'm not sure I'm using the srgb/unorm modules in the right way...

I'm using the srgb module when  the DFD transfer function is sRGB, else I'm using the unorm module: @lexaknyazev is it how it's supposed to be used?

On my test files, using the srgb or unorm modules does not change anything visually...

Local test PG: http://localhost:1338/Playground/index-local.html#EIJH8L#19
Online PG: https://www.babylonjs-playground.com/#EIJH8L#20 => this will use the new modules only after the PR is merged (look at the console)


lexaknyazev:
> I'm using the srgb module when the DFD transfer function is sRGB, else I'm using the unorm module: @lexaknyazev is it how it's supposed to be used?

**TL;DR** Yes. The difference is _very_ subtle and is related to UASTC internal design goals. Brace yourself.

UASTC was designed to be a strict subset of ASTC with lossless conversion. The latter has a very complex specification that changes slightly internal interpolation function depending on whether the final image will use sRGB -> linear decoding. This design is unique to ASTC. For UASTC to RGBA32 (software) results to exactly match UASTC to ASTC (software) to RGBA32 (hardware) results, we had to replicate these differences. Here's a flowchart that shows where they could be observed:
image

Specifically, Alpha values will always be the same, while RGB may differ a bit.

Popov72:
Ok, thanks for the clarification!

bghgary (approved):

PR #9138, opened 2020-10-09
NME: improvements to post processes

Popov72:
I have ported [this shader](https://www.shadertoy.com/view/XsKGRW) (the post process part) from shadertoy to NME: https://nme.babylonjs.com/#8MNHL1#3

In doing so, I had to modify a number of things to make it possible:
* `smoothstep` allowed only floats, I updated it to allow any type of inputs
* you can now use multiple `CurrentScreenBlock` blocks in the material to be able to sample multiple times the screen texture
* the sampling code generated by `CurrentScreenBlock` needs to run in the fragment shader, not in the vertex, else a whole class of post processes can't be done

Here are two PGs that are using the post process:
* https://playground.babylonjs.com/#7CR6WZ
* https://playground.babylonjs.com/#7CR6WZ#1

Screenshots without and with the post process (note that the post process is animated, which you won't see in the screenshots, except if you're plastered):
image
image
image
image


RaananW:
validation tests seem to fail

Popov72:
In the same way than for `CurrentScreenBlock`, I have allowed to use multiple `ParticleTextureBlock` blocks in a particle material: http://localhost:1338/nodeEditor/public/index-local.html#BDK1QF

deltakosh (approved):
All good!

deltakosh:
There is just a bit of conflict and you can merge

PR #9149, opened 2020-10-11
Update boundingBoxRenderer.ts

aWeirdo:
Can we add a onRessourcesReadyObservable to boundingBoxRenderer? 🙂 

I need to access the material after creation, but the beforeRendering observable is too late


RaananW (approved):
Thou I can't see the main use-case for this, I guess you've found one :smile:. Got no problem with an extra observable.

Popov72:
nit: Shouldn't it be `resources` instead of `ressources`?

aWeirdo:
> nit: Shouldn't it be `resources` instead of `ressources`?

Yup, you're right, copy/paste biting my behind 😄 
i'll do a PR in a minute

PR #9154, opened 2020-10-12
Remove element references from camera

RaananW:
The first variable of the attachControl function is no longer an HTML element.

A function is Tools was added to parse the arguments and make sure to return the correct value for noPreventDefaults.


src/Misc/tools.ts
```diff
@@ -724,6 +742,18 @@ export class Tools {
         window.URL.revokeObjectURL(url);
     }
 
+    public static BackCompatCameraNoPreventDefault(args: IArguments): boolean {
+        // is it used correctly?
+        if (typeof args[0] === "boolean") {
+            return args[0];
+        } else if (typeof args[1] === "boolean") {
+            Tools.Warn("Make sure you don't pass HTML element to attachControl function");
```
deltakosh, line 750: We should mention engine.inputElement

RaananW: Will do!

aWeirdo:
I think this broke PG for older babylon versions 🤔 
< 4.2 all console " e.addEventListener is not a function"

deltakosh:
Yeah I just release that...
That sucks but I guess we need to revert it

deltakosh:
Which examples are you referring to?

aWeirdo:
default example, sorry :)

deltakosh:
oh sure!

PR #9157, opened 2020-10-12
Float comparison using != instead of epsilon difference

CedricGuillemet:
Always, like ALWAYS, compare floats using epsilon. Float representation has limits !!!

Follow up on this thread : https://forum.babylonjs.com/t/static-physics-objects-still-moving-ammojs/14823/2


deltakosh:
LOL

RaananW:
We are the 0.00000001%!!!

RaananW (approved):

NullSoldier:
Thanks for following up on this Cedric! :)

PR #9160, opened 2020-10-13
Update stlSerializer.ts

aWeirdo:
Fix wrong orientation, flipped faces and bloated binary export.

-- Original mesh
head_original

-- After current export
head_export_now

-- After fixed export
head_export_fixed

### DIFF
--- a/serializers/src/stl/stlSerializer.ts
+++ b/serializers/src/stl/stlSerializer.ts
@@ -53,7 +53,7 @@ export class STLExport {
             for (let i = 0; i < meshes.length; i++) {
                 let mesh = meshes[i];
                 let indices = mesh.getIndices();
-                faceCount += !mesh.isUnIndexed ? indices.length/3 : 0;
+                faceCount += !mesh.isUnIndexed ? indices.length / 3 : 0;
             }
 
             let bufferSize = 84 + (50 * faceCount);
--- a/serializers/src/stl/stlSerializer.ts
+++ b/serializers/src/stl/stlSerializer.ts
@@ -52,8 +52,7 @@ export class STLExport {
         if (binary) {
             for (let i = 0; i < meshes.length; i++) {
                 let mesh = meshes[i];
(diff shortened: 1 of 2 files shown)


serializers/src/stl/stlSerializer.ts
```diff
@@ -12,23 +12,23 @@ export class STLExport {
...
-                new Vector3(vertices[id[1]], vertices[id[1] + 1], vertices[id[1] + 2]),
-                new Vector3(vertices[id[2]], vertices[id[2] + 1], vertices[id[2] + 2])
+                new Vector3(vertices[id[0]], vertices[id[0] + 2], vertices[id[0] + 1]),
+                new Vector3(vertices[id[1]], vertices[id[1] + 2], vertices[id[1] + 1]),
+                new Vector3(vertices[id[2]], vertices[id[2] + 2], vertices[id[2] + 1])
             ];
             let p1p2 = v[0].subtract(v[1]);
             let p3p2 = v[2].subtract(v[1]);
-            let n = (Vector3.Cross(p1p2, p3p2)).normalize();
+            let n = (Vector3.Cross(p1p2, p3p2)).normalize().negateInPlace();
```
RaananW, line 31: Just wondering about the normal negation - why is this needed?

RaananW (approved):

serializers/src/stl/stlSerializer.ts
Popov72: It must be because the vertices are now x/z/y instead of x/y/z.

But it may be easier to simply do: `let n = (Vector3.Cross(p3p2, p1p2)).normalize()`

aWeirdo:
Last minute changes,
Removed n vector3 negating & switched the input values as suggested.
Updated default fileName to match binary mesh name (which is always 'stlmesh')
Updated ASCII mesh name to match binary mesh name for consistency
And removed fileName check in download code as it is defaulted at beginning.

RaananW (approved):

Issue #9159, opened 2020-10-12
Add Value entry box to all sliders in NME


PirateJC:
Take the same solution we did for the Gradient Node slider in NME and add that ability to all sliders in NME.

https://github.com/BabylonJS/Babylon.js/issues/9103

PR #9164, opened 2020-10-13
NME - Adding input to slider

msDestiny14:
slider

https://github.com/BabylonJS/Babylon.js/issues/9159


nodeEditor/src/sharedComponents/sliderLineComponent.tsx
```diff
@@ -100,14 +103,26 @@ export class SliderLineComponent extends React.Component<ISliderLineComponentPro
     }
 
     render() {
+
+        const input = this.props.globalState ?
+        <FloatLineComponent globalState={this.props.globalState} smallUI={true} label="" target={this.state} propertyName="value" min={this.prepareDataToRead(this.props.minimum)} max={this.prepareDataToRead(this.props.maximum)}
```
deltakosh, line 108: why not having the input all the time?

msDestiny14: do you think we should? I'm not sure all the places that include it are wanted but I can check.

deltakosh: I think so unless we find an obvious reason not to do so (same for inspector by the way)

deltakosh:
Ok excellent please do the same for the inspector :)

PR #9168, opened 2020-10-13
Introduce Texture.ForceSerializeBuffers

aWeirdo:
Reference; https://forum.babylonjs.com/t/option-to-include-textures-in-babylon-files/14070

Adds static Texture.ForceSerializeBuffers boolean
~~Adds prototype Texture.toBase64()~~
Adds static textureTools.GenerateBase64StringFromTexture

Scrapping the original thought of passing a boolean down the serialization process,
Adding a second static option to Texture.SerializeBuffers seems like the better approach, 

Future/to-do;
Test cubeTexture.. is it even possible with a single base64string?
~~Test EncodeArrayBufferToBase64 instead of canvas.~~

Would like to hear thoughts & input, pick your brains 🙂

### DIFF
--- a/src/Materials/Textures/baseTexture.ts
+++ b/src/Materials/Textures/baseTexture.ts
@@ -743,6 +743,52 @@ export class BaseTexture implements IAnimatable {
     }
 
     /**
+     * Reads the pixels stored in the webgl texture and returns them as a base64 string.
+     * @returns The base64 encoded string or null.
+     */
+    public toBase64(): Nullable<string> {
+        if (!this._texture) {
+            return null;
+        }
+
+        var size = this.getSize();
+        var width = size.width;
+        var height = size.height;
+        var pixels = this.readPixels();
+        var isFloatBuffer = (pixels instanceof Float32Array);
+        var invertY = this._texture.invertY;
+
+        var canvas = document.createElement('canvas');
+        canvas.width = width;
+        canvas.height = height;
+
+        var ctx = canvas.getContext('2d');
+        ctx.fillStyle = 'rgb(0, 0, 0)';
+        ctx.fillRect(0, 0, width, height);
+
+        var wCount = 0;
+        var hCount = 0;
+        var drawHeight = 0;
+        for (var i = 0; i < pixels.length; i+=4) {
+            if (wCount >= width) {
+                wCount = 0;
+                hCount++;
+            }
+            drawHeight = (invertY ? height - hCount - 1 : hCount);
+
+            if (isFloatBuffer) {
+                ctx.fillStyle = `rgba( ${pixels[i]*255}, ${pixels[i+1]*255}, ${pixels[i+2]*255}, ${pixels[i+3]})`;
+            } else {
+                ctx.fillStyle = `rgba( ${pixels[i]}, ${pixels[i+1]}, ${pixels[i+2]}, ${pixels[i+3]/255})`;
+            }
+
+            ctx.fillRect(wCount, drawHeight, 1, 1);
+            wCount++;
+        }
+        return canvas.toDataURL('image/png');
+    }
+
+    /**
      * Release and destroy the underlying lower level texture aka internalTexture.
      */
     public releaseInternalTexture(): void {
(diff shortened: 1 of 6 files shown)


deltakosh:
I kind of like it
What I do not like too much is adding code to baseTexture as we try to keep it as slim as possible
Maybe we can move it to the TextureTools class? (as a static function)

aWeirdo:
> Maybe we can move it to the TextureTools class? (as a static function)

Sure thing!

First i wanted to look into avoiding the canvas,
This is not an area i'm strong in, not sure what i'm looking at, 

    BABYLON.StringTools.EncodeArrayBufferToBase64(pixelsInUint8Array);

is returning a string like, (total 1.4 MB)
"KzYN/0RWF/97f0P/...."

canvas returns same image, (total 1 MB)
"iVBORw0KGgoAAAANS...."

src/Materials/Textures/baseTexture.ts
```diff
@@ -742,6 +742,52 @@ export class BaseTexture implements IAnimatable {
...
+            } else {
+                ctx.fillStyle = `rgba( ${pixels[i]}, ${pixels[i + 1]}, ${pixels[i + 2]}, ${pixels[i + 3] / 255})`;
+            }
+
+            ctx.fillRect(wCount, drawHeight, 1, 1);
+            wCount++;
+        }
+        return canvas.toDataURL('image/png');
+    }
+
```
Popov72, line 790: I think this may be slow, as the fillStyle string must be decoded for each character.

Maybe something like this would be better: https://playground.babylonjs.com/#KPV56G (lines 35-62)

Popov72:
> is returning a string like, (total 1.4 MB)
> canvas returns same image, (total 1 MB)

That's because `EncodeArrayBufferToBase64` simply base64 encodes your raw array whereas when you take the canvas route you get a png encoded picture (`canvas.toDataURL('image/png')`).

src/Materials/Textures/baseTexture.ts
aWeirdo: It is a bit slow yeah, 
i'll try with yours, looks better already in PG :)
and i'll look into the faceIndex & mipmap

src/Materials/Textures/texture.ts
```diff
@@ -12,6 +12,7 @@ import { TimingTools } from '../../Misc/timingTools';
 import { InstantiationTools } from '../../Misc/instantiationTools';
 import { Plane } from '../../Maths/math.plane';
 import { StringTools } from '../../Misc/stringTools';
+import { TextureTools } from '../../Misc/textureTools';
```
aWeirdo, line 15: Tests report circular dependency, how can this be avoided? :)

deltakosh:
checks are failing :(

src/Misc/textureTools.ts
```diff
@@ -72,4 +73,77 @@ export class TextureTools {
...
+        canvas.width = width;
+        canvas.height = height;
+
+        var ctx = canvas.getContext('2d');
+        if (!ctx) {
+            return null;
+        }
+
+        var imageData = ctx.createImageData(width, height);
+        imageData.data.set(pixels);
```
aWeirdo, line 127: @deltakosh 
I don't know what it wants :)
Pixels is always a Uint8Array at this point in the code, it has a length property..
Any idea?

    [at-loader] ../../src/Misc/textureTools.ts:127:28 
    TS2345: Argument of type 'ArrayBufferView' is not assignable to parameter of type 'ArrayLike<number>'.
    Property 'length' is missing in type 'ArrayBufferView' but required in type 'ArrayLike<number>'.

deltakosh: you should be able to cast it

aWeirdo: Okay, that should fix it.
Now there's just the linting left, it started again with circular dependencies.
TextureTools import Texture, Texture import TextureTools.. thoughts?

deltakosh: we can create a new file then like CopyTools

aWeirdo: I'll have it finished by tomorrow at the latest
Thanks for the guidance both @deltakosh & @Popov72

Popov72:
LGTM!

PR #9191, opened 2020-10-15
glTF KHR_materials_translucency loading

MiiBond:
I finally got around to writing this.

A few notes:
1. Babylon currently loads the translucency mask from the G channel of the thickness texture. Neither `KHR_materials_transmission` or `KHR_materials_translucency` loaders are handling this properly. When we support the volume extension (with thickness map), this will definitely need to be dealt with. Related to this, according to the specs, transmission should override translucency.
2. Translucency in Babylon currently only considers lights and not background objects. Refraction can use a cubemap probe or screen space (2D) texture so it's possible to see objects through a transmissive surface (as the transmission loader currently sets up). It might be worth considering allowing translucency to work this way too. Once realtime filtering of 2D textures is working, we could use this for screen space translucency.
3. I noticed that translucency is added on top of other lighting in a way that I don't think is energy-conserving. For example, adding translucency to a leaf would brighten both sides as the usual irradiance wasn't being decreased in proportion to the light being transmitted. I therefore added a `1-transmittance` factor to the diffuse lighting. @sebavan Please check this as I'm not sure about this at all.


sebavan:
LGTM :-) @bghgary  could you check about the GLTF part ???

@MiiBond totally agree that we might need:

1. a special flag to setup map inputs according to the spec
2. reuse the screen space for filtering

Would you be able to push them with the volume one ???

MiiBond:
Push the changes at the same time as the volume extension? Yeah, that makes sense for the thickness texture changes.
When do you think the realtime filtering of 2D maps could be done?

loaders/src/glTF/2.0/Extensions/KHR_materials_translucency.ts
```diff
@@ -0,0 +1,98 @@
...
+            promises.push(this._loadTranslucentPropertiesAsync(extensionContext, material, babylonMaterial, extension));
+            return Promise.all(promises).then(() => { });
+        });
+    }
+
+    private _loadTranslucentPropertiesAsync(context: string, material: IMaterial, babylonMaterial: Material, extension: IKHRMaterialsTranslucency): Promise<void> {
+        if (!(babylonMaterial instanceof PBRMaterial)) {
+            throw new Error(`${context}: Material type not supported`);
+        }
+        let pbrMaterial = babylonMaterial as PBRMaterial;
```
bghgary, line 64:
```suggestion
        const pbrMaterial = babylonMaterial as PBRMaterial;
```

loaders/src/glTF/2.0/Extensions/KHR_materials_translucency.ts
```diff
@@ -0,0 +1,98 @@
...
+    public order = 175;
+
+    private _loader: GLTFLoader;
+
+    /** @hidden */
+    constructor(loader: GLTFLoader) {
+        this._loader = loader;
+        this.enabled = this._loader.isExtensionUsed(NAME);
+        if (this.enabled) {
+            (loader as any)._parent.transparencyAsCoverage = true;
```
bghgary, line 40: Instead of casting to `any`, it's probably better to expose the `_parent` object in `GLTFLoader`.

bghgary (commented):
Minor comments. Looks good otherwise.

loaders/src/glTF/2.0/Extensions/KHR_materials_translucency.ts
bghgary: Never mind, it's already exposed:

```suggestion
            this._loader.parent.transparencyAsCoverage = true;
```

MiiBond: Done

sebavan:
> Push the changes at the same time as the volume extension? Yeah, that makes sense for the thickness texture changes.

Thanks

> When do you think the realtime filtering of 2D maps could be done?

This is a great question as I have absolutely no clue how to do it at the moment :-) I would prefer to not introduce it at least before our 4.2 release coming in a month. Would that fit with the extension release ?

MiiBond:
@sebavan did you have any thoughts on the energy conservation logic that I added? It will change the look of existing translucent materials so perhaps it should be behind a new flag?
For example, this mesh is set at fully translucent with a red tint colour and a constant 0 thickness:
Screen Shot 2020-10-15 at 2 29 41 PM
Screen Shot 2020-10-15 at 2 30 28 PM

Before, it was glowing red all over. Now, with full translucency, the reflected irradiance is scaled with 1.0 - transmittance, resulting in less red being reflected.
Note that this example isn't meant to look like real skin :) I'm using `translucency=1` and `thickness=0` to demonstrate the extremes.

sebavan:
This actually makes much more sense cause otherwise we are creating energy so I am kinda considering it as a bug fix :-)

@Popov72 any thoughts ?

@MiiBond, and @bghgary could you check with the GLTF group for this one ?

Popov72:
It seems good to me, as the `refractionIrradiance` is modulated by `transmittance` in the sub surface block.

I think a new flag for this would be nice in case people want to get the same rendering than before after the change, but if it is a bug fix maybe we shouldn't add it? @sebavan will tell.

sebavan:
I currently flagged as draft and will reenable once ready, I will manually backport the fixes to master to ensure they make it through 4.2.

sebavan:
Actually @MiiBond the volume PR is not ready but is this one ready to merge ???

MiiBond:
@sebavan Yes, this one should be good to go now.

PR #9193, opened 2020-10-15
Replace input handling in InputManager with DeviceSourceManager/DeviceInputSystem

PolygonalSun:
This PR contains all of the code to replace the current input functionality inside of the  InputManager with the DeviceSourceManager.  As of right now, the event-based functionality as been replaced with calls to the DeviceSourceManager and the returned data is then injected into a simulated event that will work with the existing code.  There are a couple of other changes to support this.

DeviceInputSystem now supports MouseWheel data and stores the PointerEvent.movementX/Y values.  This second one is added to support pointer locked interactions because the actual cursor position doesn't change so the deltas need to be tracked.


deltakosh:
462 files seems a bit too much :D

PolygonalSun:
> 462 files seems a bit too much :D

Yeah, I think that when I squashed the original pull from master, I messed up the commit history and GitHub hates that.  I'm just gonna back things up and redo the changes.

PR #9194, opened 2020-10-15
Replace input handling in InputManager with DeviceSourceManager/DeviceInputSystem

PolygonalSun:
This PR contains all of the code to replace the current input functionality inside of the InputManager with the DeviceSourceManager. As of right now, the event-based functionality as been replaced with calls to the DeviceSourceManager and the returned data is then injected into a simulated event that will work with the existing code. There are a couple of other changes to support this.

DeviceInputSystem now supports MouseWheel data and stores the PointerEvent.movementX/Y values. This second one is added to support pointer locked interactions because the actual cursor position doesn't change so the deltas need to be tracked.


src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
-                return;
-            }
-            elementToAttachTo.addEventListener("keydown", this._onKeyDown, false);
-            elementToAttachTo.addEventListener("keyup", this._onKeyUp, false);
-            this._keyboardIsAttached = true;
-        };
+        this._deviceSourceManager.onDeviceConnectedObservable.add((device) => {
+            // Keyboard Events
+            if (device.deviceType == DeviceType.Keyboard) {
+                this._deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!.onInputChangedObservable.add((deviceSource) => {
```
deltakosh, line 812: why do we need this?
this._deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)

from an API standpoint I would expect to be able to do device.onInputChangedObservable...

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
-            if (!elementToAttachTo || this._keyboardIsAttached) {
-                return;
-            }
-            elementToAttachTo.addEventListener("keydown", this._onKeyDown, false);
-            elementToAttachTo.addEventListener("keyup", this._onKeyUp, false);
-            this._keyboardIsAttached = true;
-        };
+        this._deviceSourceManager.onDeviceConnectedObservable.add((device) => {
+            // Keyboard Events
+            if (device.deviceType == DeviceType.Keyboard) {
```
deltakosh, line 811: please make sure to always use ===

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
+                this._deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!.onInputChangedObservable.add((deviceSource) => {
+                    if (elementToAttachTo && document.activeElement === elementToAttachTo) {
+                        // Get Keyboard key and create event init data
+                        const key = String.fromCharCode(deviceSource.inputIndex);
+                        const init = {key: key, keyCode: deviceSource.inputIndex};
+
+                        if (deviceSource.currentState == 1) {
+                            const evt = new KeyboardEvent("keydown", init);
```
deltakosh, line 819: why do we need to create a KeyboardEvent? Can't we reuse the source one?

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
+            }
+
+            // Pointer Events
+            if (device.deviceType == DeviceType.Mouse || device.deviceType == DeviceType.Touch) {
+                this._deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!.onInputChangedObservable.add((deviceSource) => {
+                    // Get current cursor position and create event init data
+                    const pointerX = this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.Horizontal) || 0;
+                    const pointerY = this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.Vertical) || 0;
+                    // If dealing with a change to the delta, grab values for event init
+                    const deltaX = (deviceSource.inputIndex == PointerInput.DeltaHorizontal) ? this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.DeltaHorizontal) : 0;
```
deltakosh, line 838: having that all over the code seems not efficient: this._deviceSourceManager.getDeviceSource(device.deviceType)

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
+                        movementX: deltaX,
+                        movementY: deltaY
+                    };
+                    const evt = (device.deviceType == DeviceType.Mouse) ? new PointerEvent("mouse", init) : new PointerEvent("touch", init);
+                    Object.defineProperty(evt, 'target', {value: elementToAttachTo});
 
-        // Keyboard events
-        this._onCanvasFocusObserver = engine.onCanvasFocusObservable.add(
-            (() => {
+                    if (attachDown && deviceSource.inputIndex >= PointerInput.LeftClick && deviceSource.inputIndex <= PointerInput.RightClick && deviceSource.currentState == 1) {   // Pointer Down
```
deltakosh, line 850: well this seems wrong as this only need to happen when the canvas is getting focused

deltakosh, line 850: What about still having pointerdown?

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
-            }
-            elementToAttachTo.addEventListener("keydown", this._onKeyDown, false);
-            elementToAttachTo.addEventListener("keyup", this._onKeyUp, false);
-            this._keyboardIsAttached = true;
-        };
+        this._deviceSourceManager.onDeviceConnectedObservable.add((device) => {
+            // Keyboard Events
+            if (device.deviceType == DeviceType.Keyboard) {
+                this._deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!.onInputChangedObservable.add((deviceSource) => {
+                    if (elementToAttachTo && document.activeElement === elementToAttachTo) {
```
deltakosh, line 813: I would recommand early exit to clean the code:
```
if (!elementToAttachTo || document.activeElement !== elementToAttachTo) {
return;
}
```

deltakosh: That being said I like that approach, it is better than the focus/blur way we did before

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
+            // Pointer Events
+            if (device.deviceType == DeviceType.Mouse || device.deviceType == DeviceType.Touch) {
+                this._deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!.onInputChangedObservable.add((deviceSource) => {
+                    // Get current cursor position and create event init data
+                    const pointerX = this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.Horizontal) || 0;
+                    const pointerY = this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.Vertical) || 0;
+                    // If dealing with a change to the delta, grab values for event init
+                    const deltaX = (deviceSource.inputIndex == PointerInput.DeltaHorizontal) ? this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.DeltaHorizontal) : 0;
+                    const deltaY = (deviceSource.inputIndex == PointerInput.DeltaVertical) ? this._deviceSourceManager.getDeviceSource(device.deviceType)!.getInput(PointerInput.DeltaVertical) : 0;
+                    const init = {
```
deltakosh, line 840: by doing this we are losing information (source, target, etc...)

src/Inputs/scene.inputManager.ts
```diff
@@ -804,64 +806,81 @@ export class InputManager {
...
 
-        attachedFunction();
-
-        // Pointer events
-        var eventPrefix = Tools.GetPointerPrefix(engine);
-
-        if (attachMove) {
-            elementToAttachTo.addEventListener(eventPrefix + "move", <any>this._onPointerMove, false);
+        // TODO: Do we need this?
+        this._onCanvasFocusObserver = engine.onCanvasFocusObservable.add(() => {
```
deltakosh, line 874: no need anymore

src/Inputs/scene.inputManager.ts
```diff
@@ -870,23 +889,15 @@ export class InputManager {
...
-
-        // Pointer
-        this._alreadyAttachedTo.removeEventListener(eventPrefix + "move", <any>this._onPointerMove);
-        this._alreadyAttachedTo.removeEventListener(this._wheelEventName, <any>this._onPointerMove);
-        this._alreadyAttachedTo.removeEventListener(eventPrefix + "down", <any>this._onPointerDown);
-        window.removeEventListener(eventPrefix + "up", <any>this._onPointerUp);
+        this._deviceSourceManager.dispose();
 
         // Blur / Focus
+        // TODO: Do we need this?
```
deltakosh, line 900: nope this can be removed

src/Inputs/scene.inputManager.ts
```diff
@@ -65,7 +66,7 @@ export class InputManager {
     private _alreadyAttachedTo: HTMLElement;
 
     // Pointers
-    private _wheelEventName = "";
+    private _wheelEventName = "wheel";
```
deltakosh, line 69: why do we need to keep this?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -249,25 +265,25 @@ export class DeviceInputSystem implements IDisposable {
...
-                    this.onInputChanged(DeviceType.Keyboard, 0, evt.keyCode, kbKey[evt.keyCode], 0);
+                    this.onInputChanged(DeviceType.Keyboard, 0, evt.keyCode, 1, kbKey[evt.keyCode]);
                 }
-                kbKey[evt.keyCode] = 0;
             }
         });
 
-        window.addEventListener("keydown", this._keyboardDownEvent);
-        window.addEventListener("keyup", this._keyboardUpEvent);
+        this._elementToAttachTo.addEventListener("keydown", this._keyboardDownEvent);
```
deltakosh, line 285: This needs to be HEAVILY tested. The choice of window was not random. It was simply not working without it on some browsers (I look at you Safari and IE)

deltakosh (changes requested):
I did a first pass
Also please do a perf review by testing how long it takes to process a constant flow of input (say mouse moves) with and without the DeviceInputSystem.

src/Inputs/scene.inputManager.ts
PolygonalSun: I originally kept this just to serve as a constant string rather than use "wheel" in several places

src/Inputs/scene.inputManager.ts
PolygonalSun: I'll make those changes

src/Inputs/scene.inputManager.ts
PolygonalSun: That's an old change from before my last pushed PR.  Good catch!  I'll change it

src/Inputs/scene.inputManager.ts
PolygonalSun: I'll remove all focus/blur stuff

src/Inputs/scene.inputManager.ts
PolygonalSun: ditto

src/Inputs/scene.inputManager.ts
PolygonalSun: There's going to be a bunch more information added to this in the next commit.  We may still lose information but the information that we'll lose we won't need.

src/Inputs/scene.inputManager.ts
PolygonalSun: I've changed this to just reuse the provided device source (eg. device.getInput(...))

src/Inputs/scene.inputManager.ts
PolygonalSun: This will be changed to use something similar to PointerEvent because Babylon Native will not know what a KeyboardEvent is.

src/Inputs/scene.inputManager.ts
deltakosh: but do we use it in that file

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: For the time being, I'm going to revert this.  This change was more for convenience so that typing in the code window in the PG wouldn't trigger items on the canvas.

src/Inputs/scene.inputManager.ts
PolygonalSun: I just changed the refs to this to use the string literal and are no longer storing this value since we aren't transforming this.

src/Inputs/scene.inputManager.ts
PolygonalSun: This comment may no longer be applicable with the new code changes.

src/Inputs/scene.inputManager.ts
PolygonalSun: Changed

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -755,7 +756,7 @@ export class AdvancedDynamicTexture extends DynamicTexture {
             this._shouldBlockPointer = false;
             // Do picking modifies _shouldBlockPointer
             let pointerId = (pi.event as PointerEvent).pointerId || this._defaultMousePointerId;
-            this._doPicking(x, y, pi, pi.type, pointerId, pi.event.button, (<MouseWheelEvent>pi.event).deltaX, (<MouseWheelEvent>pi.event).deltaY);
+            this._doPicking(x, y, pi, pi.type, pointerId, pi.event.button, (pi.event as unknown as IWheelEvent).deltaX, (pi.event as unknown as IWheelEvent).deltaY);
```
deltakosh, line 759: why as unknown?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -127,6 +133,16 @@ export class DeviceInputSystem implements IDisposable {
             throw `Unable to find input ${inputIndex} for device ${DeviceType[deviceType]} in slot ${deviceSlot}`;
         }
 
+        // TODO: Determine best way to handle wheel
```
deltakosh, line 136: we cannot merge with TODO

PolygonalSun: Removed comment

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -127,6 +133,16 @@ export class DeviceInputSystem implements IDisposable {
             throw `Unable to find input ${inputIndex} for device ${DeviceType[deviceType]} in slot ${deviceSlot}`;
         }
 
+        // TODO: Determine best way to handle wheel
+        // When the mouse wheel is moved, only clear the value if that input is polled for
+        if (deviceType == DeviceType.Mouse && (inputIndex >= PointerInput.MouseWheelX && inputIndex <= PointerInput.MouseWheelZ)) {
```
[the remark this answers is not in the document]
PolygonalSun: NGL, this was a bit confusing but I'm assuming that you mean to change == to ===

src/Inputs/scene.inputManager.ts
```diff
@@ -214,7 +211,7 @@ export class InputManager {
         }
 
         if (pickResult) {
-            let type = evt.type === this._wheelEventName ? PointerEventTypes.POINTERWHEEL : PointerEventTypes.POINTERMOVE;
+            let type = evt.type === "wheel" ? PointerEventTypes.POINTERWHEEL : PointerEventTypes.POINTERMOVE;
```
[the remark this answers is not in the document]
PolygonalSun: The PointerEventTypes enum is what tells the InputManager what type of input we're working with (Single Click, Move, Double Click, Pick, etc.)  This context exists outside of the standard event data so we need to determine which value to use when we send it to the PointerInfo objects.

src/Inputs/scene.inputManager.ts
```diff
@@ -804,100 +803,175 @@ export class InputManager {
...
+        this._deviceInputSystem.onInputChanged = (deviceType, deviceSlot, inputIndex, previousState, currentState) => {
+            let isKeyboardActive = false;
+            try {
+                isKeyboardActive = this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, 18) >= 0;
             }
-            elementToAttachTo.addEventListener("keydown", this._onKeyDown, false);
-            elementToAttachTo.addEventListener("keyup", this._onKeyUp, false);
-            this._keyboardIsAttached = true;
-        };
+            catch { } // This is just a test to see if the keyboard is present
```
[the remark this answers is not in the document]
PolygonalSun: I just implemented a check in the DeviceInputSystem to check if a device exists.  Not really sure why I didn't do this earlier.

gui/src/2D/advancedDynamicTexture.ts
PolygonalSun: Originally, that was a carry over from how we originally handled PointerEvents.  I did take another look at things and made some changes to our PointerInfo classes and this has now been changed to just:
```
pi.event as IWheelEvent
````

File: src/Cameras/Inputs/BaseCameraMouseWheelInput.ts [hunk 17]
```diff
@@ -61,7 +62,7 @@ export abstract class BaseCameraMouseWheelInput implements ICameraInput<Camera>
                 return;
             }
 
-            const event = <MouseWheelEvent>pointer.event;
+            const event = (pointer.event as unknown as IWheelEvent);
```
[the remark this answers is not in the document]
PolygonalSun: Looks like I missed a ref.  Removed _as unknown_

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -127,6 +133,16 @@ export class DeviceInputSystem implements IDisposable {
             throw `Unable to find input ${inputIndex} for device ${DeviceType[deviceType]} in slot ${deviceSlot}`;
         }
 
+        // TODO: Determine best way to handle wheel
+        // When the mouse wheel is moved, only clear the value if that input is polled for
+        if (deviceType == DeviceType.Mouse && (inputIndex >= PointerInput.MouseWheelX && inputIndex <= PointerInput.MouseWheelZ)) {
+            const currentValue = 0 + device[inputIndex];
```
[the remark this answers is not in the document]
PolygonalSun: I think that was a remnant of an older code change.  Removed.

src/Cameras/Inputs/arcRotateCameraMouseWheelInput.ts
```diff
@@ -62,7 +63,7 @@ export class ArcRotateCameraMouseWheelInput implements ICameraInput<ArcRotateCam
             if (mouseWheelLegacyEvent.wheelDelta) {
                 wheelDelta = mouseWheelLegacyEvent.wheelDelta;
             } else {
-                wheelDelta = -(event.deltaY || event.detail) * 60;
+                wheelDelta = -(event.deltaY || event.detail);
```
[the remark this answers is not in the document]
PolygonalSun: This was causing a wheelDelta change of 6000 as opposed to the original 100.  This is because we were just pulling the value for wheelDelta (any change, as opposed to a change on a specific axis).  This leads me to believe that this was outdated and never got triggered so we didn't know there was a problem.

src/Inputs/scene.inputManager.ts
```diff
@@ -804,100 +803,175 @@ export class InputManager {
...
+
+            const altKey = (isKeyboardActive && this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, 18) == 1);
```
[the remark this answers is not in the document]
PolygonalSun: Created Constants for these keys

src/Inputs/scene.inputManager.ts
ryantrem: With the current contract, the problem with using `device` directly is that even within that if statement, TypeScript has not inferred that `device` is specifically `DeviceSource<DeviceType.Keyboard>` and so methods on it like `getInput` are not type safe. On the other hand, if you call `deviceSourceManager.getDeviceSource` the returned value's exact type *is* inferred, and it is type safe. I think the problem with the current contract is that `DeviceSourceManager.onDeviceConnectedObservable` is of type `Observable<DeviceSource<DeviceType>>`. We could introduce a new "distributive conditional type" that expands to a union type of all the possible specific `DeviceSource<T>` types, and then TypeScript would properly infer the type of `device` within the `if` check. It would look like this:

```typescript
type DistributedDeviceSource<T> = T extends DeviceType ? DeviceSource<T> : never;
export type KnownDeviceSource = DistributedDeviceSource<DeviceType>;
...
public readonly onDeviceConnectedObservable = new Observable<KnownDeviceSource>((observer) => { ... }
```

In this case, `KnownDeviceSource` resolves to `DeviceSource<DeviceType.Generic> | DeviceSource<DeviceType.Keyboard> | DeviceSource<DeviceType.Mouse> | DeviceSource<DeviceType.Touch> | DeviceSource<DeviceType.DualShock> | DeviceSource<DeviceType.Xbox> | DeviceSource<DeviceType.Switch>`. This would be fully type safe, but would introduce what appears to be a new type (`KnownDeviceSource`). It would also be backward compatible with what we currently have.

A simpler alternative would have been to have the event payload of `onDeviceConnectedObservable` only expose the `deviceType` and `deviceSlot` and force the user to retrieve the device from `DeviceSourceManager.getDeviceSource`, but we can't change that now without breaking backward compatibility.

src/Inputs/scene.inputManager.ts
```diff
@@ -709,7 +708,7 @@ export class InputManager {
             this._updatePointerPosition(evt);
 
             if (scene.preventDefaultOnPointerUp && elementToAttachTo) {
-                evt.preventDefault();
+                //evt.preventDefault();
```
sebavan, line 711: Is this Sure ???

src/Inputs/scene.inputManager.ts
```diff
@@ -804,100 +803,171 @@ export class InputManager {
...
 
-            // Wheel
-            this._wheelEventName =
-                "onwheel" in document.createElement("div")
-                    ? "wheel" // Modern browsers support "wheel"
-                    : (<any>document).onmousewheel !== undefined
-                    ? "mousewheel" // Webkit and IE support at least "mousewheel"
-                    : "DOMMouseScroll"; // let's assume that remaining browsers are older Firefox
+                if (attachUp && inputIndex >= PointerInput.LeftClick && inputIndex <= PointerInput.RightClick && currentState == 0) {   // Pointer Up
+                    Object.defineProperties(evt, {
```
sebavan, line 905: This creates GC ??? like in the previous functions wonder if there are ways to prevent it

src/Inputs/scene.inputManager.ts
```diff
@@ -804,100 +803,171 @@ export class InputManager {
...
+                        /*
+                         * Since wheel inputs stay until they are read, we'll read everything at once.  We'll then fire off
+                         * a move event if any of them are not zero.  This will allow us to only fire off events with a proper
+                         * delta.
+                         */
+                        const deltaX = this._deviceInputSystem.pollInput(deviceType, deviceSlot, PointerInput.MouseWheelX);
+                        const deltaY = this._deviceInputSystem.pollInput(deviceType, deviceSlot, PointerInput.MouseWheelY);
+                        const deltaZ = this._deviceInputSystem.pollInput(deviceType, deviceSlot, PointerInput.MouseWheelZ);
+
+                        Object.defineProperties(evt, {
```
sebavan, line 932: Same here

src/Inputs/scene.inputManager.ts
```diff
@@ -659,7 +658,7 @@ export class InputManager {
             this._updatePointerPosition(evt);
 
             if (scene.preventDefaultOnPointerDown && elementToAttachTo) {
-                evt.preventDefault();
+                //evt.preventDefault();
```
sebavan, line 661: Same here is it safe to comment cause we should either remove or keep ?

PolygonalSun:
Closing because this needs to be updated to 5.0.  Will create a new one for this feature

PR #9198, opened 2020-10-16
NME: Fix bug when changing the types of the lerp/nlerp inputs

### LINKED ISSUE #9197: Replacing right port in lerp node with the same type breaks lerp node
**Repro**
- Create a lerp with two color3 inputs and a float in gradient
- create a new color3 node
- wire the output of the new color3 to replace the right port on lerp
- note that the original wire is broken, but the new wire is not connected
- click away and try to wire the new (or original) color3 to the empty right port on lerp and an incompatible type error pops
- Current result: so long as we are trying to replace the input with the same type, we should be able to drop a new wire on a port and replace what was already there.

**Video Repro**

[lerpBug.zip](https://github.com/BabylonJS/Babylon.js/files/5393947/lerpBug.zip)

Popov72:
Fix #9197


deltakosh (approved):
All good but the checks

Issue #9199, opened 2020-10-16
Custom Frames, removing connections changes the port order of the ports on the collapsed frame.


PirateJC:
With a custom frame, if you have a specific port order. If you connect to a port as in this picture:

![image.png](https://images.zenhubusercontent.com/5dcb34efa1442a0001945170/3e912a74-edc2-4edd-8497-4bd9e9056517)

everything is fine, but if you remove the connection, the port order of the custom frame changes...the connected port moves to the bottom of the list as in this picture:

![image.png](https://images.zenhubusercontent.com/5dcb34efa1442a0001945170/f42577d8-0c06-4241-aaed-9e848b4f82d9)

PR #9200, opened 2020-10-17
Fixing Node Port Bugs

msDestiny14:
When the link was being removed, the port was getting removed but then re-added if the user requested it. This would then push it to the bottom. Now it should remain in place. 

Also fixed an issue related to port positions not getting reupdated when connecting links. 

https://github.com/babylonjs/babylon.js/issues/9199


PirateJC (approved):
Looks good to me!

Issue #9201, opened 2020-10-17
Issue when taking a screenshot with multi-cameras


gabrielheming:
The method [BABYLON.ScreenshotTools.CreateScreenshotUsingRenderTarget](https://doc.babylonjs.com/api/classes/babylon.screenshottools#createscreenshotusingrendertarget) is not respecting the selected render target when the scene has multiple cameras.

[Playground demo](https://www.babylonjs-playground.com/pg/E9IRIF/revision/266).

Line 82-89 has the following method:

```
BABYLON.ScreenshotTools.CreateScreenshotUsingRenderTarget(
    engine,
    camera1,
    { width: engine.getRenderWidth(), height: engine.getRenderHeight() },
    (data) => {
        console.log(data);
    }
);
```

The `camera1` variable has the view from the upper box (big box). However, the result, is a screenshot using the last camera from the list of active cameras (`scene.activeCameras`).

Expected output:

![expected-output](https://user-images.githubusercontent.com/1536369/96351920-52453f80-10bf-11eb-8ef3-c278fc70c7e0.png)

Actual output:

![actual-output](https://user-images.githubusercontent.com/1536369/96351922-52ddd600-10bf-11eb-85a9-55ae7aa89294.png)

[BabylonJS Forum thread discussion](https://forum.babylonjs.com/t/issue-when-taking-a-screenshot-with-multi-cameras/14964)

gabrielheming:
Seems that a new commit undone this fix. I'm reopening this issue.

sebavan:
Fix by the latest PRs :-) will be in the next nightly thanks for reporting it

PR #9202, opened 2020-10-17
Fixed screenshot when scene has multi-cameras. Issue #9201

gabrielheming:
Issue #9201


src/Misc/screenshotTools.ts
```diff
@@ -126,12 +126,18 @@ export class ScreenshotTools {
 
         var scene = camera.getScene();
         var previousCamera: Nullable<Camera> = null;
+        let previousActiveCameras: Camera[] = [];
 
         if (scene.activeCamera !== camera) {
             previousCamera = scene.activeCamera;
             scene.activeCamera = camera;
         }
```
aWeirdo, line 134: Doesn't this already handle multi-camera setup?
are we sure the bug is not somewhere else? 
maybe scene.activeCamera is not being respected if there are multiple activeCameras?

gabrielheming: Well, that is a possibility since I haven't looked at it (it is far beyond my knowledge on the BabylonJS core). Also, if some method depends on that behaviour, it could bring some backward incompatibilities.

aWeirdo, line 134: Shouldn't this already handle multi-camera setup?

gabrielheming: Nope, in fact, it is only needed to work with multi-camera setup. That is because `activeCameras` has I higher priority for both. 

This is because of this code:
https://github.com/BabylonJS/Babylon.js/blob/130f91dd9eb5becb2495737013aee5b46111b665/src/scene.ts#L3986

I am testing and sending a refactored code that simplifies the solution

gabrielheming: Shouldn't the call of `scene.render()` necessary? Because:

- if Yes, we should only work with `activeCameras`;
- if No, we should remove the `scene.render()` call.

aWeirdo: works good, of course the most performant method would be to remove the call.
but better wait for core members, shouldn't be more than a couple hours max, 
maybe there's a good reason it's there and has to stay

gabrielheming: In fact, my simplified solution doesn't work calling an hidden camera in single camera scene. My fancy solution solves that: https://playground.babylonjs.com/#Y0CG29#3 
(space bar will capture screenshot)

So, we have two ways. Removing the call of `scene.render()` or cleaning `activeCameras` also.

aWeirdo: as a note, you can actually just move the __scene.activeCamera__ check below __scene.render()__ aswell (in-case it must stay)
https://playground.babylonjs.com/#Y0CG29#4

deltakosh: I kind of like this fix maybe a bit better as it does not create a new temp array

RaananW:
Any update on this? want to resolve the conflicts?

gabrielheming:
Conflict fixed.

On my side, I would like to know about the call of `scene.render()` as it has been [pointed here](https://github.com/BabylonJS/Babylon.js/pull/9202#discussion_r507100227). Even though there is a solution that moves the calling of `scene.render()` before changing the cameras, it still unanswered for me if it is mandatory to call `scene.render()` after changing the camera.

aWeirdo:
Better just keep the scene.render() call & go with this one 
https://playground.babylonjs.com/#Y0CG29#4

Now thinking it's to avoid problems, if the function is called before any rendering has been done, like from PG without delay

deltakosh:
yes @gabrielheming yes this is mandatory to update everything (matrices, viewport, cache, etc..)

src/Misc/screenshotTools.ts
gabrielheming: @deltakosh so, with your answer:
> yes @gabrielheming yes this is mandatory to update everything (matrices, viewport, cache, etc..)

I have to use the temp array, because the render is called after changing the camera.

deltakosh: I mean with @aWeirdo  fix we do not need to store the previousActiveCameras at all

gabrielheming: That's why I asked if it was mandatory call `scene.render()` **after** changing the activeCamera. That solution call `scene.render()` **before**, and does not work calling after.

deltakosh: Oh ok..yeah! calling before is totally fine

gabrielheming: Good. I will update it.

PR #9209, opened 2020-10-20
Add some accessors to the edge renderer

Popov72:
See https://forum.babylonjs.com/t/plane-intersection-to-then-extrude/15051/2

Also, the doc for `enableEdgesRendering` was not ok, fixed that.


RaananW:
LGTM, want to add the missing doc (probably the `@return` comments)

PR #9211, opened 2020-10-20
Thin sprites

sebavan:
Allow sprites use from thin engine with thin texture and add VAO support


deltakosh (approved):
love it
we can merge as soon as the tests are up

PR #9212, opened 2020-10-20
Fix NME dds files used as environment texture

### LINKED ISSUE #9180: Reflection node is rendering an .env differently than a .dds IBL
When loading a precomputed cube map as a .dds file, the preview on the node and the lighting in the 3D preview look correct. When loading the same IBL saved as an .env (generated from the previous .dds file) the lighting in the node preview and in the 3D preview look wrong. 

**Repro**
- Create standard PBR-MR node material and connect a reflection node
- Set the cube map option on the node to true and then load the attached .dds and then replace it with the attached .env
- Expected result: The rendering on the reflection node preview and in the 3D preview would be the same since this .env was made from the .dds
- Current result: The lighting looks incorrect for the .env file.

**Screenshots**
This is how the .env renders:
image

This is how the .dds renders: 
image

Attaching the .dds, .env, and sample shader below:
[reflectionTest.zip](https://github.com/BabylonJS/Babylon.js/files/5381385/reflectionTest.zip)

Popov72:
Fix #9180 

I added a switch to enable/disable gammaSpace for the PBR Reflection block as well as the possibility to load a prefiltered cube texture: there's a specific code path to load prefiltered texture, so the enable/disable gammaSpace switch was not enough (but can still be useful for general purpose, that's why I added it).

How I did it:
image

I don't know if it's the best UI for the prefiltered part... I have shifted "Texture is prefiltered" a bit to show it works with "Load as cube texture"...


deltakosh (approved):

Popov72:
WebGL2 tests failed with this error:
image
I don't know what it is...

deltakosh:
Merging!

PR #9214, opened 2020-10-20
Instances with different materials

breakin:
Hoping to replace this PR with #9606 that solves the same problem.

Status:
* The code in this PR is being polished such that it can be merged.

This PR:
* Clone geometry without creating full new buffers on CPU/GPU
* Enables possibility to have two meshes with different instance buffers but with mostly the same geometry.

The following playground now works and reuses VBOs for all the spheres even though some are instances, some are not and they have different materials.
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    light.intensity = 0.7;
    var sphere = BABYLON.MeshBuilder.CreateSphere("sphere", {diameter: 1, segments: 32}, scene);
    sphere.setEnabled(false)

    const m1 = new BABYLON.PBRMaterial('m1', scene)
    const m2 = new BABYLON.PBRMaterial('m2', scene)
    const m3 = new BABYLON.PBRMaterial('m2', scene)
    m1.metallic = 0.0
    m2.metallic = 0.0
    m3.metallic = 0.0
    m1.albedoColor.set(1.0, 0.0, 0.0)
    m2.albedoColor.set(0.0, 1.0, 0.0)
    m3.albedoColor.set(0.0, 0.0, 1.0)

    var s1 = sphere.clone('s1')
    var s2 = sphere.clone('s2')
    var s3 = sphere.clone('s3')
    s1.material = m1
    s2.material = m2
    s3.material = m3
    s1.cloneGeometry()
    s2.cloneGeometry()

    // If we don't use cloneGeometry then .makeGeometryUnique will be called at createInstance (in 4.2) or wrong result (in 4.1).
    // Using makeGeometryUnique copies geometry and uses a new buffer on both CPU/GPU

    const instances = []

    for (let i=0; i<2; i++) {
        // Instances using material 1
        instances.push(s1.createInstance())
    }
    for (let i=0; i<2; i++) {
        // Instances using material 2
        instances.push(s2.createInstance())
    }
    for (let i=0; i<2; i++) {
        // Not instances at all
        instances.push(s3.clone())
    }

    for (let idx=0; idx<instances.length; idx++) {
        const i = instances[idx]
        i.position.x = idx;
        i.position.y = 0.5;
        i.setEnabled(true);
    }

    var ground = BABYLON.MeshBuilder.CreateGround("ground", {width: 6, height: 6}, scene);
    ground.receiveShadows = true

    // Loose context after 1 second
    setTimeout(() => {
        if (looser === null) {
            looser = engine._gl.getExtension('WEBGL_lose_context');
        }
        looser.loseContext();
    }, 1000)

    // Restore context after 2 seconds
    setTimeout(() => {
        if (looser === null) {
            looser = engine._gl.getExtension('WEBGL_lose_context');
        }
        looser.restoreContext();
    }, 2000)

    return scene;
};
~~~~~~~~~~~~~~~~~~~~~~~~~~~~


src/Meshes/buffer.ts
```diff
@@ -325,6 +325,16 @@ export class VertexBuffer {
         this._instanceDivisor = instanced ? divisor : 0;
     }
 
+    /*
+        Return a clone of this VertexBuffer that references the same data buffer.
+        Data buffers are reference counted so it will only be released once even if we have two VertexBuffers referencing it.
+    */
+    public clone(): VertexBuffer {
+        var useBytes = false; // No idea what to set here
```
deltakosh, line 333: you can simply call the constructor with false as all the parameter are already set up

src/Meshes/geometry.ts
```diff
@@ -45,6 +45,7 @@ export class Geometry implements IGetSetVerticesData {
     public onGeometryUpdated: (geometry: Geometry, kind?: string) => void;
 
     // Private
+    private _borrowed_index_buffer: boolean = false;
```
deltakosh, line 48: I'm afraid that cannot work because what will happen if I dispose the source index buffer?
We need to reference count the index buffers as well if you want to go down this path

breakin: That makes sense. I thought this through in the VertexBuffer case but index buffer is still hacky!

src/Meshes/mesh.ts
```diff
@@ -1436,6 +1436,20 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         return this;
     }
 
+    /**
+     * Creates a un-shared specific occurence of the geometry for the mesh.
+     * @returns the current mesh
+     */
+    public makeGeometrySomewhatUnique(): Mesh {
```
deltakosh, line 1443: I'm not a big fan of that naming.
What about CloneGeometry instead?

src/Meshes/geometry.ts
```diff
@@ -1000,6 +1001,24 @@ export class Geometry implements IGetSetVerticesData {
         return geometry;
     }
 
+    public make_override(id: string): Geometry {
```
deltakosh, line 1004: no need for id
we can get it with "clone" + this.id

I also suggest naming it clone()

src/Meshes/mesh.ts
breakin: Yeah I left it with a bad name to discuss it if we get that far. What it does is that it clones the geometry instead of copying it... and you can can change vertex buffers if you want to, forgetting the old ones.. but they are there until you do... If we do sort out the index buffer then cloneGeometry might be a good name,

src/Meshes/instancedMesh.ts
```diff
@@ -171,6 +171,40 @@ export class InstancedMesh extends AbstractMesh {
...
     }
 
+    /**
+     * Changes the source mesh for this instance
+     * Does not change instance buffers for the instance (assumes both the source meshes uses the same instance buffers)
+     * @see https://doc.babylonjs.com/how_to/how_to_use_instances
+     * @param newSourceMesh defines the new source mesh for the instance
+     * Returns nothing.
+     */
+    public changeSourceMesh(newSourceMesh: Mesh): void {
```
deltakosh, line 181: not sure to get what is the goal of this function
This seems really specific and probably not suited to be in the framework itself (more a client code)

src/Meshes/geometry.ts
breakin: I'll rename it. It is a bit of a weird clone function...!

src/Meshes/instancedMesh.ts
breakin: The idea from my POV is that it is something the user might want, but easy to forget something that needs to happen. By putting it in the API it forces the API to test it before each release (which is a burden).

If we want to put it on the user then we might need to open up some things since this is impossible in typescript (but possible in javascript) from client code.

src/Meshes/geometry.ts
breakin: What if it is cloned many times (as in my example in the PR)? Then clone+this.id will be taken. Not sure if it needs to be unique, there is an uniqueId for that.

src/Meshes/geometry.ts
breakin: It seems that the index buffer is also a DataBuffer and those are referenceCounted, so I might be able to do something similar. Not sure how the clone-function would look though.

src/Meshes/geometry.ts
deltakosh: yes this is supposed to be like that. all the clone will have the same id but different uniqueId

src/Meshes/instancedMesh.ts
deltakosh: yes but so far this code is far from complete and complex to maintain. I would see it in a subsequent PR

breakin: Right! I will keep it for now just for my testing but I agree that it should be moved to another PR and not be part of this discussion.

src/Meshes/geometry.ts
breakin: Ok tried solving this using references on the index_buffer

breakin:
Ok it feels like index_buffer now works more or less as intended but I have two worries:

* For now I do not copy _indices which is probably wrong. Cloning it feels like the wrong move here, maybe a comment is ok
* My big worry is rebuild. Since there are now multiple owners of the databuffer it will be rebuilt multiple times when context is lost (or whenver rebuild is called for everything). Not sure how to resolve this.

deltakosh:
I think indexbuffer should have a clone as well and behave exactly like the vertex buffer

breakin:
> I think indexbuffer should have a clone as well and behave exactly like the vertex buffer

I tried to do it like that but failed. The reason is that there is no IndexBuffer object, only a DataBuffer object. In the case of the VertexBuffer I cloned the VertexBuffer object but retained the DataBuffer-object in it since that is what is needed in order to share the GPU/CPU-data. So I can't clone the DataBuffer object since that is where the referenceCounter lives and we only want one reference counter for the GPU data.

I realize now that the data for rebuilding the VertexBuffer is not cloned and thus the cloned data buffer can't rebuild properly for VertexBuffers. It is the same situation with _indices in the Geometry. It is not cloned and thus rebuild will not work for the Geometry._indexBuffer.

That being said we don't want to rebuild the DataBuffers more than once so my PR has a problem there... :(

deltakosh:
>> no IndexBuffer object

Right..this is an excellent point. So yeah I think your reference count is enough then

For the rebuild, I do not follow you on the problem for VertexBuffer. If we clone it the way you propose it, the same buffer will be shared between clones so the problem is mostly about rebuilding it multiple times. But this is already an issue. Not something you are adding. 

For the index buffer, you are right after a context lost, the index buffer will not be cloned anymore

I think we can fix both with the following idea: instead of having the geometry to call the rebuild, we could simply add an observer to engine.onContextLost for each buffer, the buffer itself could then rebuild itself only once. This could fix both VB and IB issue

breakin:
@deltakosh: I will try to resolve this tomorrow!

deltakosh:
You rock buddy!

breakin:
> You rock buddy!

Thanks! I realized that there is no way that this will be 4.2 material due to how big it is and how little time there is left so I am slowing down my speed a little but I will try work more on it next week.

breakin:
@deltakosh I need to keep working on this but this is what I have now. I felt I needed to move index buffer handling into an object so it could be shared in many geometries. Rebuild should work for both index/vertex buffers in theory. Most playgrounds seems to work and tests locally as well but the CI hates me so need to figure out why that is.

I realize that you probably have your hands full with the 4.2 release but would love to hear if this PR is going in the right direction or not.

src/Meshes/buffer.ts
```diff
@@ -1,6 +1,159 @@
...
+    }
+
+    /**
+     * Update content of the buffer
+     * @param indices new indices to use
+     * @param offset optional offset in the target buffer
+     * @param gpuMemoryOnly specifies if the CPU memory should be updated
+     * @returns true if size changed
+     */
+    public update(indices: IndicesArray, offset?: number, gpuMemoryOnly = false): boolean {
```
deltakosh, line 115: I'm wondering if we need a new class. Can we inherit from vertexbuffer and just change what is different? There is a lot of duplicated code

Other option: have a new parent class for vb and ib?

deltakosh:
This looks pretty solid to me! Thanks!

as soon as 4.2 will be out we can think about merging it!

src/Meshes/buffer.ts
breakin: I agree.

There is one more weird thing. I feel that the Buffer-class should be a parent-class to the VertexBuffer but instead the VertexBuffer HAS a Buffer. I can't think of a reason for this except that clones of VertexBuffers always should have the same buffer even if the internal DataBuffer is recreated. But there was no clone-function before...

My IndexBuffer class has a more direct relation to the DataBuffer (which is why the clone-function is called makeNewReference instead of clone).

deltakosh: You can create a new VB and use a shared buffer (through constructor)

deltakosh:
OK ALL GOOD now that 4.2 is out
do you mind fixing the checks?

breakin:
> OK ALL GOOD now that 4.2 is out
> do you mind fixing the checks?

Haha good work! Congratulations on the release! Many new exciting things!

I had some fun on my own with this PR and managed to dig myself into a hole. Hopefully next week I can dig myself out of it again to have something new for you to react to :) Trying to make it so that IndexBuffer and VertexBuffer are more similar and act in the same way (like you suggested).

breakin:
@deltakosh: OK puh. I took the time to rewrite the PR such that IndexBuffer works exatly like VertexBuffer. I feel like I wanted to share this half-way progress to make sure I am on the right track. If I get an OK I will test and check all details of the updates such that I haven't changed anything.

NOTE: I gave the two new buffers extremely silly temporary names such that we nobody gets attached to the new temporary names.

To break no existing code do the following renamed in my PR:
* Buffer->BufferBase
* VertexBufferBuffer->Buffer
* IndexBufferBuffer -> ?

I guess this makes the most sense but we need a good name for IndexBufferBuffer.

Optimal (from perspective of vanilla babylon)
* Buffer is new with a new meaning
* Buffer->VertexBuffer
* IndexBuffer (called IndexBufferBuffer in my PR)
* IndexBuffer -> IndexBufferView or IndexBufferAccessor
* VertexBuffer -> VertexBufferView or VertexBufferAccessor

The views hold a reference count and ownership over the buffers.

breakin:
Also this PR is now based on a recent master, post 4.2.

dist/preview release/what's new.md
```diff
@@ -2,9 +2,124 @@
 
 ## Major updates
 
+- Added particle editor to the Inspector ([Deltakosh](https://github.com/deltakosh))
```
deltakosh, line 5: That is the old what's new, you should get the latest one (far shorter ;))

deltakosh:
I feel like I'm missing something obvious 👍 
Why not keeping the buffer as it is and have two children? One for VertexBuffer and one for IndexBuffer ?

Why do we need to change that? It is already working for VertexBuffer and the only thing I thought we would have to change was to add a new IndexBuffer class?

breakin:
> I feel like I'm missing something obvious 👍
> Why not keeping the buffer as it is and have two children? One for VertexBuffer and one for IndexBuffer ?
> 
> Why do we need to change that? It is already working for VertexBuffer and the only thing I thought we would have to change was to add a new IndexBuffer class?

It might be me who is missing something :) If we look at how Vertex Buffers was handled before we had:
* DataBuffer: The WebGL-buffer...  might not be created yet if we have postponed creation
* Buffer: This owns the DataBuffer and represents a physical VertexBuffer
* VertexBuffer: This is really ownership over a Buffer that can be shared. It also has an offset so it can bee seen as a view into the Buffer

Now the physical IndexBuffer and VertexBuffer are quite different. They both hold a DataBuffer but they need to use different operations on them. That is why there is a base-class now with the shared logic and then two classes that inherits. If you look at them you can see that they are quite different.

Then we have two classes that represents ownership/references/views of those. 

Another complicating factor here is that VertexBuffer does not inherit Buffer (in the previous design) so I can't just have a shared Buffer and then implement the differences in VertexBuffer/IndexBuffer since VertexBuffer/IndexBuffer doens't implement Buffer (it just has one).

Maybe I should draw a picture! I am not sure if all of these complications needs to exists but it seems like we at least need a way to create an offsetted view into a (Vertex)Buffer and for clone to work we an object that the clones can share.

deltakosh:
Well my understanding was that we wanted to only create a IndexBuffer class that works like the VertexBuffer. Why do we need more? (sorry iwas playing World of warcraft 24/7 last week so I may be disconnected :))

Index or vertex buffer are just views (technically at the metal layer) on the buffer right?

breakin:
Well hm... I think the challenge is that we want the following:
* Index Buffer shareable between many geometries
* Index Buffer restore itself when context lost (so much own CPU-data)

In order to have it shareable we either need all the geometries to have the same object OR the IndexBuffer need to have some internal object that is shared between all the clones. The DataBuffer-object is not enough to be shared since if we do get a context lost we loose it and if we recreate it our clones no longer know about each other.

My first approach was to let all the geometries have the same IndexBuffer and clone only increased the refCount... But then you suggested I should do it like vertex buffer and now I though I had but maybe you had something else in mind...!

deltakosh:
Yes I do agree with the goal. But my point was (if I recall correctly) that the buffer will handle the recreation on context lost. so vertexbuffer and indexbuffer will keep pointing to the same buffer and that buffer will manage to handle context lost and recreate its own data

breakin:
@deltakosh: Yes... but it takes different engine-functions to recreate an index buffer or an vertex buffer... or at least to upload the data again... and those differences is what I tried to put into VertexBufferBuffer and IndexBufferBuffer that share what they can from Buffer...

deltakosh:
Well, the buffer in the vb constructor is created like that:
`
this._buffer = new Buffer(engine, data, updatable, stride, postponeInternalCreation, instanced, useBytes);
`

so the buffer itself knows everything it has to know to recreate itself right?

If you look at the _rebuild of the VB class it simply call rebuild on the buffer right?

So we should simply remove the _rebuild from the VB, let the buffer register to context restored and recreates itself, add a new class for IndexBuffer which is almost the same as vb

deltakosh:
(again please do not hesitate if I miss something obvious)

breakin:
Yes that works for the VertexBuffer+Buffer+DataBuffer in the existing case.
I needed to add
IndexBuffer+Buffer-for-Index-Buffer+DataBuffer
to mirror to that for IndexBuffers.

So with my design there is no problem, but maybe too many classes and absolutely the wrong names.

Bit tired over here so gotta go to bed, I might be thinking about this the wrong way too...

deltakosh:
I would like to keep everything almost unchanged so basically I would like to maybe add a flag or an enum to the Buffer class so basically we know we need to call engine.createVertexBuffer or engine.createIndexbuffer

I would like to get:
(vertexbuffer / Index Buffer) -> Buffer (which contains a databuffer)

So no change on vertexbuffer, tiny changes on buffer to have a "mode" and a new class used by geometries to reference indexBUffer (instead of pointing directly to buffers)
Plus: moving the context restore to buffer to avoid restoring the buffers many times (which is already the case so more a bug fix :D))

breakin:
@deltakosh: Right let me give it another try!

breakin:
@deltakosh: An I think that the restore-on-context is fine now, it is just that once multiple geometries share a vertex buffer we are in trouble.. today the geometry is responsible for triggering the rebuild as far as I understand it

deltakosh:
well if the buffer is responsible for its own rebuild there is no problem right? geometries will not call the rebuild anymore

breakin:
@deltakosh: I'm still just trying to put it together. Now it is one class (Buffer) but I am still trying to make data be able to be both IndicesArray or DataArray (depending on type of buffer). Now going super-well but will try some more :)

src/Meshes/buffer.ts
```diff
@@ -1,19 +1,32 @@
-import { Nullable, DataArray } from "../types";
+import { Nullable, DataArray, IndicesArray } from "../types";
 import { ThinEngine } from "../Engines/thinEngine";
 import { DataBuffer } from './dataBuffer';
+import { Observer } from '../Misc/observable';
+
+// TODO: Try to retain old Buffer constructor
+// QUESTION: We have instanced and divisor on both VertexBuffer and Buffer... I guess different VertexBuffers could have different divisors as long as they are not views into the same buffer and used at the same time... but what do we want?
```
deltakosh, line 7: yes this is what we want as we want to cover two kinds of vertex buffers:
- a vertex buffer can simply be a full buffer
- but it can also be just part of a buffer (interleaved)

src/Meshes/buffer.ts
```diff
@@ -1,19 +1,32 @@
...
 
 /**
  * Class used to store data that will be store in GPU memory
+ * Can represent a Vertex Buffer or an Index Buffer
  */
 export class Buffer {
-    private _engine: ThinEngine;
-    private _buffer: Nullable<DataBuffer>;
+    protected _engine: ThinEngine;
+    protected _buffer: Nullable<DataBuffer>;
```
deltakosh, line 20: I guess I would say:
protected _buffer: Nullable<DataBuffer | IndicesArray>;

src/Meshes/buffer.ts
```diff
@@ -30,45 +43,74 @@ export class Buffer {
...
-     * @param useBytes defines if the offset and stride are in bytes     *
-     * @param divisor sets an optional divisor for instances (1 by default)
-     * @returns the new vertex buffer
+     * Store data into the buffer. If the buffer was already used it will be either recreated or updated depending on isUpdatable property
+     * @param data defines the data to store
      */
-    public createVertexBuffer(kind: string, offset: number, size: number, stride?: number, instanced?: boolean, useBytes = false, divisor?: number): VertexBuffer {
-        const byteOffset = useBytes ? offset : offset * Float32Array.BYTES_PER_ELEMENT;
-        const byteStride = stride ? (useBytes ? stride : stride * Float32Array.BYTES_PER_ELEMENT) : this.byteStride;
+    public create(data: Nullable<DataArray> = null): void {
```
deltakosh, line 79: same here:
`data: Nullable<DataArray | IndicesArray>`

src/Meshes/buffer.ts
```diff
@@ -89,6 +131,11 @@ export class Buffer {
         return this._data;
     }
 
+    /** Set current buffer's data */
+    public setData(data: IndicesArray | DataArray | null): void {
```
deltakosh, line 135: not sure we want that honestly

breakin:
@deltakosh: Ok now it at least runs again. I'm sure there are some bugs left and I need to test it more, especially loosing context. But if we forget about that I want to make sure that we get this right. I'm not super-happy about the as IndicesArray and as DataArray I got in there. I have not bee able to detect that setData/set data doesn't take something that is of the wrong format since some of the parts of those are so hard to detect...

I can make it such that getData/data on the VertexBuffer/IndexBuffer works with the right type so you get type-checking when using those at least. I think that would make it much more type-safe to work with it then..

Do you feel that the larger parts of this PR is ok now and I should just polish, or is it still in the wrong direction? I almost feel good about it now... but in need of polish of course...!

src/Meshes/buffer.ts
deltakosh: Did it answer your question?

src/Meshes/buffer.ts
```diff
@@ -30,45 +43,74 @@ export class Buffer {
      * @param instanced whether the buffer is instanced (optional)
      * @param useBytes set to true if the stride in in bytes (optional)
      * @param divisor sets an optional divisor for instances (1 by default)
+     * @param bufferType the type of the buffer (VertexBuffer or IndexBuffer) (default VertexBuffer)
      */
-    constructor(engine: any, data: DataArray, updatable: boolean, stride = 0, postponeInternalCreation = false, instanced = false, useBytes = false, divisor?: number) {
+    constructor(engine: any, data: DataArray, updatable: boolean, stride = 0, postponeInternalCreation = false, instanced?: boolean, divisior?: number, useBytes = false, bufferType: BufferType = BufferType.VertexBuffer) {
```
deltakosh, line 48: we cannot change constructor (for backward compat reasons)

deltakosh:
I feel good as well on the general direction. I added some comments and I'm not sure if you saw them though

But overall it seems to work for me

src/Meshes/buffer.ts
breakin: Yes. Thanks for reminding me! Divisor is per attribute and not per buffer, makes sense! I had forgotten about that.

src/Layers/effectLayer.ts
```diff
@@ -817,12 +817,6 @@ export abstract class EffectLayer {
      * @hidden Internal use only.
      */
     public _rebuild(): void {
-        let vb = this._vertexBuffers[VertexBuffer.PositionKind];
```
breakin, line 820: It is not strictly required for me to remove these. I could add a dummy _rebuild-function in VertexBuffer that does nothing. If possible it could be marked as deprecated.

deltakosh: No I like it, less code!

breakin:
@deltakosh: The issue with IndicesArray vs DataArray feels much better now, almost like a non-issue. I have a few todos in my code that relates to offsets that aren't always respected and some confusion around that.

There was one higher-level question though.

Today Many meshes share a geometry. If something happens to that geometry all meshes are notified (bounds changed, vbos changed I guess). This invalidates VAO and does other stuff.

All changes to VertexBuffers are initiated via the Geometry so this works.

Now we can have many geometris that share VBs/IBs. If one of them modify the Buffer such that it is recreated the VAOs will be corrupted on the other geometries and there is right now no way to invalidate them. Also the Geometry might have the wrong idea about the size of the buffer.

The most natural way would be for VBs/IBs to notify their users when they change. Often this will be one Geometry. With cloning it will be one Geometry per clone. So not one per mesh. I don't find a nice way around this except turning all VBs/IBs immutable when they are cloned.

Not sure how this would look but sounds like something we probably need. If we do get that in place then it will actually be quite robust, being able to have an animated object with many clones, all being synched and updating their bounds etc..

deltakosh:
I agree on the notification. The buffer should notify the geometry to invalidate their VAO 

That sounds like a good plan!

src/Engines/thinEngine.ts
```diff
@@ -675,6 +676,7 @@ export class ThinEngine {
                         // Cache
                         this.wipeCaches(true);
                         Logger.Warn("WebGL context successfully restored.");
+                        this.onContextRestoreBuffersObservable.notifyObservers(this);
```
deltakosh, line 679: Why not using onContextRestoredObservable ?

breakin: This is just a thought I had. If we move when they are created from a well-defined point before onContextRestoredObservable to onContextRestoredObservable then rebuilds of buffers will happen randomly intercepted with whatever the user might want to do at a rebuild. I was thinking that if the user did something that required the inner buffers to actually exist, or if she did something that forced a premature/extra rebuild of the buffers. That could cause confusion and maybe bugs.

I'm thinking that thinEngine could have an array of all buffers and just loop and update them (if it supports lost context). That is probably more efficient compared to having an observer at all.. But I could be wrong on that point.

If is very easy to move everything over to onContextRestoredObservable again of course if we feel that it is sufficient.

breakin:
@deltakosh: I am afraid that I can't really test context lost since it seems to fail without my changes. I added some code to drop/restore context in my playground (see description) but doing similar things in the standard babylon playground results in issues. I could be wrong... I will turn my attention no to the signaling which for now seems like the last issue with this PR...

breakin:
So taking a brief pause cause there is so much code happening now. My current thoughts are these:

Mesh>Geometry>VertexBuffer/IndexBuffer>Buffer>WebGLBuffer
Buffer can change and we need to notify from Buffer to Geometry. VertexBuffer/IndexBuffer are not shared between geometries (we instead create a new one using the same Buffer). That means that we have each buffer can be used by many vertex buffers, but a vertex buffer in general only have own owner.

I think I make it so that the Buffer-object knows about all the VertexBuffer/IndexBuffer that uses it. It can then notify them all. They have ONE (optional) geometry-owner that they can notify.

breakin:
TBH this feels very messy. For my personal usercase I would be ok with immutable shared buffers. Maybe I should try that first and then someone else can pick it up later and make it mutable?

breakin:
Not that an alterantive to this PR is that the mesh can have an override VAO and also own its own instance buffers. That way they can all share a Geometry object... If we want mutable geometries I am suddenly leaning on that being cleaner but I haven't actually implemented it...

deltakosh:
> TBH this feels very messy. For my personal usercase I would be ok with immutable shared buffers. Maybe I should try that first and then someone else can pick it up later and make it mutable?

Yes I agree ths PR starts to be too complex and will need to be splitted. So totally aligned

deltakosh:
Let's do it steps by steps :)

src/Engines/thinEngine.ts
deltakosh: I don't think this is an issue. Let's not change too many things at once ;)
I would like to merge it asap because we have too many moving pieces

Issue #9206, opened 2020-10-19
Mesh rotation in inspector breaks when changing sliders, particularly on X axis


PatrickRyanMS:
**Repro**
- Load a mesh into the sandbox
- select the mesh root and open the inspector
- use the rotation sliders to change the mesh rotation one at a time on X, Y, and Z and slide the value from 0-360 using euler angles. 
- Expected result: you can set any angle on an axis without incurring a change on the other axes. You can also smoothly rotate the mesh 360 degrees in any axis
- Current result: For Y and Z the mesh rotates fairly smoothly, but there are a couple of points where the values seem to jump. On X once you hit around 90 degrees, the Y and Z axes start jumping around and the X axis actually reduces in value as the slider continues to move to the right. The mesh also seems to jump to different rotations on incremental changes in the X rotation making the mesh feel as though it's spinning randomly.

**Video**
Repro of issue attached below:
[rotationBug.zip](https://github.com/BabylonJS/Babylon.js/files/5404192/rotationBug.zip)

deltakosh:
I'm afraid this is inherent to how we compute the eulers from the rotationQuaternion. This is the problem with the UI hiding the quaternion. We are constanly converting from euler to quaternion and at some point (like 90) we can have multiple solutions hence the jump.

Adding @RaananW  to get a second thought

RaananW:
I guess an input field instead of the slider will make it feel a little less buggy. Showing quaternion axes make no sense, and the only other solution would be to actually show a box with orientation texture on it.

deltakosh:
@PatrickRyanMS ok for an input field?

msDestiny14:
Yep. Was on my list since I already did the work for NME! 👍

PatrickRyanMS:
I've been thinking about it and while a slider is much nicer for quickly rotating an object, we do have the rotation gizmo that duplicates that behavior. So if we need to go to an input field, I think that is ok since we still have a backup path to do interactive rotation.

RaananW:
Another suggestion (very unpopular, i guess :-) ) would be - whenever someone changes the input, revert back to euler instead of quaternion in code. Instead of constantly converting, convert once and store it in code.

The only issue i see is that the user will still think that quaternions are being used. What we could do is - on component unmount, convert back to quaternions. this might change the euler values if the user goes back to this tab, but will allow using the sliders correctly.

PatrickRyanMS:
[...]

PR #9215, opened 2020-10-20
Adding Input to Sliders in Inspector

msDestiny14:
slider

### DIFF
--- a/inspector/src/components/actionTabs/actionTabs.scss
+++ b/inspector/src/components/actionTabs/actionTabs.scss
@@ -972,10 +972,13 @@ $line-padding-left: 2px;
                 }
 
                 .sliderLine {
-                    padding-left: 2px;
+                    padding-left: $line-padding-left;
                     height: 30px;
                     display: grid;
-                    grid-template-columns: 1fr auto;
+                    grid-template-rows: 100%;
+                    grid-template-columns: 1fr 40px;
+            
+
 
                     .label {
                         grid-column: 1;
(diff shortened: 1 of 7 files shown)


deltakosh:
I feel like we are missing some margin on the left of x,y and z

msDestiny14:
> I feel like we are missing some margin on the left of x,y and z

I would agree but that affects all the labels using slider. Those are all currently aligned. Should I make a special flag?

slider

off

deltakosh:
yes probably ;)

msDestiny14:
Added the margins!

PatrickRyanMS:
Am I mistaken or would having the sliders still there run into the same issue we saw before when converting between quaternion and euler? If so, we should probably remove the sliders as the UX is fairly broken using the sliders and we should encourage the use of the rotation gizmo instead.

msDestiny14:
> Am I mistaken or would having the sliders still there run into the same issue we saw before when converting between quaternion and euler? If so, we should probably remove the sliders as the UX is fairly broken using the sliders and we should encourage the use of the rotation gizmo instead.

That is up to you! Note this PR may not be the final solution to that specific issue (I did not link it for that reason). This was to be consistent with NME, which got input added to sliders last week in another PR.

deltakosh:
Well I still use the slider a lot. so maybe we can do this:
only display the slider when we do NOT have a rotationQuaternion?

PatrickRyanMS:
That could be a solution though I worry a bit about the UX changing without the user understanding why. We should take a look at it and see how it feels.

RaananW:
I have commented on the issue connected to this (this - https://github.com/BabylonJS/Babylon.js/issues/9206#issuecomment-713436564), maybe this would be an acceptable solution?

deltakosh:
>> That could be a solution though I worry a bit about the UX changing without the user understanding why. We should take a look at it and see how it feels.

This could be associated with an info indicating (like a label) that the current rotation is using a quaternion instead of euler

msDestiny14:
Close for the moment since I noticed a weird glitch that happens on float inputs with values sliders between 0-1. Will investigate,

msDestiny14:
Also including fix for https://github.com/BabylonJS/Babylon.js/issues/9206

rotation

PR #9221, opened 2020-10-21
Make it possible to move an instance from one source mesh to another

breakin:
Afaict the animations doesn't explicitly mention the source node in any form so it should be saft to leave them as is.

The use-case I see is that we have two source meshes that are the same except for having different materials (or some other per-sourceMesh property). Now we want to change material on one instance and then we change the source mesh.

Possibly this entire method could be replaced with some other general methods on instanceMesh such that the client can move it themselves. This includes:
* Changing soruceMesh attribute (in typescript)

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -57,6 +57,7 @@
 - Removed all references to HTML element from cameras' attach and detach control functions ([RaananW](https://github.com/RaananW))
 - Added `boundingBoxRenderer.onResourcesReadyObservable` ([aWeirdo](https://github.com/aWeirdo))
 - Added `copyTools.GenerateBase64StringFromTexture` ([aWeirdo](https://github.com/aWeirdo))
+- Added `instancedMesh.changeSourceMesh` ([breakin](https://github.com/breakin))
 
 ### Engine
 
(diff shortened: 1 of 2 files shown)


deltakosh:
I still do not think this is valuable in the framework honestly. This seems very niche for me

breakin:
Let me just describe what I am doing to explain why someone would need this feature.

You have a lot of instances but want to be able to change material.
Babylon doesn't solve this. It could but it doesn't.
So you build a system on the outside where you change the sourceMesh such that properties on the instanceMesh (such as material, recieveShadows, ...) choose the appropriate sourceMesh.
Each such sourceMesh needs its own instance buffers (for matrices and other attributes).

So I need the ability to do a light clone of a Mesh so that I can have multiple source meshes (#9214) and then I need to move the instance between the source meshes easily.

The part I would need at least is to make the sourceMesh attribute not being read-only or I can't do this in pure typescript. Right now I have (in client code):

@ts-ignore
instance._sourceMesh = newSource;

Which I guess is ok. But if someone else wants to build a system like the one I am building then they might find this method nice!

deltakosh:
I understand your need but this is too niche for your needs only. I will close this PR if you don't mind but will gladly merge the other one

breakin:
> 
> 
> I understand your need but this is too niche for your needs only. I will close this PR if you don't mind but will gladly merge the other one

No problem, I can make a workaround client-side for this, or change manually in my copy of babylon. No Problem!

PR #9223, opened 2020-10-21
Clip multilines by camera near plane when point is attached to a mesh

CedricGuillemet:
Follow up on this thread: https://forum.babylonjs.com/t/multiline-weird-behaviour/15020/2
multline point attached on a mesh are not clipped by the near plane. They appear flipped verticaly when behind the camera. So, I've added the z in the point value and clip the line if both extremities are not in front of the near plane.

### DIFF
--- a/gui/src/2D/controls/multiLine.ts
+++ b/gui/src/2D/controls/multiLine.ts
@@ -5,6 +5,7 @@ import { Control } from "./control";
 import { MultiLinePoint } from "../multiLinePoint";
 import { Measure } from "../measure";
 import { _TypeStore } from 'babylonjs/Misc/typeStore';
+import { Vector3 } from 'babylonjs';
 
 /**
  * Class used to create multi line control
@@ -193,6 +194,7 @@ export class MultiLine extends Control {
         context.beginPath();
 
         var first: boolean = true; //first index is not necessarily 0
+        var previousPoint: Vector3;
 
         this._points.forEach((point) => {
             if (!point) {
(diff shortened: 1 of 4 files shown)


gui/src/2D/controls/multiLine.ts
```diff
@@ -205,8 +207,13 @@ export class MultiLine extends Control {
                 first = false;
             }
             else {
-                context.lineTo(point._point.x, point._point.y);
+                if (point._point.z < 1 && previousPoint.z < 1) {
```
Popov72, line 210: Shouldn't you also test that z is >= 0?

If the point is behind the camera, z will be < 0 (I think z is the screen depth?)

Also, I think that when a point is visible but not the other, the line will disappear completely? [...] I have just tested, it's what's happening.

I guess the right way would be to clip the line "by hand" in homogenous space against the view frustum...

CedricGuillemet: nop because we are in clip space and not in view space.
Yes, the right way to do it is to clip the lines in world space with the near plane. But I don't want to go down that route 2 weeks before release as it needs a bigger refactor.

sebavan:
LGTM besides the lint issue, Let me know when to merge it @Popov72 and @CedricGuillemet  :-)

gui/src/2D/controls/multiLine.ts
CedricGuillemet: I tried to lerp with the plan Z to get clipped lines. it almost work but I'm not sure I can interpolate with clipspace Z. I prefer to let it that way and see if it's an issue later.

Popov72:
That's ok for me!

sebavan (approved):

PR #9224, opened 2020-10-21
Fix issue(s) with bone map

RaananW:
Using the bones array is not enough - some bones have a negative index and should not be displayed.

This fixes two issues - the name of the bones were displayed incorrectly (because of a difference between index in the array and the actualy index of the bone), and the length of the sliders was fixed (now according to the length of active bones).


sebavan (approved):

sebavan:
@RaananW LGTM feel free to merge when all good on your side.

PR #9234, opened 2020-10-21
Update meshSimplification.ts

aWeirdo:
Would it be ok to add a check if setting.distance exists?

I'd like to use the function without LOD :)


sebavan (changes requested):
could you change `void 0` to the usual checks we do in the rest of the code ??? Some typescript ppl might be less confortable with this notation

aWeirdo:
problem is it can be 0, 
then it'll have to be typeof() === 'undefined' check to avoid false negative

sebavan:
Why not simply settings.distance !== undefined ?

aWeirdo:
you're right :)
done

sebavan (approved):

Issue #9220, opened 2020-10-21
NME Not updating preview window


msDestiny14:
Under certain circumstances, the NME preview window shows "Please wait loading..." This is then fixed as soon as any property of a node or the node material is updated. 

Confirmed circumstances: 
 -when something is loaded into NME: a new NME file, frame etc.

Unconfirmed circumstances:
-Loading into NME?

sebavan:
@msDestiny14 can we close this one now the PR has been merged ?

PR #9237, opened 2020-10-21
NME - Setting isLoading to false after loading

msDestiny14:
Now the preview window will not be stuck on "Please wait loading.." after you load in a file or frame. https://github.com/BabylonJS/Babylon.js/issues/9220


msDestiny14:
@deltakosh I'm curious what you think of this. Not really sure why it wouldn't be set to false after since the function is not async?

deltakosh (approved):

PR #9240, opened 2020-10-22
Remove epsilon check for PolygonPoints.add

bghgary:
This playground (https://www.babylonjs-playground.com/#MEJFT3#2) is broken because there are two different logic to remove extraneous points in the polygon creation code.

1. https://github.com/BabylonJS/Babylon.js/blob/7fb7f7facaa564bd8cedc88e9ac97c2ca26f6f7d/src/Meshes/polygonMesh.ts#L33-L39
2. https://github.com/BabylonJS/Babylon.js/blob/d1348ad7c27ff8451a13a6229cbba6bcd84e97a7/src/Meshes/Builders/polygonBuilder.ts#L168-L171

The data in the PG has two points at the end that are close with the first point. The first logic removes both points. The second logic removes just the last point. This results in the logic creating extra indices that don't fit into the number of vertices.

I removed the check in 1, but I'm not sure this the right fix.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -368,3 +368,4 @@
 - Fix width/height GUI container computation to take into account paddings when `adaptWithToChildren = true` ([Popov72](https://github.com/Popov72))
 - `smoothstep` in NME is now taking any type of parameters for its `value` input. If you use generated code from the NME ("Generate code" button), you may have to move the smoothstep output connection AFTER the input connections ([Popov72](https://github.com/Popov72))
 - `SoundTrack.RemoveSound` and `SoundTrack.AddSound` were renamed to `SoundTrack.removeSound` and `SoundTrack.addSound` ([Deltakosh](https://github.com/deltakosh))
+- `PolygonPoints.add` no longer filters out points that are close to the first point ([bghgary](https://github.com/bghgary))
--- a/src/Meshes/polygonMesh.ts
+++ b/src/Meshes/polygonMesh.ts
@@ -31,11 +31,9 @@ class PolygonPoints {
 
         var result = new Array<IndexedVector2>();
         originalPoints.forEach((point) => {
-            if (result.length === 0 || !point.equalsWithEpsilon(result[0])) {
-                var newPoint = new IndexedVector2(point, this.elements.length);
-                result.push(newPoint);
-                this.elements.push(newPoint);
-            }
+            var newPoint = new IndexedVector2(point, this.elements.length);
+            result.push(newPoint);
+            this.elements.push(newPoint);
         });
 
         return result;


sebavan:
Adding @Popov72 to have a look in the morning.

Popov72:
The problem comes in fact from the point list being cleaned by the code you show in `#1` but the point list being passed to earcut is not cleaned.

The constructor is doing (only with the relevant bits):
```typescript
constructor(name: string, contours: Path2 | Vector2[] | any, scene?: Scene, earcutInjection = earcut) {
    var points: Vector2[];
    if (contours instanceof Path2) {
        points = (<Path2>contours).getPoints();
    } else {
        points = (<Vector2[]>contours);
    }

    this._addToepoint(points);

    this._points.add(points);
```
It's the call to `this._points.add(points)` that will clean the point list. So, in the end, `this._points` may have less points than `this._epoints` (which is the array filled by `this._addToepoint(points)`).

A possible fix:

Change the type taken by `_addToepoint` from `Vector2[]` to `Array<IndexedVector2>`:
```typescript
private _addToepoint(points: Array<IndexedVector2>) {
    ...
}
```

Update the constructor:
```typescript
constructor(name: string, contours: Path2 | Vector2[] | any, scene?: Scene, earcutInjection = earcut) {
    this.bjsEarcut = earcutInjection;
    this._name = name;
    this._scene = scene || Engine.LastCreatedScene;

    var points: Vector2[];
    if (contours instanceof Path2) {
        points = (<Path2>contours).getPoints();
    } else {
        points = (<Vector2[]>contours);
    }

    this._points.add(points);
    this._outlinepoints.add(points);

    this._addToepoint(this._points.elements);

    if (typeof this.bjsEarcut === 'undefined') {
        Logger.Warn("Earcut was not found, the polygon will not be built.");
    }
}
```
Update the call to `_addToepoint` in `addHole`:
```typescript
addHole(hole: Vector2[]): PolygonMeshBuilder {
    this._points.add(hole);
    var holepoints = new PolygonPoints();
    holepoints.add(hole);
    this._holes.push(holepoints);

    this._eholes.push(this._epoints.length / 2);
    this._addToepoint(holepoints.elements);

    return this;
}
```

bghgary:
Also note that the epsilon used by the two code snippets are different which is quite odd.

Popov72:
I guess it was different people that made both codes.

An epsilon of 0.001 seems quite big to me, however (it is the one you get by default when using `equalsWithEpsilon`)...

Letting the user choose the epsilon they want would be the best thing to do I think.

bghgary:
I should also note that the two algorithms for removing points are a bit different. The first algorithm removes all points that match the first point. The second removes only the last point if it matches the first point.

The first algorithm doesn't make any sense to me. Why would you want to remove all points that match the first point?

Popov72:
I agree it's a bit suspect...

Maybe it's to avoid some degenerate cases, but in that case why only checking against the first point and not all points against all others to avoid having the same point multiple times?

RaananW:
What's the current state of this PR?

Popov72:
I think it can be merged, @bghgary has performed the changes.

sebavan:
Perfect we ll wait for @bghgary and merge right after

bghgary:
I need a review

sebavan (approved):

sebavan:
Feel free to merge when all ok ;-)

sebavan (approved):

PR #9243, opened 2020-10-22
Fix 404 error when reloading textures after a context lost

Popov72:
See https://forum.babylonjs.com/t/many-performance-issues-with-4-2-and-ios-mobile/15098/40

### DIFF
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -2886,6 +2886,10 @@ export class ThinEngine {
             url = this._transformTextureUrl(url);
         }
 
+        if (originalUrl !== url) {
+            texture._originalUrl = originalUrl;
+        }
+
         // establish the file extension, if possible
         const lastDot = url.lastIndexOf('.');
         let extension = forcedExtension ? forcedExtension : (lastDot > -1 ? url.substring(lastDot).toLowerCase() : "");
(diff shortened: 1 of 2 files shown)


sebavan (approved):

src/Engines/thinEngine.ts
```diff
@@ -2886,6 +2886,10 @@ export class ThinEngine {
             url = this._transformTextureUrl(url);
         }
 
+        if (originalUrl !== url) {
```
abogartz, line 2889: Just a thought on this, but if there were a second variable like `transformedUrl`, then subsequent code could just look for `transformedUrl || originalUrl` which might make it easier to reason about this part of the code.

sebavan: sorry I merged before seeing your comment :-( @Popov72  do not hesitate to make another PR if needed... Sorry again

Popov72: Not sure there's a big difference between having `originalUrl / url` and `url / transformedUrl` (if that's what your comment is about)?

abogartz: My thinking is that `url` is being used as a source of truth, but it is being overwritten by the transformed URL, so I was thinking it's better to have unique variables with descriptive names so that there are no side effects, which is what caused this bug to slip by.

PR #9251, opened 2020-10-22
GUI: Make the inputText and inputPassword work with unicode strings

### LINKED ISSUE #9242: Incorrect behaviour of PasswordInput control
The behaviour of PasswordInput control is a little bit strange.

**Repro**
- Bug repro on [playground](https://www.babylonjs-playground.com/#W68283#1)
- Expected result:
* The number of dots should be equal to the number of characters (7, in this example)
* The selection should work correctly. Try to select the dots from the end of the dots to the front.
* If I do such selection in WebXR on Oculus Go, FPS rate drops by 15 frames per second.

**Screenshots**
chrome_5FpRuv4GED

**Desktop:**
 - OS: MS Windows 10 Pro x64
 - Browser: Chrome 86.0.4240.111 (Official Build) (64-bit)

**Oculus Go:**
 - Device: Oculus Go
 - Browser stock browser, Oculus Browser 10.11.0.0.0.238890172, Chromium 84.0.4147.125

Popov72:
Fix #9242


sebavan:
Thanks a lot

gui/src/2D/controls/inputText.ts
```diff
@@ -11,11 +11,102 @@ import { VirtualKeyboard } from "./virtualKeyboard";
 import { _TypeStore } from 'babylonjs/Misc/typeStore';
 import { Measure } from '../measure';
 
+/** @hidden */
+export class TextWrapper {
```
deltakosh, line 15: can we just make it its own file?

sebavan (approved):
Just wondering how this can impact perf ???

deltakosh (approved):
glad you did it ;)

gui/src/2D/controls/inputText.ts
Popov72: Done

Popov72:
> Just wondering how this can impact perf ???

Did not see any slowdown in my testing compared to the previous version.

Popov72:
I think some additional testing would be nice (on mobiles too): I did some on my end, but a lot of things has changed.

sebavan:
I ll merge and make a nightly when the build is green and would be cool if you prepare a test page so that we can all test tomorrow ???

Popov72:
Here's the PG: https://www.babylonjs-playground.com/#W68283#2

And you should replace the emoji line by:

```javascript
const emoji = "😊a😉b😏c😘d😍e😇f🤑g";
```

as the PG does not save the code in utf8/unicode.

PR #9259, opened 2020-10-23
Fix plane removal callback event

chrisfromwork:
It looks like an index offset isn't being calculated correctly. Without this change, undefined plane objects are obtained and handed to subscribers to onPlaneRemovedObservable. This change fixes the index offset.


Alex-MSFT (approved):

RaananW (approved):
Thanks @chrisfromwork !

PR #9294, opened 2020-10-28
Add increment to materials with duplicate id

### LINKED ISSUE #9290: Node Materials created in sandbox share id
See https://forum.babylonjs.com/t/sandbox-material-binding-problem/14951/9

**Repro**
Load 3 Cubes.glb into the babylon sandbox
[3Cubes.zip](https://github.com/BabylonJS/Babylon.js/files/5447350/3Cubes.zip)

Assign each cube a different node material created in sandbox:
image

The resulting exported scene will have all the meshes assigned to the same node material:
[scene.zip](https://github.com/BabylonJS/Babylon.js/files/5447415/scene.zip)

Fix should probably be to append a unique identifier to the material's id to avoid ambiguity.

Drigax:
This PR adds some extra logic to our material creation logic, adding a subscript to any materials created with a duplicate id.

Fixes #9290

### DIFF
--- a/src/Materials/material.ts
+++ b/src/Materials/material.ts
@@ -661,6 +661,10 @@ export class Material implements IAnimatable {
     constructor(name: string, scene: Scene, doNotAdd?: boolean) {
         this.name = name;
         this.id = name || Tools.RandomId();
+        let idSubscript = 0;
+        while (scene.getMaterialByID(this.id) != null) {
+            this.id = name + "_" + idSubscript;
+        }
 
         this._scene = scene || EngineStore.LastCreatedScene;
         this.uniqueId = this._scene.getUniqueId();
--- a/src/Materials/material.ts
+++ b/src/Materials/material.ts
@@ -660,13 +660,14 @@ export class Material implements IAnimatable {
      */
     constructor(name: string, scene: Scene, doNotAdd?: boolean) {
         this.name = name;
-        this.id = name || Tools.RandomId();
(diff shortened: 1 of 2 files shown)


Drigax:
Adding some extra eyes, not sure if this is a breaking change or not, I mostly wanted to follow some of the patterns you see in game engines and DCC tools that actually require unique node naming/ids

bghgary:
Is there a specific issue this is addressing?

src/Materials/material.ts
```diff
@@ -661,6 +661,10 @@ export class Material implements IAnimatable {
     constructor(name: string, scene: Scene, doNotAdd?: boolean) {
         this.name = name;
         this.id = name || Tools.RandomId();
+        let idSubscript = 0;
+        while (scene.getMaterialByID(this.id) != null) {
+            this.id = name + "_" + idSubscript;
```
Popov72: You probably wanted to do `this.id = name + "_" + (idSubscript++);`

Also you may use `!==` instead of `!=`.

deltakosh (approved):

Drigax:
@bghgary linked the issue. Basically this comes into play when creating materials via sandbox and how it affects the resulting serialized scene.

src/Materials/material.ts
bghgary: I would recommend removing `!= null` completely.

bghgary (approved):

PR #9303, opened 2020-10-29
max scale for gizmo

CedricGuillemet:
follow up for https://forum.babylonjs.com/t/scale-gizmo-goes-nuts/15248

compute scale component if clamp value if > max

### DIFF
--- a/src/Gizmos/axisScaleGizmo.ts
+++ b/src/Gizmos/axisScaleGizmo.ts
@@ -150,7 +150,7 @@ export class AxisScaleGizmo extends Gizmo {
                 var newScaledMatrix = scalingMatrix.multiply(this.attachedNode.getWorldMatrix());
                 var tempVector = new Vector3();
                 newScaledMatrix.decompose(tempVector);
-                
+
                 let maxScale = 100000;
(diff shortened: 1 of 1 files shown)


src/Gizmos/axisScaleGizmo.ts
```diff
@@ -146,7 +146,15 @@ export class AxisScaleGizmo extends Gizmo {
 
                 const scalingMatrix = new Matrix();
                 Matrix.ScalingToRef(1 + tmpVector.x, 1 + tmpVector.y, 1 + tmpVector.z, scalingMatrix);
-                this.attachedNode.getWorldMatrix().copyFrom(scalingMatrix.multiply(this.attachedNode.getWorldMatrix()));
+
+                var newScaledMatrix = scalingMatrix.multiply(this.attachedNode.getWorldMatrix());
```
RaananW, line 150: Maybe multiplyToRef to a temp matrix instead? This seems to be executed onDrag, which means on each frame when the user drags - This matrix will be GCed after each frame).

CedricGuillemet: I'm fixing that !

PR #9307, opened 2020-10-29
Enable WebXRMeshDetector and add additional native components to XRPlane

chrisfromwork:
This review contains the following changes:
1. webxr.nativeextensions.d.ts is updated to allow for obtaining additional real world geometry information
    * XRPlane is extended to have optional fields for complex geometry ids and geometry types.
    * XRMesh is introduced as a type for observing real world geometry meshes
    * XRSession is extended to allow for enabling mesh detection
    * XRFrame is updated to contain a set of detected real world geometry meshes
1. webxr.native.extensions.d.ts is updated to allow for customization of real world geometry detection behaviors
    * XRDetectionBoundary is introduced as a mechanism for specifying the region in the physical world to use to detect real world geometries
    * XRSession is extended to allow for setting a preferred detection region for plane detection
    * XRSession is extended to allow for setting a preferred detection region for mesh detection
1. WebXRMeshDetector is introduced as the xr feature for detecting real world geometry meshes
1. WebXRPlaneDetector is updated to apply preferred detection regions

> Additional Notes:
> 1. This change has been validated with the following on HoloLens 2: https://github.com/BabylonJS/BabylonNative/pull/476


src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
-// This file contains native only extensions for WebXR  These APIs are not supported in the browser yet.
+// This file contains native only extensions for WebXR. These APIs are not supported in the browser yet.
 // They are intended for use with either Babylon Native https://github.com/BabylonJS/BabylonNative or
 // Babylon React Native: https://github.com/BabylonJS/BabylonReactNative
 
+// Open question: Do better types exist to use for vectors/quaternions?
```
chrisfromwork, line 5: Open question: Do better types exist to use for vectors/quaternions?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+}
+
+interface XRQuaternion {
+    x: number;
+    y: number;
+    z: number;
+    w: number;
+}
+
+// Open question: Should XRGeometryType be defined or should all types be strings?
```
chrisfromwork, line 19: Open question: Should XRGeometryType be defined or should all types be strings?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+
+interface XRFrustum {
+    position: XRVector;
+    rotation: XRQuaternion;
+    fieldOfView: XRFieldOfView;
+    farDistance: number;
+}
+
+interface XRPlane {
+    // Open question: Should geometry id's and types be declared on XRPlanes or queried through other means?
```
chrisfromwork, line 37: Open question: Should geometry id's and types be declared on XRPlanes or queried through other means?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+    geometryId?: number;
+    geometryType?: XRGeometryType | string;
+}
+
+// Open question: Should XRMesh have a concept of a XRReferenceSpace similar to XRPlane?
```
chrisfromwork, line 42: Open question: Should XRMesh have a concept of a XRReferenceSpace similar to XRPlane?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+interface XRMesh {
+    positions: Float32Array;
+    indices: Uint32Array;
+    normals?: Float32Array;
+    // Open question: Do we need lastChangedTime declared for XRMeshes?
```
chrisfromwork, line 47: Open question: Do we need lastChangedTime declared for XRMeshes?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+    lastChangedTime: number;
+    // Open question: Should geometry id's and types be declared on XRMeshes or queried through other means?
```
chrisfromwork, line 49: Open question: Should geometry id's and types be declared on XRMeshes or queried through other means?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+    trySetPreferredPlaneDetectorOptions(preferredOptions: XRGeometryDetectorOptions): boolean;
+    trySetMeshDetectorEnabled(enabled: boolean): boolean;
+    trySetPreferredMeshDetectorOptions(preferredOptions: XRGeometryDetectorOptions): boolean;
 }
 
+type XRMeshSet = Set<XRMesh>;
+
 interface XRFrame {
     featurePointCloud? : Array<number>;
+    // Open question: Should meshes be declared on the XRFrame or queried for the session?
```
chrisfromwork, line 80: // Open question: Should meshes be declared on the XRFrame or queried for the session?

src/XR/features/WebXRPlaneDetector.ts
```diff
@@ -93,12 +96,11 @@ export class WebXRPlaneDetector extends WebXRAbstractFeature {
...
         this.xrNativeFeatureName = "plane-detection";
+        this._initPlaneDetector = this._initPlaneDetector.bind(this);
         if (this._xrSessionManager.session) {
-            this._init();
+            this._initPlaneDetector(this._xrSessionManager.session);
         } else {
-            this._xrSessionManager.onXRSessionInit.addOnce(() => {
-                this._init();
-            });
+            this._xrSessionManager.onXRSessionInit.addOnce(this._initPlaneDetector);
```
chrisfromwork, line 103: Open Issue: This callback is not fired if the WebXRMeshDetector is enabled prior to the WebXRPlaneDetector when no session is currently running

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+    // Open question: Should geometry id's and types be declared on XRMeshes or queried through other means?
+    geometryId?: number;
+    geometryType?: XRGeometryType | string;
+}
+
+type XRDetectionBoundaryType = "frustum" | "sphere" | "box";
+
+interface XRDetectionBoundary {
+    isStationary?: boolean;
+    type?: XRDetectionBoundaryType | string;
```
ryantrem, line 58: Why do we need the `| string` part here? I thought maybe for additional "unknown" types, but then I'm not sure how that would be useful because if it's unknown we wouldn't know what to do with it right?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+    trySetMeshDetectorEnabled(enabled: boolean): boolean;
+    trySetPreferredMeshDetectorOptions(preferredOptions: XRGeometryDetectorOptions): boolean;
 }
 
+type XRMeshSet = Set<XRMesh>;
+
 interface XRFrame {
     featurePointCloud? : Array<number>;
+    // Open question: Should meshes be declared on the XRFrame or queried for the session?
+    detectedMeshes? : XRMeshSet;
```
ryantrem, line 81: Would `surfaceMeshes` better convey what this property represents?

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: Good catch. I agree we don't need a string here. Having to implement the native backend for a specific detection boundary type would warrant updating the XRDetectionBoundaryType definition.

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+
+type XRDetectionBoundaryType = "frustum" | "sphere" | "box";
+
+interface XRDetectionBoundary {
+    isStationary?: boolean;
+    type?: XRDetectionBoundaryType;
+    frustum?: XRFrustum;
+    sphereRadius?: number;
+    boxDimensions?: XRVector;
+}
```
ryantrem, line 62: Since the intention here is that it is one of the three types of boundaries, we should type this more strongly:
```suggestion
interface XRFrustrumDetectionBoundary {
  type: "frustum";
  frustum: XRFrustum;  
}

interface XRSphereDetectionBoundary {
  type: "sphere";
  radius: number;
}

interface XRBoxDetectionBoundary {
  type: "box";
  extent: XRVector;
}

type XRDetectionBoundary = XRFrustrumDetectionBoundary | XRSphereDetectionBoundary | XRBoxDetectionBoundary;
```
This will make intellisense help you do the right thing when constructing a boundary or reading a boundary. For example:
```ts
function Test(boundary: XRDetectionBoundary) {
  if (boundary.type === "frustum") {
    let f = boundary.frustum;
  }
}
```
Here, inside the `if` block where you have already determined that `type` is `"frustum"`, then `boundary` has been type narrowed and intellisense and the compiler know that the object has a property named `frustum` of type `XRFrustum`.

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+    position: XRVector;
+    rotation: XRQuaternion;
+    fieldOfView: XRFieldOfView;
+    farDistance: number;
+}
+
+interface XRPlane {
+    // Open question: Should geometry id's and types be declared on XRPlanes or queried through other means?
+    geometryId?: number;
+    geometryType?: XRGeometryType | string;
```
ryantrem, line 39: Do we need `| string`?

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+// Open question: Should XRMesh have a concept of a XRReferenceSpace similar to XRPlane?
+interface XRMesh {
+    positions: Float32Array;
+    indices: Uint32Array;
+    normals?: Float32Array;
+    // Open question: Do we need lastChangedTime declared for XRMeshes?
+    lastChangedTime: number;
+    // Open question: Should geometry id's and types be declared on XRMeshes or queried through other means?
+    geometryId?: number;
+    geometryType?: XRGeometryType | string;
```
ryantrem, line 51: Do we need `| string`?

src/XR/features/WebXRPlaneDetector.ts
RaananW: it should actually be fired. If no session is available (meaning - enabled but not in XR), the callback ias added to the XR Session Init observable. If it isn't fired, something is wrong.

Note also that there was a reason for adding the anonymous function to call this._init(), as when you add a callback directly, its context changes (so `this` is no longer what you expect it to be). There are ways around it, I find the anonymous function) the most readable. Especially when it only happens once. On each frame I would do it differently.

src/LibDeclarations/webxr.nativeextensions.d.ts
```diff
@@ -1,11 +1,82 @@
...
+}
+
+interface XRPlane {
+    // Open question: Should geometry id's and types be declared on XRPlanes or queried through other means?
+    geometryId?: number;
+    geometryType?: XRGeometryType | string;
+}
+
+// Open question: Should XRMesh have a concept of a XRReferenceSpace similar to XRPlane?
+interface XRMesh {
```
RaananW, line 43: Is it a mesh, or is it a geometry? I know we spoke semantics, but for me this is a geometry, that can be used to construct a mesh. The same way XRPlane is Plane the concept and not Plane the mesh-type.

src/LibDeclarations/webxr.nativeextensions.d.ts
RaananW: I guess there will be devs who will use it. I can't find a specific use-case right now, but if it is already available, why not add it? Is there a performance hit when doing that?

src/XR/features/WebXRMeshDetector.ts
```diff
@@ -0,0 +1,238 @@
...
+    public onMeshUpdatedObservable: Observable<IWebXRMesh> = new Observable();
+
+    constructor(_xrSessionManager: WebXRSessionManager, private _options: IWebXRMeshDetectorOptions = {}) {
+        super(_xrSessionManager);
+        this.xrNativeFeatureName = "mesh-detection";
+        this._initMeshDetector = this._initMeshDetector.bind(this);
+        if (this._xrSessionManager.session) {
+            this._initMeshDetector(this._xrSessionManager.session);
+        } else {
+            this._xrSessionManager.onXRSessionInit.addOnce(this._initMeshDetector);
```
RaananW, line 100: same issue as mentioned before with context.

src/XR/features/WebXRMeshDetector.ts
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    public detach(): boolean {
+        if (!super.detach()) {
+            return false;
+        }
+
+        if (!!this._xrSessionManager.session.trySetMeshDetectorEnabled) {
+            this._xrSessionManager.session.trySetMeshDetectorEnabled(false);
```
RaananW, line 110: The one thing that really bothers me with the WebXR Plane detector is that it can't do that :-) it is either on or off throughout the scene. Awesome :-)

src/XR/features/WebXRMeshDetector.ts
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+        else {
+            mesh.positions = xrMesh.positions;
+            mesh.indices = xrMesh.indices;
+            mesh.normals = xrMesh.normals;
+        }
+
+        mesh.transformationMatrix = this._options.worldParentNode?.getWorldMatrix() ?? Matrix.Identity();
+        return <IWebXRMesh>mesh;
```
RaananW, line 218: expanding the discussion earlier about the mesh vs. geometry - this object is actually not a mesh. A mesh should extend BABYLON.AbstractMesh (or TransformNode), otherwise it will be confusing for the developer.

I have internally discussed (with myself), whether or not the plane detector should construct a PlaneMesh on each frame, or let the user do it. As I assumed people will use those polygons in different ways, i have decided not to do that. But in the case of this one, generating (and constantly updating) a Babylon VertextBuffer with this data (or maybe BABYLON.Geometry?) would make a lot of sense. You can use the geometry to generate a mesh, you can keep a reference to it and constantly update it. 

One could argue that the plane detector could do the same, and I would probably agree with that :-) The issue with the plane detector is that we have a polygon builder that takes points and space on a single plane and extrude them, which is a bit different that generating a reusable geometry (as it builds a mesh directly).

src/XR/features/WebXRPlaneDetector.ts
```diff
@@ -187,23 +189,29 @@ export class WebXRPlaneDetector extends WebXRAbstractFeature {
...
         const internalInit = () => {
             this._enabled = true;
             if (this._detectedPlanes.length) {
                 this._detectedPlanes.length = 0;
             }
         };
-        if (!this._xrSessionManager.session.updateWorldTrackingState) {
+
+        if (!!this._options.preferredDetectorOptions &&
+            !!session.trySetPreferredPlaneDetectorOptions) {
```
RaananW, line 201: I wonder - this function is added to the XRSession in babylon native, but is not included in WebXR.d.ts, which is the one most web users will use. Won't that fail with unknown function for web devs?

src/XR/features/WebXRMeshDetector.ts
```diff
@@ -0,0 +1,238 @@
...
+        if (!!this._options.preferredDetectorOptions &&
+            !!session.trySetPreferredMeshDetectorOptions) {
+            session.trySetPreferredMeshDetectorOptions(this._options.preferredDetectorOptions);
+        }
+    }
+
+    private _updateMeshWithXRMesh(xrMesh: XRMesh, mesh: Partial<IWebXRMesh>, xrFrame: XRFrame): IWebXRMesh {
+        mesh.xrMesh = xrMesh;
+        if (!this._xrSessionManager.scene.useRightHandedSystem) {
+            mesh.positions = new Float32Array(xrMesh.positions.length);
```
RaananW, line 188: why are you creating a new array on each frame? This will be heavy on the garbage collection. If the size changes constantly (which I assume it does), it would be better to use a native js array and keep on expanding it. if it doesn't, there is no need to create a new array. 

Another idea would be to allocate more space than needed space and keep on exponentially increasing it to prevent creating a new Float32Array (which is admittingly faster) while keeping a reference to the actual length. 

Having said all of that, all of those arrays are a direct copy of the arrays in the XRMesh. why not provide them directly? why do we need to recreate them on each frame?

---EDIT---

I just realized this is done because of the right handed system. There is a way to avoid recreating the arrays, which is the way we are loading glTF files (which are right-handed per definition). There is a combination of scaling rotation definition that flips the z correctly, which can be applied to the transformation matrix that the babylon devs will be used (See here - https://github.com/BabylonJS/Babylon.js/blob/master/loaders/src/glTF/2.0/glTFLoader.ts#L481). This is an acceptable solution - adding a root node that the dev won't change, and a child-mesh in it that has this defined geometry.

src/LibDeclarations/webxr.nativeextensions.d.ts
RaananW: DOMPoint (https://drafts.fxtf.org/geometry-1/#DOMPoint) can be used for both types (and is being used extensively in XR). And you can use DOMPointReadOnly when it makes sense
Or https://immersive-web.github.io/webxr/#xrrigidtransform directly, if it makes more sense?

src/XR/features/WebXRMeshDetector.ts
```diff
@@ -0,0 +1,238 @@
...
+            this._xrSessionManager.onXRSessionInit.addOnce(this._initMeshDetector);
+        }
+    }
+
+    public detach(): boolean {
+        if (!super.detach()) {
+            return false;
+        }
+
+        if (!!this._xrSessionManager.session.trySetMeshDetectorEnabled) {
```
RaananW, line 109: Something about the native vs. babylon issue that I discussed somewhere else already - 
It will be great to have it readable in code that this is a native-only feature. Something like we are doing here - https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/webXRSessionManager.ts#L137 which clearly tells the developer - this is not for you if you are not in native.

RaananW (commented):
Many comments, I am happy about this discussion. 

As I mentioned in one or two of my comments, I have a semantic issue with Mesh vs. Geometry, but this is a great starting point!!

There is also the need to check how we can support web devs that should not see native functions in their webxr declaration. We can discuss this next week, I am still thinking of an acceptable solution for everyone.

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: I am going to rename these variables to parentGeometryId and parentGeometryType

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: I believe that it should have an XRReferenceSpace for consistency in API. I will update here accordingly.

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: We will add last changed time for the time being

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: Moving detectedMeshes to a new XRWorldInformation interface for the time being

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: Going to stick to a strict type definition for now. we can take another look at this when enabling more platforms if we feel it makes for a fractured experience.

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: @RaananW does parentGeometryId/parentGeometryType for names help clear up confusion?

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: @ryantrem  I won't claim to be a typescript expert, but is it better to have interface inheritance here where type is guaranteed in XRDetectionBoundary. Then other detection boundaries extend XRDetectionBoundary. Or is it better to have a type definition as described above? It seems like with the type definition we could accidentally end up with types added to the type definition that don't populate a type field.

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: I've gone ahead and moved things to a stricter type definition.

src/XR/features/WebXRMeshDetector.ts
chrisfromwork: @RaananW do we know whether colliders will be calculated for meshes after applying this right handed to left handed scale transform?

Also, it seems like we may still need to change the indices orders here. Is it better to have a new indices array without updated positions/normals?

src/XR/features/WebXRMeshDetector.ts
chrisfromwork: @RaananW with this suggestion, would you be open to exposing a native flag on the webXRSessionManager? It looks like currenty the xrNavigator is a private property on the session manager?

src/XR/features/WebXRPlaneDetector.ts
chrisfromwork: @RaananW after script debugging with BabylonNative, it appears this is an issue with Observable._deferUnregister. In BabylonNative, _deferUnregister results in Observable._observers being edited during a for loop. It looks like the intention of this logic is to avoid removing observers in a synchronous manner. But for BabylonNative UWP + Chakra, this removal ends up being synchronous. 

In babylon.max.js, a for loop over _observers gets reduced to a for loop based on array length. This array length gets edited due to the synchronous removal of an observer. It causes the for loop to skip the n observers based on n previous observers being removed from the array length.

Is there an expectation that Observable._deferUnregister executes asynchronously? It seems like a lot of things may actually be broken based on this observable bug in BabylonNative.

chrisfromwork: filed an issue here: https://github.com/BabylonJS/Babylon.js/issues/9362

Happy to help drive a fix for this problem. I'm not sure what the intended behavior is for Observable._deferUnregister

src/LibDeclarations/webxr.nativeextensions.d.ts
thetuvix: The term "geometry" may be ambiguous here.  This parent object is what we are calling `SceneObject`s in our native OpenXR extension.  I think of a scene object as being made up of some number of planes and meshes, with each of those being kinds of geometry.

chrisfromwork: not opposed to renaming this as scene object. @RaananW does scene object or parentSceneObjectId/parentSceneObjectType make more sense here on your end?

RaananW: As you say, this will have to be documented correctly, but I have no problem with SceneObject and both parentGeometry variables

src/XR/features/WebXRMeshDetector.ts
RaananW: I am perfectly fine with exposing this. Will make native vs. browser easier to differentiate.

src/XR/features/WebXRMeshDetector.ts
RaananW: It's better to have as little as possible changes that will require garbage collection (preferably 0 :-) ). I will leave it up to you, as you know the structure much better than me.

Colliders can be applied, based on the babylon-mesh geometry data, and should be updated when the vertex buffer changes as well. Left-hand vs. right-hand is being checked on each place in the framework where it is needed to differentiate between them, so you need to store according to the scene's handedness, and the other components will take it from there. I hope this answers the question. if it doesn't, please let me know.

src/XR/features/WebXRPlaneDetector.ts
chrisfromwork: This should now be resolved, the issue was specific to BabylonNative

src/XR/features/WebXRMeshDetector.ts
chrisfromwork: I have added isNativeSession() to WebXRSessionManager instances

src/XR/features/WebXRMeshDetector.ts
chrisfromwork: This should now be reverted

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: calling these parentSceneObjectType and parentSceneObjectId for now

chrisfromwork: @RaananW would you prefer if we created a new object type XRSceneObject that then was available in place of parentSceneObjectId?

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: Lets call these sceneobjects for now. I'll work to change these to be a class instance for comparison compared to relying on ids

src/LibDeclarations/webxr.nativeextensions.d.ts
chrisfromwork: there are different questions open for this now. Let's follow up in the separate conversations.

src/LibDeclarations/webxr.nativeextensions.d.ts
ryantrem: I think the union type I suggest above is a stronger type assertion than having a base interface with additional interfaces that extend it, so I would recommend that route.

src/LibDeclarations/webxr.nativeextensions.d.ts
RaananW: I think it will make more sense semantically, yes.

src/XR/webXRSessionManager.ts
```diff
@@ -332,6 +332,13 @@ export class WebXRSessionManager implements IDisposable {
         }
     }
 
+    /**
+     * Returns true if the session's backend is BabylonNative, otherwise false
+     */
+    public isNativeSession(): boolean {
```
RaananW, line 338: can be a getter, so it can be used as a readonly flag

```javascript
public get isNative(): boolean {
    return !!this._xrNavigator.xr.native;
}
```

I also think it is not the session, but the environment (this is why i removed the `Session`. But that's arguable :-)

RaananW:
@chrisfromwork - I will merge this the minute it builds correctly. can you make sure all of the imports are correct and are pointing to a file (and not a directory). The build process is very strict on those kind of things.
You can try building the framework locally and see what the errors are. Contact me if you encounter some cryptic message you can't decipher :-)

src/XR/features/WebXRMeshDetector.ts
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * an array of indices in babylon space. right/left hand system is taken into account.
+     */
+    indices: Uint32Array;
+    /**
+     * an array of vertex normals in babylon space. right/left hand system is taken into account.
+     */
+    normals?: Float32Array;
+    /**
+     * data required for construction a mesh in Babylon.js
```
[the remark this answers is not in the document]
chrisfromwork: good catch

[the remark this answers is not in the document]
chrisfromwork: i think this may be fixed in a newer iteration

src/XR/features/WebXRMeshDetector.ts
chrisfromwork: Is backface culling set as a mesh or material property? IMO, most developers may not know that they will need to enable backface culling to get a valid winding order. I'd prefer to error on the side of exposing the correct winding order when meshes are auto-calculated (now an option when setting things up). If auto-calculations are turned off, folks can go down the backface culling/negative scale path on their own.

---
In reply to: [518369210](https://github.com/BabylonJS/Babylon.js/pull/9307#discussion_r518369210) [](ancestors = 518369210)

RaananW (approved):
Initial version. let's take it from here and improve where needed.

PR #9314, opened 2020-10-30
PhotoDome: Fix textureMode lost after changing the texture

Popov72:
See https://forum.babylonjs.com/t/changing-texture-in-photodome/15292


deltakosh:
Can you check the validation error? Or is it just transient?

deltakosh:
ok unrelated we are good!

PR #9315, opened 2020-10-30
gltfloader owned buffers are disposed with engine

CedricGuillemet:
We experienced leaked bgfx resources in babylon native when loading .gltf.
Those leaks are caused by a buffer that is owned by gltfloader and never disposed.
To fix that, I've added an onDisposed callback on thinengine and dispose those buffers when the engine is disposed.
The buffer array to dispose is static because it's possible to load multiple gltf in one engine.

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -1675,8 +1675,17 @@ export class GLTFLoader implements IGLTFLoader {
             const bufferView = ArrayItem.Get(`${context}/bufferView`, this._gltf.bufferViews, accessor.bufferView);
             accessor._babylonVertexBuffer = this._loadVertexBufferViewAsync(bufferView, kind).then((babylonBuffer) => {
                 const size = GLTFLoader._GetNumComponents(context, accessor.type);
-                return new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
+                var vertexBuffer = new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
                     false, accessor.byteOffset, size, accessor.componentType, accessor.normalized, true);
+
+                // gltfload owns this buffer and is responsible to dispose it at some point.
+                // for now, it's doing that procedure when the engine is disposed
+                if (babylonBuffer instanceof Buffer) {
+                    this._babylonScene.getEngine().onDispose = () => {
+                        vertexBuffer._buffer.dispose();
+                    };
+                }
+                return vertexBuffer;
             });
         }
 
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -116,6 +116,7 @@ export class GLTFLoader implements IGLTFLoader {
     private _babylonScene: Scene;
     private _rootBabylonMesh: Mesh;
     private _defaultBabylonMaterialData: { [drawMode: number]: Material } = {};
+    private static _buffersToDispose = new Array<VertexBuffer>();
 
     private static _RegisteredExtensions: { [name: string]: IRegisteredExtension } = {};
 
@@ -1681,8 +1682,15 @@ export class GLTFLoader implements IGLTFLoader {
                 // gltfload owns this buffer and is responsible to dispose it at some point.
                 // for now, it's doing that procedure when the engine is disposed
                 if (babylonBuffer instanceof Buffer) {
+                    GLTFLoader._buffersToDispose.push(vertexBuffer);
                     this._babylonScene.getEngine().onDispose = () => {
-                        vertexBuffer._buffer.dispose();
+                        while (GLTFLoader._buffersToDispose.length)
+                        {
+                            var bufferToDispose = GLTFLoader._buffersToDispose.pop();
+                            if (bufferToDispose) {
+                                bufferToDispose._buffer.dispose();
+                            }
+                        }
                     };
                 }
                 return vertexBuffer;
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -299,6 +299,22 @@ export class ThinEngine {
     protected _renderingQueueLaunched = false;
     protected _activeRenderLoops = new Array<() => void>();
 
+    /**
+    * An event triggered when the thin engine is disposed.
+    */
+    public onDisposeObservable = new Observable<ThinEngine>();
+
+    private _onDisposeObserver: Nullable<Observer<ThinEngine>> = null;
+    /** Sets a function to be executed when this engine is disposed. 
+     * This callback has been introduced to allow gltf loader to dispose its own buffer.
+    */
+    public set onDispose(callback: () => void) {
+        if (this._onDisposeObserver) {
+            this.onDisposeObservable.remove(this._onDisposeObserver);
+        }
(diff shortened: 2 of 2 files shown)


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1675,8 +1676,24 @@ export class GLTFLoader implements IGLTFLoader {
             const bufferView = ArrayItem.Get(`${context}/bufferView`, this._gltf.bufferViews, accessor.bufferView);
             accessor._babylonVertexBuffer = this._loadVertexBufferViewAsync(bufferView, kind).then((babylonBuffer) => {
                 const size = GLTFLoader._GetNumComponents(context, accessor.type);
-                return new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
+                var vertexBuffer = new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
```
ryantrem, line 1679:
```suggestion
                const vertexBuffer = new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
```

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1675,8 +1676,24 @@ export class GLTFLoader implements IGLTFLoader {
...
                     false, accessor.byteOffset, size, accessor.componentType, accessor.normalized, true);
+
+                // gltfload owns this buffer and is responsible to dispose it at some point.
+                // for now, it's doing that procedure when the engine is disposed
+                if (babylonBuffer instanceof Buffer) {
+                    GLTFLoader._buffersToDispose.push(vertexBuffer);
+                    this._babylonScene.getEngine().onDispose = () => {
+                        while (GLTFLoader._buffersToDispose.length)
+                        {
+                            var bufferToDispose = GLTFLoader._buffersToDispose.pop();
```
ryantrem, line 1689:
```suggestion
                            const bufferToDispose = GLTFLoader._buffersToDispose.pop();
```

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -116,6 +116,7 @@ export class GLTFLoader implements IGLTFLoader {
     private _babylonScene: Scene;
     private _rootBabylonMesh: Mesh;
     private _defaultBabylonMaterialData: { [drawMode: number]: Material } = {};
+    private static _buffersToDispose = new Array<VertexBuffer>();
```
ryantrem, line 119:
```suggestion
    private static readonly _buffersToDispose = new Array<VertexBuffer>();
```

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1675,8 +1676,24 @@ export class GLTFLoader implements IGLTFLoader {
...
                 const size = GLTFLoader._GetNumComponents(context, accessor.type);
-                return new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
+                var vertexBuffer = new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
                     false, accessor.byteOffset, size, accessor.componentType, accessor.normalized, true);
+
+                // gltfload owns this buffer and is responsible to dispose it at some point.
+                // for now, it's doing that procedure when the engine is disposed
+                if (babylonBuffer instanceof Buffer) {
+                    GLTFLoader._buffersToDispose.push(vertexBuffer);
+                    this._babylonScene.getEngine().onDispose = () => {
```
ryantrem, line 1686: Engine dispose seems too coarse grained. Can we do this on scene scene dispose?

ryantrem (commented):
Can you explain this a bit more:

> The buffer array to dispose is static because it's possible to load multiple gltf in one engine.

I would have thought the buffer is owned by the mesh it is loaded for, but I don't really know this code.

loaders/src/glTF/2.0/glTFLoader.ts
CedricGuillemet: correct me if I'm wrong @bghgary but assets loading with gltf belong to an asset container. and that asset container can be used to clone assets to different scenes. 1st scene dispose would result in invalid buffer for the other ones.

loaders/src/glTF/2.0/glTFLoader.ts
CedricGuillemet: Is it possible to push item into a readonly array?

CedricGuillemet:
> I would have thought the buffer is owned by the mesh it is loaded for, but I don't really know this code.
No, gltfloader creates the buffer, and in bjs, it's flagged as !_ownsBuffer here :https://github.com/BabylonJS/Babylon.js/blob/ace7268706b8f87fb4ea0e954c87cddc00e76e44/src/Meshes/buffer.ts#L287
Then, when the vertexbuffer is disposed, that buffer is not disposed :
https://github.com/BabylonJS/Babylon.js/blob/ace7268706b8f87fb4ea0e954c87cddc00e76e44/src/Meshes/buffer.ts#L448
As no one is disposing the buffer, we experience a leak on native.
In a browser, I guess the GC will collect everything at some point. especially when the canvas is diposed and all webgl resources are destroyed.

loaders/src/glTF/2.0/glTFLoader.ts
deltakosh: yes

ryantrem: Yes the `readonly` here just means the `_buffersToDispose` variable cannot be reassigned. That's different from the array itself being readonly (`ReadonlyArray<VertexBuffer>` or `Readonly<Array<VertexBuffer>>` or `Readonly<VertexBuffer[]>` or `readonly VertexBuffer[]` 😬), in which case items cannot be pushed.

loaders/src/glTF/2.0/glTFLoader.ts
deltakosh: asset container lives within a scene so we are good at the scene level (and it is better for memory pressure as well)

deltakosh: Furthermore this will help not adding code to thinengine

Popov72:
Why not adding a method like `takeBufferOwnership` to `VertxBuffer`, or even simpler add a `takeBufferOwnership` boolean to the `VertexBuffer` constructor? We would just set this flag to `true` when creating the vertex buffers in `_loadVertexAccessorAsync` and that would be it (note that there are multiple occurrences of vertex buffer creation in `_loadVertexAccessorAsync` and your current changes handle only one).

It seems logical to me that the meshes created by glTF handles the vertex buffer lifespan as any other mesh, all the more that the vertex buffers created by `_loadVertexAccessorAsync` are specific to a mesh and are not shared (if not mistaken).

deltakosh:
I concur

sebavan:
The buffer might unfortunately be shared  :-(

deltakosh:
it can be shared with the option offered by Evgeni

deltakosh:
Sorry folks but can you tell me why we need all this?
If you create multiple vertex buffers on a common buffer, the buffer will have a reference count increased. Meaning that as soon as the mesh will be dispose, it will call the buffer dispose which in turn will reduce the ref count. 

When ref count hits 0 the it will be disposed

So I'm not sure why we need all that?

Can someone shine my light ?

PR #9318, opened 2020-10-30
Fix gltf memory leak

deltakosh:


src/Meshes/buffer.ts
```diff
@@ -168,6 +169,20 @@ export class Buffer {
...
     }
 
+    /** @hidden */
+    public _increaseReferences() {
+        if (!this._buffer) {
+            return;
+        }
+
+        if (!this._isAlreadyOwned) {
+            this._isAlreadyOwned = true;
```
deltakosh, line 179: I'm doing that because the buffer has already a counter of 1 at his creation

Popov72 (approved):
LGTM!

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1676,7 +1676,7 @@ export class GLTFLoader implements IGLTFLoader {
             accessor._babylonVertexBuffer = this._loadVertexBufferViewAsync(bufferView, kind).then((babylonBuffer) => {
                 const size = GLTFLoader._GetNumComponents(context, accessor.type);
                 return new VertexBuffer(this._babylonScene.getEngine(), babylonBuffer, kind, false, false, bufferView.byteStride,
-                    false, accessor.byteOffset, size, accessor.componentType, accessor.normalized, true);
+                    false, accessor.byteOffset, size, accessor.componentType, accessor.normalized, true, 1, true);
```
Popov72, line 1679: Actually, I forgot to say that this change should also be done above 3 other times are there are 3 other `new VertexBuffer()...` calls.

CedricGuillemet (approved):
Tested with babylon native and it works great!

PR #9331, opened 2020-11-02
Update stlSerializer.ts

aWeirdo:
Add boolean option to not bake mesh transforms on export


deltakosh (approved):

aWeirdo:
that was fast! 😄

deltakosh:
we are code freezing so it was last minute but it made it :)

PR #9337, opened 2020-11-03
Added `fixedDragMeshBoundsSize` for boundingbox gizmo to size anchors…

CedricGuillemet:
… from bounds size

Follow up https://forum.babylonjs.com/t/gizmo-bounding-box-scale-problem/15354/5


Popov72:
LGTM, maybe you should document that `fixedDragMeshScreenSize` takes precedence over `fixedDragMeshBoundsSize` if both are `true`?

deltakosh (approved):

PR #9341, opened 2020-11-03
Update math.vector.ts

aWeirdo:
Adds static "Center**ToRef**" for Vector2/3/4

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -651,9 +651,7 @@ export class Vector2 {
      * @returns a new Vector2
      */
     public static Center(value1: DeepImmutable<Vector2>, value2: DeepImmutable<Vector2>): Vector2 {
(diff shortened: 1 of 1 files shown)


deltakosh:
We are in code freeze so we can only merge bug fixes

I will merge this PR as soon as 4.2 will be out :)

src/Maths/math.vector.ts
```diff
@@ -656,6 +656,17 @@ export class Vector2 {
         return center;
     }
 
+    /**
```
deltakosh, line 659: can you fix conflicts and make sure that CEnter is using CenterToRef?

aWeirdo: Will be done by monday :)

aWeirdo:
Github giving problems :)
https://github.com/BabylonJS/Babylon.js/pull/9406

PR #9342, opened 2020-11-03
Arcrotatecamera rebuild angles fix

nilss0n:
Fix for the following issue: https://forum.babylonjs.com/t/issue-involving-arcrotatecamera-collisions-and-angular-limits/15397

ArcRotateCamera.rebuildAnglesAndRadius calculates alpha in the range [0, 2*Pi]. If the previous alpha value was outside this range (e.g. < 0) it could cause the new value to  be outside the alpha limits, causing the camera to jump to the closest limit. 

This fix adds a correction to the alpha value such that the resulting numerical value is the closest one to the old value.


deltakosh:
thanks a lot!

PR #9355, opened 2020-11-05
Fix parsing the camera upVector from serialization

Popov72:
See https://forum.babylonjs.com/t/sceneloader-fails-to-load-data-if-upvector-of-a-camera-is-different-than-0-1-0/15433


deltakosh:
checks are failing

PR #9359, opened 2020-11-05
Reworked motion blur post-process to include screen based mode.

julien-moreau:
By default the motion blur post-process is object based. To save performances on large scenes, it can be useful to fallback on screen base mode which doesn't require to compute any velocity texture. In screen based, only depth texture is required.

Also fixed badly decorated members for serialization.

I can see that prepass renderer fallbacks automatically on geometry buffer renderer when failed. Do you think the constructor parameter "forceGeometryBuffer" should not set to "false" by default now? If yes, the same should be done for SSAO2

### DIFF
--- a/src/PostProcesses/motionBlurPostProcess.ts
+++ b/src/PostProcesses/motionBlurPostProcess.ts
@@ -1,6 +1,6 @@
 import { Nullable } from "../types";
 import { Logger } from "../Misc/logger";
-import { Vector2 } from "../Maths/math.vector";
+import { Matrix, Vector2 } from "../Maths/math.vector";
 import { Camera } from "../Cameras/camera";
 import { Effect } from "../Materials/effect";
 import { PostProcess, PostProcessOptions } from "./postProcess";
@@ -41,6 +41,7 @@ export class MotionBlurPostProcess extends PostProcess {
     /**
      * Gets the number of iterations are used for motion blur quality. Default value is equal to 32
      */
+    @serialize()
     public get motionBlurSamples(): number {
         return this._motionBlurSamples;
     }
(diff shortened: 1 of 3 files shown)


deltakosh:
We are setting forceGeometryBuffer = true for now because of all the problem we face lately. We are going to ship 4.2 in a week:)

we will turn it off by default for 5.0

I will also merge your change after shipping 4.2

deltakosh:
cc @CraigFeldspar

src/PostProcesses/motionBlurPostProcess.ts
```diff
@@ -181,6 +187,98 @@ export class MotionBlurPostProcess extends PostProcess {
...
+            if (this._prePassRenderer && this._prePassEffectConfiguration) {
+                this._prePassEffectConfiguration.texturesRequired[0] = Constants.PREPASS_DEPTHNORMAL_TEXTURE_TYPE;
+            }
+
+            this.onApply = (effect: Effect) => this._onApplyScreenBased(effect);
+        }
+    }
+
+    /**
+     * Called on the effect is applied when the motion blur post-process is in object based mode.
```
CraigFeldspar, line 224: missing doc for param effect

src/PostProcesses/motionBlurPostProcess.ts
```diff
@@ -181,6 +187,98 @@ export class MotionBlurPostProcess extends PostProcess {
...
+            const velocityIndex = this._geometryBufferRenderer.getTextureIndex(GeometryBufferRenderer.VELOCITY_TEXTURE_TYPE);
+            effect.setTexture("velocitySampler", this._geometryBufferRenderer.getGBuffer().textures[velocityIndex]);
+        } else if (this._prePassRenderer) {
+            const velocityIndex = this._prePassRenderer.getIndex(Constants.PREPASS_VELOCITY_TEXTURE_TYPE);
+            effect.setTexture("velocitySampler", this._prePassRenderer.prePassRT.textures[velocityIndex]);
+        }
+    }
+
+    /**
+     * Called on the effect is applied when the motion blur post-process is in screen based mode.
```
CraigFeldspar, line 242: missing doc for param effect

CraigFeldspar (commented):
couple of nits, but overall LGTM 👍

src/PostProcesses/motionBlurPostProcess.ts
julien-moreau: Thanks! This is a private method but I can add if needed :)

deltakosh (approved):

PR #9364, opened 2020-11-06
Scene Manager: GLTF Loader Fix

MackeyK24:
Fixed GLTF Loader Dispose Issue


sebavan:
It looks like this also removes the nav mesh and does not revert the other parts: https://github.com/BabylonJS/Babylon.js/pull/9360

The PR above should show you what to uncomment and where if that helps.

deltakosh:
I would like to offer an option here: In order to NOT load your framework when not required, I can add a bit of code that will detect the presence of SceneManager in the PG code and then dynamically load your lib

so no more problem

MackeyK24:
Well i removed the offending code completely. So that should not be an issue anymore. As far as the GLTF Loader. That checks the extension in use variable and enables the loader ONLY if CVTOOLS_Unity_Metadata extension is in use
```
    constructor(loader: BABYLON.GLTF2.GLTFLoader) {
          this.enabled = this._loader.isExtensionUsed(CVTOOLS_NAME);
    }
```

MackeyK24:
The legacy navmesh has been removed because really old... All navigation has been (for a while now) using RecastJSPlugin. So no need the older style navmesh library

deltakosh:
What about my proposal?

MackeyK24:
If having my Scene Manager extension loaded by default with script tag is messing things up (Which it should not anymore) i am game for whatever you see fit @deltakosh

deltakosh:
Ok I will make that work for you buddy
can you share a PG that is actually using your toolkit?

MackeyK24:
Well that's just it... You dont really use the extension directly... Its all in the GLTF Loader.

How my Unity Exporter works with GLTF

1... I export a standard GLTF 2.0 of the unity content (converts everything to PBR)
2...I serialize a little bit of extra metadata on the nodes and materials
3...My GLTF Loader hooks in to the GLTF parsing process and parses out that extra metadata and instantiates script components

I made it as seamless as possible... So wheather you use SceneLoader.Load or Append or ImportMeshes or asset containers or ANYTHING that goes thru the Babylon GLTF parser... IF CVTOOLS_Unity_Metadata is used in the gltf
it auto parses everthing at load time.

So you literally have nothing to do in the playground except load your Unity Created content... Its all in the GLTF Export/Import process

But i removed that offending code... so you can really put back in the script tag to load the **babylon.manager.js** extension

deltakosh:
ok gotcha! But I do not like having the PG to load resources that we do not have direct control over...

deltakosh:
What about adding a query url? if the url contains &UnityToolkit then I also load the file?

deltakosh:
I want to make sure you can use the PG but not add more load time / risks for the rest

deltakosh:
Something like that: https://www.babylonjs-playground.com/#QYFDDP#1?UnityToolkit

bghgary:
Why not just have the playground script load it?

deltakosh:
Because I do not want the PG to load it all the time (slow down) and we do not control the code so it could generate issues (like this one) that we struggle to find

bghgary:
I mean the PG code (the code that the user puts in the PG).

deltakosh:
That is also correct but I wanted to make it handy for @MackeyK24

MackeyK24:
I really wanted folks to be able to do ANYTHING i do in the editor, but right in the playground WITH or WITHOUT loading an actual scene file (gltf)

I did strip the GLTF code that causing problems... And i get you dont want hard code the script tag on the host web page because you dont have complete control of that particular library

But if you have to remove completely, i guess i understand.

I guess the playground folks just cant do all the kool unity exported stuff unless they KNOW they need to manually load the babylon.mananger.js extension before loading any UNITY content or call any Scene Manager API functions or use BABYLON.ScriptComponents.

Bummer, but i guess i can live that option

MackeyK24:
Oh yeah... That a thing now... The Hard Core Babylon Toolkit Developer... Thats an actual Job Market Classification ... We (and other small dev shops) are actually search for **Babylon Toolkit Developers** full time positions as well as contractors.

We come along way 👍

deltakosh:
> We come along way

Man! I'm so glad of that...Seriously I cannot be prouder

deltakosh:
> Bummer, but i guess i can live that option.
You simply need to add "?UnityToolkit" in the url :)

Other option: I can add a checkbox in the menu if you want to?

MackeyK24:
im ok with both of those options

bghgary:
I think this should be tied with the PG snippet somehow. If people load a PG that needs Unity Toolkit without the checkbox/url option, it will fail.

MackeyK24:
Have we decided what we are going to do here ?

deltakosh:
So far I've added support for "?UnityToolkit" in the url 

Will close this issue for now
@MackeyK24 please ping me on the forum if you need more :)

MackeyK24:
Yo @deltakosh ... can you please ALSO add the a menu item as mentioned here:

Other option: I can add a checkbox in the menu if you want to? (something in the gear menu that people can check and then I will save it locally so it will be loaded all the time for that user?)

deltakosh:
Ok I will

deltakosh:
https://github.com/BabylonJS/Babylon.js/issues/9382

deltakosh:
Here we are:
image

MackeyK24:
Thanks @deltakosh 👍

MackeyK24:
Yo @deltakosh .. Im getting some errors using the ?UnityToolkit option. Looks like it does not like loading the babylon.manager.js like we are now. I think is CORS and the mime type being text/plain

Screen Shot 2020-11-12 at 8 13 54 AM

deltakosh:
make sure to clear your cache (I have no issue here)
Or even close / restart your browser

MackeyK24:
I tried that... Even tried a different browser... Is says its refused because its not a script mime type

Screen Shot 2020-11-12 at 10 49 10 AM

deltakosh:
can you share a PG that repro it?

MackeyK24:
https://playground.babylonjs.com/#G16I4Y

Didnt show error in console.. But none of the BABYLON.SceneManager functions are there either

deltakosh:
Ok I had to move your file there:
https://github.com/BabylonJS/Babylon.js/blob/master/Playground/libs/babylon.manager.js

Please make sure to update it if you update the source ;D

MackeyK24:
Thats even better, i never really liked host that file on my github.io anyways.

So i can just update the playground lib babylon.manager.js with a regular Pull Request... Sweet

MackeyK24:
Yo @deltakosh ...
Now that you are loading the Unity toolkit from a switch...

Is there **ANY** way at all to also **INJECT** my **babylon.manager.d.ts** file

Otherwise all i get is editor syntax errors when actually using the BABYLON.SceneManager on the playground

ANYTHING that help so i dont errors when using SceneManager API

sebavan:
@MackeyK24 could you create a PR to put the babylon.manager.d.ts aside the lib ? I ll take care after to include it in the playground.

MackeyK24:
Is there some kind of RULE to prevent babyon.manager.d.ts file in the Playground/Libs folder.

I added both scene manager .js and .d.ts files... But the .d.ts does NOT show up in Github Change list

Screen Shot 2020-11-22 at 10 39 01 AM

MackeyK24:
It seems the project .gitignore is ingoring the .d.ts files... Why ???

```
# javascript files
src/**/*.js
!src/webpack.config.js
src/**/*.d.ts
!src/LibDeclarations/*.d.ts
```

sebavan:
cause we were having to many left over there at some points. feel free to add it back in.

PR #9368, opened 2020-11-07
Implement support for EXT_meshopt_compression

zeux:
This change implements support for EXT_meshopt_compression extension
using the WebAssembly implementation from meshoptimizer repository.

Extension spec: https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/EXT_meshopt_compression
Some performance / size numbers are available in the PR: https://github.com/KhronosGroup/glTF/pull/1830

To integrate the extension it's sufficient to override loadBufferView
and decode the provided buffer into a fresh Uint8Array. meshopt_decoder
module handles the relevant decoding procedure, including dynamic
dispatch to one of decodeVertexBuffer/decodeIndexBuffer/decodeIndexSequence
based on the mode.

The WebAssembly binary is embedded into the .js file, so no extra setup
is necessary; a .d.ts file is provided for type checking.

glTF files for this extension can be produced using gltfpack with -c/-cc
options.

Here's an example file packed with this extension using gltfpack (also using BasisU to reduce file size):
[Alien-packed.glb.zip](https://github.com/BabylonJS/Babylon.js/files/5504444/Alien-packed.glb.zip)

This file is an example of why this extension is useful - it contains morph target data which can't be compressed by Draco, but can be compressed by this extension.

I'm very new to TypeScript so any feedback on the change is welcome, as well as on the location of the decoder.


zeux:
Ah and one more test file for good measure just in case, this one doesn't have textures but has a fair amount of animation data; the pure compression ratio here is 8.5x compared to the original glb file from glTF-Sample-Models.

[brainstem.glb.zip](https://github.com/BabylonJS/Babylon.js/files/5504456/brainstem.glb.zip)

deltakosh:
You have some checks issues

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.d.ts
```diff
@@ -0,0 +1,11 @@
+// This file is part of meshoptimizer library and is distributed under the terms of MIT License.
```
bghgary, line 1: External files should sit in `dist/preview release`, perhaps under a sub directory.

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.js
```diff
@@ -0,0 +1,113 @@
...
...QILKOSQfbPden8c8d8e8fCggFDtD9OD/6FAKAEDQBFGENVcMTtmYi8ZpyHECTD+sFD/6FHID/gFAECTD+rFCTD+sFD/6FHLD/gFD/kFD/lFHE9DB/+g6DYALAEAOD+2FHOALCUUUU94DtHcD9OD9RD/kFHLALD/mFAEAED/mFAIAOAIAcD9OD9RD/kFHEAED/mFD/kFD/kFD/jFD/nFHID/mF9DBBX9LDYHOD/kFCTD+rFALAID/mFAOD/kFCggEDtD9OD9QHLAEAID/mFAOD/kFCaDbCBDnGCBDnECBDnKCBDnOCBDncCBDnMCBDnfCBDnbD9OHEDQNVi8ZcMpySQ8c8dfb8e8fD9QDMBBABAKAND9OALAEDQBFTtGEmYILPdKOenD9QDMBBABCAJRBAGCIJHGAF9JQBMMM/hEIGaF97FaL978jUUUUBCTlREGXAF9FQBCBRIEXAEABDBBBHLABCTJHKDBBBHODQILKOSQfbPden8c8d8e8fHNCTD+sFHVCID+rFDMIBAB9DBBU8/DY9D/zI818/DYAVCEDtD9QD/6FD/nFHVALAODQBFGENVcMTtmYi8ZpyHLCTD+rFCTD+sFD/6FD/mFHOAOD/mFAVALCTD+sFD/6FD/mFHcAcD/mFAVANCTD+rFCTD+sFD/6FD/mFHNAND/mFD/kFD/kFD/lFCBDtD+4FD/jF9DB/+g6DYHVD/mF9DBBX9LDYHLD/kFCggEDtHMD9OAcAVD/mFALD/kFCTD+rFD9QHcANAVD/mFALD/kFCTD+rFAOAVD/mFALD/kFAMD9OD9QHVDQBFTtGEmYILPdKOenHLD8dBAEDBIBDyB+t+J83EBABCNJALD8dFAEDBIBDyF+t+J83EBAKAcAVDQNVi8ZcMpySQ8c8dfb8e8fHVD8dBAEDBIBDyG+t+J83EBABCiJAVD8dFAEDBIBDyE+t+J83EBABCAJRBAICIJHIAF9JQBMMM9jFF97GXAGCGrAF9sHG9FQBCBRFEXABABDBBBHECND+rFCND+sFD/6FAECiD+sFCnD+rFCUUU/8EDtD+uFD/mFDMBBABCTJRBAFCIJHFAG9JQBMMM9TFEaCBCB8oGUkUUBHFABCEJC98ZJHBjGUkUUBGXGXAB8/BCTWHGuQBCaREABAGlCggEJCTrXBCa6QFMAFREMAEMMMFBCUNMIT9tBB";
```
bghgary, line 9: Just curious, I've not seen this kind of wasm. Are these efficient in terms of file size?

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
```diff
@@ -0,0 +1,66 @@
...
+            var view = this._loader.loadBufferViewAsync(context, extensionDef);
+            var decoder = MeshoptDecoder;
+
+            extensionDef._decoded = Promise.all([view, decoder.ready]).then(function (res) {
+                var source = res[0] as Uint8Array;
+                var count = extensionDef.count;
+                var stride = extensionDef.byteStride;
+                var result = new Uint8Array(new ArrayBuffer(count * stride));
+
+                decoder.decodeGltfBuffer(result, count, stride, source, extensionDef.mode, extensionDef.filter);
```
bghgary, line 54: Is this synchronous? How long does this take? Should we use web workers?

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.d.ts
zeux: Ah, I should have looked at where Draco decoder goes.
I also see similar files in `dist/`. Is the idea that the contents of `dist/preview release` will be copied to `dist` at some point so I should just add the files to `dist/preview release`, not to both?

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.js
zeux: This is a pseudo-Base64 encoding described in https://github.com/zeux/meshoptimizer/commit/df03c2080eb9874f249487d8062cfd96e4ce7eb9. It's pretty efficient, and was designed as the best possible option of encoding Wasm binary blobs with good .js size and good size post gzip. It probably isn't a great fit for very large .wasm binaries where streaming compilation is likely to be superior from the CPU cost perspective, and the cost of an extra HTTP request is not very significant.

For example, stats for the SIMD version of the decoder, which is ~10KB of raw Wasm:

- decoder_simd.wasm 10080 bytes
- decoder_simd.wasm.gz 2911 bytes
- decoder_simd.base64 13440 bytes
- decoder_simd.base64.gz 4696 bytes
- decoder_simd.wasmpack 10800 bytes
- decoder_simd.wasmpack.gz 2934 bytes

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
zeux: This is synchronous; the decoders work at est. 1GB/s on a modern desktop CPU so should take tens of ms for pretty large meshes. It's possible to implement this as a web worker but given the speed it doesn't seem necessary for now at least, and it would make things a bit more complicated since you'd need to load the .js file into each worker and distribute the work.

Breakdown for Alien model in Playground example, where it takes ~1.7ms total to run the decoder, which is ~15% of the total parse time (the actual loading time is much larger here, as Mesh._refreshBoundingInfo takes ~100ms at some point later, and there's some other processing that also takes more time than raw glTF scene loading...

image

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.d.ts
zeux: ... actually, if I put the file to `dist/`, does this mean I can't just import the file from the .ts extension, and would have to resort to using a URL-based configuration, sourcing the file from `preview.babylonjs.com/meshopt_decoder.module.js`? I was hoping to be able to just import the module and let webpack figure it out from there...

bghgary: >  Is the idea that the contents of dist/preview release will be copied to dist at some point

Yup, `dist/preview release` will be copied to `dist` when we release.

> would have to resort to using a URL-based configuration

I believe so, yes. Ideally, we want this to be configurable anyways, so that people can host their own versions.

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.js
bghgary: That's neat. Is this documented somewhere so I can find out more info?

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
bghgary: Yeah, there is certainly overhead when using workers. Any idea what the perf is like for mobile? Either way, it's okay for this to be synchronous for now as long as the contract is async (which it is), so we're good.

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
```diff
@@ -0,0 +1,66 @@
...
+import { MeshoptDecoder } from "./meshopt_decoder.module.js";
+
+const NAME = "EXT_meshopt_compression";
+
+/**
+ * [Specification](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/EXT_meshopt_compression)
+ *
+ * This extension uses a WebAssembly decoder module from https://github.com/zeux/meshoptimizer/tree/master/js
+ */
+export class EXT_meshopt_compression implements IGLTFLoaderExtension {
```
bghgary, line 15: Is there a use case for using this compression scheme outside of glTF? Draco, for example, can be used by itself.

zeux: It's possible, for example Filament uses this compression algorithm in their custom binary mesh format. I'd propose though that for now we limit the scope here to just glTF extension support - particularly, the compression algorithms are currently not exposed through a nice JS library, so in a JS-only ecosystem right now there's no readily available way to compress the data short of using gltfpack as a full-scene processor.

loaders/src/glTF/2.0/Extensions/meshopt_decoder.module.js
zeux: The commit message above and the comment are all the information available atm. Sounds like I should write a blog post about it :)

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
zeux: Depends on the mobile device and whether Wasm SIMD is supported; I haven't done Wasm benchmarks on mobile. In general I would probably expect the rest of JS processing to be the bottleneck here (which is different with Draco), since the performance of Wasm code should scale according to the device power (e.g. it can easily be 100MB/s on a slow mobile CPU, but then I'd expect all the JS work to also be hit proportionally).

And yeah it's definitely possible to move this to a web worker right now, it's just that based on the testing I've done so far it doesn't seem to be necessary - if need arises, it should be straightforward to do.

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
zeux: (note that the JS library used here exposes `decodeVertexBuffer` etc, so it's not glTF specific - it just has a convenience wrapper for glTF)

bghgary: Ok, no problem. Just to be clear, if we were to do something, I'm thinking we can add another class under [src/Meshes/Compression](https://github.com/BabylonJS/Babylon.js/tree/master/src/Meshes/Compression) that then allows people to use this compression scheme without using glTF.

Playground/index-local.html
```diff
@@ -59,8 +59,9 @@
             BABYLONDEVTOOLS.Loader       
                 .require("index.js")
                 .load(() => {
+                    BABYLON.GLTF2.Loader.Extensions.EXT_meshopt_compression.DecoderPath = "../dist/preview%20release/meshopt_decoder.module.js";
```
bghgary, line 62: Huh.. Looks like we are missing some code here for the local PG.

```javascript
                    BABYLON.DracoCompression.Configuration.decoder = {
                        wasmUrl: "../../dist/preview%20release/draco_wasm_wrapper_gltf.js",
                        wasmBinaryUrl: "../../dist/preview%20release/draco_decoder_gltf.wasm",
                        fallbackUrl: "../../dist/preview%20release/draco_decoder_gltf.js"
                    };
                    BABYLON.GLTFValidation.Configuration = {
                        url: "../../dist/preview%20release/gltf_validator.js"
                    };
```

If you search for `dracoCompression.Configuration` in the code base, it would be good to add your DecoderPath to all the various places that should have it.

zeux: Yup - will do - I stole this from Sandbox, but I was using Playground for testing.

bghgary (approved):

PR #9380, opened 2020-11-10
[bug] Update texture.ts 

aWeirdo:
Texture._rowGenerationMatrix can be null while Texture._cachedTextureMatrix is not.

If a texture is applied as ReflectionTexture, then applied as another texture type, Texture._rowGenerationMatrix can be null,
throws error in Matrix.RotationYawPitchRollToRef() chain
>Cannot read property '_m' of null

https://www.babylonjs-playground.com/#2FDQT5#564


deltakosh:
Well all the tests are failing :)

deltakosh:
Thanks for the heads up I will add it the the next RC :D

PR #9387, opened 2020-11-10
Occlusion material

rgerd:
This PR introduces a material for situations where we want to occlude geometry without rendering to the color buffer. The motivating example for this material is using [surface reconstruction meshes](https://docs.microsoft.com/en-us/windows/mixed-reality/develop/unity/tutorials/holograms-230) generated by HoloLens 2 + [BabylonNative](https://github.com/BabylonJS/BabylonNative) to occlude scene geometry without impacting framerate on the [fillrate](https://en.wikipedia.org/wiki/Fillrate)-limited device.

[Here's a playground featuring a prototype version of this material.](https://www.babylonjs-playground.com/#BDU1ES#21)


deltakosh:
Hello in order to not grow the main lib, do you mind adding the material in the material lib folder?

rgerd:
> Hello in order to not grow the main lib, do you mind adding the material in the material lib folder?

That was my first instinct, but since we're going to use this inside of the XR module for AR occlusion I was thinking it would make more sense to put it in the core so XR doesn't need to take a dependency on the materials lib.

We are still thinking through the rendering story for AR occlusion though so I'm also ok with holding off on this until we have a clearer picture of how we want to do things.

deltakosh:
Makes total sense:)

deltakosh (approved):
It is ok to merge:) please update the what's new file in  /dist/preview release before

PR #9425, opened 2020-11-18
Creating a new Toggle GUI element

kintz09:
**Summary:**

This pull request creates a Toggle class, extended from the Button class. It can be used as an individual Toggle or with a group of toggles.

**Features**
`isActive` property with getter/setter to define if the toggle is active/on or inactive/off.

`toActiveAnimation` and `toInactiveAnimation` functions allow the developer to easily define the toggle's properties when in an active and inactive state. These functions fire when the `isActive` property changes.

`pointerEnterActiveAnimation`, `pointerOutActiveAnimation`, `pointerDownActiveAnimation`, `pointerUpActiveAnimation`, `pointerEnterInactiveAnimation`, `pointerOutInactiveAnimation`, `pointerDownInactiveAnimation`, `pointerUpInactiveAnimation` functions allow the developer to easily define separate pointer animations for the Active and Inactive states. These were implemented following Button.ts as a guide.

`group` property allows the developer to define the group the toggle belongs to. When a Toggle in a group is toggled on, the other toggles in the group will be toggled off. This was implemented with radioButton.ts as a guide.

`onIsActiveChangedObservable` an Observable raised when the `isActive` property value changes. This was implemented with radioButton.ts as a guide.

Note: The Project compiled (minified) for me, but failed toward the end of the build due to a memory issues in the GitPod.

[Original Discussion](https://forum.babylonjs.com/t/best-way-to-create-and-animate-custom-button-styles-like-for-a-disabled-state/15583)


kintz09:
Just an FYI, I didn't mean to include that function in `control.ts` in this pull request. Planned to do it in a separate request. I can remove it when I get feedback on this request.

deltakosh:
Well I'm a bit surprised to not see the draw function. Why not having a draw?

gui/src/2D/controls/toggle.ts
```diff
@@ -0,0 +1,228 @@
...
+    private _isActive = false;
+    private _group = "";
+
+    /**
+     * Function called to generate the toActive animation
+     */
+    public toActiveAnimation: () => void;
+
+    /**
+     * Function called to generate the toActive animation
```
kintz09, line 22: This comment should say 'Function called to generate the toInactive animation

kintz09:
I didn't see a draw function in the button class so it didn't even cross my mind. Could you elaborate on what you'd expect the draw function in the Toggle class would do?

Here's my notes after reviewing the draw functions in these classes:
_RadioButton_
- The radioButton class extends Control. Control has an empty `_draw()` function.
- It looks like the purpose of the radioButton's draw function is to create the base shape since that's not handled by Control.

_Toggle_
- The Toggle extends Button. Button doesn't have its own draw function.
- Button extends Rectangle. Rectangle doesn't have a draw function, but does have a protected `_localDraw()`.
- Rectangle extends Container. Container has a `_localDraw()` function and a `_draw()` function.
- Container extends Control. Control doesn't have a `_localDraw() function but does have the empty `_draw()` function.
- It appears that a the Container class handles the function for a Button?

If you don't mind educating me here, I'd be happy to try updating this pull request.

deltakosh:
Well I thought the toggleButton will have its own rendering mode (something that indicates the current state of the toggle right?)
So in a sense the button has no draw as it is really generic and works as a container as you said (because we could have text button or image button)

For the toggle I was really thinking about it like a "pure" control that needs to render itself

kintz09:
This is a little beyond my expertise, but I am a quick learner so I'm happy to try again. I would certainly find value in being able to contribute to this project. 😃 

For clarity sake for anyone else reading this, what I'm trying to build is a Toggle Button and not a [Toggle Switch](https://uxmovement.com/mobile/stop-misusing-toggle-switches/#:~:text=It's%20easy%20for%20designers%20to,buttons%20are%20for%20contextual%20ones.)

So if I'm understanding you correctly, you weren't looking for the Toggle class to extend Button but be a base control. If this is correct, would I need to rework the class by extending from Control? Or would it be acceptable to extend from Container or Rectangle to avoid reimplementing properties like thickness, cornerRadius,adaptWidthToChildren, adaptHeightToChildren, etc.?

deltakosh:
Ok I was more on the toggle switch side of the house :)

So I guess you have a good idea here.  It deserves to be merge

would you be ok to update the doc as well to include examples of your control running in the PG?

kintz09:
Ah, ok, I'm glad I clarified. I can certainly fix the doc and put together some playground examples.

I do have one other thought first...  Would it be wiser for Toggle to extend Rectangle instead of Button?

- I found it weird the Toggle class exposed the static functions from Button. Mostly I found it confusing because the functions all have the word 'button' in it, but also, they currently create a button not a toggle.
- The Toggle class extended from Button still has the standard `pointerDownAnimation` properties which duplicates the toggle's pointer animation properties which seems a bit confusing unless you've read the code
- Button is pretty simple, so recreating its properties in Toggle is little effort
- A Toggle class extending Rectangle could have its own static functions to quickly create various types of toggles.

What do you think? Should Toggle extend Rectangle or Button?

deltakosh:
1. Yeah you are right! that sounds a better plan

Let's try that approach?

kintz09:
Sound like a plan. Should I reuse this pull request or start with a fresh one?

deltakosh:
Maybe start a new one

PR #9427, opened 2020-11-18
Handedness and probes.

sebavan:
Trying to address https://forum.babylonjs.com/t/reflection-probes-cubetexture-doesnt-respect-right-handed-system/15820


bghgary (commented):
I don't know this code well, but it looks right.

bghgary (approved):

Issue #9238, opened 2020-10-21
Add ability to make an input node's properties visible in the properties of a custom frame


PirateJC:
We currently have an option to allow inputs to be visible in the inspector:
![image.png](https://images.zenhubusercontent.com/5dcb34efa1442a0001945170/610c4ccf-cfd5-473a-93be-4649eb067944)

We should had the ability to let the same properties be visible to the property list when a custom frame is selected.

Say we have a custom frame that looks like this:
![image.png](https://images.zenhubusercontent.com/5dcb34efa1442a0001945170/8380a197-3251-4286-8fdb-96d2ea0d6930)

When you select the top left float node you see these properties:
![image.png](https://images.zenhubusercontent.com/5dcb34efa1442a0001945170/c7b3ce27-4bb2-4457-afd1-2e5b414d9299)

When you collapse the frame and select it, the properties window looks like this:
![image.png](https://images.zenhubusercontent.com/5dcb34efa1442a0001945170/76e80c4a-9d9f-4f3f-930f-f68f7f8b1872)

It would be amazing if we could allow the user to make the float properties visible to the collapsed frame. So if I have a collapsed frame, I could select it and have access to specified properties within the frame.

@PatrickRyanMS  - Just adding you in case there is any design considerations.

PR #9444, opened 2020-11-20
[NME] Adding ability to make input node's properties visible in the properties of a custom frame

msDestiny14:
The visibleOnFrame setting is also serialized.
MicrosoftTeams-image

https://github.com/BabylonJS/Babylon.js/issues/9238


nodeEditor/src/diagram/properties/framePropertyComponent.tsx
```diff
@@ -31,7 +39,116 @@ export class FramePropertyTabComponent extends React.Component<IFramePropertyTab
             this.props.frame.onExpandStateChanged.remove(this.onFrameExpandStateChangedObserver);
             this.onFrameExpandStateChangedObserver = null;
         }
-    }    
+    }
+    
+    renderInputBlock(block: InputBlock) {
```
deltakosh, line 44: I bet this can be shared with what exists in PropertyTabComponent.tsx (Like having a common component to avoid code duplication)

msDestiny14:
New PR now here: https://github.com/BabylonJS/Babylon.js/pull/9446

PR #9446, opened 2020-11-20
[NME] Adding ability to make input node's properties visible in the properties of a custom frame

msDestiny14:
MicrosoftTeams-image

https://github.com/BabylonJS/Babylon.js/issues/9238


msDestiny14:
@deltakosh I agree with not reusing code so I mode the repeated code to a stand alone file and used it in both places.

deltakosh (approved):
LGTM

PR #9450, opened 2020-11-21
KTX2 loader: sets the hasAlpha property of the texture

Popov72:
Automatically sets the `Texture.hasAlpha` property if the loaded file has an alpha channel.

Also, configure the local environment to point to the local ktx2Decoder files.


sebavan:
Would there be a way to not move HasAlpha on the internal texture ? This was originally a user set flag to turn alpha test on and not meant to be filled by what is actually in the texture. The texture type should provide us with the underlying data ?

Popov72:
All the underlying loading code deals only with internal textures, so I don't know how to bubble this data up until we have a Texture available...

Note that having `hasAlpha` on the internal texture is not a breaking change and the user can still update the value as they see fit (meaning setting it to true even if the texture has no alpha channel and setting it to false even if the texture has an alpha channel). It's the same change that I already did for `gammaSpace`, which was a `BaseTexture` property at start and which was moved to `InternalTexture` but without breaking things.

sebavan:
Ok so all good to me :-) I ll let @bghgary review and merge it in

Playground/index-local.html
```diff
@@ -55,6 +55,12 @@
         <div id="host-element">
         </div>
         <script>
+            function GetAbsoluteUrl(url) {
+                const a = document.createElement("a");
+                a.href = url;
+                return a.href;
+            }
+
```
bghgary, line 63: I know this is copied this from elsewhere, but I just noticed that this won't work in Babylon Native. We will need to do something different here and elsewhere.

tests/validation/config.json
```diff
@@ -921,7 +921,8 @@
         },
         {
             "title": "KTX2 decoder test",
-            "playgroundId": "#EIJH8L#1",
+            "playgroundId": "#EIJH8L#26",
+            "renderCount": 5,
```
bghgary, line 925: Why do you need a render count?

Popov72: I don't remember why, it works without it - I have removed it.

sebavan:
I guess it is now all good to merge @Popov72 ???

Popov72:
Ok for me!

PR #9451, opened 2020-11-21
Add ability to view images (ktx2, png, jpg) to the sandbox

bghgary:
This adds the ability to drag and drop image files to the sandbox which creates a plane to host the image. It also opens the inspector to the texture preview to quickly be able to open the texture inspector with the Edit button.

image

Also fixes some bugs found during development/testing:
- Clear glTF validation results for all loaders instead of only when loading glTF files
- Fix missing light when loading non-PBR scenes.

### DIFF
--- a/sandbox/src/components/renderingZone.tsx
+++ b/sandbox/src/components/renderingZone.tsx
@@ -12,7 +12,7 @@ import { EnvironmentTools } from "../tools/environmentTools";
 import { Tools } from "babylonjs/Misc/tools";
 import { FilesInput } from "babylonjs/Misc/filesInput";
 import { Animation } from "babylonjs/Animations/animation";
-import { PBRMaterial, StringTools, Texture } from "babylonjs";
+import { PBRBaseMaterial, PBRMaterial, StringTools, Texture } from "babylonjs";
 import { Mesh } from "babylonjs/Meshes/mesh";
 
 require("../scss/renderingZone.scss");
@@ -213,8 +213,8 @@ export class RenderingZone extends React.Component<IRenderingZoneProps> {
             }
         } else {
             var pbrPresent = false;
-            for (var i = 0; i < this._scene.materials.length; i++) {
-                if (this._scene.materials[i].transparencyMode !== undefined) {
+            for (const material of this._scene.materials) {
+                if (material instanceof PBRBaseMaterial) {
                     pbrPresent = true;
                     break;
                 }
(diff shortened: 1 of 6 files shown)


sebavan (commented):
LGTM, adding @RaananW for a second pair of eyes

sebavan (approved):

sandbox/src/components/renderingZone.tsx
```diff
@@ -22,8 +24,18 @@ interface IRenderingZoneProps {
     expanded: boolean;
 }
 
+function isTextureAsset(name: string): boolean {
```
RaananW, line 27: This will fail if the texture is loaded with query param (i.e. `https://url/image.jpg?atag=123456`). Are those being filtered? might be safer to use contains rather than ends with

sandbox/src/components/renderingZone.tsx
```diff
@@ -176,8 +213,8 @@ export class RenderingZone extends React.Component<IRenderingZoneProps> {
             }
         } else {
             var pbrPresent = false;
-            for (var i = 0; i < this._scene.materials.length; i++) {
-                if (this._scene.materials[i].transparencyMode !== undefined) {
+            for (const material of this._scene.materials) {
+                if (material instanceof PBRBaseMaterial) {
```
RaananW, line 217: instanceof is relatively expensive (and sometimes unreliable). Is there a different way of checking its type? will `getClassName()` work here?

sandbox/src/components/renderingZone.tsx
RaananW: (note - if it is just for drag-and-drop it should be perfectly fine the way it is)

bghgary: I suppose it could have query param. It's an unlikely scenario though. I'll check for the `?`. Using contains isn't perfect either since the query param could have `.<ext>` in them.

sandbox/src/components/renderingZone.tsx
bghgary: I'm sure there is. I don't like it because that is what caused the bug to begin with. We checked for transparencyMode on PBR, but later we added transparencyMode to StandardMaterial.

`getClassName` can work, but we would have to check a few strings since we want all materials inherited from `PBRBaseMaterial`.

I'm not too worried about the expensive part since this only runs once. How is it unreliable?

RaananW: it checks prototypes. A class with the same prototypes will return true. Rare but possible :-)

bghgary: Yes, I agree, but if it has the same prototype, it's probably okay?

PR #9463, opened 2020-11-23
Creating a new ToggleButton GUI control

kintz09:
**Summary:**

[Original Pull Request](https://github.com/BabylonJS/Babylon.js/pull/9425)

This pull request creates a ToggleButton class, extended from the Rectangle class. It functions as an individual Toggle or as a toggle group. The design of the class followed the Button class, as this is just another button type with some unique properties to support the use as a toggle button.

**Features**
`isActive` property with getter/setter to define if the toggle is active/on or inactive/off.

`toActiveAnimation` and `toInactiveAnimation` functions allow the developer to easily define the toggle's properties when in an active and inactive state. These functions fire when the `isActive` property changes.

`pointerEnterActiveAnimation`, `pointerOutActiveAnimation`, `pointerDownActiveAnimation`, `pointerUpActiveAnimation`, `pointerEnterInactiveAnimation`, `pointerOutInactiveAnimation`, `pointerDownInactiveAnimation`, `pointerUpInactiveAnimation` functions allow the developer to easily define separate pointer animations for the Active and Inactive states. These were implemented following Button.ts as a guide.

`group` property allows the developer to define the group the toggle belongs to. When a Toggle in a group is toggled on, the other toggles in the group will be toggled off. This was implemented following radioButton.ts as a guide.

`onIsActiveChangedObservable` an Observable raised when the `isActive` property value changes. This was implemented with radioButton.ts as a guide.

I attached the typescript playground code I used to test the class which can also be used as an example.
[toggleButton-playground-code.txt](https://github.com/BabylonJS/Babylon.js/files/5586074/toggleButton-playground-code.txt)

[Original Discussion](https://forum.babylonjs.com/t/best-way-to-create-and-animate-custom-button-styles-like-for-a-disabled-state/15583)


sebavan:
Looks pretty great, @Popov72  will make a deeper review of it, and we might wait on @deltakosh next week for the final review if necessary.

gui/src/2D/controls/toggleButton.ts
```diff
@@ -0,0 +1,308 @@
...
+     * Returns the image part of the button (if any)
+     */
+    public get image(): Nullable<Image> {
+        return this._image;
+    }
+
+    private _textBlock: Nullable<TextBlock>;
+    /**
+     * Returns the image part of the button (if any)
+     */
```
Popov72, line 79: Wrong doc (copy/paste of the previous one)

gui/src/2D/controls/toggleButton.ts
```diff
@@ -0,0 +1,308 @@
...
+        this.onIsActiveChangedObservable.notifyObservers(value);
+
+        if (this._isActive && this._host) {
+            // Update all controls from same group
+            this._host.executeOnAllControls((control) => {
+                if (control == this) {
+                    return;
+                }
+
+                if ((<any>control).group === undefined) {
```
Popov72, line 128: Maybe you can get use `_getTypeName() === "ToggleButton"` instead of testing for the `group` property (in case one day other control(s) define a `group` property)?

gui/src/2D/controls/toggleButton.ts
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        this._isActive = value;
+
+        // Update the visual state based on the new value
+        if (this._isActive) {
+            this.toActiveAnimation();
+        } else {
+            this.toInactiveAnimation();
+        }
```
Popov72, line 115: You test event handlers for not being undefined further down, so I guess you should also do it here.

Popov72 (changes requested):
LGTM, only some small comments.

gui/src/2D/controls/toggleButton.ts
kintz09: Would an optional chaining operator be acceptable in the BabylonJS project? 

_The traditional way:_
`if (this._isActive) {
    if (this.toActiveAnimation) {
        this.toActiveAnimation();
    }
} else {
    if (this.toInactiveAnimation) {
        this.toInactiveAnimation()
    }
}`

_Using Optional Chaining:_
`if (this._isActive) {
    this.toActiveAnimation.?();
} else {
    this.toInactiveAnimation.?();
}`

From the [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), it looks safe to use, but wasn't sure if there were compatibility concerns.

sebavan: should be good as transpiled by ts anyway

gui/src/2D/controls/toggleButton.ts
kintz09: I did a rewrite of this function utilizing `_getTypeName()` and tried to simplify the logic. Hopefully the comments make sense, but if not let me know and I can give you pseudocode to review.

gui/src/2D/controls/toggleButton.ts
kintz09: This should be resolved in the latest update to this pull request.

gui/src/2D/controls/toggleButton.ts
kintz09: This should be resolved in the latest update to this pull request.

gui/src/2D/controls/toggleButton.ts
```diff
@@ -0,0 +1,310 @@
...
+        this._markAsDirty();
+
+        this.onIsActiveChangedObservable.notifyObservers(value);
+
+        if (this._isActive && this._host && this._group) {
+            // A toggle button in a group can only have 1 active element at a given time.
+            // If this toggle button has a group, set other toggle buttons in the group to inactive.
+            this._host.executeOnAllControls((control) => {
+                // Check for control type ToggleButton
+                if ((<any>control)._getTypeName === "ToggleButton") {
```
Popov72, line 126: Something like this should work:
```javacsript
if (control.typeName === "ToggleButton") {
```
as `typeName` is a property of the base class `Control`.

kintz09: Ah, that feels much cleaner. This pull request has been updated with this change.

sebavan:
Looks overall good to me but as we are trying to merge a huge change in today, is that a problem if we wait for @deltakosh on Monday to validate the PR ???

kintz09:
Not sure if you're asking me, but if so, there's no urgency to get this merged.

sebavan:
Yup that was exactly my question, thanks a lot.

deltakosh (approved):

PR #9464, opened 2020-11-24
First try symlinked shared ui components

sebavan:
Allow sharing visual components across the various editors.


RaananW (approved):
You know how much I like this :-)

LGTM. 

I assume the symlink lib knows how to deal with an already-existing link (technically a double call of this function - `generateSharedUiComponents`).

PR #9469, opened 2020-11-24
less choppy movement for navigation agents

CedricGuillemet:
Follow up https://forum.babylonjs.com/t/how-to-use-recastjsplugin/8256/22
more consistent sub time steps with provided delta time instead of fixed delta time.


RaananW (approved):
Just need to correct linting. I would also calculate 

```javascript
deltaTime/iterationCount
```

beforehand to avoid constant division

PR #9470, opened 2020-11-24
Fix ArcRotateCamera panning with axis decomposition

CedricGuillemet:
Follow up on https://forum.babylonjs.com/t/arcrotatecamera-does-not-work-at-beta-0-with-panningaxis-set-to-1-0-1/15850
Avoid the gimbal lock and 0lengthed axis by decomposing vectors and blending them with intended axis.

### DIFF
--- a/src/Cameras/arcRotateCamera.ts
+++ b/src/Cameras/arcRotateCamera.ts
@@ -892,7 +892,7 @@ export class ArcRotateCamera extends TargetCamera {
             }
 
             this._viewMatrix.invertToRef(this._cameraTransformMatrix);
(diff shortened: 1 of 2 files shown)


src/Cameras/arcRotateCamera.ts
```diff
@@ -891,14 +891,20 @@ export class ArcRotateCamera extends TargetCamera {
                 this._transformedDirection = Vector3.Zero();
```
Popov72, line 891: I think `_localDirection` is not used anymore, you should be able to remove it.

Else LGTM (trusting you on the math, I have not really understood them!).

sebavan:
I am just waiting on the test to merge and make the nightly :-)

PR #9474, opened 2020-11-24
meshopt fixes and misc cleanup

bghgary:
The main fix is to stop using `import` in EXT_meshopt_compression which doesn't work in Babylon Native scenarios. Also added a validation test for this extension. The rest are changes to make everything more consistent.


sebavan (approved):
LGTM

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
```diff
@@ -23,59 +30,54 @@ export class EXT_meshopt_compression implements IGLTFLoaderExtension {
...
     }
 
     /** @hidden */
     public dispose() {
+        (this._loader as any) = null;
+        delete this._decoderPromise;
     }
 
     /** @hidden */
     public loadBufferViewAsync(context: string, bufferView: IBufferView): Nullable<Promise<ArrayBufferView>> {
```
RaananW, line 60: I noticed you removed the `return null;` section (which is great!!) this means that the return signature can be changed, and that this function can be converted to an async/await function (if it makes sense).

RaananW (approved):
LGTM! with just a small comment

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
```diff
@@ -23,59 +30,54 @@ export class EXT_meshopt_compression implements IGLTFLoaderExtension {
...
-                return MeshoptDecoder.ready.then(() => MeshoptDecoder);
+                return MeshoptDecoder.ready;
             });
         }
     }
 
     /** @hidden */
     public dispose() {
+        (this._loader as any) = null;
+        delete this._decoderPromise;
```
Popov72, line 56: What does calling `delete` on a promise do?

loaders/src/glTF/2.0/Extensions/EXT_meshopt_compression.ts
bghgary: The helper function `LoadExtensionAsync` will still return null.

bghgary: Also, we don't want to use async/await because we target es5.

RaananW: Typescript takes care of the conversion to es5. Use async await as much as you can

sebavan: actually one of the polyfill for generators is not in for ie (I guess it is only ie) that using async await would require if I remember well the issue.

bghgary: it also generates a whole ton of code (bad for code size) and is super hard to debug in js without the map file

sebavan: that was why we were not relying on it in the code.

bghgary: I wish we can use async/await. It would simplify the code a lot.

sebavan: OHHHHH YEESSSSS ME TOOO !!!!

PR #9481, opened 2020-11-25
added support for orthographic cameras to ssao2RenderingPipeline

Kesshi:
With the help of @CraigFeldspar i was able to fix this issue: https://forum.babylonjs.com/t/ssao-bug-with-othographic-cameras/12350

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -45,6 +45,7 @@
 - Fix an issue with keyboard control (re)attachment. ([#9411](https://github.com/BabylonJS/Babylon.js/issues/9411)) ([RaananW](https://github.com/RaananW))
 - Fix issue where PBRSpecularGlossiness materials were excluded from export [#9423](https://github.com/BabylonJS/Babylon.js/issues/9423)([Drigax](https://github.com/drigax))
 - Fix direct loading of a glTF string that has base64-encoded URI. ([bghgary](https://github.com/bghgary))
+- Fix ssao2RenderingPipeline for orthographic cameras ([Kesshi](https://github.com/Kesshi))
 
 ## Breaking changes
 
--- a/src/PostProcesses/RenderPipeline/Pipelines/ssao2RenderingPipeline.ts
+++ b/src/PostProcesses/RenderPipeline/Pipelines/ssao2RenderingPipeline.ts
@@ -401,7 +401,7 @@ export class SSAO2RenderingPipeline extends PostProcessRenderPipeline {
             [
                 "sampleSphere", "samplesFactor", "randTextureTiles", "totalStrength", "radius",
                 "base", "range", "projection", "near", "far", "texelSize",
-                "xViewport", "yViewport", "maxZ", "minZAspect"
+                "xViewport", "yViewport", "maxZ", "minZAspect", "orthoProjection"
             ],
             samplers,
             ratio, null, Texture.BILINEAR_SAMPLINGMODE,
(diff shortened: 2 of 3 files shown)


CraigFeldspar:
GG ! Just a small performance overhead, can you use a `#define` instead of a dynamic test in shader with the uniform `orthoProjection` ?

Thanks !

Kesshi:
> GG ! Just a small performance overhead, can you use a `#define` instead of a dynamic test in shader with the uniform `orthoProjection` ?
I was thinking about that but a camera can be switched dynamically between perspective and orthographic (this is also the case in my app). I'm not sure how to automatically update the defines if the camera mode gets changed.

Kesshi:
One idea would be to use the `onActivateObservable` of the post process and update the effect if the camera mode was changed.
Is it ok to call `updateEffect` inside the `onActivateObservable` handler? What happens if the pipeline is used for an perspective and orthographic camera at the same time? `updateEffect` would be called multiple times per frame then. Is that a problem?

sebavan (changes requested):
Flagging as request changes to prevent wild merge :-)

sebavan:
@CraigFeldspar will provide more guidance soon

CraigFeldspar:
Actually it requires a little bit of work, you would need an `updateEffect` function, like in `motionBlurPostProcess.ts` : 

```
    /**
     * Called on the effect must be updated (changed mode, samples count, etc.).
     */
    private _updateEffect(): void {
        if (this._geometryBufferRenderer || this._prePassRenderer) {
            const defines: string[] = [
                "#define GEOMETRY_SUPPORTED",
                "#define SAMPLES " + this._motionBlurSamples.toFixed(1),
                this._isObjectBased ? "#define OBJECT_BASED" : "#define SCREEN_BASED"
            ];

            this.updateEffect(defines.join("\n"));
        }
    }
```

Then in `onApply` you check what kind of camera is used, and call `updateEffect` accordingly.

Do you think you can do that ? that would be really awesome :)

Kesshi:
I changed it on my side as you wrote and it works but `updateEffect` does not update the effect immediately. Atleast not if its called in onApply(). There seems to be a delay of one frame. That also means it will not work if someone uses a perspective and an orthographic camera at the same time in one pipeline.
You can see the delay here:

ezgif-6-89db58a2be9c

If only the branching is a problem in my current solution then i could do something like this to avoid branching (not tested yet):
```
vec3 vViewRay = vec3((vUV.x * 2.0 - 1.0)*xViewport, (vUV.y * 2.0 - 1.0)*yViewport, depthSign);
float projectionDepthFactor = depth * float(1 - orthoProjection) + float(1 * orthoProjection);
vec3 origin = vec3(vViewRay.x * projectionDepthFactor, vViewRay.y * projectionDepthFactor, vViewRay.z * depth);
```

CraigFeldspar:
I think you are right and it becomes unnecessarily complicated.
The most elegant solution in my opinion would be to build a 3*3 matrix depending on the camera, then you multiply it with a vec3(x, y, depth) to get the pixel position.

It would ressemble something like this for an ortho camera : 
( ..   ...   ... )   ( x )
(...   ...  ...  )   ( y )
( 0   0   ... )   ( depth )

and non-zero coefficients on the bottom line for a perspective.

Kesshi:
> I think you are right and it becomes unnecessarily complicated.
> The most elegant solution in my opinion would be to build a 3*3 matrix depending on the camera, then you multiply it with a vec3(x, y, depth) to get the pixel position.

Done

sebavan:
LGTM I ll merge tomorrow after confirmation from @CraigFeldspar if no fire are starting due to our previous merge :-)

sebavan:
And it actually looks great now !!! GG @Kesshi , I bet @CraigFeldspar will love it

CraigFeldspar:
yep, @Kesshi great job ! and thanks

PR #9490, opened 2020-11-26
Fix crash of some node materials using instances on iOS

Popov72:
See https://forum.babylonjs.com/t/compiling-node-materials-with-instances-on-safari/16054/9


src/Materials/Node/Blocks/Vertex/instancesBlock.ts
```diff
@@ -175,7 +176,11 @@ export class InstancesBlock extends NodeMaterialBlock {
         state.compilationString += `#ifdef THIN_INSTANCES\r\n`;
         state.compilationString += `${output.associatedVariableName} = ${this.world.associatedVariableName} * ${output.associatedVariableName};\r\n`;
         state.compilationString += `#endif\r\n`;
-        state.compilationString += this._declareOutput(instanceID, state) + ` = float(gl_InstanceID);\r\n`;
+        if (Engine.LastCreatedEngine && !Engine.LastCreatedEngine._caps.canUseGLInstanceID) {
```
sebavan, line 179: Can we instead access it from the scene ? if would prefer to not rely on the lastCreated one ?

Popov72: The problem is that I have no scene / engine instance available in the class... That's why I used the last created engine.

sebavan: Wondering if are any way to pass the information in ? do not worry if it is too complex ?

Popov72: Ok, I was able to find the scene somewhere :)

sebavan (approved):

sebavan:
LGTM I ll merge it tomorrow with the latest PRs :-)

PR #9519, opened 2020-12-01
Support "#include <otherShader>" and "#include<otherShader>" (the space character)

Jarred-Sumner:
Currently, if your shader has a space between `#include` and `<filename>` Babylon.js will throw an error when processing the shader.

This is a little frustrating if you're using `clang-format` to format GLSL for you because it inserts the space automatically, which breaks the shader. 

So I made the regex support an optional space character.

I tested this by manually overriding this function with the diff below in my codebase.

I also wonder if `sourceCode` needs to be copied in `returnValue` on line 307, since strings are immutable.


deltakosh (approved):
Good one thanks!

Jarred-Sumner:
np

PR #9553, opened 2020-12-04
Camera fixes for glTF loader and sandbox

bghgary:
This fixes a couple of issues with camera handling in glTF loader.

- glTF cameras with [infinite projection](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#infinite-perspective-projection) now works correctly.
- The skybox will now render even if the camera's maxZ is smaller than the size of the skybox.

### DIFF
--- a/loaders/src/glTF/2.0/glTFLoader.ts
+++ b/loaders/src/glTF/2.0/glTFLoader.ts
@@ -4,7 +4,7 @@ import { Quaternion, Vector3, Matrix } from "babylonjs/Maths/math.vector";
 import { Color3 } from 'babylonjs/Maths/math.color';
 import { Tools } from "babylonjs/Misc/tools";
 import { Camera } from "babylonjs/Cameras/camera";
-import { TargetCamera } from "babylonjs/Cameras/targetCamera";
+import { FreeCamera } from "babylonjs/Cameras/freeCamera";
 import { AnimationGroup } from "babylonjs/Animations/animationGroup";
 import { Animation } from "babylonjs/Animations/animation";
 import { Bone } from "babylonjs/Bones/bone";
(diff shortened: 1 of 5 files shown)


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1189,7 +1189,7 @@ export class GLTFLoader implements IGLTFLoader {
         this.logOpen(`${context} ${camera.name || ""}`);
 
         this._babylonScene._blockEntityCollection = this._forAssetContainer;
-        const babylonCamera = new FreeCamera(camera.name || `camera${camera.index}`, Vector3.Zero(), this._babylonScene, false);
+        const babylonCamera = new TargetCamera(camera.name || `camera${camera.index}`, Vector3.Zero(), this._babylonScene, false);
```
deltakosh, line 1192: why targetcamera? It is supposed to be the parent of free and arcrotate

deltakosh: Plus it is a breaking change

src/Helpers/sceneHelpers.ts
```diff
@@ -179,21 +179,21 @@ Scene.prototype.createDefaultSkybox = function(environmentTexture?: BaseTexture,
...
         if (skyboxMaterial.reflectionTexture) {
             skyboxMaterial.reflectionTexture.coordinatesMode = Texture.SKYBOX_MODE;
         }
         skyboxMaterial.disableLighting = true;
-        hdrSkybox.infiniteDistance = true;
         hdrSkybox.material = skyboxMaterial;
     }
     hdrSkybox.isPickable = false;
+    hdrSkybox.infiniteDistance = true;
+    hdrSkybox.ignoreCameraMaxZ = true;
```
bghgary, line 196: Are we okay with this? It will make all default skyboxes have additional perf cost.

deltakosh: Yes it is ok as we have only one skybox.

loaders/src/glTF/2.0/glTFLoader.ts
bghgary: I'm going to revert this for now. Creating a free camera this way doesn't work right because the transform of the camera is on a parent transform node instead of the camera and thus the panning controls for the free camera is wrong if the camera transform has rotation in the glTF. This change was trying to prevent the camera from moving.

deltakosh (approved):

PR #9557, opened 2020-12-05
Add a smoothing threshold to smooth the sides of an extruded polygon.

snagy:
Smooth the normals above a certain dotp threshold between adjacent vertex sides.

If the dot product of the normals of two adjacent side polygons is greater than this threshold, the normal on that edge are set to the average of the two.


src/Meshes/polygonMesh.ts
```diff
@@ -239,9 +241,10 @@ export class PolygonMeshBuilder {
     /**
      * Creates the polygon
      * @param depth The depth of the mesh created
+     * @param smoothingThreshold Dot product threshold for smoothed normals
      * @returns the created VertexData
      */
-    buildVertexData(depth: number = 0): VertexData {
+    buildVertexData(depth: number = 0, smoothingThreshold: number = 2): VertexData {
```
deltakosh, line 247: we should have a default value that behaves like before (is it the case here?)

src/Meshes/polygonMesh.ts
```diff
@@ -310,43 +313,78 @@ export class PolygonMeshBuilder {
...
+            vc = vc.normalizeFromLength(vc_len);
+
+            const p0: IndexedVector2 = points.elements[(i + points.elements.length - 1) % points.elements.length];
+            const p2: IndexedVector2 = points.elements[(i + 2) % points.elements.length];
+            let vp = new Vector3(-(p.y - p0.y), 0,  p.x - p0.x);
+            const vp_len = vp.length();
+            vp = vp.normalizeFromLength(vp_len);
+            let vn = new Vector3(-(p2.y - p1.y), 0, p2.x - p1.x);
+            const vn_len = vn.length();
+            vn = vn.normalizeFromLength(vn_len);
```
deltakosh, line 339: why not simply normalize?

src/Meshes/polygonMesh.ts
```diff
@@ -310,43 +313,78 @@ export class PolygonMeshBuilder {
...
+
+            if (!flip) {
+                vp = vp.scale(-1);
+                vn = vn.scale(-1);
+                vc = vc.scale(-1);
+            }
+
+            const dotp = Vector3.Dot(vp, vc);
+            if (dotp > smoothingThreshold) {
```
deltakosh, line 348: ok gotcha, so the dotP should never be greater than 2 so 2 is a good default parameter :)

snagy: yep!

src/Meshes/polygonMesh.ts
snagy: I was going to weigh the result by the edge length of the two sides and forgot.  I just pushed a PR with that change, but done a little bit cleaner.

deltakosh (approved):

PR #9561, opened 2020-12-06
rotate3dBlock

Pryme8:
Can we have this please?

Its really useful and I tested it on my local seems to work well!


Pryme8:
I added blocks for  x*0.5+0.5 and x*2.0-1.0 as well... cause it was getting annoying making frames for those.

deltakosh:
I'm not convinced honestly. 

Some questions: 
- What is the difference between ROtate3D and Transform (existing block?)
- zeroOneblock can be done with normalize
-zeroOneInverseBlock cane be done with remap

Pryme8:
How would I make the matrix to rotate the vector by with the transform block?

Normalize only accepts vectors and I just found remap, so the inverse is not necessary.

deltakosh:
No problem buddy!

PR #9568, opened 2020-12-08
First iteration of WebGPU cache implementation

Popov72:
For the time being, I have created a separate `WebGPUEngineCache` class that implements the WebGPU interface with the caching mechanism enabled.

This way, we can test things in the "no caching" engine (`WebGPUEngine`) in case of something failing with the cache...

Once we consider everything is ok I will remove `WebGPUEngineCache` and will make `WebGPUEngine` the only one engine, with the caching mechanism.

[...] After some after thoughts, maybe `WebGPUEngineCache` should be renamed `WebGPUEngine` and `WebGPUEngine` something like `WebGPUEngineNoCache`... What do you think?


sebavan (commented):
I am just wondering if we should have both Engines for WebGPU ?

Popov72:
In my mind we only had the two engines at start to test things in case the "cache engine" would fail... But after some time we would remove it.

So, in that regard, having `WebGPUEngine` being the "cache engine" could be the best way to go as in the end it would be the only one engine remaining.

sebavan:
Ok but in this case we need to have nothing to merge in this code in between ???

Popov72:
Indeed...

So, going to remove the "no cache" version and keep only the other one. We are still able to disable caching by using `disableCacheRenderPipelines = true` if necessary.

sebavan (approved):

deltakosh:
>  will make WebGPUEngine the only one engine

Yes please

I'm really NOT for having two engines

PR #9569, opened 2020-12-08
Small tidy up around Camera Inputs Base classes.

mrdunk:
This fixes the following:

1. `BaseCameraMouseWheelInput` and `BaseCameraPointersInput` appear in the API documentation but are not available for include as they are not in `Camera/Inputs/index.ts`.
When i originally created these i presumed they didn't really need to be visible to the outside world but since then i've tried to use `BaseCameraPointersInput`. If i tried, maybe it's of use to others too. Also it's in the API documentation so needs to be available.
2. `FreeCameraMouseWheelInput._updateCamera()` creates the function `updateCameraProperty(...)` every time it is called.
Do i understand the JS GC correctly to say it is not smart enough to re-use the allocation and that this causes unnecessary GC load?


mrdunk:
@PolygonalSun 
I think camera inputs is your area right?

PolygonalSun:
> @PolygonalSun
> I think camera inputs is your area right?

Yeah, thanks for the tag.  I'll take a look at things

PolygonalSun:
Looks good and works fine

PolygonalSun (approved):

deltakosh (approved):

mrdunk:
thanks!

PR #9572, opened 2020-12-09
Fix uv set selection for textures in GBuffer

ladevieq:
Hello BabylonJS folks

Here is a small PR to fix a bug encountered using SSAO on a scene with meshes having 2 uv sets.

The old code doesn't choose for each textures which uv set to use.

Now the set to use for textures is defined using the `coordinatesIndex` of the texture.


deltakosh:
Pinging @CraigFeldspar  for review as well

deltakosh:
Also webgl1 test is failing

deltakosh (approved):

sebavan (approved):

CraigFeldspar (approved):

PR #9575, opened 2020-12-09
Avoid changing index winding order when fixing coordinate system handedness for WebXRMeshDetector

chrisfromwork:
There appears to be inconsistency in what winding order is used for left/right handed coordinate systems. WinRT APIs use right handed coordinate systems + clockwise winding. OpenXR uses right handed coordinate systems + counter clockwise winding. Until WebXR has a formalized winding order, we should avoid attempting to fix winding order based on the Babylon.js coordinate system used.

### DIFF
--- a/src/XR/features/WebXRMeshDetector.ts
+++ b/src/XR/features/WebXRMeshDetector.ts
@@ -55,8 +55,8 @@ export interface IWebXRVertexData {
      */
     positions?: Float32Array;
     /**
-     * An array of indices in babylon space. right/left hand system is taken into account.
-     * Indices will only be calculated if convertCoordinateSystems is set to true in the IWebXRMeshDetectorOptions.
+     * An array of indices in babylon space. Indices have a counterclockwise winding order.
+     * Indices will only be populated if convertCoordinateSystems is set to true in the IWebXRMeshDetectorOptions.
      */
     indices?: Uint32Array;
     /**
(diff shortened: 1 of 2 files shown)


src/XR/features/WebXRMeshDetector.ts
```diff
@@ -238,13 +231,23 @@ export class WebXRMeshDetector extends WebXRAbstractFeature {
...
-            }
-            else {
+            } else {
                 mesh.positions = xrMesh.positions;
-                mesh.indices = xrMesh.indices;
                 mesh.normals = xrMesh.normals;
             }
 
+            if (xrMesh.isClockwiseWindingOrder) {
+                // Babylon.js expects counter clockwise winding order for meshes
```
bghgary, line 240: Babylon.js winding order is configurable in the material as `sideOrientation` or can be overridden with the `overrideMaterialSideOrientation` property.

RaananW: Is that solving the issue here? Using the sideOrientation of the material / mesh.overrideMaterialSideOrientation ?

chrisfromwork: that could solve the problem. If setting a material property is a discoverable/known way to fix winding orders we could do that instead. I think we will still need to expose this value in the WebXRVertexData type so that folks know to use/not use this flag.

chrisfromwork: granted, after talking with OpenXR owners they also felt we could standardize on counter clockwise winding order for indices provided from webxr. @bghgary  and @RaananW  what would you two prefer?

bghgary: I'm not sure I'm the right person for an opinion for this. I don't have all the context. Is this part of the proposed contract for WebXR or Babylon.js? Maybe we can chat offline a bit.

chrisfromwork: I think we should do the following:

1. Assume that WebXR provides mesh coordinates in a right handed coordinate system with counterclockwise winding order.
2. Update the WebXR mesh detector to not assume a clockwise winding order for right handed coordinate systems like it does currently.

The pr is now up to date with changes based on these two principles

RaananW (approved):

PR #9589, opened 2020-12-13
Playing with ideas for attaching arbitrary camera properties to arbitrary UI

mrdunk:
Please note, this PR is not intended to be merged.
It's purpose is to start conversation.

It does compile and run though. `Translations/customUserDefinedTranslation.ts` has a sample user defined plugin that can be attached to an arbitrary MouseWheel axis.

My opinions:

1. I believe the code to affect the various Cameras properties should be independent of the code monitoring User Input.

2. Cameras should have handler methods to do common tasks. (In the attached code i've called these `Translations`.)

3. It should be possible to connect UI to any of the Camera Translation methods.

4. It should also be possible for a user to easily write their own Translation method plug-in.

5. The API for the various Cameras, Camera Translations, Input methods, etc should be a similar as possible.
We'll need to provide extra methods to handle the existing API but going forward it would be good to have a common way of linking UI to Camera property.
The legacy API methods can simple wrap the new API.

Issues with this PR:

1. Translations are keyed by `string`. Ideally there would be an Enum to index these but we also want the user to be able to create their own Translations... and i can't think of a way for users to extend an Enum... Not sure what the solution is here?

2. Babylon already has `Behaviors`. Can we use these for the Translations?

3. The demo code uses MouseWheel as input. This is a trivially simple input. How will we handle inputs with many sub-components. eg: MouseMove vs MouseDrag-button1 vs MouseDrag-button2 vs Mouse-Drag_ShiftKey, etc.

4. We'd need to add more code to `FreeCameraMouseWheelInputDemo` if we want it to replace `FreeCameraMouseWheelInput` to handle the existing API.

Stretch goals:

1. There is no reason why a Translation has to affect just a Camera. eg: Imagine if the MouseWheel could change light intensity.

2. There is no reason why a Translation only needs to be called in response to UI. Handy methods to translate a camera would be handy for other events to trigger too.

I my opinion, getting the API design right for this is the hardest single task.
I'm not happy with my previous submission for the `FreeCameraMouseWheelInput` but you guys were busy with a release at the time and didn't have time to listen to my incoherent ramblings on the subject.


mrdunk:
@PolygonalSun
https://forum.babylonjs.com/t/improve-experience-of-camera-control/15951/13?u=dunk

src/Cameras/index.ts
```diff
@@ -17,4 +18,4 @@ export * from "./Stereoscopic/index";
 export * from "./universalCamera";
 export * from "./virtualJoysticksCamera";
 export * from "./VR/index";
-export * from "./RigModes/index";
\ No newline at end of file
+export * from "./RigModes/index";
```
mrdunk, line 21: huh.
unrelated to this PR.
Must be a merge collision.

deltakosh:
Pinging @PolygonalSun

PolygonalSun:
> Please note, this PR is not intended to be merged.
> It's purpose is to start conversation.
> 
> It does compile and run though. `Translations/customUserDefinedTranslation.ts` has a sample user defined plugin that can be attached to an arbitrary MouseWheel axis.
> 
> My opinions:
> 
> 1. I believe the code to affect the various Cameras properties should be independent of the code monitoring User Input.

I agree with this as having them more decoupled would make updating and modifying either of them easier.

> 2. Cameras should have handler methods to do common tasks. (In the attached code i've called these `Translations`.)

I just want to make sure that I understand your Translations.  Is a Translation essentially a defined behavior that can be triggered for a compatible camera or would they just be tied to a specific type of camera (Could we have a mix-and-match setup or something to reduce repeated code)?

> 3. It should be possible to connect UI to any of the Camera Translation methods.

I'm having trouble seeing this one.  Would you be able to provide some example or use case for this?

> 4. It should also be possible for a user to easily write their own Translation method plug-in.

100% agree.

> 5. The API for the various Cameras, Camera Translations, Input methods, etc should be a similar as possible.
>    We'll need to provide extra methods to handle the existing API but going forward it would be good to have a common way of linking UI to Camera property.
>    The legacy API methods can simple wrap the new API.

An API for the camera systems would be awesome but this would require discussing what kind of use cases and scenarios this API would cover.  I'd definitely like to talk more on this one.

> 
> Issues with this PR:
> 
> 1. Translations are keyed by `string`. Ideally there would be an Enum to index these but we also want the user to be able to create their own Translations... and i can't think of a way for users to extend an Enum... Not sure what the solution is here?

I've been thinking on this one and I feel like there's a better option than strings but I can't think of better way yet.  I'll reply to this if something else comes to mind.

> 2. Babylon already has `Behaviors`. Can we use these for the Translations?

I think that Translations could be built off of Behaviors but then all Translation compatible cameras would need to also implement IBehaviorAware<T>.  This isn't a big deal but just something to keep in mind.

> 3. The demo code uses MouseWheel as input. This is a trivially simple input. How will we handle inputs with many sub-components. eg: MouseMove vs MouseDrag-button1 vs MouseDrag-button2 vs Mouse-Drag_ShiftKey, etc.

Additional checks or params would be required to hold the status of what buttons/key are active during the time of the press.  As of right now, we work with events, which contain that information.  Plus, we pass an enum with the type of input being provided.  There may be a more efficient way to handle this though.

> 4. We'd need to add more code to `FreeCameraMouseWheelInputDemo` if we want it to replace `FreeCameraMouseWheelInput` to handle the existing API.
> 
> Stretch goals:
> 
> 1. There is no reason why a Translation has to affect just a Camera. eg: Imagine if the MouseWheel could change light intensity.

I think that this could probably be handled by Behaviors.

> 2. There is no reason why a Translation only needs to be called in response to UI. Handy methods to translate a camera would be handy for other events to trigger too.

As long as there's a way for some object or class to access the active Translations for a given object, I don't see any issue with this.

> 
> I my opinion, getting the API design right for this is the hardest single task.
> I'm not happy with my previous submission for the `FreeCameraMouseWheelInput` but you guys were busy with a release at the time and didn't have time to listen to my incoherent ramblings on the subject.

Overall, I think that this is a good start and could result in a much cleaner design for the camera but I'd like to talk more on the goals and use cases for the camera API.

deltakosh:
> Babylon already has Behaviors. Can we use these for the Translations?

For me, all of that should be more behaviors

mrdunk:
> > 1. Cameras should have handler methods to do common tasks. (In the attached code i've called these `Translations`.)
> 
> I just want to make sure that I understand your Translations. Is a Translation essentially a defined behavior that can be triggered for a compatible camera or would they just be tied to a specific type of camera (Could we have a mix-and-match setup or something to reduce repeated code)?

I envisage them as working on a compatible camera.
They are methods that when called change the configuration of the referenced camera.
I think the logical conclusion of this is that many of these Translations should affect the base Camera class so be applicable to all camera types.
Imagine a future where all that differs between ArcRotateCamera and FollowCamera is the Translation called by the UI (and a few helper functions for configuration by legacy code).

> 
> > 1. It should be possible to connect UI to any of the Camera Translation methods.
> 
> I'm having trouble seeing this one. Would you be able to provide some example or use case for this?

Imagine mouse input as having 2 axis: X and Y.
When X changes, one callback function (Translation) gets called.
When Y changes, a different callback function (Translation) gets called.
Exactly which Translation gets called should be user definable (with sensible defaults out of the box).
 
> > 1. The API for the various Cameras, Camera Translations, Input methods, etc should be a similar as possible.
> >    We'll need to provide extra methods to handle the existing API but going forward it would be good to have a common way of linking UI to Camera property.
> >    The legacy API methods can simple wrap the new API.
> 
> An API for the camera systems would be awesome but this would require discussing what kind of use cases and scenarios this API would cover. I'd definitely like to talk more on this one.

Sounds like we are on the same page.
A whiteboard would be the easiest medium to discuss this but as Covid and geography rule that out, 
we'll need to make do with code samples.
I'm open to voice/video calls to discuss further if that is easiest.

> 
> > Issues with this PR:
> > 
> > 1. Translations are keyed by `string`. Ideally there would be an Enum to index these but we also want the user to be able to create their own Translations... and i can't think of a way for users to extend an Enum... Not sure what the solution is here?
> 
> I've been thinking on this one and I feel like there's a better option than strings but I can't think of better way yet. I'll reply to this if something else comes to mind.

I've been thinking about this too.
The way camera inputs are currently added would work here too: `camera.inputs.add(new SomeCameraInput());`.
I presume that's why camera inputs need the `getClassName()` method....
It's still strings but less nasty than this CR.

> 
> > 1. Babylon already has `Behaviors`. Can we use these for the Translations?
> 
> I think that Translations could be built off of Behaviors but then all Translation compatible cameras would need to also implement IBehaviorAware. This isn't a big deal but just something to keep in mind.

Sounds like a plan.
i'll look into Behaviors.
Let's start calling "Translations" "Behaviors" instead.

> 
> > 1. The demo code uses MouseWheel as input. This is a trivially simple input. How will we handle inputs with many sub-components. eg: MouseMove vs MouseDrag-button1 vs MouseDrag-button2 vs Mouse-Drag_ShiftKey, etc.
> 
> Additional checks or params would be required to hold the status of what buttons/key are active during the time of the press. As of right now, we work with events, which contain that information. Plus, we pass an enum with the type of input being provided. There may be a more efficient way to handle this though.

I actually have a further branch of this CR that does this quite nicely.
I was waiting for you to have a look at the current submission before i further complicate things...
I'll give it a last look and merge later tonight.

> > I my opinion, getting the API design right for this is the hardest single task.
> > I'm not happy with my previous submission for the `FreeCameraMouseWheelInput` but you guys were busy with a release at the time and didn't have time to listen to my incoherent ramblings on the subject.
> 
> Overall, I think that this is a good start and could result in a much cleaner design for the camera but I'd like to talk more on the goals and use cases for the camera API.

Just for the sake of discussion,
Do we need to subtly change our focus and think of this as a User Input API who's primary customer is an instance of a Camera?
Imagine if the event handler listening for Mouse/Keyboard/Whatever events didn't have to be attached to a camera, but instead are attached to the Scene.
The event listener would be configured to call Behaviors (Translations).
The Behaviors would modify Camera (or other Node) properties.

Am i going too far here?
It's likely outside the scope of what i have time to contribute
but i feel it would be a more logical approach.

sebavan:
@PolygonalSun I let you handle the review for this.

@mrdunk Can I move it to draft state until we would be closer to merge it ? it helps the community ppl who would look into current PR/Issues ?

mrdunk:
> @mrdunk Can I move it to draft state until we would be closer to merge it ? it helps the community ppl who would look into current PR/Issues ?

Seems like i can do that myself.
I didn't know about github Drafts. That's useful.

deltakosh:
I wonder if you can instead create the code on your repo and use it for discussions. PR should not stay open for that long

I'm closing for now if you don't mind :) and I let you keep discussing on @mrdunk repo :)

PR #9591, opened 2020-12-14
Prepass enhancements

CraigFeldspar:
+ prepass can be used in render targets
+ fixing prepass bugs with transparency
+ fixing prepass bugs with layers
+ composition is only used in prepass when necessary for subsurface scattering
+ post processes can now be shared between cameras/render targets without recreating textures on each frame
+ fixed a video memory leak with multirendertarget causing an unnecessary texture to be instanciated
+ post processes from base class `PostProcess` can now use `.clone()` and are properly serialized
+ adding lots of visual tests for prepass


CraigFeldspar:
marking as draft to prevent wild merge

src/Engines/Extensions/engine.multiRender.ts
```diff
@@ -22,18 +22,20 @@ declare module "../../Engines/thinEngine" {
          * @see https://doc.babylonjs.com/features/webgl2#multiple-render-target
          * @param size defines the size of the texture
          * @param options defines the creation options
+         * @param noDrawBuffers if set to true, the engine won't make an initializing call of drawBuffers
          * @returns the cube texture as an InternalTexture
          */
-        createMultipleRenderTarget(size: any, options: IMultiRenderTargetOptions): InternalTexture[];
+        createMultipleRenderTarget(size: any, options: IMultiRenderTargetOptions, noDrawBuffers?: boolean): InternalTexture[];
```
Popov72, line 28: I think we decided that `noDrawBuffers` should be renamed something like `initializeBuffers` instead?

src/Materials/Textures/mirrorTexture.ts
```diff
@@ -174,7 +174,8 @@ export class MirrorTexture extends RenderTargetTexture {
 
             this._mirrorMatrix.multiplyToRef(this._savedViewMatrix, this._transformMatrix);
 
-            scene.setTransformMatrix(this._transformMatrix, scene.getProjectionMatrix());
+            // Clone to not mark matrices as updated
+            scene.setTransformMatrix(this._transformMatrix.clone(), scene.getProjectionMatrix().clone());
```
Popov72, line 178: I don't understand why we need to clone the matrices before passing them?

src/Materials/Textures/prePassRenderTarget.ts
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * @hidden
+     */
+    public _beforeCompositionPostProcesses: PostProcess[] = [];
+    /**
+     * Image processing post process for composition
+     */
+    public imageProcessingPostProcess: ImageProcessingPostProcess;
+
+    public _engine: Engine;
```
Popov72, line 27: I think(?) you need to add /** @hidden */ before the public variables declared with a leading underscore, else they will be visible in the doc.

src/Materials/Textures/prePassRenderTarget.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * A multi render target designed to render the prepass.
+ * Prepass is a scene component used to render information in multiple textures
+ * alongside with the scene materials rendering.
+ * Note : This is an internal class, and you should NOT need to instanciate this.
+ * Only the `PrePassRenderer` should instanciate this class.
+ * It is more likely that you need a regular `MultiRenderTarget`
+ */
+export class PrePassRenderTarget extends MultiRenderTarget {
```
Popov72, line 17: Shouldn't we add a `@hidden` as this class is not meant to be used by the user?

src/Materials/Textures/renderTargetTexture.ts
```diff
@@ -133,9 +134,20 @@ export class RenderTargetTexture extends Texture {
...
+    /**
+     * Post-processes for this render target
+     */
+    public get postProcesses() {
+        return this._postProcesses;
+    }
     private _postProcesses: PostProcess[];
     private _resizeObserver: Nullable<Observer<Engine>>;
 
+    private get _prePass() {
```
Popov72, line 147: Maybe add `: boolean` after `private get _prePass()` to make the return type clearer (or rename it to `_prePassEnabled()`?).

src/PostProcesses/postProcess.ts
```diff
@@ -477,6 +490,71 @@ export class PostProcess {
         this.width = -1;
     }
 
+    private _createRenderTargetTexture(textureSize: { width: number, height: number }, textureOptions: RenderTargetCreationOptions, channel = 0) {
+        for (let i = 0; i < this._textureCache.length; i++) {
+            if (this._textureCache[i].width === textureSize.width &&
+                this._textureCache[i].height === textureSize.height &&
+                this._textureCache[i]._postProcessChannel === channel) {
```
Popov72, line 497: Should `_postProcessChannel` really be a member of `InternalTexture`? It seems it's only of use in the `PostProcess` class (and only in this method), so I would see more something like:
```javascript
private _textureCache: Array<{texture: InternalTexture, postProcessChannel: number}> = [];
```

src/PostProcesses/postProcess.ts
```diff
@@ -477,6 +490,71 @@ export class PostProcess {
...
+
+    private _flushTextureCache() {
+        const currentRenderId = this._renderId;
+
+        for (let i = this._textureCache.length - 1; i >= 0; i--) {
+            if (currentRenderId - this._textureCache[i]._lastUsedRenderId > 100) {
+                let currentlyUsed = false;
+                for (var j = 0; j < this._textures.length; j++) {
+                    if (this._textures.data[j] === this._textureCache[i]) {
+                        currentlyUsed = true;
```
Popov72, line 517: nit: You could `break` here

src/PostProcesses/postProcess.ts
```diff
@@ -477,6 +490,71 @@ export class PostProcess {
...
+        this._textureCache.push(tex);
+
+        return tex;
+    }
+
+    private _flushTextureCache() {
+        const currentRenderId = this._renderId;
+
+        for (let i = this._textureCache.length - 1; i >= 0; i--) {
+            if (currentRenderId - this._textureCache[i]._lastUsedRenderId > 100) {
```
Popov72, line 513: Same thing for `_lastUsedRenderId` than for `_postProcessChannel`, I think it's better to have:
```typescript
private _textureCache: Array<{texture: InternalTexture, postProcessChannel: number, _lastUsedRenderId : number}> = [];
```
and not make `InternalTexture` grow only for this post process usage.

src/PostProcesses/postProcess.ts
```diff
@@ -677,17 +732,32 @@ export class PostProcess {
 
     private _disposeTextures() {
         if (this._shareOutputWithPostProcess || this._forcedOutputTexture) {
+            this._disposeTextureCache();
```
Popov72, line 735: Was it a bug of the previous code to not release the textures in this case?

src/PostProcesses/postProcess.ts
```diff
@@ -677,17 +732,32 @@ export class PostProcess {
...
                 this._engine._releaseTexture(this._textures.data[i]);
+
+                const cacheIndex = this._textureCache.indexOf(this._textures.data[i]);
+                if (cacheIndex !== -1) {
+                    this._textureCache.splice(cacheIndex, 1);
+                }
             }
         }
+
+        this._disposeTextureCache();
```
Popov72, line 750: Is the code inside the loop above necessary given that `this._disposeTextureCache();` is called unconditionally here?

src/PostProcesses/postProcess.ts
```diff
@@ -677,17 +732,32 @@ export class PostProcess {
...
         this._textures.dispose();
     }
 
+    private _disposeTextureCache() {
+        for (let i = this._textureCache.length - 1; i >= 0; i--) {
+            this._engine._releaseTexture(this._textureCache[i]);
+            this._textureCache.splice(i, 1);
```
Popov72, line 757: nit: you could remove this line and add `this._textureCache.length = 0;` after the loop

src/Rendering/prePassRenderer.ts
```diff
@@ -178,47 +216,32 @@ export class PrePassRenderer {
...
+     * Creates a new PrePassRenderTarget
+     * This should be the only way to instanciate a `PrePassRenderTarget`
+     * @param name Name of the `PrePassRenderTarget`
+     * @param renderTargetTexture RenderTarget the `PrePassRenderTarget` will be attached to.
+     * Can be `null` if the created `PrePassRenderTarget` is attached to the scene (default framebuffer).
+     * @hidden
+     */
+    public _createRenderTarget(name: string, renderTargetTexture: Nullable<RenderTargetTexture>) : PrePassRenderTarget {
+        const rt = new PrePassRenderTarget(name, renderTargetTexture, { width: this._engine.getRenderWidth(), height: this._engine.getRenderHeight() }, 0, this._scene,
+            { generateMipMaps: false, generateStencilBuffer: true, defaultType: Constants.TEXTURETYPE_UNSIGNED_INT, types: [], drawOnlyOnFirstAttachmentByDefault: true });
```
Popov72, line 233: To save some resources, you can do: `generateStencilBuffer: true` => `generateStencilBuffer: this._engine.isStencilEnable` 

But I wonder if we really need to enable stencil at all on the `PrePassRenderTarget`?

src/Rendering/prePassRenderer.ts
```diff
@@ -178,47 +216,32 @@ export class PrePassRenderer {
...
-        this.imageProcessingPostProcess.autoClear = false;
+        return rt;
     }
 
     /**
      * Indicates if rendering a prepass is supported
      */
     public get isSupported() {
-        return this._scene.getEngine().getCaps().drawBuffersExtension;
+        return this._engine.webGLVersion > 1 || this._scene.getEngine().getCaps().drawBuffersExtension;
```
Popov72, line 244: Please remove `this._engine.webGLVersion > 1`, it won't work in WebGPU.

`this._scene.getEngine().getCaps().drawBuffersExtension` should be enough for the pre pass renderer to work? Is there another feature that would not make it work in webgl 1?

src/Rendering/prePassRenderer.ts
```diff
@@ -377,7 +347,109 @@ export class PrePassRenderer {
...
+        if (this._enabled && this._currentTarget.enabled) {
+            this._currentTarget._checkSize();
+            var internalTexture = this._currentTarget.getInternalTexture();
+            if (internalTexture) {
+                this._engine.bindFramebuffer(internalTexture);
+            }
+        }
+    }
+
+    private _setState(enabled: boolean) {
```
Popov72, line 443: Maybe `_setEnabled` would be clearer? Or do you plan to add other states than simply `enabled`?

src/Rendering/prePassRenderer.ts
```diff
@@ -377,7 +347,109 @@ export class PrePassRenderer {
...
+        this._enabled = enabled;
+    }
+
+    private _setRenderTargetState(prePassRenderTarget: PrePassRenderTarget, enabled: boolean) {
```
Popov72, line 447: Same comment here than just above on `_setState`.

src/Rendering/prePassRenderer.ts
```diff
@@ -413,90 +476,156 @@ export class PrePassRenderer {
...
+            return this._scene.activeCamera ? this._scene.activeCamera._postProcesses : [];
+        }
+    }
+
+    private _setupOutputForThisPass(prePassRenderTarget: PrePassRenderTarget, camera?: Camera) {
+        // Order is : draw ===> prePassRenderTarget._postProcesses ==> ipp ==> camera._postProcesses
+        const secondaryCamera = camera && this._scene.activeCameras && !!this._scene.activeCameras.length && this._scene.activeCameras.indexOf(camera) !== 0;
+        this._postProcessesSourceForThisPass = this._getPostProcessesSource(prePassRenderTarget, camera);
+        this._postProcessesSourceForThisPass = (this._postProcessesSourceForThisPass.filter((pp) => { return pp != null; }));
+        this._scene.autoClear = true;
```
Popov72, line 543: Shouldn't `scene.autoClear` be saved when enabling the pre pass renderer and restored when disabling it to make sure we don't interfere when the pre pass renderer is not used?

src/Rendering/prePassRenderer.ts
```diff
@@ -413,90 +476,156 @@ export class PrePassRenderer {
...
-        } else {
-            const pp = this._scene.activeCamera?._getFirstPostProcess();
-            if (pp) {
-                pp.inputTexture = this.prePassRT.getInternalTexture()!;
+    private _hasImageProcessing(postProcesses: Nullable<PostProcess>[]): boolean {
+        let isIPPAlreadyPresent = false;
+        if (postProcesses) {
+            for (let i = 0; i < postProcesses.length; i++) {
+                if (postProcesses[i]?.getClassName() === "ImageProcessingPostProcess") {
+                    isIPPAlreadyPresent = true;
```
Popov72, line 610: nit: you could `break` here

src/Rendering/prePassRenderer.ts
```diff
@@ -526,24 +655,46 @@ export class PrePassRenderer {
     private _update() {
         this._disable();
         let enablePrePass = false;
+        this._scene.imageProcessingConfiguration.applyByPostProcess = false;
 
         for (let i = 0; i < this._scene.materials.length; i++) {
             if (this._scene.materials[i].setPrePassRenderer(this)) {
                 enablePrePass = true;
+
+                this._setRenderTargetState(this.defaultRT, true);
```
Popov72, line 664: This line is in the loop but does not depend on it. Shouldn't the call be done after the loop if `enablePrePass = true`?

src/Rendering/prePassRendererSceneComponent.ts
```diff
@@ -147,14 +187,15 @@ export class PrePassRendererSceneComponent implements ISceneComponent {
      * context lost for instance.
      */
     public rebuild(): void {
-        // Nothing to do for this component
+        this.scene.disablePrePassRenderer();
+        this.scene.enablePrePassRenderer();
```
Popov72, line 191: Maybe that would deserve a comment as we may not understand why `disableXXX` has to be called before `enableXXX`?

src/Shaders/default.fragment.fx
```diff
@@ -476,10 +476,12 @@ color.rgb = max(color.rgb, 0.);
 
 #define CUSTOM_FRAGMENT_BEFORE_FRAGCOLOR
 #ifdef PREPASS
+	float writeGeometryInfo = color.a > 0.4 ? 1.0 : 0.0;
```
Popov72, line 479: Would it be advisable to have a #define constant instead of 0.4? I'm not sure, just asking.

src/Shaders/pbr.fragment.fx
```diff
@@ -536,8 +536,10 @@ void main(void) {
     #define CUSTOM_FRAGMENT_BEFORE_FRAGCOLOR
 
 #ifdef PREPASS
+    float writeGeometryInfo = finalColor.a > 0.4 ? 1.0 : 0.0;
```
Popov72, line 539: Same comment than above on the 0.4 value.

src/Materials/Textures/mirrorTexture.ts
CraigFeldspar: If we use the scene matrices, there is a cache in `setTransformMatrix` : 
```
        if (this._viewUpdateFlag === viewL.updateFlag && this._projectionUpdateFlag === projectionL.updateFlag) {
            return;
        }

        this._viewUpdateFlag = viewL.updateFlag;
        this._projectionUpdateFlag = projectionL.updateFlag;
```

Then in specific cases, it can mess up the regular update of the transform matrix outside of `MirrorTexture`

CraigFeldspar: We can also add a parameter `disableCache` in `setTransformMatrix`

src/Materials/Textures/prePassRenderTarget.ts
CraigFeldspar: you are probably right, let's add it just in case

src/Materials/Textures/prePassRenderTarget.ts
CraigFeldspar: yep

src/Materials/Textures/renderTargetTexture.ts
CraigFeldspar: I like the `_prePassEnabled`

src/PostProcesses/postProcess.ts
CraigFeldspar: Totally, it's definitely only used in `PostProcess`

src/PostProcesses/postProcess.ts
CraigFeldspar: I think in case `this._shareOutputWithPostProcess` or `this._forcedOutputTexture` was defined, textures in the texture array are released in the setter. Anyhow, defensively this makes sense.

src/PostProcesses/postProcess.ts
CraigFeldspar: yeah, `_textureCache` should always contain `_textures`, I will remove the loop

src/Rendering/prePassRenderer.ts
CraigFeldspar: It is necessary for effects like `HighlightLayer`, but your remark about the engine level check is definitely valid

src/Rendering/prePassRenderer.ts
CraigFeldspar: subsurfacescattering needs webgl2, but now since prepass is uncorrelated from it, we can remove the webGL2 check

src/Rendering/prePassRenderer.ts
CraigFeldspar: That's a fair question. In `DefaultRenderingPipeline.ts`, this flag is also hijacked. Maybe @sebavan has some insight ?

src/Rendering/prePassRenderer.ts
CraigFeldspar: It sure can be, it's just a boolean flag switch

src/Shaders/default.fragment.fx
CraigFeldspar: we could add a define `ALPHA_GEOMETRY_CUTOFF` in `Material.ts`

CraigFeldspar: it should also be accessible from `GeometryBufferRenderer`

src/Materials/Textures/mirrorTexture.ts
Popov72: I still don't get it... Do you have an example where not cloning would fail?

The `scene.setTransformMatrix(this._savedViewMatrix, scene.getProjectionMatrix());` call in `onAfterRenderObservable` should set the right matrix back when the mirror class has finished working.

This pattern is used in several classes (water material, shadow/cascaded shadow generator, reflection probe) and we didn't have problems with it until now. Is there something in this PR that makes this pattern fail?

CraigFeldspar: It is connected to prepass with mirror textures.

The call in `onAfterRenderObservable ` is not the problematic one. The problem is the first call to `setTransformMatrix` updates the `updateFlag` cache system on the projection matrix. Then the `PrePassRenderTarget` calls `setTransformMatrix` itself (in `RenderTargetTexture.render`) , which is ineffective because of cache, and causes the post-process to have the wrong projectionMatrix.

The repro is this playground : 
http://localhost:1338/Playground/index-local.html#PIZ1GK#206

Press '2' to get the SSR effect, and change your browser size to alter the projectionMatrix

Popov72: Thanks for the repro.

The problem comes from the fact that the SSAO postprocess is updating the projection matrix (by calling `camera.getProjectionMatrix(true)`) with values depending on the aspect ratio of the RTT, which may not be the same one than the aspect ratio of the output buffer. In the `onAfterRenderObservable` observer of `MirrorTexture`, when we do `scene.setTransformMatrix(this._savedViewMatrix, scene.getProjectionMatrix())`, the projection matrix we get is the last computed one, so it is the one computed by SSAO, which may not have been computed with the right aspect ratio for the output buffer.

Your workaround (cloning the matrices) works because it passes a new matrix which replaces the camera matrix (`scene._projectionMatrix` is normally the same thing than `camera.getProjectionMatrix()`). So, even if SSAO is (indirectly) modifying the camera projection matrix, `scene.getProjectionMatrix()` will return the initial cloned matrix which did not change and has the right values.

Instead of cloning the matrices, I think the right way to fix the problem is to restore the view/projection matrices of the active camera (it's what shadow/cascaded shadow generator and reflection probe are doing) by replacing:
```javascript
scene.setTransformMatrix(this._savedViewMatrix, scene.getProjectionMatrix());
```
by:
```javascript
scene.updateTransformMatrix(); // restore the view/projection matrices of the active camera
```

CraigFeldspar: works, and is way cleaner than cloning, thanks a ton !

sebavan (approved):

deltakosh (approved):

PR #9601, opened 2020-12-14
Feature/sky material up vector

Michalzr:
Feature requested on the forum:
https://forum.babylonjs.com/t/skymaterial-up-vector/16298/4

I also added `FromUnitVectors` method on the quaternion, because I believe it is useful not only for this scenario. I took the implementation from [three.js quaternion](https://github.com/mrdoob/three.js/blob/dev/src/math/Quaternion.js).

Looking forward to any feedback.


materialsLibrary/src/sky/skyMaterial.ts
```diff
@@ -298,6 +306,9 @@ export class SkyMaterial extends PushMaterial {
             this.sunPosition.x = this.distance * Math.cos(phi);
             this.sunPosition.y = this.distance * Math.sin(phi) * Math.sin(theta);
             this.sunPosition.z = this.distance * Math.sin(phi) * Math.cos(theta);
+
+            const rotation = Quaternion.FromUnitVectors(Vector3.Up(), this.up);
```
deltakosh, line 310: I would recommend using a ToRef function here as we do not want to create a quaternion per frame

src/Maths/math.vector.ts
```diff
@@ -3452,6 +3452,35 @@ export class Quaternion {
         return result;
     }
 
+    /**
+     * Creates a rotation quaternion that rotates vector vecFrom to vector vecTo
+     * @param vecFrom defines the direction vector from which to rotate
+     * @param vecTo defines the direction vector to which to rotate
+     * @returns the new quaternion
+     */
+    public static FromUnitVectors(vecFrom: DeepImmutable<Vector3>, vecTo: DeepImmutable<Vector3>): Quaternion {
```
deltakosh, line 3461: We need to make it "ToRef" to save GC

materialsLibrary/src/sky/skyMaterial.ts
```diff
@@ -113,6 +113,12 @@ export class SkyMaterial extends PushMaterial {
     @serialize()
     public cameraOffset: Vector3 = Vector3.Zero();
 
+    /**
+     * Defines the vector the skyMaterial should consider as up. (default is Vector3(0, 1, 0) as returned by Vector3.Up())
+     */
+    @serialize()
```
julien-moreau, line 119: I think @seriazeAsVector3 is needed here instead of @serialize

materialsLibrary/src/sky/skyMaterial.ts
Michalzr: Changed it to `FromUnitVectorsToRef` and made the quaternion a private property so it doesn't get created every frame.

But I have a question:
the way I understand it, the `Vector3.Up()` will create a new vector every frame as well. Same goes for the cross product in the `FromUnitVectorsToRef` method. Shouldn't we do something with them as well? If not, isn't then making the quaternion a private property a microoptimization?

deltakosh: yes totally. I think (I'm on my phone now) that we should have a UpReadOnly static property

and we should have a CrossToRef

Michalzr: Nice, I used the UpReadOnly (kind of feel stupid for not finding it on my own), but I'm not sure how to use the cross product to ref, since I need to set the result to Quaternion.
I guess I'll just wait till you have a proper look.

src/Maths/math.vector.ts
```diff
@@ -3452,6 +3452,34 @@ export class Quaternion {
...
+        const r = Vector3.Dot(vecFrom, vecTo) + 1;
+
+        if (r < Epsilon) {
+            if (Math.abs(vecFrom.x) > Math.abs(vecFrom.z)) {
+                result.set(-vecFrom.y, vecFrom.x, 0, 0);
+            } else {
+                result.set(0, - vecFrom.z, vecFrom.y, 0);
+            }
+        } else {
+            const crossProduct = Vector3.Cross(vecFrom, vecTo);
```
deltakosh, line 3471: Something like Vector3.CrossToRef(vecFrom, vecTo, result)

?

MichalZru: I must be missing something.
This would (probably) work in js, but typescript complains when passing quaternion as vector.

Michalzr: I must be missing something.
This would (probably) work in js, but typescript complains about passing a quaternion as vector3

Michalzr: I can maybe inline the cross product implementation? They do that in the three js as well.
Or I can cast the quaternion to any and pass it like that, but that doesn't feel right.

deltakosh: this is what I meant:
```
 Vector3.CrossToRef(vecFrom, vecTo, result);
```

deltakosh: oh wait.. I misread
so something like:
```
 Vector3.CrossToRef(vecFrom, vecTo, MathTmp.Vector3[0]);
```

Michalzr: Nice, did not notice `MathTmp`.
I'll change it in a few hours.

sebavan:
Looks all good aside of one error in the build for the function documentations....

image

PR #9606, opened 2020-12-15
Make Geometry shareable by sourceMeshes

breakin:
This PR hope to be a better alternative than the approach tried in #9214 .
It can use the same Geometry object with many meshes that has instances (or not).
The VertexArrayObject and per-instance VertexBuffers are handled by the Mesh itself.

The goal is the not have to copy the entire geometry (increases CPU/GPU storage) whenever we want to create instances from it (as otherwise it will destroy the other users).

```
    camera.minZ = 0.1;
    camera.maxZ = 100.0;
    light.intensity = 0.7;

    BABYLON.Effect.ShadersStore["myVertexShader"] = `
		#ifdef GL_ES
		precision highp float;
		#endif
		// Attributes
		attribute vec3 position;
		attribute vec2 uv;
		#ifdef INSTANCES
			attribute vec4 world0;
			attribute vec4 world1;
			attribute vec4 world2;
			attribute vec4 world3;
			attribute vec3 per_instance_data;
		#endif
		uniform mat4 world;
        uniform vec3 albedoColor;
		uniform mat4 viewProjection;
		varying vec3 color;
		varying vec2 vUV;
		void main(void) {
			vec3 positionUpdated = position;

            #ifdef INSTANCES
            	positionUpdated.xz *= per_instance_data.x;
            #endif

            #include<instancesVertex>

            vec4 worldPos = finalWorld * vec4(positionUpdated, 1.0);
            gl_Position = viewProjection * worldPos;
			color = albedoColor;
			vUV = uv;
		}	        
    `;

    BABYLON.Effect.ShadersStore["myFragmentShader"] = `
		#ifdef GL_ES
		precision highp float;
		#endif
		varying vec2 vUV;
		varying vec3 color;
		void main(void) {
            vec3 c = color * vUV.x;
			gl_FragColor = vec4(c, 1.0);
		}    
    `;

    var sett = {needAlphaBlending: false, attributes: ["position", "uv", "per_instance_data"], uniforms: ["world", "viewProjection", "albedoColor"] };
    var m1 = new BABYLON.ShaderMaterial("my1", scene, {vertexElement: "my", fragmentElement: "my"}, sett );
    var m2 = new BABYLON.ShaderMaterial("my2", scene, {vertexElement: "my", fragmentElement: "my"}, sett );
    var m3 = new BABYLON.ShaderMaterial("my3", scene, {vertexElement: "my", fragmentElement: "my"}, sett );

    m1.metallic = 0.0
    m2.metallic = 0.0
    m3.metallic = 0.0
    m1.setVector3("albedoColor", new BABYLON.Vector3(1.0, 0.0, 0.0));
    m2.setVector3("albedoColor", new BABYLON.Vector3(0.0, 1.0, 0.0));
    m3.setVector3("albedoColor", new BABYLON.Vector3(0.0, 0.0, 1.0));

    var sphere = BABYLON.MeshBuilder.CreateSphere("sphere", {diameter: 1, segments: 32}, scene);
    sphere.setEnabled(false)

    var s1 = sphere.clone('s1')
    var s2 = sphere.clone('s2')
    var s3 = sphere.clone('s3')
    s1.material = m1
    s2.material = m2
    s3.material = m3

    s1.registerInstancedBuffer('per_instance_data', 3);
    s2.registerInstancedBuffer('per_instance_data', 3);
    s3.registerInstancedBuffer('per_instance_data', 3); // This one will not be used for instances, lets see how it goes

    const instances = []

    for (let i=0; i<20; i++) {
        // Instances using material 1
        instances.push(s1.createInstance())
    }
    for (let i=0; i<20; i++) {
        // Instances using material 2
        instances.push(s2.createInstance())
    }
    for (let i=0; i<20; i++) {
        // Not instances at all
        instances.push(s3.clone())
    }


    var xp = -3.0, zp = 0.0;
    for (let idx=0; idx<instances.length; idx++) {
        const i = instances[idx]
        var s = 1.0
        if (i instanceof BABYLON.InstancedMesh) {
        	s = Math.random()*0.3+0.7;
        	i.instancedBuffers.per_instance_data = new BABYLON.Vector3(s, 0, 0);
        }

        i.position.x = xp;
        xp += s*1.1;

        
        i.position.z = zp
        i.position.y = 0.5;

        if (xp > 8.0) {
        	xp = -3.0;
        	zp += 1.1;
        }
        i.setEnabled(true);
    }

    var test = -1;

    // Here are some tests to see if things propagate
    setTimeout(() => {
        if (test === 0) { // Dispose
            for (let idx=0; idx<instances.length; idx++) {
                instances[idx].dispose();
            }
            s1.dispose();
            s2.dispose();
            s3.dispose();
        } else if (test === 1) { // Dispose
            s1.dispose();
            s2.dispose();
            s3.dispose();
        } else if (test === 2) { // Turn all spheres into a box by modifying geometry of one of them
	        var vertexData = BABYLON.VertexData.CreateBox({});
    	    vertexData.applyToMesh(sphere, false);
        } else if (test === 3) { // Turn all spheres into a box by modifying geometry of one of them
	        var vertexData = BABYLON.VertexData.CreateBox({});
    	    vertexData.applyToMesh(s1, false);
            // NOTE: Clicking on objects give an error right now, see https://forum.babylonjs.com/t/applying-vertexdata-to-mesh-used-as-instancedmesh-source/16670
        } else if (test === 4) { // Make sure the VAO of the instances are notified when geometry buffers are modified
            // Add a new Vertex channel, see if all objects get it
            var num = sphere.geometry.getTotalVertices() * 2;
            var uv = new Float32Array(num);
            for (let i = 0; i<num; i++) {
                uv[i] = Math.random();
            }
            // Make sure the VAO of the instances are notified that they need to be rebuilt
            sphere.setVerticesData("uv", uv, false)
        } else if (test === 5) { // Make sure the VAO of the instances are notified when geometry buffers are modified
            // Add a new Vertex channel, see if all objects get it
            var num = sphere.geometry.getTotalVertices() * 2;
            var uv = new Float32Array(num);

            setInterval(() => {
                for (let i = 0; i<num; i++) {
                    uv[i] = Math.random();
                }
                // Make sure the VAO of the instances are notified that they need to be rebuilt
                sphere.setVerticesData("uv", uv, true)

            }, 500)
        }
    }, 500);

    var ground = BABYLON.MeshBuilder.CreateGround("ground", {width: 12, height: 12}, scene);
    ground.receiveShadows = true

    return scene;
};
```

### DIFF
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -1766,7 +1766,7 @@ export class ThinEngine {
         }
     }
 
-    private _bindVertexBuffersAttributes(vertexBuffers: { [key: string]: Nullable<VertexBuffer> }, effect: Effect, extraVertexBuffers?: { [kind: string]: Nullable<VertexBuffer>}): void {
+    private _bindVertexBuffersAttributes(vertexBuffers: { [key: string]: Nullable<VertexBuffer> }, effect: Effect, overrideVertexBuffers?: { [kind: string]: Nullable<VertexBuffer>}): void {
         var attributes = effect.getAttributesNames();
 
         if (!this._vaoRecordInProgress) {
@@ -1782,8 +1782,8 @@ export class ThinEngine {
                 var ai = attributes[index];
                 var vertexBuffer: Nullable<VertexBuffer> = null;
 
-                if (extraVertexBuffers) {
-                    vertexBuffer = extraVertexBuffers[ai];
+                if (overrideVertexBuffers) {
+                    vertexBuffer = overrideVertexBuffers[ai];
                 }
 
                 if (!vertexBuffer) {
@@ -1821,9 +1821,10 @@ export class ThinEngine {
      * @param vertexBuffers defines the list of vertex buffers to store
      * @param indexBuffer defines the index buffer to store
      * @param effect defines the effect to store
+     * @param overrideVertexBuffers defines optional list of avertex buffers that overrides the entries in vertexBuffers
      * @returns the new vertex array object
      */
-    public recordVertexArrayObject(vertexBuffers: { [key: string]: VertexBuffer; }, indexBuffer: Nullable<DataBuffer>, effect: Effect, extraVertexBuffers?: { [kind: string]: Nullable<VertexBuffer>}): WebGLVertexArrayObject {
+    public recordVertexArrayObject(vertexBuffers: { [key: string]: VertexBuffer; }, indexBuffer: Nullable<DataBuffer>, effect: Effect, overrideVertexBuffers?: { [kind: string]: Nullable<VertexBuffer>}): WebGLVertexArrayObject {
         var vao = this._gl.createVertexArray();
 
         this._vaoRecordInProgress = true;
@@ -1831,7 +1832,7 @@ export class ThinEngine {
         this._gl.bindVertexArray(vao);
 
         this._mustWipeVertexAttributes = true;
-        this._bindVertexBuffersAttributes(vertexBuffers, effect, extraVertexBuffers);
+        this._bindVertexBuffersAttributes(vertexBuffers, effect, overrideVertexBuffers);
 
         this.bindIndexBuffer(indexBuffer);
 
@@ -1913,13 +1914,14 @@ export class ThinEngine {
      * @param vertexBuffers defines the list of vertex buffers to bind
      * @param indexBuffer defines the index buffer to bind
      * @param effect defines the effect associated with the vertex buffers
+     * @param overrideVertexBuffers defines optional list of avertex buffers that overrides the entries in vertexBuffers
      */
-    public bindBuffers(vertexBuffers: { [key: string]: Nullable<VertexBuffer> }, indexBuffer: Nullable<DataBuffer>, effect: Effect, extraVertexBuffers?: {[kind: string]: Nullable<VertexBuffer>}): void {
+    public bindBuffers(vertexBuffers: { [key: string]: Nullable<VertexBuffer> }, indexBuffer: Nullable<DataBuffer>, effect: Effect, overrideVertexBuffers?: {[kind: string]: Nullable<VertexBuffer>}): void {
         if (this._cachedVertexBuffers !== vertexBuffers || this._cachedEffectForVertexBuffers !== effect) {
             this._cachedVertexBuffers = vertexBuffers;
             this._cachedEffectForVertexBuffers = effect;
 
-            this._bindVertexBuffersAttributes(vertexBuffers, effect, extraVertexBuffers);
+            this._bindVertexBuffersAttributes(vertexBuffers, effect, overrideVertexBuffers);
         }
 
         this._bindIndexBufferWithCache(indexBuffer);
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -1780,10 +1780,14 @@ export class ThinEngine {
 
             if (order >= 0) {
                 var ai = attributes[index];
-                var vertexBuffer = vertexBuffers[ai];
+                var vertexBuffer: Nullable<VertexBuffer> = null;
 
-                if (!vertexBuffer && extraVertexBuffers) {
+                if (extraVertexBuffers) {
                     vertexBuffer = extraVertexBuffers[ai];
+                }
+
+                if (!vertexBuffer) {
+                    vertexBuffer = vertexBuffers[ai];
                 }
 
                 if (!vertexBuffer) {
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -1766,7 +1766,7 @@ export class ThinEngine {
         }
     }
 
-    private _bindVertexBuffersAttributes(vertexBuffers: { [key: string]: Nullable<VertexBuffer> }, effect: Effect): void {
+    private _bindVertexBuffersAttributes(vertexBuffers: { [key: string]: Nullable<VertexBuffer> }, effect: Effect, extraVertexBuffers?: { [kind: string]: Nullable<VertexBuffer>}): void {
         var attributes = effect.getAttributesNames();
 
         if (!this._vaoRecordInProgress) {
@@ -1779,7 +1779,12 @@ export class ThinEngine {
             var order = effect.getAttributeLocation(index);
 
             if (order >= 0) {
-                var vertexBuffer = vertexBuffers[attributes[index]];
+                var ai = attributes[index];
+                var vertexBuffer = vertexBuffers[ai];
+
+                if (!vertexBuffer && extraVertexBuffers) {
+                    vertexBuffer = extraVertexBuffers[ai];
+                }
 
                 if (!vertexBuffer) {
                     continue;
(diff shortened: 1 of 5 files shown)


breakin:
@deltakosh: Ok puh. I think this is the embryo to a much better approach than the one I tried before. It is not done at all but it feels like less overhead and just as powerful. Let me know what you think of the approach.

I am very happy with this development. Gotta polish it like the old one but I have a feeling that it might be less painful. Hopefully!

src/Meshes/mesh.ts
```diff
@@ -2949,18 +2962,6 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
      * @returns a new InstancedMesh
      */
     public createInstance(name: string): InstancedMesh {
```
breakin, line 2964: This is the important goal we are after. No per-source mesh operation mutates the Geometry so we don't need to make it unique in order to create instances. This lets us clone a template-mesh many times. From each clone we can create instances. They all share the same heavy Geometry obejct.

The reason we need a template is that cloning doesn't work well once we start having instances.

src/Meshes/instancedMesh.ts
```diff
@@ -607,13 +614,28 @@ Mesh.prototype._processInstancedBuffers = function(visibleInstances: InstancedMe
...
+    }
+
+    if (!lostContext) {
+        for (var kind in this._userInstancedBuffersStorage.vertexArrayObjects) {
+            this.getEngine().releaseVertexArrayObject(this._userInstancedBuffersStorage.vertexArrayObjects[kind]);
+        }
+    }
+
+    this._userInstancedBuffersStorage.vertexArrayObjects = {}
+    console.log('Invalidate mesh VAO')
```
deltakosh, line 636: make sure to remove all console.log

src/Meshes/mesh.ts
deltakosh: I genuinely love it!!!!!

src/Meshes/geometry.ts
```diff
@@ -915,6 +931,8 @@ export class Geometry implements IGetSetVerticesData {
             }
             this._vertexArrayObjects = {};
         }
+
+        // TODO: Do the meshes too?
```
deltakosh, line 935: what do you mean?

deltakosh:
I honestly prefer this one BY FAR!

My only question is: we do not tackle the problem when multiple cloned geometries try to restore the same buffer right? (after a context lost). This is ok as I can tackle it in a different PR

breakin:
@deltakosh: Ah yes I am also very happy with this approach. This is my third attempt and this one is a simpler variant of my first approach. The second approach got really messy. After a while it just felt wrong; why clone the geometry when everything shared really is fully shared? Now everything should work exactly like it does with cloned meshes without instances in terms of geometries/buffers being synched. Just gotta polish and try some more.

On a sidenote; I think a button in the playground that simulates a context-lost might be really good for people to try if their changes can work over a lost context. The code is really for it; there is an extension that simulates it happening for real.

```
    var looser
    setTimeout(() => {
        if (looser === null) {
            looser = engine._gl.getExtension('WEBGL_lose_context');
        }
        looser.loseContext();
    }, 1000)

    // Restore context after 2 seconds
    setTimeout(() => {
        if (looser === null) {
            looser = engine._gl.getExtension('WEBGL_lose_context');
        }
        looser.restoreContext();
    }, 2000)

```
Will work some more on the PR in the evening (it is 19:00 over here now).

deltakosh:
Regarding lost context button on the Playground UI, I'm not for it but we can totally do it INSIDE a playground sample (and honestly also add it to our visual tests so we can validate it for each PR)

Also as we are on the same page, can you close the other PR?

Please note that I'm on holidays starting now until 1/4 so no rush :) You have all that time to clean this PR

breakin:
> Regarding lost context button on the Playground UI, I'm not for it but we can totally do it INSIDE a playground sample (and honestly also add it to our visual tests so we can validate it for each PR)
> 
> Also as we are on the same page, can you close the other PR?
> 
> Please note that I'm on holidays starting now until 1/4 so no rush :) You have all that time to clean this PR

Makes sense. Sounds like a good thing for test-suite! Closing other PR now and will polish this one instead!

src/Engines/thinEngine.ts
```diff
@@ -1779,7 +1779,16 @@ export class ThinEngine {
             var order = effect.getAttributeLocation(index);
 
             if (order >= 0) {
-                var vertexBuffer = vertexBuffers[attributes[index]];
+                var ai = attributes[index];
+                var vertexBuffer: Nullable<VertexBuffer> = null;
+
+                if (extraVertexBuffers) {
+                    vertexBuffer = extraVertexBuffers[ai];
```
breakin, line 1786: Here I decided that if the user adds a non-instanced VertexBuffer on the Mesh it will override the Geometry VertexBuffer of the same kind. That can come in handy if you want a one-off different UV-set or something.

src/Meshes/geometry.ts
breakin: Was just thinking that we should invalidate mesh-vaos as well but it ended up in Mesh._rebuild instead.

src/Meshes/geometry.ts
```diff
@@ -297,6 +296,11 @@ export class Geometry implements IGetSetVerticesData {
         if (this._vertexArrayObjects) {
             this._disposeVertexArrayObjects();
             this._vertexArrayObjects = {}; // Will trigger a rebuild of the VAO if supported
+
+            var meshes = this._meshes;
+            for (var index = 0; index < numOfMeshes; index++) {
+                meshes[index].invalidateInstanceVertexArrayObject();
```
breakin, line 302: Here we only want to do this if instances are supported. I'm not sure how to do that so I need some help. This is also part of the pieces of Mesh that has been split off into instancedMesh.ts.

src/Meshes/geometry.ts
```diff
@@ -672,6 +678,10 @@ export class Geometry implements IGetSetVerticesData {
 
         meshes.splice(index, 1);
 
+        if (this._vertexArrayObjects) {
+            mesh.invalidateInstanceVertexArrayObject();
```
breakin, line 682: Same here. How do I know that this function exist? Should I maybe do a dummy function on Mesh that is overwritten by instancedMesh.ts, like _disposeInstanceSpecificData ?

src/Engines/thinEngine.ts
deltakosh: Good for me!

src/Meshes/geometry.ts
deltakosh: We can simply cast it as any and test for invalidateInstanceVertexArrayObject

if ((meshes[index] as any).invalidateInstanceVertexArrayObject()) {
     (meshes[index] as InstanceMesh).invalidateInstanceVertexArrayObject();
}

src/Meshes/geometry.ts
deltakosh: same

src/Meshes/geometry.ts
breakin: I tried doing it using an empty function in Mesh and then a specialized function in instancedMesh.ts. Just for now, lets see how it feels.

breakin:
@deltakosh: I feel that all code is there now. I can't test context lost since no playgrounds seems to handle context lost atm. I do want to test it more though such that changing a buffer on the geometry properly updates VAO on the mesh etc so I think unless you have something big on the code itself I will start a testing phase now.

deltakosh:
Nope I like the code as is
Let's do a throughout testing and we can call it done ;)

deltakosh (approved):

deltakosh:
@sebavan: If all tests are coming green and @breakin is giving you a go from his own testing,we can merge this PR

breakin:
@sebavan: I've done what testing I can. I found and fixed one issue (the latest commit).

Also found this bug https://forum.babylonjs.com/t/applying-vertexdata-to-mesh-used-as-instancedmesh-source/16670 but not trying to fix it here.

breakin:
@sebavan: I am also interested in wording for the what's new and the commit message. I am assuming I will do a squashed merge, otherwise I want to rebase all my commits into one.

sebavan:
Do not worry about the commit message or rebasing :-)

About what s new, just a short sentence redirecting to the documentation and attributing it to you. Your message is perfect. It looks good to go, Congrats !!!!!

Can I merge it ????

breakin:
Yes. Happy!

breakin:
@sebavan: Yes, feel free to merge it!

PR #9619, opened 2020-12-17
Fix an issue with clear for Babylon Native

bghgary:


syntheticmagus (approved):
This is marked as a draft, so I'm not sure if I'm supposed to approve it yet, but I have no qualms about getting this (and its partner in the other repo) checked in to unblock someone now since the more permanent solution will take a bit more work.

sebavan (approved):

bghgary:
I marked it as draft since it has to be coordinated with the native change.

PR #9626, opened 2020-12-18
Fix IE11 shader processor creation

bghgary:


bghgary:
@Popov72 FYI, I didn't get a chance to test this with webgpu.

sebavan (approved):

PR #9635, opened 2020-12-21
Async serialization: Fix bug with serialization of base64 texture data

### LINKED ISSUE #9631: Make serialization asynchronous
See https://forum.babylonjs.com/t/loading-serialized-scene-containing-mesh-es-with-texture-s-leads-to-failed-to-execute-teximage2d-on-webgl2renderingcontext-overload-resolution-failed-error/16764/22

Popov72:
Fix #9631.

I did not want to update all the serialization methods and tried to perform minimal changes to avoid breaking things.

So, I ended up modifying only `SceneSerializer.Serialize` by adding a final step which scans all the properties of the serialization object, retrieves the ones that are promises and replace the value (the promise) by the result of the promise. That way, the various serialization methods don't need to be made async/changed, they only need to store a promise (which is what `Texture.Serialize` did) and this final stage is waiting for the promise resolutions before returning the result.

Note however that this is a breaking change, as `SceneSerializer.Serialize` now returns a promise. `SceneRecoder.getDelta` also now returns a promise.


sebavan:
I am wondering if we should do that or add a onSuccessCallback to the serialize function + an equivalent serializeAsync version wrapping the first one ??? 

Let s check what @deltakosh might think ???

This is kind of a pretty big breaking change, but I would say it is normal for serialization to be async.

sebavan:
Please note, this will need to be changed in the inspector relying on those functions but let s wait first on the final design so you ll have less changes to do :-)

sebavan (approved):

sebavan:
Just a tiny lint issue and I ll merge it in :-)

PR #9643, opened 2020-12-22
Add safety check to deleting textures in nativeEngine

Alex-MSFT:
It's possible to call deleteTexture on a null or undefined texture, which is not handled in BabylonNative and results in an access violation.  This adds a check to native engine to check the texture is not null before passing it on to native code.

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -1503,7 +1503,9 @@ export class NativeEngine extends Engine {
     }
 
     protected _deleteTexture(texture: Nullable<WebGLTexture>): void {
-        this._native.deleteTexture(texture);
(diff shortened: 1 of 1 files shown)


sebavan (approved):

src/Engines/nativeEngine.ts
```diff
@@ -1503,7 +1503,9 @@ export class NativeEngine extends Engine {
     }
 
     protected _deleteTexture(texture: Nullable<WebGLTexture>): void {
-        this._native.deleteTexture(texture);
+        if (texture) {
```
bghgary, line 1506: This check probably should be where _deleteTexture is called in thinEngine and this function shouldn't take a Nullable.

PR #9650, opened 2020-12-23
Various fixes after changes to instanced mesh vertex buffer management

Popov72:
See https://forum.babylonjs.com/t/why-instance-instancedbuffers-color-it-doesnt-work-on-bjs-version-5/16878/5


src/Meshes/buffer.ts
```diff
@@ -1,6 +1,7 @@
-import { Nullable, DataArray } from "../types";
+import { Nullable, DataArray, FloatArray } from "../types";
 import { ThinEngine } from "../Engines/thinEngine";
-import { DataBuffer } from './dataBuffer';
+import { DataBuffer } from "./dataBuffer";
+import { Tools } from "../Misc/tools";
```
sebavan, line 4: Ohhhhh not sure about this one cause size :-)

sebavan (commented):
Aside of the import it looks good to me

sebavan (approved):

PR #9657, opened 2020-12-24
Add a OptimizeNormals option to the OBJ loader to smooth lighting

Popov72:
See https://forum.babylonjs.com/t/obj-loader-computed-normals-are-wrong-if-optimize-with-uv-is-set-to-true/16896


sebavan (approved):

sebavan:
@Popov72 just a small lint issue and I ll merge it :-)

PR #9667, opened 2020-12-27
Added WSAD keys to freeCameraKeyboardMoveInput.ts 

geonom:
Added WSAD for movement in addition to the arrow keys to support default ego perspective movement as commonly used in first person games.

Would help usability and be more friendly to newbies like me, so I do not have to add dozens of lines of code just to get the standard WSAD movement going :smile:

I think, it does none of these:

    You cannot add code that will break backward compatibility
    You cannot add code that will slow down the rendering process
    You cannot add code that will make things complex to use

### DIFF
--- a/dist/what's new.md
+++ b/dist/what's new.md
@@ -128,6 +128,7 @@
 - Added upwards and downwards keyboard input to `FreeCamera` ([Pheater](https://github.com/pheater))
 - Handle scales in camera matrices ([Popov72](https://github.com/Popov72))
 - Added mouse wheel controls to FreeCamera. ([mrdunk](https://github.com/mrdunk))
+- Added WASD movement to FreeCamera. ([geonom](https://github.com/geonom))
 
 ### Debug
 - Added new view modes to the `SkeletonViewer` class. ([Pryme8](https://github.com/Pryme8))
--- a/src/Cameras/Inputs/freeCameraKeyboardMoveInput.ts
+++ b/src/Cameras/Inputs/freeCameraKeyboardMoveInput.ts
@@ -22,7 +22,7 @@ export class FreeCameraKeyboardMoveInput implements ICameraInput<FreeCamera> {
      * Gets or Set the list of keyboard keys used to control the forward move of the camera.
      */
     @serialize()
-    public keysUp = [38];
+    public keysUp = [38, 87];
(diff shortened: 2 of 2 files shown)


geonom:
Not sure why the two checks are still failing - added the Whats new entry but it does not help for some reason 😢 

On the other hand, I created my first push request and most checks worked out (though honestly, I added four numbers, but yay 🎉 ) 😄

sebavan:
Thanks a lot for the PR. The what s new fails cause you modified the one in dist and not dist/preview release folder. 

I am not sure we should integrate it as this is language specific unlike arrows. I would prefer to close the PR and you use this setup locally. I will nevertheless add @PolygonalSun who is working on inputs and would probably want to integrate that as a standard on the future version.

Would that work for you ?

geonom:
Not sure why the Whats New Update check is still failing 😢 

Thought about backwards compatibility a bit more - but as far as I understand, any keys set via custom code should still work as they are set, as this just extends a default value, no?

> 
> 
> Thanks a lot for the PR. The what s new fails cause you modified the one in dist and not dist/preview release folder.
> 
> I am not sure we should integrate it as this is language specific unlike arrows. I would prefer to close the PR and you use this setup locally. I will nevertheless add @PolygonalSun who is working on inputs and would probably want to integrate that as a standard on the future version.
> 
> Would that work for you ?

Sure, after your input I read a bit more and now I feel a bit stupid for not having thought about that myself 😅 

Found this article helpful though a bit dated perhaps: [https://hacks.mozilla.org/2017/03/internationalize-your-keyboard-controls/](https://hacks.mozilla.org/2017/03/internationalize-your-keyboard-controls/)

One proposed "solution" here is using that feature and trying to use the physical location of the keys instead of the alphatic values, as far as I understand: [https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Handle_keyboard_events_in_a_game](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Handle_keyboard_events_in_a_game)

So far, thanks for your input and time - and fully agree on the reason not to merge this 👍

sebavan:
And rest ensured that @PolygonalSun will put it in once he ll reach this step :-)

PR #9670, opened 2020-12-28
NME texture blocks: dispose textures when disposing of blocks

Popov72:
Also add some more info when logging the error message when a gl `createShader` call fails.

### DIFF
--- a/src/Materials/Node/Blocks/Particle/particleTextureBlock.ts
+++ b/src/Materials/Node/Blocks/Particle/particleTextureBlock.ts
@@ -218,6 +218,11 @@ export class ParticleTextureBlock extends NodeMaterialBlock {
             this.texture = Texture.Parse(serializationObject.texture, scene, rootUrl) as Texture;
         }
     }
+
+    public dispose() {
+        super.dispose();
+        this.texture?.dispose();
+    }
 }
 
 _TypeStore.RegisteredTypes["BABYLON.ParticleTextureBlock"] = ParticleTextureBlock;
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -2249,10 +2249,13 @@ export class ThinEngine {
 
     private _compileRawShader(source: string, type: string): WebGLShader {
         var gl = this._gl;
+
+        while (gl.getError() != gl.NO_ERROR) { }
+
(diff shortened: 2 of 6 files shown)


sebavan (approved):

src/Materials/Node/Blocks/Particle/particleTextureBlock.ts
```diff
@@ -218,6 +218,11 @@ export class ParticleTextureBlock extends NodeMaterialBlock {
             this.texture = Texture.Parse(serializationObject.texture, scene, rootUrl) as Texture;
         }
     }
+
+    public dispose() {
+        super.dispose();
+        this.texture?.dispose();
```
deltakosh, line 224: I would challenge that texture is always defined :)

Also, I recommend to NOT dispose the texture or at least give the user the opportunity to NOT do it:
https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/material.ts#L1397

(Think about the case of users using the same texture  in different blocks)

@Popov72  @sebavan

Popov72: > I would challenge that texture is always defined :)

I don't think so, it is defined as `public texture: Nullable<BaseTexture>;` and the user may not bind a texture in the block in the nme :)

> Also, I recommend to NOT dispose the texture or at least give the user the opportunity to NOT do it:

I did not add this because the `dispose` method in `InternalTexture` is doing a reference counting and does not release the texture as long as there is a reference pending.

This doc in `Material.dispose` is a bit misleading:
```
forceDisposeTextures specifies if textures should be forcefully disposed
```
because of the ref count: the texture won't be forcefully disposed if ref count > 0 after decrementing it.

It seems strange to me that `forceDisposeTextures` exists at all: it `false`, it means a texture can be referenced by no code and still exists, meaning it won't ever be released (as it is not reachable by any code)... Or it means a copy of the texture has been done but the ref count has not been incremented (which does not seem right). I think that if the user wants the texture to stay alive, they should copy it / increment the ref count(?)

deltakosh: I disagree :) ref counting is on internal textures not on BaseTexture

PR #9686, opened 2020-12-31
Light Gizmo expression statement is not assignment or call

jasonsturges:
Presume the intention here is equality assignment?


sebavan (changes requested):
As the light gizmo is well behaving, I guess we should just remove the line instead. @jasonsturges  could you make the change ?

sebavan:
I Updated it :-)

PR #9689, opened 2020-12-31
Sound switch fallthrough

jasonsturges:
Is it intentional for case `String` to fallthrough to `Array` here?

Presume a `break` is needed?


sebavan:
this one is ok and normal to fall through

PR #9715, opened 2021-01-02
Prevent AmmoJS applyForce modifying position param

Fowlet:
Took a while for me to work out that the AmmoJS physics plugin's `applyForce()` function was modifying the `contactPoint` parameter for the purposes of converting it to local space; this broke my surrounding code that was relying on using the passed parameter as a cached value.

I've worked around it for now, but I'm guessing it shouldn't be like this? Maybe the physics plugin interfaces need modifying to take a `DeepImmutable<Vector3>` instead of a `Vector3` to enforce good practice?

### DIFF
--- a/src/Physics/Plugins/ammoJSPlugin.ts
+++ b/src/Physics/Plugins/ammoJSPlugin.ts
@@ -377,13 +377,14 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
             impostor.physicsBody.activate();
             var worldPoint = this._tmpAmmoVectorA;
             var impulse = this._tmpAmmoVectorB;
+            var localContactPoint = new Vector3().copyFrom(contactPoint);
 
             // Convert contactPoint relative to center of mass
             if (impostor.object && impostor.object.getWorldMatrix) {
-                contactPoint.subtractInPlace(impostor.object.getWorldMatrix().getTranslation());
+                localContactPoint.subtractInPlace(impostor.object.getWorldMatrix().getTranslation());
             }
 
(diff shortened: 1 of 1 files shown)


src/Physics/Plugins/ammoJSPlugin.ts
```diff
@@ -377,13 +377,14 @@ export class AmmoJSPlugin implements IPhysicsEnginePlugin {
             impostor.physicsBody.activate();
             var worldPoint = this._tmpAmmoVectorA;
             var impulse = this._tmpAmmoVectorB;
+            var localContactPoint = new Vector3().copyFrom(contactPoint);
```
sebavan, line 380: should not be a new here for garbage collection, would be great to use a tmp vector :-)

sebavan (changes requested):
LGTM aside the GC issue Thanks a lot.

RaananW:
Any update on this one?

src/Physics/Plugins/ammoJSPlugin.ts
Fowlet: (Maybe) addressed in https://github.com/BabylonJS/Babylon.js/pull/9715/commits/744a4dcdbd9e96442340b4dea40b4e3cc34ad08f

Fowlet:
`getWorldMatrix().getTranslation()` still results in a call of `new Vector3()` [here](https://github.com/BabylonJS/Babylon.js/blob/master/src/Maths/math.vector.ts#L4030), but I think its no worse than it was now.

sebavan (approved):

PR #9723, opened 2021-01-04
PBR Custom Material

MackeyK24:
Inline Shader Support For PBRCustomMaterial


materialsLibrary/src/custom/pbrCustomMaterial.ts
```diff
@@ -109,7 +109,16 @@ export class PBRCustomMaterial extends PBRMaterial {
         return arr;
     }
 
-    public Builder(shaderName: string, uniforms: string[], uniformBuffers: string[], samplers: string[], defines: MaterialDefines | string[], attributes?: string[]): string {
+    public Builder(shaderName: string, uniforms: string[], uniformBuffers: string[], samplers: string[], defines: MaterialDefines | string[], attributes?: string[], options?: BABYLON.ICustomShaderNameResolveOptions): string {
+        options!.processFinalCode = (type: string, code: string) => {
```
sebavan, line 113: This will break if no options are passed in. You should check for null before assigning here ?

materialsLibrary/src/custom/pbrCustomMaterial.ts
```diff
@@ -109,7 +109,18 @@ export class PBRCustomMaterial extends PBRMaterial {
...
+            options.processFinalCode = (type: string, code: string) => {
+                if (type === "vertex") {
+                    return code;
+                }
+                const sci = new BABYLON.ShaderCodeInliner(code);
+                sci.inlineToken = "#define pbr_inline";
+                sci.processCode();
+                return sci.code;
+            };
+        }
```
Popov72, line 123: I don't think this code should be added unconditionnally as it is a processing that the user may not want to do. Maybe add a switch somewhere (`false` by default to be back compat) to trigger it?

Also, `options.processFinalCode` may already have a value, in which case you should call the existing callback before doing your own processing.

sebavan: Yup @MackeyK24 do you prefer a flag added for it ? I can do a PR for that.

PR #9724, opened 2021-01-04
ShaderMaterial: Add support for morph targets

Popov72:
https://forum.babylonjs.com/t/animation-support-on-custom-shaders/12628/10


EricEisaman:
Does this enable per instance morph target influences?

sebavan (approved):

sebavan:
@Popov72 looks overall good but the tests are failing due to 
image

Popov72:
> Does this enable per instance morph target influences?

No, this PR adds morph target support for ShaderMaterial, but all instances are using the same targets.

PR #9729, opened 2021-01-04
If indices are empty use unindexed intersection

RaananW:
Unindexed meshes are now supported when being picked.


CedricGuillemet:
Does it work with point cloud? IIRC pointclouds have an empty index array. See this PG : https://www.babylonjs-playground.com/#FHHM51#34

sebavan:
Good question, in this case I ll remove it from the support, let me check

sebavan:
All good cause in this case they are point list draw mode not supported here.

PR #9740, opened 2021-01-05
Depth of field post process

jasonsturges:
Unsure whether this is a breaking change - but suggest changing verticle => vertical as the `name` parameter.

I don't immediately see how this would break something; so, hoping for your expertise here.


deltakosh:
All good! thanks

PR #9741, opened 2021-01-05
Probes documentation

jasonsturges:
Minor `Probes` folder documentation suggested updates:

- flaot => float
- relfection => refection
- resepective => respective


RaananW:
I think flaot is a great data type. somewhere between a float and a carrot.

PR #9747, opened 2021-01-05
support for native shadows

CedricGuillemet:
JS part of it


sebavan (approved):
approving blindly, but I trust you :-) and I like pressing buttons !!!

RaananW:
Merged, because I like pressing buttons as well.

PR #9752, opened 2021-01-06
WebGPU: optimize render pipeline and bind group caching

Popov72:
Note that `WebGPUCacheRenderPipelineString` (it is the render pipeline cache using strings as keys to the cache) is not used in the final code but was used during testing and as it is very small (18 lines of code), I kept it in the source base.

`WebGPUCacheRenderPipelineTree` is implementing the caching strategy using a tree of nodes (one node==one render state (or multiple render states when multiple states could be combined into a single number)). To speed things up, the traversing of the tree starts at the latest modified node and not at the root of the tree. To optimize this (meaning, starting at the deepest node possible to traverse as few nodes as possible), the tree is constructed so that the first nodes (stencil read mask, stencil write mask, ...) are less likely to change between two draw calls than the latest nodes (..., color states, shader stage, vertex state).

The bind group cache also uses a tree of nodes (instead of a lookup in a hash map in the previous iteration) but is a lot simpler and so does not have its own class (only the node has its class: `WebGPUBindGroupCacheNode`).

This PR also fixes the warnings generated by Chrome when using anisotropy.


sebavan (approved):

src/Materials/Textures/mirrorTexture.ts
```diff
@@ -171,8 +171,7 @@ export class MirrorTexture extends RenderTargetTexture {
             Matrix.ReflectionToRef(this.mirrorPlane, this._mirrorMatrix);
             this._mirrorMatrix.multiplyToRef(scene.getViewMatrix(), this._transformMatrix);
 
-            // Clone to not mark matrices as updated
-            scene.setTransformMatrix(this._transformMatrix.clone(), scene.getProjectionMatrix().clone());
+            scene.setTransformMatrix(this._transformMatrix, scene.getProjectionMatrix());
```
sebavan, line 174: Only one question on this one ??? wonder if it does not impact  ?

src/Meshes/buffer.ts
```diff
@@ -261,6 +261,7 @@ export class VertexBuffer {
         } else {
             this._instanced = true;
         }
+        this._computeHashCode();
```
sebavan, line 264: Is this also not heavy for gl ?

Popov72: The computation is only doing some bit operations and is done once per vertex buffer, so I don't think it's too much of a problem (we don't create a buffer each frame).

I have moved this computation here because else it was (re)done each time we access the render pipeline cache in WebGPU, so once per draw call.

src/Materials/Textures/mirrorTexture.ts
Popov72: Removing the clone calls has been forgotten, see https://github.com/BabylonJS/Babylon.js/pull/9591#discussion_r548036894

PR #9755, opened 2021-01-06
updatable capsule

avin:
added missed `updatable` param

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = crlf
+charset = utf-8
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+# editorconfig.org
+root = true
```
deltakosh, line 2: why this change? I guess this file should be removed

deltakosh (approved):

.editorconfig
avin: it defines setups for editors (analogue of .vscode/settings.json, but without reference to a specific one)

deltakosh: ok why not:)

you have some checks failing and I can merge

PR #9769, opened 2021-01-08
[Shared Component] LineContainerComponent

msDestiny14:
For this move there were 2 different implementations in GUI and the Inspector. I ended up choosing the GUI implementation since it did not rely on global state.


msDestiny14:
@deltakosh I tested in local dev and there seemed to be no difference but if you want to double check this is fine. :)

RaananW (approved):
Tested locally, LGTM!

PR #9782, opened 2021-01-12
Mark node materials dirty when texture or reflection texture is changed

Popov72:
Hoping that replacing the `texture` property by getter/setter does not break anything...

### DIFF
--- a/src/Materials/Node/Blocks/Dual/reflectionTextureBaseBlock.ts
+++ b/src/Materials/Node/Blocks/Dual/reflectionTextureBaseBlock.ts
@@ -76,17 +76,17 @@ export abstract class ReflectionTextureBaseBlock extends NodeMaterialBlock {
         const scene = texture?.getScene() ?? Engine.LastCreatedScene;
 
         if (!texture && scene) {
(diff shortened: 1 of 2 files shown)


src/Materials/Node/Blocks/Dual/reflectionTextureBaseBlock.ts
```diff
@@ -59,10 +60,35 @@ export abstract class ReflectionTextureBaseBlock extends NodeMaterialBlock {
...
+
+    public set texture(texture: Nullable<BaseTexture>) {
+        if (this._texture === texture) {
+            return;
+        }
+
+        const scene = texture?.getScene() ?? Engine.LastCreatedScene;
+
+        if (!texture && scene) {
+            /*scene.markAllMaterialsAsDirty(Constants.MATERIAL_TextureDirtyFlag, (mat) => {
```
deltakosh, line 79: why this comments?

Popov72: Hum, I thought I had pushed a fix but it seems not...

deltakosh (approved):

PR #9785, opened 2021-01-12
[GUI Components] Slider

msDestiny14:
Moving Slider property tab from inspector to sharer. Then hooked up to the GUI Editor. 
Users can now full modify the properties of slider. 

Slider


guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -26,14 +29,22 @@ interface IPropertyTabComponentState {
...
     }
 
+    timerRefresh() {
+        if (!this._lockObject.lock) {
+            this.forceUpdate();
+        }
+    }
+
     componentDidMount() {
+        this._timerIntervalId = window.setInterval(() => this.timerRefresh(), 500);
```
RaananW, line 47: If i remember correctly, this was done somewhere else as well. The question is, can it be prevented here? Is there no callback to check that it had changed, instead of checking every 500 ms?

RaananW (approved):
LGTM! I do have one comment about the interval used. If there is no other solution I'm perfectly fine with it. If you think there might be and want to look for one together, let me know

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Discussed online :)

PR #9789, opened 2021-01-12
Add DeviceInputSystem to Input Manager

PolygonalSun:
This PR contains all of the changes to update the InputManager (and other systems using the InputManager) to use the DeviceInputSystem for its event foundation.  This is a remake of a previous PR with changes to account for the jump to 5.0 (https://github.com/BabylonJS/Babylon.js/pull/9194)


src/Cameras/Inputs/BaseCameraMouseWheelInput.ts
```diff
@@ -61,9 +62,9 @@ export abstract class BaseCameraMouseWheelInput implements ICameraInput<Camera>
                 return;
             }
 
-            const event = <MouseWheelEvent>pointer.event;
+            const event = <IWheelEvent>pointer.event;
 
-            const platformScale = event.deltaMode === WheelEvent.DOM_DELTA_LINE ? this._ffMultiplier : 1;
+            const platformScale = event.deltaMode === 0x01 ? this._ffMultiplier : 1;  // If this happens to be set to DOM_DELTA_LINE, adjust accordingly
```
deltakosh, line 67: we need to NOT use unnamed constants

PolygonalSun: I changed this to a constant that's maintained in the file that has our event interfaces.

RaananW (commented):
LGTM!

The only thing I would ask it to change all `==` to `===` and `!=` to `!==` to keep type safety in equations. Perf will be slightly better as well (though probably negligible)

PolygonalSun:
> LGTM!
> 
> The only thing I would ask it to change all `==` to `===` and `!=` to `!==` to keep type safety in equations. Perf will be slightly better as well (though probably negligible)

Okay, I changed over what I saw in the InputManager and DeviceInputSystem

PolygonalSun:
Tagging @deltakosh and @CedricGuillemet

deltakosh (approved):

PR #9791, opened 2021-01-13
[GUI Components] Adding all GUI Property Tab Components

msDestiny14:
Adding the remainder property tab components to shared UI and then hooking up in GUI Editor.

### DIFF
--- a/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
+++ b/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
@@ -204,90 +204,93 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
     renderProperties()
     {
         var className = this.state.currentNode?.guiControl.getClassName();
-        if (className === "TextBlock") {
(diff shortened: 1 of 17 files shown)


guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -175,12 +204,96 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
     renderProperties()
     {
         var className = this.state.currentNode?.guiControl.getClassName();
+        if (className === "TextBlock") {
```
RaananW, line 207: A switch would be much more efficient in this case (especially since it can't two classes at the same time)

RaananW (changes requested):
My only comment is about the switch (or `else if` instead of `if`). Otherwise good to go :-)

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: There we go :)

deltakosh (approved):

PR #9797, opened 2021-01-14
P8/nine patch patch

Pryme8:
I have this set as a draft right now because I want to see if I should simplify my math equations and save some calculations.

Then also verify that removing "populateNinePatchSlicesFromImage" stuff is not an issue with something I did not see.  But from looking over it, there was no real use for it anywhere I saw and was never called essentially.


Pryme8:
https://www.babylonjs-playground.com/#G5H9IN#27

here is the working proof.

gui/src/2D/controls/image.ts
```diff
@@ -818,51 +746,35 @@ export class Image extends Control {
...
-
-        // Borders
-        this._drawImage(context, left, this._sliceTop, leftWidth, this._sliceBottom - this._sliceTop, this._currentMeasure.left, this._currentMeasure.top + topHeight, leftWidth, targetTopHeight - topHeight);
-
-        this._drawImage(context, this._sliceRight, this._sliceTop, leftWidth, this._sliceBottom - this._sliceTop, this._currentMeasure.left + this._currentMeasure.width - rightWidth, this._currentMeasure.top + topHeight, leftWidth, targetTopHeight - topHeight);
-
-        this._drawImage(context, this._sliceLeft, top, centerWidth, topHeight, this._currentMeasure.left + leftWidth, this._currentMeasure.top, targetCenterWidth, topHeight);
-
-        this._drawImage(context, this._sliceLeft, this._sliceBottom, centerWidth, bottomHeight, this._currentMeasure.left + leftWidth, this._currentMeasure.top + targetTopHeight, targetCenterWidth, bottomHeight);
+        let height = this._imageHeight, width = this._imageWidth;
```
deltakosh, line 750: please use the same formatting with a let per variable

Pryme8: yesir

Pryme8:
oop missed one

Pryme8:
Gonna do a build and get that lint error fixed.

Pryme8:
"Validation Tests Displacement map FAILED"

What would this have to do with my commit?

Popov72:
> What would this have to do with my commit?

It's a timeout problem, you can ignore this error.

> Then also verify that removing "populateNinePatchSlicesFromImage" stuff is not an issue with something I did not see. But from looking over it, there was no real use for it anywhere I saw and was never called essentially.

I don't think you can remove this code, it's not because it is not used inside the engine itself that it is not used by end users: `populateNinePatchSlicesFromImage` has public getter / setter. For eg, there are two PGs that are using it: https://doc.babylonjs.com/playground?q=populateNinePatchSlicesFromImage&type=code

Also, there's a problem with your code: https://playground.babylonjs.com/#G5H9IN#29

The second image should be twice as big as the first one, but it seems it is half of it instead.

See this PG where it is ok: https://playground.babylonjs.com/#UW6XWQ

Note that the code of `_renderNinePatch` in this PG is the existing one but with the `if (this._populateNinePatchSlicesFromImage) { ... }` commented out, as else there's a hole in the display. I don't really know why there would be some corrections applied when the slice left/top/right/bottom values are computed from a picture (meaning when `populateNinePatchSlicesFromImage = true`) instead of being given by the user directly...

I think that when everything is sorted out we should add new validation tests to cover the 9-patch code.

Pryme8:
Agreed!
How does populateNinePatchSlicesFromImage actually grab those slice values?  Where is that information stored on the image,  I don't understand how it works I guess and when I looked over the code I cant seem to establish why using that flag works.  

I guess it checks the images alpha and then tries to approximate it?  That seems pretty shotty and very specific use cases.

Also thank  you for the review, I was able to fix the scale thingy.  Now just need to figure out the last param stuff that with the computer from image.

Pryme8:
Ok this should be close.

Pryme8:
Why was this closed?

deltakosh:
image

deltakosh:
Ok gotcha! You wanted to merge it actually :D

lol...Merging now!

Pryme8:
<3 thank you sir

Popov72:
`populateNinePatchSlicesFromImage` simply assumes that the corner / side parts have alpha <= 127 and the center part has alpha > 127.

So, for the horizontal dimension, it detects slice_left as the first time the alpha value of the pixel goes above 127 and it detects slice_right when slice_left has already been set and the first time the alpha value of the pixel goes below 127.

Same algorithm for the vertical dimension.

deltakosh (approved):

deltakosh:
You're ok to merge @Popov72 ?

Popov72:
I don't think it's ok to merge, we still have:
```typescript
        let widthFactor = width / this._currentMeasure.width;
        let heightFactor = height / this._currentMeasure.height;
```
whereas it is:
```typescript
    let widthFactor = this._currentMeasure.width/width;
    let heightFactor = this._currentMeasure.height/height;
```
In the PG that works.

Popov72:
Also I think there should be an entry in the bug fix section of the what's new file.

deltakosh:
Agree on both

Pryme8:
Whoops sorry I forgot to make that fix on my local!

deltakosh:
Please update what's new ;)
and waiting for @Popov72 validation and we can merge

Pryme8:
What should I say? That there where corrections made to accommodate for pixel error on firefox?

deltakosh:
yep

Popov72:
That's ok for me when the what's new file is updated.

Pryme8:
I will do that today after my wife can help me with the baby here! Just need both my hands.

Pryme8:
Hold off on merging this, I think I found a bug.

Pryme8:
So after lots more testing I can validate that the the "_populateNinePatchSlicesFromImage" does not work as intended and distorts both where the slices should be and the final output image, so I had to go back and fix that.  While doing that I realized that the new stuff I was doing also distorted the image and did no let the user keep the fixed border size.

image
Old/Original Method

image
New/Purposed Method

https://playground.babylonjs.com/#G5H9IN#46

I will revise my PR accordingly.

Pryme8:
and after more testing it looks like I need a hair more corrections... dang it one second...

Pryme8:
Oooooohhhhkay, I think this is it.  I was able to get the populate from Image working correctly and the pixel gaps in Firefox are now gone.

deltakosh:
Thanks buddy!

PR #9798, opened 2021-01-15
[Gui Editor] Adding Loading and Saving

msDestiny14:
Users can now save and load their gui locally or to the snippet server. This also be used in playgrounds. 
There is also code for the beginning of the parenting system.


guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -96,109 +98,76 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+        let content = JSON.stringify(adt.serializeContent());
 
+        var xmlHttp = new XMLHttpRequest();
         xmlHttp.onreadystatechange = () => {
             if (xmlHttp.readyState == 4) {
                 if (xmlHttp.status == 200) {
                     var snippet = JSON.parse(xmlHttp.responseText);
-                    const oldId = material.snippetId;
-                    material.snippetId = snippet.id;
+                    const oldId = adt.snippetId || "_BLANK";
```
RaananW, line 121: where is _BLANK coming from?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -96,109 +98,76 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+                        navigator.clipboard.writeText(adt.snippetId);
                     }
 
                     let windowAsAny = window as any;
 
                     if (windowAsAny.Playground && oldId) {
                         windowAsAny.Playground.onRequestCodeChangeObservable.notifyObservers({
-                            regex: new RegExp(oldId, "g"),
-                            replace: material.snippetId
+                            regex: new RegExp(`parseFromSnippetAsync\\("${oldId}`, "g"),
```
RaananW, line 135: can you explain the regex/replace paradigm?

I am asking since regex has `\\` (which in this case is actually a single `\` after a break), but the replace doesn't even have a single `\`

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -301,6 +270,8 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
     }
 
     render() {
+
+        //var myScene=this.props.globalState.guiTexture.getScene();
```
RaananW, line 274: nit - cleanup?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -344,15 +315,9 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
                         <ButtonLineComponent label="Save" onClick={() => {
                             this.save();
                         }} />
-                        {
-                            this.props.globalState.customSave &&
-                            <ButtonLineComponent label={this.props.globalState.customSave!.label} onClick={() => {
```
RaananW, line 349: Was this added somewhere else? or is it simply not needed?

guiEditor/src/diagram/guiNode.ts
```diff
@@ -102,19 +110,23 @@ export class GUINode {
     public constructor(globalState: GlobalState, public guiControl: Control) {
         this._globalState = globalState;
         this._ownerCanvas = this._globalState.workbench;
-        
+        this.x = guiControl.leftInPixels;
+        this.y = guiControl.topInPixels;
         guiControl.onPointerUpObservable.add(evt => {
             this.clicked = false;
             console.log("up");
```
RaananW, line 115: (not part of this PR, but...) is this console (and the others in this file)  needed?

guiEditor/src/diagram/guiNode.ts
```diff
@@ -102,19 +110,23 @@ export class GUINode {
...
-        
+        this.x = guiControl.leftInPixels;
+        this.y = guiControl.topInPixels;
         guiControl.onPointerUpObservable.add(evt => {
             this.clicked = false;
             console.log("up");
         });
 
         guiControl.onPointerDownObservable.add( evt => {
+            if(!this._ownerCanvas.isUp) return;
```
RaananW, line 119: Our tslint/prettier should prevent this from passing lint, but we are not yet linting those files. Will you be able to run a formatter (prettier?) on these files?

guiEditor/src/diagram/guiNode.ts
```diff
@@ -151,16 +165,53 @@ export class GUINode {
...
+        });
+
         return true;
         //evt.stopPropagation();
     }
 
+    renderContainer(): React.ReactNode {
+        if(!this._isContainer) return null;
+
+        return React.createElement(ContainerPropertyTabComponent, {
```
RaananW, line 177: any reason this isn't a tsx file and render jsx/html fragement here?

guiEditor/src/diagram/properties/containerPropertyComponent.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { IContainerComponentProps } from "./propertyComponentProps";
+
+export class ContainerPropertyTabComponent extends React.Component<IContainerComponentProps> {
+    constructor(props: IContainerComponentProps) {
+        super(props);
+    }
+
+    AddElement()
+    {
+        let newElementName = window.prompt("Please enter name of the element");
```
RaananW, line 16: I assume this will be extended? Want to remove it from here for now?

guiEditor/src/guiEditor.ts
```diff
@@ -57,7 +56,7 @@ export class GUIEditor {
 
         if (options.customLoadObservable) {
             options.customLoadObservable.add((data) => {
-                SerializationTools.Deserialize(data, globalState);
+                //TODO: Add deserilization here.
```
RaananW, line 59: are you tracking this somehow?

guiEditor/src/stringTools.ts
```diff
@@ -0,0 +1,87 @@
...
+    private static _SaveAs(blob: Blob, name: string, document: HTMLDocument) {
+        if ('download' in HTMLAnchorElement.prototype) {
+            var URL = window.URL || window.webkitURL;
+            var a = document.createElement('a');
+        
+            a.download = name;
+            a.rel = 'noopener'; // tabnabbing
+
+            a.href = URL.createObjectURL(blob)
+            setTimeout(() => { URL.revokeObjectURL(a.href) }, 4E4) // 40s
```
RaananW, line 20: 40 seconds?

guiEditor/src/stringTools.ts
```diff
@@ -0,0 +1,87 @@
+export class StringTools {
```
RaananW, line 1: Don't we have some sort of file saver in the framework already? I am pretty sure I have seen similar code. Doesn't NME have the same thing?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Yes this is not needed at the moment but will come back eventaully.

guiEditor/src/diagram/guiNode.ts
msDestiny14: For now I'd like to keep it in since I still do a lot of debugging around that. Is that considered proper?

guiEditor/src/diagram/guiNode.ts
msDestiny14: Not really. Originally this was paired besides the other render but since that function is gone maybe I should rethink this.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -96,109 +98,76 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+        let json = JSON.stringify(this.props.globalState.guiTexture.serializeContent());
+        StringTools.DownloadAsFile(this.props.globalState.hostDocument, json, "guiTexture.json");
     }
 
     saveToSnippetServer() {
-        /*const material = this.props.globalState.nodeMaterial;
-        const xmlHttp = new XMLHttpRequest();
-
-        let json = SerializationTools.Serialize(material, this.props.globalState);
+        var adt = this.props.globalState.guiTexture;
```
Popov72, line 113: Use `let` or `const` instead of `var` (there are other occurrences of `var` througout the code).

Also, you should favor `const` over `let` whenever possible.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Ya. I'm not sure when there would be a time when snippetId doesn't exist.

guiEditor/src/guiEditor.ts
msDestiny14: You mean in github?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
Popov72: Actually the regex is `parseFromSnippetAsync\(...` but because it is hosted in a js string the `\` must be escaped to `\\`, so I think it's fine, it's really the `(` which is escaped in the expression (but something like `parseFromSnippetAsync\\s*\\(` would be safer).

But I don't understand why the previous code had to be changed: is it for additional safety (though there's little chance that the snipped id string would appear somewhere in the code without really being the snipped id we need to replace)?

guiEditor/src/diagram/guiNode.ts
```diff
@@ -151,16 +163,53 @@ export class GUINode {
...
+        });
+    }
+
     public updateVisual()
     {
         this.guiControl.leftInPixels = this.x;
         this.guiControl.topInPixels = this.y;
     }
 
+    private isContainer() {
```
Popov72, line 189: Missing the leading underscore to the function name.

guiEditor/src/diagram/properties/containerPropertyComponent.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { LineContainerComponent } from "../../sharedUiComponents/lines/lineContainerComponent";
+import { TextLineComponent } from "../../sharedUiComponents/lines/textLineComponent";
+import { IContainerComponentProps } from "./propertyComponentProps";
+
+export class ContainerPropertyTabComponent extends React.Component<IContainerComponentProps> {
+    constructor(props: IContainerComponentProps) {
+        super(props);
+    }
+
+    AddElement()
```
Popov72, line 14: Missing a public/protected/private modifier and probably the first letter should be lowercase (except if it is a static function?).

guiEditor/src/diagram/properties/propertyComponentProps.ts
```diff
@@ -1,7 +1,13 @@
 import { Control } from "babylonjs-gui/2D/controls/control";
 import { GlobalState } from "../../globalState";
+import { GUINode } from "../guiNode";
 
 export interface IPropertyComponentProps {
     globalState: GlobalState;
-    guiControl: Control;
+    guiBlock: Control;
```
Popov72, line 7: Did you have a specific reason to rename `guiControl` to `guiBlock`? `guiControl` seems better than `guiBlock` given that the type is `Control` . Or was it too generic?

msDestiny14:
I removed some of the container related code since it's going to be reworked anyways in the future.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -110,15 +109,15 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
     }
 
     saveToSnippetServer() {
-        var adt = this.props.globalState.guiTexture;
+        let adt = this.props.globalState.guiTexture;
```
Popov72, line 112: Most of the `let` you added could be `const` instead :)

If in doubt, just always use `const` and change to `let` if the transpiler does complain.

nodeEditor/src/diagram/display/inputDisplayManager.ts
```diff
@@ -46,15 +45,19 @@ export class InputDisplayManager implements IDisplayManager {
...
 
         if (inputBlock.isAttribute) {
-            name = StringTools.GetBaseType(inputBlock.output.type);
+            name = this.GetBaseType(inputBlock.output.type);
         }
 
         return name;
     }
 
+    public GetBaseType(type: NodeMaterialBlockConnectionPointTypes): string {
```
Popov72, line 57: You should either make this method static or have the first letter of its name be lowercase.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: I think it was just implementations taken form different places since original the code was commented. I can revert back to the original (assuming it works)

msDestiny14: Yep it does! :)

guiEditor/src/diagram/guiNode.ts
RaananW: As long as it goes out for production, i am totally fine with it. Maybe just add a TODO to follow up

guiEditor/src/guiEditor.ts
RaananW: yep, some issue you can reference here, or some external todo list?
TODOs tend to stay in place...

guiEditor/src/diagram/guiNode.ts
msDestiny14: Totally!

RaananW (approved):

guiEditor/src/guiEditor.ts
msDestiny14: Making git issues for next PR.

guiEditor/src/diagram/guiNode.ts
msDestiny14: I'll make and issue for that.

RaananW:
Just need to resolve the conflicts and hit the merge button :-)

PR #9801, opened 2021-01-15
Fix for undefined value on keyup event

PolygonalSun:
It was reported in the forums that there was an error with the _keyboardUpEvent function producing an error if a keyup event occurred before a keydown event.  For the DeviceInputSystem, we only initialize the Keyboard input when we get a keydown event as we shouldn't get an up event before a down event.  Because of this, if we get a keyup event, we have no "keyboard" to reference and change.

This PR added a check for this edge case to initialize the keyboard on any key events.


deltakosh (approved):

thebravoman:
Thanks guys. That was fast.

PR #9803, opened 2021-01-16
Add a helper class to display the frustum of a directional light

Popov72:
Used in the inspector:

image

It's updated in real time if any change in the light parameters impacts the frustum.


src/Debug/directionalLightFrustumViewer.ts
```diff
@@ -0,0 +1,349 @@
...
+
+        this._tlLinesPoints[0] = TmpVectors.Vector3[4];
+        this._tlLinesPoints[1] = TmpVectors.Vector3[8];
+
+        this._blLinesPoints[0] = TmpVectors.Vector3[5];
+        this._blLinesPoints [1] = TmpVectors.Vector3[9];
+    }
+
+    protected _getInvertViewMatrix(): Matrix {
+        Matrix.LookAtLHToRef(this._light.position, this._light.position.add(this._light.direction), Vector3.Up(), this._inverseViewMatrix);
```
RaananW, line 345: Is this called on each frame (or only on update?) `Vector.Up()` will create a new vector on each call.

deltakosh: We have Vector.Upreadonly I guess

RaananW (approved):

PR #9813, opened 2021-01-19
obstacles and corners

CedricGuillemet:
followup on https://forum.babylonjs.com/t/navigation-mesh-and-crowd-agents/4609/86
Doc in progress


src/Navigation/INavigationEngine.ts
```diff
@@ -156,12 +156,42 @@ export interface INavigationEnginePlugin {
...
     getMaximumSubStepCount(): number;
 
+    /**
+     * Creates a cylinder obstacle and add it to the navigation
+     * @param position world position
+     * @param radius cylinder radius
+     * @param height cylinder height
+     * @returns the obstacle freshly created
+     */
+    addCylinderObstacle(position: Vector3, radius: number, height: number): IObstacle;
```
Popov72, line 166: Does this signature mean the cylinder is always parallel to the Y axis? We can't have an arbitrary rotated cylinder?

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -692,6 +726,25 @@ export class RecastJSCrowd implements ICrowd {
...
     }
 
+    /**
+     * Get the next corner points composing the path (max 4 points)
+     * @param index agent index returned by addAgent
+     * @returns array containing world position composing the path
+     */
+    getCorners(index: number): Vector3[]
+    {
+        var pt: number;
```
Popov72, line 736: `let`?

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -692,6 +726,25 @@ export class RecastJSCrowd implements ICrowd {
...
+        let navPath = this.recastCrowd.getPath(index);
+        let pointCount = navPath.getPointCount();
+        var positions = [];
```
Popov72, line 739: `const`?

Popov72 (commented):
LGTM!

src/Navigation/INavigationEngine.ts
CedricGuillemet: No ellipse. But, a PI/2 on X or Z axis gives a rectangle, so you can use a box for that case ;)

PR #9814, opened 2021-01-19
Fix for ArcRotateCamera's mouse wheel zoom issues

PolygonalSun:
This PR changes and refactors the mouse wheel code to fix some issues with how the ArcRotateCamera handles zooming with the mouse wheel.  It also simplifies and cleans up the code a bit.


deltakosh:
Pinging @MiiBond  as he is also using that feature

deltakosh (approved):

PR #9816, opened 2021-01-19
Add support for event.key values and handle blur events to DeviceInputSystem

PolygonalSun:
This PR contains two fixes:
- onInputChanged has added eventData parameter
Since there are several people who use the direct event data from the browser, the DeviceInputSystem will pass the direct event to the InputManager, as part of the onInputChanged call.  This should ensure that any event.key values are exactly what they are from the DOM event.  Additionally, the direct event data will also be accessible for the various pointer events as well.  Note: this DOM event data is only accessible via the onInputChanged callback.

- Keyboard values will be properly reset on "blur" event
When the canvas loses focus, the keyboard values in the DeviceInputSystem will reset to unpressed so that no unintended values will occur as users move between tabs.


src/Inputs/scene.inputManager.ts
```diff
@@ -823,7 +823,7 @@ export class InputManager {
             // Keyboard Events
             if (deviceType === DeviceType.Keyboard) {
                 evt.type = ((currentState === 1) ? "keydown" : "keyup");
-                evt.key = String.fromCharCode(inputIndex);
+                evt.key = eventData.key;
```
RaananW, line 826: Since we support IE you can fall back to String.fromCharCode in case .key is not defined

RaananW (commented):
If possible you can use the browser event interfaces instead of any (KeyboardEvent, PointerEvent etc). But as this is internal it's a nice to have

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -276,7 +282,10 @@ export class DeviceInputSystem implements IDisposable {
             if (kbKey) {
                 kbKey[evt.keyCode] = 1;
                 if (this.onInputChanged) {
-                    this.onInputChanged(DeviceType.Keyboard, 0, evt.keyCode, 0, kbKey[evt.keyCode]);
+                    const eventData: { [k: string]: any } = {};
```
Popov72, line 285: Doesn't this create a new object each time? Maybe the object could be created outside (at the file level, or once in the constructor) and reused here (and below)?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -276,7 +282,10 @@ export class DeviceInputSystem implements IDisposable {
             if (kbKey) {
                 kbKey[evt.keyCode] = 1;
                 if (this.onInputChanged) {
-                    this.onInputChanged(DeviceType.Keyboard, 0, evt.keyCode, 0, kbKey[evt.keyCode]);
+                    const eventData: { [k: string]: any } = {};
+                    eventData.key = evt.key;
+
+                    this.onInputChanged(DeviceType.Keyboard, 0, evt.keyCode, 0, kbKey[evt.keyCode], eventData);
```
Popov72, line 288: Couldn't it be interesting to know which event triggered the `onInputChanged` (by passing an additional data to the call that would let know the event was a key down/up/blur/...)? Just thinking out loud here.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -480,6 +541,7 @@ export class DeviceInputSystem implements IDisposable {
         this._elementToAttachTo.addEventListener(this._eventPrefix + "move", this._pointerMoveEvent);
         this._elementToAttachTo.addEventListener(this._eventPrefix + "down", this._pointerDownEvent);
         this._elementToAttachTo.addEventListener(this._eventPrefix + "up", this._pointerUpEvent);
+        this._elementToAttachTo.addEventListener("blur", this._pointerBlurEvent);
```
Popov72, line 544: I just have to ask that you meant to not add the `this._eventPrefix` to "blur" as you did for the other events, even if I'm sure it's done on purpose!

src/DeviceInput/deviceInputSystem.ts
RaananW: native (browser) event data is usually immutable (the object itself), i think that in this case it is fine to not reuse the object to prevent from people using the vent data incorrectly.

src/DeviceInput/deviceInputSystem.ts
RaananW: the blur event is not pointer/mouse event, bur an event on the element itself (like `focus`). so it is "standalone" :-)

src/Inputs/scene.inputManager.ts
PolygonalSun: I'll add code to cover for that scenario.  Great catch!

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: This sounds like it could be worth exploring.  The only thing that could would give me hesitation, it would be how that functionality would play with Babylon Native.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -291,13 +307,42 @@ export class DeviceInputSystem implements IDisposable {
...
+        };
+
+        this._onCanvasFocusObserver = this._engine.onCanvasFocusObservable.add(
+            (() => {
+                if (document.activeElement === this._elementToAttachTo) {
+                    attachedFunction();
+                }
+                return attachedFunction;
+            })()
+        );
```
bghgary, line 327: I'm really confused by this code. What is this trying to do?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -291,13 +307,42 @@ export class DeviceInputSystem implements IDisposable {
...
+
+            this._elementToAttachTo.removeEventListener("keydown", this._keyboardDownEvent);
+            this._elementToAttachTo.removeEventListener("keyup", this._keyboardUpEvent);
+
+            // Reset all keyboard inputs
+            if (this._inputs[DeviceType.Keyboard][0]) {
+                for(let i = 0; i < DeviceInputSystem._MAX_KEYCODES; i++) {
+                    this._inputs[DeviceType.Keyboard][0][i] = 0;
                 }
             }
```
bghgary, line 342: Does this need to run if this.elementToAttachTo if falsy?

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -291,13 +307,42 @@ export class DeviceInputSystem implements IDisposable {
...
+            if (!this._elementToAttachTo) {
+                return;
+            }
+
+            this._elementToAttachTo.removeEventListener("keydown", this._keyboardDownEvent);
+            this._elementToAttachTo.removeEventListener("keyup", this._keyboardUpEvent);
+
+            // Reset all keyboard inputs
+            if (this._inputs[DeviceType.Keyboard][0]) {
+                for(let i = 0; i < DeviceInputSystem._MAX_KEYCODES; i++) {
```
bghgary, line 339: nit

```suggestion
                for (let i = 0; i < DeviceInputSystem._MAX_KEYCODES; i++) {
```

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -411,6 +456,36 @@ export class DeviceInputSystem implements IDisposable {
...
+            // Handle mouse buttons
+            if (this.isDeviceAvailable(DeviceType.Mouse)) {
+                const pointer = this._inputs[DeviceType.Mouse][0];
+                for (let i = 0; i < DeviceInputSystem._MAX_POINTER_INPUTS; i++) {
+                    if (pointer[i + 2] === 1) {
+                        pointer[i + 2] = 0;
+
+                        if (this.onInputChanged) {
+                            this.onInputChanged(DeviceType.Mouse, 0, i + 2, 1, pointer[i + 2]);
+                        }
```
bghgary, line 469: I don't understand why we need to add fake events.

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -411,6 +456,36 @@ export class DeviceInputSystem implements IDisposable {
...
+            // Handle Active Touches
+            if (this.isDeviceAvailable(DeviceType.Touch)) {
+                const pointer = this._inputs[DeviceType.Touch];
+                for (let i = 0; i < pointer.length; i++) {
+                    if (pointer[i][PointerInput.LeftClick] === 1) {
+                        pointer[i][PointerInput.LeftClick] = 0;
+
+                        if (this.onInputChanged) {
+                            this.onInputChanged(DeviceType.Touch, i, PointerInput.LeftClick, 1, pointer[i][PointerInput.LeftClick]);
+                        }
```
bghgary, line 483: ditto

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Talked offline

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: No longer applicable

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Talked offline

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: No longer applicable

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -291,13 +295,26 @@ export class DeviceInputSystem implements IDisposable {
...
+                    if (kbKey[i] !== 0) {
+                        kbKey[i] = 0;
+                    }
                 }
             }
         });
 
-        window.addEventListener("keydown", this._keyboardDownEvent);
-        window.addEventListener("keyup", this._keyboardUpEvent);
+        this._elementToAttachTo.addEventListener("keydown", this._keyboardDownEvent);
```
RaananW, line 315: without defining `tabindex` on an html element, the window would be the only element to attach to. `window` of `document.body`, actually. But if you try to attach a keyboard event to a canvas (without setting its `tabindex`), events won't go through. Is that accounted for here?
It will work in the playground, as we are adding tabindex to the canvas.

This relates to keyboard events only. blur should be on the canvas element itself.

RaananW (commented):
Just a comment about the event to attach to, otherwise LGTM!

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Talked offline, adding line to add engine's tabIndex to elementToAttachTo if current one is -1

RaananW: Resolved with setting `.tabIndex`

RaananW (approved):

PR #9817, opened 2021-01-20
Inspector: Clone only the mesh, not the children when using render wireframe over mesh

Popov72:
See https://forum.babylonjs.com/t/render-wireframe-over-mesh-broken-with-parenting/17650


deltakosh (approved):

deltakosh:
Tests are failing :(

sebavan:
I am really not sure that the gltf loader should be touched as it currently works like other gltf viewers I am more thinking this could be an export issue.

@bghgary  ???

Popov72:
Rollbacked my bug for the baseColorFactor!

bghgary:
> I am really not sure that the gltf loader should be touched as it currently works like other gltf viewers I am more thinking this could be an export issue.

> Rollbacked my bug for the baseColorFactor!

Right. `baseColorFactor` in glTF is already in linear space.

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#pbrmetallicroughnessbasecolorfactor

> These values are linear.

deltakosh (approved):

PR #9830, opened 2021-01-22
exposed collision retry count

CedricGuillemet:
follow up on https://forum.babylonjs.com/t/internal-collision-system-acting-weird-after-calling-mesh-dispose/17700 and maybe https://forum.babylonjs.com/t/collision-problem/17751/2

by default, stops collision detection after 3 retries which might not be enough in some cases. I exposed the value so it can be tweaked .


msDestiny14:
Looks good to me! :)

msDestiny14 (approved):

Popov72 (approved):
LGTM!

PR #9831, opened 2021-01-22
Fix spatial sound played reversed in right handed mode

### LINKED ISSUE #8836: Spatialization of sound is inverted when useRightHandedSystem is true
Try this PG:

https://playground.babylonjs.com/#7X0ELW

It's working as expected, the sound comes from the left when the ufo is on the left and from the right when it is on the right.

Now change `scene.useRightHandedSystem` to `true`: the sound is inverted.

Popov72:
See https://forum.babylonjs.com/t/spatial-audio-is-reversed-with-userighthandedsystem/17493

This PR reverts back  what was done to fix #8836 as the existing code at that time was ok. The right/left handedness of the scene is taken care in the camera view matrix, so there's no need to use a different camera direction vector when in right handed system.

The problem is in fact with the way the MSFT_audio_emitter extension deals with sounds. It sets `Sound._positionInEmitterSpace = true`, which leads to transforming the origin of the 3D world to the local space of the sound emitter, which does not sound right to me as all sound positions you should be expressed in the same world coordinate for audio engine to work as expected (also, why transforming the origin of the 3D space?).

The code is (first 3 lines):
```typescript
if (this._positionInEmitterSpace) {
    node.worldMatrixFromCache.invertToRef(TmpVectors.Matrix[0]);
    this.setPosition(TmpVectors.Matrix[0].getTranslation());
} else {
    if (!(<any>node).getBoundingInfo) {
        this.setPosition(node.absolutePosition);
    } else {
        let mesh = node as AbstractMesh;
        let boundingInfo = mesh.getBoundingInfo();
        this.setPosition(boundingInfo.boundingSphere.centerWorld);
    }
}
if (Engine.audioEngine.canUseWebAudio && this._isDirectional && this.isPlaying) {
    this._updateDirection();
}
```

Waiting for feedback about this before removing the "Draft" flag.

@najadojo If you are around, could you shed some light on this as you did integrate the MSFT_audio_emitter extension and wrote the code related to `_positionInEmitterSpace`? Thanks in advance!

### DIFF
--- a/loaders/src/glTF/2.0/Extensions/MSFT_audio_emitter.ts
+++ b/loaders/src/glTF/2.0/Extensions/MSFT_audio_emitter.ts
@@ -194,7 +194,6 @@ export class MSFT_audio_emitter implements IGLTFLoaderExtension {
                     sound.maxDistance = emitter.maxDistance || 256;
                     sound.rolloffFactor = emitter.rolloffFactor || 1;
                     sound.distanceModel = emitter.distanceModel || 'exponential';
-                    sound._positionInEmitterSpace = true;
                 }));
             }
 
--- a/src/Audio/audioSceneComponent.ts
+++ b/src/Audio/audioSceneComponent.ts
@@ -249,8 +249,7 @@ Object.defineProperty(Scene.prototype, "audioPositioningRefreshRate", {
  * in a given scene.
  */
 export class AudioSceneComponent implements ISceneSerializableComponent {
-    private static _CameraDirectionLH = new Vector3(0, 0, -1);
-    private static _CameraDirectionRH = new Vector3(0, 0, 1);
+    private static _CameraDirection = new Vector3(0, 0, -1);
 
     /**
      * The component name helpful to identify the component in the list of scene components.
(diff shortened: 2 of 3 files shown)


najadojo:
Sorry I'm not recalling the reason's for the flag setting. The spec for the extension is at [MSFT_audio_emitter].(https://github.com/KhronosGroup/glTF/blob/8109e86a26c9210814f8d236589ce92a2f9fa8f8/extensions/2.0/Vendor/MSFT_audio_emitter/README.md)

Popov72:
Thanks for answering.

I don't see anything in the spec that would explain the code (but I'm no sound expert).

As I understand it, what you perceive of a sound only depends on your position (so the camera position) and the sound source position in world space (+direction of viewing + cone if sound is spatial, but it does not matter for the discussion). So, you only need to transform the sound source position (expressed in its local space) to world space by multiplying this position by the world matrix of the node to which the sound is attached to. That's what the code is doing when the MSFT extension is not used.

Maybe @bghgary you have an idea about that (I saw you were in the committee for the MSFT extension)?

bghgary:
As far as I know, the position of the audio node is no different than any other nodes in the scene graph. Thus, there should not be a special flag on the sound object to transform it in any way different than normal.

We also discussed this offline and @Popov72 will remove the odd flag that was added to Sound.

PR #9834, opened 2021-01-22
Add code to DeviceInputSystem to handle Pointer Blur events

PolygonalSun:
During the testing for the InputManager conversion to the DeviceInputSystem, I found a bug that's been around since at least v3.3 in the InputManager where pointer lock will remain if you lose focus on the canvas while any pointer buttons are pressed.


RaananW (approved):
LGTM!

My only comment would be - blur has nothing to do with pointer events, but the focus in the browser itself (which might be triggered by a pointer click outside of it).

PR #9835, opened 2021-01-22
Fix envTextureLoader leaking an exception that can not be caught by the consumer

CoPrez:
I have been doing some stress testing of Babylon.js by loading a bunch of instances of a basic scene all in the same web page. Eventually the webgl context is lost and Babylon starts throwing exceptions (that's expected). I however found that one of the exceptions I was seeing could not be caught because it throwing an exception internally rather than catching it and notifying the onError callback.

In the scenario I encountered this issue was where the consumer calls the [constructor for a CubeTexture](https://github.com/BabylonJS/Babylon.js/blob/39596ca3cfcffa44f299c235540e153b4ab2280c/src/Materials/Textures/cubeTexture.ts#L168) which takes in a URI and an onError callback (as well as other parameters). From there it goes through a bunch of internal layers and eventually to create the texture because the webgl context has been lost. Specifically it fails at [this line in thinEngine](https://github.com/BabylonJS/Babylon.js/blob/382452035ada5b3e34d856d0be1d2f9eed301177/src/Engines/thinEngine.ts#L3044). As the exception bubbles up through the stack it reaches envTextureLoader and synchronously fails in the call to UploadEnvLevelsAsync. The exception is not caught there and continues to bubble up. Eventually it gets to the layer where the file operation callback was called ([here](https://github.com/BabylonJS/Babylon.js/blob/b8238023e50655c58361136a88fcd7e88f06772d/src/Misc/fileTools.ts#L400)). The only layer above that is the XMLHttpRequest which means the exception becomes unhandled.

This change fixes the exception pipeline by ensuring envTextureLoader catches any synchronous or asynchronous errors and pipes them to the onError callback rather than letting them bubble up.


src/Materials/Textures/Loaders/envTextureLoader.ts
```diff
@@ -42,15 +42,21 @@ export class _ENVTextureLoader implements IInternalTextureLoader {
...
+            try {
+                EnvironmentTextureTools.UploadEnvSpherical(texture, info);
+                EnvironmentTextureTools.UploadEnvLevelsAsync(texture, data, info).then(() => {
+                    texture.isReady = true;
+                    texture.onLoadedObservable.notifyObservers(texture);
+                    texture.onLoadedObservable.clear();
+                    if (onLoad) {
+                        onLoad();
+                    }
+                }).catch((reason) => {
```
bghgary, line 54: nit: it's a tiny bit cheaper to do this

```suggestion
                }, (reason) => {
```

bghgary (approved):
Looks fine to me. @sebavan Can you also take a look?

deltakosh (approved):

src/Materials/Textures/Loaders/envTextureLoader.ts
CoPrez: Looking at it a bit more I actually think the `.catch()` approach here is what we want. It does add an extra layer of promise resolution but with that approach it will also catch if the .then() throws an error inside of the onSuccess callback. Which in this case is quite plausible since it is triggering the onLoadedObservable callbacks.

Happy to change it though if you think it should still use the passed in onReject rather than the catch chaining.

bghgary: Interesting. I think the code probably shouldn't call both onLoad and onError. With the way you had it, if onLoad threw an exception, then it would call onError. The way I suggested doesn't quite work either since it won't trap the error coming [lines 48-50](from https://github.com/BabylonJS/Babylon.js/pull/9835/files/cdd5d02666605eb4a681d60042d1cd8aeca1d947#diff-e7a1c1509ff5bf5a669adbcb9718d819b466e40b0916e7bfa9e8b5b95810a70aR48-R50).

We probably need something like this:
```javascript
                EnvironmentTextureTools.UploadEnvLevelsAsync(texture, data, info).then(() => {
                    texture.isReady = true;
                    texture.onLoadedObservable.notifyObservers(texture);
                    texture.onLoadedObservable.clear();
                }.then(() =>{
                    if (onLoad) {
                        onLoad();
                    }
                }, reason) => {
                    onError?.("Can not upload environment levels", reason);
                });
```

bghgary: Wait, this isn't right either. We shouldn't throw for the onLoadedObservable either. So what I suggested is right I think.

ryantrem: One thing that is a bit weird currently is if the `onLoad` callback throws, then the `onError` callback is also called. It might make sense to do something more like:
```ts
EnvironmentTextureTools.UploadEnvLevelsAsync(texture, data, info).then(() => {
    texture.isReady = true;
    texture.onLoadedObservable.notifyObservers(texture);
    texture.onLoadedObservable.clear();
}).then(onLoad, (reason) => onError?.("Can not upload environment levels", reason));
```

bghgary: That is more or less my suggestion although originally for the wrong reason.

CoPrez: Going back to the original suggestion so that it's up to the onLoad callbacks to handle any errors they might generate themselves.

bghgary (approved):

PR #9838, opened 2021-01-23
Fix a bug in scene loader and add some unit tests

### LINKED ISSUE #9837: Root path should not combine with data URI
Following #9430, some changes were made to the way glTFs can be loaded from data URIs.  In particular, these lines were added:

https://github.com/BabylonJS/Babylon.js/blob/cfde2ef094705213bf95547b5d1a2ef0adae79ec/src/Loading/sceneLoader.ts#L603-L606

It looks like a root URL is being combined with a data URI here, and I think that's the cause of a bug I'm seeing.

In the VSCode glTF extension, a string-based glTF can be supplied along with a non-blank root path.  This is because the glTF might reference binary data and textures by relative path.  So the glTF itself needs to be read with JSON.parse, without any root path on the front, but all relative resources must be combined with the root path.

As a result of this change, the full URL get converted to something along the lines of:

```
'schema://root/path/' + 'data:{"asset":{"version":2}, ...'
```

This results in a `404` showing up in the network tab of the debugger in VSCode, with the address being the root path followed by the entire contents of the glTF JSON document.

You can also reproduce this bug by editing the [playground from #9430](https://playground.babylonjs.com/#PTLWQQ#2). On line 32, change the second argument from an empty string `''` to a root path such as `'https://playground.babylonjs.com/'`.  The root path should be ignored for data URIs, and only used for relative assets referenced from inside the data URI.

(As an aside, `Tools.IsBase64` is a misnomer here.  The data in question is `'data:' + JSON.stringify(gltfAsObject)`, which is plain text but is not base64-encoded in any way.  But that's not the cause of the bug here, it's just an inaccurate choice of function name for what it's testing.  Perhaps it should be `IsDataUri` or similar).

/cc @bghgary

bghgary:
Fixes #9837


emackey (approved):
Confirmed, this fixes the issue for me.

Also the new unit test "Direct load" appears representative of how the VSCode preview window loads glTFs into Babylon.  It will be great to have that protected by the test suite.

Thanks @bghgary!

RaananW (approved):

PR #9841, opened 2021-01-25
Added check for PointerDragBehavior to reenable camera controls, if not reenabled before behavior is disabled

PolygonalSun:
This PR adds an additional check to the PointerDragBehavior's pointer observable where it will run the releaseDrag function if the behavior was disabled before this function got the chance to run.  This prevents the issue of the camera controls being disabled in a situation where dragging could be set as a toggleable function.


dist/preview release/what's new.md
```diff
@@ -120,6 +120,7 @@
 - Fix crash when cloning material in `AssetContainer.instantiateModelsToScene` when mesh is an instanced mesh ([Popov72](https://github.com/Popov72))
 - Fix issue with NinePatch displaying half pixel gaps between slices on Firefox browsers. ([Pryme8](https://github.com/Pryme8))
 - Fix issue when canvas loses focus while holding a pointer button ([PolygonalSun](https://github.com/PolygonalSun))
+- Fix issue where camera controls stay detached if PointerDragBehavior is disabled prematurely ([PolygonalSun](https://github.com/PolygonalSun))
```
bghgary, line 123: Is this a pre-existing bug before your input changes? If not, then you don't want a what's new entry.

PR #9853, opened 2021-01-27
Handle changes in XRWebGLLayer framebuffer size

kainino0x:
XRWebGLLayer's framebuffer size (and framebuffer object) are allowed to
change between frames. WebXRSessionManager's IRenderTargetProvider would
create a RenderTargetTexture just once, so it would miss changes coming
from WebXR.

Obviously I have no idea if this is the right fix, but it works :)
Guidance would be appreciated.

I found this bug while working on a personal project:
https://github.com/kainino0x/holoplay-webxr
It is possible to verify this fix by loading the Babylon playground with this browser extension installed, entering XR, then changing the "number of views to render".

### DIFF
--- a/src/XR/webXRSessionManager.ts
+++ b/src/XR/webXRSessionManager.ts
@@ -238,8 +238,21 @@ export class WebXRSessionManager implements IDisposable {
             this._rttProvider = this._xrNavigator.xr.getNativeRenderTargetProvider(this.session, this._createRenderTargetTexture.bind(this));
         } else {
             // Create render target texture from xr's webgl render target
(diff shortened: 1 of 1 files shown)


RaananW (approved):
LGTM!

Thanks for submitting the fix.

Wanted to ask about the extension you mentioned - I know the WebXR emulator and have been using it for some time now, but I don't remember seeing a way to change the number of views there.

kainino0x:
> Wanted to ask about the extension you mentioned - I know the WebXR emulator and have been using it for some time now, but I don't remember seeing a way to change the number of views there.

Indeed the WebXR emulator doesn't have a way to change the number of views (beyond 1-2) or the framebuffer size. That control is specific to my project which outputs to a Looking Glass display, which can display an arbitrary number of different angles in a 30 degree "cone" around the display

Issue #9751, opened 2021-01-06
[XR] Move to the new hand tracking API


RaananW:
As part of the changes there is now a way to get all poses for all 25 joints in one go (instead of querying per joint):

https://immersive-web.github.io/webxr-hand-input/#xrframe-interface

cabanier:
Also, the WebXR Hands API was updated so you need to update your library.

RaananW:
Yep, working on that 😊
Thanks!

RaananW:
This was implemented already

PR #9855, opened 2021-01-27
initial Hand-API update

RaananW:
Partly addressing #9751,
Update to hand joint tracking to support the new changes in the API

### DIFF
--- a/src/LibDeclarations/webxr.d.ts
+++ b/src/LibDeclarations/webxr.d.ts
@@ -315,6 +315,9 @@ interface XRJointPose extends XRPose {
     radius: number | undefined;
 }
 
+// to be extended
+type XRHandJoint = string;
+
 interface XRHand extends Iterable<XRJointSpace> {
     readonly size: number;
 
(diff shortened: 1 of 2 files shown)


src/LibDeclarations/webxr.d.ts
```diff
@@ -315,11 +315,16 @@ interface XRJointPose extends XRPose {
     radius: number | undefined;
 }
 
+// to be extended
+type XRHandJoint = string;
+
 interface XRHand extends Iterable<XRJointSpace> {
```
cabanier, line 321: XRHand is not an array anymore; instead it's a map: `iterable<XRHandJoint, XRJointSpace>`

cabanier (commented):
This looks good!
Maybe you can also have a parallel implementation that uses the high performance API. I've heard anecdotal evidence that it uses far less CPU resources

src/LibDeclarations/webxr.d.ts
RaananW: This PR is to first get the hands working again.
I am already implementing a new version using the `fillXXX` functions to create less objects per frame. 
XRHandJoint is technically not a string but a collection  of predefined strings. To avoid back-compat issues i didn't change it yet, but this is coming together with the new PR as well.

deltakosh (approved):

PR #9856, opened 2021-01-27
Scene Manager: V5.0.0-A70

MackeyK24:
Babylon Toolkit Alpha 7 Update


RaananW:
Hey @MackeyK24 - I totally understand the need for the declaration update, but - can the .js file be served somewhere else?

MackeyK24:
It was an issue serving that file elsewhere and the fix @deltakosh mare to run toolkit in playground. So delta recommended the spot it’s in now.

what the problem, too many updates... I was just trying to keep playground up to date with latest builds. But I can update the playground less... will be behind the current toolkit in that case

RaananW:
It's not about updating less, we want it updated! There are a few implications regarding commiting a minified file that I am a bit concerned about. We can take that offline :-)

PR #9865, opened 2021-01-29
KHR_materials_transmission: fixes and improvements

Popov72:
* the refraction texture was rendered two times. Setting the `.refractionTexture` of a PBR material is enough to have the RTT generated, so I have removed the push in `scene.customRenderTargets`
* the observer was not removed from `mesh.onMaterialChangedObservable` when a mesh is disposed
* I have removed the layer handling. I don't think we want to have all layers automatically drawn in the refraction texture RTT: it should be a user decision. For eg, if we have a GUI, we get:
image
That's not what we want, the button should not be part of the refraction texture. I think it's better to let the user decide if they want to add the refraction texture to the list of RTTs of any layer they can have in their scene => they can access the refraction texture through `scene._transmissionHelper.getOpaqueTarget()`
* made some options configurable (samples, lodGenerationScale and lodGenerationOffset). Those values can be changed by accessing directly the RTT texture, but I think adding theses in the option list make it clear they can/must be changed (and we get their default values in the docs). Notably, lodGenerationScale and lodGenerationOffset will depend on your scene and the size of the opaque objects.

FYI @MiiBond

### DIFF
--- a/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
+++ b/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
@@ -12,6 +12,7 @@ import { Mesh } from "babylonjs/Meshes/mesh";
 import { Texture } from "babylonjs/Materials/Textures/texture";
 import { RenderTargetTexture } from "babylonjs/Materials/Textures/renderTargetTexture";
 import { Observable, Observer } from "babylonjs/Misc/observable";
+import { Constants } from "babylonjs/Engines/constants";
 
 interface ITransmissionHelperHolder {
     /**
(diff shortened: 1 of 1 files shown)


bghgary (approved):

Popov72:
Failing test in webgl 1 is "Displacement map" which has nothing to do with this PR, so probably a transient error.

MiiBond:
@Popov72 Very nice!
If you have multiple transmissive materials, will setting .refractionTexture on each cause multiple renders per frame? I think that's what I was afraid of when I initially wrote this.

Popov72:
No, the texture will be rendered a single time even if used by multiple materials.

Popov72:
That's ok for me and for Gary. I guess we can merge.

PR #9867, opened 2021-01-29
Fix triplanar material when mesh position is not at the origin

Popov72:
See https://forum.babylonjs.com/t/triplanar-shader-strange-behavior-when-meshs-are-translated-possible-bug/17985


Pryme8:
Not sure why setting the forth component of the quat to 0  would fix it? I have not had to do that before with triplanar shaders. Can you please explain how this works, not saying its wrong just want to understand!

deltakosh:
because tangent or normals are directions and not positions so you do not want to apply a translation transform to them

Pryme8:
Got it!  that actually makes a lot of sense as soon as you say it.

julien-moreau:
Thanks @Popov72 You rox !!

PR #9872, opened 2021-01-31
Fix Mesh.increaseVertices not working with updatable buffers

Popov72:
See https://forum.babylonjs.com/t/updatable-mesh-and-increasevertices/18037/2

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -2807,9 +2807,9 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
     public increaseVertices(numberPerEdge: number): void {
         var vertex_data = VertexData.ExtractFromMesh(this);
         var uvs = vertex_data.uvs;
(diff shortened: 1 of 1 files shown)


RaananW (commented):
Will that not fail on IE11? I remember adding Array.from once and the one issue it had was browser support. Is there a different way of doing this?

Popov72:
Ah, IE11...

I don't know if we should put [the polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#polyfill) in the `ArrayTools` class?

RaananW:
So much code to support the best browser in the world...

If there is no other way, maybe we should I just think - we increase the codebase just for a single browser that is probably used in less than 1% of the experiences. My suggestion would be - to keep the old implementation for IE11 only. Meaning - check if Array.from exists, and if not - simply continue without it, as it was until now. This way we support this function in other browsers, while slowly removing advance-functionality support for IE11. 
@deltakosh - is that acceptable?

deltakosh:
I agree with @RaananW idea. We will not increase our size for ie11 anymore.

Popov72:
Ok, done!

deltakosh (approved):

PR #9873, opened 2021-02-01
Optimize collider

ottoville:
This will add some optimization to collider, the length of velocity is no longer calculated when not needed and some formulas has been modified to minimize instructions.


src/Collisions/collider.ts
```diff
@@ -138,7 +140,14 @@ export class Collider {
     /** @hidden */
     public _initialize(source: Vector3, dir: Vector3, e: number): void {
         this._velocity = dir;
-        Vector3.NormalizeToRef(dir, this._normalizedVelocity);
+        this._velocitySquaredLength = this._velocity.lengthSquared();
+        const len = Math.sqrt(this._velocitySquaredLength)
+        if (len === 0 || len === 1.0) {
+            this._normalizedVelocity.copyFromFloats(dir._x, dir._y, dir._z);
```
RaananW, line 146: could use copyFrom instead of using the 3 floats

RaananW (commented):
Just for me to understand - this entire change is actually to avoid calculating the distanceSquared twice, right?

src/Collisions/collider.ts
```diff
@@ -347,8 +355,8 @@ export class Collider {
             edgeDotVelocity = Vector3.Dot(this._edge, this._velocity);
             edgeDotBaseToVertex = Vector3.Dot(this._edge, this._baseToVertex);
 
-            a = edgeSquaredLength * (-velocitySquaredLength) + edgeDotVelocity * edgeDotVelocity;
-            b = edgeSquaredLength * (2.0 * Vector3.Dot(this._velocity, this._baseToVertex)) - 2.0 * edgeDotVelocity * edgeDotBaseToVertex;
+            a = edgeSquaredLength * (-this._velocitySquaredLength) + edgeDotVelocity * edgeDotVelocity;
+            b = 2 * (edgeSquaredLength * Vector3.Dot(this._velocity, this._baseToVertex) - edgeDotVelocity * edgeDotBaseToVertex)
```
RaananW, line 359: (Linting failed, probably the missing semicolon here)

src/Collisions/collider.ts
ottoville: Yes, sure

Popov72: `copyFrom` call `copyFromFloats`, so directly using `copyFromFloats` saves a call ;)

However, we should probably change the implementation of `Vector3.copyFrom` to directly store the values instead of calling `Vector3.copyFromFloats`, as it is what `Vector2.copyFrom`, `Vector4.copyFrom` and `Quaternion.copyFrom` are doing...

RaananW: Yes, that would be wonderful. The reason I commented is that when I see `copyFrom` it is obvious what type this input object is, but copy from float might mean that even if it IS a vector something is manipulated between input and output. 
Internally it does the same, of course. We can change it - @Popov72 want to submit this?

RaananW:
Thanks a lot! Linting still fails. Try formatting the file using prettier (we have a configuration file in place), it should eliminate all linting issues.

deltakosh:
you can also run "npm run build" and it will clean your linting

CedricGuillemet:
This PR is breaking some PG. See this thread : https://forum.babylonjs.com/t/jumping-camera-movement/18352/8
@ottoville Can you please take a look ?

ottoville:
It does not seems to be problem with arcrotatecamera: https://playground.babylonjs.com/#SRZRWV#437
Its somewhat releated to gravity?

CedricGuillemet:
Yes, it appears the most once gravity is enabled. The camera jumps up and down each frame. Like collision is detected the frame after and then it's reapplied.

ottoville:
I dont see how this PR could have broke it. Its just simple math formulas re-arranged. Can someone have a second eye to see if there is some mistake I cannot see.
The algorithm for collision detection itself is very much depated and there are plenty of known precision errors. Could it be that this PR changed the precision to be too much accurate to gravity collision response for handle.

CedricGuillemet:
No worries, I'll do a deeper check and let you know if I find something :)

PR #9877, opened 2021-02-01
Add HoloLens 2 shell fluent button material

rgerd:
This PR adds the same fluent design shaders that we use in the HoloLens 2 shell, which are exported from a cross-engine tool we use to provide identical design functionality across Unity, Unreal,  our internal holographic engine, and now Babylon.js.

ezgif-6-9ae7eb121c9f


gui/src/3D/materials/fluentButton/fluentButtonBuilder.ts
```diff
@@ -0,0 +1,45 @@
+import { Scene } from "babylonjs/scene";
+import { Mesh } from "babylonjs/Meshes/mesh";
+import { VertexData } from "babylonjs/Meshes/mesh.vertexData";
+import { Nullable } from "babylonjs/types";
+
+function CreateFluentButtonVertexData(): VertexData {
```
RaananW, line 6: Is it possible to have a mesh for this instead? Load it async the first time it is loaded? I am just thinking about the package size for non-VR scenes that don't require this. 
Unless it is used any place other than creating the mesh in the builder

gui/src/3D/materials/fluentButton/fluentButtonMaterial.ts
```diff
@@ -0,0 +1,494 @@
...
+    constructor() {
+        super();
+        this.rebuild();
+    }
+}
+
+/**
+ * Class used to render square buttons with fluent desgin
+ */
+export class FluentButtonMaterial extends PushMaterial {
```
RaananW, line 41: This file doesn't follow our naming convention - https://doc.babylonjs.com/divingDeeper/developWithBjs/approvedNamingConventions

RaananW (commented):
The only change I ask is the naming convention, especially in `fluentButtonMaterial.ts`. (https://doc.babylonjs.com/divingDeeper/developWithBjs/approvedNamingConventions).

Also, and this is just an open conversation and not a change request, will it be possible to offload the binary information (image data and vertex data) and host it somewhere else? If a develop requires it as part of the package they will be able to include it themselves, otherwise it can be loaded async. What do you think?

deltakosh:
Any update?

rgerd:
> Any update?

The shader is functionally ready, we were focusing on the input manager updates required to support the shader properly (beyond the hacked demo). Since those changes are taking longer than expected I will work on addressing Raanan's comments this week. Thank you for your patience!

RaananW:
If you need any place for the digital assets - we have an assets repository that is deployed to a CDN.

rgerd:
Almost there-- currently blocked as I'm trying to figure out how to get all of the vertex attributes defined in the `fluentButtonBuilder` to preserve across gltf export / import.

RaananW (approved):

deltakosh (approved):

PR #9885, opened 2021-02-03
homogeneous depth range Native

CedricGuillemet:
Viewport depth range is dependant on rendering API.
[-1..1] for webgl and opengl
[0..1] for webgl and d3d

For babylon.js, the fix is done in the shaderprocessor depending on API (webgl vs webgpu).
It has never been fixed for Native, assuming always [-1..1] even on d3d. near plane clipping was happening too soon but no one noticed.

This looks like the inverted UV with the exception we can't use the same trick on native with flipped projection matrix. 
Because all the computations (like shadows) are done assuming a depth range of [-1..1] and the depth range remapping is done at the shader end. 

So, here, the same trick is done as for webgpu with a shader processor.

### DIFF
--- a/src/Engines/Native/nativeShaderProcessors.ts
+++ b/src/Engines/Native/nativeShaderProcessors.ts
@@ -1,0 +1,24 @@
+import { Nullable } from '../../types';
+import { WebGL2ShaderProcessor } from '../WebGL/webGL2ShaderProcessors';
+import { ShaderProcessingContext } from '../Processors/shaderProcessingOptions';
+
+declare type ThinEngine = import("../thinEngine").ThinEngine;
+declare type NativeEngine = import("../nativeEngine").NativeEngine;
+
+/** @hidden */
+export class NativeShaderProcessor extends WebGL2ShaderProcessor {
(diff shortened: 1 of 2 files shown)


src/Engines/Native/nativeShaderProcessors.ts
```diff
@@ -0,0 +1,24 @@
...
+declare type NativeEngine = import("../nativeEngine").NativeEngine;
+
+/** @hidden */
+export class NativeShaderProcessor extends WebGL2ShaderProcessor {
+    public postProcessor(code: string, defines: string[], isFragment: boolean, processingContext: Nullable<ShaderProcessingContext>, engine: ThinEngine) {
+        code = super.postProcessor(code, defines, isFragment, processingContext, engine);
+
+        // Depending on API, depth range is [-1..1] or [0..1]
+        // This is defined by homogeneousDepth in bgfx caps
+        // Flip Y + convert z range from [-1,1] to [0,1]
```
Popov72, line 15: I think "Flip Y" should be removed in the comment, as the code is doing only the z range remapping.

Popov72 (approved):
LGTM!

PR #9887, opened 2021-02-04
recast friendly module definition

CedricGuillemet:
followup on https://forum.babylonjs.com/t/using-recast-detour-package-server-side-with-null-engine-and-typescript-definitions/18068/9
More friendly recast module def
I copy/pasted your code @RaananW . Is there something else to do?


RaananW (approved):
There is a catch here - this will work for modules, but won't work if Recast is populated in the global namespace. I am not sure what the case here is, but I assume that this package was meant to be used as a module.

Issue #9293, opened 2020-10-28
Implement Quaternion.LookRotation


Alex-MSFT:
Unity provides a nice helper function [Quaternion.LookRotation](https://docs.unity3d.com/ScriptReference/Quaternion.LookRotation.html) to help create a look rotation quaternion vector for a given direction and normal.  This helps with operations such as orienting objects to face a specific direction along a fixed plane.

Babylon.js has a similar mesh.lookAt function but this has two limitations:
1. You can't lock the normal for the look direction
2. You can't perform this operation with just two vectors, as it requires operating against a mesh.

In the past I have implemented it using this algorithm: https://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToQuaternion/

deltakosh:
Hello any news?

Alex-MSFT:
Thanks for reminding me I have this issue assigned to me!  It got a bit lost on my personal backlog as I have been working on some other priorities.  I will try to prepare a PR for this soon.

deltakosh:
Thanks a lot!

PR #9890, opened 2021-02-04
Implement Quaternion.FromLookDirectionLH/RH and Matrix LookDirectionLH/RH

Alex-MSFT:
This PR adds new helper functions for Quaternion and Matrices which allow creation of a rotation specifying just a direction and up vector as discussed in #9293.

These utilities are similar to LookAtLH/RH, however differ in that they do not take into account the eye orientation, and are based only on two orthogonal direction vectors for up and forward.  This will generate a rotation in an upright position that is facing in the same direction as the forward direction vector.

Here are some examples:
LookRotation_1

LookRotation_2

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -3597,6 +3597,7 @@ export class Quaternion {
      * This function works in left handed mode
      * @param forward defines the forward direction - Must be normalized and orthogonal to up.
      * @param up defines the up vector for the entity - Must be normalized and orthogonal to forward.
+     * @param ref defines the target quaternion.
      */
     public static FromLookDirectionLHToRef(forward: DeepImmutable<Vector3>, up: DeepImmutable<Vector3>, ref: Quaternion): void {
         var rotMat = MathTmp.Matrix[0];
@@ -3621,7 +3622,7 @@ export class Quaternion {
      * This function works in right handed mode
      * @param forward defines the forward direction - Must be normalized and orthogonal to up.
      * @param up defines the up vector for the entity - Must be normalized and orthogonal to forward.
-     * @returns A new quaternion oriented toward the specified forward and up.
+     * @param ref defines the target quaternion.
      */
     public static FromLookDirectionRHToRef(forward: DeepImmutable<Vector3>, up: DeepImmutable<Vector3>, ref: Quaternion): void {
         var rotMat = MathTmp.Matrix[0];
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -3583,7 +3583,7 @@ export class Quaternion {
      * Creates a new rotation value to orient an object to look towards the given forward direction, the up direction being oriented like "up".
      * This function works in left handed mode
      * @param forward Defines the forward direction of the desired quaternion.
-     * @param up  Defines the up direction of the desired quaternion.
+     * @param up Defines the up direction of the desired quaternion.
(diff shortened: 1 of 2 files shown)


src/Maths/math.vector.ts
```diff
@@ -5134,6 +5156,97 @@ export class Matrix {
...
+    /**
+     * Sets the given "result" Matrix to a rotation matrix used to rotate an entity so that it looks in the direction of forward, the up direction being oriented like "up".
+     * This function works in left handed mode
+     * @param forward defines the forward direction
+     * @param up defines the up vector for the entity
+     * @param result defines the target matrix
+     */
+    public static LookDirectionLHToRef(forward: DeepImmutable<Vector3>, up: DeepImmutable<Vector3>, result: Matrix): void {
+        // First ensure that two vectors are valid and approximately orthogonal.
+        if (forward.length() < Epsilon || up.length() < Epsilon || Math.abs(Vector3.Dot(forward, up)) >= Epsilon) {
```
Popov72, line 5181: I don't think we should have these checks, as as a general rule Babylon does not validate the input parameters passed to a function (for performance reasons): it's up to the caller to make sure that the parameters are valid.

Simply updating the doc to indicate that both vectors must be orthogonal (and probably of unit length? that would save normalizing in the function itself as most of the time the user will already have unit vectors) should be enough.

src/Maths/math.vector.ts
```diff
@@ -5134,6 +5156,97 @@ export class Matrix {
...
+        }
+
+        // Next normalize the two inputted vectors and take the cross product to find the third orthonormal vector that defines the rotation.
+        const forwardNorm = MathTmp.Vector3[0];
+        const upNorm = MathTmp.Vector3[1];
+        const right = MathTmp.Vector3[2];
+        forward.normalizeToRef(forwardNorm);
+        up.normalizeToRef(upNorm);
+        Vector3.CrossToRef(upNorm, forwardNorm, right);
+        right.normalize();
```
Popov72, line 5238: You can save this normalization: as up and forward are orthogonal and of unit length, so their cross product is also of unit length.

src/Maths/math.vector.ts
Alex-MSFT: Done!

src/Maths/math.vector.ts
Alex-MSFT: Good catch, done.

src/Maths/math.vector.ts
```diff
@@ -3579,6 +3579,28 @@ export class Quaternion {
...
     }
 
+    /**
+     * Creates a new rotation value to orient an object to look towards the given forward direction, the up direction being oriented like "up".
+     * This function works in left handed mode
+     * @param forward defines the forward direction - Must be normalized and orthogonal to up.
+     * @param up defines the up vector for the entity - Must be normalized and orthogonal to forward.
+     * @returns A new quaternion oriented toward the specified forward and up.
+     */
+    public static FromLookDirectionLH(forward: DeepImmutable<Vector3>, up: DeepImmutable<Vector3>): Quaternion {
```
RaananW, line 3589: Would be great to also have toRef functions for those two, and use a temp Matrix instead of creating a new one. You have TmpVector available for you (in the same file), you could use TmpVector[7]

Alex-MSFT: Done

deltakosh:
LGTM. Just a doc issue and we are good!

PR #9892, opened 2021-02-05
Add config object to nativeEngine constructor, Add device scaling to NativeEngine constructor.

Drigax:


Drigax:
This one is an interesting change, I'm currently torn between matching the Engine constructor definition, and using our own, thoughts?

bghgary:
@sebavan @Popov72 What are we doing for WebGPU?

Popov72:
We are using a custom options object:
```typescript
    public constructor(canvas: HTMLCanvasElement, options: WebGPUEngineOptions = {}) {
        super(null);
        ....
```
https://github.com/BabylonJS/Babylon.js/blob/master/src/Engines/webgpuEngine.ts#L77-L143

But I can see we are not dealing with the hardware scaling level, we are setting it to 1...

I guess it's something to add on my list.

deltakosh:
up?

sebavan:
Any progress on this one ? It now has merge conflicts

Drigax:
It doesn't seem like anyone else has any strong opinions about this change, so I'll fix the conflicts and merge

Drigax:
@bghgary, I updated to follow the pattern we're using in the other engine configuration objects, the only thing I'm unsure of is that we have an [EngineOptions class defined](https://doc.babylonjs.com/typedoc/interfaces/babylon.engineoptions) for the thinEngine, but I'm not sure if we should derive our configuration object from this, since I'm unsure the extent to which we support any of these options

PR #9895, opened 2021-02-06
Updating near interaction for new api

rickfromwork:
Correcting the outdated reference to hand joints caused by the update to the new webXR hand API.

### DIFF
--- a/src/XR/features/WebXRHandTracking.ts
+++ b/src/XR/features/WebXRHandTracking.ts
@@ -241,7 +241,7 @@ export class WebXRHand implements IDisposable {
         /** the controller to which the hand correlates */
         public readonly xrController: WebXRInputSource,
         /** the meshes to be used to track the hand joints */
-        public readonly trackedMeshes: AbstractMesh[],
+        public readonly trackedMeshes: Map<string, AbstractMesh>,
         private _handMesh?: AbstractMesh,
         private _rigMapping?: string[],
         disableDefaultHandMesh?: boolean,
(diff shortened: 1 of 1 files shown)


src/XR/features/WebXRHandTracking.ts
```diff
@@ -370,9 +370,10 @@ export class WebXRHand implements IDisposable {
                 }
             }
         });
+
         // Update the invisible fingertip collidable
         if (this._nearInteractionMesh) {
-            const indexTipPose = this.trackedMeshes[hand.INDEX_PHALANX_TIP].position;
+            const indexTipPose = this.trackedMeshes[handJointReferenceArray.indexOf(XRHandJoint["index-finger-tip"])].position;
```
rickfromwork, line 376: @RaananW Is there some easier way to get the index needed for this.trackedMeshes than this? 

I was also using joint indices in some custom scenes, but now I have to just use the raw index instead:
Instead of `xrHand.xrController.inputSource.hand!.THUMB_PHALANX_TIP` for example, I'm now forced to use the raw index, `4` (which seems more error-prone in development).

RaananW: good question.  Since the API changed a bit (to strings instead of numbers) it is a little different to find the index. indexOf should work, but maybe we should either have a reverse-lookup map for string-to-index, or store the trackedMeshes in an object instead of an array. Will make sense when updating the tracked meshes. and their index is not so important.

Want to add it in this PR?

rickfromwork: Honestly, I'd be okay with using indexOf, but since neither XRHandJoint or HandJointReferenceArray are exposed (perhaps intentionally), app developers are limited to using index values or looping through the entirety of trackedMeshes.

RaananW (approved):

PR #9911, opened 2021-02-10
Fix typo of EventState.initialize, and 'initialize' in a comment

### LINKED ISSUE #9909: EventState::initalize should be EventState::initialize (typo)
There's really not much to it, just a typo I've noticed:

https://github.com/BabylonJS/Babylon.js/blob/8da127906fd3929fe4536a56cec586eae7e01e01/src/Misc/observable.ts#L27

seritools:
Fixes #9909 

Also found a comment with the same typo so I just went ahead and fixed that as well.


deltakosh (approved):

seritools:
Oops, missed the line about adding info in what's new! Since this is public API it should be added under Breaking Changes, right?

deltakosh:
Correct :)

RaananW:
If we want to keep back-compat, you can actually leave `public initalize` and proxy it to the actual function.

PR #9913, opened 2021-02-10
Properly register pointerUp event for touch input when window is blurred

PolygonalSun:
This PR contains a fix for an issue that would cause an error when the canvas would register a touch and then lose focus.  Forum post illustrating the issue can be found [here](https://forum.babylonjs.com/t/typeerror-cannot-read-property-2-of-undefined/18216).


src/DeviceInput/deviceInputSystem.ts
```diff
@@ -478,13 +478,16 @@ export class DeviceInputSystem implements IDisposable {
             // Handle Active Touches
             if (this.isDeviceAvailable(DeviceType.Touch)) {
                 const pointer = this._inputs[DeviceType.Touch];
+
                 for (let i = 0; i < pointer.length; i++) {
-                    if (pointer[i][PointerInput.LeftClick] === 1) {
+                    if (pointer[i] && pointer[i][PointerInput.LeftClick] === 1) {
```
Popov72, line 483: If you wanted to impress your friends, you could write:
```typescript
    if (pointer[i]?.[PointerInput.LeftClick] === 1) {
```
I have just learnt that the optional chaining syntax also works with array indexing, so I thought I should share my discovery ;)

PolygonalSun: Ooh, I appreciate this discovery.  I wasn't sure if there was an optional chaining syntax for multidimensional arrays.

deltakosh:
up?

PolygonalSun:
> up?

That title may be a bit of a misnomer.  Basically, the same code that'd run for a touch's pointerup event would run for all active touches when the window is blurred.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Modified to use optional chaining

deltakosh:
LGTM..merge when you want ;)

sebavan (approved):

PR #9925, opened 2021-02-13
improve hints for floor and fract vector2 methods

nekochanoide:
Also VS Code document format changed some weird (imo) formating, I hope this is fine.

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -355,6 +355,7 @@ export class Vector2 {
 
     /**
      * Gets a new Vector2 from current Vector2 floored values
+     * eg (1.2, 2.31, 1.54) returns (1, 2, 1)
      * @returns a new Vector2
      */
     public floor(): Vector2 {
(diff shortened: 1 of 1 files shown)


src/Maths/math.vector.ts
```diff
@@ -355,14 +355,16 @@ export class Vector2 {
 
     /**
      * Gets a new Vector2 from current Vector2 floored values
+     * eg (1.2, 2.31, 1.54) returns (1, 2, 1)
```
RaananW, line 358: if you are already adding a comment like that to vector2, i think it will be better to have the example contain only 2 values and not 3.

RaananW (changes requested):
Thanks for the doc update! 
If we are already adding an example, would be great to match it with the class in hand.

src/Maths/math.vector.ts
nekochanoide: Yes, makes sense

PR #9928, opened 2021-02-15
KHR_materials_pbrSpecularGlossiness: the texture was flagged as "non color data" incorrectly

Popov72:
Note however that the texture was still handled correctly (meaning in sRGB space), as this flagging is only used when dealing with .ktx2 textures, to force using RGBA over a compressed texture if ASTC/BC7 is not available (as per the GLTF doc recommendations).


deltakosh:
Test fails

PR #9929, opened 2021-02-16
Fix collision squared distance computation

CedricGuillemet:
Followup on https://forum.babylonjs.com/t/jumping-camera-movement/18352
Fixes issues introduced with https://github.com/BabylonJS/Babylon.js/pull/9873

Side note: a PR claiming 'optimize something ... ' shall not be merged unless there are metrics clearly showing a performance improvement. I doubt a significant improvement can benefit from that PR. Moreover, adding more states for the same value (here, velocity) almost always make the work more difficult as values must be properly synced.

- `distToCollisionSquared` is not `t * this._velocitySquaredLength` but `(t * velocityLength) * (t * velocityLength)` which is `t2 * _velocitySquaredLength`

### DIFF
--- a/src/Collisions/collider.ts
+++ b/src/Collisions/collider.ts
@@ -373,7 +373,7 @@ export class Collider {
         }
 
         if (found) {
-            var distToCollisionSquared = (t * Math.sqrt(this._velocitySquaredLength)) ** 2;
+            var distToCollisionSquared = t * t * this._velocitySquaredLength;
 
             if (!this.collisionFound || distToCollisionSquared < this._nearestDistanceSquared) {
                 // if collisionResponse is false, collision is not found but the collidedMesh is set anyway.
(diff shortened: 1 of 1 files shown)


Popov72:
> (t * velocityLength) * (t * velocityLength)

You probably meant: `(t + velocityLength) * (t + velocityLength)`

src/Collisions/collider.ts
```diff
@@ -260,6 +260,7 @@ export class Collider {
         if (!embeddedInPlane) {
             this._basePoint.subtractToRef(trianglePlane.normal, this._planeIntersectionPoint);
             this._velocity.scaleToRef(t0, this._tempVector);
+            this._velocitySquaredLength = this._velocity.lengthSquared();
```
Popov72, line 263: Is this line necessary? It is already done in `initialize` (I guess this function is always called before `_testTriangle`?) and `_velocity` is not modified anywhere in the class (and is private, so can't be modified by the outside), so the value computed in `initialize` should still be the right one.

CedricGuillemet:
@Popov72 No, actually, my simplification was not correct. It's actualy t ^ 2 * distanceSquared

src/Collisions/collider.ts
CedricGuillemet: Good catch! removed now

deltakosh:
can we merge?

Popov72:
That's ok for me.

PR #9933, opened 2021-02-16
createImageBitmap Abstraction

CedricGuillemet:
Followup for https://forum.babylonjs.com/t/babylon-native-on-ubuntu/15615/21
createImageBitmap is used for `CreateGroundFromHeightMap` . This PR is the JS part. Native part will be done once this is OK.

The goal here is also for me to learn how to properly wrap JS function as the same work will be needed for the Canvas/GUI.

### DIFF
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -25,6 +25,7 @@ import "./Extensions/engine.readTexture";
 import "./Extensions/engine.dynamicBuffer";
 import { IAudioEngine } from '../Audio/Interfaces/IAudioEngine';
 import { IPointerEvent } from "../Events/deviceInputEvents";
+import { CanvasGenerator } from '../Misc/canvasGenerator';
 
 declare type Material = import("../Materials/material").Material;
 declare type PostProcess = import("../PostProcesses/postProcess").PostProcess;
@@ -295,10 +296,30 @@ export class Engine extends ThinEngine {
      */
     public createImageBitmap(...args: any[]): Promise<ImageBitmap> //image: ImageBitmapSource, options?: ImageBitmapOptions
     {
-        if (args.length <= 2) {
-            return createImageBitmap(args[0], args[1]);
+        return createImageBitmap.apply(null, args);
+    }
+
+    /**
+     * Resize an image and returns the image data as an uint8array
+     * @param image image to resize
+     * @param bufferWidth destination buffer width
+     * @param bufferHeight destination buffer height
+     * @returns an uint8array containing RGBA values of bufferWidth * bufferHeight size
+     */
+    public resizeImageBitmap(image: HTMLImageElement | ImageBitmap, bufferWidth: number, bufferHeight: number): Uint8Array {
+        var canvas = CanvasGenerator.CreateCanvas(bufferWidth, bufferHeight);
+        var context = canvas.getContext("2d");
+
+        if (!context) {
+            throw new Error("Unable to get 2d context for resizeImageBitmap");
         }
-        return createImageBitmap(args[0], args[1], args[2], args[3], args[4], args[5]);
+
+        context.drawImage(image, 0, 0);
+
+        // Create VertexData from map data
+        // Cast is due to wrong definition in lib.d.ts from ts 1.3 - https://github.com/Microsoft/TypeScript/issues/949
+        var buffer = <Uint8Array>(<any>context.getImageData(0, 0, bufferWidth, bufferHeight).data);
+        return buffer;
     }
 
     /**
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -291,6 +291,17 @@ export class Engine extends ThinEngine {
     }
 
     /**
+     * Engine abstraction for createImageBitmap
+     */
+    public createImageBitmap(...args: any[]): Promise<ImageBitmap> //image: ImageBitmapSource, options?: ImageBitmapOptions
+    {
+        if (args.length <= 2) {
+            return createImageBitmap(args[0], args[1]);
+        }
+        return createImageBitmap(args[0], args[1], args[2], args[3], args[4], args[5]);
+    }
+
+    /**
      * Will flag all materials in all scenes in all engines as dirty to trigger new shader compilation
      * @param flag defines which part of the materials must be marked as dirty
      * @param predicate defines a predicate used to filter which materials should be affected
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -293,12 +293,12 @@ export class Engine extends ThinEngine {
 
     /**
      * Engine abstraction for createImageBitmap
-     * @param args parameters for createImageBitmap implementation
+     * @param image source for image
+     * @param options An object that sets options for the image's extraction.
      * @returns ImageBitmap
      */
-    public createImageBitmap(...args: any[]): Promise<ImageBitmap>
-    {
-        return createImageBitmap.apply(null, args);
+    public createImageBitmap(image: ImageBitmapSource, options?: ImageBitmapOptions): Promise<ImageBitmap> {
+        return createImageBitmap(image, options);
     }
 
     /**
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -293,8 +293,10 @@ export class Engine extends ThinEngine {
 
     /**
      * Engine abstraction for createImageBitmap
+     * @param args parameters for createImageBitmap implementation
+     * @returns ImageBitmap
      */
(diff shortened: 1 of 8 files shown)


deltakosh (approved):

src/Engines/engine.ts
```diff
@@ -290,6 +290,17 @@ export class Engine extends ThinEngine {
...
 
+    /**
+     * Engine abstraction for createImageBitmap
+     */
+    public createImageBitmap(...args: any[]): Promise<ImageBitmap> //image: ImageBitmapSource, options?: ImageBitmapOptions
+    {
+        if (args.length <= 2) {
+            return createImageBitmap(args[0], args[1]);
+        }
+        return createImageBitmap(args[0], args[1], args[2], args[3], args[4], args[5]);
```
bghgary, line 301: I think you can probably use `createImageBitmap.apply` here?

src/Engines/nativeEngine.ts
```diff
@@ -1771,6 +1772,20 @@ export class NativeEngine extends Engine {
         }
     }
 
+    /**
+     * Engine abstraction for createImageBitmap with Native implementation
+     */
+    public createImageBitmap(...args: any[]): Promise<ImageBitmap>
+    {
+        return new Promise((resolve, reject) => {
+            if (args.length <= 2) {
```
bghgary, line 1781: ditto on this

src/XR/features/WebXRImageTracking.ts
```diff
@@ -190,7 +191,8 @@ export class WebXRImageTracking extends WebXRAbstractFeature {
                         img.src = image.src;
                         img.onload = () => {
                             img.decode().then(() => {
-                                createImageBitmap(img).then((imageBitmap) => {
+                                const engine = EngineStore.LastCreatedEngine;
```
bghgary, line 194: `this._xrSessionManager.scene.getEngine()`

src/Engines/webgpuEngine.ts
```diff
@@ -2084,7 +2086,8 @@ export class WebGPUEngine extends Engine {
             gpuTextureWrapper = this._textureHelper.createGPUTextureForInternalTexture(texture);
         }
 
-        createImageBitmap(video).then((bitmap) => {
+        const engine = EngineStore.LastCreatedEngine;
+        engine && engine.createImageBitmap(video).then((bitmap) => {
```
bghgary, line 2090: This is the engine class right?

```suggestion
        this.createImageBitmap(video).then((bitmap) => {
```

src/Engines/webgpuEngine.ts
```diff
@@ -2030,7 +2031,8 @@ export class WebGPUEngine extends Engine {
             gpuTextureWrapper = this._textureHelper.createGPUTextureForInternalTexture(texture, width, height);
         }
 
-        createImageBitmap(canvas).then((bitmap) => {
+        const engine = EngineStore.LastCreatedEngine;
+        engine && engine.createImageBitmap(canvas).then((bitmap) => {
```
bghgary, line 2035: ditto

Popov72: You should be able to simply do `this.createImageBitmap(...)`?

src/Engines/webgpuEngine.ts
Popov72: ..ah, Gary already pointed to it!

src/Misc/fileTools.ts
```diff
@@ -160,7 +160,7 @@ export class FileTools {
 
         if (typeof Image === "undefined" || (engine?._features.forceBitmapOverHTMLImageElement ?? false)) {
             FileTools.LoadFile(url, (data) => {
-                createImageBitmap(new Blob([data], { type: mimeType }), { premultiplyAlpha: "none" }).then((imgBmp) => {
+                engine?.createImageBitmap(new Blob([data], { type: mimeType }), { premultiplyAlpha: "none" }).then((imgBmp) => {
```
Popov72, line 163: You should probably do `engine!` instead, as else we will be stuck without any error message if engine is null/undefined (even if it's only theoritical).

src/Engines/engine.ts
CedricGuillemet: Exactly what I was looking for!

CedricGuillemet:
I had to add a new abstraction function `resizeImageBitmap` that resize/convert to RGBA32 and returns an uint8array. I'm not sure the function name is appropriate as it does more than just resizing. Let me know.

src/Engines/engine.ts
```diff
@@ -290,6 +291,39 @@ export class Engine extends ThinEngine {
         return EngineStore.LastCreatedScene;
     }
 
+    /**
+     * Engine abstraction for createImageBitmap
+     * @param args parameters for createImageBitmap implementation
+     * @returns ImageBitmap
+     */
+    public createImageBitmap(...args: any[]): Promise<ImageBitmap>
```
ryantrem, line 299: Why does this have to be `any[]` for parameters? I think there is something fundamental about this API that I'm not understanding.

CedricGuillemet: createImageBitmap has 2 signatures : https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap
So, aby number of parameters are supported here.

bghgary: Hmm, having two signatures doesn't imply that you have to use any or `...`. You can just create overloads in TypeScript (as long as there are different number of arguments).

bghgary: From lib.dom.d.ts:

```typescript
declare function createImageBitmap(image: ImageBitmapSource, options?: ImageBitmapOptions): Promise<ImageBitmap>;
declare function createImageBitmap(image: ImageBitmapSource, sx: number, sy: number, sw: number, sh: number, options?: ImageBitmapOptions): Promise<ImageBitmap>;
```

src/Engines/nativeEngine.ts
```diff
@@ -143,6 +143,8 @@ interface INativeEngine {
     setTextureAnisotropicLevel(texture: WebGLTexture, value: number): void;
     setTexture(uniform: WebGLUniformLocation, texture: Nullable<WebGLTexture>): void;
     deleteTexture(texture: Nullable<WebGLTexture>): void;
+    createImageBitmap(parameters: any): ImageBitmap;
```
bghgary, line 146: this doesn't need to be `any` anymore.

src/Engines/nativeEngine.ts
```diff
@@ -1771,6 +1773,38 @@ export class NativeEngine extends Engine {
...
+     */
+    public createImageBitmap(...args: any[]): Promise<ImageBitmap>
+    {
+        return new Promise((resolve, reject) => {
+            if (args.length && Array.isArray(args[0]) && args[0].length) {
+                const image = this._native.createImageBitmap(args[0][0]);
+                if (image) {
+                    resolve(image);
+                } else if (reject) {
+                    reject();
```
bghgary, line 1789: should probably put a reason in the reject

src/Engines/nativeEngine.ts
```diff
@@ -1771,6 +1773,38 @@ export class NativeEngine extends Engine {
...
+                }
+                return;
+            }
+            throw new Error(`Unsupported data for createImageBitmap.`);
```
bghgary, line 1793: this is a bit odd, although this will translate to a reject call... why not just call reject here?

deltakosh:
What is the current status of this PR?

src/Engines/engine.ts
CedricGuillemet: from what I understand, you can declare overloadings but you can't have different signature for the implementation. I case either check parameters types for choosing the correct overloading or use `any` with `apply`. Which one shall I do?

CedricGuillemet:
@deltakosh still working on TS function signature.

src/Engines/engine.ts
bghgary: https://stackoverflow.com/questions/12688275/is-there-a-way-to-do-method-overloading-in-typescript

Here is an example: https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8CwAoaXoAcCuARiAJbDQBCYATgBRgBc0AdngLYECmVANNAYy3ZcAlIwBu8YgBMA3Bmy5CJMpVoNoEAC5ViTAOajoE6XMzZ8RUhWp0BbTlWgAfDdt17eBAPx2hVQ8ZSKPLYAL4YoUA

CedricGuillemet: I was taking the problem from the wrong side. there is only 1 override used in whole engine. So I changed the signature to only support it.

bghgary (approved):

PR #9943, opened 2021-02-19
Bind frame buffer and clear all rigCameras for XR

bghgary:
This is a patch fix for clearing both eyes when using XR in Babylon Native. The code is currently getting lucky and is clearing whatever is the last bound framebuffer for XR. This happens to work in a browser since the frame buffer of either eye is the same frame buffer. This is not true for Babylon Native and thus it only clears one eye which ever is the last one bound. Ideally, this code needs to be refactored so that it's not full of these specific branches for XR.


RaananW (approved):
Tested on emulator, Oculus Link (Desktop) and the oculus quest.

PR #9944, opened 2021-02-20
Creating a touchButton subset for togglable buttons and adding steady-state logic for touch

rickfromwork:
This change adds a new class intended to make toggle-state buttons easier to manage (when implemented).
This change also adds steady-state logic to the touch button, so we get updated positions for touch interactions every update, rather than once on each state change.

### DIFF
--- a/gui/src/3D/controls/touchButton3D.ts
+++ b/gui/src/3D/controls/touchButton3D.ts
@@ -96,7 +96,7 @@ export class TouchButton3D extends Button3D {
         }
 
         this._collisionMesh = collisionMesh;
-        this._collisionMesh.metadata = this;
+        this._injectGUI3DMetadata(this._collisionMesh).control = this;
 
         this.collidableFrontDirection = collisionMesh.forward;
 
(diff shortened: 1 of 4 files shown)


gui/src/3D/controls/touchButton3D.ts
```diff
@@ -250,30 +249,37 @@ export class TouchButton3D extends Button3D {
             newPushDepth = Math.max(newPushDepth, value);
         });
 
-        if (newPushDepth == ButtonState.Press) {
-            if (previousPushDepth == ButtonState.Hover) {
+        this._firePointerEvents(newPushDepth, previousPushDepth, pointOnButton);
+    }
+
+    protected _firePointerEvents(newButtonState: ButtonState, previousButtonState: ButtonState, pointOnButton: Vector3) {
+        const dummyPointerId = 0;
```
RaananW, line 256: If i understand correctly, this ID will not be used outside of this scope, but `0` as ID is the first a pointer will receive. that might confuse developers who relay on pointerId for pointer detection. It aht a problem in this scenario?

gui/src/3D/controls/touchToggleButton3D.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    /** Callback used to start toggle on animation */
+    public toggleOnAnimation: () => void;
+    /** Callback used to start toggle off animation */
+    public toggleOffAnimation: () => void;
+
+    /**
+     * An event triggered when the button is toggled on
+     */
+    public onToggleOnObservable = new Observable<Vector3WithInfo>();
```
RaananW, line 25: make sure to clear this and the next observable on dispose

gui/src/3D/controls/touchToggleButton3D.ts
```diff
@@ -0,0 +1,79 @@
...
+import { TouchButton3D, ButtonState } from "./touchButton3D";
+
+/**
+ * Class used as base class for touch-enabled toggleable buttons
+ */
+export class TouchToggleButton3D extends TouchButton3D {
+    private _isPressed = false;
+
+    /** Callback used to start toggle on animation */
+    public toggleOnAnimation: () => void;
```
RaananW, line 18: are these callbacks needed? why not use the observables? We are trying to avoid direct callbacks and are always pushing towards using observables instead

rickfromwork: Oh. I did this since Control3D has callbacks for the animations.

gui/src/3D/controls/touchButton3D.ts
```diff
@@ -250,30 +249,37 @@ export class TouchButton3D extends Button3D {
             newPushDepth = Math.max(newPushDepth, value);
         });
 
-        if (newPushDepth == ButtonState.Press) {
-            if (previousPushDepth == ButtonState.Hover) {
+        this._firePointerEvents(newPushDepth, previousPushDepth, pointOnButton);
```
rgerd, line 252: nit: While you're here, I would call these `newPushState`/`updatedPushState`. `previousPushState`, to me the depth name implies a continuous value for how deeply the button is being pushed

gui/src/3D/controls/touchButton3D.ts
rgerd: We're also going to need to eventually be able to distinguish between multiple pointers (near/far, left/right hand). Might be out of scope for this PR, but should we come up with special values to signify those pointers? Or maybe pass an  extra enum value to distinguish between id's for traditional pointers and id's for holographic hand/controller pointers?

rickfromwork: It hasn't caused any issues yet, though I also don't think anyone has really used this class yet either. I originally left it as 0 since I couldn't figure out how to get the pointer ID from the hand.

gui/src/3D/controls/touchToggleButton3D.ts
rickfromwork: I actually was never sure why there were both in Control3D, since they were both called at the same point. But if callbacks are deprecated, then it makes more sense.

rgerd (approved):

gui/src/3D/controls/touchToggleButton3D.ts
RaananW: Not really deprecated but not needed. Unless i am missing some inheritance issue?

gui/src/3D/controls/touchButton3D.ts
RaananW: The pointer ID can be set by you, as long as it is persistent between calls. This is what being done in the pointer-selection feature - https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/features/WebXRControllerPointerSelection.ts#L424 . Each connected controller gets a new ID, and keeps it for as long as it is connected.

RaananW:
@rgerd - Pointer IDs are assigned by the framework itself and are not fixed. Each pointer gets its own ID, and the idea is that you can identify a pointer source using the pointer id (for example if you want to make sure that the pointer up event has the same source as the pointer down event, you can check if their pointerIds match).
We might need to differ between the events in a different matter, or provide some sort of an interface to know what device correlates to what pointer ID, but we can't have fixed IDs for specific devices.

RaananW (approved):

PR #9953, opened 2021-02-23
WebGPU: add a bind groups cache

Popov72:
The biggest change is the addition of a "material context" (`IMaterialContext`) property to materials that is created and handled by the engine and that keeps track of the texture uniform values (in the case of WebGPU - for WebGL, the `engine.createMaterialContext()` call returns **undefined** as we don't need this).

Texture values were previously tracked in `effect._pipelineContext.textures` but it was not the right place because a given effect can be used by multiple materials. The engine needs access to those texture values to be able to create the binding groups (or to access existing groups from the cache). It also needs the list of uniform buffers, but those are passed by other means (calls to `engine.bindUniformBufferBase`).

Another change is the addition of a `DrawWrapper` (tentative name, can be changed) class that wraps the data passed to the engine through the call to `enableEffect` (which takes now a `Effect | DrawWrapper`): it holds the shader context (through the effect, and more precisely `effect._pipelineContext`), the material context (the new `IMaterialContext` structure) and a new "draw context" which is also created/handled by the engine and allows to manage data related to the draw call. For WebGPU, it is used for the `CompatibilityMode=false` case where the render pipeline / bind groups is cached the first time and reused for subsequent calls (see below for explanations about `CompatibilityMode`). For WebGL, this is not used and `engine.createDrawContext()` returns **undefined**. Note that we could also add a `geometry context` property to `DrawWrapper` (the property would be a member of `SubMesh`) for full coverage of data, but I didn't have a need for it (yet?) so did not create it for the time being.

In the process, I have updated the code to the latest WebGPU spec and corrected two bugs:
* Fixed instanced line meshes not displayed
* Sometimes we had some gl warnings when using the `ShadowDepthWrapper` class

I have also added a `compatibilityMode` flag to the engine as a proof of concept, but there is still a lot of work to do to have the "fast" mode (meaning, `compatibilityMode = false`) usable, so I have made it hidden.


src/Materials/standardMaterial.ts
```diff
@@ -1229,7 +1229,7 @@ export class StandardMaterial extends PushMaterial {
             DetailMapConfiguration.AddSamplers(samplers);
 
             PrePassConfiguration.AddUniforms(uniforms);
-            PrePassConfiguration.AddSamplers(uniforms);
+            PrePassConfiguration.AddSamplers(samplers);
```
deltakosh, line 1232: good catch!

deltakosh (approved):
Nothing seems wrong for me
LGTM

src/Materials/standardMaterial.ts
Popov72: No impact currently because `PrePassConfiguration.AddSamplers` has an empty implementation, but who knows in the future!

src/Materials/effectRenderer.ts
```diff
@@ -8,6 +8,7 @@ import { Constants } from '../Engines/constants';
 import { Observable } from '../Misc/observable';
 import { Effect } from './effect';
 import { DataBuffer } from '../Meshes/dataBuffer';
+import { DrawWrapper } from "./drawWrapper";
```
sebavan, line 11: Not sure about why we need the drawwrapper here ?

src/Materials/effectRenderer.ts
```diff
@@ -214,6 +215,9 @@ export class EffectWrapper {
      */
     public effect: Effect;
```
sebavan, line 216: Should this now be always going through drawwrapper and not being assignable ?

sebavan (approved):
Approve with tiny comments on EffectWrapper ?

src/Materials/effectRenderer.ts
Popov72: That's because `EffectRenderer` is calling `engine.enableEffect` and this call should always be made with a `DrawWrapper` as the parameter for WebGPU to work as expected.

src/Materials/effectRenderer.ts
Popov72: You're right, I have made accessors for `effect`.

deltakosh:
Some conflicts
LGTM (@sebavan  ack?)

PR #9958, opened 2021-02-23
Custom shader material for line mesh

FullStackForger:
PR is a follow up to BabylonJS [forum thread](https://forum.babylonjs.com/t/ability-to-change-line-material-eg-for-animated-line-shader/18625/3)

PR changes involve making use of `LineMesh.material` setter allowing to override custom line mesh material.
Custom material is stored internally as 
```
private _customShader: Nullable<ShaderMaterial>;
```
without overriding default `_colorShader` shader material. 

Additional `resetMaterial(dispose:boolean)` have been added allowing to restore original material with auto disposal (by default) of assigned custom shader material. 

Change doesn't break existing APIs, thus should be backward compatible. 

Notable change:  
`material` getter now returns `ShaderMaterial` (same for setter) instead of Material type, it is to do with calling shader defines, eg: clipping planes, etc.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -2,168 +2,169 @@
 
 ## Major updates
 
-- Infinite Morph Targets: When supported (WebGL2+) you are no more limited to 4 morph targets per mesh ([Deltakosh](https://github.com/deltakosh))
+-   Infinite Morph Targets: When supported (WebGL2+) you are no more limited to 4 morph targets per mesh ([Deltakosh](https://github.com/deltakosh))
 
 ## Updates
 
 ### General
 
-- Added static CenterToRef for vectors 2/3/4  ([aWeirdo](https://github.com/aWeirdo))
-- Added ability to view images (ktx2, png, jpg) to the sandbox. ([bghgary](https://github.com/bghgary))
-- Added optional smoothed normals for extruded procedural polygons. ([snagy](https://github.com/snagy))
-- Added support for infinite perspective cameras ([Deltakosh](https://github.com/deltakosh))
-- Added ability to enable/disable ArcRotateCamera zoom on multiTouch event ([NicolasBuecher](https://github.com/NicolasBuecher))
-- Moving button to shared uI folder.([msDestiny14](https://github.com/msDestiny14))
-- Added `collisionRetryCount` to improved collision detection ([CedricGuillemet](https://github.com/CedricGuillemet))
-- Moved sharedUI component to shared UI folder. ([msDestiny14](https://github.com/msDestiny14))
-- Added encapsulate and encapsulateBoundingInfo methods to BoundingInfo. ([Tolo789](https://github.com/Tolo789))
-- Added onLoadObservable to the textureDome class(es) ([RaananW](https://github.com/RaananW))
-- Modified InputManager to use DeviceInputSystem ([PolygonalSun](https://github.com/PolygonalSun))
-- Added a [helper class](https://doc.babylonjs.com/typedoc/classes/babylon.debug.directionallightfrustumviewer) to display the frustum of a directional light ([Popov72](https://github.com/Popov72))
-- Improved collision detection performance ([ottoville](https://github.com/ottoville/))
-- Added new helper functions for Quaternion.FromLookDirection and Matrix.LookDirection ([Alex-MSFT](https://github.com/Alex-MSFT))
+-   Added static CenterToRef for vectors 2/3/4 ([aWeirdo](https://github.com/aWeirdo))
+-   Added ability to view images (ktx2, png, jpg) to the sandbox. ([bghgary](https://github.com/bghgary))
+-   Added optional smoothed normals for extruded procedural polygons. ([snagy](https://github.com/snagy))
+-   Added support for infinite perspective cameras ([Deltakosh](https://github.com/deltakosh))
+-   Added ability to enable/disable ArcRotateCamera zoom on multiTouch event ([NicolasBuecher](https://github.com/NicolasBuecher))
+-   Moving button to shared uI folder.([msDestiny14](https://github.com/msDestiny14))
+-   Added `collisionRetryCount` to improved collision detection ([CedricGuillemet](https://github.com/CedricGuillemet))
+-   Moved sharedUI component to shared UI folder. ([msDestiny14](https://github.com/msDestiny14))
+-   Added encapsulate and encapsulateBoundingInfo methods to BoundingInfo. ([Tolo789](https://github.com/Tolo789))
+-   Added onLoadObservable to the textureDome class(es) ([RaananW](https://github.com/RaananW))
+-   Modified InputManager to use DeviceInputSystem ([PolygonalSun](https://github.com/PolygonalSun))
+-   Added a [helper class](https://doc.babylonjs.com/typedoc/classes/babylon.debug.directionallightfrustumviewer) to display the frustum of a directional light ([Popov72](https://github.com/Popov72))
+-   Improved collision detection performance ([ottoville](https://github.com/ottoville/))
+-   Added new helper functions for Quaternion.FromLookDirection and Matrix.LookDirection ([Alex-MSFT](https://github.com/Alex-MSFT))
 
 ### Engine
 
-- Moved all instance data from Geometry to Mesh such that the same Geometry objects can be used by many meshes with instancing. Reduces memory consumption on CPU/GPU. ([breakin](https://github.com/breakin)
-- Added NativeEngine configuration object parameter. ([drigax](https://github.com/drigax))
+-   Moved all instance data from Geometry to Mesh such that the same Geometry objects can be used by many meshes with instancing. Reduces memory consumption on CPU/GPU. ([breakin](https://github.com/breakin)
+-   Added NativeEngine configuration object parameter. ([drigax](https://github.com/drigax))
 
 ### Loaders
 
-- Added support for EXT_meshopt_compression for glTF loader. ([zeux](https://github.com/zeux))
-- Increased KHR_materials_transmission render target texture default size. ([Drigax](https://github.com/drigax))
-- Changed glTF loader to remove empty animation groups if there are no animation channels loaded with the given options. ([bghgary](https://github.com/bghgary))
-- Update glTF validator to `2.0.0-dev.3.3`. ([bghgary](https://github.com/bghgary))
-- Added support for KHR_xmp_json_ld for glTF loader. ([Sebavan](https://github.com/sebavan/), [bghgary](https://github.com/bghgary))
-- Added a `OptimizeNormals` option to the OBJ loader to smooth lighting ([Popov72](https://github.com/Popov72))
-- Added a `Prefiltered` option to the CubeTextureAssetTask ([MackeyK24](https://github.com/MackeyK24))
+-   Added support for EXT_meshopt_compression for glTF loader. ([zeux](https://github.com/zeux))
+-   Increased KHR_materials_transmission render target texture default size. ([Drigax](https://github.com/drigax))
+-   Changed glTF loader to remove empty animation groups if there are no animation channels loaded with the given options. ([bghgary](https://github.com/bghgary))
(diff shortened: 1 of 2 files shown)


FullStackForger:
It hurts but I haven't added test. Could use bit of help with that.

Popov72:
I'm not sure we should add so much complications only to retain both the existing material alongside the new one...

I would have simply implemented the setter as:
```javascript
    public set material(value: ShaderMaterial) {
        this._colorShader = value;
    }	    
```
and that's it.

If a user wants to reuse the old shader after having set a new one (something I don't think will really happen anyway - if you change the shader you will likely use it until the end of your app), then they can simply save it somewhere by calling the getter and later on reset it by calling the setter with this saved value.

FullStackForger:
Cheers Evgeni @Popov72

> I'm not sure we should add so much complications only to retain both the existing material alongside the new one

Fair point. Perhaps change was driven too much by my project requirement. Happy to simplify. Will recreate PR.

How do you guys test it? Can't see any tests for `LineMesh`?

Popov72:
We have the `LineEdgesRenderer` test that create line meshes and enable the edge renderer for them.

I don't really think we need a specific test for this PR, it's straightforward enough (only 2/3 lines changed).

You don't need to create a new PR (if it's what you meant by "recreating the PR"), you can simply update this one.

FullStackForger:
Do you guys have a guide on what is required in a PR?

> you can simply update this one.

Yep, already done. Not sure why it is failing though. 

Looking at Azure logs it is quite difficult to work out why step 1, 2 fails. Also in step 4 TS error is to do with scene file (nothing to do with my changes). Step 5 seems to have the same TS related error as step 4.

src/Meshes/linesMesh.ts
```diff
@@ -155,15 +155,15 @@ export class LinesMesh extends Mesh {
     /**
      * @hidden
      */
-    public get material(): Material {
+    public get material(): ShaderMaterial {
```
deltakosh, line 158: I will not force it to be ShaderMaterial actually. We could want to use NME for instance

deltakosh (changes requested):

src/Meshes/linesMesh.ts
FullStackForger: No probs. Wasn't sure what impact it might have. Better safe than sorry but happy to change for both getter & setter.

@deltakosh David did you guys have a chance to look at those Azure errors. I believe those have little to do with actual PR change, unless I am missing something.

deltakosh: Yes I just did fix them :) It was because of TS being TS. They break back compat on every tiny release
It is fixed now

I will merge your PR after your change

Thanks again for your help btw :)

FullStackForger: Hey no probs! Re your earlier comment... color shader is of `ShaderMaterial` type
```
private _colorShader: ShaderMaterial;
```
which seems fine looking at operations on shader defines. 
I believe it is better to leave `ShaderMaterial` in getter/setter until NME (what is it?) use case. 

One more small commit coming in.

FullStackForger:
@Popov72 @deltakosh Any reason for using `var` over `const` / `let` (especially with TS). 
Happy to clean the code up a bit if there are no objections.

deltakosh:
No please go ahead, this code was probably written before cost / let even existed ;D

FullStackForger:
> No please go ahead, this code was probably written before cost / let even existed ;D

Will do as a separate PR within next few days. 👍🏻

src/Meshes/linesMesh.ts
deltakosh: Well my comment is if we open it we need to use Material because you can use NME this way

deltakosh:
up?

FullStackForger:
> up?

Just pulled from upstream master. No conflicts.

deltakosh:
I mean, I  still see the ShaderMaterial in the PR ;0

src/Meshes/linesMesh.ts
Popov72: @FullStackForger You should be able to switch to `Material` for the base class of `_colorShader` without too much hassle by performing those changes:

```typescript
private _isShaderMaterial(shader: Material): shader is ShaderMaterial {
    return shader.getClassName() === "ShaderMaterial";
}

private _addClipPlaneDefine(label: string) {
    if (!this._isShaderMaterial(this._colorShader)) {
        return;
    }
    ...
}

private _removeClipPlaneDefine(label: string) {
    if (!this._isShaderMaterial(this._colorShader)) {
        return;
    }
    ...
}

public _bind(subMesh: SubMesh, effect: Effect, fillMode: number): Mesh {
    ...
    if (!this.useVertexColor && this._isShaderMaterial(this._colorShader)) {
        ...
    }
    ...
}
```

FullStackForger: 👌🏻

FullStackForger:
Sorry about the MIA status. Got stuck with company's admin stuff for couple of days. Haven't forgotten about you guys! 

I have followed @Popov72 guard suggestion. 

@deltakosh It should be done now but you might want to have a quick look to check if we should perhaps call `MaterialHelper.BindClipPlane` conditionally as well.

https://github.com/BabylonJS/Babylon.js/blob/ae6c0df2ce7754f9af2e8866920b6635966f4b93/src/Meshes/linesMesh.ts#L212

src/Meshes/linesMesh.ts
```diff
@@ -102,43 +107,51 @@ export class LinesMesh extends Mesh {
...
         // Clip planes
-        scene.clipPlane ? this._addClipPlaneDefine("CLIPPLANE") : this._removeClipPlaneDefine("CLIPPLANE");
-        scene.clipPlane2 ? this._addClipPlaneDefine("CLIPPLANE2") : this._removeClipPlaneDefine("CLIPPLANE2");
-        scene.clipPlane3 ? this._addClipPlaneDefine("CLIPPLANE3") : this._removeClipPlaneDefine("CLIPPLANE3");
-        scene.clipPlane4 ? this._addClipPlaneDefine("CLIPPLANE4") : this._removeClipPlaneDefine("CLIPPLANE4");
-        scene.clipPlane5 ? this._addClipPlaneDefine("CLIPPLANE5") : this._removeClipPlaneDefine("CLIPPLANE5");
-        scene.clipPlane6 ? this._addClipPlaneDefine("CLIPPLANE6") : this._removeClipPlaneDefine("CLIPPLANE6");
-
-        if (!this._colorShader.isReady(this)) {
+        if (!this._isShaderMaterial(this._lineMaterial)) {
```
Popov72, line 145: Should be `if (this._isShaderMaterial(this._lineMaterial)) {` instead.

Also, if you keep this check here, you should remove the checks in `_addClipPlaneDefine` and `_removeClipPlaneDefine` as they are redundant.

Popov72:
> It should be done now but you might want to have a quick look to check if we should perhaps call MaterialHelper.BindClipPlane conditionally as well.

I think you can keep it unconditionally. If the `vClipPlaneX` variables are not defined in the shader it should simply do nothing.

src/Meshes/linesMesh.ts
FullStackForger: Good catch. Would only prefer checks here but they are added to  `_(add/remove)ClipPlane` due to TS errors. Will remove them if you guys accept code with `//@ts-ignore` (not recommended)

FullStackForger:
> > It should be done now but you might want to have a quick look to check if we should perhaps call MaterialHelper.BindClipPlane conditionally as well.
> 
> I think you can keep it unconditionally. If the `vClipPlaneX` variables are not defined in the shader it should simply do nothing.

Perfect! It wasn't clear to me.

src/Meshes/linesMesh.ts
FullStackForger: Alternatively we can pass material as argument or cast to `ShaderMaterial` in both functions but neither seem elegant.

FullStackForger: On 2nd thought ....idc! If statement is gone all together from `isReady()` method. Enjoy. 😄

src/Meshes/linesMesh.ts
```diff
@@ -39,7 +40,11 @@ export class LinesMesh extends Mesh {
      */
     public intersectionThreshold: number;
 
-    private _colorShader: ShaderMaterial;
+    private _lineMaterial: Material;
+
+    private _isShaderMaterial(shader: Material): shader is ShaderMaterial {
```
deltakosh, line 45:
```suggestion
    private _isShaderMaterial(shader: Material):  {

```

deltakosh (changes requested):
just a tiny change

src/Meshes/linesMesh.ts
Popov72: I think that won't work because this method is used to let Typescript know that the parameter is a `ShaderMaterial` or not. If simply returning a `boolean`, I think we will have an error when using the parameter and assuming it is a `ShaderMaterial`.

For eg:
```typescript
        if (!this._isShaderMaterial(this._lineMaterial)) {
            return;
        }

        const define = "#define " + label;
        const index = this._lineMaterial.options.defines.indexOf(define);
```
`this._lineMaterial.options` won't work because Typescript does not know that `this._lineMaterial` is a `ShaderMaterial`. We would need to cast all usages as `as ShaderMaterial`: it's easier to have `_isShaderMaterial` returns a `shader is ShaderMaterial` type I think. I don't think there's a performance penality of doing it like that?

FullStackForger: @Popov72  is right. We used [guard function](https://www.typescriptlang.org/docs/handbook/advanced-types.html) for type check. If that method returns `boolean` casting is unavoidable forcing:
```
(this._lineMaterial as ShaderMaterial).options.defines.indexOf(define);
```
Surely we don't want that...

FullStackForger:
> just a tiny change

@deltakosh see above.

FullStackForger:
Oops. Accidental PR close....

FullStackForger:
@deltakosh / @Popov72 What's the status with this PR guys? Any chance we can get it merged?

Popov72:
That's ok for me but you should update the what's new file.

FullStackForger:
> That's ok for me but you should update the what's new file.

@Popov72 / @sebavan  I am not sure where I should inject it. Are you guys ok for me to create additional header. Thinking of something like...

> ### Meshes
> - `LineMesh` now allows assigning custom material with `material` setter. ([FullStackForger](https://github.com/fullstackforger))

Popov72:
That's ok for me.

FullStackForger:
@deltakosh / @Popov72  All green! 🚀

dist/preview release/what's new.md
```diff
@@ -2,167 +2,170 @@
...
-- Added an `OcclusionMaterial` to simplify depth-only rendering of geometry ([rgerd](https://github.com/rgerd))
-- PrePass can now be used in `RenderTargets` speeding up effects like SSAO2 or MotionBlur ([CraigFeldspar](https://github.com/CraigFeldspar))
-- Added support for morph targets to `ShaderMaterial` ([Popov72](https://github.com/Popov72))
+-   Added an `OcclusionMaterial` to simplify depth-only rendering of geometry ([rgerd](https://github.com/rgerd))
+-   PrePass can now be used in `RenderTargets` speeding up effects like SSAO2 or MotionBlur ([CraigFeldspar](https://github.com/CraigFeldspar))
+-   Added support for morph targets to `ShaderMaterial` ([Popov72](https://github.com/Popov72))
+
+### Meshes
+
+-   `LineMesh` now allows assigning custom material via `material` setter. ([FullStackForger]
```
Popov72, line 55: I think you missed your github address?

FullStackForger: Oops 🗡️

FullStackForger: Sorry! Too much context switching. It is now done.

deltakosh:
a last conflict and go ;)

FullStackForger:
Yep. See it. To do with spacing in _what's new_ file. Merged accepting incoming and re-added my note.

Issue #9921, opened 2021-02-13
Implement Parenting System 


msDestiny14:
Features:
- [ ] Implement how we select down the tree.
- [x] Implement serialization hookups

Bugs:
- [x] Fix bug with parenting order
- [x] Fix bug with alignment not parenting
- [ ] Fix bug related to update dropdown. (react bug)

msDestiny14:
Outdated.

PR #9963, opened 2021-02-24
[GE] Parenting System

msDestiny14:
Controls can now be parented to containers using a drop down selection. (UI temporary).

Each GUINode contains information of the nodes children in order to iterate through deleting/updating. The child node also contains information of its parent.

https://github.com/BabylonJS/Babylon.js/issues/9921


guiEditor/src/diagram/workbench.tsx
```diff
@@ -203,151 +217,148 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
-        
+
         // Create our first scene.
         this._scene = new Scene(engine);
         this._scene.clearColor = new Color4(0.2, 0.2, 0.3, 1.0);
-        let camera = new ArcRotateCamera(
-            "Camera", -Math.PI / 2, 0, 1024, Vector3.Zero(), this._scene);
-        const light = new HemisphericLight(
-            "light1", Axis.Y, this._scene);
+        let camera = new ArcRotateCamera("Camera", -Math.PI / 2, 0, 1024, Vector3.Zero(), this._scene);
```
RaananW, line 247: `let` suggests that the camera might be reassigned later in the code. Is that the case here?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -65,44 +63,49 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
         });
 
         this.props.globalState.hostDocument!.addEventListener("keyup", () => this.onKeyUp(), false);
-        this.props.globalState.hostDocument!.addEventListener("keydown", evt => {         
-            this._ctrlKeyIsPressed = evt.ctrlKey;
-        }, false);
-        this.props.globalState.hostDocument!.defaultView!.addEventListener("blur", () => {
-            this._ctrlKeyIsPressed = false;
-        }, false);     
+        this.props.globalState.hostDocument!.addEventListener(
```
RaananW, line 88: seems like this is mainly formatting and not a code change, but - are those events being released at a later time?

guiEditor/src/diagram/guiNode.ts
```diff
@@ -181,18 +185,28 @@ export class GUINode {
         }
     }
 
-    public addGui(childNode: GUINode) {
-        if (!this._isContainer) return;
+    public addChildGui(childNode: GUINode) {
+        if (!this.isContainer) return;
+        childNode.parent = this;
```
RaananW, line 190: is that needed? can't you get this information from the scene itself, after setting the element's parent (or container)?

RaananW (commented):
Looks great! 
I still have doubts about the need to keep the parenting references in the react context. Isn't it enough to analyze the scene and use it as the source-of-truth?

msDestiny14:
@RaananW so I thought about removing the parent. But then there is no way to extract the GUINode information (needed for the editor) from the gui control. Unless we want to remove the GUINode wrapper class entirely. I'm not against that but need to look more into it.

guiEditor/src/diagram/guiNode.ts
msDestiny14: See below comments.

guiEditor/src/diagram/workbench.tsx
msDestiny14: Like this? Seems like a lot of code. Is there a better way to do this?

RaananW:
@msDestiny14 - managing two states might be a challenging task once you get to more complicated tasks. 
If the scene has all of the information you need, it should be your source of information. Of course, it is possible to add some form of a proxy class that augments the scene's object, if this is needed, and then you have a 1-to-1 connection between a GUI element and the object. But I don't recommend managing state based on that (like checking parenting using this object instead of the scene), as this could lead to complications in a later stage.
If there is no way around it, then it is needed, of course! I just want to make sure you don't have a clean way to go around the issue

guiEditor/src/diagram/workbench.tsx
RaananW: That won't work, as the function you are passing is not the same object as the one passed to the addEventListener. To get that to work you will need to define local functions that can be passed as a variable to both add and removeEventListener

RaananW:
> I started removing the parent and children members which was fine initially because you can get that information from the control itself. But how does it then know which GUINode it belongs too. If it is needed, you should keep it, of course. But it should only contain the augmented information, not the same information as the scene. the GUINode should have a reference to the scene-object, which has the parenting information. Will that be possible in this usecase?

BTW - you have a small conflict that prevents us from merging this :-)

msDestiny14:
I'm gonna think about this a little more over the weekend and let you know Monday :)

guiEditor/src/components/parentingPropertyGridComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+    globalState: GlobalState;
+}
+
+export class ParentingPropertyGridComponent extends React.Component<IParentingPropertyGridComponentProps> {
+    constructor(props: IParentingPropertyGridComponentProps) {
+        super(props);
+    }
+    public parentIndex: number = 0;
+
+    private _isContainer(guiControl: Control) {
```
RaananW, line 20: Would it make sense having it in the main repository?

RaananW (approved):

guiEditor/src/components/parentingPropertyGridComponent.tsx
msDestiny14: I ad it there originally but I didn't see it being used any where but here for now.

PR #9964, opened 2021-02-24
Prevent removal of undefined pointer from DeviceInputSystem

PolygonalSun:
This PR fixes an issue where an error will be thrown if a touch is already down before the DeviceInputSystem is initialized.  The new behavior will just ignore the initial pointerup when it is released.


deltakosh (approved):

ryantrem:
We have exactly the same problem in `DeviceSourceManager`: https://github.com/BabylonJS/Babylon.js/blob/10cfc6880bd8c4637b7ae29331f48bd2d786af70/src/DeviceInput/InputDevices/deviceSourceManager.ts#L174
This is the one that is hit in the context of Babylon Native. Will you include a fix for that as well?

PolygonalSun:
> We have exactly the same problem in `DeviceSourceManager`:
> 
> https://github.com/BabylonJS/Babylon.js/blob/10cfc6880bd8c4637b7ae29331f48bd2d786af70/src/DeviceInput/InputDevices/deviceSourceManager.ts#L174
> 
> 
> This is the one that is hit in the context of Babylon Native. Will you include a fix for that as well?

I can also fix that.  It's just a matter checking if the device exists in the _devices array before deleting when the _removeDevice function is called.  I can add this change to the DeviceSourceManager but the change to the DeviceInputSystem should also be reflected in the NativeInput implementation.

CoPrez:
I was testing the fix to DeviceSourceManager in the context of React Native. It looks like the change passes the exception into the _updateFirstDevices function now.
MicrosoftTeams-image

The callstack will look a bit funky because I am overwriting the _removeFunction in our TransformationSystem.ts file to match the changes you've made in this PR (as a quick an easy way to test the changes rather than pulling in and locally building everything). My injected function in TransformationSystem.ts is simply this:
```ts
DeviceSourceManager.prototype._removeDevice = function (deviceType: DeviceType, deviceSlot: number) {
    if (this._devices[deviceType]?.[deviceSlot]) {
        delete this._devices[deviceType][deviceSlot];
    }
    this._updateFirstDevices(deviceType);
};
```

PolygonalSun:
> I was testing the fix to DeviceSourceManager in the context of React Native. It looks like the change passes the exception into the _updateFirstDevices function now.
> MicrosoftTeams-image
> 
> The callstack will look a bit funky because I am overwriting the _removeFunction in our TransformationSystem.ts file to match the changes you've made in this PR (as a quick an easy way to test the changes rather than pulling in and locally building everything). My injected function in TransformationSystem.ts is simply this:
> 
> ```ts
> DeviceSourceManager.prototype._removeDevice = function (deviceType: DeviceType, deviceSlot: number) {
>     if (this._devices[deviceType]?.[deviceSlot]) {
>         delete this._devices[deviceType][deviceSlot];
>     }
>     this._updateFirstDevices(deviceType);
> };
> ```

Hey @CoPrez,

I recently pushed a change to the _updateFirstDevices function that I think should address that issue.  Would you be able to try it out in your React Native instance?  I created a way that I think could be used to test it but I just wanted to make sure before merging.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -171,7 +171,10 @@ export class DeviceSourceManager implements IDisposable {
      * @param deviceSlot "Slot" or index that device is referenced in
      */
     private _removeDevice(deviceType: DeviceType, deviceSlot: number) {
-        delete this._devices[deviceType][deviceSlot];
+        if (this._devices[deviceType]?.[deviceSlot]) {
```
bghgary, line 174: I'm not sure I get this change. How is it possible for this if check to be false?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -192,10 +195,12 @@ export class DeviceSourceManager implements IDisposable {
             case DeviceType.Generic:
                 const devices = this._devices[type];
                 delete this._firstDevice[type];
```
bghgary, line 197: nit: logically speaking it would make more sense if these lines were reversed
```suggestion
                delete this._firstDevice[type];
                const devices = this._devices[type];
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Answered offline

Issue #9922, opened 2021-02-13
Implement Z Order Changing


msDestiny14:
Be able to move a control up or down in the zorder. This can be just with simple buttons for now.

PR #9967, opened 2021-02-25
[GE] Adding ability to change zOrder

msDestiny14:
zorder

https://github.com/BabylonJS/Babylon.js/issues/9922


RaananW (approved):
One line - and all good. this is great :-)

PR #9968, opened 2021-02-25
Normalize quat/update matrix for drag behavior

CedricGuillemet:
follow up for https://forum.babylonjs.com/t/massive-bug-gizmo-rotationquaternion/18666/6
and https://forum.babylonjs.com/t/undesirable-bounding-box-gizmo-custom-drag-behavior/18660

- normalize quaternion out of matrix update
- compute absolute position before using it in pointerdragbehavior plane

Not much to say for the quaternion normalization. It drifted after a while when decomposing/recomposing the values.
For the dragbehavior, this happens when select and dragging the node (with button up), the absolute position is (0,0,0) and the mesh got an offset. Fixed by computing the absolute position before using it.


Popov72 (approved):
LGTM

PR #9969, opened 2021-02-25
KTX2 decoder: select a transcode format with a decision tree

Popov72:
The transcode format now depends on the source texture format (ETC1S/UASTC) as per the guidelines from Khronos.


deltakosh:
Webgl1 issue? or just something to ignore?

Popov72:
You can ignore, it's the "Displacement map" test that timed out.

PR #9995, opened 2021-03-02
Teleportation fix

syntheticmagus:
Fix for issue with reference space change logic in WebXRCamera and teleportation logic which was causing the camera to become positionally headlocked during continuous motion.

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -4125,6 +4125,19 @@ export class Matrix {
     }
 
     /**
+     * Switch transform matrix handedness (invert z). Note that this is an in-place operation.
+     * @returns this matrix
+     */
+    public switchHandedness(): Matrix {
+        this._m[2] *= -1;
+        this._m[6] *= -1;
+        this._m[8] *= -1;
+        this._m[9] *= -1;
+        this._m[14] *= -1;
+        return this;
+    }
+
+    /**
      * Copy the current matrix from the given one
      * @param other defines the source matrix
      * @returns the current updated matrix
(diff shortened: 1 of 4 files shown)


src/Maths/math.vector.ts
```diff
@@ -4124,6 +4124,19 @@ export class Matrix {
         return result;
     }
 
+    /**
+     * Switch transform matrix handedness (invert z). Note that this is an in-place operation.
+     * @returns this matrix
+     */
+    public switchHandedness(): Matrix {
```
RaananW, line 4131: This function already exists (toggleModelMatrixHandInPlace)

src/XR/webXRCamera.ts
```diff
@@ -268,63 +268,22 @@ export class WebXRCamera extends FreeCamera {
...
             }
-            pos.negateInPlace();
 
-            const transform2 = new XRRigidTransform({ x: pos.x, y: pos.y, z: pos.z });
-            // Update offset reference to use a new originOffset with the teleported
-            // player position and orientation.
-            // This new offset needs to be applied to the base ref space.
-            this._xrSessionManager.referenceSpace = referenceSpace.getOffsetReferenceSpace(transform2);
+            transformMat.decompose(undefined, this._referenceQuaternion, this._referencedPosition);
+            const transform = new XRRigidTransform(this._referencedPosition, this._referenceQuaternion);
```
RaananW, line 285: Does this work as expected in all XRRigidTransform implementation?

RaananW (approved):
I just have a quick question (and a duplicate function)

PR #9998, opened 2021-03-03
Unnecessary check in physicsImposter removed

Quadtree:
I recently encountered the bug that MackeyK24 encountered in 2018: https://forum.babylonjs.com/t/move-kinematic-rigidbodies-in-code/17568/6

Basically, as mentioned in the original forum thread it's impossible to move a kinematic body without issues unless you first set `.rotationQuaternion` on the object (unless the object started with one, some do). The workaround is easy, but it's annoying, and pretty confusing for new users.

This Playground is an example of the issue (very slightly adapted from Cedric's in the above thread): https://playground.babylonjs.com/#66PS52#120

If you remove the `ground.rotationQuaternion = null;` line, the ground moves as expected.

This PR simply removes the check for `.rotationQuaternion`. Since the check on line 849 explicitly handles the case where the quaternion is invalid, I can't imagine the check on line 852 is intended. Looking at the history of the file, I think the check was accidentally left in from an earlier version, but I could be wrong about that.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -146,6 +146,7 @@
 - Fix an issue with hand-detachment when using hand tracking in WebXR ([#9882](https://github.com/BabylonJS/Babylon.js/issues/9882)) ([RaananW](https://github.com/RaananW))
 - Fix issue with cursor and 'doNotHandleCursors' on GUI ([msDestiny14](https://github.com/msDestiny14))
 - Fix thin instances + animated bones not rendered in the depth renderer ([Popov72](https://github.com/Popov72))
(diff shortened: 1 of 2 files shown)


RaananW:
All physics objects must have rotation quaternion. If they don't there was an issue when creating the impostor. Checking for one is just classic defensive programming in action. 
The fix will ignore the initial rotation of the body, as the quaternion is not present (unless, of course, I'm missing something). I'll run a few tests soon.
It feels like a fix for a symptom that is integrated a little deeper than we might think.

Quadtree:
Ah, you're right. This fix would cause it to ignore Euler angles. I didn't test that case, sorry about that. The situation is a little weird, but I can't think of a completely safe fix, so I think I'm just gonna close this for now.

PR #10001, opened 2021-03-03
Provide additional attributes to customize the posture of the gizmo

ecojust:
Sometimes we want the gizmo to be displayed in a specific posture，

Neither attachedMesh’s  posture nor default value(new Quaternion(0, 0, 0, 1)).

So，i think it is necessary to provide an attribute(_customRotationQuaternion) to support this demand.

### DIFF
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -203,7 +203,7 @@ export class Gizmo implements IDisposable {
             else {
                 if (this._customRotationQuaternion) {
                     this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion);
-                }else {
+                } else {
                     this._rootMesh.rotationQuaternion!.set(0, 0, 0, 1);
                 }
             }
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -167,10 +167,15 @@ export class Gizmo implements IDisposable {
     }
 
     /**
-     * update customRotationQuaternion
-     * @param customRotationQuaternion Quaternion that defines the gizmo‘s pose
+     * poseture that the gizmo will be display
+     * * When set null, default value will be used (Quaternion(0, 0, 0, 1))
      */
-    public setCustomRotationQuaternion(customRotationQuaternion: Nullable<Quaternion>) {
+
+    public get customRotationQuaternion(): Nullable<Quaternion> {
+        return this._customRotationQuaternion;
+    }
+
+    public set customRotationQuaternion(customRotationQuaternion: Nullable<Quaternion>) {
         this._customRotationQuaternion = customRotationQuaternion;
     }
 
@@ -196,7 +201,7 @@ export class Gizmo implements IDisposable {
                 effectiveNode.getWorldMatrix().decompose(undefined, this._rootMesh.rotationQuaternion!);
             }
             else {
-                this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion || new Quaternion(0, 0, 0, 1));
+                this._customRotationQuaternion ? (this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion)) : (this._rootMesh.rotationQuaternion!.set(0, 0, 0, 1));
             }
 
             // Scale
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -201,7 +201,11 @@ export class Gizmo implements IDisposable {
                 effectiveNode.getWorldMatrix().decompose(undefined, this._rootMesh.rotationQuaternion!);
             }
             else {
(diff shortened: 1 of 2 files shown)


CedricGuillemet:
Please update what'snew.md

ecojust:
> Please update what'snew.md

Already added now.

CedricGuillemet (approved):
LGTM!

src/Gizmos/gizmo.ts
```diff
@@ -188,7 +196,7 @@ export class Gizmo implements IDisposable {
                 effectiveNode.getWorldMatrix().decompose(undefined, this._rootMesh.rotationQuaternion!);
             }
             else {
-                this._rootMesh.rotationQuaternion!.set(0, 0, 0, 1);
+                this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion || new Quaternion(0, 0, 0, 1));
```
Popov72, line 199: I think the `_update` function may be called often (each frame?), so you should avoid creating a new object each time.

Maybe:
```typescript
if (this._customRotationQuaternion) {
    this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion);
} else {
    this._rootMesh.rotationQuaternion!.set(0, 0, 0, 1);
}
```

src/Gizmos/gizmo.ts
```diff
@@ -166,6 +166,14 @@ export class Gizmo implements IDisposable {
         });
     }
 
+    /**
+     * update customRotationQuaternion
+     * @param customRotationQuaternion Quaternion that defines the gizmo‘s pose
+     */
+    public setCustomRotationQuaternion(customRotationQuaternion: Nullable<Quaternion>) {
```
Popov72, line 173: Why not using standard getter / setter?

```typescript
public get customRotationQuaternion():  Nullable<Quaternion>) {
    return this._customRotationQuaternion;
}

public set customRotationQuaternion(customRotationQuaternion: Nullable<Quaternion>) {
    this._customRotationQuaternion = customRotationQuaternion;
}

File: src/Gizmos/gizmo.ts (hunk 1 above)
Comment (ecojust) [reply src/Gizmos/gizmo.ts:199]: how about this:

`this._rootMesh.rotationQuaternion!.set(...(this._customRotationQuaternion?.asArray() || [0, 0, 0, 1]));`

i think the if branch makes the code look redundant,

and,literal representations can be more efficient than object representations

File: src/Gizmos/gizmo.ts (hunk 2 above)
Comment (ecojust) [reply src/Gizmos/gizmo.ts:173]: that could be better!

and,I wonder if it's necessary to open the getter method. Are we going to use the getter?

**Comment by deltakosh** [comment]
@Popov72 you can merge if you are good with the PR

File: src/Gizmos/gizmo.ts (hunk 1 above)
Comment (Popov72) [reply src/Gizmos/gizmo.ts]: No, that won't be better because `asArray` is creating a new array each time you call it.

I don't see a way to make it faster than calling `copyFrom` / `set` as in my snippet because those two methods simply take the parameters and put them in the object.

File: src/Gizmos/gizmo.ts (hunk 2 above)
Comment (Popov72) [reply src/Gizmos/gizmo.ts]: I don't know, but even if it's not, it is a one liner so that's not a big deal and if someone ever wants it one day it is there!

File: src/Gizmos/gizmo.ts [hunk 3]
```diff
@@ -196,7 +201,7 @@ export class Gizmo implements IDisposable {
                 effectiveNode.getWorldMatrix().decompose(undefined, this._rootMesh.rotationQuaternion!);
             }
             else {
-                this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion || new Quaternion(0, 0, 0, 1));
+                this._customRotationQuaternion ? (this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion)) : (this._rootMesh.rotationQuaternion!.set(0, 0, 0, 1));
```
Popov72, line 204: I'm not fond of using the ternary operator to hide a `if` statement when the expressions on the right don't return anything...

IMO the intent is clearer if simply using an `if` in the first place.

I don't know if I'm too picky or if it's something that should be avoided according to the Babylon code writing rules: @sebavan, @RaananW, @deltakosh any thoughts?

sebavan: nope totally agree :-) we should keep it for assignments

ecojust: @Popov72  I agree with you，when the expressions on the right don't return anything，use a `if` statement make the logic clearer

src/Gizmos/gizmo.ts
```diff
@@ -201,7 +201,11 @@ export class Gizmo implements IDisposable {
                 effectiveNode.getWorldMatrix().decompose(undefined, this._rootMesh.rotationQuaternion!);
             }
             else {
-                this._customRotationQuaternion ? (this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion)) : (this._rootMesh.rotationQuaternion!.set(0, 0, 0, 1));
+                if (this._customRotationQuaternion) {
+                    this._rootMesh.rotationQuaternion!.copyFrom(this._customRotationQuaternion);
+                }else {
```
Popov72, line 206: nit: missing space after }

PR #10019, opened 2021-03-05
WebGPU: fix scissor, add cache for stencil ref and blend color

Popov72:
Also added a validation test for scissor and the initial computation for `hardwareScalingLevel` (same computation than in `ThinEngine`).


src/Engines/webgpuEngine.ts
```diff
@@ -883,16 +947,94 @@ export class WebGPUEngine extends Engine {
...
+                this._startMainRenderPass(!hasScissor, backBuffer ? color : null, depth, stencil);
+            }
+            if (hasScissor) {
+                this._applyScissor(this._currentRenderPass!);
+                this._clearFullQuad(backBuffer ? color : null, depth, stencil);
+            }
         }
     }
 
+    private _clearFullQuad(clearColor?: Nullable<IColor4Like>, clearDepth?: boolean, clearStencil?: boolean): void {
```
sebavan, line 975: As this creates GC should we merge and you do the cache in another PR ?

Popov72: Ok.

sebavan:
@Popov72 Looks like there is a conflict :-) I ll build the nightly right after the merge.

PR #10026, opened 2021-03-08
WebGPU: Refactor and optimize scissor clearing

Popov72:


sebavan (commented):
My only recommendation would be to rename webgpuMisc to something closer from clearQuad of you can ?

Popov72:
In fact I had named this file Misc because I thought that other things could go in this file, `clearQuad` being a small class. Going to rename it for the time being.

Issue #9983, opened 2021-02-27
[GE] Implement Tree View


msDestiny14:
- [x] Create basic visual representation for all gui elements in the tree (including children)
- [x] Be able to select elements from the tree

PR #10029, opened 2021-03-08
[GE] Adding Scene Explorer 

msDestiny14:
scene

https://github.com/BabylonJS/Babylon.js/issues/9983

Implementation from the Inspector.


msDestiny14:
Since there is a good amount of files that are simply 1:1 copied from the inspector with 0 changes, I wonder if I can do a shared system like the other components from NME. Will look into this.

PR #10031, opened 2021-03-09
MeshImpostor local position transform

CedricGuillemet:
followup https://forum.babylonjs.com/t/ammo-js-plugin-left-handed-coordinate-bug/18943/4
positions for mesh impostor must be in toplevelObject space once its ammo transform is setup (without scale). So, compute a local matrix that is the mesh world transform in ammo top level transform space.


Popov72 (approved):
LGTM

PR #10042, opened 2021-03-11
Add support for local cube map to refraction cube textures

Popov72:
Also fixed support for local cube map to node material reflection textures.


deltakosh (approved):

Popov72:
Failing tests in WebGL1 can be ignored (there's a timeout for "Basis loader" and it seems the following test fails because of that as the error makes no sense to me: "TypeError: currentScene is null").

PR #10043, opened 2021-03-11
Canvas abstraction

CedricGuillemet:


gui/src/2D/controls/image.ts
```diff
@@ -307,14 +309,16 @@ export class Image extends Control {
...
     public _rotate90(n: number, preserveProperties: boolean = false): Image {
-        let canvas = document.createElement("canvas");
-
-        const context = canvas.getContext("2d")!;
         const width = this._domImage.width;
         const height = this._domImage.height;
 
-        canvas.width = height;
-        canvas.height = width;
+        const engine = Engine.LastCreatedEngine;
```
bghgary, line 315: This doesn't feel right. If all of our platform abstractions are on engine, then this class needs to depend on engine. Using `LastCreatedEngine` seems wrong.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "babylonjs",
-    "version": "5.0.0-alpha.16",
+    "version": "5.0.0-alpha.18",
```
bghgary, line 3: why is this file changing?

src/Engines/thinEngine.ts
```diff
@@ -4835,4 +4836,20 @@ export class ThinEngine {
...
+     */
+    public createCanvas(width: number, height: number) : ICanvas {
+        if (typeof document === "undefined") {
+            return <ICanvas>(<any>(new OffscreenCanvas(width, height)));
+        }
+        let canvas = <ICanvas>(<any>(document.createElement("canvas")));
+        canvas.width = width;
+        canvas.height = height;
+        return canvas;
+    }
```
bghgary, line 4854: Do you need to remove canvasGenerator.ts?

src/Engines/nativeEngine.ts
```diff
@@ -2224,6 +2225,22 @@ export class NativeEngine extends Engine {
...
+     * @return ICanvas interface
+     */
+    public createCanvas(width: number, height: number) : ICanvas {
+        if (_native.NativeCanvas) {
+            const canvas = new _native.NativeCanvas();
+            canvas.width = width;
+            canvas.height = height;
+            return canvas;
+        }
+        throw new Error("Native Canvas plugin not available.");
```
bghgary, line 2241: nit: i think this is clearer

```suggestion
        if (!_native.NativeCanvas) {
            throw new Error("Native Canvas plugin not available.");
        }

        const canvas = new _native.NativeCanvas();
        canvas.width = width;
        canvas.height = height;
        return canvas;
```

bghgary (commented):
The only thing I don't think should go in is the usage of LastCreatedEngine. Everything else looks good to me.

package-lock.json
CedricGuillemet: I don't know what's changing this file but I should not have commited it.

gui/src/2D/controls/image.ts
CedricGuillemet: I agree, I will change that.

src/Engines/thinEngine.ts
CedricGuillemet: I did that a month ago, what I remember is cyclic dependency and I had to move it there. I don't remember the details.

bghgary: Right now there is duplicate code in canvasGenerator.ts right? So something needs to happen. :-)

gui/src/2D/controls/image.ts
CedricGuillemet: Actually no, I don't see how to get an engine instance without `LastCreatedEngine`. Any idea?

src/Engines/nativeEngine.ts
```diff
@@ -2225,6 +2226,22 @@ export class NativeEngine extends Engine {
...
     }
 
+    /**
+     * Create a canvas
+     * @param width width
+     * @param height height
+     * @return ICanvas interface
+     */
+    public createCanvas(width: number, height: number) : ICanvas {
+        if (!_native.NativeCanvas) {
```
chrisfromwork, line 2236: should the native engine expose static functions for loading ttfs? where will that logic live?

src/Materials/Textures/dynamicTexture.ts
```diff
@@ -45,7 +45,7 @@ export class DynamicTexture extends Texture {
             this._canvas = options;
```
chrisfromwork, line 45: we end up down this path for AdvancedDynamicTexture.CreateFullscreenUI. it generates exceptions because we submit a 0x0 texture size which bgfx doesn't support. we may want to check that the options are a canvas element compared to assuming they are a canvas element because they have a getContext function

src/Engines/nativeEngine.ts
CedricGuillemet: No strong opinion for that yet.

src/Materials/Textures/dynamicTexture.ts
CedricGuillemet: 0x0 texture should not crash when calling `AdvancedDynamicTexture.CreateFullscreenUI`. I guess the layer will resize the texture later at runtime. NativeCanvas should discard texture creation and rendering until a valid size is provided.

src/Engines/thinEngine.ts
CedricGuillemet: I removed canvasGenerator.ts

bghgary (approved):

Issue #10041, opened 2021-03-10
GUI linkWithMesh wrong behavior


neymanushka:
Control.linkWithMesh does not work correctly and superslow if i set AdvancedDynamicTexture.renderAtIdealSize = true.

[forum entry](https://forum.babylonjs.com/t/gui-renderatidealsize-and-linktomesh/19052)

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com/#SZV9D1#1):
- Expected result:
TextBlock control in the center of mesh
- Current result:
TextBlock far from the mesh

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: chrome,ff
 - Version: chrome  89.0.4389.82,firefox 86

**Additional context**
perfomance down issue only visible in firefox browser @msDestiny14

PR #10045, opened 2021-03-11
Fixes for RenderAtIdealSize

msDestiny14:
I noticed that the offset seemed to be calculated twice so I removed one of them. 

https://github.com/BabylonJS/Babylon.js/issues/10041


msDestiny14:
If anyone knows of additional examples I should test with to make sure I'm not breaking backwards compatibility please let me know.

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -501,15 +501,6 @@ export class AdvancedDynamicTexture extends DynamicTexture {
         var renderWidth = engine.getRenderWidth() * this._renderScale;
         var renderHeight = engine.getRenderHeight() * this._renderScale;
 
-        if (this._renderAtIdealSize) {
```
deltakosh, line 504: Can you elaborate more on why you remove that code?

deltakosh:
You have plenty here: https://doc.babylonjs.com/playground?q=renderAtIdealSize&type=code

gui/src/2D/valueAndUnit.ts
```diff
@@ -75,7 +75,7 @@ export class ValueAndUnit {
      * @returns the value
      */
     public getValue(host: AdvancedDynamicTexture): number {
-        if (host && !this.ignoreAdaptiveScaling && this.unit !== ValueAndUnit.UNITMODE_PERCENTAGE) {
+        if (host && !this.ignoreAdaptiveScaling && this.unit !== ValueAndUnit.UNITMODE_PERCENTAGE && host.renderAtIdealSize) {
```
deltakosh, line 78: This seems incorrect as ideadlWith and height can still be used: when renderAtIdealSize = false;
https://doc.babylonjs.com/divingDeeper/gui/gui#adaptive-scaling

renderAtIdealSize is here to force the texture to rescale always at the expected ideal size

gui/src/2D/advancedDynamicTexture.ts
msDestiny14: This line was changing the render w/h but then it appearing to be re-accounted for other file. Calculating it twice?

deltakosh: but this needs to happen as the goal of renderAtIdealSize is precisely to do that scale (check the doc link I shared)

msDestiny14:
@deltakosh here is the new version :) @RaananW recommended making a visual test case for this but Having some local build issues at the moment. Prioritizing the fix.

deltakosh (approved):
Looks good!

msDestiny14:
wait @deltakosh does it look like I'm taking @Popov72 's changes from somewhere too? I pull from master before committing. Bad merge?

PR #10048, opened 2021-03-12
do not dispose anything parented to a handtracking sphere when disposing

Palmer-JC:


Palmer-JC:
I have put in an override to my mesh sub-class for an Arm button panel.  Toggles, sliders & dials hold state, which I do not want to be required to have to mess with if someone leaves & comes back.

```javascript
/**
 * @override
 * Change to ignore
 *
 * @param {boolean} doNotResurce - Do children as well
 */
public dispose(doNotRecurse : boolean) : void {}

/**
 * never called.
 */
public actuallyDispose(doNotRecurse : boolean) : void {
    super.dispose(doNotRecurse);
}
```

So not hurry, but also most people do not have mesh source code generation, so this is not so easily available.

RaananW (approved):
LGTM. And makes sense when it comes to the spheres themselves.

PR #10061, opened 2021-03-16
NME: Fix the strength input parameter of the `PerturbNormal` block

Popov72:
See https://forum.babylonjs.com/t/strange-behaviours-in-node-material-normals/19216

This parameter was passed as `1/strength` to the shader instead of simply `strength`.

### DIFF
--- a/src/Materials/Node/Blocks/Fragment/perturbNormalBlock.ts
+++ b/src/Materials/Node/Blocks/Fragment/perturbNormalBlock.ts
@@ -151,7 +151,7 @@ export class PerturbNormalBlock extends NodeMaterialBlock {
 
         state._emitUniformFromString(this._tangentSpaceParameterName, "vec2");
 
-        let replaceForBumpInfos = this.strength.isConnectedToInputBlock && this.strength.connectInputBlock!.isConstant ? `${state._emitFloat(1.0 / this.strength.connectInputBlock!.value)}` : `1.0 / ${this.strength.associatedVariableName}`;
+        let replaceForBumpInfos = this.strength.isConnectedToInputBlock && this.strength.connectInputBlock!.isConstant ? `${state._emitFloat(this.strength.connectInputBlock!.value)}` : `${this.strength.associatedVariableName}`;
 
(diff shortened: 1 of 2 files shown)


src/Materials/Node/Blocks/Fragment/perturbNormalBlock.ts
```diff
@@ -151,7 +151,7 @@ export class PerturbNormalBlock extends NodeMaterialBlock {
 
         state._emitUniformFromString(this._tangentSpaceParameterName, "vec2");
 
-        let replaceForBumpInfos = this.strength.isConnectedToInputBlock && this.strength.connectInputBlock!.isConstant ? `${state._emitFloat(1.0 / this.strength.connectInputBlock!.value)}` : `1.0 / ${this.strength.associatedVariableName}`;
+        let replaceForBumpInfos = this.strength.isConnectedToInputBlock && this.strength.connectInputBlock!.isConstant ? `${state._emitFloat(this.strength.connectInputBlock!.value)}` : `${this.strength.associatedVariableName}`;
```
deltakosh, line 154: This was done that way because the shader code expect 1/level:
image

deltakosh: Ignore me. I read the forum ;)

deltakosh (approved):

src/Materials/Node/Blocks/Fragment/perturbNormalBlock.ts
Popov72: The PBR material is passing the level "as is" whereas the `StandardMaterial` is passing `1/level`: the value is not used at the same place in the shader code.

Popov72: However, we do have a problem because if the block is not used in connection with the PBR block, it will fail with my changes...

deltakosh: Oh yeah this is bad..any idea?

PR #10077, opened 2021-03-18
UBO matrix cache fix

sebavan:
Let s wait on the tests to be sure

### DIFF
--- a/src/Materials/uniformBuffer.ts
+++ b/src/Materials/uniformBuffer.ts
@@ -705,12 +705,13 @@ export class UniformBuffer {
     private _cacheMatrix(name: string, matrix: IMatrixLike): boolean {
         this._checkNewFrame();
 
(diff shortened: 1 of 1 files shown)


deltakosh (approved):

src/Materials/uniformBuffer.ts
```diff
@@ -705,12 +705,13 @@ export class UniformBuffer {
...
 
-        var cache = this._valueCache[name];
-        var flag = matrix.updateFlag;
+        const cache = this._valueCache[name];
+        const flag = matrix.updateFlag;
         if (cache !== undefined && cache === flag) {
             return false;
         }
 
+        this._valueCache[name] = flag;
```
Popov72, line 714: Good catch! It seems I simplified a bit too much when I reorganized the code end of last year...

PR #10089, opened 2021-03-22
Shadows: improve performance by using mesh and scene UBOs

Popov72:
Some numbers (PG Powerplant with CSM: https://playground.babylonjs.com/#SYQW69#842):
* WebGL2: from 115 fps to 145 fps - we also save some GPU time, but it's hard to give some figures because it changes a lot from run to run, and even inside a run, sometimes a frame takes 2ms sometimes it takes 13ms... However, it seems we save something like between 1.5 and 2.5ms with the changes
* WebGPU: from 95 fps to 125 fps - can't give numbers for saved GPU time as instrumentation is not implemented yet in Babylon (and PIX does not work anymore for timings...)

Made some improvements by also removing unneeded bindings of matrices (we only need the `world` and `viewProjection` matrices).

I could also remove quite some code from the `ShadowDepthWrapper` class that was handling the NME matrix updates because the corresponding uniforms are filled from the scene view/projection matrices which are set by the shadow generator at the start of the process in `onBeforeRenderObservable`.

I fixed a bug in WebGPU with the UBO binding cache that was done in the `Effect` class when calling `bindUniformBufferBase`: I have moved this code in the `ThinEngine` class instead, so that there's no caching taking place for WebGPU.

I thought I needed to support the scene/mesh ubos in `ShaderMaterial` so I have added this support, but it turned out I didn't need it... I have kept the code, as it can still be useful on its own.

As for the `Tools.FloatRound` calls added in the `UniformBuffer` class, it is a bug fix: we were generating more UBO updates than necessary because of a float 32 / 64 bits comparison mismatch.


deltakosh:
OMG this is quite a fantastic update buddy!

dist/preview release/what's new.md
```diff
@@ -24,6 +24,7 @@
 - Added a [helper class](https://doc.babylonjs.com/typedoc/classes/babylon.debug.directionallightfrustumviewer) to display the frustum of a directional light ([Popov72](https://github.com/Popov72))
 - Improved collision detection performance ([ottoville](https://github.com/ottoville/))
 - Added new helper functions for Quaternion.FromLookDirection and Matrix.LookDirection ([Alex-MSFT](https://github.com/Alex-MSFT))
+- Improved performance when using the shadow / cascacaded shadow generator ([Popov72](https://github.com/Popov72))
```
deltakosh, line 27: please move it to main features. This is massive

deltakosh (approved):

dist/preview release/what's new.md
Popov72: Done!
I also needed to change a bit the UBO binding cache problem because of a circular dependency between `Effect` and `ThinEngine`.

PR #10096, opened 2021-03-24
KHR_materials_transmission - Make roughness affect thin surface refraction again

MiiBond:


sebavan:
Would be great to not use branch with if when the value comes out from texture ?

MiiBond:
@sebavan Indeed. I realized that at lunch today. I'll change it when I get a chance to work on this again. There is at least one more thing I need to do.

MiiBond:
Okay, so this is fixed now.
However, I’m wondering if we need the volume ior VS surface ior logic at all. We can just use the thickness check to see if the material is volumetric or not.
Originally, I think I separated the two IOR's so that we could have IOR affect the reflections and refractions independently but I don't think that's really necessary.
This might change existing scenes that use refraction with a thickness of 0 though?

sebavan:
I would prefer to avoid breaking changes but let s see what @Popov72 and @bghgary think ?

bghgary:
I told @MiiBond to check with you, haha. From a glTF perspective, there is no change.

src/Materials/Node/Blocks/PBR/refractionBlock.ts
```diff
@@ -233,7 +233,7 @@ export class RefractionBlock extends NodeMaterialBlock {
 
         effect.setFloat4(this._vRefractionInfosName, refractionTexture.level, 1 / indexOfRefraction, depth, this.invertRefractionY ? -1 : 1);
 
-        effect.setFloat3(this._vRefractionMicrosurfaceInfosName, refractionTexture.getSize().width, refractionTexture.lodGenerationScale, refractionTexture.lodGenerationOffset);
+        effect.setFloat4(this._vRefractionMicrosurfaceInfosName, refractionTexture.getSize().width, refractionTexture.lodGenerationScale, refractionTexture.lodGenerationOffset, indexOfRefraction);
```
Popov72, line 236: I think it's `1 / indexOfRefraction` instead?

MiiBond: Quite right. Thank you.

sebavan (changes requested):

sebavan (approved):

src/Materials/PBR/pbrSubSurfaceConfiguration.ts
```diff
@@ -353,6 +355,7 @@ export class PBRSubSurfaceConfiguration {
                     }
                 }
 
+                defines.SS_HAS_THICKNESS = this.minimumThickness !== 0.0 && this.maximumThickness !== 0.0;
```
Popov72, line 358: Why not doing `defines.SS_HAS_THICKNESS = (this.maximumThickness - this.minimumThickness) !== 0.0;` instead?

I'm not sure that `this.minimumThickness !== 0.0` should be a requirement(?)

Popov72 (approved):

src/Materials/PBR/pbrSubSurfaceConfiguration.ts
MiiBond: Quite right.

Popov72:
I think this one can be merged?

Issue #10087, opened 2021-03-22
Simultaneous button presses not working properly


PolygonalSun:
In the PG below, there are two cubes and when a mouse button is pressed, a corresponding cube will rotate.  Both should rotate at the same time when the left and right are pressed but in 5.0, only one will rotate at a time.

**Repro**
https://playground.babylonjs.com/#T4MJ6I#5

In 5.0, only one cube will rotate.
In 4.2, both cubes will move when the left and right mouse buttons are pressed

Forum post for reference: https://forum.babylonjs.com/t/detect-simultaneously-pressed-mousebuttons/6278

sebavan:
Closed by https://github.com/BabylonJS/Babylon.js/pull/10098

PR #10098, opened 2021-03-24
Fix Simultaneous button click

sebavan:
https://forum.babylonjs.com/t/detect-simultaneously-pressed-mousebuttons/6278/7


sebavan:
Fix #10087

src/DeviceInput/deviceInputSystem.ts
```diff
@@ -355,17 +355,29 @@ export class DeviceInputSystem implements IDisposable {
                 pointer[PointerInput.DeltaVertical] = evt.movementY;
 
                 if (this.onInputChanged) {
+                    // The browser might use a move event in case
+                    // of simultaneous mouse buttons click for instance. So
+                    // in this case we stil need to propagate it.
+                    let fireFakeMove = true;
```
deltakosh, line 361: does it work if you click AND move simulteanously? Don't we swallow one in this case?

sebavan: not that I have seen so far let me double check

sebavan: nope all good

PR #10106, opened 2021-03-29
SkyMatrial: fix cameraOffset serialize

happyPaner:
SkyMatrial: fix cameraOffset serialize

### DIFF
--- a/materialsLibrary/src/sky/skyMaterial.ts
+++ b/materialsLibrary/src/sky/skyMaterial.ts
@@ -110,7 +110,7 @@ export class SkyMaterial extends PushMaterial {
      * Defines an offset vector used to get a horizon offset.
      * @example skyMaterial.cameraOffset.y = camera.globalPosition.y // Set horizon relative to 0 on the Y axis
      */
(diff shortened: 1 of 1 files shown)


julien-moreau:
LGTM! @happyPaner anyway can you test that when you parse a sky material serialized with the old decorator then parsing with the new decorator still works? 
Else, we’ll have to create the vector 3 manually in the .Parse function to keep the backward compatibility
Thanks ! :)

happyPaner:
Hi! @julien-moreau, yes. I has tested. it's a bug. When i parse a sky material serialized with the old decorator get a wrong result. Use new decorator, get right result.

deltakosh:
Any update?

julien-moreau:
@deltakosh it's ok for me, and it looks like Chibi Rex and Yeti valitation tests are not using the Sky Material. I don't understand why this pipeline is failing

deltakosh:
Let me run that again :)

julien-moreau:
CI looks good now @deltakosh :)

PR #10108, opened 2021-03-30
Enabled near hover, pick and grab for WebXRControllerPointerSelection 

satyapoojasama:
Introduced the following to the webXRControllerPointerSelection code:
nearPick
nearHover 
nearGrab
pick PickingInfo (it is now common for both near and far interaction)


src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -344,6 +360,35 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
                 controllerData.xrController.getWorldPointerRayToRef(controllerData.tmpRay);
+                const hand = controllerData.xrController.inputSource.hand;
+                if(hand)
+                {
+                    const xrJoint = hand.get("index-finger-tip");
+                    if(xrJoint)
+                    {
+                        let pose = _xrFrame.getJointPose!(xrJoint, this._xrSessionManager.referenceSpace);
+                        if (!pose || !pose.transform) {
+                            return;
```
rickfromwork, line 371: Probably shouldn't return here. We can still do far interaction processing with no joint data.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,9 +411,33 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                pick = this._scene.pickWithRay(controllerData.tmpRay, this._scene.pointerMovePredicate || this.raySelectionPredicate);
+                if (pick && pick.pickedPoint && pick.hit) {
+                    // Update laser state
+                    this._updatePointerDistance(controllerData.laserPointer, pick.distance);
+                }
+            }
+     
+            controllerData.pick = pick;
 
             if (pick && pick.pickedPoint && pick.hit) {
```
rickfromwork, line 442: There's some logic in here specific to the laser that shows up for far interaction that we shouldn't be doing if we have a near interaction.

src/Meshes/abstractMesh.ts
```diff
@@ -350,6 +350,8 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
      */
     public isPickable = true;
 
+    public isNearInteractable = false;
```
rickfromwork, line 353: nit: I think isNearPickable is more aligned with the other predicate boolean, isPickable.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,9 +411,33 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                     scene.pointerY = this._screenCoordinatesRef.y;
                 }
             }
-            controllerData.pick = this._scene.pickWithRay(controllerData.tmpRay, this._scene.pointerMovePredicate || this.raySelectionPredicate);
 
-            const pick = controllerData.pick;
+            let hover = this._pickWithMesh(controllerData.hoverMesh, false);
+            let nearHover = hover && hover.pickedPoint && hover.hit;
+            if(nearHover)
```
rickfromwork, line 417: If we're in hover range, we should probably commit fully to the near interaction model. The way its laid out here, we can send hover info, but still do a far interaction, which creates interactions on potentially two different objects at the same time (not supported by the BJS pointer system/controls)

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -344,6 +360,35 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                        if (!pose || !pose.transform) {
+                            return;
+                        }
+                        const pos = pose.transform.position;
+                        const originalMesh = SphereBuilder.CreateSphere("ParentSphere", { diameter: 1 });
+                        originalMesh.scaling.set(0.01, 0.01, 0.01);
+
+                        // populate information for near pick
+                        const pickRadius = (pose.radius || 0.008)/2;
+                        controllerData.pickMesh = originalMesh.createInstance("IndexPickSphere");
```
rickfromwork, line 379: I know the webXRHandTracker holds information about the joint radius somewhere... Can't we just use a reference to the index tip joint mesh itself here, and the radius information that comes with it?

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: good catch, i was thinking about it and slipped my mind

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: so, it should be added?

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: love this idea, not sure how to do that though. let me dig some and see

satyapoojasama: we are using that radius information infact. "pose.radius"

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,9 +408,42 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
-            const pick = controllerData.pick;
+            let pick = null;
+            if(controllerData.hoverMesh)
+            {
+                let hoverInfo = this._pickWithMesh(controllerData.hoverMesh, false);
+                let nearHover = hoverInfo && hoverInfo.pickedPoint && hoverInfo.hit;
+                if(nearHover)
+                {
+                    console.log("HOVER SUCCESS");
+                    controllerData.hover = hoverInfo;
```
satyapoojasama, line 420: currently hover pick information is being populated in a different property "controllerData.hover". this is different from "controllerData.pick" that is used by pointermove events. Should i update hover to even use "controllerData.pick"  instead?

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: moved it to only apply for far interaction

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -344,6 +368,28 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                        if (indexTipPose && indexTipPose.transform) {
+                            let zAxisMultiplier = 1;
+                            if(!this._scene.useRightHandedSystem)
+                            {
+                                zAxisMultiplier = -1;
+                            }
+                            const indexTipPos = indexTipPose.transform.position;
+                            // set positions for near pick and hover
+                            if(controllerData.pickIndexMesh) {
+                                controllerData.pickIndexMesh.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z*zAxisMultiplier));
```
rickfromwork, line 385: nit - spacing around the *

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -344,6 +368,28 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
                 controllerGlobalPosition = controllerData.xrController.pointer.position;
                 controllerData.xrController.getWorldPointerRayToRef(controllerData.tmpRay);
+                const hand = controllerData.xrController.inputSource.hand;
+                if(hand) {
+                    const xrIndexTip = hand.get("index-finger-tip");
+                    if(xrIndexTip) {
+                        let indexTipPose = _xrFrame.getJointPose!(xrIndexTip, this._xrSessionManager.referenceSpace);
+                        if (indexTipPose && indexTipPose.transform) {
+                            let zAxisMultiplier = 1;
+                            if(!this._scene.useRightHandedSystem)
```
rickfromwork, line 378: This can also be one-lined as the following if you want:
let zAxisMultiplier = this._scene.useRightHandedSystem ? 1 : -1;

src/XR/features/WebXRControllerPointerSelection.ts
rickfromwork: There's still ray-specific logic being called no matter what below

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -582,13 +661,45 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                 }
             };
 
+            const squeezeStartListener = (event: XRInputSourceEvent) => {
+                if(event.inputSource.hand)
```
rickfromwork, line 665: nit - conditionals should have spacing before the '(' here and elsewhere

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +412,47 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
 
-            const pick = controllerData.pick;
+            let pick = null;
+            let hoverAtOrigin = false;
+            if(controllerData.hoverIndexMesh) {
+                let hoverInfo = this._pickWithMesh(controllerData.hoverIndexMesh, false, this.nearPickPredicate || this.nearGrabPredicate);
+                let nearHover = hoverInfo && hoverInfo.pickedPoint && hoverInfo.hit;
+                if(hoverInfo?.pickedPoint)
+                {
+                    hoverAtOrigin = hoverInfo.pickedPoint.x == 0 && hoverInfo.pickedPoint.y == 0 && hoverInfo.pickedPoint.z == 0;
```
rickfromwork, line 423: What situation causes this to be true?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +412,47 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                {
+                    pickAtOrigin = pickInfo.pickedPoint.x == 0 && pickInfo.pickedPoint.y == 0 && pickInfo.pickedPoint.z == 0;
+                }
+                if(nearPick && !pickAtOrigin) {
+                    controllerData.nearPick = true;
+                    pick = pickInfo;
+                }
+            }
 
+            if (!controllerData.nearHover && !controllerData.nearPick && !controllerData.nearGrab) {
```
rickfromwork, line 445: Where are these values ever reset?

rickfromwork:
Do you have any plans to open up onXRFrame to support both hands in near interaction? Right now it looks like the check at the start of the function limits usage to the dominant device/hand, since Babylon.JS traditionally supports just one pointer.

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: this happens right at the moment the hand tip spheres are created( they are created at origin and then moved to the position of index tip on OnXRFrame above) and if there is a  mesh that happens to have nearinteraction turned on at origin, this is true automatically even if there was no hover or pick. As we know the hand tip spheres are created every time the hand goes out of the scene and comes back in, this does happen quite a few times

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: very good catch! probably they should be reset at the beginning o OnXRFrame. Currently they are not reset.

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: I see, i did not know selection Mesh is only supposed to be populated for ray. Do you think none of this applies to near interaction at all?

deltakosh:
I feel like this PR has too many file change :)

deltakosh:
For instance all dist folder should not be pushed

satyapoojasama:
> For instance all dist folder should not be pushed

You are right, my latest push should not have contained all that. i think there is something wrong with it, lemme fix it!

PR #10121, opened 2021-04-01
WebGPU: perf optimization & misc

Popov72:
(sorry for the big PR!)

* added time query to WebGPU to have meaningful values in "GPU Frame Time" in the inspector
* added `engine.compatibilityMode = false` ("fast path") and `engine.snapshotRendering/engine.snapshotRenderingMode` as ways to improve perfs:
  * `engine.compatibilityMode = false`: will create a bundle the first time a draw is done and will be reused afterwards. Preliminary work
  * `engine.snapshotRendering = true` enables a mode where the GPU commands of the next frame are recorded in bundle lists and replayed in subsequent frames. When `engine.snapshotRenderingMode=1` we also don't execute some JS code, like the `_evaluateActiveMeshes` and the mesh `render` methods. This is meant to improve perf in mostly static scenes (like the forest demo, where fps > 60fps - GPU bound)
* did some small changes to improve perf in the legacy mode ("slow path")
* added `alphaCutOff` and `vAmbientColor` to the ubo for the standard material, so that all standard parameters of the material are in the ubo and not in the LeftOver ubo
* made some changes in shadow generator/CSM + `ShadowDepthWrapper` to be able to use one `DrawWrapper` per cascade in CSM mode (to enable fast path to work with CSM and `ShadowDepthWrapper`) and one `DrawWrapper` per cube face for point lights (in standard shadow generator)
* added a `checkUbosContentBeforeUpload` feature. When true, it checks that the current content of the ubo is really different from the content of the corresponding GPU buffer before uploading. We need this in WebGPU to avoid having a lot of unncessary uploads (that WebGL does not have because Angle is doing some smart job about this)
* fixed a bug where `ShadowGenerator.bindShadowLight` was called for a material even if the mesh this material was applied to was not a shadow receiver
* synced with latest WebGPU spec


src/Engines/Extensions/engine.occlusionQuery.ts
```diff
@@ -271,6 +296,47 @@ Engine.prototype.endTimeQuery = function(token: _TimeToken): int {
...
+    if (value === this._captureGPUFrameTime) {
+        return;
+    }
+
+    this._captureGPUFrameTime = value;
+
+    if (value) {
+        this._onBeginFrameObserver = this.onBeginFrameObservable.add(() => {
+            if (!this._gpuFrameTimeToken) {
+                this._gpuFrameTimeToken = this.startTimeQuery();
```
deltakosh, line 316: does it work when using occlusion query? I was under the impression that only one query was available at time

src/Engines/constants.ts
```diff
@@ -594,4 +594,10 @@ export class Constants {
      * Constant used as key code for Shift key
      */
     public static readonly INPUT_SHIFT_KEY = 16;
+
+    /** Standard snapshot rendering. In this mode, some form of dynamic behavior is possible (for eg, uniform buffers are still updated) */
+    public static readonly SNAPSHOTRENDERING_STANDARD = 0;
+
+    /** Fast snapshot rendering. In this mode, everything is static and only some limited form of dynamic behaviour is possible */
+    public static readonly SNAPSHOTRENDERING_FAST = 1;
```
deltakosh, line 602: will this be enabled when users call scene.frezzeActiveMeshes?

Popov72: No, it's independent from freezeActiveMeshes. It's activated when setting `engine.snapshotRendering = true`.

src/Engines/Extensions/engine.occlusionQuery.ts
Popov72: Not occlusion query but timer query. But it was already there before :) In this file, there were occlusion + time query functions

src/Engines/constants.ts
deltakosh: do you think it is an option to do i? Does it make sense?

src/Engines/Extensions/engine.occlusionQuery.ts
deltakosh: oh right..I'm a dick with file names

can you rename the file name to something simply like engine.query ?

src/Engines/constants.ts
Popov72: No, I think it's really two different things, you could want freezing active meshes without enabling snapshot rendering as snapshot rendering have more constraints than freezing meshes (meaning, when snapshot rendering is enabled a number of things won't work as expected if you do them that would work if you only froze active meshes).

Popov72:
Compilation fails with:
image

I have added `es2020.bigint` in the ts lib, I guess it's what makes the compil fail but I have no idea why.

Maybe @sebavan you know?

Popov72:
Everything is ok except for these errors in "Unit tests" and "Module tests":
```
/home/vsts/work/1/s/dist/preview release/babylon.module.d.ts(86999,18): error TS2300: Duplicate identifier 'Observer'.
/home/vsts/work/1/s/dist/preview release/babylon.module.d.ts(145969,10): error TS2300: Duplicate identifier 'Observer'.
```

Popov72:
Ok, problems fixed, I switched the PR to "Ready for review"

deltakosh:
Waiting for @sebavan approval

src/scene.ts
```diff
@@ -763,6 +763,54 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
 
+    /**
+     * Bind the current view position to an effect.
+     * @param effect The effect to be bound
+     * @param scene The scene the eyes position is used from
+     * @param variableName name of the shader variable that will hold the eye position
+     * @param isVector3 true to indicates that variableName is a Vector3 and not a Vector4
+     * @return the computed eye position
+     */
+     public static BindEyePosition(effect: Nullable<Effect>, scene: Scene, variableName = "vEyePosition", isVector3 = false): Vector4 {
```
sebavan, line 774: Why static if we pass the scene in ?

materialsLibrary/src/cell/cellMaterial.ts
```diff
@@ -263,7 +263,7 @@ export class CellMaterial extends PushMaterial {
                 this._activeEffect.setFloat("pointSize", this.pointSize);
             }
 
-            MaterialHelper.BindEyePosition(effect, scene);
+            Scene.BindEyePosition(effect, scene);
```
sebavan, line 266: Why static if it is on scene ? I ll put the comment in scene as well

src/Engines/WebGPU/webgpuBundleList.ts
```diff
@@ -0,0 +1,182 @@
...
+        this._currentItemIsBundle = false;
+        this.numDrawCalls = 0;
+    }
+
+    public clone(): WebGPUBundleList {
+        this.close();
+
+        const cloned = new WebGPUBundleList(this._device);
+
+        cloned._list = [];
```
sebavan, line 172: Could we init the array at the right size ? (prevents over allocation when it grows)

src/Materials/material.ts
```diff
@@ -992,7 +992,7 @@ export class Material implements IAnimatable {
         if (this._needToBindSceneUbo) {
             if (effect) {
                 this._needToBindSceneUbo = false;
-                MaterialHelper.FinalizeSceneUbo(this.getScene());
+                Scene.FinalizeSceneUbo(this.getScene());
```
sebavan, line 995: Same here for the static ?

src/scene.ts
```diff
@@ -763,6 +763,54 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
+
+        return TmpVectors.Vector4[0];
+    }
+
+    /**
+     * Update the scene ubo before it can be used in rendering processing
+     * @param scene the scene to retrieve the ubo from
+     * @returns the scene UniformBuffer
+     */
+    public static FinalizeSceneUbo(scene: Scene): UniformBuffer {
```
sebavan, line 800: Same static comment but not sure why it had been moved to scene ?

sebavan (commented):
Sounds goood aside the static move to Scene ?

src/scene.ts
Popov72: I preferred to retain the exact same signature to avoid too many breaking changes, but the move is a breaking change itself so you're right it's better to just make it non static. Going to perform the change.

src/scene.ts
Popov72: Because of circular dependency problems. I had to move it along with `BindEyePosition` as the latter is calling `FinalizeSceneUbo`.

materialsLibrary/src/cell/cellMaterial.ts
Popov72: Not static anymore

src/Engines/WebGPU/webgpuBundleList.ts
Popov72: Done

src/Materials/material.ts
Popov72: Not static anymore

sebavan:
> Compilation fails with:
> image
> 
> I have added `es2020.bigint` in the ts lib, I guess it's what makes the compil fail but I have no idea why.
> 
> Maybe @sebavan you know?
> 
> [...] Trying to add "es2015.symbol.wellknown", let's see how it comes...

No Idea :-) Let s check offline

sebavan (approved):

Popov72:
> No Idea :-) Let s check offline

That's ok, there's no error anymore.

Popov72:
If ok, can this be merged?

PR #10125, opened 2021-04-02
Allow DOM style transparency for containers

blindinghues:
By rendering children to an intermediate texture, containers can have their alpha function like in the DOM. Opaque children can overlap, and their parent container transparency set, without darkened seams appearing in the overlapping region.

https://forum.babylonjs.com/t/babylongui-container-transparency/19177/2
(my use case explained, and an example of the darkened seams [here](https://playground.babylonjs.com/#XCPP9Y#6346))

This is a bit of a niche case, but I believe the DOM style of alpha blending is quite universal, and should at least be an option. 

image

With this change I'm finally able to have split up SVG image elements overlap without issue for the GUI in my upcoming game.

### DIFF
--- a/gui/src/2D/controls/container.ts
+++ b/gui/src/2D/controls/container.ts
@@ -37,7 +37,7 @@ export class Container extends Control {
         return this._renderToIntermediateTexture;
     }
     public set renderToIntermediateTexture(value: boolean) {
-        if (this._renderToIntermediateTexture == value) {
+        if (this._renderToIntermediateTexture === value) {
             return;
         }
         this._renderToIntermediateTexture = value;
--- a/gui/src/2D/controls/container.ts
+++ b/gui/src/2D/controls/container.ts
@@ -27,9 +27,22 @@ export class Container extends Control {
     /** @hidden */
     protected _adaptHeightToChildren = false;
     /** @hidden */
-    protected _intermediateTexture: Nullable<DynamicTexture> = null;
+    protected _renderToIntermediateTexture: boolean = false;
     /** @hidden */
-    public _renderToIntermediateTexture: boolean = false;
+    protected _intermediateTexture: Nullable<DynamicTexture> = null;
+
+    /** Gets or sets boolean indicating if children should be rendered to an intermediate texture rather than directly to host, useful for alpha blending */
+    @serialize()
+    public get renderToIntermediateTexture(): boolean {
+        return this._renderToIntermediateTexture;
+    }
+    public set renderToIntermediateTexture(value: boolean) {
+        if (this._renderToIntermediateTexture == value) {
+            return;
+        }
+        this._renderToIntermediateTexture = value;
+        this._markAsDirty();
+    }
 
     /**
      * Gets or sets a boolean indicating that layout cycle errors should be displayed on the console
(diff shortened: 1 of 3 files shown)


gui/src/2D/controls/container.ts
```diff
@@ -98,9 +105,11 @@ export class Container extends Control {
     /**
      * Creates a new Container
      * @param name defines the name of the container
+     * @param renderToIntermediateTexture renders children to an intermediate texture rather than directly to host, useful for alpha blending
      */
-    constructor(public name?: string) {
+    constructor(public name?: string, renderToIntermediateTexture?: boolean) {
         super(name);
+        this._renderToIntermediateTexture = renderToIntermediateTexture ? true : false;
```
deltakosh, line 112:
```suggestion
        this._renderToIntermediateTexture = !!renderToIntermediateTexture;

```

deltakosh (approved):
Like it a lot! Great contrib!

gui/src/2D/controls/container.ts
```diff
@@ -23,6 +26,10 @@ export class Container extends Control {
     protected _adaptWidthToChildren = false;
     /** @hidden */
     protected _adaptHeightToChildren = false;
+    /** @hidden */
+    protected _intermediateTexture: Nullable<DynamicTexture> = null;
+    /** @hidden */
+    public _renderToIntermediateTexture: boolean = false;
```
Popov72, line 32: Can't you remove the leading underscore (and the `hidden` modifier) and let the property be updatable by the user at any point? I think with the current code it should work?

gui/src/2D/controls/container.ts
```diff
@@ -387,11 +409,22 @@ export class Container extends Control {
 
     /** @hidden */
     public _draw(context: CanvasRenderingContext2D, invalidatedRectangle?: Measure): void {
+        let contextToDrawTo: CanvasRenderingContext2D = (this._intermediateTexture) ? this._intermediateTexture.getContext() : context;
```
Popov72, line 412: nit: `const` instead of `let`

gui/src/2D/controls/container.ts
```diff
@@ -484,6 +525,9 @@ export class Container extends Control {
         for (var index = this.children.length - 1; index >= 0; index--) {
             this.children[index].dispose();
         }
+        if (this._intermediateTexture) {
+            this._intermediateTexture.dispose();
+        }
```
Popov72, line 530: You can simplify the 3 lines into:
```typescript
this._intermediateTexture?.dispose();
```

blindinghues: Short'n' sweet, thanks

gui/src/2D/controls/container.ts
blindinghues: Needed to make a slight change, as if we changed this property to false as it was we would still render to the intermediate texture if it had been created

gui/src/2D/controls/container.ts
blindinghues: Changed this to work with the above point

gui/src/2D/controls/container.ts
blindinghues: I removed this from the constructor, and made it a settable property after construction. Is this perhaps a better approach?

blindinghues:
Addressed reviews, thanks for the feedback :)

gui/src/2D/controls/container.ts
```diff
@@ -27,9 +27,22 @@ export class Container extends Control {
...
-    public _renderToIntermediateTexture: boolean = false;
+    protected _intermediateTexture: Nullable<DynamicTexture> = null;
+
+    /** Gets or sets boolean indicating if children should be rendered to an intermediate texture rather than directly to host, useful for alpha blending */
+    @serialize()
+    public get renderToIntermediateTexture(): boolean {
+        return this._renderToIntermediateTexture;
+    }
+    public set renderToIntermediateTexture(value: boolean) {
+        if (this._renderToIntermediateTexture == value) {
```
Popov72, line 40: nit: ===

Popov72:
Only a nit and LGTM!

You should update the what's new file, however.

PR #10137, opened 2021-04-05
add slider-3d gui control

binyammathewos:
Added a 3D slider GUI control

### DIFF
--- a/src/Engines/WebGPU/webgpuCacheRenderPipeline.ts
+++ b/src/Engines/WebGPU/webgpuCacheRenderPipeline.ts
@@ -219,11 +219,14 @@ export abstract class WebGPUCacheRenderPipeline {
     protected abstract _getRenderPipeline(param: { token: any, pipeline: Nullable<GPURenderPipeline> }): void;
     protected abstract _setRenderPipeline(param: { token: any, pipeline: Nullable<GPURenderPipeline> }): void;
 
-    public vertexBuffers: VertexBuffer[];
+    public readonly vertexBuffers: VertexBuffer[];
 
     public get colorFormats(): GPUTextureFormat[] {
         return this._mrtAttachments1 > 0 ? this._mrtFormats : this._webgpuColorFormat;
     }
+
+    public readonly mrtAttachments: number[];
+    public readonly mrtTextureArray: InternalTexture[];
 
     public getRenderPipeline(fillMode: number, effect: Effect, sampleCount: number): GPURenderPipeline {
         if (this.disabled) {
@@ -351,6 +354,9 @@ export abstract class WebGPUCacheRenderPipeline {
             // so we can encode 5 texture formats in 32 bits
             throw "Can't handle more than 10 attachments for a MRT in cache render pipeline!";
         }
+        (this.mrtAttachments as any) = attachments;
+        (this.mrtTextureArray as any) = textureArray;
+
         let bits: number[] = [0, 0], indexBits = 0, mask = 0, numRT = 0;
         for (let i = 0; i < attachments.length; ++i) {
             const index = attachments[i];
--- a/src/Engines/WebGPU/webgpuClearQuad.ts
+++ b/src/Engines/WebGPU/webgpuClearQuad.ts
@@ -1,4 +1,5 @@
 import { Effect } from "../../Materials/effect";
+import { InternalTexture } from "../../Materials/Textures/internalTexture";
 import { IColor4Like } from "../../Maths/math.like";
 import { VertexBuffer } from "../../Meshes/buffer";
 import { WebGPUDataBuffer } from "../../Meshes/WebGPU/webgpuDataBuffer";
@@ -17,13 +18,20 @@ export class WebGPUClearQuad {
     private _cacheRenderPipeline: WebGPUCacheRenderPipeline;
     private _effect: Effect;
     private _bindGroups: { [id: number]: GPUBindGroup[] } = {};
+    private _depthTextureFormat: GPUTextureFormat | undefined;
+    private _bundleCache: { [key: number]: GPURenderBundle } = {};
 
     public setDepthStencilFormat(format: GPUTextureFormat | undefined): void {
+        this._depthTextureFormat = format;
         this._cacheRenderPipeline.setDepthStencilFormat(format);
     }
 
     public setColorFormat(format: GPUTextureFormat): void {
         this._cacheRenderPipeline.setColorFormat(format);
+    }
+
+    public setMRTAttachments(attachments: number[], textureArray: InternalTexture[]): void {
+        this._cacheRenderPipeline.setMRTAttachments(attachments, textureArray);
     }
 
     constructor(device: GPUDevice, engine: WebGPUEngine, emptyVertexBuffer: VertexBuffer) {
@@ -37,7 +45,33 @@ export class WebGPUClearQuad {
         this._effect = engine.createEffect("clearQuad", [], ["color", "depthValue"]);
     }
 
-    public clear(renderPass: GPURenderPassEncoder, clearColor?: Nullable<IColor4Like>, clearDepth?: boolean, clearStencil?: boolean, sampleCount = 1) {
+    public clear(renderPass: Nullable<GPURenderPassEncoder>, clearColor?: Nullable<IColor4Like>, clearDepth?: boolean, clearStencil?: boolean, sampleCount = 1): Nullable<GPURenderBundle> {
+        let renderPass2: GPURenderPassEncoder | GPURenderBundleEncoder;
+        let bundle: Nullable<GPURenderBundle> = null;
+        let bundleKey = 0;
+
+        if (renderPass) {
+            renderPass2 = renderPass;
+        } else {
+            bundleKey = (clearColor ? clearColor.r + clearColor.g * 256 + clearColor.b * 256 * 256 + clearColor.a * 256 * 256 * 256 : 0) +
+                (clearDepth ? 2 ** 32 : 0) +
+                (clearStencil ? 2 ** 33 : 0) +
+                (this._engine.useReverseDepthBuffer ? 2 ** 34 : 0) +
+                sampleCount * (2 ** 35);
+
+            bundle = this._bundleCache[bundleKey];
+
+            if (bundle) {
+                return bundle;
+            }
+
+            renderPass2 = this._device.createRenderBundleEncoder({
+                colorFormats: this._cacheRenderPipeline.colorFormats,
+                depthStencilFormat: this._depthTextureFormat,
+                sampleCount,
+            });
+        }
+
         this._cacheRenderPipeline.setDepthWriteEnabled(!!clearDepth);
         this._cacheRenderPipeline.setStencilEnabled(!!clearStencil);
         this._cacheRenderPipeline.setStencilWriteMask(clearStencil ? 0xFF : 0);
@@ -85,11 +119,17 @@ export class WebGPUClearQuad {
             );
         }
 
-        renderPass.setStencilReference(this._engine._clearStencilValue);
-        renderPass.setPipeline(pipeline);
+        renderPass2.setPipeline(pipeline);
         for (let i = 0; i < bindGroups.length; ++i) {
-            renderPass.setBindGroup(i, bindGroups[i]);
+            renderPass2.setBindGroup(i, bindGroups[i]);
         }
-        renderPass.draw(4, 1, 0, 0);
+        renderPass2.draw(4, 1, 0, 0);
+
+        if (!renderPass) {
+            bundle = (renderPass2 as GPURenderBundleEncoder).finish();
+            this._bundleCache[bundleKey] = bundle;
+        }
+
+        return bundle;
     }
 }
--- a/src/Engines/webgpuEngine.ts
+++ b/src/Engines/webgpuEngine.ts
@@ -1033,7 +1033,7 @@ export class WebGPUEngine extends Engine {
 
     /** @hidden */
     public _applyStencilRef(renderPass: GPURenderPassEncoder): void {
-        renderPass.setStencilReference(this._stencilStateComposer.funcRef);
+        renderPass.setStencilReference(this._stencilStateComposer.funcRef ?? 0);
     }
 
     private _blendColorsCurrent: Array<Array<Nullable<number>>> = [[null, null, null, null], [null, null, null, null]];
@@ -1097,7 +1097,11 @@ export class WebGPUEngine extends Engine {
                 if (!this._rttRenderPassWrapper.renderPass) {
                     this._startRenderTargetRenderPass(this._currentRenderTarget!, false, backBuffer ? color : null, depth, stencil);
                 }
-                this._applyScissor(this._currentRenderPass!);
+                if (!this.compatibilityMode) {
+                    this._bundleList.addItem(new WebGPURenderItemScissor(this._scissorCached.x, this._scissorCached.y, this._scissorCached.z, this._scissorCached.w));
+                } else {
+                    this._applyScissor(this._currentRenderPass!);
+                }
                 this._clearFullQuad(backBuffer ? color : null, depth, stencil);
             } else {
                 if (this._currentRenderPass) {
@@ -1110,20 +1114,38 @@ export class WebGPUEngine extends Engine {
                 this._startMainRenderPass(!hasScissor, backBuffer ? color : null, depth, stencil);
             }
             if (hasScissor) {
-                this._applyScissor(this._currentRenderPass!);
+                if (!this.compatibilityMode) {
+                    this._bundleList.addItem(new WebGPURenderItemScissor(this._scissorCached.x, this._scissorCached.y, this._scissorCached.z, this._scissorCached.w));
+                } else {
+                    this._applyScissor(this._currentRenderPass!);
+                }
                 this._clearFullQuad(backBuffer ? color : null, depth, stencil);
             }
         }
     }
 
     private _clearFullQuad(clearColor?: Nullable<IColor4Like>, clearDepth?: boolean, clearStencil?: boolean): void {
-        const renderPass = this._getCurrentRenderPass();
+        const renderPass = !this.compatibilityMode ? null : this._getCurrentRenderPass();
 
         this._clearQuad.setColorFormat(this._colorFormat);
         this._clearQuad.setDepthStencilFormat(this._depthTextureFormat);
-        this._clearQuad.clear(renderPass, clearColor, clearDepth, clearStencil, this.currentSampleCount);
+        this._clearQuad.setMRTAttachments(this._cacheRenderPipeline.mrtAttachments ?? [], this._cacheRenderPipeline.mrtTextureArray ?? []);
 
-        this._applyStencilRef(renderPass);
+        if (!this.compatibilityMode) {
+            this._bundleList.addItem(new WebGPURenderItemStencilRef(this._clearStencilValue));
+        } else {
+            renderPass!.setStencilReference(this._clearStencilValue);
+        }
+
+        const bundle = this._clearQuad.clear(renderPass, clearColor, clearDepth, clearStencil, this.currentSampleCount);
+
+        if (!this.compatibilityMode) {
+            this._bundleList.addBundle(bundle!);
+            this._bundleList.addItem(new WebGPURenderItemStencilRef(this._stencilStateComposer.funcRef ?? 0));
+            this._reportDrawCall();
+        } else {
+            this._applyStencilRef(renderPass!);
+        }
     }
 
     //------------------------------------------------------------------------------
@@ -3740,7 +3762,7 @@ export class WebGPUEngine extends Engine {
                 this._bundleList.addItem(new WebGPURenderItemScissor(this._scissorCached.x, this._scissorCached.y, this._scissorCached.z, this._scissorCached.w));
             }
             if (mustUpdateStencilRef) {
-                this._bundleList.addItem(new WebGPURenderItemStencilRef(this._stencilStateComposer.funcRef));
+                this._bundleList.addItem(new WebGPURenderItemStencilRef(this._stencilStateComposer.funcRef ?? 0));
             }
             if (mustUpdateBlendColor) {
                 this._bundleList.addItem(new WebGPURenderItemBlendColor(this._alphaState._blendConstants.slice()));


gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Creates a new slider
+     * @param name defines the control name
+     */
+     constructor(name?: string, options?: ISlider3DOptions) {
+        super(name);
+
+        this._minimum =  options?.minimum ? options.minimum : 0;
+        this._maximum = options?.maximum ? options.maximum : 100;
```
rickfromwork, line 44: Is there anything stopping a user from mixing up min/max, or making max < min in general?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+    public get minimum(): number {
+        return this._minimum;
+    }
+
+    public set minimum(value: number) {
+        if (this._minimum === value) {
+            return;
+        }
+
+        this._minimum = Math.min(value, this._maximum);
```
rickfromwork, line 58: Personally I think the value should adapt to the new min/max value, instead of having the new min/max value adapt to the current value (Clamp the value to the new range, instead of expanding the range to accommodate the current value).

gui/src/3D/controls/slider3D.ts
rickfromwork, line 44: So by default, there are 100 different values the slider can be set to. Do we want that many as a default?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+        material.specularColor = Color3.Black();
+
+        return material;       
+    }
+
+    private _createBehavior(): PointerDragBehavior {
+        const pointerDragBehavior = new PointerDragBehavior({ dragAxis: new Vector3(1, 0, 0) });
+        pointerDragBehavior.moveAttached = false;
+
+        pointerDragBehavior.onDragObservable.add((event) => {
```
rickfromwork, line 153: Is it possible to have the value be dynamically updated, instead of just on release? While keeping it set to specific values, of course. I've personally always gotten annoyed with scroll bars/other draggables that don't show me what I want until I release the button, meaning I can't browse options as easily.

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+    public onValueChangedObservable = new Observable<number>();
+
+    /**
+     * Creates a new slider
+     * @param name defines the control name
+     */
+     constructor(name?: string, options?: ISlider3DOptions) {
+        super(name);
+
+        this._minimum =  options?.minimum ? options.minimum : 0;
```
chrisfromwork, line 43: nit: can we declare these inline number values (0, 100, 50) as constants?

gui/src/3D/controls/slider3D.ts
chrisfromwork, line 58: should we allow setting minimum and maximum independent of one another? it seems like with this Math min/max check developers can end up in scenarios where the values don't get set correctly if they are trying to set a range (going from (0, 10) to (20, 30) for example could result in (10, 30) if setting min before max).

I feel like these min/max checks could unintentionally cause confusing value behaviors in their current form.

gui/src/3D/controls/slider3D.ts
chrisfromwork, line 43: how do we want to handle scenarios where a minimum is specified but no maximum, etc. Do we need to do things here to ensure that the minimum is less than the maximum or that the value falls within the minimum and maximum range?

binyammathewos: Agreed, I will move those as constants. thanks

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+         if (this._maximum === value) {
+             return;
+         }
+ 
+         this._maximum = Math.max(value, this._minimum);
+         this.value = this._value;
+     }
+
+    /** Gets or sets current value */
+    public get value(): number {
```
chrisfromwork, line 77: nit: do we need separate comments for get vs set? Or will this comment auto resolve for both get and set in an ide?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+        if(!this.node){
+            return 1;
+        }
+
+        return this._sliderBar.position.x + this._sliderBar.scaling.y / 2;
+    }
+
+    // Mesh association
+    protected _createNode(scene: Scene): TransformNode {
+        const anchor = new AbstractMesh(`${this.name}_slider`, scene);
```
chrisfromwork, line 110: >     [](start = 70, length = 4)

nit: remove extra spaces

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+
+        this._sliderThumbMaterial = this._createThumbMaterial(mesh.getScene());
+        this._sliderThumb.material =  this._sliderThumbMaterial;
+    }
+    
+    private _createBarMaterial(scene: Scene): Material {
+        const material  =  new StandardMaterial(`${this.name}_sliderbar_material`, scene);
+        material.specularColor = Color3.Black();
+
+        return material;
```
chrisfromwork, line 139: >        [](start = 24, length = 7)

nit: there are a lot of extra spaces throughout. please remove.

gui/src/3D/controls/slider3D.ts
binyammathewos: The property setter won't allow that. But you are right the constructor is missing this validation. I will address that in the next pr. thanks

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+        return this._sliderBar.position.x + this._sliderBar.scaling.y / 2;
+    }
+
+    // Mesh association
+    protected _createNode(scene: Scene): TransformNode {
+        const anchor = new AbstractMesh(`${this.name}_slider`, scene);    
+        
+        this._sliderBar = MeshBuilder.CreateCylinder(`${this.name}_sliderbar`, { diameter: 0.03, height: 1.0 }, scene);
+        this._sliderBar.rotation.z = -Math.PI / 2;
+        this._sliderBar.scaling.y = 2;
```
chrisfromwork, line 114: >2 [](start = 36, length = 1)

should these arbitrary scale values be set through the options provided in the constructor? regardless, can we get any fallback values declared as nonexported constants at the top of the file/class?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+    protected _affectMaterial(mesh: AbstractMesh) {
+        this._sliderBarMaterial = this._createBarMaterial(mesh.getScene());
+        this._sliderBar.material = this._sliderBarMaterial;
+
+        this._sliderThumbMaterial = this._createThumbMaterial(mesh.getScene());
+        this._sliderThumb.material =  this._sliderThumbMaterial;
+    }
+    
+    private _createBarMaterial(scene: Scene): Material {
+        const material  =  new StandardMaterial(`${this.name}_sliderbar_material`, scene);
```
chrisfromwork, line 136: >  [](start = 26, length = 1)

nit: its unclear why some double spaces exist for this material constructor/material constructors below

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed, I will fix this in the next pr. Thanks

gui/src/3D/controls/slider3D.ts
chrisfromwork: I feel like I have seen value intervals added here as well for other slider controls. we probably want this so that we can support high precision as well as integer value scenarios

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+    private _convertToPosition(value: number): number {
+        const distance = (value - this.minimum) / (this.maximum - this.minimum) * (this.end - this.start);
+        return Math.min(Math.max(this.start + distance, this.start), this.end); 
+    }
+
+    /**
+     * Releases all associated resources
+     */
+     public dispose() {
+        super.dispose();
```
chrisfromwork, line 175: nit: wrong spacing

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+        this._sliderThumb.scaling = new Vector3(2, 2, 2);
+        this._sliderThumb.position.x = this._convertToPosition(this.value);
+        this._sliderThumb.addBehavior(this._createBehavior());
+        this._sliderThumb.setParent(anchor);
+
+        return anchor;
+    }
+
+    protected _affectMaterial(mesh: AbstractMesh) {
+        this._sliderBarMaterial = this._createBarMaterial(mesh.getScene());
```
chrisfromwork, line 128: >_createBarMaterial [](start = 39, length = 18)

should we expose the ability to specify a material as an option?

gui/src/3D/controls/slider3D.ts
binyammathewos: I took this from 2D slider control, which allows the slider be set to 100 values by default. We don't have to use 100 values. Please let me know your recommendation.  Thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed. May be we can add this as an option for the end user to pass when creating the slider. By default we can dynamically update the value onDrag. Thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed, I have a similar issues reported by rickfromwork , I will address this in my next pr. thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed, We are missing validation here. we can throw exception for invalid ranges. I will address this in the next pr. thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: I followed the other 3D controls style. But I  am okay to put the comment for both setter and getter. Thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed, thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed, will do that. Thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: I will address that in my next pr. thanks

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed, I will expose the materials as a public members. Thanks

gui/src/3D/controls/slider3D.ts
Popov72, line 44: Is it expected that we can't set `maximum` to 0?

If 0 is a valid value for `maximum`, you can use:
```javascript
this._maximum = options?.maximum ?? 100;
```

For `minimum`, you can also use:
```javascript
this._minimum = options?.minimum ?? 0;
```

gui/src/3D/controls/slider3D.ts
rgerd, line 110: nit: I'd make this a `TransformNode` since you don't actually need a mesh.

gui/src/3D/controls/slider3D.ts
rickfromwork: I was thinking about this too. I think having a single comment is fine, since the documentation generated from this puts both the setter and getter into a single entry.

For example, Mesh.ts's isUnIndexed
https://doc.babylonjs.com/typedoc/classes/babylon.mesh#isunindexed

rickfromwork: It does look like the comment gets duplicated, though...

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+    
+    private _createBarMaterial(scene: Scene): Material {
+        const material  =  new StandardMaterial(`${this.name}_sliderbar_material`, scene);
+        material.specularColor = Color3.Black();
+
+        return material;       
+    }
+
+    private _createThumbMaterial(scene: Scene): Material {
+        const material  =  new StandardMaterial(`${this.name}_sliderthumb_material`, scene);
```
msDestiny14, line 143: super tiny nit: spacing

ryantrem:
Do you have any screenshots or animated gifs you can share here so we can see what it actually looks like?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,193 @@
...
+    minimum?: number,
+    maximum?: number,
+    value?: number
+};
+
+/**
+ * Class used to create a slider in 3D
+ */
+ export class Slider3D extends Control3D {
+    private _sliderBarMaterial: Material;
```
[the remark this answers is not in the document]
binyammathewos: Agreed, thanks

msDestiny14:
> Do you have any screenshots or animated gifs you can share here so we can see what it actually looks like?

Second on this.

gui/src/3D/controls/index.ts
```diff
@@ -13,4 +13,5 @@ export * from "./touchButton3D";
 export * from "./touchMeshButton3D";
 export * from "./touchHolographicButton";
 export * from "./touchToggleButton3D";
-export * from "./volumeBasedPanel";
\ No newline at end of file
+export * from "./volumeBasedPanel";
+export * from "./slider3D";
```
rickfromwork, line 17: Please place alphabetically in the list

gui/src/3D/controls/slider3D.ts
binyammathewos: This issue has been resolved. Take a look at the updated version.

gui/src/3D/controls/slider3D.ts
binyammathewos: This issue has been resolved. Take a look at the updated version.

gui/src/3D/controls/slider3D.ts
binyammathewos: Agreed with Rick, having a single for both setter and getter make sense.

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+    public get minimum(): number {
+        return this._minimum;
+    }
+
+    public set minimum(value: number) {
+        if (this._minimum === value) {
+            return;
+        }
+
+        this._minimum = Math.max(value, SLIDER_MIN);
```
rickfromwork, line 78: Ah, so the SLIDER_* values are used as limiting values, not just the initial state. Interesting. I want to say there should also be a check in here that the new minimum is not larger than the current maximum, but I remember talking already about how that check doesn't exist for the 2D slider...

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+        this._value = Math.max(Math.min(value, this._maximum), this._minimum);
+        this.onValueChangedObservable.notifyObservers(this._value);
+    }
+
+    protected get start(): number {
+        if (!this.node) {
+            return -SLIDER_SCALING / 2;
+        }
+
+        return this._sliderBar.position.x - this._sliderBar.scaling.y / 2;
```
rickfromwork, line 128: Would it be possible to re-orient the thumb/bar such that the axis are aligned, letting us write something more readable like "this._sliderBar.position.x - this._sliderBar.scaling.x / 2"?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+    private _convertToPosition(value: number): number {
+        const position = ((value - this.minimum) / (this.maximum - this.minimum)) * (this.end - this.start) + this.start;
+        return Math.min(Math.max(position, this.start), this.end);
+    }
+
+    private _convertToValue(position: number): number {
+        let value = ((position - this.start) / (this.end - this.start)) * (this.maximum - this.minimum) + this.minimum;
+        value = this.step ? Math.round(value / this.step) * this.step : value;
+
+        return Math.max(Math.min(this.roundingToDecimal ? Math.round(value * 100) / 100 : value, this._maximum), this._minimum);
```
rickfromwork, line 221: Why is the decimal rounding done to two decimal places?

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Releases all associated resources
+     */
+    public dispose() {
+        super.dispose();
+
+        if (this._sliderBar) {
```
rickfromwork, line 230: nit: If it's like this elsewhere that's fine, but I'm always a fan of condensing one-line conditionals:
this._sliderBar?.dispose();

gui/src/3D/controls/slider3D.ts
binyammathewos: We are assuming the slider values are only positive numbers. Since the default min value is 0, we are using it here to limit the min value the user trying to set to the default min value. Since 2D slider don't prevent the user from setting min > max, I am assuming we can leave it like this. The slider won't crash but will not work as intended. I was trying to avoid this circular validation for min & max. It is very confusing for the end user.

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+    private _step: number;
+
+    /** Observable raised when the sldier value changes */
+    public onValueChangedObservable = new Observable<number>();
+
+    /**
+     * Creates a new slider
+     * @param name defines the control name
+     */
+    constructor(name?: string, options?: ISlider3DOptions) {
```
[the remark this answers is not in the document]
binyammathewos: Yes, I do agree with that. It was Rick's idea to pass this as optional parameters. I will reach out to him to see if he has strong argument against it.

gui/src/3D/controls/slider3D.ts
binyammathewos: Yes, we can. I even found it very confusing every time I see this line.

gui/src/3D/controls/slider3D.ts
binyammathewos: No specific reason. I looked up the MRTK slider example, which uses two decimal places for values. One idea I have is to let the use set this as an option. We can default it to two decimals and we don't even have to check if  we need rounding or not (I am using the min, max or step values to determine that)

gui/src/3D/controls/slider3D.ts
binyammathewos: Nice, I like that.

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+        const thumbMaterial = new StandardMaterial(`${this.name}_sliderthumb_material`, mesh.getScene());
+        thumbMaterial.specularColor = Color3.Black();
+        mesh.material = thumbMaterial;
+
+        this._sliderBarMaterial = barMaterial;
+        this._sliderThumbMaterial = thumbMaterial;
+    }
+
+    private _createBehavior(): PointerDragBehavior {
+        const pointerDragBehavior = new PointerDragBehavior({ dragAxis: new Vector3(1, 0, 0) });
```
chrisfromwork, line 196: >new Vector3(1, 0, 0) [](start = 72, length = 20)

nit: fwiw, some of these more common constants have static values. Like this vector3 here exists as a static Vector3.Right() value.

gui/src/3D/controls/slider3D.ts
chrisfromwork: sounds good

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,246 @@
...
+     */
+    public get sliderThumbMaterial(): StandardMaterial {
+        return this._sliderThumbMaterial;
+    }
+
+    // Mesh association
+    protected _createNode(scene: Scene): TransformNode {
+        const anchor = new TransformNode(`${this.name}_slider`, scene);
+
+        const sliderBar = MeshBuilder.CreateCylinder(`${this.name}_sliderbar`, { diameter: 0.03, height: 1.0 }, scene);
```
chrisfromwork, line 164: >0.03 [](start = 91, length = 4)

Do we think that an external developer will need to customize these values? or can they be reset after slider control creation?

binyammathewos: I was assuming the developer can use scaling get/set property to scale the control  up/down to the desired size. I think of these values as scale of 1 for the control as a whole. I don't think we need these to be customizable.  Let me know your thought.

gui/src/3D/controls/slider3D.ts
binyammathewos: Nice, I did not know there is values for this. I only know Vector3.Zero().

gui/src/3D/controls/slider3D.ts
binyammathewos: This issue might be resolved. Can you take a look at the update on this PR? Thanks,

gui/src/3D/controls/slider3D.ts
binyammathewos: This issue might be resolved. Can you take a look at the update on this PR? Thanks,

chrisfromwork (approved):
:shipit:

binyammathewos:
Please find attached here the Slider3D control in action. Thanks

slider3d

gui/src/3D/controls/slider3D.ts
Popov72: It is solved indeed as the constructor does not take an `options` object anymore :)

gui/src/3D/controls/slider3D.ts
```diff
@@ -0,0 +1,220 @@
...
+    public onValueChangedObservable = new Observable<number>();
+
+    /**
+     * Creates a new slider
+     * @param name defines the control name
+     */
+    constructor(name?: string) {
+        super(name);
+
+        this._minimum = SLIDER_MIN;
```
RaananW, line 44: Would it be possible to pass those values in the constructor, defaulting to those values? In the current implementation it will take 5 different calls if I want to change all values.

gui/src/3D/controls/slider3D.ts
RaananW: I am for allowing customization in this case. Mainly because you are changing the scaling.y value. A user that will manipulate the scaling values as they see fit will have these values overwritten. Might be enough to define a `scaleFactor` that will be applied to all of those default values.

RaananW (commented):
LGTM! with two comments waiting for your feedback

sebavan:
LGTM as well, I ll merge a soon as @RaananW comments have been addressed.

gui/src/3D/controls/slider3D.ts
binyammathewos: To be consistent with the other GUI control, we decided not to pass these parameters in the constructor. The initial PR uses constructor parameters, but there was a code review feedback to remove it for the reason I mentioned earlier.

gui/src/3D/controls/slider3D.ts
binyammathewos: The user can use getter/setter public property for scaling the control up or down.  Those default values are like a seed that will give the control the default look.

RaananW:
I just noticed you are pushing the changes to the preview branch and not to master. Any reason for that?

sebavan:
Such a good catch ranaan !!!!! yup please retarget the pr to master.

RaananW (changes requested):
Sorry, just to make sure it is not merged to the wrong branch. You have my approval when merged to master

binyammathewos:
> I just noticed you are pushing the changes to the preview branch and not to master. Any reason for that?

No, reason. I re-target the PR to master. Thanks

RaananW (approved):

RaananW:
Just wanted to make sure all of the conversations are resolved before merging this. Are we good to go?

PR #10149, opened 2021-04-07
Fix instance picking when in billboard mode

Popov72:
See https://forum.babylonjs.com/t/billboardmodeall-breaks-pick-order-for-instances/19673


sebavan:
LGTM but just for my information why the TMP 0 was not working ? was it overriden later on in the code ?

Popov72:
Yes, `TmpVectors.Matrix[0]` is used in ray intersection code, so the world matrix was modified along the way.

sebavan:
In this case could we not use another TMP Matrix like the TmpVectors.Matrix[1] ? or at least a static in the file ? to prevent over allocating and adding properties to the mesh class and not go over the limit ?

Popov72:
That seems dangerous to me: any code can be executed after the time one calls `InstancedMesh.getWorldMatrix()`. So, if some code is using the same temp variable we used, it will fail.

I think the life time of those temp variables should be limited to the scope they are used in and that we should never return a temp vector/matrix from a function the user can call: we can't know which code will be called afterwards. If this code is using for internal computation this same temp, we are screwed up.

I don't see another safe way to do it than having a local per instance variable to hold the matrix... What I can do is to let it undefined and create it the first time we need it (as billboard mode is rarely used, the penalty will occur only in this case).

sebavan:
Ok Totally agree if not used directly here so all good :-)

src/Meshes/instancedMesh.ts
```diff
@@ -34,6 +34,7 @@ Mesh._instancedMeshFactory = (name: string, mesh: Mesh): InstancedMesh => {
 export class InstancedMesh extends AbstractMesh {
     private _sourceMesh: Mesh;
     private _currentLOD: Mesh;
+    private _billboardWorldMatrix = new Matrix();
```
deltakosh, line 37: can we make it static please? So we don't create one per mesh

Popov72: We can't have it static because else all instances would use the same matrix.

deltakosh: oh ok, I did not check the code corectly, I thought it was a temporary matrix used as an intermediary

In this case please simply delay creating it

Popov72: Done

deltakosh (approved):

PR #10155, opened 2021-04-08
Fix Scene.getPointerOverMesh returning disposed mesh

Popov72:
See https://forum.babylonjs.com/t/scene-getpointerovermesh-can-pick-the-disposed-mesh/19786


RaananW (approved):

Popov72:
The error in WebGL2 tests is transient and can be ignored.

deltakosh (approved):

PR #10157, opened 2021-04-08
test deviceInputSystem is valid before disposing

CedricGuillemet:
Nightly BabylonNative VT fails with latest babylonjs nightly, 
`this.deviceInputSystem` is undefined and it stops VT.

### DIFF
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -1837,7 +1837,9 @@ export class Engine extends ThinEngine {
         this.disableVR();
 
         // DeviceInputSystem
-        this.deviceInputSystem.dispose();
+        if (this.deviceInputSystem) {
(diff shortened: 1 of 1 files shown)


src/Engines/engine.ts
```diff
@@ -1837,7 +1837,9 @@ export class Engine extends ThinEngine {
         this.disableVR();
 
         // DeviceInputSystem
-        this.deviceInputSystem.dispose();
+        if (this.deviceInputSystem) {
```
Popov72, line 1840: You could do a one liner fix with:
```typescript
this.deviceInputSystem?.dispose();
```
:)

Popov72 (approved):

deltakosh (approved):

PR #10162, opened 2021-04-08
Add stencil properties to the material

Popov72:
I needed to keep separate stencil states at the engine level independent from the material stencil states for some effects to work (like the outline renderer). So, there's a new `StencilStateComposer` class which is given the (engine) global states and the material states and is responsible for setting the final states on the pipeline (stencil states from the material take precedence over states from the engine if `material.stencil.enabled == true` and if `stencilStateComposer.useStencilGlobalOnly == false`).


deltakosh (approved):
pretty smart! I like it

deltakosh:
waiting for @sebavan approval (and there is a tiny conflict)

sebavan (approved):

PR #10164, opened 2021-04-09
Use observables for inputsystem singleton

CedricGuillemet:
Fixes bug introduced in this commit : https://github.com/BabylonJS/Babylon.js/commit/669a5134437a931fccd3be7f9bbad1722c18eb03
DeviceInputSystem becomes a singleton but onInputChange was a callback that was overriden.
Using Observables now to fix the issue.

I've only tested camera mouse movement and gizmo. Let me know if I must test any other PG.


CedricGuillemet:
Closing this PR as it fixes web but will break Native.

Issue #10146, opened 2021-04-06
Slider control breaks/disappears when passing in a step value greater than 1


rickfromwork:
The Slider control supports passing in a step value, such that dragging the slider increments it by <step> at a time. This works as expected for values under or equal to 1, but fails on values greater than 1. For greater values, the slider disappears as soon as it receives a pointer down (it is clicked).

Looking at the code, it is hitting a divide-by-zero error in baseSlider.ts's _updateValueFromPointer call,specifically at
`const mult = (1 / this._step) | 0;`
`this.value = this._step ? ((value * mult) | 0) / mult : value;`
It seems like Javascript does not handle decimal values when performing _bitwise_ operations, so the _bitwise OR_ on the first line resolves to 0. Then the next line divides by that value.

Why are bitwise operations being used? Why is **mult** ever being set to 0, given the next line divides by it?

**Repro**
- Bug repro on https://www.babylonjs-playground.com/#NGS9AU#343
- Expected result: Step should increase by 2 at a time
- Current result: Slider disappears (the editor shows the sphere's scaling is NaN)

msDestiny14:
Thank you. Going to look into it now! :)

thomlucc:
Closing as fixed!

PR #10169, opened 2021-04-09
Fix for GUI slider step values greater than one

msDestiny14:
https://github.com/babylonjs/babylon.js/issues/10146

### DIFF
--- a/gui/src/2D/controls/sliders/baseSlider.ts
+++ b/gui/src/2D/controls/sliders/baseSlider.ts
@@ -300,7 +300,7 @@ export class BaseSlider extends Control {
         else {
             value = this._minimum + ((x - this._currentMeasure.left) / this._currentMeasure.width) * (this._maximum - this._minimum);
         }
-    
+
         this.value = this._step ? ((value / this._step) | 0) * this._step : value;
     }
 
--- a/gui/src/2D/controls/sliders/baseSlider.ts
+++ b/gui/src/2D/controls/sliders/baseSlider.ts
@@ -301,7 +301,7 @@ export class BaseSlider extends Control {
             value = this._minimum + ((x - this._currentMeasure.left) / this._currentMeasure.width) * (this._maximum - this._minimum);
         }
 
-        const mult = (1 / this._step) | 0;
+        const mult = (1.0 / this._step);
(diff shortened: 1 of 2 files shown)


deltakosh:
Does it still work with values smaller than 1?

msDestiny14:
@deltakosh yep tested it with 0.5 and 0.25

gui/src/2D/controls/sliders/baseSlider.ts
```diff
@@ -301,7 +301,7 @@ export class BaseSlider extends Control {
             value = this._minimum + ((x - this._currentMeasure.left) / this._currentMeasure.width) * (this._maximum - this._minimum);
         }
 
-        const mult = (1 / this._step) | 0;
+        const mult = (1.0 / this._step);
```
Popov72, line 304: Note that it is a breaking change when `this._step < 1`.

Maybe you should do instead:
```
        const mult = this._step < 1 ? (1.0 / this._step) | 0 : (1.0 / this._step);
```

gui/src/2D/controls/sliders/baseSlider.ts
```diff
@@ -301,7 +301,7 @@ export class BaseSlider extends Control {
             value = this._minimum + ((x - this._currentMeasure.left) / this._currentMeasure.width) * (this._maximum - this._minimum);
         }
 
-        const mult = (1 / this._step) | 0;
+        const mult = (1.0 / this._step);
         this.value = this._step ? ((value * mult) | 0) / mult : value;
```
Popov72, line 305: I don't understand this formula (I know it is not changed by this PR)...

Either we want `this.value` to be independent from `this._step` and then `this.value = value` or we want `this.value` to be a multiple of `this._step`, in which case the formula is wrong and should be something like:

```javascript
this.value = this._step ? ((value / this._step) | 0) * this._step : value;
```
I think we are in the latter case when `this._step != 0` because we are in the former case when `this._step == 0`.

deltakosh: I tend to agree

msDestiny14: Yea I agree this makes a lot more sense.

deltakosh (approved):

PR #10182, opened 2021-04-13
Make the canvas optional in the process pointer move function

RaananW:
The canvas can be optional when processing the pointer move event (which can be emulated in environments without a canvas (i.e. XR).

There is a potential breaking change here, @sevan / @deltakosh - would be great to hear your opinion about it. The breaking change is only in the typescript typing of  the `PointerMoveStageAction` - it is possible that if a typescript project is using it directly they will now need to check if the canvas is not null.

### DIFF
--- a/src/Inputs/scene.inputManager.ts
+++ b/src/Inputs/scene.inputManager.ts
@@ -70,7 +70,12 @@ export class InputManager {
     private _onPointerDown: (evt: IPointerEvent) => void;
     private _onPointerUp: (evt: IPointerEvent) => void;
 
-    private _initClickEvent: (obs1: Observable<PointerInfoPre>, obs2: Observable<PointerInfo>, evt: IPointerEvent, cb: (clickInfo: _ClickInfo, pickResult: Nullable<PickingInfo>) => void) => void;
+    private _initClickEvent: (
+        obs1: Observable<PointerInfoPre>,
+        obs2: Observable<PointerInfo>,
+        evt: IPointerEvent,
+        cb: (clickInfo: _ClickInfo, pickResult: Nullable<PickingInfo>) => void
+    ) => void;
     private _initActionManager: (act: Nullable<AbstractActionManager>, clickInfo: _ClickInfo) => Nullable<AbstractActionManager>;
     private _delayedSimpleClick: (btn: number, clickInfo: _ClickInfo, cb: (clickInfo: _ClickInfo, pickResult: Nullable<PickingInfo>) => void) => void;
     private _delayedSimpleClickTimeout: number;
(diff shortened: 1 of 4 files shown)


deltakosh (approved):
No worries for that breaking change for me

src/Inputs/scene.inputManager.ts
```diff
@@ -861,35 +858,32 @@ export class InputManager {
...
+                    this._onPointerUp(currentEvent as IPointerEvent);
                 }
 
                 if (attachMove) {
-                    if (inputIndex === PointerInput.Horizontal
-                        || inputIndex === PointerInput.Vertical
-                        || inputIndex === PointerInput.DeltaHorizontal
-                        || inputIndex === PointerInput.DeltaVertical
-                        || inputIndex === PointerInput.FakeMove) {
+                    if (inputIndex === PointerInput.Horizontal || inputIndex === PointerInput.Vertical || inputIndex === PointerInput.DeltaHorizontal || inputIndex === PointerInput.DeltaVertical || inputIndex === PointerInput.FakeMove) {
```
rickfromwork, line 881: nit: same here, always a fan of multi-lining longer conditional checks.

src/Inputs/scene.inputManager.ts
```diff
@@ -806,37 +804,36 @@ export class InputManager {
...
-            const metaKey = (isKeyboardActive && (this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_META_KEY1) === 1
-                || this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_META_KEY2) === 1
-                || this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_META_KEY3) === 1));
-            const shiftKey = (isKeyboardActive && this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_SHIFT_KEY) === 1);
+            const altKey = isKeyboardActive && this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_ALT_KEY) === 1;
+            const ctrlKey = isKeyboardActive && this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_CTRL_KEY) === 1;
+            const metaKey =
+                isKeyboardActive &&
+                (this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_META_KEY1) === 1 || this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_META_KEY2) === 1 || this._deviceInputSystem.pollInput(DeviceType.Keyboard, 0, Constants.INPUT_META_KEY3) === 1);
```
rickfromwork, line 811: nit: I do like the readability that comes with placing each OR clause on a different line. Maybe just indent them more so it's clear they aren't grouped in alongside the AND operator?

src/Inputs/scene.inputManager.ts
RaananW: That's the auto-formatter causing havoc. Let me see if I can reconfigure it

src/Inputs/scene.inputManager.ts
RaananW: same as above... let me check if i can reconfigure it

RaananW: should be fine now :-)

rickfromwork (approved):

PR #10192, opened 2021-04-14
`level` output for the TextureBlock / switch for disabling level multiplication

rassie:
This PR is a playground for the discussion in https://forum.babylonjs.com/t/nme-make-texture-level-work-for-normal-maps-as-it-does-in-other-materials/19443/14. This code currently does not work (at all!), but might work someday.

### DIFF
--- a/src/Materials/Node/Blocks/Dual/textureBlock.ts
+++ b/src/Materials/Node/Blocks/Dual/textureBlock.ts
@@ -74,6 +74,11 @@ export class TextureBlock extends NodeMaterialBlock {
     public convertToLinearSpace = false;
 
     /**
+     * Gets or sets a boolean indicating if multiplication of texture with level should be disabled
+     */
+    public disableLevelMultiplication = false;
+
+    /**
      * Create a new TextureBlock
      * @param name defines the block name
      */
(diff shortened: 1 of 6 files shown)


sebavan:
LGTM aside the // FIXME block which needs to be removed. @Popov72  will provide a more in depth analysis :-)

rassie:
> LGTM aside the // FIXME block which needs to be removed.

Yeah, I couldn't find the reason for that. Would look into it again later today, but would also appreciate any help with that. Maybe there is something obvious that I've overlooked.

rassie:
@sebavan @Popov72 have found and fixed the FIXME problem. This patch should be complete now.

sebavan:
@rassie, you have one lint issue:
image

And you should credit yourself in the dist/preview release/whats new.md file

So that the build will be all green and happy.

@Popov72 will do the thorough review soon.

rassie:
Done!

sebavan (approved):

src/Materials/Node/Blocks/Dual/textureBlock.ts
```diff
@@ -73,6 +73,11 @@ export class TextureBlock extends NodeMaterialBlock {
      */
     public convertToLinearSpace = false;
 
+    /**
+     * Gets or sets a boolean indicating if content needs to be converted to linear space
+     */
+    public disableLevelMultiplication = false;
```
Popov72, line 79: nit: The doc is wrong

src/Shaders/ShadersInclude/bumpFragment.fx
```diff
@@ -14,7 +14,7 @@
 	#elif defined(BUMP)
 		// flip the uv for the backface
 		vec2 TBNUV = gl_FrontFacing ? vBumpUV : -vBumpUV;
-		mat3 TBN = cotangent_frame(normalW * normalScale, vPositionW, TBNUV);
+		mat3 TBN = cotangent_frame(normalW * normalScale, vPositionW, TBNUV,  vTangentSpaceParams);
```
Popov72, line 17: nit: two spaces instead of one

Popov72 (commented):
LGTM: Great job!

One small thing is that 3 functions have been removed from `bumpFragmentFunctions.fx` and that could be a breaking change if someone was using them...

However, I don't think this code was supposed to be called by external users and having `3 x perturbNormal + 2 x cotangent_frame` functions reduced to `1 x perturbNormal + 1 x cotangent_frame` sounds good to me, so let's keep it!

rassie:
> LGTM: Great job!
> 
> One small thing is that 3 functions have been removed from `bumpFragmentFunctions.fx` and that could be a breaking change if someone was using them...
> 
> However, I don't think this code was supposed to be called by external users and having `3 x perturbNormal + 2 x cotangent_frame` functions reduced to `1 x perturbNormal + 1 x cotangent_frame` sounds good to me, so let's keep it!

It's a good thing there is a 5.0 release coming then ;) I'll fix the nits in a couple of hours when I'm not on mobile! Thanks!

rassie:
Nits are done, thanks for the feedback!

PR #10209, opened 2021-04-16
Fix onSuccess exception handling in FileTools.RequestFile

mrlika:
**Background**

If put under `.gltf` extension a non-json file and try to load it this `await` never resolves and leaves `scene` in loading state forever:

```typescript
try {
  await SceneLoader.AppendAsync('https://some.site.com/models/', 'bad_non_json_file.gltf', scene);
  console.log('Model loading success');
} catch (e) {
  console.log('Model loading failed');
}
```
None of `console.log` is never called in such case.

It is because `onSuccess` callback throws exception `SyntaxError: JSON.parse: unexpected character ...` here: https://github.com/BabylonJS/Babylon.js/blob/6626433abbee8f1d1e82499915645cc54b785a50/loaders/src/glTF/glTFFileLoader.ts#L546

**Solution**

The solution is to avoid pitfalls like this one globally by fixing `FileTools.RequestFile` to behave more like a promise and call `onError` if `onSuccess` fails.


sebavan:
I like it, let s see what @bghgary thinks about it ?

mrlika:
Probably it is better to re-throw error if error handler is missing:

```javascript
if (onError) {
    onError(e);
} else {
    throw e;
}
```

bghgary:
This is like here: https://github.com/BabylonJS/Babylon.js/blob/43a9bb796520175c0460326f71aeb3cb75fa13d2/src/Loading/sceneLoader.ts#L725-L730

Ideally, we should fix this for all the callbacks (`onProgress`, `onOpened`) and not just `onSuccess`, in case any of these throw an exception. `onError` can mimic this:
https://github.com/BabylonJS/Babylon.js/blob/43a9bb796520175c0460326f71aeb3cb75fa13d2/src/Loading/sceneLoader.ts#L702-L707

mrlika:
@bghgary, you are right. I added `onOpened` error handling because exception in it also breaks promise flow. Please check.

Error handling of `onProgress` can be skipped since it doesn't hang promise.

In overall type of error handler `onError?: (error: RequestFileError) => void` looks incorrect to me because `RequestFile` method can throw not only `RequestFileError`s.

Because of this I don't know a good way to add error handling to offline branch of this method - there is no `WebRequest` object required by `RequestFileError` constructor.

Looks like it is a good idea to change error handler type to `onError?: (error: unknown) => void` that touches about 5 files in the project and can be done in separate PR.

src/Misc/fileTools.ts
```diff
@@ -363,11 +364,25 @@ export class FileTools {
                 }
             };
 
+            const handleError = (error: unknown) => {
+                const message = error instanceof Error ? error.message : 'Unknown error';
```
bghgary, line 368:
```suggestion
                const message = error.message || "Unknown error";
```

bghgary (commented):
One minor comment. Looks good otherwise.

bghgary:
> Error handling of onProgress can be skipped since it doesn't hang promise.

I suppose that is true. To be consistent, we shouldn't do it in the [SceneLoader](https://github.com/BabylonJS/Babylon.js/blob/43a9bb796520175c0460326f71aeb3cb75fa13d2/src/Loading/sceneLoader.ts#L883-L890) either.

> In overall type of error handler onError?: (error: RequestFileError) => void looks incorrect to me because RequestFile method can throw not only RequestFileErrors.
> 
> Because of this I don't know a good way to add error handling to offline branch of this method - there is no WebRequest object required by RequestFileError constructor.
> 
> Looks like it is a good idea to change error handler type to onError?: (error: unknown) => void that touches about 5 files in the project and can be done in separate PR.

It's a bit more complicated than it seems as this will propagate all the way to the scene loader plugins. I'm not sure updating this is worth it.

bghgary (approved):

deltakosh:
Tests are failing :(

bghgary:
I reran them to make sure it's not from yesterday's bug.

mrlika:
Merged with master again. Test pass.

bghgary (approved):

PR #10210, opened 2021-04-17
Remove unnecessary bind calls when using instancing

Popov72:
When using instancing, the mesh bind call is done in the `_processRendering` method, so no need to do it earlier.


sebavan (approved):
GG

PR #10219, opened 2021-04-19
ShaderMaterial: Add support for morph target texture

Popov72:
See https://forum.babylonjs.com/t/shader-material-doesnt-work-with-instances/793/11


deltakosh (approved):

Popov72:
The error makes no sense to me: "Invalid Chai property: getClassName"

It happens on two different tests for WebGL1/WebGL2. I think we should restart the tests and see what happens.

deltakosh:
I restarted the tests

Popov72:
I think this one can be merged.

PR #10221, opened 2021-04-19
Fix for wrongly rendered GUI rectangle on resize with adaptWidthToChildren

msDestiny14:
https://forum.babylonjs.com/t/rectangle-gets-corrupted-on-resize-when-using-adaptwidthtochildren-with-textblock-and-padding/20124

### DIFF
--- a/gui/src/2D/controls/container.ts
+++ b/gui/src/2D/controls/container.ts
@@ -375,6 +375,7 @@ export class Container extends Control {
                         if (this.adaptHeightToChildren && child._height.isPixel) {
                             computedHeight = Math.max(computedHeight, child._currentMeasure.height + child.paddingTopInPixels + child.paddingBottomInPixels);
                         }
+                        this._markAsDirty();
                     }
                 }
 
@@ -395,6 +396,7 @@ export class Container extends Control {
 
                 this._postMeasure();
             }
+            
             rebuildCount++;
         }
         while (this._rebuildLayout && rebuildCount < this.maxLayoutCycle);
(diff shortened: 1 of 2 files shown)


deltakosh (approved):

gui/src/2D/controls/container.ts
```diff
@@ -375,6 +375,7 @@ export class Container extends Control {
                         if (this.adaptHeightToChildren && child._height.isPixel) {
                             computedHeight = Math.max(computedHeight, child._currentMeasure.height + child.paddingTopInPixels + child.paddingBottomInPixels);
                         }
+                        this._markAsDirty();
```
deltakosh, line 378: This should not happen every time. (only when computeWidth or height change)

deltakosh (changes requested):
A subtle change

gui/src/2D/controls/container.ts
msDestiny14: So setting inside markAsDirty() inside the ifs will not fix because the roto container which is the one that needs to be updated does not have adaptiveScaling. Does it not make sense if the children get updated then the root container will get updated as well?

deltakosh: ok I get it so maybe calling it only when adaptivescaling is on. My point is that we need to figure out when to call markasdirty because we cannot call it per frame as it will defy to overall dirtiness feature ;D

gui/src/2D/controls/container.ts
```diff
@@ -371,11 +371,12 @@ export class Container extends Control {
 
                         if (this.adaptWidthToChildren && child._width.isPixel) {
                             computedWidth = Math.max(computedWidth, child._currentMeasure.width + child.paddingLeftInPixels + child.paddingRightInPixels);
+                            this.parent?._markAsDirty();
```
deltakosh, line 374: can we store the value of computedWitdh before the previous line and only call markasdirty if the value changed?

msDestiny14: sure can.

deltakosh (approved):

PR #10222, opened 2021-04-19
Add near interaction to WebXRControllerPointerSelection

satyapoojasama:
Added new WebXR feature "**WebXRNearInteraction**" to enable pointer simulations for near interactions. The following functionality has been added:
- new parameters "**isNearPickable**" and "**isNearGrabbable**" are added to abstractMesh class. 
- pick info is populated and pointermove is simulated during a **Near Pick** action on a mesh with "**isNearPickable**" parameter turned on.
- pick info is populated and pointermove is simulated during a **Near Hover** action on a mesh with "**isNearPickable**" parameter turned on.
- pick info is populated, pointer down, up and move are simulated during a **Near Grab** action on a mesh with "**isNearGrabbable**" parameter turned on.
- Far ray interactions are turned off when the index tip of the dominant hand comes in the hoverRange of a mesh whose "isNearPickable" or "isNearGrabbable" parameters turned on. 

Testing done: This code has been verified and it works great with 6dof and bounding box.


src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -724,6 +858,34 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                    pickingInfo.hit = result;
+                    pickingInfo.pickedMesh = mesh;
+                    pickingInfo.pickedPoint = handMesh.position;
+                    pickingInfo.originMesh = handMesh;
+                    pickingInfo.distance = Vector3.Distance(mesh.position, handMesh.position);
+                    mesh.updateFacetData();
+                    const closestFacet = mesh.getClosestFacetAtCoordinates(handMesh.position.x, handMesh.position.y, handMesh.position.z, undefined, true, true);
+                    if (closestFacet)
+                    {
+                        pickingInfo.ray = new Ray(handMesh.position, mesh.getFacetNormal(closestFacet).negate(), pickingInfo.distance);
```
RaananW, line 881: Is it possible to reuse this ray instead of creating a new one?  This happens on each frame and will have impact on the GC.

Same goes to the `getFacetNormal` call (use `getFacetNormalToRef` instead, and negate in place, so you don't create new vectors)

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -118,6 +119,7 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
         }
 
         const { laserPointer, selectionMesh } = this._generateNewMeshPair(xrController.pointer);
+        const { hoverIndexMesh, pickIndexMesh} = this._generateNewHandTipMeshes();
```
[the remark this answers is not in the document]
RaananW: Also, this should only execute for controllers with hand defs, no?

RaananW (commented):
Got a few comments and questions.

Some nits - 
* Would be great if you can format the code using `prettier` (and the .prettier configuration file we have in the repo).
* Try to create as little objects as possible and reuse as much as you can.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -264,6 +283,11 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                 meshUnderPointer: null,
                 pick: null,
                 tmpRay: new Ray(new Vector3(), new Vector3()),
+                pickIndexMesh,
```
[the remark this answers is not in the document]
satyapoojasama: currently it does not add anything to gaze mode Raanan.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +410,51 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                     scene.pointerY = this._screenCoordinatesRef.y;
                 }
             }
-            controllerData.pick = this._scene.pickWithRay(controllerData.tmpRay, this._scene.pointerMovePredicate || this.raySelectionPredicate);
 
-            const pick = controllerData.pick;
+            let pick = null;
+            let hoverAtOrigin = false;
+            if (controllerData.hoverIndexMesh) {
```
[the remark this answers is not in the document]
satyapoojasama: you are right, but as the value is initialized as nullable, ts is making me check it before using it. If you have better idea, i am all ears :)

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
satyapoojasama: yeah i believe so. and it is a great idea to add the options for better performance.

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: I would love to do that, but it changes for every frame right? I create the ray depending on the index finger position which can change every frame.

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
RaananW: Is there a need to create the meshes here? Or is it just for a future implementation?

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
RaananW: I see! So this is not only for hands, but for controllers as well. Got it.

src/XR/features/WebXRControllerPointerSelection.ts
RaananW: The controller data object has a tmpRay that is reused on each frame. You can use it (if it makes sense) or create a new one for this interaction.

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
satyapoojasama: It is actually not required Raanan, but if i dont add them it errors out saying the controller definition is missing these things. Probably the controller definition is common for both ray and gaze mode. I am not sure how to separate them.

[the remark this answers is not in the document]
RaananW: I thought this was the reason they are optional. I will try pulling this branch later and understand what the issue is

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -558,6 +659,34 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                         }
                     }
                 });
+
+                // squeezeComponent is a new member on the controllerData
+                controllerData.squeezeComponent = motionController.getComponent("grasp");
```
RaananW, line 664: Is it possible that this is null on certain controllers?  If it can be null, the next line will fail.

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
satyapoojasama: that would make sense, true! thank you Raanan

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +422,63 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
                 }
             }
-            controllerData.pick = this._scene.pickWithRay(controllerData.tmpRay, this._scene.pointerMovePredicate || this.raySelectionPredicate);
 
-            const pick = controllerData.pick;
+            let pick = null;
+            let hoverAtOrigin = false;
+            let hoverRange = false; 
+            if (controllerData.hoverIndexMeshTip) {
+                let hoverPickInfo = this._pickWithMesh(controllerData.hoverIndexMeshTip, false, this.nearPickPredicate);
```
rickfromwork, line 430: These could probably be combined, to reduce the number of pickWithMesh calls.
(mesh)=>{return this.nearPickPredicate(mesh) || this.nearGrabPredicate(mesh);}

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +422,63 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                if (hoverPickInfo?.pickedPoint) {
+                    hoverAtOrigin = hoverPickInfo.pickedPoint.x === 0 && hoverPickInfo.pickedPoint.y === 0 && hoverPickInfo.pickedPoint.z === 0;
+                }
+                if (nearHover && !hoverAtOrigin) {   
+                    controllerData.nearHover = true;
+                    pick = hoverPickInfo;
+                }
+            }
+
+            if (controllerData.pickIndexMeshTip) {
```
rickfromwork, line 449: This contains roughly the same code as the hover section above. It shouldn't be too hard to pull it out into a lambda, returning the pick info.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +422,63 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                }
+                if (nearPick && !pickAtOrigin) {
+                    controllerData.nearPick = true;
+                    pick = pickInfo;
+                }
+            }
+
+            if (controllerData.grabRay) {
+                let pinchInfo = this._scene.pickWithRay(controllerData.grabRay, this.nearGrabPredicate);
+                if (pinchInfo && pinchInfo.pickedPoint && pinchInfo.hit) {
```
rickfromwork, line 464: No check for a pick at the origin here?

rickfromwork: Also this section looks like it's close enough to the previous two to share the same lambda

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -724,6 +882,27 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
         // here due to a typo
         return this.laserPointerDefaultColor;
     }
+
+    private _pickWithMesh(indexTipMesh: AbstractMesh, precise: boolean, predicate: (mesh: AbstractMesh) => boolean): Nullable<PickingInfo> {
+        var pickingInfo = new PickingInfo();
```
rickfromwork, line 887: May as well not do this PickingInfo allocation unless we know we can use it. Since you're returning a Nullable object, you can just return null when the pick fails, instead of returning a default picking info. Either that or don't make the return nullable if there's no way to return null.

rickfromwork (changes requested):
Some formatting/nitpicks, but also some issues that need to be addressed.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -390,7 +495,9 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                     Vector3.RotationFromAxisToRef(axis2, pickNormal, axis1, controllerData.selectionMesh.rotation);
                     controllerData.selectionMesh.position.addInPlace(pickNormal.scale(deltaFighting));
                 }
-                controllerData.selectionMesh.isVisible = true && this.displaySelectionMesh;
+                if (!controllerData.nearGrab) {
+                    controllerData.selectionMesh.isVisible = true && this.displaySelectionMesh;
```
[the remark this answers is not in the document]
satyapoojasama: you mean if not false?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -337,13 +366,40 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                                controllerData.pickIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
+                            }
+                            if (controllerData.hoverIndexMeshTip) {
+                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
+                            }
+                            // set near interaction grab ray parameters
+                            let nearGrabRayLength = 0.5;
+                            let indexTipZaxisMultiplier = -1;
+                            controllerData.grabRay.origin.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
+                            controllerData.grabRay.direction.set(indexTipOrientation.x, indexTipOrientation.y, (indexTipOrientation.z) * indexTipZaxisMultiplier);
```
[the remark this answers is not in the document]
satyapoojasama: interesting, even the indexTipPos is a Quaternion and I dont observe anything wrong apparently with the way grab works, may be I should account for 'w'. I actually started using it this way as I observed webXRHandTracking does the same conversion for position.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -558,6 +665,37 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                // squeezeComponent is a new member on the controllerData
+                controllerData.squeezeComponent = motionController.getComponent("grasp");
+
+                // onSqueezeButtonChangedObserver is also a new variable
+                if(controllerData.squeezeComponent)
+                {
+                    controllerData.onSqueezeButtonChangedObserver = controllerData.squeezeComponent.onButtonStateChangedObservable.add((component) => {
+                        if (component.changes.pressed) {
+                            const pressed = component.changes.pressed.current;
+                            if (controllerData.pick) {
```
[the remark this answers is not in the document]
satyapoojasama: Won't the "grasp" take care of filtering differences between nearTouch vs nearGrab? I assumed this will be triggered for grab and not for touch. If that is not the case, I should definitely have a filter

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,14 +422,63 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+            let hoverAtOrigin = false;
+            let hoverRange = false; 
+            if (controllerData.hoverIndexMeshTip) {
+                let hoverPickInfo = this._pickWithMesh(controllerData.hoverIndexMeshTip, false, this.nearPickPredicate);
+                let hoverGrabInfo = this._pickWithMesh(controllerData.hoverIndexMeshTip, false, this.nearGrabPredicate);
+                if ((hoverPickInfo && hoverPickInfo.hit) || (hoverGrabInfo && hoverGrabInfo.hit)) {
+                    hoverRange = true;
+                } 
+                if (hoverRange) {
+                    controllerData.laserPointer.isVisible = false;
```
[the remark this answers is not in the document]
satyapoojasama: So, that was the initial plan - to populate pick info from hover during grab action. But, this is creating more confusion since we just use "pick" for both hover and grab. It will anyways be overwritten by neargrab pick info when it is successful. Although when it is not successful, it is creating problems with pointermove in attachtrackedRayMode.  I will have to still investigate more on this. but yeah if this could be done, that will be neat!

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: grab is using pickWithRay, so it does not issues with origin. the problem for pickwithmesh is that it uses the indextip position that is created at origin when initializing and it is causeing false positive. Pickwithray uses ray and it does not have to deal with this.

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
RaananW: Position is always a vector3, rotation (in WebXR) is always a quaternion.

src/XR/features/WebXRControllerPointerSelection.ts
rickfromwork: Oh, I missed the fact that it uses a different picking method

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -423,42 +423,42 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                 }
             }
 
+            let populateNearInteractionInfo = (nearInteractionInfo: Nullable<PickingInfo>, nearHover: boolean): Nullable<PickingInfo> =>{
+                let result = null;
+                let nearInteractionAtOrigin = false;
+                let nearInteraction = nearInteractionInfo && nearInteractionInfo.pickedPoint && nearInteractionInfo.hit;
+                if (nearInteractionInfo ?.pickedPoint) {
```
rickfromwork, line 430: Does this work with a space before the '?' ? I learned something new...

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
rickfromwork: Oh, yeah. I'm also curious about why you're using a "true && X" format, since that will just resolve to "X" by the compiler.

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -893,14 +888,15 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                    pickingInfo = new PickingInfo();
                     pickingInfo.hit = result;
                     pickingInfo.pickedMesh = mesh;
                     pickingInfo.pickedPoint = indexTipMesh.position;
                     pickingInfo.originMesh = indexTipMesh;
                     pickingInfo.distance = Vector3.Distance(mesh.position, indexTipMesh.position);
                 }
             }
-        }
+        }
```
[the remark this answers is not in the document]
satyapoojasama: I did not end up using the formatter actually. will use it though

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: this code is tested and it works, so i suppose yes

src/XR/features/WebXRControllerPointerSelection.ts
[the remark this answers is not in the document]
satyapoojasama: Honestly no idea, all I did is to move that statement within the if clause. I guess Raanan would have a better idea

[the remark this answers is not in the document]
RaananW: If I'm responsible for those 8 wasted characters, i apologize 😔
No need to do that, it's like writing if(true). You can remove it.

Surprised the compiler never complained 😂

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -384,18 +387,23 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                             let zAxisMultiplier = this._scene.useRightHandedSystem? 1 : -1;
                             const indexTipPos = indexTipPose.transform.position;
                             const indexTipOrientation = indexTipPose.transform.orientation;
+                            controllerData.indexTipQuaternion.set(indexTipOrientation.x, indexTipOrientation.y, (indexTipOrientation.z * zAxisMultiplier), (indexTipOrientation.w * zAxisMultiplier));
```
rickfromwork, line 390: nit: should probably rename zAxisMultiplier to better reflect that it's related to right-hand coordinate systems, since it's used for more than just the z axis

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -384,18 +387,23 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
                             const indexTipPos = indexTipPose.transform.position;
                             const indexTipOrientation = indexTipPose.transform.orientation;
+                            controllerData.indexTipQuaternion.set(indexTipOrientation.x, indexTipOrientation.y, (indexTipOrientation.z * zAxisMultiplier), (indexTipOrientation.w * zAxisMultiplier));
+
                             // set positions for near pick and hover
                             if (controllerData.pickIndexMeshTip) {
                                 controllerData.pickIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
                             }
                             if (controllerData.hoverIndexMeshTip) {
-                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
```
rickfromwork: Why was this removed?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -384,18 +387,23 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, indexTipPos.z);
                             }
+
                             // set near interaction grab ray parameters
                             let nearGrabRayLength = 0.5;
-                            let indexTipZaxisMultiplier = -1;
+                            let webxrZaxisMultiplier = -1;
```
rickfromwork, line 402: Why is this a hard-coded value?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -384,18 +387,23 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                            let indexTipOrientationVector = Vector3.Zero();
                             controllerData.grabRay.origin.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
-                            controllerData.grabRay.direction.set(indexTipOrientation.x, indexTipOrientation.y, (indexTipOrientation.z) * indexTipZaxisMultiplier);
+                            controllerData.indexTipQuaternion.toEulerAnglesToRef(indexTipOrientationVector);
+                            controllerData.grabRay.direction.set(indexTipOrientationVector.x, indexTipOrientationVector.y, (indexTipOrientationVector.z) * webxrZaxisMultiplier);
```
rickfromwork, line 406: This seems to apply a negation on top of the negation applied earlier, cancelling it out?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -384,18 +387,23 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
                             if (controllerData.hoverIndexMeshTip) {
-                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
+                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, indexTipPos.z);
                             }
+
                             // set near interaction grab ray parameters
                             let nearGrabRayLength = 0.5;
-                            let indexTipZaxisMultiplier = -1;
+                            let webxrZaxisMultiplier = -1;
+                            let indexTipOrientationVector = Vector3.Zero();
```
rickfromwork, line 403: Probably good to cache this too, instead of creating a new one for each controller every frame.

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: i think you are viewing outdated file for some reason

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: let me check

satyapoojasama: remember the ray projected from the index tip is in the opposite direction, that is why we negate it here. i dont think it has to do with Right handed system here. that is why i hardcoded it

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: this is not making much difference in observed performance. do we still want to cache it?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -117,9 +119,9 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
             // already attached
             return;
         }
-
+        const disableNearInteraction = this._options.disableNearInteraction as boolean;
```
rickfromwork, line 122: Is there a need to mark this as a boolean here, since you defined the parameter as a boolean above?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -834,21 +837,26 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+            pickIndexMeshTip = SphereBuilder.CreateSphere("IndexPickSphere", { diameter: 1 }, sceneToRenderTo);
+            pickIndexMeshTip.scaling.set(0.03, 0.03, 0.03);
+            pickIndexMeshTip.isVisible = false;
+            
+        }
         return {
             hoverIndexMeshTip,
             pickIndexMeshTip
         }; 
+
```
rickfromwork, line 859: nit: remove extra space

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -834,21 +837,26 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
         };
     }
 
-    private _generateNewHandTipMeshes() {
+    private _generateNewHandTipMeshes(disableNearInteraction: boolean) {
```
rickfromwork, line 840: What is the intended use case for using this flag?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -834,21 +837,26 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
-
-        const hoverIndexMeshTip = SphereBuilder.CreateSphere("IndexHoverSphere", { diameter: 1});
-        hoverIndexMeshTip.scaling.set(0.1, 0.1, 0.1);
-        hoverIndexMeshTip.isVisible = false;
-
-        const pickIndexMeshTip = SphereBuilder.CreateSphere("IndexPickSphere", { diameter: 1 });
-        pickIndexMeshTip.scaling.set(0.03, 0.03, 0.03);
-        pickIndexMeshTip.isVisible = false;
-
+        const sceneToRenderTo = this._options.useUtilityLayer ? this._options.customUtilityLayerScene || UtilityLayerRenderer.DefaultUtilityLayer.utilityLayerScene : this._scene;
```
rickfromwork, line 842: We are only supporting near interaction on **either** the main scene **or** the utility layer scene?

src/XR/features/WebXRControllerPointerSelection.ts
rickfromwork: I guess it's not too important if there's no noticeable perf improvement

src/XR/features/WebXRControllerPointerSelection.ts
rickfromwork: Just saw Raanan's comment from a few weeks back, makes sense. Is it possible to set it up in a way that also allows for disabling/enabling near interaction after creation?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -690,6 +837,28 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
     }
 
+    private _generateNewHandTipMeshes(disableNearInteraction: boolean) {
+        // populate information for near hover, pick and pinch
+        const sceneToRenderTo = this._options.useUtilityLayer ? this._options.customUtilityLayerScene || UtilityLayerRenderer.DefaultUtilityLayer.utilityLayerScene : this._scene;
+        var hoverIndexMeshTip = null;
+        var pickIndexMeshTip = null;
+        if (!disableNearInteraction) {
+            hoverIndexMeshTip = SphereBuilder.CreateSphere("IndexHoverSphere", { diameter: 1}, sceneToRenderTo);
+            hoverIndexMeshTip.scaling.set(0.1, 0.1, 0.1);
```
rickfromwork, line 847: nit: make the scaling values for the two spheres const and pull them out, something like
const _hoverMeshRadius = 0.1;
...
hoverIndexMeshTip.scaling.set(this._hoverMeshRadius, this._hoverMeshRadius, this._hoverMeshRadius);

Then in _onXRFrame, you can scale and use the hover radius for the grabRay length, instead of hard-coding it as well. This way, if we tweak the hover/touch radius down the road, things will just work with a tweak to the const values at the top ofthe class, without needing to dig into the code to see where the values are used.

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: but when undefined it is not setting to a boolean value instead it is assigned to be undefined which the below method would not take in. So, i had to take its value as boolean in order to account for when it is undefined

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -337,13 +366,40 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+                            const indexTipOrientation = indexTipPose.transform.orientation;
+                            // set positions for near pick and hover
+                            if (controllerData.pickIndexMeshTip) {
+                                controllerData.pickIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
+                            }
+                            if (controllerData.hoverIndexMeshTip) {
+                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, (indexTipPos.z * zAxisMultiplier));
+                            }
+                            // set near interaction grab ray parameters
+                            let nearGrabRayLength = 0.5;
```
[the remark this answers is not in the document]
satyapoojasama: you mean the indexTipOrienationVector as well? Its value is modified in toEulerAnglesToRef

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: you mean on every frame? i don't think so

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: sure, what would you like it to be. again, I used this from the way it is used in a different function in the same file :)

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -105,6 +105,9 @@ export interface IWebXRControllerPointerSelectionOptions {
      */
     maxPointerDistance?: number;
 
+    /**
+     * Disable near interaction if not used to improve performance.
```
rickfromwork, line 109: nit: punctuation, or it'll sound like one of the uses for near interaction is to _improve_ performance

src/XR/features/WebXRControllerPointerSelection.ts
rickfromwork: Yeah, I've got no great suggestions. You don't have to change it, but if you did maybe something like sceneToCreateMeshOn or meshCreationScene.

src/XR/features/WebXRControllerPointerSelection.ts
rickfromwork: Ah, that's because disableNearInteraction is an optional parameter. You can usually account for that with "!!", which converts the value to a boolean (essentially the same thing you're doing here).
Typescript usually drops all the "as ..." code when generating the JS files, so I'm concerned that this approach will lose information, since I think "as ..." is basically telling it to interpret the undefined value as a boolean value (a hard cast).

src/XR/features/WebXRControllerPointerSelection.ts
satyapoojasama: i  did rename this to zaxisRHSmultiplier, do you want it to be renamed something else?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -242,11 +247,23 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
      */
     public raySelectionPredicate: (mesh: AbstractMesh) => boolean;
 
-    /** filter used for near interaction pick and hover */
-    public nearPickPredicate(mesh: AbstractMesh): boolean { return mesh.isEnabled() && mesh.isVisible && mesh.isPickable && mesh.isNearPickable};
-    
-    /** filter used for near interaction grab */
-    public nearGrabPredicate(mesh: AbstractMesh): boolean {return mesh.isEnabled() && mesh.isVisible && mesh.isPickable && mesh.isNearGrabbable};
+    /**
```
rickfromwork, line 250: Hm. I don't think the predicate functions here can be locally overwritten in this setup. If we want to allow developers to use their own predicates for near interactions, I think it needs to be set up like so:
public nearGrabPredicate = (...)=>{...};

Otherwise if this stays as a function, there's probably no need to publicly expose it.

satyapoojasama: That is actually true. I  will make this private

RaananW:
Related (draft) PR - https://github.com/BabylonJS/Babylon.js/pull/10363

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+            pickIndexMeshTip,
+            hoverIndexMeshTip,
+            grabRay: new Ray(new Vector3(), new Vector3()),
+            nearInteraction: false,
+            nearGrab: false,
+            nearGrabInProcess: false,
+            id: WebXRNearInteraction._idCounter++,
+        };
+
+        if (this._attachedController) {
```
rickfromwork, line 80: It's a little unclear on what this part does. It looks like it overwrites the attached controller with the current controller ID unless the pointer is only enabled on the other controller?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+    private static _idCounter = 200;
+
+    private _attachController = (xrController: WebXRInputSource) => {
+        if (this._controllers[xrController.uniqueId]) {
+            // already attached
+            return;
+        }
+        const { hoverIndexMeshTip, pickIndexMeshTip } = this._generateNewHandTipMeshes();
+
+        // get two new meshes
```
rickfromwork, line 66: Should this comment go before the call to _generateNewHandTipMeshes()?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+                xrController.inputSource.handedness === this._options.preferredHandedness
+            ) {
+                this._attachedController = xrController.uniqueId;
+            }
+        } else {
+            if (!this._options.enablePointerSelectionOnAllControllers) {
+                this._attachedController = xrController.uniqueId;
+            }
+        }
+        // NOT SURE: if i should return null for gaze and screen modes
```
rickfromwork, line 93: I don't think it makes sense to perform any near interaction logic using the gaze/screen, so since this is now a standalone class, you can return even earlier in those cases, maybe before even calling into _attachController().

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+        if (!super.attach()) {
+            return false;
+        }
+
+        this._options.xrInput.controllers.forEach(this._attachController);
+        this._addNewAttachObserver(this._options.xrInput.onControllerAddedObservable, this._attachController);
+        this._addNewAttachObserver(this._options.xrInput.onControllerRemovedObservable, (controller) => {
+            // REMOVE the controller
+            this._detachController(controller.uniqueId);
+        });
```
rickfromwork, line 168: Why not just have _detachController take in a controller instead of an id?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+
+    /**
+     * Filter used for near interaction pick and hover
+     */
+    private nearPickPredicate(mesh: AbstractMesh): boolean {
+        return mesh.isEnabled() && mesh.isVisible && mesh.isPickable && mesh.isNearPickable;
+    }
+
+    /**
+     * Filter used for near interaction garb
```
rickfromwork, line 241: > garb [](http://example.com/codeflow?start=39&length=5)

nit: grab

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+            }
+        }
+        return null;
+    }
+
+    /**
+     * This function sets webXRControllerPointer Selection feature that will be disabled when
+     * the hover range is reached for a mesh and will be reattached when not in hover range.
+     * This is used to remove the selection rays when moving.
+     * @param farInteractionFeature the feature to disable when finger is in hover range for a mesh
```
rickfromwork, line 227: >e feature to disable when finger is in hover range for a mesh [](http://example.com/codeflow?start=38&length=61)

This is fine, though I'm not the biggest fan on how vaguely it refers to the pointerSelection feature here (since I doubt this would ever be used with a different feature).

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+    private readonly _hoverRadius = 0.1;
+    private readonly _pickRadius = 0.03;
+    private _indexTipQuaternion = new Quaternion();
+    private _indexTipOrientationVector = Vector3.Zero();
+
+    protected _onXRFrame(_xrFrame: XRFrame) {
+        Object.keys(this._controllers).forEach((id) => {
+            // only do this for the selected pointer
+            const controllerData = this._controllers[id];
+            if (!this._options.enablePointerSelectionOnAllControllers && id !== this._attachedController) {
```
rickfromwork, line 256: >enablePointerSelectionOnAllControllers [](http://example.com/codeflow?start=31&length=38)

nit: some naming doesn't make as much sense in the strictly near interaction context, as it talks about the pointer. Maybe something more in-line with the new class name like enableNearInteractionOnAllControllers?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+                        let indexTipPose = _xrFrame.getJointPose!(xrIndexTip, this._xrSessionManager.referenceSpace);
+                        if (indexTipPose && indexTipPose.transform) {
+                            let zAxisRHSMultiplier = this._scene.useRightHandedSystem ? 1 : -1;
+                            const indexTipPos = indexTipPose.transform.position;
+                            const indexTipOrientation = indexTipPose.transform.orientation;
+                            this._indexTipQuaternion.set(
+                                indexTipOrientation.x,
+                                indexTipOrientation.y,
+                                indexTipOrientation.z * zAxisRHSMultiplier,
+                                indexTipOrientation.w * zAxisRHSMultiplier
```
rickfromwork, line 279: nit: drop the "z" from the front since this applies to the w axis as well.

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+                            // set positions for near pick and hover
+                            if (controllerData.pickIndexMeshTip) {
+                                controllerData.pickIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, indexTipPos.z * zAxisRHSMultiplier);
+                            }
+                            if (controllerData.hoverIndexMeshTip) {
+                                controllerData.hoverIndexMeshTip.position.set(indexTipPos.x, indexTipPos.y, indexTipPos.z * zAxisRHSMultiplier);
+                            }
+
+                            // set near interaction grab ray parameters
+                            const nearGrabRayLength = 5 * this._hoverRadius;
```
rickfromwork, line 291: >5 [](http://example.com/codeflow?start=54&length=1)

nit: make this (5) or nearGrabRayLength a const value up alongside this._hoverRadius

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+                }
+                let originalSceneHoverGrab = this._pickWithMesh(controllerData.hoverIndexMeshTip, this._scene, false, this.nearGrabPredicate);
+
+                let hoverPickInfo = accuratePickInfo(originalSceneHoverPick, utilitySceneHoverPick);
+                let hoverGrabInfo = accuratePickInfo(originalSceneHoverGrab, utilitySceneHoverGrab);
+                if ((hoverPickInfo && hoverPickInfo.hit) || (hoverGrabInfo && hoverGrabInfo.hit)) {
+                    // turn off far interaction if in the hover range
+                    if (this._farInteractionFeature) {
+                        this._farInteractionFeature.detach();
+                        hoverRange = true;
```
rickfromwork, line 355: Should this be set in the case where there is no feature to disable?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+                }
+            }
+
+            // Turn on far interaction if near interaction is unsuccessful
+            if (!controllerData.nearInteraction && !hoverRange && !controllerData.nearGrabInProcess) {
+                if (this._farInteractionFeature) {
+                    this._farInteractionFeature.attach();
+                }
+            }
+            controllerData.pick = pick;
```
rickfromwork, line 393: nit: whitespace between closing brace and next "chunk" of logic

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+            pointerType: "xr",
+        };
+        controllerData.onFrameObserver = this._xrSessionManager.onXRFrameObservable.add(() => {
+            if (controllerData.pick) {
+                this._scene.simulatePointerMove(controllerData.pick, pointerEventInit);
+            }
+        });
+        if (xrController.inputSource.gamepad) {
+            const init = (motionController: WebXRAbstractMotionController) => {
+                // squeezeComponent is a new member on the controllerData
```
rickfromwork, line 420: This is one of the comments I added for your reference when implementing, it doesn't make much sense having it stick around

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+                controllerData.meshUnderPointer = null;
+            }
+        });
+    }
+
+    private get _utilityLayerScene() {
+        return this._options.customUtilityLayerScene || UtilityLayerRenderer.DefaultUtilityLayer.utilityLayerScene;
+    }
+
+    private _attachTrackedPointerRayMode(xrController: WebXRInputSource) {
```
rickfromwork, line 407: This function can be reduced and renamed, since it's no longer attaching a tracked pointer ray.

controllerData.SqueezeComponent is what you want for hands, as the grasp event triggers consistently, which the primary trigger triggers only on certain hand orientations (for holoLens, at least). It doesn't make sense to have active listeners to both though, as that should be doubling the input events.

controllerData.SelectComponent would be what we want for other, physical motion controllers, except that the spheres we track hit-testing with are tracking specific hand joints that don't exist on non-hand controllers. So if other hands (such as Oculus) do not send grasp, you would need to listen to this event, otherwise just listening to the squeeze should be more than enough.

Also the event listeners at the bottom can probably be removed, as from what I remember Raanan said they were added to get hand selection working prior to hands gaining the gamepad object.

I remember suggesting some of these extra parts could be kept in to make testing easier while the API has changed, let me know if I mentioned a different reason to keep them that I've now forgotten.

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+        const meshCreationScene = this._options.useUtilityLayer ? this._options.customUtilityLayerScene || UtilityLayerRenderer.DefaultUtilityLayer.utilityLayerScene : this._scene;
+        var hoverIndexMeshTip = null;
+        var pickIndexMeshTip = null;
+        hoverIndexMeshTip = SphereBuilder.CreateSphere("IndexHoverSphere", { diameter: 1 }, meshCreationScene);
+        hoverIndexMeshTip.scaling.set(this._hoverRadius, this._hoverRadius, this._hoverRadius);
+        hoverIndexMeshTip.isVisible = false;
+
+        pickIndexMeshTip = SphereBuilder.CreateSphere("IndexPickSphere", { diameter: 1 }, meshCreationScene);
+        pickIndexMeshTip.scaling.set(this._pickRadius, this._pickRadius, this._pickRadius);
+        pickIndexMeshTip.isVisible = false;
```
rickfromwork, line 552: nit: It's small-scale, but since these two chunks are pretty much the same, they could be lambda'd

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,590 @@
...
+            nearInteraction: boolean;
+            nearGrab: boolean;
+            nearGrabInProcess: boolean;
+            // event support
+            eventListeners?: { [event in XREventType]?: (event: XRInputSourceEvent) => void };
+        };
+    } = {};
+    private _scene: Scene;
+
+    private _attachedController: string;
```
rickfromwork, line 127: >_attachedController [](http://example.com/codeflow?start=12&length=19)

There can only be one controller used for near interaction at a time?

rickfromwork:
Note that since this is a standalone class now, that means we no longer get to use/see the cursor when we have a near interaction, unless you want to re-add that chunk of code (I believe it was the logic around controllerData.selectionMesh)

RaananW:
thanks @rickfromwork for the review



> Note that since this is a standalone class now, that means we no longer get to use/see the cursor when we have a near interaction, unless you want to re-add that chunk of code (I believe it was the logic around controllerData.selectionMesh)

We should be able to connect them both, and disable the rays when in near-interaction mode. Similar to the way the teleportation and pointer selection work seemingly together:

https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/features/WebXRControllerTeleportation.ts#L853-L859

src/XR/features/WebXRNearInteraction.ts
satyapoojasama: Sure, I agree with renaming. Initially i did not have the select component added but added it later for supporting oculus specifically after Raanan got to test with it. @RaananW  Could you please confirm if I can remove event listeners?

src/XR/features/WebXRNearInteraction.ts
satyapoojasama: attachedController is populated only when enablePointerSelectionOnAllControllers is not enabled. In case it is enabled, we have more than a single Controller and it is not populated. It is kind of placeholder for knowing which controller is being used if only single controller is enabled in the scene. Also, this is not new logic that I added, it is used from pointerselection code to support more than one controller at a time.

src/XR/features/WebXRNearInteraction.ts
satyapoojasama: Explained above. in short no, there can be more than one.

src/XR/features/WebXRNearInteraction.ts
satyapoojasama: I understand that it is possible to use other abstract feature in place of this. But the reason i had it as an abstract feature is because of detach problems that occurred when I confine it just to be a pointer selection class. Let me add in a check inside the class to make sure it is actually controller pointer selection feature that is passed in.

src/XR/features/WebXRNearInteraction.ts
satyapoojasama: please refer to detach controller function. It is because the id is required to verify if it is attachedController or not (at the bottom of the function).

src/XR/features/WebXRNearInteraction.ts
rickfromwork: If it's expected that other classes might be used instead of the pointer feature, then it's fine to leave it as-is

src/XR/features/WebXRNearInteraction.ts
rickfromwork: Ah yeah, that was the part I was unsure about. I'm still worried about repeated events on HoloLens though - is it possible to do something where if "grasp" exists listen to squeeze, otherwise listen to the select action?

src/XR/features/WebXRNearInteraction.ts
satyapoojasama: I really dont find big difference passing a controller vs id in this context

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -0,0 +1,592 @@
...
+            let resultMesh = null;
+            resultMesh = SphereBuilder.CreateSphere(name, { diameter: 1 }, sceneToUse);
+            resultMesh.scaling.set(scale, scale, scale);
+            resultMesh.isVisible = visibility;
+
+            return resultMesh;
+        };
+
+        hoverIndexMeshTip = createSphereMesh("IndexHoverSphere", this._hoverRadius, false, meshCreationScene);
+        pickIndexMeshTip = createSphereMesh("IndexPickSphere", this._pickRadius, false, meshCreationScene);
```
rickfromwork, line 554: nit: probably don't need to pass in visibility as a parameter since it's always false.

src/XR/features/WebXRNearInteraction.ts
rickfromwork: Gotcha, thanks for the overview

rickfromwork (approved):

RaananW:
I pulled this and am working on the build errors. The issue is the new properties that were added to the mesh class.

src/XR/features/WebXRNearInteraction.ts
RaananW: I find it the best as it is. Resolving this.

src/XR/features/WebXRNearInteraction.ts
RaananW: Isn't that the case right now? This is what I do in the pointer selection - if no gamepad is available, use the events.

RaananW: I don't see a case where events will be repeated. And hololens will never go into the event listener registration loop

RaananW (approved):

PR #10231, opened 2021-04-21
Updating the XR components to handle the latest webXR changes for hand input

### LINKED ISSUE #10144: WebXR sends select/squeeze events for hands in a gamepad object now
Need to update Babylon.JS to not assume a gamepad object always implies a motion controller is used, as that causes it to create models for the controllers (and probably other things as well) that hands do not support.

rickfromwork:
Hands now pass information for two "buttons" through the gamepad object, and so it no longer makes sense to load controller models if and only if a gamepad component exists. We now have to check if the controller is a hand or not (and if so, don't load the model).

This is done by creating a motion controller class for the hands that uses the hand-specific profile strings, and tells the controller manager to skip loading the model. Also adds some logic around other classes that relied on the gamepad component for determining if a model existed.

Resolves #10144


src/XR/motionController/webXRGenericHandController.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * The motion controller class for the standard HTC-Vive controllers
+ */
+export class WebXRGenericHandController extends WebXRAbstractMotionController {
+
+    public profileId = "generic-hand-select-grasp";
+
+    /**
+     * Create a new Vive motion controller object
```
RaananW, line 16: nit - documentation. WOuld be great to document the class correctly for the typedoc page

src/XR/motionController/webXRGenericHandController.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    protected _processLoadedModel(_meshes: AbstractMesh[]): void {
+        // nothing to do
+    }
+
+    protected _setRootMesh(meshes: AbstractMesh[]): void {
+        this.rootMesh = new Mesh(this.profileId + " " + this.handedness, this.scene);
+
+        meshes.forEach((mesh) => {
```
RaananW, line 44: probably not needed

RaananW (commented):
LGTM!
Just one question - this profile exists in the webxr-input repository, which we load per default. Won't it load this profile automatically from the repository if the profile is set in the XRInputSource?

satyapoojasama (approved):

src/XR/motionController/webXRGenericHandController.ts
rickfromwork: Oh jeez I forgot to update the comments

src/XR/motionController/webXRGenericHandController.ts
rickfromwork: > LGTM!
> Just one question - this profile exists in the webxr-input repository, which we load per default. Won't it load this profile automatically from the repository if the profile is set in the XRInputSource?

It looks like the motion controller manager's UpdateProfilesList() pulls from the list found here:
https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/profiles/profilesList.json
But this doesn't hold any data for hands, since it looks like a list of controller models.

rickfromwork:
> LGTM!
> Just one question - this profile exists in the webxr-input repository, which we load per default. Won't it load this profile automatically from the repository if the profile is set in the XRInputSource?

It looks like the UpdateProfilesList() call in the controller manager is pulling from the following spot, which only holds data for controller models. Since there's no model for hands, there's no entry for them.
https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/profiles/profilesList.json

src/XR/motionController/webXRAbstractMotionController.ts
```diff
@@ -291,7 +291,10 @@ export abstract class WebXRAbstractMotionController implements IDisposable {
          * handedness (left/right/none) of this controller
          */
         public handedness: MotionControllerHandedness,
-        _doNotLoadControllerMesh: boolean = false
+        /**
+         * @hidden
+         */
+        public _doNotLoadControllerMesh: boolean = false
```
chrisfromwork, line 297: >_doNotLoadControllerMesh [](http://example.com/codeflow?start=15&length=24)

if this is public should it be renamed?

src/XR/motionController/webXRGenericHandController.ts
```diff
@@ -0,0 +1,138 @@
+import { IMotionControllerLayoutMap, IMinimalMotionControllerObject, MotionControllerHandedness, WebXRAbstractMotionController } from "./webXRAbstractMotionController";
+import { Scene } from "../../scene";
+import { AbstractMesh } from "../../Meshes/abstractMesh";
+import { Mesh } from "../../Meshes/mesh";
+import { Quaternion } from "../../Maths/math.vector";
+import { WebXRMotionControllerManager } from "./webXRMotionControllerManager";
+
+/**
+ * The motion controller class for the standard HTC-Vive controllers
```
chrisfromwork, line 9: > HTC-Vive contro [](http://example.com/codeflow?start=47&length=16)

nit: is this file really htc-vive specific or should this comment be updated?

src/XR/features/WebXRControllerPhysics.ts
```diff
@@ -87,46 +87,55 @@ export class WebXRControllerPhysics extends WebXRAbstractFeature {
         // if no motion controller available, create impostors!
         if (this._options.physicsProperties!.useControllerMesh && xrController.inputSource.gamepad) {
             xrController.onMotionControllerInitObservable.addOnce((motionController) => {
-                motionController.onModelLoadedObservable.addOnce(() => {
-                    const impostor = new PhysicsImpostor(motionController.rootMesh!, PhysicsImpostor.MeshImpostor, {
-                        mass: 0,
-                        ...this._options.physicsProperties,
+                if (!motionController._doNotLoadControllerMesh) {
+                    motionController.onModelLoadedObservable.addOnce(() => {
+                        const impostor = new PhysicsImpostor(motionController.rootMesh!, PhysicsImpostor.MeshImpostor, {
```
rgerd, line 92: nit: You could put each parameter on its own line instead of having a long line here

src/XR/features/WebXRControllerPhysics.ts
```diff
@@ -87,46 +87,55 @@ export class WebXRControllerPhysics extends WebXRAbstractFeature {
...
-                    const controllerMesh = xrController.grip || xrController.pointer;
-                    this._controllers[xrController.uniqueId] = {
-                        xrController,
-                        impostor,
-                        oldPos: controllerMesh.position.clone(),
-                        oldRotation: controllerMesh.rotationQuaternion!.clone(),
-                    };
-                });
+                } else {
+                    // This controller isn't using a model, create imposters instead
```
rgerd, line 105: *impostor

src/XR/features/WebXRControllerPhysics.ts
```diff
@@ -87,46 +87,55 @@ export class WebXRControllerPhysics extends WebXRAbstractFeature {
...
-            this._controllers[xrController.uniqueId] = {
-                xrController,
-                impostor,
-                impostorMesh,
-            };
+            this._createPhysicsImposter(xrController);
         }
     };
 
+    private _createPhysicsImposter = (xrController: WebXRInputSource) => {
```
chrisfromwork, line 114: >_createPhysicsImposter [](http://example.com/codeflow?start=12&length=22)

nit: can this just be a tradditional member function compared to a declared variable?

src/XR/motionController/webXRAbstractMotionController.ts
rickfromwork: It should be a hidden public member, with the @hidden tag on it, though I've never seen it used inline in this way. @RaananW just double-checking, this would work as expected and hide the member from the public, right?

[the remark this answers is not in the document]
rickfromwork: Probably. I was sticking with the format seen elsewhere though, so I'm tempted to not change it.

src/XR/features/WebXRControllerPhysics.ts
rickfromwork: Oh hah, my word wrap did that visually for me so I didn't even notice.

src/XR/motionController/webXRAbstractMotionController.ts
RaananW: yes, this is "internal public". Underscore means - don't use outside of the framework. There is mainly because there is really no private and public in JS (in ES5 to be exact)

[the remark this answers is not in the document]
RaananW: yeah, it can get confusing, but as @rickfromwork said - since it is an optional parameter in the constructor it should stay this way ("falsy" being the default).

RaananW (approved):
LGTM!

PR #10237, opened 2021-04-21
Add KHR_materials_clearcoat serialization

### LINKED ISSUE #10181: Implement KHR_Materials_Clearcoat serialization
See https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_clearcoat for the full schema.

Originally implemented in the loader via https://github.com/BabylonJS/Babylon.js/pull/7006

Since more people are looking to Babylon.js as an asset assembling/modification tool, we should support this as part of our effort to support full I/O support of the glTF spec.

Drigax:
Implement KHR_material_clearcoat exporter extension class. 

Add additional logic to avoid pruning of UV coordinates if extensions include textures.

Fix #10181


serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -181,9 +181,33 @@ export class _GLTFMaterialExporter {
             }
         }
 
+        // Do a naieve check for materials that contain extensions that have texture-related members.
```
Drigax, line 184: I'm not a fan of this implementation, but it seems effective. I also considered adding a mapping if which IMaterials include textures that is created as the IMaterials are created, I can change if there are any serious considerations of the efficiency of iterating over all members in our exported materials.

This could use some unit tests for the various extensions that include texture information.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -181,9 +181,33 @@ export class _GLTFMaterialExporter {
...
+                        }
+                    }
+                }
+            }
+        }
+
         return false;
     }
 
+    public _getTextureInfo(babylonTexture: Nullable<BaseTexture>): Nullable<ITextureInfo> {
```
Drigax, line 201: Its actually a little strange that other extensions like `KHR_materials_sheen` populate their own list of exported textures, that doesn't seem to scale well when we get to a point where materials may have multiple shared textures used across multiple extensions for the same material. Not a huge concern now, but something I want to make sure I take a note of.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public postExportMaterialAdditionalTextures?(context: string, node: IMaterial, babylonMaterial: Material): BaseTexture[] {
+        let additionalTextures: BaseTexture[] = [];
+        if (babylonMaterial instanceof PBRMaterial) {
+            if (babylonMaterial.clearCoat.isEnabled) {
+                if (babylonMaterial.clearCoat.texture) {
+                    additionalTextures.push(babylonMaterial.clearCoat.texture);
+                }
+                if (!babylonMaterial.clearCoat.useRoughnessFromMainTexture && babylonMaterial.clearCoat.textureRoughness) {
```
bghgary, line 49: What about `remapF0OnInterfaceChange`?

Also, it looks like the textures will be ignored if the flags are not set like it is necessary for glTF. Either we should handle those cases or we should throw a warning or something to indicate that they weren't exported.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+                if (!babylonMaterial.clearCoat.isEnabled) {
+                    resolve(node);
+                    return;
+                }
+
+                this._wasUsed = true;
+
+                if (node.extensions == null) {
+                    node.extensions = {};
+                }
```
bghgary, line 74: nit: can be simplified if you want

```suggestion
                node.extensions = node.extensions || {};
```

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+
+                let clearCoatTextureInfo = this._exporter._glTFMaterialExporter._getTextureInfo(babylonMaterial.clearCoat.texture);
+                let clearCoatTextureRoughnessInfo;
+                if (babylonMaterial.clearCoat.useRoughnessFromMainTexture) {
+                    clearCoatTextureRoughnessInfo = this._exporter._glTFMaterialExporter._getTextureInfo((babylonMaterial as PBRMaterial).metallicTexture
+                                                                                                        ?? (babylonMaterial as PBRMetallicRoughnessMaterial).metallicRoughnessTexture);
```
bghgary, line 80: This is unfortunate. I wonder if we can expose `_metallicTexture` somehow instead of having to switch. The two properties both point to `_metallicTexture`.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+                } else {
+                    clearCoatTextureRoughnessInfo = this._exporter._glTFMaterialExporter._getTextureInfo(babylonMaterial.clearCoat.textureRoughness);
+                }
+
+                let clearCoatNormalTextureInfo = this._exporter._glTFMaterialExporter._getTextureInfo(babylonMaterial.clearCoat.bumpTexture);
+
+                const clearCoatInfo: IKHRMaterialsClearcoat = {
+                    clearcoatFactor: babylonMaterial.clearCoat.intensity,
+                    clearcoatTexture: clearCoatTextureInfo ?? undefined,
```
bghgary, line 89: I don't think this is what you want.

`foo.bar = undefined` is not the same as `foo.bar` doesn't exist. I believe this will serialize into `{ bar: undefined }`.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -181,9 +181,33 @@ export class _GLTFMaterialExporter {
...
+                let extensionObject = material.extensions[extension];
+                for (let member in extensionObject){
+                    if (member.indexOf("Texture") !== -1){
+                        if (extensionObject[member]){
+                            return true;
+                        }
+                    }
+                }
+            }
+        }
```
bghgary, line 196: Hmm, are you sure you want to leave it like this? It seems better if the code asks the extension code whether there are textures or not?

serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -181,9 +181,33 @@ export class _GLTFMaterialExporter {
             }
         }
 
+        // Do a naive check for materials that contain extensions that have texture-related members.
+        if (material.extensions) {
+            for (let extension in material.extensions){
```
bghgary, line 186: nit:
```suggestion
            for (let extension in material.extensions) {
```

Might consider formatting the file.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: So far that doesn't seem to be the case, undefined/nulled fields appear to not be serialized. I'm not sure if thats a feature of the JSON serializer, or something we're intentionally doing.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: I suppose, but this feels right as an external consumer of these materials

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    public dispose() {
+    }
+
+    /** @hidden */
+    public get wasUsed() {
+        return this._wasUsed;
+    }
+
+    public postExportMaterialAdditionalTextures?(context: string, node: IMaterial, babylonMaterial: Material): BaseTexture[] {
```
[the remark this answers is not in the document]
Drigax: good question, this started as a copy of the sheen serializer, which was probably a copy of the interface, so this syntax seemed to propagate.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: > Also, it looks like the textures will be ignored if the flags are not set like it is necessary for glTF

Can you expand on this? I'm not sure what you mean here.


> What about remapF0OnInterfaceChange?

I dont understand, why would this affect the other serializable fields? It seems like something that we just can't handle.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Yeah, I considered that, as well ^^^ https://github.com/BabylonJS/Babylon.js/pull/10237#discussion_r617731942 as a mapping. Most likely this can be done as an interface inherited by the other material extension interfaces.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
bghgary: Yup, confirmed with JSON.stringify.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
bghgary: > > Also, it looks like the textures will be ignored if the flags are not set like it is necessary for glTF
>
> Can you expand on this? I'm not sure what you mean here.

When serializing, we shouldn't just drop textures if they are not compatible with glTF. Either we need to convert them to work with glTF or log/throw some kind of error/warning.

>> What about remapF0OnInterfaceChange?
> 
> I dont understand, why would this affect the other serializable fields? It seems like something that we can't express and should drop, unless I'm missing something.

The loader sets this flag. I don't know exactly what this flag does, but do we need to check it here? If this isn't set, is the material compatible with glTF?

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
bghgary: Doesn't feel right to me. We shouldn't have to cast to the leaf type in order to access common properties. If another child material were added that remaps `_metallicTexture` to something else, we will have to add it here.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: > When serializing, we shouldn't just drop textures if they are not compatible with glTF. Either we need to convert them to work with glTF or log/throw some kind of error/warning.

Can you be more specific? I still don't follow what you're getting at and why its relevant here

Drigax: If you're referring to `pbrClearCoatConfiguration._tintTexture` being dropped, I can add a warning indicating that it's not supported. This texture specifically has no analog.

bghgary: If `babylonMaterial.clearCoat.useRoughnessFromMainTexture` is true, then the roughness texture will be missing in the glTF. Right?

bghgary: I'm not familiar with this code to know whether this is expected or not. Looks like it's being used in `_finishMaterial` and `_exportTextureAsync` will be called on it. If `useRoughnessFromMainTexture` is true, it looks like the main `texture` won't be exported since it's not in the list. Even if you add it to the list, it will still be wrong since this won't match the glTF spec. Have you tried exporting a clearcoat material where `useRoughnessFromMainTexture` is true?

deltakosh:
@bghgary ok to merge?

bghgary (changes requested):
Still have outstanding comments.

deltakosh:
up @Drigax

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Ended up adding a base class to our glTF interface, now when we initialize these serialized members, we'll define a function. I don't think this function gets serialized by default, but I'll check.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: Exposed the PBRBaseMaterial properties.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -12,10 +12,9 @@ import { RawTexture } from "babylonjs/Materials/Textures/rawTexture";
 import { Material } from "babylonjs/Materials/material";
 import { StandardMaterial } from "babylonjs/Materials/standardMaterial";
 import { PBRBaseMaterial } from "babylonjs/Materials/PBR/pbrBaseMaterial";
-import { PBRBaseSimpleMaterial } from "babylonjs/Materials/PBR/pbrBaseSimpleMaterial";
 import { PBRMaterial } from "babylonjs/Materials/PBR/pbrMaterial";
 import { PBRMetallicRoughnessMaterial } from "babylonjs/Materials/PBR/pbrMetallicRoughnessMaterial";
-import { PBRSpecularGlossinessMaterial } from "babylonjs/Materials/PBR/pbrSpecularGlossinessMaterial";
+import { IMaterialExtension } from  'babylonjs-gltf2interface';
```
bghgary, line 17: nit:
```suggestion
import { IMaterialExtension } from  "babylonjs-gltf2interface";

```

You can change the defaults in vscode:
image

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+        let additionalTextures: BaseTexture[] = [];
+        if (babylonMaterial instanceof PBRMaterial) {
```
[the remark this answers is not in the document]
Drigax: Done.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
```diff
@@ -573,7 +569,7 @@ export class _GLTFMaterialExporter {
             else {
                 resizedTexture1 = this._createWhiteTexture(texture2Size.width, texture2Size.height, scene);
             }
-            resizedTexture2 = texture2;
+            resizedTexture2 = texture2!;
```
bghgary, line 572: Comment of this function says:
> If both textures are null, returns null

That doesn't seem to be the case. If it were the case, then you shouldn't need this `!` here.

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -363,26 +363,30 @@ export abstract class PBRBaseMaterial extends PushMaterial {
     /**
      * Intensity of the direct lights e.g. the four lights available in your scene.
      * This impacts both the direct diffuse and specular highlights.
+     * @hidden
      */
-    protected _directIntensity: number = 1.0;
+    public _directIntensity: number = 1.0;
```
bghgary, line 368: I don't know about this. We should check with Seb but he is out on vacation for a while. @deltakosh or @RaananW Do you have an opinion on this?

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
```diff
@@ -0,0 +1,102 @@
...
+    public dispose() {
+    }
+
+    /** @hidden */
+    public get wasUsed() {
+        return this._wasUsed;
+    }
+
+    public postExportMaterialAdditionalTextures?(context: string, node: IMaterial, babylonMaterial: Material): BaseTexture[] {
+        let additionalTextures: BaseTexture[] = [];
```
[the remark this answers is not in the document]
Drigax: Done.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Refactored sheen to be less strange, reused my material exporter helper functions to simplify how to get the textureInfo for a given texture this extension adds.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Refactored the implementation to be more object oriented.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Our style is all over the place in this source.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: Its needed, this comment is wrong. This whole function is wrong. But i did the minimum to get it to work and pass static analysis.

src/Materials/PBR/pbrBaseMaterial.ts
Drigax: I asked seb before he left. He suggested this as our simplest way to expose this functionality.

serializers/src/glTF/2.0/glTFMaterialExporter.ts
Drigax: fixed

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: Yes, it actually works when we point it to the proper texture. I don't believe we should be including this texture explicitly when useRoughessFromMainTexture is enabled, however. we should just try to find the texture if it exists. Similarly to the sheen implementation.

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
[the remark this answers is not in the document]
Drigax: removed from this and `KHR_materials_sheen`

serializers/src/glTF/2.0/Extensions/KHR_materials_clearcoat.ts
Drigax: Also added warnings for if we can't export features out of glTF's spec.

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1079,12 +1137,6 @@ export abstract class PBRBaseMaterial extends PushMaterial {
                         }
                     }
```
Drigax, line 1138: ugh bad merge. Fixing...

tests/validation/config.json
```diff
@@ -1222,57 +1296,75 @@
...
+                "webgl1"
+            ],
             "referenceImage": "instances-renderself-mb.png"
         },
         {
             "title": "Thin instances + render self motion blur",
             "renderCount": 10,
             "playgroundId": "#217750#35",
-            "excludedEngines": ["webgl1"],
+            "excludedEngines": [
```
Drigax, line 1364: hmm, looks like the autoformatter is expanding the arrays, however I'm not super oppsed to this change in format...

Drigax: Ehh, nevermind.

src/Materials/PBR/pbrBaseMaterial.ts
Drigax: fixed,

bghgary (approved):

PR #10239, opened 2021-04-21
[GE] Bug fix for panning

msDestiny14:
Removed extra HTML divs that were causing the blur function to be called.


msDestiny14:
Not sure where the extra code is coming from? Shouldn't this already be in master? @deltakosh

guiEditor/src/diagram/workbench.tsx
```diff
@@ -344,11 +335,12 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
                 scene.onPointerObservable.add(panningFn, PointerEventTypes.POINTERMOVE);
                 this._panning = true;
             } else {
-                this._panning = false;
+               this._panning = false;
```
sebavan, line 338: NIT: looks like indentation is not the same as line 336 :-)

guiEditor/src/diagram/workbench.tsx
```diff
@@ -199,9 +192,7 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
     }
 
     componentDidMount() {
-        this._rootContainer = this.props.globalState.hostDocument.getElementById("workbench-container") as HTMLDivElement;
-        this._gridCanvas = this.props.globalState.hostDocument.getElementById("workbench-canvas-container") as HTMLDivElement;
-        this._svgCanvas = this.props.globalState.hostDocument.getElementById("workbench-svg-container") as HTMLElement;
+        this._rootContainer = this.props.globalState.hostDocument.getElementById("workbench-canvas") as HTMLDivElement;
```
sebavan, line 195: Why not using React Ref here as the element is part of the same component ? (won t change a lot cause there can only be one workbench per page in this case but still sounds like a good practice)

sebavan (approved):
Just nit comments

guiEditor/src/diagram/workbench.tsx
msDestiny14: OH! ty!

PR #10242, opened 2021-04-22
onMeshReadyObservable

RaananW:
Will notify if a mesh is completely ready.

The reason it is done this way is to avoid the unneeded work (is completely ready) if no callback was registered.

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -243,6 +243,13 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
     // Internal data
     private _internalMeshDataInfo = new _InternalMeshDataInfo();
 
+    private _isCompletelyReady: boolean = false;
+
+    /**
+     * Will notify when the mesh is completely ready, including materials.
+     */
+    public onMeshReadyObservable: Observable<Mesh>;
+
     public get computeBonesUsingShaders(): boolean {
         return this._internalAbstractMeshDataInfo._computeBonesUsingShaders;
     }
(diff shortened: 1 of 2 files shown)


src/Meshes/mesh.ts
```diff
@@ -1155,6 +1181,28 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         return true;
     }
 
+    private _checkReadinessObserver: Nullable<Observer<Scene>>;
+
+    private _onMeshReadyObserverAdded = (observer: Observer<Mesh>) => {
+        // only notify once! then unregister the observer
+        observer.unregisterOnNextCall = true;
+        if (this._isCompletelyReady) {
```
Popov72, line 1189: A mesh can be ready at some point, then its state changed and so not be ready anymore for some frames until it is ready again. With the current implementation, once it has been ready at some point it will always fire the `MeshReady` event (if someone registers for it) even if it becomes not ready.

Maybe `this._isCompletelyReady` should be reset to `false` here and the first part of the `if` removed...

[...] In fact, `this._isCompletelyReady` could be completely removed. In that case, you could do a function `executeWhenReady` instead (much like `scene.executeWhenReady`) and avoid creating new properties (so not hitting the 128 max properties limit). `Mesh.ExecuteWhenReady` could maybe have some value too, that would wait for an array of meshes to be ready.

RaananW: Discussed offline :-)

I agree with the fact that the private _isCompletelyReady is not needed and have removed it, moved the other privates to internal. I now add only one extra parameter

deltakosh (approved):

PR #10249, opened 2021-04-22
DeviceInputSystem: Fix pointer events when cursor leaves canvas

PolygonalSun:
This PR changes the behavior of the pointer events to better emulate the behavior from 4.2.  The pointerout event listener code has been removed and the pointerup eventlistener has been updated to use the window as the target to catch any pointer button releases outside of the canvas.


RaananW (approved):
~~Won't native have a problem with the window object?~~

Answered myself, all good :-)

PR #10251, opened 2021-04-23
Persistent "disable level multiplication" switch

rassie:
Follow-up to 4a17889b7118fad8edbc934fc8fbf1d4d433dd75 (PR #10192). "Disable level multiplication" needs to be shown even if no texture is loaded, since otherwise developing a material from NME becomes more difficult.


Popov72:
LGTM

Popov72 (approved):
LGTM

PR #10259, opened 2021-04-26
Fix PhysicsImposter for negatively scaled mesh

wilcoschoneveld:
Hi!

This is my first pull request and I would like to contribute!

When importing glb (gltf) files from Blender I've noticed the meshes are scaled negatively in one axis. As a consequence the physics imposters on these meshes are not created correctly. In example playgrounds ([#66PS52](https://playground.babylonjs.com/#66PS52), [#DGEP8N](https://playground.babylonjs.com/#DGEP8N)) this is 'corrected' with the following code:

```
m.scaling.x = Math.abs(m.scaling.x);
m.scaling.y = Math.abs(m.scaling.y);
m.scaling.z = Math.abs(m.scaling.z);
```

There are two problems with this workaround:
1. It forces the user to flip the mesh, thereby not allowing the user to use the mesh in the original orientation.
2. The calculated mesh center coordinate is wrong, this is apparent when using a non-symmetric mesh.

**Minimal playground reproduction of the issue: https://playground.babylonjs.com/#TWVYMQ#5**

In the playground, you can immediately see that the calculated physics imposter is wrong. You can try the workaround by enabling line 28 (`cube.scaling.y = -cube.scaling.y;`) but you'll see that the box is aligned incorrectly, also the mesh is flipped which is not what I want.

The proposed solution is to adjust for the sign of the scaling when calling `getObjectExtendSize` and `getObjectCenter`. This seems to resolve my issue.

I haven't found a nice `Vector3` helper method for this operation, any ideas?

What are your thoughts?

Thanks a lot!

Related discussion: https://forum.babylonjs.com/t/how-to-correctly-import-gltf-and-add-physics/20234/3

### DIFF
--- a/src/Physics/physicsImpostor.ts
+++ b/src/Physics/physicsImpostor.ts
@@ -621,6 +621,10 @@ export class PhysicsImpostor {
             const boundingInfo = this.object.getBoundingInfo();
             // get the global scaling of the object
             const size = boundingInfo.boundingBox.extendSize.scale(2).multiplyInPlace(scaling);
+            // correct the size for negative scaling (e.g. flipped mesh)
+            size.x *= this.object.scaling.x < 0 ? -1 : 1;
+            size.y *= this.object.scaling.y < 0 ? -1 : 1;
+            size.z *= this.object.scaling.z < 0 ? -1 : 1;
             //bring back the rotation
             this.object.rotationQuaternion = q;
             //calculate the world matrix with the new rotation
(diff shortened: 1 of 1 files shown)


RaananW:
Thanks :-)

You are totally right, we don't take the negative scaling into account.

Does it work with mesh impostors? The reason for the negative z is to allow left-handed systems to show right-handed GLTFs. the combination of rotation and negative scaling generates the right result without changing the glTF object itself. I guess ignoring the scaling will work well with spheres and boxes, but I wonder how will more complex impostors react  to the change?

wilcoschoneveld:
Cool, thanks for the reply!

I'll investigate the impact of this change on mesh imposters and other cases. I don't have time tomorrow but I will check in on wednesday.

wilcoschoneveld:
I had another idea to _work around_ this issue by baking the scaling into the mesh, like so: https://playground.babylonjs.com/#TWVYMQ#6

This also seems to work, which eliminates the need for this PR. But I'm not sure about any downsides/side-effects...

Any thoughts?

RaananW:
> I had another idea to _work around_ this issue by baking the scaling into the mesh

This is what i usually do when i have complex transformation on meshes. Babylon should support it out of the box, so it does make sense to find a solution that fits all cases. If you want to look into it, it would be great! if you don't find a solution, would be great if you created an issue and describe the issue. We'll fix that for 5.0

wilcoschoneveld:
Thanks @RaananW for the comment.

I'm not yet proficient enough with BabylonJS to feel confident about the fix, so I'm focusing on gaining some experience first. I've created an issue about this: https://github.com/BabylonJS/Babylon.js/issues/10283

PR #10261, opened 2021-04-26
KHR_materials_transmission: Fix crash when using instanced meshes

Popov72:
See https://forum.babylonjs.com/t/gltf-importer-errors-on-load/20214/6

### DIFF
--- a/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
+++ b/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
@@ -12,6 +12,7 @@ import { Texture } from "babylonjs/Materials/Textures/texture";
 import { RenderTargetTexture } from "babylonjs/Materials/Textures/renderTargetTexture";
 import { Observable, Observer } from "babylonjs/Misc/observable";
 import { Constants } from "babylonjs/Engines/constants";
+import { Tools } from "babylonjs/Misc/tools";
 
 interface ITransmissionHelperHolder {
     /**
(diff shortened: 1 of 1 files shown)


deltakosh (approved):

loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -150,11 +151,16 @@ class TransmissionHelper {
...
     private _addMesh(mesh: AbstractMesh): void {
         this._materialObservers[mesh.uniqueId] = mesh.onMaterialChangedObservable.add(this.onMeshMaterialChanged.bind(this));
-        if (this.shouldRenderAsTransmission(mesh.material)) {
-            this._transparentMeshesCache.push(mesh);
-        } else {
-            this._opaqueMeshesCache.push(mesh);
-        }
+        // we need to defer the processing because _addMesh may be called as part as an instance mesh creation, in which case some
+        // internal properties are not setup yet, like _sourceMesh (needed when doing mesh.material below)
+        Tools.SetImmediate(() => {
```
bghgary, line 156: Does this cause one frame to be rendered incorrectly?

Popov72: No, that causes a crash when doing `mesh.material` because `mesh._sourceMesh` is not setup yet (in the `InstancedMesh` constructor) when the `onNewMesh` observable is triggered.

Popov72: I think I didn't understand your question...

Yes, the first frame won't have the refraction texture setup for the transparent meshes. However, it is likely that the materials will also load some textures, so the material won't be ready the first few frame(s) anyway, meaning the meshes won't be rendered and so this problem won't show up.

bghgary (approved):

PR #10263, opened 2021-04-27
Compute BoundingInfo for morph target

EricBeetsOfficial-Opuscope:


EricBeetsOfficial-Opuscope:
Added support for morph targets to the mesh `BoundingInfo` refresh as mentioned [here](https://forum.babylonjs.com/t/blendshape-morph-target-bounding-box-mesh-baking/20261/2?u=eric_beets)

src/Meshes/mesh.ts
```diff
@@ -1302,18 +1302,49 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
 
+    /** @hidden */
+    public _getPositionData(applySkeleton: boolean, applyMorph: boolean): Nullable<FloatArray> {
+        var data = super._getPositionData(applySkeleton, applyMorph);
+
+        if (data && applyMorph && this.morphTargetManager) {
+            data = Tools.Slice(data);
+            this._generatePointsArray();
+
+            var influencedPositions = new Float32Array(data.length);
```
RaananW, line 1326: will that have any implications on garbage collection? how often is that expected to run?

RaananW (commented):
Thanks for the PR! Makes total sense :-)

Just have one question around garbage collection

src/Meshes/mesh.ts
EricBeetsOfficial-Opuscope: Hi,
You're totally right, beyond the memory consumption, this update will consumme CPU cycles. I think `refreshBoundingInfo` hasn't to be called for each frame !

To be honest, I tried to reproduce the `AbstractMesh._getPositionData` schema (which is called just before). If you think `_generatePointsArray`/`ToolsSlice` is too much (overkill), we can remove it..

bghgary: I think it should be possible to not need this specific allocation. If you make the outer loop the vertex index, and loop through the morphs inside, then you shouldn't need to allocate another buffer for the influenced positions.

src/Meshes/mesh.ts
```diff
@@ -1302,18 +1302,41 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
+        var data = super._getPositionData(applySkeleton, applyMorph);
+
+        if (data && applyMorph && this.morphTargetManager) {
+
+            for (let vertexCount = 0; vertexCount < data.length; vertexCount++) {
+
+                for (let targetCount = 0; targetCount < this.morphTargetManager.numTargets; targetCount++) {
+                    var influence = this.morphTargetManager.getTarget(targetCount).influence;
+                    if (influence > 0.0) {
+                        var morphTargetPositions = this.morphTargetManager.getTarget(targetCount).getPositions();
```
RaananW, line 1329: even thou it is 0(1) operation, could you reuse the `this.morphTargetManager.getTarget(targetCount)` call? something like:

```suggestion
                        const target = this.morphTargetManager.getTarget(targetCount);
                            if(target.influence > 0) {
                                const morphTargetPositions = target .getPositions();
```

RaananW (commented):
Small comments, otherwise LGTM! I will let Gary comment as well before merging.

src/Meshes/mesh.ts
```diff
@@ -1302,18 +1302,42 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
-        this._refreshBoundingInfo(this._getPositionData(applySkeleton), bias);
+        this._refreshBoundingInfo(this._getPositionData(applySkeleton, applyMorph), bias);
         return this;
     }
 
+    /** @hidden */
+    public _getPositionData(applySkeleton: boolean, applyMorph: boolean): Nullable<FloatArray> {
+        var data = super._getPositionData(applySkeleton, applyMorph);
+
+        if (data && applyMorph && this.morphTargetManager) {
```
bghgary, line 1322: We still need this when the super function doesn't do it:
```javascript
            data = Tools.Slice(data);
            this._generatePointsArray();
```

Though I'm not sure what is the best way to test this. Maybe recheck `(data && applySkeleton && this.skeleton)`?

bghgary (changes requested):

src/Meshes/mesh.ts
EricBeetsOfficial-Opuscope: You're.. I didn't see that ..
What do you think of:
```
        var data = super._getPositionData(applySkeleton, applyMorph);

        if (data) {
            if (!applySkeleton || !this.skeleton) {
                data = Tools.Slice(data);
                this._generatePointsArray();
            }

            if (applyMorph && this.morphTargetManager) {

                // ... 
            }
        }
        return data;
```

bghgary: Yeah, I guess this works. It's a bit dangerous since if someone modifies the logic in the super function, it's not obvious that the child function needs to change, but I don't see a clean way around this. We could maybe pass an out parameter (via an object) to the super function so that we know if the super function has already copied the data, but that's clunky.

RaananW: maybe a  `protected` flag that being set to true every `_generatePointsArray` call and is being reset on each frame? It does feel over-engineered, but I agree with @bghgary that with the suggested implementation we will need to be careful changing the logic in the super function.

EricBeetsOfficial-Opuscope: I don't know if it's relevant, but I am not sure to understand why `skeleton` and `morph target` aren't in the same scope (AbstractMesh). If both were in it, it wouldn't be a problem anymore.

bghgary: @deltakosh?

deltakosh: Most a historical reason. We can move both them no problem

RaananW: any update on this?

EricBeetsOfficial-Opuscope: Sorry, not yet (Iam working on BabylonNative for now :p). Next week, I will investigate on the BoundingBox !

EricBeetsOfficial-Opuscope: Hi all,

Here we go, I moved the `morphTargetManager` into `abstractMesh` class and have uniformized the `_getPositionData` function.
What do you think ?

Eric

src/Meshes/abstractMesh.ts
```diff
@@ -1267,12 +1291,15 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
 
-        if (data && applySkeleton && this.skeleton) {
+        if (data && ((applySkeleton && this.skeleton) || (applyMorph && this.morphTargetManager))) {
             data = Tools.Slice(data);
             this._generatePointsArray();
+        }
+
+        if (data && applySkeleton && this.skeleton) {
 
             var matricesIndicesData = this.getVerticesData(VertexBuffer.MatricesIndicesKind);
```
bghgary, line 1304: nit: spacing

```suggestion
        if (data && applySkeleton && this.skeleton) {
            var matricesIndicesData = this.getVerticesData(VertexBuffer.MatricesIndicesKind);
```

src/Meshes/abstractMesh.ts
```diff
@@ -1320,6 +1347,22 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
                 }
             }
         }
+        if (data && applyMorph && this.morphTargetManager) {
+
+            for (let vertexCount = 0; vertexCount < data.length; vertexCount++) {
+
+                for (let targetCount = 0; targetCount < this.morphTargetManager.numTargets; targetCount++) {
```
bghgary, line 1354: nit: spacing

```suggestion
        }

        if (data && applyMorph && this.morphTargetManager) {
            for (let vertexCount = 0; vertexCount < data.length; vertexCount++) {
                for (let targetCount = 0; targetCount < this.morphTargetManager.numTargets; targetCount++) {
```

bghgary (approved):
LGTM

PR #10272, opened 2021-04-28
Touch holographic button fix

rickfromwork:
Changing the visibility type so that the material is transparent but actually hit-testable. Also renaming the collision mesh for debug readability.

### DIFF
--- a/gui/src/3D/controls/touchHolographicButton.ts
+++ b/gui/src/3D/controls/touchHolographicButton.ts
@@ -270,7 +270,7 @@ export class TouchHolographicButton extends TouchButton3D {
             depth: 1.0,
         }, scene);
         collisionMesh.isPickable = true;
-        collisionMesh.isVisible = false;
+        collisionMesh.visibility = 0;
         collisionMesh.scaling = new Vector3(0.032, 0.032, 0.016);
 
         SceneLoader.ImportMeshAsync(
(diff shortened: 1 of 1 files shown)


gui/src/3D/controls/touchHolographicButton.ts
```diff
@@ -264,13 +264,13 @@ export class TouchHolographicButton extends TouchButton3D {
...
     // Mesh association
     protected _createNode(scene: Scene): TransformNode {
-        const collisionMesh = BoxBuilder.CreateBox(this.name ?? "TouchHolographicButton", {
+        const collisionMesh = BoxBuilder.CreateBox((this.name ?? "TouchHolographicButton") + "_CollisionMesh", {
             width: 1.0,
             height: 1.0,
             depth: 1.0,
         }, scene);
         collisionMesh.isPickable = true;
-        collisionMesh.isVisible = false;
```
rickfromwork, line 273: Kind of misleading that isVisible influences the picking logic, but visibility does not. @RaananW Do you know if there is any underlying reason for having these two similar-sounding values perform completely different tasks?

satyapoojasama (approved):

gui/src/3D/controls/touchHolographicButton.ts
hansagames: it similar in all other tech, that alpha value dont influence picking as it gives possibility to create so called blind buttons that are not visible but can react to inputs, and if element is not visible at all then it dont react to inputs any more

RaananW: In HTML is would be the difference between display hidden and visibility (or even opacity). The element can be there but invisible (with a rate of invisibility), or not visible at all

RaananW (approved):

Issue #10076, opened 2021-03-18
[GE] [Alpha 1B] Create Navigation Hotkeys


msDestiny14:
- [x] Zoom
- [x] Pan

PR #10279, opened 2021-04-29
[GE] Navigation hotkeys 

msDestiny14:
Currently: q = select, w = pan, e = zoom (Subject to change in future date)
https://github.com/BabylonJS/Babylon.js/issues/10076


guiEditor/src/components/commandBarComponent.tsx
```diff
@@ -13,10 +13,33 @@ interface ICommandBarComponentProps {
...
 export class CommandBarComponent extends React.Component<ICommandBarComponentProps> {
-
+    private _panning: boolean;
+    private _zooming: boolean;
+    private _selecting: boolean;
     public constructor(props: ICommandBarComponentProps) {
         super(props);
 
+        props.globalState.onPanObservable.add(() => {
+            this._panning = !this._panning;
```
RaananW, line 23: have you tried passing those to the component state instead of forcing an update? the component's state should have those three elements, and when you set state the component will reload by itself.

guiEditor/src/components/commandButtonComponent.tsx
```diff
@@ -18,8 +18,8 @@ export class CommandButtonComponent extends React.Component<ICommandButtonCompon
 
     public render() {
         return (
-            <div className="command-button" onClick={this.props.onClick} title={this.props.tooltip + (this.props.shortcut ? "\n" + this.props.shortcut : "")}>
-                <div className="command-button-icon">
+            <div className={"command-button " + (this.props.isActive ? "active" : "")} onClick={this.props.onClick} title={this.props.tooltip + (this.props.shortcut ? "\n" + this.props.shortcut : "")}>
```
RaananW, line 21: Template literals are more efficient when creating a "string out of strings".

```suggestion
            <div className={`command-button ${this.props.isActive ? "active" : ""}} onClick={this.props.onClick} title={`${this.props.tooltip} ${this.props.shortcut ? "\n" + this.props.shortcut : ""`}}>

```

I hope I didn't miss anything. This will generate less strings per reload and is, in general, a good practice.

guiEditor/src/diagram/workbench.tsx
```diff
@@ -80,19 +83,50 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
             }
         });
 
-        props.globalState.onPanObservable.add( () => {
-            this._forcePanning = true;
+        props.globalState.onPanObservable.add(() => {
+            this._forcePanning = ! this._forcePanning;
```
RaananW, line 87: same thing with the component state - every time you need a class variable that will be used in the template, think "state".

guiEditor/src/diagram/workbench.tsx
```diff
@@ -80,19 +83,50 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+        props.globalState.onZoomObservable.add(() => {
+
             this._forceZooming = !this._forceZooming;
+            this._forcePanning = false;
+            this._forceSelecting = false;
+            if(!this._forceZooming) {
+                this.globalState.onSelectionObservable.notifyObservers();
+            }
+            else {
+                const canvas = document.getElementById("workbench-canvas") as HTMLCanvasElement;
```
RaananW, line 116: if this never changes (and it doesn't, since it is an element with id), why not have a reference of the element during construction? This way you will save a few `getElementById` calls.

guiEditor/src/diagram/workbench.tsx
```diff
@@ -80,19 +83,50 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+                canvas.style.cursor = "zoom-in";
+            }
         });
 
-        this.props.globalState.hostDocument!.addEventListener("keyup", () => this.onKeyUp(), false);
+        this.props.globalState.hostDocument!.addEventListener(
```
RaananW, line 121: is this being removed at a certain point?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -366,13 +396,20 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
-        const zoomFn = (p: PointerInfo, e: EventState) => {
+        const zoomFnScrollWheel = (p: PointerInfo, e: EventState) => {
             const delta = this.zoomWheel(p, e, camera);
             this.zooming(delta, scene, camera, plane, inertialPanning);
         };
 
+        const zoomFnMouse = (p: PointerInfo, e: EventState) => {
+            const newPos = this.getPosition(scene, camera, plane);
+            const deltaVector = initialPos.subtract(newPos);
+            this.zooming(deltaVector.x > 0? -15: 15, scene, camera, plane, inertialPanning);
```
RaananW, line 407: nit - formatting

guiEditor/src/diagram/workbench.tsx
```diff
@@ -397,7 +434,28 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
             removeObservers();
         }, PointerEventTypes.POINTERUP);
 
-        scene.onPointerObservable.add(zoomFn, PointerEventTypes.POINTERWHEEL);
+
+        scene.onKeyboardObservable.add((k: KeyboardInfo, e: KeyboardEventTypes) => {
```
RaananW, line 438: should that be removed at a certain point? maybe the element unmount/deconstruction?

guiEditor/src/scss/commandBar.scss
```diff
@@ -71,14 +71,14 @@
             background-color: #666666;
         } 
         
-        &:active {
+        .active {
             transform-origin: center;
-            transform: scale(0.95);
+            background-color: #666666;
```
RaananW, line 76: do you have scss variables already? a great candidate :-)

guiEditor/src/diagram/workbench.tsx
```diff
@@ -397,7 +434,28 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
             removeObservers();
         }, PointerEventTypes.POINTERUP);
 
-        scene.onPointerObservable.add(zoomFn, PointerEventTypes.POINTERWHEEL);
+
+        scene.onKeyboardObservable.add((k: KeyboardInfo, e: KeyboardEventTypes) => {
+            switch (k.event.key) {
+                case "q": //select?
+                    if (!this._forceSelecting)
+                        this.globalState.onSelectionObservable.notifyObservers();
```
RaananW, line 442: what is the reason for notifying void? is there a reason to do that on all observables? why not simply not notify?

RaananW: I might be missing some understanding as to what those observables are doing. When I see `onSelectionObservable` , i expect it to notify what was selected, and not only "something was selected".

guiEditor/src/components/commandBarComponent.tsx
RaananW: And every time you really want to use `forceUpdate`, do whatever you can not to (like using setState, as an example) :-)

guiEditor/src/diagram/workbench.tsx
RaananW: Sure, but it will be great to first fully understand what they actually do :-)

msDestiny14:
https://github.com/BabylonJS/Babylon.js/pull/10344

Issue #10278, opened 2021-04-29
Database : IOfflineProvider - .manifest URL creation


gabrielheming:
I have created the thread below on the forum concerning the .manifest URL creation.

https://forum.babylonjs.com/t/database-iofflineprovider-manifest-url-creation/18455

As I said in the post, the situations are pretty forward. However, I got a bit delayed while I was trying to cope with the URL object and IE 11 situation. Because we could use one of the below approaches:

- Find a polyfil package;
- Create its polyfill (it seems to be the case for Promise for instance);
- Use regular expressions instead of URL.

I just didn't know how to proceed.

deltakosh:
Here is a better idea: if URL object do not exist stay with the current behavior

We are slowly moving out of ie11 support so that should be fine

RaananW:
> We are slowly moving out of ie11 support so that should be fine

This kind of sentence will NEVER stop making me extremely happy. 



> if URL object do not exist stay with the current behavior

Perfect, i'll take care of that

gabrielheming:
So, if I do not need to care about IE11, the proposal on the thread is enough. In that case, I can send the PR as is.

PR #10280, opened 2021-04-29
Database : IOfflineProvider - .manifest URL creation

gabrielheming:
Issue #10278

### DIFF
--- a/src/Offline/database.ts
+++ b/src/Offline/database.ts
@@ -107,18 +107,27 @@ export class Database implements IOfflineProvider {
             this._callbackManifestChecked(false);
         };
 
+        var createManifestURL = (): string => {
+            if (typeof URL !== 'undefined') {
+                var url = new URL(this._currentSceneUrl)
+                url.pathname += '.manifest';
+                return url.toString();
+            }
+
+            return this._currentSceneUrl + ".manifest";
+        }
+
         var timeStampUsed = false;
-        var manifestURL = new URL(this._currentSceneUrl)
-        manifestURL.pathname += '.manifest';
+        var manifestURL = createManifestURL();
 
         var xhr = new WebRequest();
 
         if (navigator.onLine) {
             // Adding a timestamp to by-pass browsers' cache
             timeStampUsed = true;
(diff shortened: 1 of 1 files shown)


src/Offline/database.ts
```diff
@@ -108,16 +108,17 @@ export class Database implements IOfflineProvider {
         };
 
         var timeStampUsed = false;
-        var manifestURL = this._currentSceneUrl + ".manifest";
+        var manifestURL = new URL(this._currentSceneUrl)
```
RaananW, line 111: you will still need to check if `typeof URL !== 'undefined'`. Something like this:

```suggestion
        const manifestURL = typeof URL !== 'undefined' ? new URL(this._currentSceneUrl) : this._currentSceneUrl + ".manifest";
```

Of course the rest of the changes will need to react correctly to that as well. So maybe an `if( typeof URL !== 'undefined')` will work better

RaananW (changes requested):
Sadly we still do need to support browsers with no URL element, but it is ok if the previous functionality will still be used on older browsers.

src/Offline/database.ts
gabrielheming: Agreed. I've refactored into an internal function to avoid duplication.

RaananW:
Linting fails. Is timestamp still supported with this method? (looks like it, i am just making sure)

gabrielheming:
It should work properly. It returns a regular string.

The part of timestamp, I've reverted to the original, I am not seeing any lint error. Thoug, I've seen on `Date.now()` that should be `Date.now().toString()`

RaananW:
`Date.now()` returns a number, string + number will result in a string, so it's all good.

gabrielheming:
> `Failed to lint: /home/vsts/work/1/s/src/Offline/database.ts [118, 10]: Missing semicolon, /home/vsts/work/1/s/src/Offline/database.ts [112, 57]: Missing semicolon.`

Ops, my bad. It fixed.

> `Date.now()` returns a number, string + number will result in a string, so it's all good.

Yep, indeed. It was just showing it here before.

RaananW:
thank you so much for this! :-)

PR #10281, opened 2021-04-29
Update WebXRHitTest.ts

RaananW:
Allow a different profile for transient input, defaulting to generic-touchscreen


lehoangphu (approved):
nice!

PR #10286, opened 2021-04-30
Add check for duplicates in ShadowCaster render list

### LINKED ISSUE #9573: ShadowCaster | Avoid duplicates in render list.
In file: https://github.com/BabylonJS/Babylon.js/blob/master/src/Lights/Shadows/shadowGenerator.ts

addShadowCaster & removeShadowCaster:
`shadowGenerator.getShadowMap().renderList` should avoid duplicates.

I don't know if this has affects on rendering performance - but I had a function automatically adding meshes on load
and I wanted to remove one item from the list but the shadow kept displaying.

Reason: 
The meshes were added individually but also via descendants of another mesh.
Would be nice if some could add the `if (not in the list)` statements to add/remove for the mesh itself and the children of it.

This will need a loop or a filter on the array before push:
image

Thanks and best regards

Kevin

Tested: Babylon Version 4.1

ivankoleda:
Closes #9573

It would be really helpful for meshes with large trees of descendants to be filtered for duplicates automatically.
I create map to reduce complexity of these checks, but let me know if this is an overkill in this case.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -27,6 +27,7 @@
 - Added new helper functions for Quaternion.FromLookDirection and Matrix.LookDirection ([Alex-MSFT](https://github.com/Alex-MSFT))
 - Added support for clip planes to the edge renderer ([#10053](https://github.com/BabylonJS/Babylon.js/issues/10053)) ([Popov72](https://github.com/Popov72))
 - Added support for [cannon-es](https://github.com/pmndrs/cannon-es) to the cannonJSPlugin. ([frankieali](https://github.com/frankieali))
+- Added check for duplicates in addShadowCaster ([ivankoleda](https://github.com/ivankoleda))
 
 ### Engine
 
--- a/src/Lights/Shadows/shadowGenerator.ts
+++ b/src/Lights/Shadows/shadowGenerator.ts
@@ -731,10 +731,15 @@ export class ShadowGenerator implements IShadowGenerator {
             (renderedMesh) => { renderListMap[renderedMesh.name] = undefined; },
         );
 
-        var meshes = includeDescendants ? [mesh, ...mesh.getChildMeshes()] : [mesh];
-        for (var addedMesh of meshes) {
-            if (!renderListMap.hasOwnProperty(addedMesh.name)) {
-                this._shadowMap.renderList.push(addedMesh);
+        if (!renderListMap.hasOwnProperty(mesh.name)) {
+            this._shadowMap.renderList.push(mesh);
+        }
+        if (!includeDescendants) {
+            return this;
+        }
+        for (var childMesh of mesh.getChildMeshes()) {
+            if (!renderListMap.hasOwnProperty(childMesh.name)) {
+                this._shadowMap.renderList.push(childMesh);
             }
         }
 
--- a/src/Lights/Shadows/shadowGenerator.ts
+++ b/src/Lights/Shadows/shadowGenerator.ts
@@ -726,20 +726,15 @@ export class ShadowGenerator implements IShadowGenerator {
             this._shadowMap.renderList = [];
         }
 
-        var renderListMap: Record<string, undefined> = {};
-        this._shadowMap.renderList.forEach(
-            (renderedMesh) => { renderListMap[renderedMesh.name] = undefined; },
-        );
-
-        if (!renderListMap.hasOwnProperty(mesh.name)) {
+        if (this._shadowMap.renderList.indexOf(mesh) === -1) {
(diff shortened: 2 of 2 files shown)


dist/preview release/what's new.md
```diff
@@ -1,217 +1,218 @@
-# 5.0.0
```
deltakosh, line 1: can you check why all the file changed?

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -726,10 +726,16 @@ export class ShadowGenerator implements IShadowGenerator {
             this._shadowMap.renderList = [];
         }
 
-        this._shadowMap.renderList.push(mesh);
-
-        if (includeDescendants) {
-            this._shadowMap.renderList.push(...mesh.getChildMeshes());
+        var renderListMap: Record<string, undefined> = {};
+        this._shadowMap.renderList.forEach(
```
deltakosh, line 730: I do not feel this is required (and it will save a full loop) as by default values will be undefined

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -726,10 +726,16 @@ export class ShadowGenerator implements IShadowGenerator {
...
-        this._shadowMap.renderList.push(mesh);
-
-        if (includeDescendants) {
-            this._shadowMap.renderList.push(...mesh.getChildMeshes());
+        var renderListMap: Record<string, undefined> = {};
+        this._shadowMap.renderList.forEach(
+            (renderedMesh) => { renderListMap[renderedMesh.name] = undefined; },
+        );
+
+        var meshes = includeDescendants ? [mesh, ...mesh.getChildMeshes()] : [mesh];
```
deltakosh, line 734: In order to avoid ... (which will again generate a full traversal and a lot of GC) we can simply do something
```
this._shadowMap.renderList.push(mesh);
if (!includeDescendants) {
   return;
}

for (var addedMesh of mesh.getChildMeshes()) {
            if (!renderListMap.hasOwnProperty(addedMesh.name)) {
                this._shadowMap.renderList.push(addedMesh);
            }
        }
```

deltakosh (changes requested):
I would like to avoid all the GC as much as we can :)

src/Lights/Shadows/shadowGenerator.ts
ivankoleda: This code with spread operators may look tricky, I was trying to use minimal amount of memory (so I used `undefined` instead of `true` as value to save few bytes) to create map of existing meshes containing only keys without values and later look if mesh have been previously added to renderList at constant time. I don't see how checking if mesh is already in list can be performed without couple of traversals for initial mesh and it's children with indexOf() if map would not be created. The only possible optimization I see is not to create map if includeDescendants is false. Please correct me if I misunderstood something in your comments.

src/Lights/Shadows/shadowGenerator.ts
ivankoleda: removed spreading of children array and moved check for initial mesh into separate if statement

src/Lights/Shadows/shadowGenerator.ts
ivankoleda: That's what I ment by not creating map for case when `includeDescendants` is false

```   
        if (includeDescendants) {
            var renderListMap: Record<string, undefined> = {};
            this._shadowMap.renderList.forEach(
                (renderedMesh) => { renderListMap[renderedMesh.name] = undefined; },
            );
            if (!renderListMap.hasOwnProperty(mesh.name)) {
                this._shadowMap.renderList.push(mesh);
            }
            for (var childMesh of mesh.getChildMeshes()) {
                if (!renderListMap.hasOwnProperty(childMesh.name)) {
                    this._shadowMap.renderList.push(childMesh);
                }
            }
        } else {
            var index = this._shadowMap.renderList.indexOf(mesh);
            if (index === -1) {
                this._shadowMap.renderList.push(mesh);
            }
        }
```

ivankoleda: However I may have overestimated the difference between accessing `object` property and `array.indexOf` for small/medium sized amounts of data, https://stackoverflow.com/a/48031382/8303988, https://jsben.ch/YEecD V8 seem to be doing a lot of optimizations. I'm ready to rewrite it with multiple `indexOf` calls, that way GC wouldn't have any additional work to do.

deltakosh: That would be neat :)

deltakosh (approved):

PR #10287, opened 2021-04-30
Removed DeviceInputSystem dispose call from InputManager

PolygonalSun:
This PR adds a bit of logic to the InputManager to disable controls on detach and only add to the observables ones when attachControls is called.

I also removed a line from the WebDeviceInputSystem that wasn't doing anything.


deltakosh:
@bghgary is it related to the fix we needed this morning somehow?

bghgary:
It certainly seems like it, but the other PR removes the observable which is probably still necessary.

deltakosh:
ok so waiting for Dave to be back :)

PolygonalSun:
> It certainly seems like it, but the other PR removes the observable which is probably still necessary.

I'm not sure what the specifics are of the issue that you had yesterday but this PR fixes an issue that I found where the dispose was being called in both Engine and InputManager and was breaking gizmos.  This change will fix the behavior while preserving the single-instance per engine of the DeviceInputSystem.

deltakosh:
WAiting for @bghgary / @CedricGuillemet / @RaananW

PolygonalSun:
Closing as issue is fixed

PR #10294, opened 2021-05-02
update package-lock.json version to match package.json

egnor:
I don't know if something is supposed to do this automatically in your spiffy CI setup, but the file keeps popping up in my "git status" after I run "npm install", and... normally I think it's supposed to match...?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "babylonjs",
-    "version": "5.0.0-alpha.16",
+    "version": "5.0.0-alpha.19",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {


deltakosh:
Adding @RaananW  to have a look (I do not know why we have this in the lock in first place)

RaananW:
Someone merged it once, and even since then we update it from time to time :-)


As the project is not defined by the `package.json` in the main directory (out of historical reasons), this file doesn't matter too much. Merging, as it will be good to keep it updated.

RaananW (approved):

deltakosh:
How can we make it updated automatically? Or do we need it at all?

RaananW:
npm officially recommends comitting this file, but we are using package.json a little differently.

There is no issue in removing this file. As part of the overhaul that we want to do to the build system in 5.0, it will probably be addressed as well.

deltakosh:
Sorry I did not mean the file but the babylonjs entry. This looks weird to me

egnor:
(I am excited to hear about a build system overhaul :-))

RaananW:
> Sorry I did not mean the file but the babylonjs entry. This looks weird to me

We can discuss this offline if you want :-) In general, the package-lock is bound to the `package.json` in the same directory, meaning it is bound to this - https://github.com/BabylonJS/Babylon.js/blob/master/package.json, and this one lists the project as 5-alpha19.

deltakosh:
ok but why it does not contains the same version? because we missed to merge it when it changed?

RaananW:
It only updates if you run npm update or npm install. So to keep this updated you need to push to npm (which updates our package.json) and run npm install, and commit both.

deltakosh:
Gotcha!

PR #10295, opened 2021-05-02
Avoid the need for global install during "npm build"

egnor:
No-argument "npm link" updates the global package repository, which is unfortunate and sometimes impossible (by default on Linux, the global repository isn't writable without sudo).

This changes a couple things to avoid that:
- With `--noGlobalInstall`, avoids the "npm link" calls in the prepare{Es6,UMD,Additional}DevPackages build tasks
- Regardless of the flag, directly links to the temp-install for testing, rather than linking from the global repository

(Maybe the "npm link" calls could be removed entirely instead of flag-disabled, but I wasn't sure if other people used the global linkage in their workflow... I'm only testing "npm run build" / "npm run start" at the moment.) The way shelljs is used in these spots is also changed to be more compact and robust.

### DIFF
--- a/Tools/Gulp/tasks/gulpTasks-testsES6.js
+++ b/Tools/Gulp/tasks/gulpTasks-testsES6.js
@@ -19,31 +19,24 @@ var es6TestsWebpackFile = path.resolve(es6TestsFolder, "webpack.config.js");
  * Launches the ES6 modules tests to evaluate the min package size.
  */
 gulp.task("tests-es6Modules", function(done) {
-    colorConsole.log("Npm link dependencies");
-    shelljs.exec("npm link @babylonjs/core", {
-        async: false,
-        cwd: es6TestsFolder
-    });
+    const savedShellConfig = {...shelljs.config};
+    try {
+        Object.assign(shelljs.config, {verbose: true, silent: false, fatal: true});
 
-    shelljs.exec("npm link @babylonjs/materials", {
-        async: false,
-        cwd: es6TestsFolder
-    });
+        colorConsole.log("Link to dev copy of built modules");
+        const coreDevDir = config.core.computed.packageES6DevDirectory;
+        const materialsDevDir = config.materialsLibrary.computed.packageES6DevDirectory;
+        shelljs.exec(`npm install --no-save "${coreDevDir}"`, {cwd: es6TestsFolder});
+        shelljs.exec(`npm install --no-save "${materialsDevDir}"`, {cwd: es6TestsFolder});
 
-    colorConsole.log("Bundle test app");
-    var result = shelljs.exec("npx webpack", {
-        async: false,
-        cwd: es6TestsFolder
-    });
-
-    if (result.code != 0) {
-        colorConsole.error(result.stdout);
-        colorConsole.error(result.stderr);
-        throw "Can not build es6 dev apps."
+        colorConsole.log("Bundle test app with webpack");
+        shelljs.exec("npx webpack", {cwd: es6TestsFolder});
+    }
+    finally {
+        Object.assign(shelljs.config, savedShellConfig);
     }
 
     colorConsole.log("Gather output size");
-
     var testsBaseLine = fs.readJSONSync(baseLinePath);
     var webpackConfig = require(es6TestsWebpackFile);
     for (let entryName in webpackConfig.entry) {
--- a/Tools/Gulp/tasks/gulpTasks-testsES6.js
+++ b/Tools/Gulp/tasks/gulpTasks-testsES6.js
@@ -42,11 +42,13 @@ gulp.task("tests-es6Modules", function(done) {
         }
 
         colorConsole.log("Bundle test app with webpack");
+        shelljs.config.fatal = false;  // Special error reporting below.
         webpack_result = shelljs.exec("npx webpack", {cwd: es6TestsFolder});
-        if (webpack_result.stdout)
+        if (webpack_result.code != 0) {
             colorConsole.error(`webpack stdout:\n${webpack_result.stdout}`);
-        if (webpack_result.stderr)
             colorConsole.error(`webpack stderr:\n${webpack_result.stderr}`);
+            throw "webpack operation failed in tests-es6Modules";
+        }
     }
     finally {
         Object.assign(shelljs.config, savedShellConfig);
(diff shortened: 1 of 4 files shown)


deltakosh:
I need validation from @sebavan but he is oof for a month so please bear with us :)

Tools/Gulp/tasks/gulpTasks-testsES6.js
```diff
@@ -19,31 +19,24 @@ var es6TestsWebpackFile = path.resolve(es6TestsFolder, "webpack.config.js");
...
-        colorConsole.error(result.stderr);
-        throw "Can not build es6 dev apps."
+    const savedShellConfig = {...shelljs.config};
+    try {
+        Object.assign(shelljs.config, {verbose: true, silent: false, fatal: true});
+
+        colorConsole.log("Link to dev copy of built modules");
+        const coreDevDir = config.core.computed.packageES6DevDirectory;
+        const materialsDevDir = config.materialsLibrary.computed.packageES6DevDirectory;
+        shelljs.exec(`npm install --no-save "${coreDevDir}"`, {cwd: es6TestsFolder});
```
sebavan, line 29: would prefer to keep the npm link here and bypassing the task as you do later

Tools/Gulp/tasks/gulpTasks-testsES6.js
```diff
@@ -19,31 +19,24 @@ var es6TestsWebpackFile = path.resolve(es6TestsFolder, "webpack.config.js");
...
+        shelljs.exec(`npm install --no-save "${materialsDevDir}"`, {cwd: es6TestsFolder});
+
+        colorConsole.log("Bundle test app with webpack");
+        shelljs.exec("npx webpack", {cwd: es6TestsFolder});
```
sebavan, line 33: We are now missing our custom error log would be great to have it back

sebavan (changes requested):
I love the overall idea but I guess we are missing some of the error messages now ?

`
 if (result.code != 0) {
        colorConsole.error(result.stdout);
        colorConsole.error(result.stderr);
        throw "Can not build es6 dev apps."
    }
`

Also why not simply preventing the tasks dependencies in case of noGlobalInstall instead of changing all the tasks  and conditionnaly make them do nothing ?

sebavan:
@egnor any updates on the comments ?

egnor:
Ooh sorry the comments look quite sensible I just hadn't circled back to
it. If timing matters I can prioritize it for sure!

sebavan:
no rush and no worries I ll keep it open for now

Tools/Gulp/tasks/gulpTasks-testsES6.js
egnor: Hmm, not sure quite how that would work.

Please pardon me if I'm over-explaining:

Previously, `prepareAdditionalDevPackages`, `prepareEs6DevPackages` and `prepareUMDDevPackages` (the dependency packages) would execute `npm link` calls with no arguments. Those creates symlinks from the global environment to the package directories (`packageDevPath` for each package). Then, `tests-es6Modules` (the user of those packages) would execute `npm link @babylonjs/core` and `npm link @babylonjs/materials`, which would create a symlink from the local `node_modules` to the global environment. So you end up with a chain from local `node_modules` => global environment => `packageDevPath` directory.

With this change, to avoid the global environment in the middle, `tests-es6Modules` (the user) directly links to the `packageDevPath` directory for the packages it depends on. This is done using `npm install` (which when given a directory, makes a symlink). That way it doesn't matter whether or not a global environment link was ever made for the other packages -- the dependency will always be to (what should be?) the right place. That _seems_ generally superior to me but I could be missing some use cases!

If it were up to me, I'd just unconditionally remove the `npm link` from the other packages, but I figured people found some benefit in having them globally linked and I didn't want to rock the boat too much.

So... I _could_ check `noGlobalInstall` here in `tests-es6Modules` (the user), and if set use `npm install ${coreDevDir}`, and if not set use `npm link @babylonjs/core` (and similarly for materials). But given that the `npm install` technique should always work and seems more robust, it seemed simpler and better to just always use `npm install` here at the user. (Also it avoids needing to share the flag value.)

But I'm sure I'm missing something?

egnor (commented):
Sorry for the long delay circling back!

deltakosh:
pinging @sebavan

sebavan:
@egnor About the console, yup would be great to place them back in as I can see those easily from some monitoring tools :-)

About npm link, I am all in to do it if noGlobalInstall is not passed in. Basically, we rely on it heavily when developping locally. like anywhere in my environment I can use my local babylon or mainly one of them from any other projects. Also the dependency is then working automatically without changing the package json to know where the other babylon packages are.

I ll merge as soon as you ll have both changes in :-)

Thanks again

Tools/Gulp/tasks/gulpTasks-testsES6.js
egnor: (Updated to check noInstallGlobal, and choose "npm link" if not set.)

Tools/Gulp/tasks/gulpTasks-testsES6.js
egnor: (updated to put explicit colorConsole logging back.)

egnor (commented):
This should address your final concerns? I'm attempting a build now...

Tools/Gulp/tasks/gulpTasks-testsES6.js
```diff
@@ -19,31 +25,34 @@ var es6TestsWebpackFile = path.resolve(es6TestsFolder, "webpack.config.js");
...
+            shelljs.exec("npm link @babylonjs/materials", {cwd: es6TestsFolder});
+        }
 
-    if (result.code != 0) {
-        colorConsole.error(result.stdout);
-        colorConsole.error(result.stderr);
-        throw "Can not build es6 dev apps."
+        colorConsole.log("Bundle test app with webpack");
+        webpack_result = shelljs.exec("npx webpack", {cwd: es6TestsFolder});
+        if (webpack_result.stdout)
```
sebavan, line 46: Both should only be displayed if result.code != 0

sebavan (changes requested):
One last tiny detail for the console log based on result.errorCode and It ll be merged :-)

Thanks a lot

Tools/Gulp/tasks/gulpTasks-testsES6.js
egnor: Good catch and sorry for the oversight! Updated.

sebavan:
Thank you :-)

PR #10297, opened 2021-05-02
Create WebXR Camera Snippet

BoldBigflank:


RaananW:
Hey, thanks a lot!

I am not sure what happened with git and why it seems like the entire file is changing, will you be able to check it on your side?

And - did you intentionally add it twice?

BoldBigflank:
First I just copy/pasted it all over knowing I only added one entry. Seeing the changes which I thought were an indentation issue, I tried again only adding the new entry. Ignoring whitespace changes it shows the correct changes. I might be using different line endings. I'll tidy this up.

Ah, I didn't realize it was there twice. I'll fix that.

PR #10298, opened 2021-05-03
WebGPU: fix misc bugs

Popov72:
* after a spec change, we must set the **RenderAttachment + CopyDst** flags on all textures because `copyImageBitmapToTexture` now requires it and we don't know in advance if we will call this method or not during the lifetime of a texture. According to @Kangz, there's a memory footprint overhead and maybe some slight runtime overhead to use these flags on all textures. The alternative would be able to flag a texture at creation time as being "read only" so we know we can safely omit those flags.
* fix optimizing usage of `GPUVertexBuffer`. There are some constraints on the value the `offset` property of an attribute can take. If we are out of these limits, we can't regroup this attribute with another one and we must use a separate `GPUVertexBuffer` for this attribute and use the `offset` parameter of the `setBindGroup` call (3rd parameter) to hold the offset value.


Kangz:
> after a spec change, we must set the RenderAttachment + CopyDst flags on all textures because copyImageBitmapToTexture now requires it and we don't know in advance if we will call this method or not during the lifetime of a texture.

Why not just do an extra copy instead? It's going to be really rare that you need to have textures updated each frame.

Popov72:
> Why not just do an extra copy instead? It's going to be really rare that you need to have textures updated each frame.

It happens when using the GUI / video textures. Maybe when we have other options to deal with canvas/videos than using `copyImageBitmapToTexture` we will be able to make this extra copy instead of setting the flags?

PR #10299, opened 2021-05-03
Fix sprite not displayed in certain cases

Popov72:
See https://forum.babylonjs.com/t/newbie-question-about-sprites-not-showing-within-an-inside-out-mesh/20483


Faber-smythe (approved):

sebavan:
Could this be added under a flag as it looks like it would double the number of draw calls in some cases ?

Popov72:
Why would that double the number of draw calls? It just (re)sets the front face culling to CCW before drawing the sprite(?)

sebavan:
> Why would that double the number of draw calls? It just (re)sets the front face culling to CCW before drawing the sprite(?) (note that `culling = true`).

Yup stupid me :-) those vacations are much needed...

PR #10301, opened 2021-05-03
[Navigation] Limit new objects for garbage collection

CedricGuillemet:
follow up on this thread : https://forum.babylonjs.com/t/apparent-memory-leak-in-recastjs/20450

Too many vec3 allocations. using temp vectors allocated once and reused.

### DIFF
--- a/src/Navigation/Plugins/recastJSPlugin.ts
+++ b/src/Navigation/Plugins/recastJSPlugin.ts
@@ -33,6 +33,9 @@ export class RecastJSPlugin implements INavigationEnginePlugin {
     private _maximumSubStepCount: number = 10;
     private _timeStep: number = 1 / 60;
 
+    private _tempVec1: any;
+    private _tempVec2: any;
+
     /**
      * Initializes the recastJS plugin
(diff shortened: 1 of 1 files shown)


src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -261,9 +275,13 @@ export class RecastJSPlugin implements INavigationEnginePlugin {
      * @returns the resulting point along the navmesh
      */
     moveAlong(position: Vector3, destination: Vector3): Vector3 {
-        var p = new this.bjsRECAST.Vec3(position.x, position.y, position.z);
-        var d = new this.bjsRECAST.Vec3(destination.x, destination.y, destination.z);
-        var ret = this.navMesh.moveAlong(p, d);
+        this._tempVec1.x = position.x;
```
Popov72, line 278: Not related to the PR, but I think `moveAlong` should simply rely on `moveAlongToRef` to avoid duplicate code.

Popov72 (approved):
LGTM!

RaananW (approved):

PR #10306, opened 2021-05-03
Playground - Create WebXR Camera snippet

BoldBigflank:
Cleaned up and tested locally.


BoldBigflank:
There were also a couple other snippets that didn't have uniform indentation, fixed that too.

RaananW (approved):

PR #10307, opened 2021-05-03
WebGPU: Add support for compute shaders

Popov72:
I thought I should make a first PR even if the support is not complete yet.

What's missing:
* ~~support for storage buffers as input types. Currently, there is support for textures, storage textures and ubos~~ Added
* full serialization / parsing. This one is a bit involved because:
  * empty created raw textures are currently not serialized => need to add proper serialization to them
  * ubos are not serializable
  * when storage buffers will be in, serialization will have to be added to them too

Here are two samples:
* https://playground.babylonjs.com#3URR7V#59. Use a CS to clear the texture applied to the sphere and another CS to copy a texture into another one, applied on the ground. Note that it's for demonstration only, I set the workgroup size to be (1,1,1) which is really bad for performances! Also creates a CS to compute the multiplication of two matrices, to demonstrate usage of storage buffers
* https://playground.babylonjs.com#3URR7V#46. It is a direct port of the WebGPU sample [imageBlur](http://austin-eng.com/webgpu-samples/samples/imageBlur)

Regarding the implementation:
* The implementation is modeled after `Effect` / `PipelineContext`, so there's a `ComputeEffect` and a `ComputePipelineContext` classes (but much simpler than their rendering counter part!). The main user land class is `ComputeShader` which is modeled a bit like `ShaderMaterial`.
* To support storage textures, I needed to know that a texture should be created as a storage texture or not, so I have added a `creationFlag` parameter which is a bit field. Only bit 0 is currently used and when set means "storage texture" (there's a constant for it in `Constants.ts`)
  * I have added two new methods `CreateRGBAStorageTexture` and `CreateRStorageTexture` to ease creating storage textures that automatically set the bit when creating the corresponding raw textures
* Note that WGSL does not support preprocessing, so the `#define` that we normally generate in the source shader are prefixed by `//`.
* The browsers don't currently support reflection from WGSL code, so it's not possible to retrieve the group + binding values of a given variable by name: the user must directly provide the group and binding numbers.  The interface (see `setTexture`, `setStorageTexture`, `setUniformBuffer`) is designed to also be able to take a string instead of two numbers for when the browsers will support reflection.


deltakosh:
All good for me!! I feel like we should write the page doc asap :)

deltakosh (approved):

Popov72:
Yes, I plan to do it after I have added the support for the storage buffers.

deltakosh:
Lovely!! Let's merge and prepare a tweet ;)

deltakosh:
oh just a lint issue and you can merge :)

benaadams:
Can you add this to breaking changes (is a break if using imports from `@babylonjs/core`
```diff
-import { DataBuffer } from "../Meshes/dataBuffer";
-import { Buffer, VertexBuffer } from "../Meshes/buffer";
+import { DataBuffer } from "../Buffers/dataBuffer";
+import { Buffer, VertexBuffer } from "../Buffers/buffer";
```

deltakosh:
Excellent catch!

@Popov72

PR #10309, opened 2021-05-03
Clear engine, scene, and initFunction at the start of a playground run

egnor:
Per https://forum.babylonjs.com/t/global-variable-scene-persistent-between-playground-runs-can-be-confusing/20311/2, clears some window globals before launching a new playground run.

Also fixes a buglet where exceptions thrown during execution weren't getting shown properly.

(Be happy: I got sucked into a much larger revamp of that code, but then had a moment of sanity and backed out to make the much more strategic change...)

### DIFF
--- a/Playground/src/components/rendererComponent.tsx
+++ b/Playground/src/components/rendererComponent.tsx
@@ -121,7 +121,13 @@ export class RenderingComponent extends React.Component<IRenderingComponentProps
         }
 
         try {
+            // Set up the global object ("window" and "this" for user code).
+            // Delete (or rewrite) previous-run globals to avoid confusion.
             let globalObject = window as any;
+            delete globalObject.engine;
+            delete globalObject.scene;
+            delete globalObject.initFunction;
+
             let canvas = this._canvasRef.current!;
             globalObject.canvas = canvas;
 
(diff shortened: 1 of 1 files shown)


egnor:
(By the way, let me know if this needs a note on the what's-new page. I'm assuming it's a little too trivial for that, much like my various fixes to build nits.)

deltakosh (approved):

RaananW:
This code will be exported when a user downloads the playground snippet as well. The reason it was done the way it was is to prevent the user downloading it from using the global object and to keep scene and engine scoped, at least when the snippet is used locally outside of the playground.

In case this is fine, I would still use `globalObject` instead of window where possible (or remove `globalObject` altogether).

egnor:
Sorry, I don't understand how it kept scene and engine scoped before? It
put "var scene;" and "var engine;" right at the top level, which installs
them into the global namespace (aka window aka self).

I would love to have scene and engine scoped so you can't access them from
the global object but I bet that would break scripts that depend on having
them there.

That's actually part of the problem I was trying to improve, which was that
in the interactive playground, the old scene is still in the global object
until the end of create scene(), which can lead to confusion.

Sorry, I probably misunderstand you -- can you clarify??

RaananW:
Hmmm...

You are right. We used to scope it differently. I see the downloadable code is a little different now.
We'll need to address this in the future.

Thanks!

PR #10312, opened 2021-05-04
defensive URL parsing

RaananW:
Since we don't have a base URL (which is required as the 2nd variable in the URL constructor when using relative URLs), relative URLs will not parse correctly. 
This will fallback to returning a normal string instead.
See - https://forum.babylonjs.com/t/idbstorageenabled-throws-invalid-url-error/20475


deltakosh:
Reviewing your PR are always a pain in the heck :( 
Can you try to only change what was really part of the PR?

I trust you on this one as there are too many changes

PR #10329, opened 2021-05-06
SSAO2: Allow to choose the texture type used by the post processes

Popov72:
See https://forum.babylonjs.com/t/ssao-with-hrd-texture-causes-color-precision-loss/20598/4


RaananW (approved):

RaananW:
Can the failed visual tests be related to that change? Was that the default all along?

Popov72:
Yes, the value I set as the default one was the default one before.

I checked on my local computer and all validation tests pass successfully.

RaananW (approved):

CraigFeldspar (approved):
LGTM

PR #10330, opened 2021-05-06
Intersects (Pick) - make sure we iterate correctly when step=1

RaananW:
When Material fillMode is set to Constants.MATERIAL_TriangleStripDrawMode we iterate one time to many. We should stop when index is smaller than indexCount - 2 (is step is 1).
See - https://playground.babylonjs.com/debug.html#WIR77Z#344
Also added a line to make sure positions exist before checking the intersection to avoid exceptions thrown and to fail silently.


Popov72 (approved):
LGTM

deltakosh (approved):

PR #10333, opened 2021-05-06
Buffers: allow more generic buffer creation and wrapping of existing GPU buffers

Popov72:
The PR allows to use an existing GPU buffer (`DataBuffer`) in a `Buffer` object so that we can create a `VertexBuffer` around a storage buffer filled by a compute shader (for eg).

Also, storage buffer creation now takes a creation flag to flag the buffer for specific additional usages (as uniform buffer, vertex buffer, index buffer, etc).

Finally I have added a `Mesh.forcedInstanceCount` property as discussed offline with @deltakosh.

### DIFF
--- a/src/Buffers/buffer.ts
+++ b/src/Buffers/buffer.ts
@@ -32,7 +32,7 @@ export class Buffer {
      * @param useBytes set to true if the stride in in bytes (optional)
      * @param divisor sets an optional divisor for instances (1 by default)
      */
-    constructor(engine: any, data: DataArray, updatable: boolean, stride = 0, postponeInternalCreation = false, instanced = false, useBytes = false, divisor?: number) {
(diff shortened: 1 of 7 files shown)


src/Buffers/buffer.ts
```diff
@@ -44,7 +44,13 @@ export class Buffer {
         this._instanced = instanced;
         this._divisor = divisor || 1;
 
-        this._data = data;
+        if (data instanceof DataBuffer) {
```
deltakosh, line 47: can we replace instanceOf by anthing else? isntanceof is expensive

Popov72: Do we really create so many buffers that it could be impactful?

Note that the `VertexBuffer` constructor is doing an `if (data instanceof Buffer) {`, that's why I thought I was allowed to do it in the `Buffer` constructor too :).

But I can change that in both constructors if you want.

deltakosh (approved):

src/Buffers/buffer.ts
deltakosh: yeah no you are right it is just my TOC :) keep it like that no worries

PR #10344, opened 2021-05-07
[GE] Navigation hotkeys 

msDestiny14:
Currently: q = select, w = pan, e = zoom (Subject to change in future date)
#10076

Changes from original PR https://github.com/BabylonJS/Babylon.js/pull/10279


guiEditor/src/diagram/workbench.tsx
```diff
@@ -397,7 +440,28 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
             removeObservers();
         }, PointerEventTypes.POINTERUP);
 
-        scene.onPointerObservable.add(zoomFn, PointerEventTypes.POINTERWHEEL);
+
+        scene.onKeyboardObservable.add((k: KeyboardInfo, e: KeyboardEventTypes) => {
+            switch (k.event.key) {
+                case "q": //select?
```
RaananW, line 446: When defined above, you are using a capital "Q" , but here is is lowercase. Is that the same shortcut?

guiEditor/src/scss/commandBar.scss
```diff
@@ -68,17 +71,17 @@
...
         &:hover {
-            background-color: #666666;
+            background-color: var(--selectionGrey);
         } 
         
-        &:active {
+        .active {
             transform-origin: center;
-            transform: scale(0.95);
+            background-color: #666666;
```
RaananW, line 79: awesome! you have a variable :-) you can use it here

RaananW (approved):

PR #10351, opened 2021-05-11
Fix JsDoc of Buffer.create

barroij:
minor jsodc update for function `Buffer.create` that will do nothing if the buffer is already used but not updatable.

This change is not worth a line in 'What's new' :)


barroij:
probably for you @Popov72

RaananW (approved):

PR #10363, opened 2021-05-13
Fixing utility scene XR pick

### LINKED ISSUE #10179: Most 3D Controls aren’t pickable by WebXR ControllerPointerSelection
Test scene here: https://www.babylonjs-playground.com/#9K3MRA#384
Reference forum post: https://forum.babylonjs.com/t/most-3d-controls-arent-pickable-by-webxr-controllerpointerselection/19845

After entering VR, you can see that the ray cast by the controller passes right through the holographic button. This is because the button’s meshes are created on the scene’s utility layer, which WebXRControllerPointerSelection doesn’t check during its ray-casting phase. This in turn means that instead of the ray stopping on the surface and rendering a cursor, it passes right through without rendering a cursor.

Note: You can still interact with the button using the ray + motion controller trigger button, you just need to guess as to if the ray is hitting the button.

I feel either controllers shouldn’t interact with 3D controls (though this doesn’t make much sense to me), or the utility layer meshes should be included with the main scene’s meshes for the pick (this seems to already be the case for 2D interactions?).

Note 2: The meshes created in the scene when preparing a MeshButton3D are the only 3D control that WebXRControllerPointerSelection can pick, as the mesh is created on the main scene. This creates a visual inconsistency between the behaviour of different 3D controls.

CraigFeldspar:
This is a proposal to fix #10179

Introduces a 1 to 1 reference between a scene and an utility layer. Is this assumption valid though ? Do you see a case where we can have several utility layers ?


src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -366,7 +370,29 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
...
+            let utilityScenePick = null;
+            if (this._scene._utilityLayerRenderer) {
+                utilityScenePick = this._scene._utilityLayerRenderer.utilityLayerScene.pickWithRay(
+                    controllerData.tmpRay,
+                    this._scene._utilityLayerRenderer.utilityLayerScene.pointerMovePredicate || this.raySelectionPredicate
+                );
+            }
+
+            let originalScenePick = this._scene.pickWithRay(controllerData.tmpRay, this._scene.pointerMovePredicate || this.raySelectionPredicate);
+            if (!utilityScenePick || utilityScenePick.distance === 0) {
```
RaananW, line 383: Why not check for pick.hit ?

src/Rendering/utilityLayerRenderer.ts
```diff
@@ -10,6 +10,16 @@ import { Vector3 } from '../Maths/math.vector';
 import { Camera } from '../Cameras/camera';
 import { Color3 } from '../Maths/math.color';
 
+declare module "../scene" {
+    export interface Scene {
+        /**
+         * Associated utility layer renderer
+         * @hidden
+         */
+        _utilityLayerRenderer: UtilityLayerRenderer;
```
RaananW, line 19: Just a thought - Wouldn't it be possible to pass the utility layer renderer to the XR feature? In the options or as a setter?

I noticed the question regarding a scene-utility layer renderer relation. Though not expected, it is still possible to have more than one.

Will also avoid adding more parameters to the scene, which is very bloated as it is :-)

RaananW (commented):
Do I understand the logic correctly?:

1) If both had pick results, choose the closer one
2) If one of them has a pick (and the other doesn't), choose this one
3) If no pick found, choose the utility layer's  pick

src/XR/features/WebXRControllerPointerSelection.ts
CraigFeldspar: I took code from the utility layer pick listener. But you're right pick.hit is clearer

src/Rendering/utilityLayerRenderer.ts
CraigFeldspar: I looked and there is indeed a property `useUtilityLayer` in `WebXRControllerPointerSelection`, and also a static property `UtilityLayerRenderer.DefaultUtilityLayer.utilityLayerScene``, so we can probably use that

CraigFeldspar:
> Do I understand the logic correctly?:
> 
> 1. If both had pick results, choose the closer one
> 2. If one of them has a pick (and the other doesn't), choose this one
> 3. If no pick found, choose the utility layer's  pick

Right, except 
3. If no pick is found, original scene's pick is chosen

RaananW (approved):

Issue #10353, opened 2021-05-11
Texture Inspector middle mouse to pan does not stop when button released


PatrickRyanMS:
**Repro**
- open any playground or model in the sandbox and select a texture
- open the texture inspector and use mouse wheel to zoom in a bit
- press and hold middle mouse button to pan the image, then release button and move mouse
- Expected result: the image stops panning
- Current result: the image stays in the panning state and no other actions can be performed

PR #10364, opened 2021-05-13
DeviceInputSystem: Convert web implementation to use set/releasePointerCapture

PolygonalSun:
This PR changes the previous implementation of using the window as the event listening element back to the scene's provided HTMLElement.  Furthermore, setPointerCapture and releasePointerCapture are now being used to prevent lost of pointer button tracking.

These changes should also fix the following issue: https://github.com/BabylonJS/Babylon.js/issues/10353

### DIFF
--- a/src/DeviceInput/Implementations/webDeviceInputSystem.ts
+++ b/src/DeviceInput/Implementations/webDeviceInputSystem.ts
@@ -41,6 +41,8 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
     private _pointerBlurEvent = (evt: any) => { };
     private _wheelEventName: string;
 
+    private _mouseId = -1;
+
     private _pointerWheelClearObserver: Nullable<Observer<Engine>> = null;
 
     private _gamepadConnectedEvent = (evt: any) => { };
@@ -138,7 +140,6 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
             // Blur Events
             this._elementToAttachTo.removeEventListener("blur", this._keyboardBlurEvent);
             this._elementToAttachTo.removeEventListener("blur", this._pointerBlurEvent);
-            this._elementToAttachTo.removeEventListener(this._eventPrefix + "out", this._pointerBlurEvent);
 
             // Keyboard Events
             if (this._keyboardActive) {
@@ -150,7 +151,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
             if (this._pointerActive) {
                 this._elementToAttachTo.removeEventListener(this._eventPrefix + "move", this._pointerMoveEvent);
                 this._elementToAttachTo.removeEventListener(this._eventPrefix + "down", this._pointerDownEvent);
-                window.removeEventListener(this._eventPrefix + "up", this._pointerUpEvent);
+                this._elementToAttachTo.removeEventListener(this._eventPrefix + "up", this._pointerUpEvent);
                 this._elementToAttachTo.removeEventListener(this._wheelEventName, this._pointerWheelEvent);
 
                 if (this._pointerWheelClearObserver) {
@@ -339,6 +340,10 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 this._addPointerDevice(deviceType, deviceSlot, evt.clientX, evt.clientY);
             }
 
+            if (deviceType === DeviceType.Mouse && this._mouseId === -1) {
+                this._mouseId = evt.pointerId;
+            }
+
             const pointer = this._inputs[deviceType][deviceSlot];
             if (pointer) {
                 // Store previous values for event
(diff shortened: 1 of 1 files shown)


src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -415,12 +420,18 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
...
+                this._mouseId = evt.pointerId;
+            }
+
             const pointer = this._inputs[deviceType][deviceSlot];
             if (pointer) {
                 const previousHorizontal = pointer[PointerInput.Horizontal];
                 const previousVertical = pointer[PointerInput.Vertical];
                 const previousButton = pointer[evt.button + 2];
 
+                this._elementToAttachTo.setPointerCapture(evt.pointerId);
```
deltakosh, line 433: Was it the case before? if no what is the rationale behind this choice?

darraghjburke: I believe this is because we want to capture pointerup events even if the pointer leaves the canvas before the user releases the button. Since the element captures the pointer, it will receive the pointerup event no matter where the pointer is on screen. https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture

This is relevant to #10353  because the user may move their mouse outside the texture inspector window before releasing the pan button. The texture inspector need to catch this pointerup event and stop panning.

@PolygonalSun do I understand correctly?

src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -41,6 +41,8 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
     private _pointerBlurEvent = (evt: any) => { };
     private _wheelEventName: string;
 
+    private _mouseId = -1;
```
darraghjburke, line 44: I think we may want to track a list of buttons we have captured, rather than assuming that only one pointer is pressed. If the user presses multiple mouse buttons down, we should capture them all, and then release them all when the related pointerup event occurs.

bghgary: Capturing the pointer should capture all pointer events to the element. I don't think we need to worry about which buttons. They will all be captured. If there are multiple mouse devices connected, then we may have an issue.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
PolygonalSun: Yup, that's correct.  Initially, we were just listening to the window for pointerup events and handling it that way but that method doesn't work with situations that have multiple windows like the Texture Inspector.  This approach gives us identical behavior with single windows but also fixes pointerup listening for the pop out window.

deltakosh: Ok make sure it works when inside NME preview window, Playground, etc...

bghgary: There is an edge case where this doesn't work correctly. If you hold down two different buttons and let go of only one of them, the pointer capture will be released. Seems like we need to ref count for this to be totally correct.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
PolygonalSun: For pointer button events, a "pointerup" event is only thrown all buttons are up.  Otherwise, a "pointermove" event is thrown when a button is released.

Things do get a little tricky when the cursor leaves the canvas with multiple buttons pressed though.  There is a standing bug (since 4.0) where dragging will "stick" if the cursor leaves the canvas with multiple buttons pressed.  It may be worth fleshing out the logic to try and fix this, though I might relegate that to a different PR so that Texture Inspector gets fixed first.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
PolygonalSun: I wasn't able to repro this edge case result.  When I tried tracking this what I got was
FIRST BUTTON DOWN (pointerdown fired, capture set)
SECOND BUTTON DOWN (pointermove fired, capture still set)
ANY BUTTON UP (pointermove fired, capture still set)
OTHER BUTTON UP (pointerup fired, capture released)

The pointercapturelost event only fired after released in the pointerup (after the second button was released).  I've tried this with both the PG and Texture Inspector.  Is there a step that I might be missing to repro this?

bghgary: Sorry, I didn't actually try this code change. I mentioned this because I tried the [MDN sample](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) and it didn't work right for this. Maybe this code works and I don't understand it well enough to see it.

darraghjburke (approved):
Looks good to me! Thanks for addressing the comments.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
PolygonalSun: Just as an update to what this changes:
- NME: No change
- GUI Editor: No change
- Sandbox: PointerUp working in Texture Inspector window; Objects can be dragged off screen (drag coords can be read off window)
- Playground: Objects can be dragged off screen (drag coords can be read off window); works with Gizmos
While this is a slight change to behavior, I don't see this as a negative change.  The original behavior was that if the cursor left the screen while a button was pressed, the dragging behavior would stop but the object being dragged would "snap" to the coordinates of the cursor when it returned, assuming the button was never released.

deltakosh: This is pretty neat actually :D

deltakosh (approved):
Good job buddy! waiting for final approval by @bghgary and you can merge

bghgary (approved):
Good for me

PR #10379, opened 2021-05-17
LinesMesh and PCS: allows to use an existing material

Popov72:
See https://forum.babylonjs.com/t/stop-materials-from-unnecessary-duplicating/20860


Popov72:
Yes, it is the "Texture cache" test which is failing, as it happens often.

PR #10391, opened 2021-05-19
WebGPU: add support for GPU particles

Popov72:
I had to reorganize the structure of the particle buffers to limit the number of dummy properties (the more dummy properties we have, the bigger the buffer is) in the WebGPU case: the STD140 layout applies for the storage buffers.

I did not manage to have shader files with a **.wgsl** extension to be automatically transformed to a .ts file by the gulp task, so I have created a .fx file instead for the update compute shader (**gpuUpdateParticles.compute.fx**).

In addition to the WebGPU support, I have also performed those changes in the `GPUParticleSystem` class:
* support for the `MULTIPLYADD` blend mode
* support for custom effects
* fixed a bug with the cylinder emitter

On [this playground](https://playground.babylonjs.com/#PU4WYI#241) (it won't work without the PR, even in WebGL2) and on my computer (GTX 1080), according to PIX the compute task is 1.35 times faster than the Stream Output stage used in WebGL2 (transform feedback buffers): 1.41ms vs 1.91ms.

Test PG: https://playground.babylonjs.com/#PU4WYI#4


src/Particles/gpuParticleSystem.ts
```diff
@@ -20,13 +20,21 @@ import { EngineStore } from "../Engines/engineStore";
 import { IAnimatable } from "../Animations/animatable.interface";
 import { CustomParticleEmitter } from "./EmitterTypes/customParticleEmitter";
 import { ThinEngine } from "../Engines/thinEngine";
+import { ComputeShader } from "../Compute/computeShader";
```
deltakosh, line 23: I wonder if we should not inherit from GPUParticle to create a ComputeShaderParticle ? This way we can keep the GPUPart slimmer

THoughts?

deltakosh (approved):
Love it.. Just a change in the architecture but that should not be too painful

src/Particles/gpuParticleSystem.ts
Popov72: I can't see a GPUParticle class? Also, the compute shader is not linked to a particle (there's not a shader per particle) but to the particle system as a whole.

deltakosh: I meant GPUParticleSystem AND ComputeShaderParticleSystem

Popov72: I think that would be more of a `WebGL2ParticleSystem` and `ComputeShaderParticleSystem` that would implement the specificities of each platform and `GPUParticleSystem` that would use one of the other depending on the current engine?

Popov72:
> Just a change in the architecture but that should not be too painful

Actually that was not so straightforward ;)

deltakosh:
Sorry buddy

dist/preview release/what's new.md
```diff
@@ -220,6 +220,7 @@
 - Fix for namepsace sharing in .scss files; PropertyTab, SceneExplorer ([msDestiny14](https://github.com/msDestiny14))
 - Fix sprites not displayed in certain cases ([Popov72](https://github.com/Popov72))
 - Fix undefined camera pose in WebXR in Babylon Native ([CraigFeldspar](https://github.com/CraigFeldspar))
+- Fix some different behaviours between `ParticleSystem` and `GPUParticleSystem` when using the cylinder emitter ([Popov72](https://github.com/Popov72))
```
deltakosh, line 223: the what's new should be updated with the new class

src/Particles/gpuParticleSystem.ts
```diff
@@ -733,11 +765,18 @@ export class GPUParticleSystem extends BaseParticleSystem implements IDisposable
             this.defaultProjectionMatrix = Matrix.PerspectiveFovLH(0.8, 1, 0.1, 100);
         }
 
-        this._customEffect = { 0: customEffect };
+        this._platform = this._engine.getCaps().supportComputeShaders ? new ComputeShaderParticleSystem(this, this._engine) : new WebGL2ParticleSystem(this, this._engine);
```
deltakosh, line 768: I would love that this code to be delegated nsomehow to the children

My ultimate goal is for someone referencing the GPU particle system for webgl2 to NOT have to import webgpu one

Popov72: I tried something, let me know if it's ok for you.

deltakosh (approved):
LGTM!

PR #10402, opened 2021-05-20
WebGPU: add support for occlusion queries

Popov72:
Note that in WebGPU there aren't different algorithms for occlusion queries, so `OCCLUSION_ALGORITHM_TYPE_CONSERVATIVE` and `OCCLUSION_ALGORITHM_TYPE_ACCURATE` are not used.


deltakosh:
So our occlusion query works in webgpu? MAN YOU ROCK

Popov72:
Is it possible to restart the WebGL1 tests? There's no reason why it fails the shadow wrapper one, the PR has no relation to it (it seems one node material could not be downloaded currently from the snippet server).

RaananW:
> Is it possible to restart the WebGL1 tests? There's no reason why it fails the shadow wrapper one, the PR has no relation to it (it seems one node material could not be downloaded currently from the snippet server).

Rerunning them now.

PR #10403, opened 2021-05-21
StandardMaterial: fix not using the tangent attribute when available

Popov72:


Popov72:
The validation test failure can be ignored, it's "Texture cache" again.

deltakosh:
Man, this is so annoying. Can you try to look at why this one is always failing? maybe we can replace it?

Popov72:
I have no idea why it fails, it's so simple... Only 3 planes with 3 textures displayed. When we look at the image data we can see the output is ok, but we got a log with an gl error 1282 (invalid operation)...

I'm going to move this test somewhere else in the config.json file in my next PR and we will see if it's still failing.

PR #10405, opened 2021-05-21
DeviceInputSystem: Add checks for pointer events with no pointerId

PolygonalSun:
This PR contains code to check for and address instances where pointer functions are being triggered by MouseEvents and TouchEvents.  This includes auto-assignment of an id for mouse input and exception throwing for touch input.


src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -420,17 +429,30 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
...
-
             const pointer = this._inputs[deviceType][deviceSlot];
             if (pointer) {
                 const previousHorizontal = pointer[PointerInput.Horizontal];
                 const previousVertical = pointer[PointerInput.Vertical];
                 const previousButton = pointer[evt.button + 2];
 
-                this._elementToAttachTo.setPointerCapture(evt.pointerId);
+                if (deviceType === DeviceType.Mouse && this._mouseId === -1) { // Mouse; Among supported browsers, value is either 1 or 0 for mouse
+                    if (evt.pointerId === undefined) { // If there is no pointerId (eg. manually dispatched MouseEvent)
```
deltakosh, line 439: why not fixing it at the SimulatePointerDOwn level? This would avoid having to generate id that could collide with real ones

PolygonalSun: I've thought about this but I decided to go this route because both the InputManager's and DeviceSourceManager's input routes through the DeviceInputSystem.  If I were to fix it at the simulatePointerDown level, it would still be broken if anyone were to use MouseEvents with the DeviceSourceManager.

With the interpreting of the pointerId, we're only doing that for the mouse and in my various tests, the id for mouse is 1 with the exception of Firefox.  According to the spec, the default value for pointerId is 0 so if the event doesn't have a properly defined pointerId (either _undefined_ or the default value for a non-FF browser), we'd have to adjust it or an exception will be thrown.  If the PointerEvent is properly created, this interpretation should never happen.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -224,6 +225,14 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
      * @param numberOfInputs Number of input entries to create for given device
      */
     private _registerDevice(deviceType: DeviceType, deviceSlot: number, numberOfInputs: number) {
+        if (deviceType === undefined) {
+            throw "Unable to register undefined device";
+        }
+
+        if (deviceSlot === undefined) {
+            throw `Unable to register device ${DeviceType[deviceType]} to undefined slot.`;
+        }
```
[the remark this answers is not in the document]
PolygonalSun: So I just ran through some tests and deviceType being undefined actually isn't possible BUT you can have an undefined deviceSlot if someone creates a pointerdown event with TouchEvent.  TBH, we can't actually interpret a manually dispatched TouchEvent because it has no pointerId that we could properly extrapolate.  I could remove the deviceType check but I'd like to keep the slot one because if it's undefined, it will use the values for the touch as one level up:
Inputs[DeviceType.Touch][DeviceSlot = undefined] is interpreted as Inputs[DeviceType.Touch]

src/DeviceInput/Implementations/webDeviceInputSystem.ts
deltakosh: ok I'm just not at ease with sniffing FF. BUt not sure how to avoid it

deltakosh (approved):
LGTM

src/DeviceInput/Implementations/webDeviceInputSystem.ts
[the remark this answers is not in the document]
bghgary: Too late for this PR, but in general If it can actually be undefined, the function signature should probably represent that so that TypeScript can warn you about it.

PR #10406, opened 2021-05-21
Add support for sRGB textures

### LINKED ISSUE #10360: Support EXT_sRGB for proper texture color interpolation
See https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureLinearInterpolationTest for a test model that exhibits the incorrect behavior.

Live repro: https://sandbox.babylonjs.com/?assetUrl=https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/TextureLinearInterpolationTest/glTF/TextureLinearInterpolationTest.gltf

image

See https://github.com/KhronosGroup/glTF-Sample-Models/pull/297 for context.

See [EXT_sRGB](https://developer.mozilla.org/en-US/docs/Web/API/EXT_sRGB) for the WebGL1 extension. WebGL2 and WebGPU has built-in support via flags.

Popov72:
The glTF loader is now using sRGB textures by default (if supported by the GPU) for gamma encoded textures, but there's a flag to disable it (`glTFFileLoader.useSRGBBuffers = false`).

I have added the possibility to pass an object with the options to the `Texture` constructor as the number of parameters was becoming frightening...

The sandbox can now also use WebGPU instead of WebGL if you pass `webgpu` in the url.

I had to regenerate a number of pictures for some of the validation tests related to glTF as this PR can change slightly the output from what it was before: the tests generally didn't fail but we still could see a greater number of red pixels in the comparison pictures than before the PR.

I have also moved the "Texture cache" test at the top of the list, as this test keeps failing when executed on the Azure servers...

Fix  #10360


bghgary:
May want to put `fixes #10360` in the comment to auto close the issue.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -2014,30 +2014,40 @@ export class GLTFLoader implements IGLTFLoader {
...
+        const promise = this._createTextureAsync(context, sampler, image, assign, undefined, !texture._textureInfo.nonColorData);
 
         this.logClose();
 
         return promise;
     }
 
     /** @hidden */
-    public _createTextureAsync(context: string, sampler: ISampler, image: IImage, assign: (babylonTexture: BaseTexture) => void = () => { }, textureLoaderOptions?: any): Promise<BaseTexture> {
+    public _createTextureAsync(context: string, sampler: ISampler, image: IImage, assign: (babylonTexture: BaseTexture) => void = () => { }, textureLoaderOptions?: any, useSRGBBuffer?: boolean): Promise<BaseTexture> {
```
bghgary, line 2025: I'm a bit worried about adding more and more parameters to this function, which may potentially be exposed to the loader extensions someday. No action to do for now. I just want to point this out.

bghgary (approved):
LGTM from the glTF side

deltakosh (approved):

deltakosh:
weblg1 test is failing

Popov72:
WebGL1/2 test failing is a timeout ("Timeout of 180000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves."), so probably a transient network error.

The PG is https://playground.babylonjs.com/#EYAUPC#2, I don't see why it would fail.

Popov72:
Also "GUI Near Menu" failing here.

PR #10408, opened 2021-05-24
# Modify the branch condition test in GLTFLoader._loadVertexAccessorAsync

bp0002:
When the value of 'accessor.byteOffset' is 0, it should also pass the test. In my project, this caused a lot of waste of memory.


bp0002:
> When the value of 'accessor.byteOffset' is 0, it should also pass the test. In my project, this caused a lot of waste of memory.

bp0002:
Is my wrong understanding, should I use 'sparse'.  :(

PR #10409, opened 2021-05-24
TransmissionHelper: Add generateMipmaps option

Popov72:
See https://forum.babylonjs.com/t/transmission-helper-optimisation/21037/3

### DIFF
--- a/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
+++ b/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
@@ -44,9 +44,14 @@ interface ITransmissionHelperOptions {
     lodGenerationOffset: number;
 
     /**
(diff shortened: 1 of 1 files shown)


deltakosh (approved):

Popov72:
The test failing is "GUI Near Menu" but there's no relationship between this test and the changes from this PR. Also, the test does work in WebGL1, so I guess we can suppose it's a transient error...

deltakosh:
Adding @RaananW who should know what this test is about

deltakosh:
@CraigFeldspar can you check why the GUI near menu test is sometimes failing?

CraigFeldspar:
> @CraigFeldspar can you check why the GUI near menu test is sometimes failing?

@deltakosh I think this is because of the XHR for the pin icon that sometimes arrives late.
I will remove the test from auto testing and only keep it local

CraigFeldspar:
@deltakosh #10415

Issue #9336, opened 2020-11-03
[XR] Add sliding to the WebXR movement method


nikaera:
**Is your feature request related to a problem? Please describe.**
Whenever I create VR content in Babylon.js, I implement the same movement method as in Playground below.
https://playground.babylonjs.com/#INBVUY#58

The following issues may also be relevant.
https://github.com/BabylonJS/Babylon.js/issues/7442

**Describe the solution you'd like**
I'd like to be able to use the above-mentioned movement method with the `enableFeature` of the `FeatureManager` like the [Teleportion Module](https://doc.babylonjs.com/how_to/webxr_selected_features#enabling-teleportation).

When implementing the movement method, it would be better if you can set the following parameters.
- Movement speed
- Restrictions on movement in the horizontal direction only (like [the Playground](https://playground.babylonjs.com/#INBVUY#58))
  - If you remove the movement restrictions, you'll be able to move in all directions, as if you were flying through the air.

**Discussion**
https://forum.babylonjs.com/t/add-sliding-to-the-webxr-movement-method/15351

nikaera:
It would be nice to have the option to show a grid when you move or to darken the edge of the screen in the direction you are moving to make it less likely to cause VR sickness.

https://www.gdcvault.com/play/1023922/Full-Speed-Flying-in-VR

Issue #7442, opened 2020-01-14
[WebXR] Allow movement with controllers


RaananW:
As the WebXR controllers are no longer gamepads, there is no support for the gamepad input on the webxr camera.

Allow the user to move freely using the controller's thumbstick, if available.

RaananW:
When this is enabled, we can also turn collision detection on for the parent XR camera.

toedter:
This would be great!

PR #10410, opened 2021-05-25
add support for XR Movement Controller

brianzinn:
Wanted to kick off a starting point with an early working version.

https://github.com/BabylonJS/Babylon.js/issues/7442
https://github.com/BabylonJS/Babylon.js/issues/9336

# includes
- dynamic motion controller handling through custom configurations.  handlers passed all needed arguments for strong typing to avoid knowlege of feature class itself (like SPS).
- FeaturesManager disallows enabling while conflicting 'Teleportation' feature is enabled (and vice versa)
- configuration for movement with rotation not following viewer pose
- movement and rotation have independant thresholds and can be enabled/disabled.
- test playground `4GC64Z#13`

### DIFF
--- a/src/XR/features/WebXRControllerMovement.ts
+++ b/src/XR/features/WebXRControllerMovement.ts
@@ -1,0 +1,608 @@
+import { WebXRFeaturesManager, WebXRFeatureName } from "../webXRFeaturesManager";
+import { Observer } from "../../Misc/observable";
+import { WebXRSessionManager } from "../webXRSessionManager";
+import { Nullable } from "../../types";
+import { WebXRInput } from "../webXRInput";
+import { WebXRInputSource } from "../webXRInputSource";
+import { WebXRControllerComponent, IWebXRMotionControllerAxesValue, IWebXRMotionControllerComponentChangesValues } from "../motionController/webXRControllerComponent";
+import { AbstractMesh } from "../../Meshes/abstractMesh";
+import { Matrix, Quaternion, Vector3 } from "../../Maths/math.vector";
+import { Ray } from "../../Culling/ray";
+import { WebXRAbstractFeature } from "./WebXRAbstractFeature";
+import { MotionControllerComponentType } from "../motionController/webXRAbstractMotionController";
+import { Tools } from "../../Misc/tools";
+import { PrecisionDate } from "../../Misc/precisionDate";
+
+/**
+ * These restrictions are applied to movements (not rotations).
+ */
+export type WebXRControllerMovementRestrictions = {
+    /**
+     * Only allow movement on floor meshes within additional constraints.
+     * Default: true
+     */
+    floorMeshesOnly: boolean
+    /**
+     * Can we fall to a lower floor mesh when moving?
+     * Default: user height (or 1.8 when not available) x 3.  Set to 0 to disable.
+     */
+    fallingDistance: number
+    /**
+     * Can we go up to a higher floor (ie: stairs) when moving?
+     * Default: user height (or 1.8m when not available).  Set to 0 to disable.
+     */
+    climbingDistance: number
+};
+
+/**
+ * The options container for the controller movement module
+ */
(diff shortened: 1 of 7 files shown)


brianzinn:
@RaananW would appreciate if you can review PR when you have time.  thanks.

src/XR/features/WebXRControllerMovement.ts
```diff
@@ -0,0 +1,547 @@
...
+ */
+export type WebXRControllerMovementRegistrationConfiguration = {
+    /**
+     * handlers are filtered to these types only
+     */
+    allowedComponentTypes?: MotionControllerComponentType[]
+    /**
+     * For registering movement to specific hand only.  Useful if your app has a "main hand" and "off hand" for determining the functionality of a controller.
+     */
+    handedness?: XRHandedness
```
RaananW, line 89: teleportation calls it `forceHandedness`, would be great to stay consistent.

src/XR/features/WebXRControllerMovement.ts
```diff
@@ -0,0 +1,547 @@
...
+    /**
+     * For main component only (useful for buttons and may not trigger axis changes)
+     */
+    mainComponentOnly?: boolean
+    /**
+     * Additional predicate to apply to controllers to restrict a handler being added.
+     */
+    componentSelectionPredicate?: (xrController: WebXRInputSource) => Nullable<WebXRControllerComponent>
+} & ({
```
RaananW, line 98: formatting feels a bit off here.

src/XR/features/WebXRControllerMovement.ts
```diff
@@ -0,0 +1,547 @@
...
+
+    /**
+     * The module's name
+     */
+    public static readonly Name = WebXRFeatureName.MOVEMENT;
+
+    /**
+     * Standard controller configurations.
+     */
+    public static readonly REGISTRATIONS: Record<string, WebXRControllerMovementRegistrationConfiguration[]> = {
```
RaananW, line 153: I find `{[key:string]: WebXRControllerMovementRegistrationConfiguration[]}` better, especially since we don't use Record in the framework. Is there a difference between the two?

src/XR/features/WebXRControllerMovement.ts
```diff
@@ -0,0 +1,547 @@
...
+     * The (Babylon) version of this module.
+     * This is an integer representing the implementation version.
+     * This number does not correspond to the webxr specs version
+     */
+    public static readonly Version = 1;
+
+    /**
+     * An event triggered after XR frame completes
+     */
+    public onAfterXRFrameObservable = new Observable<WebXRControllerMovementState>();
```
RaananW, line 191: How is this different than adding a new onFrame observable after this one was created?
What will be the use for those callbacks?

Also, I think naming here is not entirely correct - it's after this component was done, not after (or before) the xr frame

RaananW (commented):
LGTM! I just have a few comments.

I also want to pull the branch and test it, approval comes right afterwards :-)

src/XR/features/WebXRControllerMovement.ts
brianzinn: I ran it through the Prettier Playground!!  I saw you added a .prettierrc some time ago and then it doesn't look like it's still there.  tslint was deprecated in 2019.  would be interesting eslint + prettier, but even on small projects that is a massive undertaking.

src/XR/features/WebXRControllerMovement.ts
brianzinn: removed those observables

src/XR/features/WebXRControllerMovement.ts
brianzinn: agreed.  i think it's better too, more natural for strings.

RaananW (approved):

RaananW:
Don't forget to add the feature to the docs

PR #10411, opened 2021-05-25
Scale gizmo & physics debug view fix

CedricGuillemet:
Follow up on https://forum.babylonjs.com/t/gltf-basic-physics/20849/2 for improved physics debug view
Missing compounds for Sphere and Cylinders

and https://forum.babylonjs.com/t/scaling-bug-using-gizmo-manager/21062/7 for uniform scale.
Mesh genuine scale was applied twice.


Popov72 (approved):
LGTM!

PR #10420, opened 2021-05-26
Stencil and back buffer YFlip

CedricGuillemet:
- Expose Stencil values and apply stencil state in Native
~~- Expose originBottomLeft readonly value~~
~~- If originBottomLeft is not true (D3D, Metal), then render a fullscreen triangle to do a YFlip or backbuffer picture will be upside down. That particular drawcall is done in `endFrame` override~~

Related PR on Native : https://github.com/BabylonJS/BabylonNative/pull/739


src/Engines/nativeEngine.ts
```diff
@@ -884,6 +935,59 @@ export class NativeEngine extends Engine {
...
+                    vUV = position * 0.5 + 0.5;
+                    vUV.y = 1.0 - vUV.y;
+                }`;
+            Effect.ShadersStore["fullscreenFlipFragmentShader"] = `
+                varying vec2 vUV;
+                uniform sampler2D backBuffer;
+                void main(void)
+                {
+                    vec3 color = texture2D(backBuffer, vUV).xyz;
+                    gl_FragColor = vec4(color, 1.0);
```
Popov72, line 956: The backBuffer can't have some alpha != 1?

src/Engines/nativeEngine.ts
```diff
@@ -884,6 +935,59 @@ export class NativeEngine extends Engine {
...
         this._shaderProcessor = new NativeShaderProcessor();
+
+        if (!this.originBottomLeft) {
+            // prepare YFlip shader and vertex stream
+            Effect.ShadersStore["fullscreenFlipVertexShader"] = `
+                attribute vec2 position;
+                attribute vec2 uv;
+                varying vec2 vUV;
+                void main(void) {
+                    gl_Position = vec4(position, 0.5, 1.0);
```
Popov72, line 946: Why not `0` for z?

src/Engines/nativeEngine.ts
```diff
@@ -884,6 +935,59 @@ export class NativeEngine extends Engine {
...
+                varying vec2 vUV;
+                uniform sampler2D backBuffer;
+                void main(void)
+                {
+                    vec3 color = texture2D(backBuffer, vUV).xyz;
+                    gl_FragColor = vec4(color, 1.0);
+                }`;
+
+            this._YFlip = {
+                vertexBuffers: { [VertexBuffer.PositionKind]: new VertexBuffer(this, [3, -1, -1, -1, -1, 3], VertexBuffer.PositionKind, false, false, 2) },
```
Popov72, line 960: Is it faster to draw a single big triangle that goes outside the limit of the windows than two triangles that fit exactly in?

Popov72 (approved):
LGTM!

src/Engines/nativeEngine.ts
```diff
@@ -87,6 +87,43 @@ interface INativeEngine {
     readonly ALPHA_SCREENMODE: number;
 
     readonly homogeneousDepth: boolean;
+    readonly originBottomLeft: boolean;
```
syntheticmagus, line 90: I might be thinking about this wrong, but it seems odd to me for JavaScript to know about and fix what I think of as a fundamentally native problem. Off the top of my head, this feels like a setting that should maybe be provided per framebuffer by the Graphics component, then leveraged where appropriate by the NativeEngine down in the C++ layer. Is the idea of screen origin location ever exposed to JavaScript by browsers?

src/Engines/nativeEngine.ts
CedricGuillemet: It won't be blended. It's a pure color copy. Alpha has no use in that case.

src/Engines/nativeEngine.ts
CedricGuillemet: Any [0..1] value is valid.

src/Engines/nativeEngine.ts
CedricGuillemet: Technically, yes. GPU computes 2x2 Tiles. With 2 triangles, the 2x2 tiles on the screen diagonal would be computed twice. In practice, it won't make a difference. It's just using a bit less memory, less code,...

src/Engines/nativeEngine.ts
CedricGuillemet: I agree YFlip is only a native problem. In practice, NativeShaderProcessor has to append the flip, so this boolean has to be exposed. Concerning the fullscreen pass, same, only a native problem and the pass 'should' be done on the c++ side. But, for maintenance reason, I don't want to maintain a precompiled shader version for all the platforms (build shaderc, have bgfx shader generated, bin2h, integrate shaders in the solution, ...). Just building shaderc and all its dependencies will blow up the compilation time on the ci. It's a trade off.

> Is the idea of screen origin location ever exposed to JavaScript by browsers?

Yes and no. There is an implicit difference between webgl and webgpu. Depending on the API you use, you have to flip or not.

Popov72:
Is it possible to restart the WebGL1 tests? They are working in #10424 that has been created more recently.

src/Engines/Native/nativeShaderProcessors.ts
```diff
@@ -15,6 +15,9 @@ export class NativeShaderProcessor extends WebGL2ShaderProcessor {
         if (!isFragment && !(<NativeEngine>engine).homogeneousDepth) {
             const lastClosingCurly = code.lastIndexOf("}");
             code = code.substring(0, lastClosingCurly);
+            if (!(<NativeEngine>engine).originBottomLeft) {
+                code += "gl_Position.y *= -1.;\n";
+            }
```
bghgary, line 20: I thought we moved all of the shader manipulation logic to the native side? When did we add this back? :)

If we moved to the native side, then you don't need to expose `homogeneousDepth` and `originBottomLeft` anymore, is that right?

CedricGuillemet: How hard would it be to do it in shader traversals @Drigax ? I didn't look at glslang yet

bghgary: I'm not sure you need shader traversal. We can manipulate the strings just the same as you are doing here.

syntheticmagus: This should not be hard to do in a traverser. All it should require is figuring out how to express this operation (most easily done by inspecting an example in the debugger) then creating a helper to replicate that, effectively injecting this line into the AST when appropriate.

deltakosh:
ok to merge?

CedricGuillemet:
@deltakosh Nop, this PR will break BabylonNative without this one : https://github.com/BabylonJS/BabylonNative/pull/739
I can set this one to draft until then ?

deltakosh:
yes please :D

PR #10421, opened 2021-05-26
Fix glTF validation tests

Popov72:


src/Engines/WebGPU/webgpuComputeContext.ts
```diff
@@ -96,6 +96,7 @@ export class WebGPUComputeContext implements IComputeContext {
                     }
                 }
             }
+            console.log(bindGroupEntries);
```
bghgary, line 99: accident?

Popov72: corrected

deltakosh (approved):

PR #10432, opened 2021-05-28
NME: Add support for parallax mapping to the PerturbNormal block

Popov72:
See https://forum.babylonjs.com/t/is-there-the-way-to-use-parallax-mapping-for-node-material/16903/18

image

For parallax (not occlusion) mapping, you need to bind something to the `parallaxHeight` input (generally the alpha component of the normal map).

For parallax occlusion mapping, what I did to overcome the problem of sampling multiple times the texture is that it assumes that the `normalMapColor` input is coming from a `Texture` block. That way, I can retrieve the sampler name used by that block and reuse it for sampling the texture myself (in the `parallaxOcclusion` shader function). If `normalMapColor` is not coming from a `Texture` block then parallax occlusion is disabled.

Parallax occlusion is enabled through a switch on the `PerturbNormal` block:
image

There's a new `uvOffset` output that ones must add to the regular uv coordinates before a texture lookup if parallax must be taken into account (it is (0,0) if parallax is not used).

Of note: `PerturbBlock` can be added a single time in a material, so it simplifies things (the `uvOffset` variable from `bumpFragment.fx` can be reused).


deltakosh (approved):
You were braver than me :)

Popov72:
I took a look to see if it could be not too difficult to add the support.

It was not too hard until I reached parallax occlusion, but I did not want to throw away what I did so I managed to make something work without too much hassle :)

I think the right thing to do would be to have something like an `InternalTexture` block (or maybe `Picture` block) which would be a wrapper around a picture and that would simply provide a sampler as its output. The `PerturbBlock` block could consume it as an input instead/in addition of/to the current `normalMapColor` input.

This `InternalTexture` block would also allow to make multiple lookups in a texture without consuming lots of samplers in the shader code: currently, if you want to sample 4 times the same texture (with different uvs!), you must create 4 `Texture` blocks and you will end up with 4 `sampler2D` in the shader code (which all points to the same picture, which is a waste). By decoupling the picture from the sampling we could reuse a single sampler multiple times. But this requires a bit of work, so I used a workaround for the parallax thing.

deltakosh:
I like this internalBlock idea. Let's put a pin on it and come back to it when we will have time

PR #10436, opened 2021-05-28
Fix #10433

### LINKED ISSUE #10433: Multiple objects leaks when loading assets with scene loader
**Repro**
1. Open https://playground.babylonjs.com/#WV6GBY in Chromium-based browser.
  (It's easier to repro with localDev to avoid playground specific issues)
2. Open the development console.
3. Type `queryObjects(Promise)`

Current result:
image
(Note that some pending ones are leaks from the calling `queryObjects` itself)

Expected result:
* No TypedArray, AssetContainer, or any other objects from scene loader should be leaked via leaked Promises

See forum for more context: https://forum.babylonjs.com/t/i-discovered-the-worst-and-most-insidious-memory-leak-in-babylonjs-history/21170

deltakosh:


bghgary (approved):
What happens if assetContainer.addAllToScene is called? Does it clear out the parentContainer for all of the objects? I'm just wondering when the asset container will be collected if the objects are all pointing to the container.

src/assetContainer.ts
```diff
@@ -469,6 +469,7 @@ export class AssetContainer extends AbstractScene {
 
             if (move) {
                 targetAssets.push(asset);
+                (asset as any)._parentContainer = this;
```
Popov72, line 472: I wonder if we could not have something like:
```javascript
class AssetStorable {
    /** @hidden */
    public_parentContainer: AssetContainer;

    public removeFromContainer(arrayName: string): void {
        // we could also pass the arrayName in the constructor of the class and keep it as a property
        // remove from this._parentContainer[arrayName]
    }
}
```
That way we could declare:
```typescript
private _moveAssets<T extends AssetStorable>(sourceAssets: T[], targetAssets: T[], keepAssets: T[]): void {
```
instead of
```typescript
private _moveAssets<T>(sourceAssets: T[], targetAssets: T[], keepAssets: T[]): void {
```
and no need to cast to any in `(asset as any)._parentContainer = this;`.

That would avoid duplicating numerous times the same code but would mean all classes that can be stored in a asset container must extend `AssetStorable`.

Popov72 (approved):
LGTM!

src/assetContainer.ts
deltakosh: I like the idea but I will keep it for another PR

PR #10440, opened 2021-05-31
3D GUI Slate advanced interactions

CraigFeldspar:
## `SixDofDragBehavior` 
 - Separated the virtual mesh position/orientation into a `BaseSixDofDragBehavior` to be able to get low-level pointer 3D positions without having the attached mesh move or rotate. This has been done in order to have more control on what to do with pointer positions.
 - Can now handle multiple pointers at the same time
 - Added a `draggableMeshes` list in order to have only a subset of descendant meshes take pointer events into account.

## `HolographicSlate`
 - Added a behavior for the content of the 3D slate. It is now draggable on x/y axii. Content scaling with 2 pointers is still TODO
 - Slate have better default world dimensions for XR experiences. They also spawn in front of the camera by default.
 - Slate now use the new `FluentBackplateMaterial` from @rgerd 

## `SlateGizmo`
 - Rewrote better structured code for the slate gizmo and its handles.
 - Added a new material for gizmo handles so they change scale/color when hovered or dragged

## `UtilityLayerRenderer`
 - Fixed a bug where pointer up/move events were not sent to 3D controls when the utility layer renderer is not hit by the picking ray.

## `WebXRDefaultExperience`
 - It is now possible to pass pointer selection options when creating the experience, so both left and right pointers can have their ray/pickpoint meshes visual cues.


src/Behaviors/Meshes/baseSixDofDragBehavior.ts
```diff
@@ -0,0 +1,378 @@
...
+    /**
+     * In case of multipointer interaction, all pointer ids currently active are stored here
+     */
+    public currentDraggingPointerIds: number[] = [];
+
+    /**
+     * Get or set the currentDraggingPointerId
+     * @deprecated Please use currentDraggingPointerId instead
+     */
+    public get currentDraggingPointerID(): number {
```
RaananW, line 78: Do we need to add those? no need to support an API we didn't have until now. Unless I am missing something :-)

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
```diff
@@ -0,0 +1,378 @@
...
+    /**
+     *  Initializes the behavior
+     */
+    public init() {}
+
+    /**
+     * In the case of multiple active cameras, the cameraToUseForPointers should be used if set instead of active camera
+     */
+    private get _pointerCamera() {
+        if (this._scene.cameraToUseForPointers) {
```
RaananW, line 131: just a thought - maybe define this as a getter  in the scene instead? so the camera that is set to be the camera to use for pointers will return (if defined), otherwise the scene active camera will return. This way we can stay consistent between components. 
Not sure what the API change requirements here, so - take that as an untested suggestion :-)

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
```diff
@@ -0,0 +1,378 @@
...
+/**
+ * Base behavior for six degrees of freedom interactions in XR experiences.
+ * Creates virtual meshes that are dragged around
+ * And observables for position/rotation changes
+ */
+export class BaseSixDofDragBehavior implements Behavior<Mesh> {
+    protected static _virtualScene: Scene;
+    private _pointerObserver: Nullable<Observer<PointerInfo>>;
+    private _attachedToElement: boolean = false;
+    protected _virtualMeshesInfo: {
```
RaananW, line 21: Couldn't find any dispose function in this class. Is detach considered to be dispose in this case?

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
```diff
@@ -0,0 +1,378 @@
...
+                        zDragFactor = 0;
+                    }
+
+                    if (this._pointerCamera && this._pointerCamera.cameraRigMode == Camera.RIG_MODE_NONE && !this._pointerCamera._isLeftCamera && !this._pointerCamera._isRightCamera) {
+                        pointerInfo.pickInfo.ray.origin.copyFrom(this._pointerCamera!.globalPosition);
+                        zDragFactor = 0;
+                    }
+
+                    // Calculate controller drag distance in controller space
+                    var originDragDifference = pointerInfo.pickInfo.ray.origin.subtract(virtualMeshesInfo.lastOriginPosition);
```
RaananW, line 301: could you use subtractToRef with temp vectors?

If not - (nit) - change to `const`

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
```diff
@@ -0,0 +1,378 @@
...
+
+                    virtualMeshesInfo.originMesh.addChild(virtualMeshesInfo.dragMesh);
+                    virtualMeshesInfo.originMesh.addChild(virtualMeshesInfo.pivotMesh);
+
+                    this._applyZOffset(virtualMeshesInfo.dragMesh, localOriginDragDifference, zDragFactor);
+                    this._applyZOffset(virtualMeshesInfo.pivotMesh, localOriginDragDifference, zDragFactor);
+
+                    // Update the controller position
+                    virtualMeshesInfo.originMesh.position.copyFrom(pointerInfo.pickInfo.ray.origin);
+                    virtualMeshesInfo.originMesh.lookAt(pointerInfo.pickInfo.ray.origin.add(pointerInfo.pickInfo.ray.direction));
```
RaananW, line 313: AddToRef or `addInPlace` (if possible)

src/Behaviors/Meshes/sixDofDragBehavior.ts
```diff
@@ -94,236 +48,181 @@ export class SixDofDragBehavior implements Behavior<Mesh> {
...
+        const worldPivot = this._ownerNode.getAbsolutePivotPoint();
+        this._ownerNode.setParent(null);
+
+        if (pointerCount === 1) {
+            this._targetPosition.copyFrom(this._ownerNode.position);
+            this._targetOrientation.copyFrom(this._ownerNode.rotationQuaternion);
+            this._targetScaling.copyFrom(this._ownerNode.scaling);
+
+            if (this.faceCameraOnDragStart && this._scene.activeCamera) {
+                const toCamera = this._scene.activeCamera.position.subtract(worldPivot).normalize();
```
RaananW, line 174: try using ref-vectors/quaternions in this and the next few lines

src/XR/webXRDefaultExperience.ts
```diff
@@ -140,10 +144,12 @@ export class WebXRDefaultExperience {
 
                 if (!options.disablePointerSelection) {
                     // Add default pointer selection
-                    result.pointerSelection = <WebXRControllerPointerSelection>result.baseExperience.featuresManager.enableFeature(WebXRControllerPointerSelection.Name, options.useStablePlugins ? "stable" : "latest", <IWebXRControllerPointerSelectionOptions>{
-                        xrInput: result.input,
-                        renderingGroupId: options.renderingGroupId,
-                    });
+                    const pointerSelectionOptions = options.pointerSelectionOptions || {} as IWebXRControllerPointerSelectionOptions;
```
RaananW, line 147: nit
```suggestion
                    const pointerSelectionOptions = {
                      ...options.pointerSelectionOptions,
                      xrInput: result.input, 
                      renderingGroupId: options.renderingGroupId
                    }
```

RaananW (commented):
Most of my comments are about reusing objects where possible (reference/temp vectors/quats/matrix).

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
CraigFeldspar: This was in `SixDofDragBehavior.ts`, so I kept it for backwards compatibility

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
CraigFeldspar: This was also in `SixDofDragBehavior.ts`, but I agree it should be in the scene as it sound like something we could reuse. But isn't it a problem to add yet another property in the scene ? I think @sebavan mentionned that we shouldn't add too many to not trigger the slow path ?

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
CraigFeldspar: Yes, all behaviors have a `detach` function but no `dispose` in the interface

gui/src/3D/controls/holographicSlate.ts
```diff
@@ -254,20 +297,65 @@ export class HolographicSlate extends ContentDisplay3D {
         return node;
     }
 
+    private _attachContentPlateBehavior() {
+        this._contentDragBehavior.attach(this._contentPlate);
+        this._contentDragBehavior.moveAttached = false;
+        this._contentDragBehavior.useObjectOrientationForDragging = true;
+        this._contentDragBehavior.updateDragPlane = false;
+
+        let origin = new Vector3();
```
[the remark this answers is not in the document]
CraigFeldspar: Actually I think this is better to keep it as local variables, since it is only used as temp storage in the subsequent functions. There are similar patterns in `Behavior`s as well. In that case, it is not putting any strain on the GC since there is no creation of objects (`_attachContentPlateBehavior` is only called once)

gui/src/3D/controls/holographicSlate.ts
```diff
@@ -254,20 +297,65 @@ export class HolographicSlate extends ContentDisplay3D {
...
+        this._backPlateMaterial = new FluentBackplateMaterial(this.name + "plateMaterial", mesh.getScene());
 
         this._pickedPointObserver = this._host.onPickedPointChangedObservable.add((pickedPoint) => {
-            if (pickedPoint) {
-                this._backPlateMaterial.hoverPosition = pickedPoint;
-                this._backPlateMaterial.hoverColor.a = 1.0;
-            } else {
-                this._backPlateMaterial.hoverColor.a = 0;
-            }
+            // if (pickedPoint) {
```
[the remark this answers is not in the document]
CraigFeldspar: Yes, the new FluentBackplateMaterial still doesn't have pointer position input. @rgerd is this tracked in an issue ? maybe I can add this in a comment next to this code

gui/src/3D/gizmos/gizmoHandle.ts
```diff
@@ -0,0 +1,216 @@
...
+    DRAG = 2,
+}
+
+/**
+ * Base class for SlateGizmo handles
+ */
+export abstract class GizmoHandle {
+    protected _scene: Scene;
+    protected _state: HandleState = HandleState.IDLE;
+    protected _materials: HandleMaterial[] = [];
```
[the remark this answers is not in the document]
CraigFeldspar: no it's a mistake, good catch !

gui/src/3D/controls/holographicSlate.ts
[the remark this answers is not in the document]
RaananW: So my only comment is that most of them can be consts :-)

RaananW:
@CraigFeldspar - can this be combined with this PR - https://github.com/BabylonJS/Babylon.js/pull/10222 ? will there be a need to make some changes after merging the near-interaction PR?

RaananW (approved):

CraigFeldspar:
> @CraigFeldspar - can this be combined with this PR - #10222 ? will there be a need to make some changes after merging the near-interaction PR?

It should since near interaction PR uses `simulatePointerXXX` and my code hooks on that. Gonna check though, I'm putting this in draft for a while to prevent wild merge

CraigFeldspar:
@RaananW it indeed broke (at least in WebXR). I'm now receiving 2 pointermove events each frame, one from `WebXRControllerPointerSelection`, and one from `WebXRNearInteraction`. And the one from the near interaction feature sends a (0, 0, 0) origin for the controller ray. I'm looking further into it and try to find a fix. Ideally, we will want to have only 1 pointermove event each frame to avoid unnecessary operations

CraigFeldspar:
Hi @satyapoojasama , I'm asking you because I see you authored #10222, should we manually call `setFarInteractionFeature` to avoid having both pointer events from far and near interaction on the same frame ?

RaananW:
The two events should have a different id. You should be able to distinguish between the two

RaananW:
But i agree, there should be only one move event, either near or far.

RaananW:
This should fix it - https://github.com/BabylonJS/Babylon.js/pull/10463

Pointer move events of the near interaction will only be triggered when the far interaction is not disabled (or doesn't exist)

CraigFeldspar:
> This should fix it - #10463
> 
> Pointer move events of the near interaction will only be triggered when the far interaction is not disabled (or doesn't exist)

Yep works, but we still send 2 pointerDown and pointerUp, right ?
Shouldn't cause any problem given that SixDofDragBehavior will just reset position twice, but maybe we should tend to something cleaner where component push their pickingInfo to a master component that will be the only one to fire events ? WDYT ?

RaananW:
the same check can be done for the simulate pointer up and down. I'll add it.

About setting the farInteraction - this should happen automatically if you are using the default experience helper. If you don't it means you enabled the feature on your own and should set the farInteraction feature yourself.

CraigFeldspar:
> the same check can be done for the simulate pointer up and down. I'll add it.
> 
> About setting the farInteraction - this should happen automatically if you are using the default experience helper. If you don't it means you enabled the feature on your own and should set the farInteraction feature yourself.

I see nowhere in the code where we set automatically `_farInteractionFeature` or `farInteractionFeature` getter. Maybe I'm missing something ?

RaananW:
> > the same check can be done for the simulate pointer up and down. I'll add it.
> > About setting the farInteraction - this should happen automatically if you are using the default experience helper. If you don't it means you enabled the feature on your own and should set the farInteraction feature yourself.
> 
> I see nowhere in the code where we set automatically `_farInteractionFeature` or `setFarInteractionFeature`. Maybe I'm missing something ?

https://github.com/BabylonJS/Babylon.js/blob/master/src/XR/webXRDefaultExperience.ts#L173

RaananW:
I pushed further changes to the near interaction - it will now only trigger if in the hover range and far interaction is not attached. 
There is some room for improvement there, but this should solve your issue.

CraigFeldspar:
I'm using this PG : http://localhost:1338/Playground/index-local.html#2YZFA0#207 with `const xr = await scene.createDefaultXRExperienceAsync(...` and somehow `this._farInteractionFeature` is undefined, I'm looking into it

CraigFeldspar:
Oh, `this._options.farInteractionFeature` is defined, but not `this._farInteractionFeature`, easy fix 😄

RaananW:
merge from master (you might have a few conflicts)

CraigFeldspar:
@RaananW everything seems in order 👍

PR #10442, opened 2021-06-01
WebGPU: add support for offscreen canvas

Popov72:


deltakosh (approved):

sebavan (approved):

Popov72:
Added an option to not perform the final copy pass to invert Y the texture, as an optimization. The user will have to add `canvas.style.transform = "scaleY(-1)"` on the canvas from which they get the offscreen canvas before passing it to WebGPUEngine.

As it does not work out of the box, this option is disabled by default.

Popov72:
The error does not seem related to the PR:
image

PR #10457, opened 2021-06-03
Add some more parameters and a camera fix

bghgary:
Some changes needed for 3D commerce certification testing.

### DIFF
--- a/sandbox/src/sandbox.tsx
+++ b/sandbox/src/sandbox.tsx
@@ -6,6 +6,11 @@ import { ReflectorZone } from "./components/reflectorZone";
 import { Footer } from "./components/footer";
 import { EnvironmentTools } from "./tools/environmentTools";
 import { Vector3 } from "babylonjs/Maths/math.vector";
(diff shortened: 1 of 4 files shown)


sandbox/src/sandbox.tsx
```diff
@@ -122,6 +141,22 @@ export class Sandbox extends React.Component<ISandboxProps, { isFooterVisible: b
...
+                    case "skybox": {
+                        this._globalState.skybox = (value === "true" ? true : false);
+                        break;
+                    }
+                    case "clearColor": {
+                        this._clearColor = value;
+                        break;
+                    }
+                    case "camera": {
+                        this._camera = +value;
```
deltakosh, line 157: why +?

deltakosh (approved):

sandbox/src/sandbox.tsx
bghgary: converts to number

deltakosh: ok makes sense

Issue #8904, opened 2020-09-08
[XR] Lighting Estimation Feature


ankitcsr:
https://immersive-web.github.io/lighting-estimation/

brianzinn:
I've started a branch:
https://github.com/brianzinn/Babylon.js/commit/ebd1f40ac64f4acbd924b9d3486c4c181bd0a3e6

I'll work to get the `getReflectionCubeMap` and `LightingEstimate` (`primaryLightDirection`/`primaryLightIntensity` and `sphericalHarmonicsCoefficients`).

I will get the XR feature to the point of making the data available.  One thing to note is that I think the babylon SphericalHarmonics is to order^2 and I think it would need to support order ^3 with the 27 element array, but I haven't checked thoroughly.

PR #10461, opened 2021-06-04
Light estimation

brianzinn:
initial support for light estimation
https://github.com/BabylonJS/Babylon.js/issues/8904

### DIFF
--- a/src/XR/features/WebXRLightEstimation.ts
+++ b/src/XR/features/WebXRLightEstimation.ts
@@ -75,6 +75,7 @@ export class WebXRLightEstimation extends WebXRAbstractFeature {
      * The XRLightProbe object created during attach (may not be available for a few frames depending on device).
      *
      * The XRLightProbe itself contains no lighting values, but is used to retrieve the current lighting state with each XRFrame.
+     * XR Light Probe also contains the probe space, may update it's pose over time as the user moves around their environment.
      */
     public get xrLightProbe(): Nullable<XRLightProbe> {
         return this._xrLightProbe;
--- a/src/XR/features/WebXRLightEstimation.ts
+++ b/src/XR/features/WebXRLightEstimation.ts
@@ -1,0 +1,114 @@
+import { Tools } from "../../Misc/tools";
+import { Nullable } from "../../types";
+import { WebXRFeatureName, WebXRFeaturesManager } from "../webXRFeaturesManager";
+import { WebXRSessionManager } from "../webXRSessionManager";
+import { WebXRAbstractFeature } from "./WebXRAbstractFeature";
+
+/**
+ * Options for Light Estimation feature
+ */
+export interface IWebXRLightEstimationOptions {
+    /**
+     * Instead of using preferred reflection format use srgba8 to initialize the light probe
+     */
+     reflectionFormatSRGBA8?: boolean;
+}
+
+/**
+ * Light Estimation Feature
+ *
+ * @since 5.0.0
+ */
+export class WebXRLightEstimation extends WebXRAbstractFeature {
+
+    /**
+     * While the estimated cube map is expected to update over time to better reflect the user's environment as they move around those changes are unlikely to happen with every XRFrame.
+     * Since creating and processing the cube map is potentially expensive, especially if mip maps are needed, pages can listen to the reflectionchange event on the XRLightProbe to
+     * determine when an updated cube map needs to be retrieved.
+     *
+     * Event Listener to for "reflectionchange" events.
+     */
+    private _reflectionChangeListener: Nullable<EventListenerOrEventListenerObject> = null;
+
+    private _xrLightProbe: Nullable<XRLightProbe> = null;
+
+    /**
+     * The module's name
+     */
+    public static readonly Name = WebXRFeatureName.LIGHT_ESTIMATION;
+    /**
+     * The (Babylon) version of this module.
+     * This is an integer representing the implementation version.
+     * This number does not correspond to the WebXR specs version
+     */
+    public static readonly Version = 1;
+
+    /**
+    * Creates a new instance of the light-estimation feature
+    * @param _xrSessionManager an instance of WebXRSessionManager
+    * @param options options to use when constructing this feature
+    */
+    constructor(
+        _xrSessionManager: WebXRSessionManager,
+        /**
+         * options to use when constructing this feature
+         */
+        public readonly options: IWebXRLightEstimationOptions
+    ) {
+        super(_xrSessionManager);
+        this.xrNativeFeatureName = "light-estimation";
+
+        // https://immersive-web.github.io/lighting-estimation/
+        Tools.Warn("light-estimation is an experimental and unstable feature.");
+    }
+
+    /**
+     * attach this feature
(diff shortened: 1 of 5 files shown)


brianzinn:
Able to retrieve primary light direction/intensity as well as ReflectionCubeMap texture.
image

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * An estimated cube map representing the users environment.
+     */
+    public get reflectionCubeMap(): Nullable<WebGLTexture> {
+        if (this._xrLightProbe === null) {
+            return null;
+        }
+
+        if (this._reflectionCubeMap = null) {
```
RaananW, line 110: I assume you meant `===`?

RaananW (commented):
We want to keep all async operations using observables or promises.

Whenever possible we should expose an updated/added/created observable, so that devs won't need to check for null (and keep on calling until it is ready). 

Another thing - I understand this is initial support, but this requires the dev to apply the light data to a light created by themselves. The options should allow the user to provide a light that will be updated, or ask to create a light to which the light data be applied.
Not sure if this is out of the scope of this PR.

src/XR/features/WebXRLightEstimation.ts
brianzinn: 🤦

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,160 @@
...
+            this._reflectionCubeMap = this._getXRGLBinding().getReflectionCubeMap(this._xrLightProbe);
+            this._xrLightProbe.addEventListener('reflectionchange', this._updateReflectionCubeMap);
```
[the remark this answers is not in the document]
brianzinn: getting the cubemap is not async (it's only getting the light probe that may take multiple frames, but on my device is available on first frame).  The event listener should actually be a separate feature option.   I'll add that change in and make it clearer.

brianzinn:
I added an observable for the Reflection Cube Map.  It is updated every 20-30 frames on my device.  I suspect you would want light direction and intensity to be `Vector3` for public API (Looks like light intensity needs to be scaled down by the highest value) instead of the Web XR specific typings.  The cube texture as well should likely update a babylon equivalent.  Hope this is enough to get light-estimation feature to next step with spherical harmonics and the WebGLTexture cube map available.

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,161 @@
...
+            this._xrWebGLBinding = new XRWebGLBinding(this._xrSessionManager.session, context);
+        }
+        return this._xrWebGLBinding;
+    }
+
+    /**
+     * Event Listener to for "reflectionchange" events.
+     */
+    private _updateReflectionCubeMap = (): void => {
+        this._reflectionCubeMap = this._getXRGLBinding().getReflectionCubeMap(this._xrLightProbe!);
```
RaananW, line 108: does the old one needs to be disposed?

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,161 @@
...
+        if (!super.attach()) {
+            return false;
+        }
+
+        this._xrSessionManager.session.requestLightProbe({
+            reflectionFormat: this.options.reflectionFormatSRGBA8 ? "srgba8" : this._xrSessionManager.session.preferredReflectionFormat ?? "srgba8"
+        }).then((value: XRLightProbe) => {
+            this._xrLightProbe = value;
+            this._xrLightProbe.addEventListener('reflectionchange', this._updateReflectionCubeMap);
+            this._reflectionCubeMap = this._getXRGLBinding().getReflectionCubeMap(this._xrLightProbe!);
```
RaananW, line 128: should probably notify here (the observable)

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,161 @@
...
+     * XR Light Probe also contains the probe space, may update it's pose over time as the user moves around their environment.
+     */
+    public get xrLightProbe(): Nullable<XRLightProbe> {
+        return this._xrLightProbe;
+    }
+
+    /**
+     * The most recent light estimate.  Available starting on the first frame where the device provides a light probe.
+     */
+    public get xrLightingEstimate(): Nullable<XRLightEstimate> {
```
RaananW, line 87: we'll need to convert those values to left-handed when used in babylon. otherwise they are flipped.

RaananW (commented):
Hey @brianzinn , thank you so much for working on that!

I added a few comments - but I will be happy to work on that if you can't find the time.

src/XR/features/WebXRLightEstimation.ts
brianzinn: @RaananW  I didn't clean up based on this:
https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.8
```
Note that underlying GL object will be automatically marked for deletion when the JS object is destroyed, however this method allows authors to mark an object for deletion early.
```

It's easy for me to mark for deletion  - I started making some changes to do so.  Shall I update with `context.deleteTexture(this._reflectionCubeMap)`?

src/XR/features/WebXRLightEstimation.ts
brianzinn: fixed.

src/XR/features/WebXRLightEstimation.ts
brianzinn: Yes @RaananW - To be consistent I think we want to remove XR specific typings from the API.  Likely that means wrapping with something like:
```typescript
export type LightEstimate = {
  primaryLightDirection: Vector3 // changed to left handed
  primaryLightIntensity: Vector3 // although not directly usable (see below)
  sphericalHarmonicCoefficients: Float32Array
}
```
I didn't go that extra step.  Following the explainer the `intensity` from the primary light is used to set color and a numeric primitive for light intensity (https://github.com/immersive-web/lighting-estimation/blob/main/lighting-estimation-explainer.md#xrlightestimate).

RaananW:
Note - I will merge this right after making a few small changes (i.e. the last discussion). Will keep this open until then.

sebavan:
Thanks @RaananW I ll leave it open for you then :-)

sebavan (changes requested):
Preventing merge before @RaananW can jump in :-)

sebavan:
@brianzinn did you have any chance to take a look at the changes ?

brianzinn:
> @brianzinn did you have any chance to take a look at the changes ?

@sebavan I have not been looking at the changes - I am under the impression that @RaananW was going to independently make changes and then to merge.  If there is something specific for me to follow up on then just let me know and I am happy to look into.  Thanks.

sebavan:
Oh my.... completely misread the thread and the worst part is I already discussed it with him.... so yes nothing required from your side for now :-)

sebavan:
Closing as @RaananW is implementing it :-)

PR #10466, opened 2021-06-04
Properly set depth function for NativeEngine

rgerd:
The shader used for the WebXR hands requires a depth prepass, and when displaying the model in native I noticed nothing was rendering to the screen. Upon further inspection, the BabylonNative engine under-the-hood defaults to a depth mode of LESS, when BabylonJS assumes the default to be LEQUAL, but there was no actual call to `_native.setDepthTest` that was setting the underlying value in the native side.

https://github.com/BabylonJS/Babylon.js/blob/d120c8e483e3e92f01deba34880552be27bfaed8/src/Engines/thinEngine.ts#L1140-L1143

NativeEngine initializes the depth test value to LEQUAL, but the value never actually gets set with a call to `_native.setDepthTest`.
https://github.com/BabylonJS/Babylon.js/blob/d120c8e483e3e92f01deba34880552be27bfaed8/src/Engines/nativeEngine.ts#L765

https://github.com/BabylonJS/Babylon.js/blob/3a9e1322b453881b59407a511fe4705fe2cd59b9/src/XR/features/WebXRHandTracking.ts#L446-L449

xr-hand
A screenshot from the BabylonNative playground with the fix applied. Previously, nothing was getting rendered to the screen because the color pass was getting discarded due to the depth values being equal.


src/Engines/engine.ts
```diff
@@ -994,7 +994,22 @@ export class Engine extends ThinEngine {
      * @param depthFunc defines the function to use
      */
     public setDepthFunction(depthFunc: number) {
-        this._depthCullingState.depthFunc = depthFunc;
+        // We delegate these calls to the named functions because
+        // NativeEngine overrides them to translate the constants.
+        switch (depthFunc) {
```
sebavan, line 999: We would in this case miss quite a few of them: 
gl.NEVER (never pass)
gl.LESS (pass if the incoming value is less than the depth buffer value)
gl.EQUAL (pass if the incoming value equals the depth buffer value)
gl.LEQUAL (pass if the incoming value is less than or equal to the depth buffer value)
gl.GREATER (pass if the incoming value is greater than the depth buffer value)
gl.NOTEQUAL (pass if the incoming value is not equal to the depth buffer value)
gl.GEQUAL (pass if the incoming value is greater than or equal to the depth buffer value)
gl.ALWAYS (always pass)

sebavan (changes requested):

src/Engines/engine.ts
sebavan: This is used with more cases than the few helper we have on the Engine class. I guess all the helpers should call into this one which is more generic and native could only implement this one instead ?

rgerd: Whoops, completely forgot about the other cases.

sebavan: nw I almost did to :-)

Popov72:
Note that in `ThinEngine` there is:
```typescript
public get depthCullingState(): DepthCullingState {
    return this._depthCullingState;
}
```
It means a user can set the depth function (and other states like zOffset, cull, cullFace, etc) directly without using the `setDepthFunction`, `setZOffset`, ... helper functions of the engine.

So, instead of overriding all those helper functions, for WebGPU I have created a `WebGPUDepthCullingState` (that extends `DepthCullingState`) to make sure all those state changes end up in the WebGPU implementation. Note that there's the same problem with the stencil, there's a specific `WebGPUStencilStateComposer` that extends `StencilStateComposer`.

rgerd:
@Popov72 This is a _much_ better idea-- to make a `NativeDepthCullingState`. As long as people are comfortable with me pushing this fix for now, I think it would be best to write an issue and consider the changes more carefully, especially since we are currently churning a lot of code on the native side around render state: https://github.com/BabylonJS/BabylonNative/pull/739.

This fix should work for now, but I totally agree that having everything tracked and applied through a state object would be much more elegant (and possibly more performant).

sebavan:
works for me let s see what @bghgary and @deltakosh  think

bghgary:
I'm okay with this for now.

bghgary (approved):

sebavan (approved):

PR #10467, opened 2021-06-04
DeviceInputSystem - Only set pointer capture when pointer lock isn't present

PolygonalSun:
This change adds some code to adjust the handling of pointer capture so that it isn't set if there is a pointer lock.


sebavan (commented):
Would be great to test accross browsers (chrome old edge safari firefox and mobile) to be fully safe ?

deltakosh:
Linting is failing also

PolygonalSun:
> Linting is failing also

Should be fixed.  I missed a space

PolygonalSun:
> Would be great to test accross browsers (chrome old edge safari firefox and mobile) to be fully safe ?

I wasn't able to test it on Old Edge but I tested this latest commit on Chrome, Edge (Chromium), Firefox (Windows/MacOS/Android), and Safari (MacOS/iOS).

sebavan:
LGTM then

PR #10472, opened 2021-06-07
SurfaceMagnetism behavior

CraigFeldspar:
This PR adds a behavior for meshes to stick on the surfaces of other meshes


gui/src/3D/behaviors/defaultBehavior.ts
```diff
@@ -69,8 +88,9 @@ export class DefaultBehavior implements Behavior<Mesh> {
      * Attaches the default behavior
      * @param ownerMesh The top level mesh
      * @param draggablesMeshes Descendant meshes that can be used for dragging the owner mesh
+     * @param sceneUnderstandingMeshes Meshes from the scene understanding that will be used for surface magnetism
      */
-    public attach(ownerMesh: Mesh, draggablesMeshes?: Mesh[]): void {
+    public attach(ownerMesh: Mesh, draggablesMeshes?: Mesh[], sceneUnderstandingMeshes?: AbstractMesh[]): void {
```
RaananW, line 93: is there a way to add/remove meshes from the meshes array?

gui/src/3D/controls/nearMenu.ts
```diff
@@ -150,8 +153,8 @@ export class NearMenu extends VolumeBasedPanel {
 
         this._pinButton.position.copyFromFloats(this._backPlate.scaling.x / 2 + 0.02, this._backPlate.scaling.y / 2, -0.01);
 
-        this._defaultBehavior.followBehavior.minimumDistance = extendSize.length() * 0.5 * this.scaling.length();
-        this._defaultBehavior.followBehavior.maximumDistance = extendSize.length() * 1.5 * this.scaling.length();
+        this._defaultBehavior.followBehavior.minimumDistance = extendSize.length() * 3 * this.scaling.length();
```
RaananW, line 156: what is the reason behind the change here? is this something we need to store as a readonly static property?

RaananW (approved):
LGTM!

Just 2 questions.

gui/src/3D/behaviors/defaultBehavior.ts
CraigFeldspar: Now yes, `control.defaultBehavior.surfaceMagnetismBehavior.meshes`. It's a bit tedious, so it may be better to keep a reference of the array

gui/src/3D/controls/nearMenu.ts
CraigFeldspar: that's a mistake, I'm reverting

CraigFeldspar: It is changed again in a more recent PR (#10485), to be a fixed value for the near menu. I will add a parameter in this PR instead

RaananW:
There are conflicts and a build error regarding `_defaultBehavior`. I'll merge right after it's fixes

PR #10473, opened 2021-06-07
Add intensityTexture to PBR sub surface

Popov72:
See https://forum.babylonjs.com/t/render-artifacts-with-khr-materials-volume-and-khr-materials-transmission/21414/2

It fixes a problem with glTF when both a transmission and a thickness textures are provided as part of the transmission / volume extensions.

Note that there's an optimisation that will only use a single sampler if both textures are identical.

### DIFF
--- a/src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
+++ b/src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
@@ -37,6 +37,9 @@ struct subSurfaceOutParams
     #ifdef SS_THICKNESSANDMASK_TEXTURE
         const in vec4 thicknessMap,
     #endif
+    #ifdef SS_INTENSITY_TEXTURE
+        const in vec4 intensityMap,
+    #endif
     #ifdef REFLECTION
         #ifdef SS_TRANSLUCENCY
             const in mat4 reflectionMatrix,
@@ -155,6 +158,23 @@ struct subSurfaceOutParams
         #endif
     #else
         float thickness = vThicknessParam.y;
+    #endif
+
+    #ifdef SS_INTENSITY_TEXTURE
+        #if defined(SS_USE_GLTF_THICKNESS_TEXTURE)
+            #ifdef SS_REFRACTION
+                refractionIntensity *= intensityMap.r;
+            #elif defined(SS_TRANSLUCENCY)
+                translucencyIntensity *= intensityMap.r;
+            #endif
+        #else
+            #ifdef SS_REFRACTION
+                refractionIntensity *= intensityMap.g;
+            #endif
+            #ifdef SS_TRANSLUCENCY
+                translucencyIntensity *= intensityMap.b;
+            #endif
+        #endif
     #endif
 
     // _________________________________________________________________________________________
--- a/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
+++ b/loaders/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
@@ -359,7 +359,7 @@ export class KHR_materials_transmission implements IGLTFLoaderExtension {
             (extension.transmissionTexture as ITextureInfo).nonColorData = true;
             return this._loader.loadTextureInfoAsync(`${context}/transmissionTexture`, extension.transmissionTexture, undefined)
                 .then((texture: BaseTexture) => {
-                    pbrMaterial.subSurface.thicknessTexture = texture;
+                    pbrMaterial.subSurface.intensityTexture = texture;
                     pbrMaterial.subSurface.useGltfStyleThicknessTexture = true;
                     pbrMaterial.subSurface.useMaskFromThicknessTexture = true;
                 });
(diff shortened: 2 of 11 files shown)


src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
```diff
@@ -157,6 +160,23 @@ struct subSurfaceOutParams
         float thickness = vThicknessParam.y;
     #endif
 
+    #ifdef SS_INTENSITY_TEXTURE
+        #if defined(SS_USE_GLTF_THICKNESS_TEXTURE)
+            #ifdef SS_REFRACTION
+                refractionIntensity *= intensityMap.r;
+            #elif defined(SS_TRANSLUCENCY)
+                translucencyIntensity *= intensityMap.r;
```
sebavan, line 168: I guess we should be able to have both here ?

sebavan (commented):
I am wondering if that works when the 3 extensions are enabled ? translucency volume and transmission ?

src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
Popov72: Need to have a look...

Popov72: We would need to support another texture...

It does not feel quite right to add a new texture on our side each time glTF uses a channel for some intensity modulation in an extension. We could end up having to support dozens of textures in the end, and we would need to add some non trivial code / obfuscate shader code with even more #defines to optimize cases when some textures are actually the same, to save samplers...

The glTF specs should rather handle cases where several extensions are used together (transmission, volume, translucency) and enforce the fact that the "intensity" texture should be the same but each extension using a different channel.

What do you think? And @bghgary?

For the time being, the translucency extension is a draft, so I don't think we should make any change on our side just yet.

Popov72: For discussion:
* rename `intensityTexture` I have just added into `refractionIntensityTexture`. Use the green channel (red in the glTF case)
* add a new `translucencyIntensityTexture`. Use the blue channel (blue also in the glTF case)

We would optimize usage of sampler only when `refractionIntensityTexture` == `translucencyIntensityTexture` == `thicknessTexture`.

We would also need to have two new properties to sub surface, `useRefractionIntensityTexture` and `useTranslucencyIntensityTexture`. For the time being, if the texture is used, we assume it is always used for both the refraction and translucency cases.

Popov72:
Added support for the translucency texture, which is required from the KHR_materials_translucency extension.

Popov72:
The error is in some WebXR unit test it seems:

```
ERROR: 'BJS - [19:54:56]: WebXRControllerMovement feature requires "xrInput" option.'
```

sebavan:
Weird, @RaananW any idea ?

RaananW:
Weird. Those tests passed when the PR was merged. You can remove them for now.

@brianzinn - want to check what happened here?

brianzinn:
I will look there - thanks for heads up @RaananW.  Those tests can be turned off and I will verify.

bghgary (approved):
LGTM

src/Shaders/ShadersInclude/pbrBlockSubSurface.fx
bghgary: We had talked about doing something smarter in terms of sampler management. I think the idea is to specify what samplers are needed and do some kind of sampler indirection (maybe with macros or something) to make it so that we don't have to do this for each new textures that is added every time.

brianzinn:
I am seeing a different test failure on this PR:
```
Babylon Scene Loader #glTF Load BoomBox twice and check texture instancing FAILED
	AssertionError: createTextureSpyCalled: expected true to be false

```

Popov72:
I have run the tests locally and all are ok.

sebavan:
Let me run the tests again just in case :-)

sebavan:
@Popov72 I guess we should remove the test and merge it ?

Popov72:
I don't know why it is failing here whereas it works on my computer...

Also, I checked using a [PG](https://playground.babylonjs.com/#HPPE3C#3) that it was really working, the `createTexture` function is never called when loading the boombox for the 2nd time...

So, I don't know what to do, I can remove the test if you want.

sebavan:
I am seeing only the xrinput as a failed test now so I guess commenting this one out should be enough ?

Popov72:
"xrInput" is generating an error log but is not counted as an error (there are other tests that also generate error logs).

The failed test is really coming from the BoomBox test (look for the first occurrence of **(1 FAILED)** line 238).

brianzinn:
@sebavan  It's just writing to log:
```
Tools.Error('WebXRControllerMovement feature requires "xrInput" option.');
```

I will sinon stub + restore that out for the test, so the logs are not confusing, but you will find there is a different failing test.

sebavan:
Ohhhh I see yup sounds guenine as it repros every time :-(

Popov72:
Ok, I found the problem, it was WebGL1 specific.

PR #10475, opened 2021-06-08
[MRTK] Hand constraint + hand menu

CraigFeldspar:
+ Adds `HandConstraintBehavior`, to make the assigned mesh follow the hand pose. This behavior requires to be linked to a XR experience with the `HandTracking` feature activated.
+ Adds `HandMenu`, a simple menu that uses `HandConstraintBehavior`.


RaananW:
This seems to be very much related to hands, but i wonder why not enable this for xr controllers as well?

CraigFeldspar:
> This seems to be very much related to hands, but i wonder why not enable this for xr controllers as well?

I'm sure it could, but I'm using a lot of hand features to calculate transforms (position of fingers, wrist, etc). Maybe in a further PR we can also make this compatible with generic controllers poses.

RaananW:
> > This seems to be very much related to hands, but i wonder why not enable this for xr controllers as well?
> 
> I'm sure it could, but I'm using a lot of hand features to calculate transforms (position of fingers, wrist, etc). Maybe in a further PR we can also make this compatible with generic controllers poses.

That's what I mean :-) 
It shouldn't be only the position of the hand but an arbitrary position (for example the location of the controller itself, or the location of a mesh parented to it).

Would be great to address this. I will create the issue

RaananW (approved):

PR #10476, opened 2021-06-08
feature: custom timeout

jamidwyer:
The default timeout for xhr and WebRequest is 0, which means there is not timeout. This allows users to set their own value for timeout.


deltakosh (approved):
Thanks!

deltakosh:
Linting is failing :(

deltakosh:
I'll take care of it

PR #10481, opened 2021-06-10
Refactor WebXRHandTracking

### LINKED ISSUE #10453: WebXRHand recreates mesh and shader on construction
**Repro**
- Test a scene on HoloLens using [BabylonNative](https://github.com/BabylonJS/BabylonNative/), and inside the call to `createDefaultXRExperienceAsync`, call:
```javascript
xr.baseExperience.featuresManager.enableFeature(
                        BABYLON.WebXRFeatureName.HAND_TRACKING,
                        "latest",
                        { xrInput: xr.input });
```
- Move your hands in and out of view, paying attention to the console, which will signal that shaders are being recompiled, and notice that performance dips considerably.

This is due to the following call in the WebXRHand constructor:
https://github.com/BabylonJS/Babylon.js/blob/3a9e1322b453881b59407a511fe4705fe2cd59b9/src/XR/features/WebXRHandTracking.ts#L297
Which causes a mesh import:
https://github.com/BabylonJS/Babylon.js/blob/3a9e1322b453881b59407a511fe4705fe2cd59b9/src/XR/features/WebXRHandTracking.ts#L432
And a material compilation:
https://github.com/BabylonJS/Babylon.js/blob/3a9e1322b453881b59407a511fe4705fe2cd59b9/src/XR/features/WebXRHandTracking.ts#L444

We should be pre-loading the meshes and materials and enabling/disabling them as hands go in and out of view, instead of recreating them every time.

### LINKED ISSUE #10454: WebXRHand doesn't efficiently obtain hand poses
The WebXR hand tracking API explainer recommends using `fillPoses` and `fillJointRadii` as a way of reducing the number of javascript objects created per-frame while updating tracked hand models:
https://github.com/immersive-web/webxr-hand-input/blob/main/explainer.md#efficiently-obtaining-hand-poses

WebXRHand still uses the inefficient method:
https://github.com/BabylonJS/Babylon.js/blob/3a9e1322b453881b59407a511fe4705fe2cd59b9/src/XR/features/WebXRHandTracking.ts#L347

In order for BabylonNative to maintain support, we will want to get https://github.com/BabylonJS/BabylonNative/issues/773 closed first.

rgerd:
This is a pretty substantial refactor of the `WebXRHandTracking` feature, which yields about a 10-15 fps improvement on HoloLens with BabylonNative.

These images kinda look the same, but the left is a representative frame timing of the original (25-35 FPS) and the right is a representative frame timing after the refactor (40-50 FPS):
<div>

</div>

The main things I did were:
1. Refactor the `IWebXRHandTrackingOptions` to make them a bit clearer by more explicitly separating hand and joint mesh options
2. Move hand mesh ownership out of `WebXRHand` and into `WebXRHandTracking` (fixes #10453)
3. Refactor the default rig mapping to generate something closer to what a user would give us, and make it easier to swap out for a different default model in the future if we choose to do so
4. Use `XRFrame.fillPoses` and `XRFrame.fillJointRadii` to more efficiently query poses (fixes #10454)
5. Refactor the logic that actually applies queried poses to the transform nodes in a way that minimizes object churn

### DIFF
--- a/src/XR/features/WebXRHandTracking.ts
+++ b/src/XR/features/WebXRHandTracking.ts
@@ -70,6 +70,10 @@ export interface IWebXRHandTrackingOptions {
          * The utilityLayer scene that contains the 3D UI elements. Passing this in turns on near interactions with the index finger tip
          */
         sceneForNearInteraction?: Scene;
+        /**
+         * Scale factor for all joint meshes (defaults to 1)
+         */
+        scaleFactor?: number;
     };
 
     /**
@@ -308,6 +312,7 @@ export class WebXRHand implements IDisposable {
      * @param _nearInteractionMesh As optional mesh used for near interaction collision checking
      * @param _leftHandedMeshes Are the hand meshes left-handed-system meshes
      * @param _jointsInvisible Are the tracked joint meshes visible
+     * @param _jointScaleFactor Scale factor for all joint meshes
      */
     constructor(
         /** The controller to which the hand correlates. */
@@ -319,7 +324,8 @@ export class WebXRHand implements IDisposable {
         readonly rigMapping: Nullable<XRHandMeshRigMapping>,
         private readonly _nearInteractionMesh?: Nullable<AbstractMesh>,
         private readonly _leftHandedMeshes: boolean = false,
-        private readonly _jointsInvisible: boolean = false
+        private readonly _jointsInvisible: boolean = false,
+        private readonly _jointScaleFactor: number = 1
     ) {
         this._scene = _jointMeshes[0].getScene();
 
@@ -430,13 +436,13 @@ export class WebXRHand implements IDisposable {
             this._tempJointMatrix.decompose(undefined, jointTransform.rotationQuaternion!, jointTransform.position);
 
             // The radius we need to make the joint in order for it to roughly cover the joints of the user's real hand.
-            const jointRadius = this._jointRadii[jointIdx];
+            const scaledJointRadius = this._jointRadii[jointIdx] * this._jointScaleFactor;
 
             const jointMesh = this._jointMeshes[jointIdx];
             jointMesh.isVisible = !this._handMesh && !this._jointsInvisible;
             jointMesh.position.copyFrom(jointTransform.position);
             jointMesh.rotationQuaternion!.copyFrom(jointTransform.rotationQuaternion!);
-            jointMesh.scaling.copyFromFloats(jointRadius, jointRadius, jointRadius);
+            jointMesh.scaling.copyFromFloats(scaledJointRadius, scaledJointRadius, scaledJointRadius);
 
             // The WebXR data comes as right-handed, so we might need to do some conversions.
             if (!this._scene.useRightHandedSystem) {
@@ -454,7 +460,7 @@ export class WebXRHand implements IDisposable {
             // Update the invisible fingertip collidable
             if (this._nearInteractionMesh && jointName == "index-finger-tip") {
                 this._nearInteractionMesh.position.copyFrom(jointTransform.position);
-                this._nearInteractionMesh.scaling.copyFromFloats(jointRadius, jointRadius, jointRadius);
+                this._nearInteractionMesh.scaling.copyFromFloats(scaledJointRadius, scaledJointRadius, scaledJointRadius);
                 if (!this._scene.useRightHandedSystem) {
                     this._nearInteractionMesh.position.z *= -1;
                 }
@@ -766,7 +772,8 @@ export class WebXRHandTracking extends WebXRAbstractFeature {
             this._handResources.rigMappings && this._handResources.rigMappings[handedness],
             touchMesh,
             this.options.handMeshes?.meshesUseLeftHandedCoordinates,
-            this.options.jointMeshes?.invisible
+            this.options.jointMeshes?.invisible,
+            this.options.jointMeshes?.scaleFactor
         );
 
         this._attachedHands[xrController.uniqueId] = webxrHand;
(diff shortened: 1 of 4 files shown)


src/XR/features/WebXRHandTracking.ts
```diff
@@ -73,32 +66,38 @@ export interface IWebXRHandTrackingOptions {
...
          * Should the default hand mesh be disabled. In this case, the spheres will be visible (unless set invisible).
          */
-        disableDefaultHandMesh?: boolean;
+        disableDefaultMeshes?: boolean;
         /**
-         * a rigged hand-mesh that will be updated according to the XRHand data provided. This will override the default hand mesh
+         * Rigged hand meshes that will be tracked to the user's hands. This will override the default hand mesh.
          */
-        handMeshes?: {
+        customMeshes?: {
```
RaananW, line 86: i totally agree with the change, but it breaks back-compat, as this was introduced in 4.2. If you can add a legacy support in the function itself it would be great.

src/XR/features/WebXRHandTracking.ts
```diff
@@ -324,112 +367,196 @@ export class WebXRHand implements IDisposable {
...
     /**
-     * Update this hand from the latest xr frame
-     * @param xrFrame xrFrame to update from
-     * @param referenceSpace The current viewer reference space
-     * @param scaleFactor optional scale factor for the meshes
+     * Update this hand from the latest xr frame.
+     * @param xrFrame The latest frame received from WebXR.
+     * @param referenceSpace The current viewer reference space.
      */
-    public updateFromXRFrame(xrFrame: XRFrame, referenceSpace: XRReferenceSpace, scaleFactor: number = 2) {
```
RaananW, line 339: did you remove scaleFactor on purpose?

src/XR/features/WebXRHandTracking.ts
```diff
@@ -324,112 +367,196 @@ export class WebXRHand implements IDisposable {
...
-                const radius = (pose.radius || 0.008) * scaleFactor;
-                mesh.scaling.set(radius, radius, radius);
-
-                // if we are using left-handed meshes, transform before applying to the hand mesh
-                if (this._leftHandedMeshes && !mesh.getScene().useRightHandedSystem) {
-                    mesh.position.z *= -1;
-                    mesh.rotationQuaternion!.z *= -1;
-                    mesh.rotationQuaternion!.w *= -1;
+
+        const anyHand: any = hand;
```
RaananW, line 400: just wondering - why is any needed? is the type not enough?

src/XR/features/WebXRHandTracking.ts
```diff
@@ -324,112 +367,196 @@ export class WebXRHand implements IDisposable {
...
+                if (jointPose) {
+                    this._jointTransformMatrices.set(jointPose.transform.matrix, jointIdx * 16);
+                    this._jointRadii[jointIdx] = jointPose.radius || 0.008;
+                } else {
+                    trackingSuccessful = false;
+                    break;
                 }
+            }
+        } else {
+            throw new Error("XRFrame does not support joint tracking!");
```
RaananW, line 420: the feature should not be enabled if joint tracking is not supported. is there a case where this will fail even if hands are supported?

RaananW (commented):
Just a fw comments. otherwise - LGTM!

Did you test with both right handed and left handed systems?

src/XR/features/WebXRHandTracking.ts
rgerd: Seems that TypeScript doesn't allow you to override the indexing ([]) operator, so I need to convert to any to allow dynamic access of members on the object

RaananW:
@rgerd - any update on that one? would be great to merge this

sebavan:
It now has couple minor conflicts as well @rgerd

rgerd:
@RaananW @sebavan Hi sorry I took some time off while working on this. Feedback will be responded to by EOD.

src/XR/features/WebXRHandTracking.ts
rgerd: I've spent a lot of time looking into this comment, and I think we're going to have to take a more in-depth look at the way we're interpreting the [WebXR spec IDL](https://immersive-web.github.io/webxr-hand-input/#xrhand-interface) for XRHand in webxr.d.ts, especially with the new changes to the spec​. 

I'm going to file an issue about exactly what I mean, but for now I think I'm going to have to leave the ​any​ cast as-is because otherwise I would have to make a bunch of crazy changes to our .d.ts.

src/XR/features/WebXRHandTracking.ts
rgerd: We leave it up to the user to not enable the feature if joint tracking / hands aren't supported? Doesn't seem like there are any explicit checks against it.

RaananW: The feature "`isCompatible` function should catch that earlier, but it's fine to stay defensive here.

src/XR/features/WebXRHandTracking.ts
RaananW: I am prefectly fine with adjusting our webxr.d.ts to fit the IDL. But you can do that in a future PR. if you do, please put a TODO above this line so we know it's not the final state

sebavan:
@rgerd Only the linting to go :-) @RaananW it looks ok to me but would like your back review :-)

RaananW (approved):

src/XR/features/WebXRHandTracking.ts
rgerd: Added legacy support by translating the options object in the constructor

PR #10485, opened 2021-06-10
Improving near interactions

### LINKED ISSUE #10471: Near interaction improvements
After trying out near interaction on 3D GUI elements, I found several improvement that I would like to discuss with you @RaananW @satyapoojasama :

+ Near grab requires a ray to be hitting the mesh to trigger when the user pinches his fingers. This is okay for bigger meshes, but for little meshes, like resize handles, i can be hard to get a hit. I feel like near grab, like near press, should pick meshes only from a distance hit test (what you implemented as `pickWithMesh`)

+ Near press is not implemented, it is not possible to press buttons just by getting your index finger close. Everything is already there though, we just need to trigger pointerDown and Up events.

+ There is a small bug where nearGrab always takes precedence on nearPick, even when no hit was found, or mesh doesn't have `isNearGrabbable` set to true (causing near pick to have a `hit === false`).

+ `pickWithMesh` is rather inefficient, we hit test a sphere mesh versus another meshes, causing a lot of triangle-triangle intersections. We can just perform distance-to-triangle tests instead.

I already started with implementing some of that because I need these improvements to test the new `NearMenu`, but I would be happy to get your opinion on these.

CraigFeldspar:
Related to #10471
+ Makes near interaction more precise and efficient (specifically by adding a sphere-to-mesh intersection test)
+ Implements near touch, to trigger button without pinching the fingers
+ Can specify a controllerId to exclude from near interactions (for hand menu to avoid near interaction with the hand that carries the menu)
+ Enables 2 pointers near interaction with SixDofDragBehavior
+ Adds a visual cue for near interaction picks

Closes #10471


src/XR/features/WebXRNearInteraction.ts
```diff
@@ -321,148 +361,172 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
-                    result = nearInteractionInfo;
-                    controllerData.nearInteraction = true;
+                    result = nearInteractionInfo!;
                 }
                 return result;
             };
 
-            let pick = null;
-            let hoverRange = false;
+            let pick = new PickingInfo();
```
RaananW, line 378: there is no reason to init pick here. It is populated (or replaced) in the next few loops. One less object to garbage-collect on each frame.

src/Meshes/abstractMesh.ts
```diff
@@ -1713,6 +1745,64 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
         return pickingInfo;
     }
 
+    /**
+     * Picks this mesh with a sphere
+     * @param sphere picking sphere in world coordinates
+     * @param skipBoundingInfo a boolean indicating if we should skip the bounding info check
+     * @returns the picking info
+     */
+    public pickWithSphere(sphere: BoundingSphere, skipBoundingInfo = false): PickingInfo {
```
RaananW, line 1754: We are trying to avoid adding new elements to mesh/abstract mesh.

Is there a reason to include the pickWithSphere here? what is the (abstract) use of this outside of XR near interactions?

RaananW (commented):
Just one comment regarding the function added to (abstract)mesh

src/Meshes/abstractMesh.ts
CraigFeldspar: I put it there because this function's logic is very similar to the above one, `intersects`. Instead of intersecting with a ray, we intersect with a sphere. So it felt like it belonged here. 
If we really want to avoid adding new members to `abstractMesh`, where do you suggest to put it ?  `WebXRNearInteraction` ?

RaananW: Yes, I would keep it there for now. 
abstractMesh has `intersectsWithMesh` as well, but this is involving picking, which should technically be on the scene level.

CraigFeldspar: So I think the best is to keep the function here, and change the name to `intersectsSphere`, which is more accurate and consistent with the above function

RaananW: intersects returns a boolean. and has little to do with picking info.

I totally understand the need to put it somewhere else, but this is a very specialized function that is only being used by the near interaction feature. The fact is that we didn't need it until now.
We are trying to minimize the impact on large classes like the mesh and the scene by not adding too much "specialized" functionality to it.

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -321,148 +361,172 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
-                if (this._utilityLayerScene) {
-                    utilitySceneNearGrab = this._utilityLayerScene.pickWithRay(controllerData.grabRay, this.nearGrabPredicate);
-                }
-                let originalSceneNearGrab = this._scene.pickWithRay(controllerData.grabRay, this.nearGrabPredicate);
-                pick = accuratePickInfo(originalSceneNearGrab, utilitySceneNearGrab);
-                if (pick && pick.pickedPoint && pick.hit) {
-                    controllerData.nearGrab = true;
+                const nearPick = populateNearInteractionInfo(pickInfo);
+                if (nearPick.hit) {
+                    // Near pick takes precedence over hfover interaction
```
rickfromwork, line 413: >hfover [](http://example.com/codeflow?start=55&length=6)

nit - hover

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -563,22 +642,26 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
                 let mesh = sceneToUse.meshes[meshIndex];
-                if (!predicate(mesh)) {
+                if (!predicate(mesh) || !this._controllerAvailablePredicate(mesh, controllerData.xrController.uniqueId)) {
                     continue;
                 }
-                let result = mesh.intersectsMesh(indexTipMesh, precise);
-                if (result) {
-                    pickingInfo = new PickingInfo();
-                    pickingInfo.hit = result;
+                let sphere = BoundingSphere.CreateFromCenterAndRadius(position, radius);
```
rickfromwork, line 656: Looks like this can be created outside of the for loop

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
```diff
@@ -223,20 +227,30 @@ export class BaseSixDofDragBehavior implements Behavior<Mesh> {
...
                     this._dragging = true;
+
                     virtualMeshesInfo.lastOriginPosition.copyFrom(pointerInfo.pickInfo.ray.origin);
 
                     // Set position and orientation of the controller
                     virtualMeshesInfo.originMesh.position.copyFrom(pointerInfo.pickInfo.ray.origin);
-                    virtualMeshesInfo.originMesh.lookAt(pointerInfo.pickInfo.ray.origin.add(pointerInfo.pickInfo.ray.direction));
+                    if (pointerInfo.pickInfo.originMesh) {
+                        // Near interaction
+                        virtualMeshesInfo.originMesh.lookAt(this._pointerCamera!.globalPosition);
```
rickfromwork, line 238: Does this mean that when dragging objects in XR with near interaction, they'll be oriented to face the user/camera? Or does the virtual mesh not relate to the real mesh in that way?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -321,148 +361,172 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
+
                 let nearInteractionAtOrigin = false;
                 let nearInteraction = nearInteractionInfo && nearInteractionInfo.pickedPoint && nearInteractionInfo.hit;
                 if (nearInteractionInfo?.pickedPoint) {
                     nearInteractionAtOrigin = nearInteractionInfo.pickedPoint.x === 0 && nearInteractionInfo.pickedPoint.y === 0 && nearInteractionInfo.pickedPoint.z === 0;
                 }
                 if (nearInteraction && !nearInteractionAtOrigin) {
-                    result = nearInteractionInfo;
-                    controllerData.nearInteraction = true;
+                    result = nearInteractionInfo!;
```
rickfromwork, line 373: Does the conditional check above not guarantee to the typescript compiler that nearInteractionInfo exists?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -321,148 +361,172 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
-                if ((hoverPickInfo && hoverPickInfo.hit) || (hoverGrabInfo && hoverGrabInfo.hit)) {
-                    hoverRange = true;
+                if (hoverPickInfo && hoverPickInfo.hit) {
                     // turn off far interaction if in the hover range
                     if (this._farInteractionFeature) {
                         this._farInteractionFeature.detach();
                     }
                     pick = populateNearInteractionInfo(hoverPickInfo);
+                    if (pick.hit) {
+                        controllerData.hoverInteraction = true;
```
rickfromwork, line 398: We probably shouldn't disable the far interaction feature if populateNearInteractionInfo gives an invalid pick info, since that would cause a feature detach/attach every frame where it happens.

Actually, does it make more sense to place both the detach and attach logic together down below, as an if/else conditional?

src/Behaviors/Meshes/baseSixDofDragBehavior.ts
CraigFeldspar: `baseSixDofDragBehavior` only moves around virtual meshes related to pointer positions/pick, then the mesh position/orientation is set in the child class `SixDofDragBehavior`. This separation was made to enable other kind of pointer interactions, like 2 pointer interactions, or the plane constrained drag for slate gizmo's handles.

So yes, even though the virtual mesh looks at camera, final mesh orientation is indeed different.

src/XR/features/WebXRNearInteraction.ts
CraigFeldspar: yep it should, but it doesn't, compiler requires the "!" sign. I think the compiler still doesn't decompose these kinds of booleans.

src/XR/features/WebXRNearInteraction.ts
CraigFeldspar: yep, good catch

rickfromwork (approved):
:shipit:

rickfromwork:
Is this ready to merge in?

sebavan:
looking at the comments and such I think so, hence the merge, I ll do a nightly in a bit

PR #10489, opened 2021-06-10
NME: Added a SceneDepth block to access the scene depth buffer

Popov72:
See https://forum.babylonjs.com/t/about-node-material-questions/21498/3


deltakosh (approved):

deltakosh:
This is super neat! thanks a lot!

PR #10491, opened 2021-06-10
Support Base64 Encoded Data URL's generically

CoPrez:
This change lets base64 data urls ([docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs)) be loaded in the same manor as a normal https:// url. The basic idea is that the GLTF, OBJ, STL, and Babylon file loaders don't need to know how to handle a data url, they can just treat it like any other URL and request the contents either in binary form as an ArrayBuffer or in text form. Effectively it can use the same path as loading a remote resource but the data is resolved immediately without performing an XHR.

The biggest caveat is that a data URL may or may not contain mime-type information and it doesn't give any hint as to the extension. That means to reliable load from a data URL the consumer should provide the plugin extension property. Otherwise the default loader (BabylonFileLoader) will likely be used.

This change makes the [ISceneLoaderPluginBase::canDirectLoad](https://github.com/BabylonJS/Babylon.js/blob/d8ccccb9ab7aeed8e43b76c56ba47d9bc7c85385/src/Loading/sceneLoader.ts#L175) property and infrastructure fairly redundant as the data URL can now be directly loaded without the plugin needing to explicitly support it. I didn't want to remove the property though for both simplicity and maintaining support of any third party plugins that might rely on the function for non data url inputs.

I have tested out these changes locally by using the playground and loading different OBJ, STL, and GLB models in their base64 encoded form.


src/Misc/fileTools.ts
```diff
@@ -348,6 +349,23 @@ export class FileTools {
...
+
+                    if (onError) {
+                        onError(new RequestFileError(message, new WebRequest()));
+                    } else {
+                        Logger.Error(message);
+                    }
+                }
+
+                return;
+            }
```
bghgary, line 367: This code probably belongs in LoadFile and not RequestFile. LoadFile, for example, can also handle `file:` protocol. It probably makes the most sense to add `data:` protocol there.

src/Misc/tools.ts
```diff
@@ -912,7 +913,7 @@ export class Tools {
      * @return True if the uri is a base64 string or false otherwise
      */
     public static IsBase64(uri: string): boolean {
-        return uri.length < 5 ? false : uri.substr(0, 5) === "data:";
+        return uri.length < 5 ? false : uri.substr(0, 5) === "data:" && StringTools.EndsWith(uri.split(',')[0], ";base64");
```
bghgary, line 916: This is not efficient since split will create the second part of the string. Probably better to use `indexOf` to get the index of the `,` character and use substr to get the string before that index.

src/Misc/tools.ts
```diff
@@ -932,6 +933,10 @@ export class Tools {
         return bufferView.buffer;
     }
 
+    public static DecodeBase64AsText(uri: string): string {
```
bghgary, line 936: need typedoc here.

src/Misc/fileTools.ts
bghgary: Hmm, just noticed that this won't work because `SceneLoader` is calling requestFile directly. Hmm..

CoPrez: I'm actually a bit confused on the purpose of LoadFile versus RequestFile. LoadFile will pass the request to RequestFile if the url doesn't start with "file:" which means with my current implementation calling `LoadFile("data:base64,...")` will actually work because it will fall through to calling RequestFile.

If I move the logic into LoadFile then I would also need to update the SceneLoader to call LoadFile instead of directly calling ReadFile. SceneLoader would then also need to know that data URLs should be treated different than standard URLs.

I kind of see it the other way around actually. I think LoadFile and RequestFile should be combined to be one function so the consumer doesn't need to know that a "file:" url should be handled differently than a "data:" or "https:" url.

src/Misc/tools.ts
CoPrez: Good call! I've updated the logic so it will only duplicate 7 characters of the string to validate that it's equal to ";base64"

src/Misc/fileTools.ts
```diff
@@ -12,6 +12,7 @@ import { ShaderProcessor } from '../Engines/Processors/shaderProcessor';
 import { ThinEngine } from '../Engines/thinEngine';
 import { EngineStore } from '../Engines/engineStore';
 import { Logger } from './logger';
+import { Tools } from './tools';
```
RaananW, line 15: this causes a circular dependency (tools -> fileTools -> tools)

src/Misc/tools.ts
```diff
@@ -17,6 +17,7 @@ import { InstantiationTools } from "./instantiationTools";
 import { GUID } from "./guid";
 import { IScreenshotSize } from "./interfaces/screenshotSize";
 import { SliceTools } from "./sliceTools";
+import { StringTools } from "./stringTools";
```
RaananW, line 20: This can be removed

src/Misc/tools.ts
```diff
@@ -912,7 +913,18 @@ export class Tools {
      * @return True if the uri is a base64 string or false otherwise
      */
     public static IsBase64(uri: string): boolean {
-        return uri.length < 5 ? false : uri.substr(0, 5) === "data:";
+        // Check that the length of the string is at least as long as the minimum base64 data url length
+        if (uri.length >= "data:;base64,".length)
+        {
+            if (uri.substr(0, 5) === "data:") {
+                let commaIndex = uri.indexOf(",");
+                if (commaIndex > 7 && uri.substr(commaIndex - 7, 7) === ";base64") {
```
RaananW, line 921: nit
```suggestion
                return (commaIndex > 7 && uri.substr(commaIndex - 7, 7) === ";base64");
```

RaananW (changes requested):

src/Misc/tools.ts
CoPrez: Indeed it can after the refactor. Removing it now.

src/Misc/fileTools.ts
bghgary: RequestFile has additional callbacks specific to http requests. This was originally just LoadFile. I split this to support http range requests in the glTF loader.

> If I move the logic into LoadFile then I would also need to update the SceneLoader to call LoadFile instead of directly calling ReadFile. SceneLoader would then also need to know that data URLs should be treated different than standard URLs.

I have to try think about this more and maybe try some things.

src/Misc/fileTools.ts
CoPrez: I moved the data url logic into fileTools so that it doesn't create a circular import. I then redeclared the functions in tools to maintain backwards compatibility.

@RaananW, I didn't update Babylon.JS's internal references to `Tools.IsBase64` and `Tools.DecodeBase64` to use the new declarations in FileTools. I wasn't sure what the standard practice was there. Should I change all of those references in this PR or keep it minimal?

RaananW: That's perfectly fine. As long as the public functions are all there you can leave it this way. Would be great to just have a deprecated comment for future reference

RaananW (commented):
Waiting on @bghgary

src/Misc/fileTools.ts
CoPrez: @deprecated docs have been added. Good call 👍

src/Misc/fileTools.ts
```diff
@@ -493,6 +510,50 @@ export class FileTools {
...
+    }
+
+    /**
+     * Decode the given base64 uri into a UTF-8 encoded string.
+     * @param uri The uri to decode
+     * @return The decoded base64 data.
+     */
+    public static DecodeBase64UrlAsText(uri: string): string {
+        return atob(uri.split(",")[1]);
+    }
```
bghgary, line 556: fileTools? it should be stringTools. we already have other base64 things in stringTools, but there is a problem with moving it to stringTools for treeshaking.

CoPrez: I didn't realize StringTools had logic for decoding and encoding Base64 strings. I'm not sure what the advantage of that is versus `atob` and `btoa`. I could see using StringTools.Decode inside of these functions, but it feels kind of odd to add logic in StringTools for understanding the URL structure. I am fine with moving the implementation wherever you think is appropriate though.

bghgary: atob and btoa only deals with strings. It takes and returns strings. They don't work for binary data.

bghgary: > I could see using StringTools.Decode inside of these functions, but it feels kind of odd to add logic in StringTools for understanding the URL structure.

Yeah, that would be weird. I think the best would be if the fileTools function parsed the url and the stringTools decode. Similar for encode.

CoPrez: Moved the string manipulation of Base64 data into StringTools now with the Data URL manipulation still being in FileTools.

src/Loading/sceneLoader.ts
```diff
@@ -479,7 +479,7 @@ export class SceneLoader {
 
         SceneLoader.OnPluginActivatedObservable.notifyObservers(plugin);
 
-        if (directLoad) {
+        if (directLoad && plugin.canDirectLoad && plugin.canDirectLoad(fileInfo.url)) {
```
CoPrez, line 482: This change is currently breaking the scenario (and unit test) for OBJ files being passed in as raw text by specifying the data as "data:<obj_file_content>"

I didn't realize that was a supported scenario because the OBJ plugin does not declare the `canDirectLoad` or `directLoad` functions. In the previous implementation if the data was a string that started with "data:" it would fallback stripping "data:" from the start of the string and passing the rest of the value to the plugins default loader method.

src/Misc/stringTools.ts
```diff
@@ -77,6 +77,32 @@ export class StringTools {
...
     }
 
+    /**
+     * Converts a given base64 string as an ASCII encoded stream of data
+     * @param base64Data The base64 encoded string to decode
+     * @returns Decoded ASCII string
+     */
+    public static DecodeBase64ToString(base64Data: string) : string {
+        return atob(base64Data);
+    }
```
bghgary, line 87: I know this is trying to make the code consistent, but extra functions are not free in JS. It's probably better to call atob directly.

RaananW: I would avoid using atob and btoa altogether. They don't support Unicode (try running `btoa('✓äßü')`. `atob` would (not) work the same). Also atob doesn't exist on node.js (was actually just added in node 16, but uses `ISO-8859-1` instead of `UTF-8`)). So if this code is running on the server (using a null engine, for example), this will either fail or return an unwanted result in certain cases. 
We are using atob in code, and it does work, as the strings being parsed are URIs, which are considered to be ASCII-Only. but this is no longer the absolute (web)truth. Some servers and DNSs accept URLs with unicode characters. 

There are different and safer ways of encoding/decoding base64 (https://stackoverflow.com/a/30106551)

Also - It doesn't have to be a member of the StringUtils (i.e. doesn't have to be a public static member). It can be exported at a height level to avoid any complications with an object being too big.

bghgary: That's probably the right thing though we have been using atob for a long while. `atob` is also being polyfilled in Babylon Native. If we change this, we need to update Babylon Native also (maybe don't even need it anymore).

bghgary:
I converted this PR to draft since I'm experimenting a bit.

src/Misc/stringTools.ts
CoPrez: I had moved the decoding into StringTools per this earlier comment:
https://github.com/BabylonJS/Babylon.js/pull/10491#discussion_r650321738

atob certainly is not the best decoder, but it was what was already in place for Babylon.JS and my goal was to minimize the impact of these changes. I'm happy to swap it out with one of the implementations from that stack overflow post. I would say not to use the MDN suggestion in this case as it only works if you're doing both the encoding and decoding.

RaananW: To summarize my entire (way too long) comment - the approach of adding some abstracted method to decode strings is  great, because we will be able to move the entire atob usage in the framework to some other implementation.

src/Misc/fileTools.ts
```diff
@@ -493,6 +518,42 @@ export class FileTools {
...
     }
+
+    /**
+     * Test if the given uri is a base64 string
+     * @param uri The uri to test
+     * @return True if the uri is a base64 string or false otherwise
+     */
+    public static IsBase64DataUrl(uri: string): boolean {
+        // Check that the length of the string is at least as long as the minimum base64 data url length
+        if (uri.length >= "data:;base64,".length)
```
bghgary, line 529: Unfortunately, this isn't cheap in JS.

```suggestion
        if (uri.length >= 13) // "data:;base64,".length
```

RaananW: As much as I do agree with the change suggested, I am not sure it is as expensive as you think:

image

src/Loading/sceneLoader.ts
CoPrez: I've resolved the regression by updating the conditional to still use the directLoad code path if the given data URI is not a valid base64 data uri. That retains the pre-existing support for non standard "data:" urls while still making it so loaders can remain oblivious to base64 data urls.

The one gotcha I found was that we still need to use the glTF directLoad function when it registers as a handler for base64 URI's. That's because it's only in it's directLoad function that it detects if it's a gltf or glb file.

src/Misc/fileTools.ts
CoPrez: @bghgary has now reworked the change so this functionality has been moved into LoadFile rather than RequestFile. With that change SceneLoader just uses LoadFile and doesn't directly call RequestFile anymore.

src/Misc/fileTools.ts
bghgary: Makes sense. I'm just saying it's not free like it is in native code.

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
```diff
@@ -731,4 +731,94 @@ describe('Babylon Scene Loader', function() {
             return BABYLON.SceneLoader.LoadAsync("file:///", "Box.gltf").then(resetPreprocessUrl, resetPreprocessUrl);
         });
     });
-});
+
+    describe('#DirectLoad', () => {
+        it('should load a base64 encoded obj with no mime type', () => {
+            var scene = new BABYLON.Scene(subject);
+            return BABYLON.SceneLoader.LoadAssetContainerAsync('data:;base64,' + objBase64, undefined, scene, () => { }, ".obj").then((container) => {
```
bghgary, line 738: Is there a reason why sometimes we use asset container and sometimes not in these tests?

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
```diff
@@ -731,4 +731,94 @@ describe('Babylon Scene Loader', function() {
...
+            var scene = new BABYLON.Scene(subject);
+            return BABYLON.SceneLoader.LoadAssetContainerAsync('data:;base64,' + objBase64, undefined, scene, () => { }, ".obj").then((container) => {
+                expect(container.meshes.length).to.eq(1);
+                expect(container.meshes[0].getTotalVertices()).to.eq(4);
+            });
+        });
+
+        it('should load a base64 encoded obj with a valid mime type', () => {
+            var scene = new BABYLON.Scene(subject);
+            return BABYLON.SceneLoader.LoadAssetContainerAsync('data:model/obj;base64,' + objBase64, undefined, scene, () => { }, ".obj").then((container) => {
```
bghgary, line 746: nit: everywhere
```suggestion
            return BABYLON.SceneLoader.LoadAssetContainerAsync('data:model/obj;base64,' + objBase64, undefined, scene, undefined, ".obj").then((container) => {

```

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
```diff
@@ -731,4 +731,94 @@ describe('Babylon Scene Loader', function() {
             return BABYLON.SceneLoader.LoadAsync("file:///", "Box.gltf").then(resetPreprocessUrl, resetPreprocessUrl);
         });
     });
-});
+
+    describe('#DirectLoad', () => {
+        it('should load a base64 encoded obj with no mime type', () => {
+            var scene = new BABYLON.Scene(subject);
```
bghgary, line 737: nit: everywhere
```suggestion
            const scene = new BABYLON.Scene(subject);
```

bghgary (approved):

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
CoPrez: The change has been made

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
CoPrez: Updated now

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
CoPrez: That was just because I copy and pasted lines from different locations. I've updated them all to consistently use `ImportMeshAsync` now.

CoPrez:
Reactivating the PR now. @bghgary has integrated his refactors to the SceneLoader and glTFFileLoader into the PR.

Pinging @RaananW so that you could have a chance to give this a once over as it's changed a fair bit since you approved it.

src/Loading/sceneLoader.ts
```diff
@@ -525,56 +514,46 @@ export class SceneLoader {
...
-                };
+        const manifestChecked = () => {
+            if (pluginDisposed) {
+                return;
+            }
 
-                const errorCallback = (error: RequestFileError) => {
-                    onError(error.message, error);
-                };
+            const successCallback = (data: any, responseURL?: string) => {
```
RaananW, line 522: as this function is exactly the dataCallback function, is there a need to define it here?

src/Misc/fileTools.ts
```diff
@@ -493,8 +518,44 @@ export class FileTools {
...
+    /**
+     * Test if the given uri is a base64 string
+     * @param uri The uri to test
+     * @return True if the uri is a base64 string or false otherwise
+     */
+    public static IsBase64DataUrl(uri: string): boolean {
+        // Check that the length of the string is at least as long as the minimum base64 data url length
+        if (uri.length >= 13) // "data:;base64,".length
+        {
+            if (uri.substr(0, 5) === "data:") {
```
RaananW, line 531: Without creating an extra string:
```suggestion
            if (uir.indexOf('data:') === 0) {
```

src/Misc/fileTools.ts
```diff
@@ -493,8 +518,44 @@ export class FileTools {
...
+                let commaIndex = uri.indexOf(",");
+                return commaIndex > 7 && uri.substr(commaIndex - 7, 7) === ";base64";
```
RaananW, line 533: you can use indexOf here as well (instead of creating a new string)

tests/unit/babylon/src/Loading/babylon.sceneLoader.tests.ts
```diff
@@ -731,4 +731,94 @@ describe('Babylon Scene Loader', function() {
             return BABYLON.SceneLoader.LoadAsync("file:///", "Box.gltf").then(resetPreprocessUrl, resetPreprocessUrl);
         });
     });
-});
+
+    describe('#DirectLoad', () => {
+        it('should load a base64 encoded obj with no mime type', () => {
+            const scene = new BABYLON.Scene(subject);
+            return BABYLON.SceneLoader.ImportMeshAsync('', 'data:;base64,' + objBase64, undefined, scene, undefined, ".obj").then((result) => {
```
RaananW, line 738: very nit-picky, only if you are making changes to the PR:

```suggestion
            return BABYLON.SceneLoader.ImportMeshAsync('', `data:;base64,${objBase64}`, undefined, scene, undefined, ".obj").then((result) => {

```

This is the same for the entire file (9 more instances)

src/Misc/fileTools.ts
bghgary: This is true, but it will search the entire string if `data:` is not present. Probably not a big deal since big strings probably start with `data:`

src/Loading/sceneLoader.ts
bghgary: Good point. This isn't necessary anymore. We can just pass dataCallback directly.

src/Misc/fileTools.ts
RaananW: so use startsWith? :-)
Or use a regexp. avoid creating the extra object

src/Misc/fileTools.ts
CoPrez: I've switched the base64 URL detection out to using regex. The previous implementation had some false positives for non base64 url's that contained base64 data url's in them. Such as a raw GLTF file.

src/Misc/fileTools.ts
bghgary: Yup. We did that below. Using startsWith is a good idea.

PR #10494, opened 2021-06-11
Add pre-defined Audio Context and Destination options to AudioEngine.

svanderbeck11:
Added the ability for the audio engine to take an existing Audio Context and a specific Audio Destination node.  Discussed on forum https://forum.babylonjs.com/t/custom-audio-context/21080.

### DIFF
--- a/src/Audio/audioEngine.ts
+++ b/src/Audio/audioEngine.ts
@@ -7,20 +7,8 @@ import { Engine } from "../Engines/engine";
 import { IAudioEngine } from './Interfaces/IAudioEngine';
 import { DomManagement } from "../Misc/domManagement";
 
-export interface AudioEngineOptions {
-     /**
-     * Specifies an existing Audio Context for the audio engine
-     */
-      audioContext?: AudioContext;
-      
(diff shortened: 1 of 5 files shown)


src/Audio/audioEngine.ts
```diff
@@ -7,8 +7,20 @@ import { Engine } from "../Engines/engine";
 import { IAudioEngine } from './Interfaces/IAudioEngine';
 import { DomManagement } from "../Misc/domManagement";
 
+export interface AudioEngineOptions {
```
deltakosh, line 10: do you mind moving it to a separate file

deltakosh:
A tiny change and we are good (plus fixing the checks)

src/Audio/audioEngine.ts
svanderbeck11: Split that off into its own class and renamed it IAudioEngineOptions.

svanderbeck11:
Should be all updated with checks complete.

deltakosh:
Thanks it rocks!

Issue #10369, opened 2021-05-14
[GE] Advance Texture Property Tab


msDestiny14:
![PT](https://user-images.githubusercontent.com/67929198/118323726-e771b080-b4b5-11eb-8741-098f08c7106c.JPG)

PR #10498, opened 2021-06-11
[GE] Advance Texture Property Tab 

msDestiny14:
adt

https://github.com/BabylonJS/Babylon.js/issues/10369

Notes: 
Shared Components will simply have an optional "icon" property. The css will then be altered in the local version associated with the gui editor.


RaananW (commented):
I see a lof of css value changes. Are these all according to the design? Are there some form of design guidelines to make it easier to orient around these changes?

I see 35px height, 25px, 32px. There is usually some sort of a union between heights of different components.

RaananW (approved):

PR #10500, opened 2021-06-14
Add an option to use a NDC range between 0..1 instead of -1..1

Popov72:
WebGPU is using a NDC range of 0..1.

This PR adds a property to the engine that allows WebGPU to use the native range and get rid of some conversions.

It also allows to use the reverse depth buffer option to fix some precision problems when geometries are too close. See https://playground.babylonjs.com/#67JFXI#5 => in WebGPU, the two planes won't flicker.

[EDIT] Also synced with WebGPU spec.


src/Engines/WebGPU/webgpuShaderProcessors.ts
```diff
@@ -366,7 +367,11 @@ export class WebGPUShaderProcessor implements IShaderProcessor {
         if (!isFragment) {
             const lastClosingCurly = code.lastIndexOf("}");
             code = code.substring(0, lastClosingCurly);
-            code += "gl_Position.y *= -1.;\ngl_Position.z = (gl_Position.z + gl_Position.w) / 2.0; }";
+            code += "gl_Position.y *= -1.;\n";
+            if (!engine.isNDCHalfZRange) {
```
RaananW, line 371: maybe passing the flag instead of the engine will make it a bit abstracter? Just a thought.

Popov72: The `engine` is already a parameter of `postProcessor` (see the [interface](https://github.com/BabylonJS/Babylon.js/blob/master/src/Engines/Processors/iShaderProcessor.ts#L15)), I simply didn't need it yet in the function.

RaananW (approved):

RaananW (approved):

Popov72:
The WebGL1 validation test error is probably a transient error, it is a 3mn timeout on the "DDS" test.

src/Engines/WebGPU/Extensions/engine.dynamicTexture.ts
```diff
@@ -16,10 +16,19 @@ WebGPUEngine.prototype.updateDynamicTexture = function(texture: Nullable<Interna
...
+    this.createImageBitmap(canvas).then((bitmap) => {
+        this._textureHelper.updateTexture(bitmap, gpuTextureWrapper.underlyingResource!, width, height, texture.depth, gpuTextureWrapper.format, 0, 0, invertY, premulAlpha, 0, 0, this._uploadEncoder);
+        if (texture.generateMipMaps) {
+            this._generateMipmaps(texture, this._uploadEncoder);
+        }
+
+        texture.isReady = true;
+    });
+
+    /*this._textureHelper.updateTexture(canvas, texture, width, height, texture.depth, gpuTextureWrapper.format, 0, 0, invertY, premulAlpha, 0, 0, this._uploadEncoder);
```
sebavan, line 28: Only one question, I wonder why keeping the comment ???

src/Engines/WebGPU/webgpuTextureHelper.ts
```diff
@@ -1084,7 +1085,18 @@ export class WebGPUTextureHelper {
...
+            if (invertY || premultiplyAlpha) {
+                const engine = EngineStore.LastCreatedEngine;
+                engine && engine.createImageBitmap(imageBitmap, { imageOrientation: invertY ? "flipY" : "none", premultiplyAlpha: premultiplyAlpha ? "premultiply" : "none" }).then((imageBitmap) => {
+                    this._device.queue.copyImageBitmapToTexture({ imageBitmap }, textureCopyView, textureExtent);
+                });
+            } else {
+                this._device.queue.copyImageBitmapToTexture({ imageBitmap }, textureCopyView, textureExtent);
+            }
+
+            /*imageBitmap = imageBitmap as (ImageBitmap | HTMLCanvasElement | OffscreenCanvas);
```
sebavan, line 1099: same :-)

sebavan (commented):
Also I am wondering why some of the render changed as well as the PGs ?

src/Engines/WebGPU/Extensions/engine.dynamicTexture.ts
Popov72: Because as soon as `copyExternalImageToTexture` is fixed in Chrome I will remove the code above line 27 and will uncomment the code starting at line 28 :)

src/Engines/WebGPU/webgpuTextureHelper.ts
Popov72: Same answer than above

Popov72:
> Also I am wondering why some of the render changed as well as the PGs ?

The two PGs that are changed are because the `camera.minZ` was very low (0.01), and when enabling the reverse depth buffer mode there were some differences because of that. I set `minZ` to 0.1 and the same picture works both for the normal/reverse depth buffer modes.

sebavan (approved):

sebavan:
I ll merge after my npm release :-)

PR #10506, opened 2021-06-15
fix i.initFunction is not a function pg error

RaananW:
Syntax errors inside a script added to the DOM are not being caught using a simple try/catch. The only way to catch those is using the window `error` event. 

This PR adds error handling in case a syntax error happened while parsing the evaluated code.

See https://forum.babylonjs.com/t/i-initfunction-is-not-a-function/21233


Popov72 (approved):
LGTM!

deltakosh (approved):

PR #10509, opened 2021-06-16
LOD screen coverage proposal

CraigFeldspar:
Implements a screen coverage LOD loading and selection to support https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/MSFT_lod


src/Cameras/camera.ts
```diff
@@ -488,16 +513,12 @@ export class Camera extends Node {
             return false;
         }
 
-        return this._cache.position.equals(this.position)
-            && this._cache.upVector.equals(this.upVector)
-            && this.isSynchronizedWithParent();
+        return this._cache.position.equals(this.position) && this._cache.upVector.equals(this.upVector) && this.isSynchronizedWithParent();
```
deltakosh, line 516: please don't change that :)

src/Cameras/camera.ts
```diff
@@ -506,17 +527,15 @@ export class Camera extends Node {
...
-                && this._cache.aspectRatio === engine.getAspectRatio(this);
-        }
-        else {
-            check = this._cache.orthoLeft === this.orthoLeft
-                && this._cache.orthoRight === this.orthoRight
-                && this._cache.orthoBottom === this.orthoBottom
-                && this._cache.orthoTop === this.orthoTop
-                && this._cache.renderWidth === engine.getRenderWidth()
-                && this._cache.renderHeight === engine.getRenderHeight();
+            check = this._cache.fov === this.fov && this._cache.fovMode === this.fovMode && this._cache.aspectRatio === engine.getAspectRatio(this);
```
deltakosh, line 530: same

CraigFeldspar: this is done automatically by prettier when linting, maybe we should change the rules in .prettierrc so it doesn't mess that much with indentation ?

deltakosh: Totally (adding our pretty man :) @RaananW )

Popov72: Or run the whole repo with the current rules to make it up to date and make everyone use prettier/those rules?

RaananW: I will pass over the entire repo very soon. probably next week. as it seems to be disturbing a lot

RaananW: Also, notice it didn't do it on the next one. it is related to the location of the "&&" characters. unifying this is not a bad thing.

src/Meshes/mesh.ts
```diff
@@ -784,12 +796,13 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
     }
 
     private _sortLODLevels(): void {
+        const useScreenCoverage = this._internalMeshDataInfo._useLODScreenCoverage ? -1 : 1;
```
Popov72, line 799: I don't find the name `useScreenCoverage` very clear given how it is used... Maybe something independent from the LOD type would be better (like `resultForLessComparison`)?

src/Meshes/mesh.ts
```diff
@@ -799,17 +812,17 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
     /**
      * Add a mesh as LOD level triggered at the given distance.
      * @see https://doc.babylonjs.com/how_to/how_to_use_lod
-     * @param distance The distance from the center of the object to show this level
+     * @param distanceOrScreenCoverage Either distance from the center of the object to show this level or the screen coverage if `useScreenCoverage` is set to `true`
```
Popov72, line 815: nit: May be add `(value between 0 and 1)` to make clear the screen coverage value is a percentage

src/Meshes/meshLODLevel.ts
```diff
@@ -8,12 +8,12 @@ import { Nullable } from '../types';
...
      * Creates a new LOD level
-     * @param distance defines the distance where this level should star being displayed
+     * @param distanceOrScreenCoverage defines either the distance or the screen coverage where this level should start being displayed
      * @param mesh defines the mesh to use to render this level
      */
     constructor(
-        /** Defines the distance where this level should start being displayed */
-        public distance: number,
+        /** Either distance from the center of the object to show this level or the screen coverage if `useLODScreenCoverage` is set to `true` on the mesh*/
+        public distanceOrScreenCoverage: number,
```
Popov72, line 16: You can't really rename the property because it would be a breaking change...

Or maybe we can accept it (@deltakosh / @sebavan ?) as I'm not sure anyone should directly work with the `MeshLODLevel` class, in which case the **Breaking changes** section of the what's new file should be updated.

Popov72 (approved):
LGTM save for some small comments!

src/Meshes/mesh.ts
deltakosh: it is ok as this is not binding for either TS or JS

src/Meshes/meshLODLevel.ts
deltakosh: It is ok as this is not binding for either TS or JS :)

sebavan:
I ll do my pass after the first batch of fixes :-)

src/Meshes/mesh.ts
CraigFeldspar: I went with `sortingOrderFactor`

RaananW (approved):

src/Meshes/mesh.ts
```diff
@@ -884,8 +898,18 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         }
 
         var distanceToCamera = bSphere.centerWorld.subtract(camera.globalPosition).length();
+        const useScreenCoverage = internalDataInfo._useLODScreenCoverage;
+        let ratio = 1;
+
+        if (useScreenCoverage) {
```
sebavan, line 904: could we here keep ratio in a variable compareValue. and in the else part make it = -distanceToCamera ?

So that in the 2 following comparison we do not need the extra if cases ???

sebavan (changes requested):
Just a minor tweak @CraigFeldspar and I ll merge it :-)

PR #10514, opened 2021-06-17
Guarding against undefined document usage in 3D controls

rickfromwork:
Prevents crashes in remaining 3D GUI components by placing 2D control usage behind 'document' checks. Some functions are unguarded, like the HolographicButton's tooltipText setter, but since that must be manually called, it doesn't pose nearly as large of a risk.


RaananW (approved):

RaananW:
There is a linting issue. Merging after it's fixed

PR #10515, opened 2021-06-17
fix crash in sixDofDragBehavior.ts

CedricGuillemet:
Follow up on https://forum.babylonjs.com/t/not-able-to-access-gizmos-when-2-cameras-are-activated/21680
If node/mesh is not attached to the behavior, a crash happens when it's detached from gizmomanager.

### DIFF
--- a/src/Behaviors/Meshes/sixDofDragBehavior.ts
+++ b/src/Behaviors/Meshes/sixDofDragBehavior.ts
@@ -233,6 +233,8 @@ export class SixDofDragBehavior extends BaseSixDofDragBehavior {
             this._ownerNode.getScene().onBeforeRenderObservable.remove(this._sceneRenderObserver);
         }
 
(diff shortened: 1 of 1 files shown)


src/Behaviors/Meshes/sixDofDragBehavior.ts
```diff
@@ -233,6 +233,8 @@ export class SixDofDragBehavior extends BaseSixDofDragBehavior {
             this._ownerNode.getScene().onBeforeRenderObservable.remove(this._sceneRenderObserver);
         }
 
-        this._virtualTransformNode.dispose();
+        if (this._virtualTransformNode) {
```
Popov72, line 236: nit: `this._virtualTransformNode?.dispose();`

Popov72 (approved):

RaananW (approved):

PR #10516, opened 2021-06-17
Fix colors refresh for gizmos

CedricGuillemet:
followup https://forum.babylonjs.com/t/major-gizmo-bug/21724
update gizmo colors when attached mesh/node changes without pointer events to refresh it.

### DIFF
--- a/src/Gizmos/axisDragGizmo.ts
+++ b/src/Gizmos/axisDragGizmo.ts
@@ -167,6 +167,7 @@ export class AxisDragGizmo extends Gizmo {
             dragBehavior: this.dragBehavior
         };
         this._parent?.addToAxisCache(collider as Mesh, cache);
+        this._axisCache = cache;
 
         this._pointerObserver = gizmoLayer.utilityLayerScene.onPointerObservable.add((pointerInfo) => {
             if (this._customMeshSet) {
@@ -187,6 +188,7 @@ export class AxisDragGizmo extends Gizmo {
     protected _attachedNodeChanged(value: Nullable<Node>) {
         if (this.dragBehavior) {
             this.dragBehavior.enabled = value ? true : false;
+            this._refreshForDragBehavior(this.dragBehavior.enabled, this._coloredMaterial, this._disableMaterial);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/Gizmos/gizmo.ts
```diff
@@ -321,6 +322,25 @@ export class Gizmo implements IDisposable {
         }
     }
 
+    /**
+     * refresh gizmo mesh material depending on state of dragBehavior
+     * @param enabledDragBehavior drag behavior is enabled or not
+     * @param enableddMaterial material to apply if drag behavior is enabled
+     * @param disableMaterial material to apply if drag behavior is disenabled
+     */
+    protected _refreshForDragBehavior(enabledDragBehavior: boolean, enabledMaterial: StandardMaterial, disableMaterial: StandardMaterial) {
```
RaananW, line 331: if `enabledDragBehavior` is only used to check which material is being used, why not only pass the material that is going to be used in the first place?

mrlooi: I believe this is used to switch the materials on enabled vs not enabled. if enabled=true, colored material. If not, gray material

RaananW: The three variables passed to the function are only being used by themselves. a boolean and two materials are passed to get a material that could have been passed on its own in the first place.

mrlooi: yep true, would probably be cleaner to just pass in the material itself and maybe rename to setMaterial or sth

src/Gizmos/axisDragGizmo.ts
```diff
@@ -188,7 +188,7 @@ export class AxisDragGizmo extends Gizmo {
     protected _attachedNodeChanged(value: Nullable<Node>) {
         if (this.dragBehavior) {
             this.dragBehavior.enabled = value ? true : false;
-            this._refreshForDragBehavior(this.dragBehavior.enabled, this._coloredMaterial, this._disableMaterial);
+            this._setGizmoMeshMaterial(this.dragBehavior.enabled ? this._coloredMaterial : this._disableMaterial);
```
mrlooi, line 191: LGTM!

Btw, just wondering (sort of a mini feature request), do you think it would be good if dragBehavior.enabled did not update based on attachedNodeChange? Just thinking that a developer might want to control it and not worry about it changing if a different node is attached

mrlooi: I can see myself setting dragBehavior.enabled = true/false manually myself and the material will update based on that, seems like quite a useful feature

RaananW (approved):

src/Gizmos/axisDragGizmo.ts
CedricGuillemet: I will do it as a second PR.

Popov72 (approved):
LGTM!

PR #10520, opened 2021-06-18
Observable for PointerDragBehavior enable state

CedricGuillemet:
Following https://github.com/BabylonJS/Babylon.js/pull/10516
Refactor the code a little and and observable pour pointerDragBehavior's enable to dynamicaly change gizmo color based on PointerDragBehavior state.

ping @mrlooi

### DIFF
--- a/src/Gizmos/axisDragGizmo.ts
+++ b/src/Gizmos/axisDragGizmo.ts
@@ -5,7 +5,6 @@ import { Vector3 } from "../Maths/math.vector";
 import { TransformNode } from "../Meshes/transformNode";
 import { Node } from "../node";
 import { Mesh } from "../Meshes/mesh";
-import { LinesMesh } from "../Meshes/linesMesh";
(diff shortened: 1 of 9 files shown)


src/Gizmos/axisDragGizmo.ts
```diff
@@ -176,19 +174,17 @@ export class AxisDragGizmo extends Gizmo {
...
             }
         });
+
+        this.dragBehavior.onEnabledObservable.add((newState) => {
+            this._setGizmoMeshMaterial(cache.gizmoMeshes, newState ? cache.material : cache.disableMaterial);
+        });
     }
     protected _attachedNodeChanged(value: Nullable<Node>) {
         if (this.dragBehavior) {
             this.dragBehavior.enabled = value ? true : false;
```
mrlooi, line 187: Probably a design decision question: would we still need _attachedNodeChanged?
If enabled is something that the developer controls, maybe we no longer need it to be toggled on/off depending on if it's attached to a node or not?

PR #10530, opened 2021-06-21
Projection plane tilt

CraigFeldspar:
+ Adding a `projectionPlaneTilt` property to `Camera.ts`, to tilt the projection plane about the X axis
+ Adding an `applyVerticalCorrection` method to `Camera.ts` to automatically tilt the projection plane so it compensates the camera pitch, and vertical lines appear vertical on the screen

I'll post on the forum a FYI detail of the maths


src/Maths/math.vector.ts
```diff
@@ -5831,10 +5832,11 @@ export class Matrix {
         let b = 2.0 * n / height;
         let c = (f + n) / (f - n);
         let d = -2.0 * f * n / (f - n);
+        let rot = Math.tan(projectionPlaneTilt);
```
Popov72, line 5835: What if `projectionPlaneTilt` = Math.PI/2 + n*Math.PI?

CraigFeldspar: When we approch Math.PI/2, things get really distorted so it shouldn't really be used, this is meant for small angles close to 0. 

But in case it would carry `NaN` somewhere, I'll put a safeguard

CraigFeldspar:
Playground : http://localhost:1338/Playground/index-local.html#L20FJ4#15

sebavan (approved):

src/Cameras/camera.ts
```diff
@@ -404,6 +412,15 @@ export class Camera extends Node {
         return ret;
     }
 
+    /**
+     * Automatically tilts the projection plane, using `projectionPlaneTilt`, to correct the perspective effect on vertical lines.
+     */
+    public applyVerticalCorrection() {
+        const rot = this.absoluteRotation.toEulerAngles();
+
+        this.projectionPlaneTilt = this._scene.useRightHandedSystem ? -rot.x : rot.x;
```
RaananW, line 421: Just want to be sure - do we need to flip rotation.x on right handed systems?

CraigFeldspar: `toEulerAngles` adds a PI rotation about the Y axis in right handed systems, and since this rotation is applied before the X rotation, the X axis is flipped

sebavan: seems legit :-) but I stuggle so much with this I am always impressed by the ones who know... A bit like right and the other right...

RaananW (approved):

PR #10533, opened 2021-06-21
[GE] Adding icons to line components

msDestiny14:


RaananW:
There is an issue with duplicate identifiers in 2 or 3 components.

msDestiny14:
Looking into it!

PR #10535, opened 2021-06-21
Fix LastCreateScene value when deleting scenes

Popov72:
I'm not sure it is a fix, maybe it is expected that `LastCreatedScene` is null when we delete the last created scene, even if there are still some other scenes alive...

However, personally I would expect to have the "last scene not disposed" in `LastCreatedScene`, so let's tell what you think.

### DIFF
--- a/src/scene.ts
+++ b/src/scene.ts
@@ -4481,6 +4481,14 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         this.metadata = null;
 
         if (EngineStore._LastCreatedScene === this) {
+            if (this._engine.scenes.length > 1) {
+                EngineStore._LastCreatedScene = this._engine.scenes[this._engine.scenes.length - 2];
+            } else {
+                EngineStore._LastCreatedScene = null;
+            }
+        }
+
+        if (EngineStore._LastCreatedScene === this) {
             EngineStore._LastCreatedScene = null;
         }
 
--- a/src/scene.ts
+++ b/src/scene.ts
@@ -4488,10 +4488,6 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
             }
         }
 
-        if (EngineStore._LastCreatedScene === this) {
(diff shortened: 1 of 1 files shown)


src/scene.ts
```diff
@@ -4480,6 +4480,14 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         this.afterRender = null;
         this.metadata = null;
 
+        if (EngineStore._LastCreatedScene === this) {
```
deltakosh, line 4483: OVerall I'm good with this change but you need to remove the next block of code (duplicate)

Popov72: I already did it before you comment ;)

src/scene.ts
```diff
@@ -4481,7 +4481,11 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         this.metadata = null;
 
         if (EngineStore._LastCreatedScene === this) {
-            EngineStore._LastCreatedScene = null;
+            if (this._engine.scenes.length > 1) {
```
RaananW, line 4484: Wouldn't this entire code fit better after removing the scene from `engine.scenes` in line 485?

sebavan (approved):

Popov72:
The error in **Module Tests** is:
image

RaananW (approved):
Nit-picky - the else is actually not needed. Simply set it to be the last element of the array. If there isn't any, it'll be undefined negative index will return undefined as well)

Popov72:
Indeed but `_LastCreatedScene` is not declared as accepting to be `undefined`:
```javascript
public static _LastCreatedScene: Nullable<Scene> = null;
```
I find it strange that:
```javascript
EngineStore._LastCreatedScene = undefined;
```
is (rightly) not allowed by Typescript but that this code does not raise any warning/error:
```javascript
EngineStore._LastCreatedScene = this._engine.scenes[this._engine.scenes.length - 1];
```

PR #10538, opened 2021-06-21
Add missing UVs to glTF draco loader extension

bghgary:
See https://forum.babylonjs.com/t/multiple-uv-sets-getting-lost-with-draco-compression/21830


sebavan:
Could we add a test for it ?

sebavan (approved):

bghgary:
> Could we add a test for it ?

We can, though I'm not sure how useful that will be.

PR #10539, opened 2021-06-22
Fix bug in IsBase64 check

bghgary:


src/Misc/fileTools.ts
```diff
@@ -13,6 +13,8 @@ import { ThinEngine } from '../Engines/thinEngine';
 import { EngineStore } from '../Engines/engineStore';
 import { Logger } from './logger';
 
+const base64DataUrlRegEx = new RegExp(/^data:([^,]+\/[^,]+)?;base64,/);
```
CoPrez, line 16: I believe you still want the case invariant flag

```suggestion
const base64DataUrlRegEx = new RegExp(/^data:([^,]+\/[^,]+)?;base64,/i);
```

bghgary: Oops, forgot to push. I already fixed it :)

CoPrez (approved):

PR #10540, opened 2021-06-22
[GE] Moving art

msDestiny14:


PatrickRyanMS:
Some of these icons may not be unique to the GUI Editor but also reused in the all-up inspector redesign. Is moving them out of the shared components directory going to cause any future problems?

msDestiny14:
You know what I think you're right!

PR #10542, opened 2021-06-22
NME: Fix potential crash when serializing a frame

Popov72:
See https://forum.babylonjs.com/t/node-material-editor-wont-allow-save/21833

It crashes because when serializing the "Displacement" frame, `this._exposedOutPorts` is an array with 7 elements, but the 3rd is `undefined`. So, the fix will avoid the crash, but I don't know if this undefined element is expected in the first place or not (it seems reloading the material after it has been saved is ok, though...). Pinging @msDestiny14.


RaananW (approved):
Validation tests fail on base64 decoding issue. Unrelated to this PR for sure.

sebavan (approved):

PR #10548, opened 2021-06-22
[GE] Copying property line files

msDestiny14:
This is the initial move. Next PR will have edits to these files. Discussed with @RaananW yesterday.


RaananW (approved):
Before merging - @msDestiny14 - please make sure you only move components that cannot be changed in `shared` (like the checkbox or colorPicker components, maybe?). If that's the case, please merge :-)

msDestiny14:
All of them reference CommonControlPropertyGridComponent which is unique to GUI Editor at the moment. Plus they will all probably get structure change to remove the LineContainerComponent.

PR #10549, opened 2021-06-22
Base64 loading fixes

bghgary:
- Fix invalid base64 directLoad from glTFloader
- Fix SceneLoader usage in unit tests


sebavan (approved):

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -2132,13 +2132,7 @@ export class GLTFLoader implements IGLTFLoader {
...
-        if (Tools.IsBase64(uri)) {
-            const data = new Uint8Array(Tools.DecodeBase64(uri));
-            this.log(`${context}: Decoded ${uri.substr(0, 64)}... (${data.length} bytes)`);
-            return Promise.resolve(data);
-        }
-
-        this.log(`${context}: Loading ${uri}`);
+        this.log(`${context}: Loading ${uri.substr(0, 64)}`);
 
         return this._parent.preprocessUrlAsync(this._rootUrl + uri).then((url) => {
```
CoPrez, line 2137: I think this will run into issues. If the uri is already a `data:` url then concatenating it with _rootUrl likely isn't what you want to do.

bghgary: Yup, reverted and updated the deprecated functions.

PR #10552, opened 2021-06-23
DeviceSourceManager - Fix touch on iOS

PolygonalSun:
Because of how iOS handles pointerIds in pointer events (uses negative numbers), I've added an array in the DeviceInputSystem to track all negative values and assign them a positive value that works with the DeviceSourceManager.

### DIFF
--- a/src/DeviceInput/Implementations/webDeviceInputSystem.ts
+++ b/src/DeviceInput/Implementations/webDeviceInputSystem.ts
@@ -624,9 +624,6 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 }
                 // Clear all active touches
                 while (this._activeTouchIds.pop() !== undefined) { }
-
-                    this._activeTouchIds.pop();
-                }
             }
         });
 
--- a/src/DeviceInput/Implementations/webDeviceInputSystem.ts
+++ b/src/DeviceInput/Implementations/webDeviceInputSystem.ts
@@ -339,7 +339,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
     private _handlePointerActions() {
         this._pointerMoveEvent = ((evt) => {
             const deviceType = this._getPointerType(evt);
-            const deviceSlot = (deviceType === DeviceType.Mouse) ? 0 : this._activeTouchIds[evt.pointerId];
+            const deviceSlot = (deviceType === DeviceType.Mouse) ? 0 : this._activeTouchIds.indexOf(evt.pointerId);
 
             if (!this._inputs[deviceType]) {
                 this._inputs[deviceType] = {};
@@ -419,7 +419,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
 
             if (deviceType === DeviceType.Touch) {
                 deviceSlot = this._rollingTouchId++;
-                this._activeTouchIds[evt.pointerId] = deviceSlot;
+                this._activeTouchIds[deviceSlot] = evt.pointerId;
             }
 
             if (!this._inputs[deviceType]) {
@@ -488,7 +488,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
 
         this._pointerUpEvent = ((evt) => {
             const deviceType = this._getPointerType(evt);
-            const deviceSlot = (deviceType === DeviceType.Mouse) ? 0 : this._activeTouchIds[evt.pointerId];
+            const deviceSlot = (deviceType === DeviceType.Mouse) ? 0 : this._activeTouchIds.indexOf(evt.pointerId);
 
             const pointer = this._inputs[deviceType]?.[deviceSlot];
             if (pointer && pointer[evt.button + 2] !== 0) {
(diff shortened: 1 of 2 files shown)


src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -587,17 +597,18 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
             if (this.isDeviceAvailable(DeviceType.Touch)) {
                 const pointer = this._inputs[DeviceType.Touch];
 
-                for (const deviceSlotKey in pointer) {
-                    const deviceSlot = +deviceSlotKey;
+                // Get list of active touch ids and clear each one in the inputs array
+                for (const deviceSlotKey in Object.keys(this._activeTouchIds)) {
+                    const deviceSlot = this._activeTouchIds[+deviceSlotKey];
```
sebavan, line 602: Could we do the +deviceSlotKey only once ?

PolygonalSun: I've changed code to just store +deviceSlotKey in a local var and use that throughout the blur code.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -609,6 +621,8 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                         this._unregisterDevice(DeviceType.Touch, deviceSlot);
                     }
                 }
+                // Clear all active touches
+                this._activeTouchIds = [];
```
RaananW, line 625: Avoid recreating the array
```suggestion
                this._activeTouchIds.length = 0;
```

RaananW (commented):
If I understand correctly from your description, pointerIds in iOS are negative?
How do you go about reading and writing those negative events from the activeTouchIds array? it seems like you are reading `evt.pointerId` or use it dorectly. Or did I miss an important line?

PolygonalSun:
> If I understand correctly from your description, pointerIds in iOS are negative?
> How do you go about reading and writing those negative events from the activeTouchIds array? it seems like you are reading `evt.pointerId` or use it dorectly. Or did I miss an important line?

If my understanding is correctly, I can use these negative values as a key to hold our assigned device slot values.  So whenever we get a pointerdown event, I'm using evt.pointerId as the key and giving it a value of our assigned, rolling touch id.  We still need the direct pointerId for pointer capturing but we use the rolling touch id from activeTouchIds for slot tracking in the DeviceInputSystem.

I do feel like there may be a more elegant way to handle that though.

RaananW:
> > If I understand correctly from your description, pointerIds in iOS are negative?
> > How do you go about reading and writing those negative events from the activeTouchIds array? it seems like you are reading `evt.pointerId` or use it dorectly. Or did I miss an important line?
> 
> If my understanding is correctly, I can use these negative values as a key to hold our assigned device slot values. So whenever we get a pointerdown event, I'm using evt.pointerId as the key and giving it a value of our assigned, rolling touch id. We still need the direct pointerId for pointer capturing but we use the rolling touch id from activeTouchIds for slot tracking in the DeviceInputSystem.
> 
> I do feel like there may be a more elegant way to handle that though.

this feels very odd  :-) 
there is no such thing as array[-1] in javascript. this should return undefined. But since you are setting it, the array acts like any other js object and just stores it. the array's length will still be 0, and no array function will work correctly (like forEach, map or pop).
why not store the rolling id as an index, and the pointer Id as the element and search for indexOf(pointerId) instead?

PolygonalSun:
> > > If I understand correctly from your description, pointerIds in iOS are negative?
> > > How do you go about reading and writing those negative events from the activeTouchIds array? it seems like you are reading `evt.pointerId` or use it dorectly. Or did I miss an important line?
> > 
> > 
> > If my understanding is correctly, I can use these negative values as a key to hold our assigned device slot values. So whenever we get a pointerdown event, I'm using evt.pointerId as the key and giving it a value of our assigned, rolling touch id. We still need the direct pointerId for pointer capturing but we use the rolling touch id from activeTouchIds for slot tracking in the DeviceInputSystem.
> > I do feel like there may be a more elegant way to handle that though.
> 
> this feels very odd :-)
> there is no such thing as array[-1] in javascript. this should return undefined. But since you are setting it, the array acts like any other js object and just stores it. the array's length will still be 0, and no array function will work correctly (like forEach, map or pop).
> why not store the rolling id as an index, and the pointer Id as the element and search for indexOf(pointerId) instead?
> 
> **EDIT** - .length will also not work as you expect. So you will have to clear the array in a different way

I'm following what you're saying.  I'll change stuff around to work with the rolling ids as the index.

src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -609,6 +622,10 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                         this._unregisterDevice(DeviceType.Touch, deviceSlot);
                     }
                 }
+                // Clear all active touches
+                while (this._activeTouchIds.length > 0) {
```
RaananW, line 626: so, not `.length` will work, but if you want to do it this way, you can always have ano-op loop like this:

```suggestion
                while (this._activeTouchIds.pop() !== undefined) { }

```

sebavan:
Sounds like smthg broke on this one @PolygonalSun

PolygonalSun:
> Sounds like smthg broke on this one @PolygonalSun

Yeah, it was a bad merge.  I've pushed a fix

RaananW (approved):

PR #10561, opened 2021-06-24
Engine: Move the shader code inliner to an extension

Popov72:
The aim is to reduce the bundle size when using the WebGL engine as it does not need shader inlining.

See https://forum.babylonjs.com/t/shadercodeinliner-bundled-even-when-not-used/21852


sebavan (changes requested):
In this case, why not placing the function in ThinEngine (default is almost no code) and overriding in WebGPU and Native ? I guess it would be enough to cute the dependency ?

As this kind of do the same :-) ??? or I am missing smthg

sebavan (approved):

Issue #10566, opened 2021-06-24
Add a Performance Profiler to display data in graphical format.


SahilTara:
## Description
This issue will be used to track the progress of the performance profiler feature.
Basically the goal is to enhance the experience of people who would like to figure out where performance issues are arising.
This will be done by allowing users to plot data over time for a variety of things such as fps, draw calls, number of active meshes, frame time. As this feature evolves more and more data sources will be added :D.

A bit of a rough visual for what I'm trying to build. Of course disclaimer that design is subject to change (and will change) this is just to illustrate what I'm trying to build.
![image](https://user-images.githubusercontent.com/32103099/123318358-1dff0a00-d4fd-11eb-800f-7d594dec7498.png)

This tool is going to be able to get engine specific data which chrome dev tools is currently not able to do.

We will be using a pop out format similar to the animation curve editor for the desktop flow, while mobile will have to record, export, then import on to the desktop.

I have made a list of milestones to cover for my internship (right now they are really rough and based on the information I know so they are subject to change around, and will continue to gain higher fidelity as I get closer to the milestone). If you've seen the [forum link](https://forum.babylonjs.com/t/babylon-performance-profiler/21826) you may notice the milestones below have already changed due to a better understanding of the requirements. 

## Milestone 1:

- [x] Ask team about what “data sources” they would like to see 
- [x] Ask community about what “data sources” they would like to see
- [x] Develop a list of data sources to be implemented.

## Milestone 2: 

- [x] Create button to open profiler
- [x] Have a pop up that displays a canvas 
- [x] Use the canvas API to create a line graph (single line)
  - [x] 	Draw axis (that dynamically adjusts to data)
  - [x] 	Draw data points
- [x] Implement graph scrolling 
- [x] Test graph with live data (can be something as simple as FPS counter)

## Milestone 3:

- [x] Allow for zooming into the graph
- [x] Allow for panning of the graph
- [x] Allow for hovering over data points and displaying the actual values.
- [x] Allow for custom colourization for lines.
- [x] Allow rescaling of window.
- [x] Allow for toggling of data sources

## Milestone 4:
- [x] Allow for the collection of performance data. 
- [x] Begin collecting FPS data.
- [x] Connect the graph to the collector.
- [x] Have tickers update with relevant info on the right side bar as in image.
- [x] Normalization of data

## Milestone 5:

- [x] Allow for the collection of draw calls, number of active meshes, frame time.
[...]

PR #10572, opened 2021-06-25
Add graph service basic methods and skeleton.

SahilTara:
Basically this pull request adds a graphing service, fully implements the clearing of the graph method and begins creating the drawing code. For now we use similar code that lived in the graph component to simply draw a line to the canvas. #10566

### DIFF
--- a/inspector/src/components/graph/graphSupportingTypes.ts
+++ b/inspector/src/components/graph/graphSupportingTypes.ts
@@ -1,10 +1,8 @@
 /**
  * Defines what data is needed to graph a point on the graph.
- *
- * @interface IPoint
  */
 interface IPoint {
-    x: number; // This will need to become a date of some kind with real data.
+    x: number; // This will probably need to become a date when we have real data.
     y: number;
 }
 
--- a/inspector/src/components/graph/graphSupportingTypes.ts
+++ b/inspector/src/components/graph/graphSupportingTypes.ts
@@ -1,20 +1,4 @@
-/**
- * Defines what data is needed to graph a point on the graph.
- */
-interface IPoint {
-    x: number; // This will probably need to become a date when we have real data.
-    y: number;
-}
-
-/**
- * Defines the shape of a dataset that our graphing service uses for drawing purposes.
- */
-export interface IPerfDataset {
-    color?: string;
-    data: IPoint[];
-    hidden?: boolean; // falsey by default!
-}
-
+import { IPerfDataset } from "babylonjs/Misc/interfaces/iPerfViewer"
 /**
  * Defines what settings our canvas graphing service accepts
  */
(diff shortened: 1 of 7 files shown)


inspector/src/components/graph/graphSupportingTypes.ts
```diff
@@ -1,3 +1,25 @@
-interface IDataset {
-    // empty for now, but will contain the shape of what a data set is.
+/**
+ * Defines what data is needed to graph a point on the graph.
+ *
+ * @interface IPoint
+ */
+interface IPoint {
+    x: number; // This will need to become a date of some kind with real data.
```
deltakosh, line 7: typo in the comment

SahilTara: I think I got it, I think it was just the wording and the interface tag (correct me if I'm wrong).

sebavan (commented):
My main question is why setting up a callback from graphComponent to perfViewerComponent to draw ???

It looks like the graphComponent is creating the service so it could as well call draw ???

SahilTara:
> My main question is why setting up a callback from graphComponent to perfViewerComponent to draw ???
> 
> It looks like the graphComponent is creating the service so it could as well call draw ???

Basically it is so the perfViewerComponent can decide when to update the data, the core will be providing data which we will subscribe to in the perfViewerComponent and we will simply call draw() again when we get the new data!

sebavan:
But you wont be able to call draw as you only have the callback :-) so as long as it is not called with the service it does not draw. Or I am missing smthg with is more than possible on Fridays :-)

SahilTara:
> But you wont be able to call draw as you only have the callback :-) so as long as it is not called with the service it does not draw. Or I am missing smthg with is more than possible on Fridays :-)

Yeah for now I only have the callback, but I will eventually be storing the instance outside of just the callback. So the callback will give the instance to perfViewerComponent, and then the perfViewerComponent will store the instance and call it when it needs to :D. It is correct that if it is not called by the service it does not draw.

sebavan:
This is discussed offline with @RaananW so it will get merge when he ll approve :-)

src/Misc/interfaces/iPerfViewer.ts
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * The y coordinate of the point.
+     */
+    y: number;
+}
+
+/**
+ * Defines the shape of a dataset that our graphing service uses for drawing purposes.
+ */
+export interface IPerfDataset {
```
RaananW, line 19: I assume that most (if not all) datasets will be time-based. This can be added later, of course, but shouldn't we have some form of a timestamp here?

Or - is the x axis of the graph anything other than time?

RaananW (commented):
Just one more question :-)

RaananW (approved):
(Left one question which can be resolved later)

src/Misc/interfaces/iPerfViewer.ts
SahilTara: Yes! We have the same idea the x axis will be plotting time! I believe the timestamp can be represented as a number. 

Would you want me to rename x: number to timestamp: number to make this more clear? Or is there a potential performance concern that you can see with storing timestamps as numbers?

PR #10590, opened 2021-06-29
Draw performance graph data and axis.

SahilTara:
This PR handles the drawing of x axis and the data! It also adds very basic scrolling capability which will be improved on later when we have zooming and panning! After this PR is merged milestone 2 will be complete :D in #10566 .

This is what it looks like when testing it out with live fps data (of course this connection isn't in the PR because it was a connection done for just test purposes).

https://user-images.githubusercontent.com/32103099/123847058-1b792780-d8e4-11eb-8a1f-d3a1b9cdfa1b.mp4

### DIFF
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -1,4 +1,4 @@
-import { ICanvasGraphServiceSettings, IMinMax } from "./graphSupportingTypes";
+import { ICanvasGraphServiceSettings, IMinMax, INiceTicks, IDrawableArea } from "./graphSupportingTypes";
 import { IPerfDataset, IPerfPoint } from "babylonjs/Misc/interfaces/iPerfViewer";
 
 /**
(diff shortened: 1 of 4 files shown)


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -38,17 +39,222 @@ export class CanvasGraphService {
...
+        // TODO: Make better sliding window code (accounting for zoom and pan).
+        // Keep only visible and non empty datasets and get a certain window of items.
+        const datasets = this.datasets.filter((dataset: IPerfDataset) => !dataset.hidden && dataset.data.length > 0).map((dataset: IPerfDataset) => ({
+            ...dataset,
+            data: dataset.data.slice(Math.max(dataset.data.length - 300, 0))
+        }));
+
+        datasets.forEach((dataset: IPerfDataset) => {
+            const timeMinMax = this._getMinMax(dataset.data.map((point: IPerfPoint) => point.timestamp));            
+            globalTimeMinMax = {min: Math.min(timeMinMax.min, globalTimeMinMax.min), max: Math.max(timeMinMax.max, globalTimeMinMax.max)};
```
RaananW, line 54: instead of reassigning, make this a `const` and set min and max of the object

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -38,17 +39,222 @@ export class CanvasGraphService {
         // First we clear the canvas so we can draw our data!
         this.clear();
 
-        // TODO: Draw each dataset
-        this.datasets.forEach((dataset: IPerfDataset) => {
-            // do drawing
+        // Get global min max of time axis (across all datasets).
+        let globalTimeMinMax = {min: Infinity, max: 0};
```
RaananW, line 43: shouldn't that be flipped? max is infinity and min is 0?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -38,17 +39,222 @@ export class CanvasGraphService {
...
-        this.datasets.forEach((dataset: IPerfDataset) => {
-            // do drawing
+        // Get global min max of time axis (across all datasets).
+        let globalTimeMinMax = {min: Infinity, max: 0};
+
+        // TODO: Make better sliding window code (accounting for zoom and pan).
+        // Keep only visible and non empty datasets and get a certain window of items.
+        const datasets = this.datasets.filter((dataset: IPerfDataset) => !dataset.hidden && dataset.data.length > 0).map((dataset: IPerfDataset) => ({
+            ...dataset,
+            data: dataset.data.slice(Math.max(dataset.data.length - 300, 0))
```
RaananW, line 49: is the `300` limit fixed? can it be changed / configured by the user in the future?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -38,17 +39,222 @@ export class CanvasGraphService {
...
+     * 
+     * @param minMax the minimum and maximum values of the axis
+     * @param spaceAvailable the total amount of space we have allocated to our axis
+     * @returns a list of "nice" tickS.
+     */
+    private _getTicks(minMax: IPerfMinMax, spaceAvailable: number): number[] {
+        const { min, max } = minMax;
+        const minTickSpacing = 40; 
+
+        let ticks: number[] = [];
```
RaananW, line 138: maybe try reusing this array instead of creating it every time you draw the graph? This will help with garbage collection and will support realtime rendering

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -38,17 +39,222 @@ export class CanvasGraphService {
...
+    }
+
+    /**
+     * Draws the time axis, adjusts the drawable area for the graph.
+     * 
+     * @param timeMinMax the minimum and maximum for the time axis. 
+     * @param drawableArea the current allocated drawable area. 
+     */
+    private _drawTimeAxis(timeMinMax: IPerfMinMax, drawableArea: IGraphDrawableArea) {
+        const { _ctx: ctx } = this;
```
RaananW, line 89: is this just for convenience, or is there a context issue in this function?

RaananW (commented):
Looking good!!

Try to create as little amount of objects as possible, and reuse as much as you can.

inspector/src/components/graph/canvasGraphService.ts
SahilTara: I don't think so because if i set max to infinity and min to 0 when i do Math.min(globalTimeMinMax.min, timeMinMax.min); then the result will always be 0 and similarly for max where the result will always be infinity!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: For now** yes, I'm still working out how zooming and panning will work but my thoughts are the user would be manipulating this formula.

I'm not entirely sure how useful being able to configure the initial size of the sliding window may be, but I could add the setting in very easily if wanted

inspector/src/components/graph/canvasGraphService.ts
SahilTara: I think I could reduce the times I need to recreate the array, but probably cant get away with reusing the array every time. 

I could potentially reuse the array when oldNiceMin === currNiceMin && oldNiceMax === currNiceMax
(Perhaps also && oldSpacing === spacing because the loop depends on spacing as well).

When any of those are different I'll probably need to create a fresh array.

Do you think this is a good solution to the problem?

inspector/src/components/graph/canvasGraphService.ts
RaananW: awesome. just keep track on it.

inspector/src/components/graph/canvasGraphService.ts
RaananW: can't you reuse the array while emptying it on each frame? setting the array's length to 0 is a quick operation which clears the array. so instead of `let ticks: number[] = [];` you just do `this.ticksArray.length = 0`

SahilTara: Oooo okay I was under the impression that the performance implications for both those operations were the same!

I will change this to .length = 0 instead :D

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts

sebavan:
@RaananW I ll let you merge once ready, thanks

PR #10598, opened 2021-06-30
WebGPU: add hasOriginBottomLeft property

Popov72:
Draft for the time being so that @CedricGuillemet can test on Native.

I need also to perform additional testing in WebGPU, but currently some validation tests are broken because of a Chrome Canary update.

There are also some various fixes:
* fix crash of fluentBackplate in WebGPU
* fix bump not working in node materials
* fix uv not working in node material when using morph targets


sebavan:
This is definitely not a joke PR :-) I am starting reviewing now, this might take a bit.

Popov72:
One problem: compressed textures are inverted...

We can't invert the texture at loading time because we can't render to a compressed texture format. That means that the texture is the same (same orientation) than when `hasOriginBottomLeft=true`. But because we flip the Y coordinate in all texture lookups, the texture appears inverted when `hasOriginBottomLeft=false`...

It's the responsibility of the user to flip their v coordinates when using compressed texture and when `hasOriginBottomLeft=false`.

There are some other cases when the user will have to flip their v coordinates: when dealing with textures that are not for display purpose. That's what I did on our end for some shaders (morph or rgbe decoding, for eg), where I do a flipY on some uv coordinates so that the automatic process of flipping Y in all texture lookups will in effect undo the flipping.

I can't see a way to avoid those manual changes in those circumstances, I don't know if it's acceptable...

For WebGPU we can live without the PR (we will only have some additional texture copies to invert Y when necessary + a final copy to invert the framebuffer - it's already working, you simply have to set `engine.hasOriginBottomLeft = true`) but for Native I'm not so sure(?)

sebavan:
Yup sounds like it might create issues with the adoption of KTX2 for instance ??? Let s discuss it offline tomorrow during stand up or on Friday ?

CedricGuillemet:
If I want to let the JS side do the final YFlip rendering pass, do I only have to set `engine.hasOriginBottomLeft = true` ?

Popov72:
> If I want to let the JS side do the final YFlip rendering pass, do I only have to set engine.hasOriginBottomLeft = true ?

No because the WebGL side has not been affected by the PR. It's only on the WebGPU side that I handle both `hasOriginBottomLeft = true` and `hasOriginBottomLeft = false` cases (and the final YFlip in the `hasOriginBottomLeft = true` case is some special WebGPU code, it would not work in the Native engine).

PR #10599, opened 2021-06-30
DeviceInputSystem - Modify system to support multiple canvas'/views

PolygonalSun:
This PR contains changes to the InputManager and DeviceInputSystem to allow for event listeners to be swapped to the active input element when attachControl is called.


sebavan (approved):
Lgtm let me do a couple local tests

sebavan (approved):

PR #10609, opened 2021-07-01
Fix #10607

### LINKED ISSUE #10607: No onError observable on videotexture
When I try to load a video I can not add an error observable to check if there was an issue when loading it.

There should be a way to add an onError Observable to videotexture.

[Forum Link](https://forum.babylonjs.com/t/how-to-add-on-error-observable-callback-on-videotexture/22047)

deltakosh:


src/Materials/Textures/videoTexture.ts
```diff
@@ -74,6 +74,39 @@ export class VideoTexture extends Texture {
     private _createInternalTextureOnEvent: string;
     private _frameId = -1;
     private _currentSrc: Nullable<string | string[] | HTMLVideoElement> = null;
+    private _onError?: Nullable<(message?: string, exception?: any) => void>;
+    private _errorFound = false;
+
+    private _processError(reason: any) {
+        this._errorFound = true;
+        if (this._onError) {
+            this._onError(reason.message);
```
sebavan, line 83: wonder if we should reason?.message here and below ?

sebavan (approved):

PR #10614, opened 2021-07-01
Added zoom in and zoom out functionality to performance viewer 🔍

SahilTara:
This pullrequest implements the "Allow zooming into the graph" feature of milestone 3 in #10566. It does this by adding the following:
- Attaches an event listener on construction of the service,
- Event listener triggers whenever we use the mouse wheel to zoom in or out (Forward = in, Backwards = out).
- When triggered we will get how much to zoom based on the mouseDelta before we triggered the event and bind the zoom level to be between 60 datapoints (at minimum) and the length of the largest dataset (at maximum).
- We remove event listeners on unmount.

We also draw a playhead at the end of the data.  To help the user know where the future data is, and to move the changing bit to the center. This gives the eye something to lock onto and reduces the anxiety of the user when zooming into the graph fully. Without this the graph would look very chaotic.

To draw the playhead we do the following:
- Adjust the maximum datapoint in the x axis based on the scale factor (this will scale all of our data to scaleFactor% of the canvas). Note: The scale factor is a number between 0.0 and 1.0 representing the percentage of the canvas. Currently the scaleFactor is a constant, but if we make this dynamic we should look at if there are any occurrences of division by zero, although I don't see a point where scaleFactor will ever be 0 (as that means that the data is fully hidden and we only see the playhead ).

- We then paint the future box and a divider to 1-scaleFactor of the screen, and then paint the playhead about 8 pixels before the divider. This is done to give the user the ability to see the delta better.

Anyways here is  a video quickly showing off what is included in this PR:

https://user-images.githubusercontent.com/32103099/124183320-0b994900-da86-11eb-8708-499dcc67be17.mp4

### DIFF
--- a/inspector/src/components/graph/canvasGraphComponent.tsx
+++ b/inspector/src/components/graph/canvasGraphComponent.tsx
@@ -14,8 +14,11 @@ export const CanvasGraphComponent: React.FC<ICanvasGraphComponentProps> = (props
         if (!canvasRef.current) {
             return;
         }
+
         // temporarily set empty array, will eventually be passed by props!
-        canvasServiceCallback(new CanvasGraphService(canvasRef.current, {datasets: []}));
+        const canvasGraphService = new CanvasGraphService(canvasRef.current, {datasets: []});
+        canvasServiceCallback(canvasGraphService);
+        return () => canvasGraphService.destroy();
     }, [canvasRef])
    
     return (
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -307,7 +307,6 @@ export class CanvasGraphService {
 
         // Bind the zoom between [minZoom, maxZoom]
         this._sizeOfWindow = Math.min(Math.max(this._sizeOfWindow - amount, minZoom), maxZoom);
-        console.log(this._sizeOfWindow);
     }
 
     /**
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -11,7 +11,7 @@ export class CanvasGraphService {
     private _width: number;
     private _height: number;
     public readonly datasets: IPerfDataset[];
-
+    private _sizeOfWindow: number = 300;
     private _ticks: number[];
 
     /**
@@ -24,9 +24,9 @@ export class CanvasGraphService {
         this._ctx = canvas.getContext && canvas.getContext("2d");
         this._width = canvas.width;
         this._height = canvas.height;
-
         this.datasets = settings.datasets;
         this._ticks = [];
+        this._attachEventListeners(canvas);
     }
 
     /**
@@ -49,7 +49,7 @@ export class CanvasGraphService {
         // Keep only visible and non empty datasets and get a certain window of items.
         const datasets = this.datasets.filter((dataset: IPerfDataset) => !dataset.hidden && dataset.data.length > 0).map((dataset: IPerfDataset) => ({
             ...dataset,
-            data: dataset.data.slice(Math.max(dataset.data.length - 300, 0))
+            data: dataset.data.slice(Math.max(dataset.data.length - this._sizeOfWindow, 0))
         }));
 
         datasets.forEach((dataset: IPerfDataset) => {
@@ -126,7 +126,6 @@ export class CanvasGraphService {
         ctx.stroke();
         ctx.restore();
     }
-
 
     /**
      * Generates a list of ticks given the min and max of the axis, and the space available in the axis.
@@ -255,6 +254,24 @@ export class CanvasGraphService {
         }
 
         return startingPixel + normalizedValue * spaceAvailable;
+    }
+
+    private _attachEventListeners(canvas: HTMLCanvasElement) {
+        canvas.addEventListener("wheel", this._handleZoom.bind(this));
+    }
+
+
+    private _handleZoom(event: WheelEvent) {
+        event.preventDefault();
+        
+        if (!event.deltaY) {
+            return;
+        }
+
+        const amount = (event.deltaY * -0.01 | 0) * 100;
+        this._sizeOfWindow = Math.max(this._sizeOfWindow - amount, 60);
+        console.log(this._sizeOfWindow);
+        return;
     }
     
     /**
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -2,6 +2,15 @@ import { ICanvasGraphServiceSettings, IPerfMinMax, IGraphDrawableArea } from "./
 import { IPerfDataset, IPerfPoint } from "babylonjs/Misc/interfaces/iPerfViewer";
 
 const defaultColor = "#000";
+const futureBoxColor = "#dfe9ed";
+const dividerColor = "#0a3066";
+const playheadColor = "#b9dbef";
+
+const playheadSize = 8;
+const dividerSize = 2;
+
+// Currently the scale factor is a constant but when we add panning this may become formula based.
+const scaleFactor = 0.8;
 
 /**
  * This class acts as the main API for graphing given a Here is where you will find methods to let the service know new data needs to be drawn,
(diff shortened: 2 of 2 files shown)


inspector/src/components/graph/canvasGraphComponent.tsx
```diff
@@ -14,8 +14,11 @@ export const CanvasGraphComponent: React.FC<ICanvasGraphComponentProps> = (props
         if (!canvasRef.current) {
             return;
         }
+
         // temporarily set empty array, will eventually be passed by props!
-        canvasServiceCallback(new CanvasGraphService(canvasRef.current, {datasets: []}));
+        const canvasGraphService = new CanvasGraphService(canvasRef.current, {datasets: []});
+        canvasServiceCallback(canvasGraphService);
```
sebavan, line 20: I am still troubled by this calling back to set it on the parent :-) and wonder if we could use a global state or send command to the component instead from the parent one ? I am afraid of ending up with the parent still holding to a destroyed service for instance ?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -11,7 +13,7 @@ export class CanvasGraphService {
     private _width: number;
     private _height: number;
     public readonly datasets: IPerfDataset[];
```
sebavan, line 15: [NIT] might be more readable to split privates and publics

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -49,7 +51,7 @@ export class CanvasGraphService {
         // Keep only visible and non empty datasets and get a certain window of items.
```
sebavan, line 51: Could add in the todo as well to limit allocations ? to prevent extra GC if draw is called at high frequency

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -65,15 +67,22 @@ export class CanvasGraphService {
             right: this._width,
         };
 
+        // we will now rescale the maximum for the playhead.
+        // Currently the scale factor is a constant but when we add panning this may become formula based.
+        const scaleFactor = 0.8;
```
sebavan, line 72: Would be great to move at the top of the file with other constants ? and move it back in if it changes.

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -256,6 +265,103 @@ export class CanvasGraphService {
...
+        const minZoom = 60;
+
+        // The max zoom is the largest dataset's length.      
+        const maxZoom = this.datasets.map((dataset: IPerfDataset) => dataset.data.length)
+                        .reduce((maxLengthSoFar: number, currLength: number) => {
+                            return Math.max(currLength, maxLengthSoFar)
+                        }, 0);
+
+        // Bind the zoom between [minZoom, maxZoom]
+        this._sizeOfWindow = Math.min(Math.max(this._sizeOfWindow - amount, minZoom), maxZoom);
```
sebavan, line 309: Could use Babylon Scalar Clamp function ?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -256,6 +265,103 @@ export class CanvasGraphService {
...
+     * @param scaleFactor The Percentage between 0.0 and 1.0 of the canvas the data gets drawn on.  
+     */
+    private _drawPlayheadRegion(drawableArea: IGraphDrawableArea, scaleFactor: number) {
+        const { _ctx: ctx } = this;
+       
+        if (!ctx) {
+            return;
+        }
+
+        const dividerSize = 2;
```
sebavan, line 325: would be great to extract constants at the top

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Oh woah you're right, I didn't notice I thing during a rebase I must've misplaced the datasets variable! good callout!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Oh I didn't know that existed! For sure

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Done

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Hmm I've added the comment but I'm having a bit of trouble right now imagining how I'd reduce the allocations happening due to the map.

The alternative I'm seeing right now is we slice on every iteration the items we need. But that feels just as bad. I may need to think of it more, but if you have any suggestions I'm open to them

sebavan: you could probably keep a data array in memory and manually filter from source into this array preventing recreation of the filter then map and also spread operator  but the code might be harder to read so it really depends of the draw frequency ?

inspector/src/components/graph/canvasGraphComponent.tsx
SahilTara: If it has a destroyed service, would you want the ability to check if your service is destroyed like "isDestroyed"?

For global state, I am wondering what would happen if we want multiple CanvasComponents?  Could you elaborate more on this? Perhaps I'm misunderstanding the global state concept you're suggesting. 

Could you also elaborate a bit more about the other method you're suggesting?

My concerns for if we handle creation in the parent is how would we know when the canvas has been unmounted?

sebavan: To keep it at the simplest, a ref could work and you call functions on it ?

sebavan: Even if I totally understand it is not the "React Way" :-)

sebavan: @RaananW might have a better alternative

SahilTara: > To keep it at the simplest, a ref could work and you call functions on it ?

So basically something like an action reducer which I send some Actions enum values? That could work.

I'll wait to see if @RaananW  also has any suggestions

SahilTara: Seb and I talked about this offline, and I have made note of possible suggestions. Concerns make a lot more sense to me now. We will be keeping this callback for now, but will remove it the moment there is a bug that arises from it, or after the data layer is implemented (whichever comes first).

The idea is after the data layer is implemented we may be able to use observables to tell the child component "Hey I've updated the data, you can draw now with this new data".   Instead of having this circular dependency type approach. I may change this approach later on but it is the idea I am leaning towards.

inspector/src/components/graph/canvasGraphService.ts
SahilTara: I have made note of this and will explore possible solutions including the ones we discussed in our offline chat (Potentially using typed arrays). Will continue thinking about how to reduce the garbage collection pressure from this filter and map.

inspector/src/components/graph/canvasGraphComponent.tsx
RaananW: I always think of a service as a stateless singleton. It's simply a collection of functions with a common theme. In this case, this service is canvas-dependent, so it requires the canvas as a reference for each draw call (or, as you do it, sent in the constructor as a reference).
The "react way" would be to have those functions here, as they are a part of this component and technically have nothing to do with the parent. why does the parent call the draw function and not this component?

SahilTara: Basically its because for now its easier to make changes in the parent without updating the state

SahilTara: So to give a full summary for everyone else:

The reason I did this originally was because there was a sibling component which needed to share data between itself and the canvas.

After speaking with both Seb and Raanan the plan is to do the following:

Basically we will be having two observables: One for metadata, another for data (wrapped by some type of class inside of core)

So there will be a toggling/colourizing component that is a sibling to the canvas component which will be able to update the metadata, and will be a view on the metadata itself too.

Then there will be the canvas component which subscribes to changes to the data and the metadata, it will simply be a view for them.

I am marking this conversation as resolved, and to fix once the data class is finished!!

sebavan:
Looks all good to me :-) @RaananW can I merge it ?

PR #10620, opened 2021-07-02
Fix support for camera output render targets

Popov72:


src/scene.ts
```diff
@@ -4009,6 +4026,10 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         this.resetCachedMaterial();
         this._renderId++;
 
+        if (!this.prePass && bindFrameBuffer) {
+            this._bindFrameBuffer(this._activeCamera);
+        }
```
Popov72, line 4031: I had to call `_bindFrameBuffer` here because the `this.setTransformMatrix` / `this.updateTransformMatrix()` calls that follow need that the right frame buffer is bound: the `camera.getProjectionMatrix()` call will compute the aspect ratio based on the currently bound frame buffer.

RaananW (approved):

deltakosh (approved):

PR #10621, opened 2021-07-02
NME: Fix uv sets 3 to 6

Popov72:


deltakosh (approved):

sebavan:
Not to sure to understand how it is fixing it but I trust you :-)

sebavan:
Ok double checked with @PirateJC so it is all good for the removal and I am adding 3 to 6 back in the drop down in the nightly.

PR #10625, opened 2021-07-04
Added parameter type in the constructor, added async method for loadi…

Null0924:
Added the typescript type in the constructor. Added an async method to load the layout.


gui/src/2D/xmlLoader.ts
```diff
@@ -327,9 +331,41 @@ export class XmlLoader {
...
         xhttp.send();
     }
+    /**
+     * Initiates the xml layout loading asynchronously
+     * @param xmlFile defines the xml layout to load
+     * @param rootNode defines the node / control to use as a parent for the loaded layout controls.
+     * @returns Promise
+
+     */
+    public async loadLayoutAsync(xmlFile: any, rootNode: any): Promise<boolean> {
```
sebavan, line 346: Could you instead wrap the LoadLayout function where you d expose a onError callback ? this would help factorize the code

sebavan (changes requested):
Looks good to me but would be great to factorize the code and type the callback ?

gui/src/2D/xmlLoader.ts
Null0924: Yes, you're right, just did another commit.

gui/src/2D/xmlLoader.ts
```diff
@@ -312,24 +315,46 @@ export class XmlLoader {
      * @param rootNode defines the node / control to use as a parent for the loaded layout controls.
      * @param callback defines the callback called on layout load.
      */
-    public loadLayout(xmlFile: any, rootNode: any, callback: any): void {
+    public loadLayout(xmlFile: any, rootNode: any, onSuccess: any = null, onError: any = null): void {
```
sebavan, line 318: could you replace the type onSuccess: any by onSuccess: Nullable<() => void> and onError: any by onError : Nullable<(error: srting) => void>

sebavan (changes requested):
just a couple typings fix and you are good to go.

gui/src/2D/xmlLoader.ts
Null0924: okay, done :) .

sebavan (approved):

sebavan:
Thanks, I ll merge tomorrow before the next nightly as I am monitoring a regression at the moment.

Null0924:
> 
> 
> Thanks, I ll merge tomorrow before the next nightly as I am monitoring a regression at the moment.



> 
> 
> Thanks, I ll merge tomorrow before the next nightly as I am monitoring a regression at the moment.

Thanks, I added a couple more commits because I just noticed you can not dispose the GUI after loading directly from the loader class.

PR #10627, opened 2021-07-04
Mesh: Fix problems when using both instances and thin instances

Popov72:
See https://forum.babylonjs.com/t/strange-thin-intances-behaviour/22112

Prioritize thin instances when using both instances and thin instances at the same time (having both enabled at the same time is not possible).


sebavan:
Looks good to me but it also looks like it broke a few tests I ll merge on Monday once the tests have been fixed

PR #10633, opened 2021-07-05
Fix Obj Serializer

sebavan:
https://forum.babylonjs.com/t/objexport-is-undefined-and-missing-from-cdn/22094

### DIFF
--- a/serializers/src/OBJ/objSerializer.ts
+++ b/serializers/src/OBJ/objSerializer.ts
@@ -21,6 +21,9 @@ export class OBJExport {
     public static OBJ(mesh: Mesh[], materials?: boolean, matlibname?: string, globalposition?: boolean): string {
         const output: string[] = [];
         let v = 1;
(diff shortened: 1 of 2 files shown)


serializers/src/OBJ/objSerializer.ts
```diff
@@ -87,14 +93,16 @@ export class OBJExport {
                 for (i = 0; i < trunkUV.length; i += 2) {
                     output.push("vt " + trunkUV[i] + " " + trunkUV[i + 1]);
                 }
+                currentTextureV++;
```
Popov72, line 96: Shouldn't this line be inside the loop?

Popov72 (approved):
LGTM save for a small thing.

serializers/src/OBJ/objSerializer.ts
sebavan: I do not understand how my tests worked like this :-)

PR #10635, opened 2021-07-06
Fix test nearmenu

CraigFeldspar:


sebavan:
Should the hardcoded values be constants in the file ?

sebavan (commented):
I wonder if the constant values should be defined as static ???

PR #10636, opened 2021-07-06
Fix negative scale support for physics impostors

### LINKED ISSUE #10283: Physics imposters are not created correctly for negatively scaled meshes.
**Repro**
- Bug repro: https://playground.babylonjs.com/#TWVYMQ#7
- Expected result: BoxImposter should match the cube mesh
- Current result: BoxImposter is incorrect

**Screenshots**
image

**Additional context**
I stumbled upon this issue when trying to create imposters for imported gltf scenes created with Blender. See playground [here](https://playground.babylonjs.com/#TWVYMQ#5). I see two use cases for which solving this problem would help;

1. When a user wants to create an imposter for a purposely flipped mesh, such that it has a desired orientation
2. When a user wants to create an imposter for an imported mesh from a different coordinate system

A workaround exists by baking the scaling into the mesh, like so: https://playground.babylonjs.com/#TWVYMQ#6

CedricGuillemet:
fixes #10283 and adds capsule debug view. Fixes scaling for cylinder physics debug display.


sebavan (commented):
LGTM, i ll merge it before the nightly

Popov72 (approved):

PR #10639, opened 2021-07-06
Reduce MergeMeshes resource utilization

ryantrem:
`MergeMeshes` tends to be heavily used in Babylon React Native to reduce the size of the scene graph since JS execution is relatively slow without JIT. However, `MergeMeshes` can be relatively slow (e.g. a few seconds when merging say 200-300 meshes for a glb, or a few hundred milliseconds when merging a few dozen meshes created at runtime, which can be a noticeable hitch). Given this, I spent a little time profiling `MergeMeshes` and improving several aspects of it:
- Rather than merge pairs of meshes repeatedly until all meshes are merged, just merge all meshes in one pass.
- Figure out how big we'll need the arrays to be up front and allocate them only once (the final size needed for all merged meshes).
- Don't use `Array.concat` as it is really slow and can't merge more than two arrays at once.

These changes make `MergeMeshes` about **5x-10x faster**, depending on the scenario. It also **reduces memory pressure about about 10x-12x**.

### Before
In this example, there were about 250 meshes. It took about 1.5 seconds and memory spiked up to about 1.4GB.
image

### After
Again, 250 meshes, but now taking about 233ms and memory climbing to only about 129mb (without big memory spikes and big GCs).
image

Unfortunately the change in perf was not as significant in Babylon Native, so I will have another round of changes coming that specifically targets native.

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -4789,6 +4789,9 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         subdivideWithSubMeshes?: boolean,
         multiMultiMaterials?: boolean
     ): Nullable<Mesh> {
+        // Remove any null/undefined entries from the mesh array
+        meshes = meshes.filter(Boolean);
+
         if (meshes.length === 0) {
             return null;
         }
@@ -4799,13 +4802,11 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
 
             // Counting vertices
             for (index = 0; index < meshes.length; index++) {
-                if (meshes[index]) {
-                    totalVertices += meshes[index].getTotalVertices();
+                totalVertices += meshes[index].getTotalVertices();
 
-                    if (totalVertices >= 65536) {
-                        Logger.Warn("Cannot merge meshes because resulting mesh will have more than 65536 vertices. Please use allow32BitsIndices = true to use 32 bits indices");
-                        return null;
-                    }
+                if (totalVertices >= 65536) {
+                    Logger.Warn("Cannot merge meshes because resulting mesh will have more than 65536 vertices. Please use allow32BitsIndices = true to use 32 bits indices");
+                    return null;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


deltakosh (approved):

sebavan:
Sounds like an amazing improvement, I ll take a closer look before merging !!!

src/Meshes/mesh.vertexData.ts
```diff
@@ -464,103 +464,121 @@ export class VertexData {
...
-        if (!other) {
+        if (nonNullOthers.length === 0) {
             return source;
         }
 
-        var len = other.length + source.length;
-        var isSrcTypedArray = source instanceof Float32Array;
-        var isOthTypedArray = other instanceof Float32Array;
+        if (!source) {
+            return this._mergeElement(nonNullOthers[1], nonNullOthers.slice(1));
```
Popov72, line 555: Shouldn't it be:
```typescript
            return this._mergeElement(nonNullOthers[0], nonNullOthers.slice(1));
```
?

src/Meshes/mesh.vertexData.ts
```diff
@@ -464,103 +464,121 @@ export class VertexData {
...
+        }
 
-        // use non-loop method when the source is Float32Array
-        if (isSrcTypedArray) {
-            var ret32 = new Float32Array(len);
+        const len = source.length + nonNullOthers.reduce((othersSumLen, other) => othersSumLen += other.length, 0);
```
Popov72, line 558: nit: same thing than above:
```typescript
nonNullOthers.reduce((curSum, other) => curSum + other.length, 0);
```

src/Meshes/mesh.ts
```diff
@@ -4871,7 +4863,15 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
             }
         }
 
-        source = <Mesh>source;
+        const source = meshes[0];
```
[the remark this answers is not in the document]
ryantrem: Yes, you're right. I changed it to filter out null/undefined entries immediately and then removed the null checks. I think this should result in the same behavior as before.

src/Meshes/mesh.vertexData.ts
ryantrem: Wow how did I do that!? Glad you caught this one!

src/Meshes/mesh.ts
```diff
@@ -4871,7 +4863,15 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
-        source = <Mesh>source;
+        const source = meshes[0];
+
+        const getVertexDataFromMesh = (mesh: Mesh) => {
+            const wm = mesh.computeWorldMatrix(true);
+            const vertexData = VertexData.ExtractFromMesh(mesh, true, true);
+            vertexData.transform(wm);
+            return vertexData;
+        };
+        const vertexData = getVertexDataFromMesh(source).merge(meshes.slice(1).map(mesh => getVertexDataFromMesh(mesh)));
```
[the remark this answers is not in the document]
ryantrem: Filtered out null/undefined meshes.

sebavan (approved):
I ll merge as soon as the build is green THX

src/Meshes/mesh.vertexData.ts
sebavan: @Popov72 King of reviews :-)

PR #10642, opened 2021-07-07
Add panning to the performance profiler graph!

SahilTara:
The following PR handles panning to the profiler graph. What this means is the user will be able to pan which will keep the user in the position they've panned to, instead of dragging them along with the real time data (as seen in real time mode).  The user can resume to real time by "catching up" to the real time data, be this through zooming out all the way or panning to catch up. 

Right now it may be a bit confusing to figure out where the real time data is due to the playhead being static. In a future PR it will become dynamic, so that it will only come fully into view as you catch up/have caught up to the real time data. This PR only deals with the panning functionality, and returning to "real time mode". 

Completes the "Allow for panning of the graph" item in #10566 

Here's a video showing what is possible with this PR:

https://user-images.githubusercontent.com/32103099/124811344-615f6c80-df30-11eb-81b9-464e1e18efc2.mp4

### DIFF
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -58,7 +58,7 @@ export class CanvasGraphService {
         this.clear();
 
         // Get global min max of time axis (across all datasets).
-        let globalTimeMinMax = {min: Infinity, max: 0};
+        const globalTimeMinMax = {min: Infinity, max: 0};
 
         // TODO: Make better sliding window code (accounting for zoom and pan).
         // TODO: Perhaps see if i can reduce the number of allocations.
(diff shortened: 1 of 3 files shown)


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -58,20 +60,45 @@ export class CanvasGraphService {
...
-        // TODO: Make better sliding window code (accounting for zoom and pan).
         // TODO: Perhaps see if i can reduce the number of allocations.
         // Keep only visible and non empty datasets and get a certain window of items.
-        const datasets = this.datasets.filter((dataset: IPerfDataset) => !dataset.hidden && dataset.data.length > 0).map((dataset: IPerfDataset) => ({
-            ...dataset,
-            data: dataset.data.slice(Math.max(dataset.data.length - this._sizeOfWindow, 0))
-        }));
+        const datasets = this.datasets.map((dataset: IPerfDataset) => {
+            // skip hidden and empty datasets!
+            if (dataset.data.length === 0 || !!dataset.hidden) {
```
sebavan, line 69: thinking through overallocation in this whole section will be necessary :-)

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -284,6 +311,9 @@ export class CanvasGraphService {
      */
     private _attachEventListeners(canvas: HTMLCanvasElement) {
         canvas.addEventListener("wheel", this._handleZoom);
+        canvas.addEventListener("mousedown", this._handlePanStart);
+        // The user may stop panning outside of the canvas size so we should add the event listener to the document.
```
sebavan, line 315: Great that you put a comment here :-)

sebavan (commented):
This looks overall ok to me and it would be great to improve the memory allocation of the draw function as well as ensure on unmount that every event have been removed from the canvas :-)

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Yep, of course!! I have it pinned as a to-do item 😄. Probably going to have a typed array hold the items, as we've discussed. Going to put it off until the data layer is implemented though!!

PR #10645, opened 2021-07-07
DeviceInputSystem: Fix FakeMove

PolygonalSun:
This PR adds a line to make sure that the input array where the device state for Mouse is correct, with respect to move events.

### DIFF
--- a/src/DeviceInput/Implementations/webDeviceInputSystem.ts
+++ b/src/DeviceInput/Implementations/webDeviceInputSystem.ts
@@ -395,10 +395,11 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                     fireFakeMove = false;
                 }
                 // Lets Propagate the event for move with same position.
-                if (fireFakeMove) {
+                if (fireFakeMove && evt.button !== -1) {
                     deviceEvent.inputIndex = PointerInput.FakeMove;
                     deviceEvent.previousState = 0;
                     deviceEvent.currentState = 0;
+                    pointer[evt.button + 2] =  Math.abs(pointer[evt.button + 2] - 1); // Reverse state of button if evt.button has value
 
                     this.onInputChangedObservable.notifyObservers(deviceEvent);
                 }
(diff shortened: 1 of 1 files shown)


src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -395,10 +395,11 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                     fireFakeMove = false;
                 }
                 // Lets Propagate the event for move with same position.
-                if (fireFakeMove) {
+                if (fireFakeMove && evt.button !== -1) {
                     deviceEvent.inputIndex = PointerInput.FakeMove;
                     deviceEvent.previousState = 0;
                     deviceEvent.currentState = 0;
+                    pointer[evt.button + 2] =  Math.abs(pointer[evt.button + 2] - 1); // Reverse state of button if evt.button has value
```
sebavan, line 402: Totally unsure what this one does ?

PolygonalSun: With any pointer events, if there are any changes to what buttons are pressed, the PointerEvent's button property will be populated with the relevant button value.  We then use this to toggle the value of that button using this line.  Since the pointer button values can only be 0 or 1 (and init to 0), we would just subtract one and take the absolute value:
if 1 then |1 - 1| = 0
if 0 then |0 - 1| = |-1| = 1

If you think that I should use a different approach, please let me know.

sebavan: pointer[evt.button + 2] = (pointer[evt.button + 2] ? 0 : 1); 
or
pointer[evt.button + 2] = 1 - pointer[evt.button + 2]
or 
pointer[evt.button + 2] ^= 1
or
pointer[evt.button + 2] = +!pointer[evt.button + 2]

:-) In your case the second one might do but I was mostly wondering about the + 2, adding a comment would help

PolygonalSun: Ah!  Okay, I can add a comment for that.

sebavan: Sounds like a plan, thanks I ll merge right after

sebavan (approved):

PR #10646, opened 2021-07-08
Use Layer render camera for Scene ray picking

CedricGuillemet:
fixes https://forum.babylonjs.com/t/gizmo-issue-with-two-viewports-and-gui/22229/3
Use layer render camera for scene ray picking when it's not null.

### DIFF
--- a/src/Rendering/utilityLayerRenderer.ts
+++ b/src/Rendering/utilityLayerRenderer.ts
@@ -200,7 +200,7 @@ export class UtilityLayerRenderer implements IDisposable {
                             scenePick = new PickingInfo();
                         }
                     } else {
-                        let previousActiveCamera: Nullable<Camera> = null
+                        let previousActiveCamera: Nullable<Camera> = null;
                         // If a camera is set for rendering with this layer
                         // it will also be used for the ray computation
                         // To preserve back compat and because scene.pick always use activeCamera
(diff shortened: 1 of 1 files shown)


src/Rendering/utilityLayerRenderer.ts
```diff
@@ -200,9 +200,26 @@ export class UtilityLayerRenderer implements IDisposable {
...
+                        // If a camera is set for rendering with this layer
+                        // it will also be used for the ray computation
+                        // To preserve back compat and because scene.pick always use activeCamera
+                        // it's substituted temporarily and a new scenePick is forced.
+                        // otherwise, the ray with previously active camera is always used.
+                        // It's set back to previous activeCamera after operation.
+                        if (this._renderCamera)
+                        {
+                            previousActiveCamera = scene.activeCamera;
+                            scene.activeCamera = this._renderCamera;
```
Popov72, line 213: You should update `activeCamera` using `_activeCamera` instead, else the `onActiveCameraChanged` notification will be fired.

src/Rendering/utilityLayerRenderer.ts
```diff
@@ -200,9 +200,26 @@ export class UtilityLayerRenderer implements IDisposable {
...
+                            prePointerInfo.ray = null;
+                        }
                         scenePick = prePointerInfo.ray
                             ? scene.pickWithRay(prePointerInfo.ray)
                             : scene.pick(originalScene.pointerX, originalScene.pointerY);
+                        if (previousActiveCamera)
+                        {
+                            scene.activeCamera = previousActiveCamera;
```
Popov72, line 221: Same thing than above.

Popov72 (approved):

sebavan (approved):

PR #10653, opened 2021-07-08
[GE] Be able to copy GUI controls

msDestiny14:
GUIs can now be copied using the "Copy Element" button.

### DIFF
--- a/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
+++ b/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
@@ -293,7 +293,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
                                 if (newControl) { //insert the new control into the adt
                                     this.props.globalState.workbench.appendBlock(newControl);
                                     let index = 1;
-                                    while (this.props.globalState.workbench.nodes.filter(  //search if thier are any copies
+                                    while (this.props.globalState.workbench.nodes.filter(  //search if there are any copies
                                         control => control.name === newControl.name).length > 1) {
                                         newControl.name = `${this.state.currentNode.name} Copy ${index++}`;
                                     }
--- a/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
+++ b/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
@@ -274,7 +274,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
                         <div id="title">{`${this.state.currentNode.name} [${this.state.currentNode.getClassName()}] (ID: ${this.state.currentNode.uniqueId.toString()})`}</div>
                     </div>
                     {this.renderProperties()}
-                   <hr/>
+                    <hr />
                     <ButtonLineComponent
                         label="REMOVE ELEMENT"
                         onClick={() => {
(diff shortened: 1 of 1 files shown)


guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -274,14 +274,27 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
                             this.state.currentNode?.dispose();
                             this.props.globalState.onSelectionChangedObservable.notifyObservers(null);
                         }}
                     />
+                    <ButtonLineComponent
+                        label="COPY ELEMENT"
+                        onClick={() => {
+                            let serializationObject = {};
+                            this.state.currentNode?.serialize(serializationObject);
+                            let newControl = Control.Parse(serializationObject, this.props.globalState.guiTexture);
```
sebavan, line 290: should it not be aborted if this.state.currentNode is null or undefined ?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -274,14 +274,27 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
                         label="REMOVE ELEMENT"
                         onClick={() => {
                             this.state.currentNode?.dispose();
                             this.props.globalState.onSelectionChangedObservable.notifyObservers(null);
                         }}
                     />
+                    <ButtonLineComponent
+                        label="COPY ELEMENT"
+                        onClick={() => {
+                            let serializationObject = {};
```
sebavan, line 288: [nit] let -> const

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
sebavan: [nit] let -> const

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -274,14 +274,27 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+                            this.state.currentNode?.serialize(serializationObject);
+                            let newControl = Control.Parse(serializationObject, this.props.globalState.guiTexture);
+
+                            if (newControl) { //insert the new control into the adt
+                                this.props.globalState.workbench.appendBlock(newControl);
```
sebavan, line 293: Should there be any special renaming ???

sebavan (commented):
LGTM minor comments only

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: I thought about this but I let @PatrickRyanMS decide what would be best :)

PatrickRyanMS: I think we should do something to indicate a copied control by adding " copy" to the end of the control name. The user can rename to whatever they want for the final name, but in the interim, it feels like we should delineate any copies as such to prevent confusion about which was the original when looking solely at the tree view. Continued copies should increment like " copy 1", " copy 2" etc. so that they all have unique names. This way, if I make three copies of a button, I can easily identify them to move to a final location and receive a new name.

Popov72 (approved):

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -285,13 +285,20 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+                                this.state.currentNode.serialize(serializationObject);
+                                const newControl = Control.Parse(serializationObject, this.props.globalState.guiTexture);
 
-                            if (newControl) { //insert the new control into the adt
-                                this.props.globalState.workbench.appendBlock(newControl);
-                                this.props.globalState.onSelectionChangedObservable.notifyObservers(newControl);
+                                if (newControl) { //insert the new control into the adt
+                                    this.props.globalState.workbench.appendBlock(newControl);
+                                    let index = 1;
+                                    while (this.props.globalState.workbench.nodes.filter(  //search if thier are any copies
```
Popov72, line 296: nit: there

PR #10654, opened 2021-07-08
Performance graph dynamic playhead

SahilTara:
This pull request adds the dynamic sizing of the play head!! So if we are in a panned state, the real-time data can fill the buffer of space we have.  As we get closer to the real-time data, the play head will seem to grow. As we move away, the play head will shrink and disappear. Related to #10566 

Here is the related video capture describing what is going on more in depth:

https://user-images.githubusercontent.com/32103099/125003510-59352900-e025-11eb-951e-e5f82b883c82.mp4


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -484,7 +571,7 @@ export class CanvasGraphService {
     private _drawPlayheadRegion(drawableArea: IGraphDrawableArea, scaleFactor: number) {
         const { _ctx: ctx } = this;
        
-        if (!ctx) {
+        if (!ctx || scaleFactor >= 0.95) {
```
sebavan, line 574: sounds like a magic number to me ?

sebavan (approved):
LGTM just one tiny comment.

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Good catch! I forgot that I do not carry a license in witchcraft and wizardry from Hogwarts. So, I have turned this magical number (and another) into a constant and added comment explaining what it is for in my latest commit. 😄

Issue #10032, opened 2021-03-09
navigationPlugin.getNavmeshData


unicomp21:
We need an awaitable version which runs on a web worker thread.  This would also allow for multiple navmesh bakes to run concurrently.

CedricGuillemet:
@unicomp21 A version using worker thread is available thanks to this PR : https://github.com/BabylonJS/Babylon.js/pull/10656
Let me know if it fits your need.

unicomp21:
perfect, thanks @CedricGuillemet !!!

PR #10656, opened 2021-07-09
navmesh building in a worker

CedricGuillemet:
This is the first PR for the navmesh + worker work. see https://github.com/BabylonJS/Babylon.js/issues/10032
navmesh building can use an optional web worker instead of blocking calls. 
Next PRs (after my vacation) will addressed few things I'd like to add:
- documentation
- playground example
- ability to use multiple navmesh instead of only a default one


deltakosh (approved):
ok so by default so far it will remain on the main thread right? 
I like this approach and after a few months of testing we could change the behavior

sebavan (approved):

PR #10659, opened 2021-07-12
Draft for Order independant transparency

CraigFeldspar:


src/Rendering/renderingGroup.ts
```diff
@@ -150,7 +150,11 @@ export class RenderingGroup {
         // Transparent
         if (this._transparentSubMeshes.length !== 0) {
             engine.setStencilBuffer(stencilState);
-            this._renderTransparent(this._transparentSubMeshes);
+            if (this._scene.useOrderIndependantTransparency) {
```
deltakosh, line 153: just a typo to fix everywhere else: it is independent (no a)

src/Rendering/depthPeelingRenderer.ts
```diff
@@ -0,0 +1,228 @@
...
+        if (!this._blendBackEffect.isReady() || !this._finalEffect.isReady()) {
+            return;
+        }
+
+        const DEPTH_CLEAR_VALUE = -99999.0;
+        const MIN_DEPTH = 0;
+        const MAX_DEPTH = 1;
+
+        // Clears
+        this._engine._bindUnboundFramebuffer(this._blendBackMrt.getInternalTexture()!._framebuffer as WebGLFramebuffer);
```
deltakosh, line 131: We should not get WebGL access in here (cc @Popov72 to confirm)

CraigFeldspar: yep this is really an early preview for discussion, all webgl calls will be added in thinEngine. In that case we will have a `MultiRenderTarget.setTexture(index: number)` to perform the framebuffer layout

Popov72:
One thing that should not be forgotten is testing with `engine.useReverseDepthBuffer = true` as you will likely need to invert some comparison for it to work as expected.

Popov72:
When calling `this._engine.enableEffect`, it's important to pass a `DrawWrapper` and not an effect, else it won't work in WebGPU. Basically, where you have an effect, you should have a `DrawWrapper` instead.

As you are using `EffectWrapper`, it's simply a matter of passing `effectWrapper._drawWrapper` instead of `effectWrapper.effect` to `engine.enableEffect`.

src/Materials/Textures/multiRenderTarget.ts
```diff
@@ -66,6 +66,7 @@ export class MultiRenderTarget extends RenderTargetTexture {
     private _multiRenderTargetOptions: IMultiRenderTargetOptions;
     private _count: number;
     private _drawOnlyOnFirstAttachmentByDefault: boolean;
+    private _framebuffer: WebGLFramebuffer;
```
Popov72, line 69: We can't have WebGL specific objects in such high level class.

src/Materials/Textures/multiRenderTarget.ts
```diff
@@ -261,14 +263,21 @@ export class MultiRenderTarget extends RenderTargetTexture {
...
         this._count = this._internalTextures.length;
-
         // TODO : update types and samplingModes
     }
 
+    /**
+     * Retrieves the underlying framebuffer associated with this texture
+     * @returns The framebuffer
+     */
+    public _getFrameBuffer(): WebGLFramebuffer {
```
Popov72, line 277: Same problem than above.

Popov72 (commented):
I think we have a global problem with abstracting framebuffers, in the current state the PR can't work in WebGPU...

src/Materials/Textures/multiRenderTarget.ts
```diff
@@ -241,17 +241,6 @@ export class MultiRenderTarget extends RenderTargetTexture {
...
-        //     // So we keep references
-        //     texture._framebuffer = this._texture!._framebuffer;
-        //     // Prevents the dispose of the framebuffer
-        //     this._texture!._framebuffer = null;
-        //     texture._MSAAFramebuffer = this._texture!._MSAAFramebuffer;
-        //     this._texture!._MSAAFramebuffer = null;
-        //     this._texture = texture;
-        // }
-
         if (disposePrevious && this._internalTextures[index] && index !== 0) {
```
Popov72, line 244: What about having a method like this in `engine.multiRender.ts`:
```typescript
ThinEngine.prototype.updateMultipleRenderTargetTexture = function(textures: InternalTexture[], texture: InternalTexture, index: number, attachmentIndex: number = 0, disposePrevious: boolean = true): void {
    if (disposePrevious && textures[index] && index !== 0) {
        textures[index].dispose();    
    }
    
    textures[index] = texture;

    this.bindTextureFramebuffer(textures[0]._framebuffer!, texture, attachmentIndex);
}
```
Then:
```typescript
public setInternalTexture(texture: InternalTexture, index: number, attachmentIndex: number = 0, disposePrevious: boolean = true) {
    this._engine.updateMultipleRenderTargetTexture (this._internalTextures, texture, attachmentIndex, disposePrevious);

    if (!this.textures[index]) {
        this.textures[index] = new Texture(null, this.getScene());
    }
    this.textures[index]._texture = this._internalTextures[index];

    this._count = this._internalTextures.length;
    // TODO : update types and samplingModes
}
```
That way you can remove the `_frameBuffer` data member, as well as `_getFrameBuffer()`.

Also, why do we need an `attachmentIndex` that would be different from `index`? We can't simply use `index` instead of `attachmentIndex`?

src/Engines/constants.ts
```diff
@@ -57,6 +57,11 @@ export class Constants {
      * Alpha will be set to DST ALPHA
      */
     public static readonly ALPHA_EXCLUSION = 16;
+    /**
+     * Defines that alpha blending is SRC * SRC ALPHA + DST * (1 - SRC ALPHA)
+     * Alpha will be set to SRC ALPHA + (1 - SRC ALPHA) * DST ALPHA
```
MiiBond, line 62: I'm not sure of the intent of this blend mode so please ignore these comments if they're out of place:

Alpha accumulation should use "over" blending if blending SRC over (i.e. in front) of DST. This is just the current `ALPHA_COMBINE`.
Alpha accumulation should use "under" blending if blending SRC under (i.e. behind) of DST. This would be `SRC * (1 - DST_ALPHA) + DST * (DST_ALPHA)`

Additionally, this alpha blend equation looks odd to me since each layer is increasing the alpha of the destination by more than SRC_ALPHA.

CraigFeldspar: Actually code comes from [this example](https://github.com/tsherif/webgl2examples/blob/master/oit-dual-depth-peeling.html) but I gave it a further look because it seemed odd to me too that we don't use DST ALPHA in the color part of the blend.

So it turns out this is for back blending. Front blending is done manually in the shader [here](https://github.com/BabylonJS/Babylon.js/blob/b54bb0366da18e2c6c3dc3b8ac65f6af189b9b02/src/Shaders/default.fragment.fx#L468). And the alpha accumulation of the value only makes sense at the very end, when front peels and back peels meet to perform the final blending. At this moment we need the total alpha accumulation of the back peels to perform one final blending of front over back ([here](https://github.com/BabylonJS/Babylon.js/blob/b54bb0366da18e2c6c3dc3b8ac65f6af189b9b02/src/Shaders/oitFinal.fragment.fx#L12)).

MiiBond: Ah, thanks for the explanation. The results look good, so 👍

src/Engines/WebGL/webGLRenderTargetWrapper.ts
```diff
@@ -47,6 +47,63 @@ export class WebGLRenderTargetWrapper extends RenderTargetWrapper {
         this._framebuffer = this._depthStencilBuffer = this._MSAAFramebuffer = this._colorTextureArray = this._depthStencilTextureArray = null;
     }
 
+    /**
+     * Shares the depth buffer of this render target with another render target.
+     * *
+     * @param renderTarget Destination renderTarget
```
Popov72, line 53: nit: @hidden is missing (or rename `_shareDepth` to `shareDepth`)

src/Engines/WebGL/webGLRenderTargetWrapper.ts
```diff
@@ -47,6 +47,63 @@ export class WebGLRenderTargetWrapper extends RenderTargetWrapper {
...
+    public _shareDepth(renderTarget: WebGLRenderTargetWrapper): void {
+        super._shareDepth(renderTarget);
+
+        const gl = this._context;
+        const depthbuffer = this._depthStencilBuffer;
+        const framebuffer = renderTarget._framebuffer;
+
+        if (renderTarget._depthStencilBuffer) {
+            gl.deleteRenderbuffer(renderTarget._depthStencilBuffer);
+            renderTarget._depthStencilBuffer = this._depthStencilBuffer;
```
Popov72, line 64: I think the line above should be outside the `if`?

src/Engines/thinEngine.ts
```diff
@@ -3406,13 +3507,132 @@ export class ThinEngine {
...
-            processFunction: (width: number, height: number, img: HTMLImageElement | ImageBitmap | { width: number, height: number }, extension: string, texture: InternalTexture, continuationCallback: () => void) => boolean, samplingMode: number) => void,
-        prepareTextureProcessFunction: (width: number, height: number, img: HTMLImageElement | ImageBitmap | { width: number, height: number }, extension: string, texture: InternalTexture, continuationCallback: () => void) => boolean,
-        buffer: Nullable<string | ArrayBuffer | ArrayBufferView | HTMLImageElement | Blob | ImageBitmap> = null, fallback: Nullable<InternalTexture> = null, format: Nullable<number> = null,
-        forcedExtension: Nullable<string> = null, mimeType?: string, loaderOptions?: any, useSRGBBuffer?: boolean): InternalTexture {
+    /**
+     * Creates an internal texture without binding it to a framebuffer
+     * @param size defines the size of the texture
+     * @param options defines the options used to create the texture
+     * @hidden
+     * @returns a new render target texture stored in an InternalTexture
```
Popov72, line 3515: nit: @hidden is missing

CraigFeldspar:
Hey @MiiBond, if I add an API that would be kinda like : 

```
this.depthPeelingRenderer.onBeforeRenderPass((passNb: number) ⇒ void);
this.depthPeelingRenderer.onAfterRenderPass((passNb: number) ⇒ void);
this.depthPeelingRenderer.render(submeshes: SubMesh[]);
this.depthPeelingRenderer.setRenderTexture(tex: InternalTexture, id : DEPTH_TEXTURE | FRONT_COLOR_TEXTURE, BACK_COLOR_TEXTURE)
this.depthPeelingRenderer.getRenderTexture(id: DEPTH_TEXTURE | FRONT_COLOR_TEXTURE, BACK_COLOR_TEXTURE);
```

would it suit you for your usage regarding #10720 ?
Thanks!

MiiBond:
@CraigFeldspar , it might. I'm not sure. What is the result after each render pass? Two new layers of transparency are rendered (front and back)?
If I can use the depth textures, after each pass, to do some custom rendering, that would work, I think. But it wouldn't be terribly efficient since I'd be doing twice the number of passes needed.

If there was an option to render with custom material properties as well as save out separate colour buffers for each layer, that would be more ideal.

CraigFeldspar:
@MiiBond after each render pass, you get 3 textures : one with the front and back depth, each respectively encoded in the red and green channel, one with the front color, and one with the back color. The idea behind this API is to retrieve them all 3 with 3 calls to `getRenderTexture` in the `onAfterRenderPass` observer.
You can also specify your own textures, with `setRenderTexture`, that you would probably call in the `onBeforeRenderPass` observer.

To be able to do more in each render pass, you would probably need to add your own code in the PBR/Standard shader, right ?

MiiBond:
Will I be able to skip the over/under alpha blending to the two colour maps and replace that with custom rendering (via changes to the pbr material) to a set of MRT's? Ultimately, that's what I need.
I just want to make sure that we're not locked in to the traditional over/under alpha blending that's usually used with dual depth peeling (and most OIT techniques).

CraigFeldspar:
@MiiBond I guess I can add a boolean to skip the blend/final compose, but you will still have no control over what's in the shader. It will render just the PBR back and front colors in 2 textures, and the depth encoded as stated above in another.
@sebavan @Popov72, should we add a "#CUSTOM_OIT" in the shaderInclude that would be replaced by the shader parser ?

Popov72:
I think there's still the need to test in reverse depth buffer mode (when setting `engine.useReverseDepthBuffer = true`)?

You probably will need to test for the `USE_REVERSE_DEPTHBUFFER` define in the shader code.

src/Rendering/depthPeelingRenderer.ts
```diff
@@ -0,0 +1,368 @@
...
+
+        const DEPTH_CLEAR_VALUE = -99999.0;
+        const MIN_DEPTH = 0;
+        const MAX_DEPTH = 1;
+
+        (this._scene.prePassRenderer! as any)._enabled = false;
+
+        // Clears
+        this._engine.bindFramebuffer(this._depthMrts[0].renderTarget!);
+        let attachments = this._engine.buildTextureLayout([true]);
```
sebavan, line 277: Could we keep all the arrays here in cache to prevent GC

src/Rendering/depthPeelingRenderer.ts
```diff
@@ -0,0 +1,368 @@
...
+        this._engine.bindAttachments(attachments);
+        this._engine.clear(new Color4(DEPTH_CLEAR_VALUE, DEPTH_CLEAR_VALUE, 0, 0), true, false, false);
+
+        this._engine.bindFramebuffer(this._depthMrts[1].renderTarget!);
+        this._engine.clear(new Color4(-MIN_DEPTH, MAX_DEPTH, 0, 0), true, false, false);
+
+        this._engine.bindFramebuffer(this._colorMrts[0].renderTarget!);
+        attachments = this._engine.buildTextureLayout([true, true]);
+        this._engine.bindAttachments(attachments);
+        this._engine.clear(new Color4(0, 0, 0, 0), true, false, false);
```
sebavan, line 287: Same for colors, let s try to have no GC

src/Shaders/pbr.fragment.fx
```diff
@@ -587,6 +590,10 @@ void main(void) {
     gl_FragData[PREPASS_VELOCITY_INDEX] = vec4(velocity, 0.0, writeGeometryInfo);
     #endif
 
+    #ifdef PREPASS_ALBEDO
+        vec3 sqAlbedo = sqrt(surfaceAlbedo); // for pre and post scatter
```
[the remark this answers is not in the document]
CraigFeldspar: simply a bugfix, old L598 should be under a #define

src/Shaders/default.fragment.fx
```diff
@@ -456,8 +459,17 @@ color.rgb = max(color.rgb, 0.);
...
 
-#if !defined(PREPASS) || defined(WEBGL2) 
+#if !defined(PREPASS) || defined(WEBGL2)
 	gl_FragColor = color;
 #endif
 
+#if ORDER_INDEPENDENT_TRANSPARENCY
+	if (fragDepth == nearestDepth) {
+		frontColor.rgb += color.rgb * color.a * alphaMultiplier;
+		frontColor.a = 1.0 - alphaMultiplier * (1.0 - color.a);
```
sebavan, line 469: I am having trouble to match this and the equations from under blending in https://developer.download.nvidia.com/SDK/10.5/opengl/src/dual_depth_peeling/doc/DualDepthPeeling.pdf Could you explain it ?

I have the same comment for the rest of the blending equations in use. Would be great to better understand the setup on my side :-)

src/Engines/Extensions/engine.alpha.ts
```diff
@@ -118,6 +118,10 @@ ThinEngine.prototype.setAlphaMode = function (mode: number, noDepthWriteChange:
             this._alphaState.setAlphaBlendFunctionParameters(this._gl.ONE_MINUS_DST_COLOR, this._gl.ONE_MINUS_SRC_COLOR, this._gl.ZERO, this._gl.ONE);
             this._alphaState.alphaBlend = true;
             break;
+        case Constants.ALPHA_LAYER_ACCUMULATE:
+            this._alphaState.setAlphaBlendFunctionParameters(this._gl.SRC_ALPHA, this._gl.ONE_MINUS_SRC_ALPHA, this._gl.ONE, this._gl.ONE_MINUS_SRC_ALPHA);
```
sebavan, line 122: Same comment as some others, would be great to know where this setups comes from :-)

src/Materials/Textures/textureCreationOptions.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Define options used to create an internal texture
+ */
+export class InternalTextureCreationOptions {
```
sebavan, line 4: Should they not all be interfaces only ?

sebavan (commented):
LGTM aside of two tiny comments which are more for my understanding.

PR #10666, opened 2021-07-13
Add tooltip to the performance graph on hover.

SahilTara:
This PR adds the necessary functionality to draw a tooltip over the data in the graph. This will allow the user to simply hover over a point of interest and the tooltip will show the value at that location for all data sources.

Completes the "Allow for hovering over data points and displaying the actual values." of #10566 

Video of what this PR adds:

https://user-images.githubusercontent.com/32103099/125531626-850aedd6-6c1d-4712-a9ec-d94c18cee50d.mp4

### DIFF
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -19,6 +19,9 @@ const spaceBetweenTextAndBox = 5;
 const playheadSize = 8;
 const dividerSize = 2;
 
+const axisLineLength = 10;
+const axisPadding = 10;
+
 // Currently the scale factor is a constant but when we add panning this may become formula based.
 const scaleFactor = 0.8;
 
@@ -51,6 +54,7 @@ export class CanvasGraphService {
     private _globalTimeMinMax: IPerfMinMax;
     private _hoverPosition: number | null;
     private _drawableArea: IGraphDrawableArea;
+    private _axisHeight: number;
     
     private readonly _tooltipLineHeight;
     private readonly _defaultLineHeight;
@@ -78,6 +82,7 @@ export class CanvasGraphService {
         if (this._ctx) {
             const defaultMetrics = this._ctx.measureText(alphabet);
             this._defaultLineHeight = defaultMetrics.actualBoundingBoxAscent + defaultMetrics.actualBoundingBoxDescent;
+            this._axisHeight = axisLineLength + axisPadding + this._defaultLineHeight + axisPadding;
 
             this._ctx.save();
             this._ctx.font = tooltipFont;
@@ -181,7 +186,7 @@ export class CanvasGraphService {
             this._globalTimeMinMax.max = Math.max(dataset.data[bounds.end - 1].timestamp, this._globalTimeMinMax.max);
         });
 
-        let updatedScaleFactor = Scalar.Clamp((this._globalTimeMinMax.max - this._globalTimeMinMax.min)/(bufferMaximum - this._globalTimeMinMax.min), 0.8, 1); 
+        const updatedScaleFactor = Scalar.Clamp((this._globalTimeMinMax.max - this._globalTimeMinMax.min)/(bufferMaximum - this._globalTimeMinMax.min), scaleFactor, 1); 
 
         // we will now set the global maximum to the maximum of the buffer.
         this._globalTimeMinMax.max = bufferMaximum;
@@ -281,10 +286,8 @@ export class CanvasGraphService {
 
         this._generateTicks(timeMinMax, spaceAvailable);
 
-        const axisHeight = this._defaultLineHeight + 30;
-
         // remove the height of the axis from the available drawable area.
-        drawableArea.bottom -= axisHeight;
+        drawableArea.bottom -= this._axisHeight;
 
         // draw time axis line
         ctx.save();
@@ -507,7 +510,7 @@ export class CanvasGraphService {
         const inferredTimestamp = this._getNumberFromPixel(pixel, this._globalTimeMinMax, start, end);
         
         const results: IPerfTooltip[] = [];
-
+        let longestText: string = "";
         // get the closest timestamps to the target timestamp, and store the appropriate meta object.
         this.datasets.forEach((dataset: IPerfDataset) => {
             if (!!dataset.hidden || dataset.data.length === 0) {
@@ -515,20 +518,17 @@ export class CanvasGraphService {
             }
 
             const closestIndex = this._getClosestPointToTimestamp(dataset, inferredTimestamp);
+            const text = `${dataset.id}: ${dataset.data[closestIndex].value.toFixed(2)}`;
             
+            if (text.length > longestText.length) {
+                longestText = text;
+            }
+
             results.push({
-                            text: `${dataset.id}: ${dataset.data[closestIndex].value.toFixed(2)}`,
+                            text,
                             color: dataset.color ?? defaultColor,
                         });
-        });
-
-        const longestText = results.reduce((longestSoFar: string, result) => {
-            if (longestSoFar.length < result.text.length) {
-                return result.text;
-            } else {
-                return longestSoFar;
-            }
-        }, "");
+        }); 
 
         let x = pixel - start;
         let y = Math.floor((drawableArea.bottom - drawableArea.top)/2);
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -19,6 +19,13 @@ const stopDrawingPlayheadThreshold = 0.95;
 // Threshold for the ratio at which we go from panning mode to live mode.
 const returnToLiveThreshold = 0.998;
 
+// Font to use on the tooltip!
+const tooltipFont = "12px Arial";
+
+// A string containing the alphabet, used in line height calculation for the font.
+const alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+
 /**
  * This class acts as the main API for graphing given a Here is where you will find methods to let the service know new data needs to be drawn,
  * let it know something has been resized, etc! 
@@ -32,6 +39,10 @@ export class CanvasGraphService {
     private _panPosition: IPerfMousePanningPosition | null;
     private _positions: Map<string, number>;
     private _datasetBounds: Map<string, IPerfIndexBounds>;
+    private _globalTimeMinMax: IPerfMinMax;
+    
+    private readonly _tooltipLineHeight;
+    private readonly _defaultLineHeight;
 
     public readonly datasets: IPerfDataset[];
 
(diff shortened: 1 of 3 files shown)


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -47,8 +69,25 @@ export class CanvasGraphService {
...
+        if (this._ctx) {
+            const defaultMetrics = this._ctx.measureText(alphabet);
+            this._defaultLineHeight = defaultMetrics.actualBoundingBoxAscent + defaultMetrics.actualBoundingBoxDescent;
+
+            this._ctx.save();
+            this._ctx.font = tooltipFont;
+            const fontMetrics = this._ctx.measureText(alphabet);
+            this._tooltipLineHeight = fontMetrics.actualBoundingBoxAscent + fontMetrics.actualBoundingBoxDescent;
+            this._ctx.restore();
+        } else {
```
sebavan, line 87: Should we not completely fail the display if no canvas/context are available ?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -138,13 +178,13 @@ export class CanvasGraphService {
             bounds.end = this._getClosestPointToTimestamp(dataset, bufferMaximum) + 1;
 
             // keep track of largest timestamp value in view!
-            globalTimeMinMax.max = Math.max(dataset.data[bounds.end - 1].timestamp, globalTimeMinMax.max);
+            this._globalTimeMinMax.max = Math.max(dataset.data[bounds.end - 1].timestamp, this._globalTimeMinMax.max);
         });
 
-        let updatedScaleFactor = Scalar.Clamp((globalTimeMinMax.max - globalTimeMinMax.min)/(bufferMaximum - globalTimeMinMax.min), 0.8, 1); 
+        let updatedScaleFactor = Scalar.Clamp((this._globalTimeMinMax.max - this._globalTimeMinMax.min)/(bufferMaximum - this._globalTimeMinMax.min), 0.8, 1);
```
sebavan, line 184: Could we move the 0.8 as a constant ?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -240,7 +281,7 @@ export class CanvasGraphService {
 
         this._generateTicks(timeMinMax, spaceAvailable);
 
-        const axisHeight = 100;
+        const axisHeight = this._defaultLineHeight + 30;
```
sebavan, line 284: Would be great to have 30 in a const so at least by looking at this line we know what is happening ? Even better as it is almost "static" you could precompute axisHeight when you check for _defaultLineHeight ? (premature optim but there are no small gains :-) )

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -416,9 +459,131 @@ export class CanvasGraphService {
...
+     * 
+     * @param event A reference to the event to be handled. 
+     */
+    private _handleDataHover = (event: MouseEvent) => {
+        if (this._panPosition) {
+            // we don't want to do anything if we are in the middle of panning
+            return;
+        }
+
+        this._hoverPosition = event.clientX;
```
sebavan, line 479: Should this not account for the canvas not being fully at 0 on the left ? (It might already be the case but just guessing)

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -416,9 +459,131 @@ export class CanvasGraphService {
...
+    }
+
+    /**
+     * Draws the tooltip given the area it is allowed to draw in and the current pixel position.
+     * 
+     * @param pixel the position of the mouse cursor in pixels. 
+     * @param drawableArea  the available area we can draw in.
+     */
+    private _drawTooltip(pixel: number | null, drawableArea: IGraphDrawableArea) {
+        const {_ctx : ctx} = this;
```
sebavan, line 499: I wonder how slow the function is and if it should debounce a bit looking into the rest of the code ?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -416,9 +459,131 @@ export class CanvasGraphService {
...
+
+            const closestIndex = this._getClosestPointToTimestamp(dataset, inferredTimestamp);
+            
+            results.push({
+                            text: `${dataset.id}: ${dataset.data[closestIndex].value.toFixed(2)}`,
+                            color: dataset.color ?? defaultColor,
+                        });
+        });
+
+        const longestText = results.reduce((longestSoFar: string, result) => {
```
sebavan, line 525: This goes over all the results again, why not doing it in the previous foreach ?

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -416,9 +459,131 @@ export class CanvasGraphService {
...
+        }
+
+        // first convert the mouse position in pixels to a timestamp.
+        const {left: start, right: end} = ctx.canvas.getBoundingClientRect();
+        const inferredTimestamp = this._getNumberFromPixel(pixel, this._globalTimeMinMax, start, end);
+        
+        const results: IPerfTooltip[] = [];
+
+        // get the closest timestamps to the target timestamp, and store the appropriate meta object.
+        this.datasets.forEach((dataset: IPerfDataset) => {
```
sebavan, line 512: Same as we discussed last time, This should try to be garbage collection friendly

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -416,9 +459,131 @@ export class CanvasGraphService {
...
+        ctx.save();
+        
+        ctx.font = tooltipFont;
+        ctx.textBaseline = "middle";
+        ctx.textAlign = "left";
+
+        const boxLength = this._tooltipLineHeight;
+        const textHeight = this._tooltipLineHeight + Math.floor(tooltipHorizontalPadding/2);
+
+        const width = ctx.measureText(longestText).width + boxLength + 2 * tooltipHorizontalPadding + spaceBetweenTextAndBox;
```
sebavan, line 545: I wonder how measureText speed is but depending on your findings, might be good to cache ?

sebavan (changes requested):
LGTM, mostly nits and also being mindful of perf mostly in the move handler

inspector/src/components/graph/canvasGraphService.ts
SahilTara: it does already!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Hmm I'll see if i can do that!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Yep! its basically just bound between the original scaleFactor of 0.8 and 1 so i will update that!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Yeah I'm actually kind of confused on how we would do that, do you have any suggestions?

sebavan: May be throwing an error in the constructor ? or just returning ?

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Would 2 arrays work in this case instead of a tooltip array we have 2 arrays of strings?

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts

inspector/src/components/graph/canvasGraphService.ts
sebavan: Even if you keep only one array you could reuse it from move to move instead of recreating it on every events

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Seems to be ~0.13ms: I could possibly debounce though if you feel it is necessary

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Ahh okay so its the array creation that is the main issue here rather than the object creation in the loop? So doing something similar to what I do with the ticks array (instance variable then .length = 0 in the event) should be fine then?

inspector/src/components/graph/canvasGraphService.ts
SahilTara: I throw an error now!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: So the performance of measureText depends entirely on the length of the text: https://jsbench.me/d4kr47hmda/1

If you look at the "Small" section it seems to be very very fast (for me it is 0.01ms or 72126.07 ops/s), and around the length of what our text would be at largest for our predefined tags. 

Although the performance does degrade pretty hard for the "large" section as compared to the small, because of this I could do some form of caching just in case, if we follow the no gain is too small rule!

inspector/src/components/graph/canvasGraphService.ts
sebavan: you might want to keep the allocated array around with the max potential size with the objects in and keep a local length to know the meaningful part of the data. Then in here you could update the length and data content for the meaningful part. This would prevent allocating and such for the array as well as the GC involved by the creation of the new objects.

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Latest commit caches this!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Latest commit pre-creates the objects in the constructor, and we loop over and define the meaningful part in the event. We simply update the objects that have been pre-allocated now!

inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -47,8 +77,33 @@ export class CanvasGraphService {
...
         this._datasetBounds = new Map<string, IPerfIndexBounds>();
+        this._globalTimeMinMax = {min: Infinity, max: 0};
+        this._drawableArea = {top: 0, left: 0, right: 0, bottom: 0};
+        this._textCache = {text: "", width: 0};
+        this._tooltipItems = [];
+    
+        for (let i = 0; i < maximumDatasetsAllowed; i++) {
+            this._tooltipItems.push({text: "", color: ""});
+        }
+
```
sebavan, line 91: looks like a lot of spaces :-)

sebavan (approved):
Only one tiny fix and you ll be good to go

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Okay so I debounced the drawing part inside of the event handler I noticed that the actual event itself needed to update the mouse position whenever, but the expensive bit inside the mousemove has been debounced. This makes the mouse move event a lot less performance heavy!

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Fixed!

PR #10667, opened 2021-07-14
Fix getPositionData issue

deltakosh:


src/Meshes/abstractMesh.ts
```diff
@@ -1305,11 +1306,19 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
+        let needUpdatePositionInSkeletonPass = false;
+        let data = this.getVerticesData(VertexBuffer.PositionKind);
 
         if (data && ((applySkeleton && this.skeleton) || (applyMorph && this.morphTargetManager))) {
             data = Tools.Slice(data);
             this._generatePointsArray();
+            needUpdatePositionInSkeletonPass = !!(applySkeleton && this.skeleton) && !(applyMorph && this.morphTargetManager);
+
+            if (this._positions) {
+                this._internalAbstractMeshDataInfo._positions = this._positions.slice();
```
Popov72, line 1318: Maybe use `Tools.Slice(this._positions)` instead (as above line 1313)?

src/Meshes/thinInstanceMesh.ts
```diff
@@ -305,7 +305,7 @@ Mesh.prototype.thinInstanceRefreshBoundingInfo = function (forceRefreshParentInf
 
     if (forceRefreshParentInfo) {
         vectors.length = 0;
-        this.refreshBoundingInfo(true);
+        this.refreshBoundingInfo();
```
Popov72, line 308: So, if the mesh has a skeleton, we don't take it into account anymore to compute the bounding box? Is it for a performance reason? Maybe we can pass a flag to `thinInstanceRefreshBoudingInfo` that would make the call to `refreshBoundingInfo` pass `true, true` as parameters?

src/Meshes/abstractMesh.ts
```diff
@@ -1305,11 +1306,19 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
     /** @hidden */
     public _getPositionData(applySkeleton: boolean, applyMorph: boolean): Nullable<FloatArray> {
-        var data = this.getVerticesData(VertexBuffer.PositionKind);
+        let needUpdatePositionInSkeletonPass = false;
+        let data = this.getVerticesData(VertexBuffer.PositionKind);
 
         if (data && ((applySkeleton && this.skeleton) || (applyMorph && this.morphTargetManager))) {
             data = Tools.Slice(data);
             this._generatePointsArray();
+            needUpdatePositionInSkeletonPass = !!(applySkeleton && this.skeleton) && !(applyMorph && this.morphTargetManager);
```
Popov72, line 1315: Why not updating the vertices based on the skeleton if there are also some morph targets? The morph may not apply on all vertices, so it's possible some vertices are still updated by the bones even where there is both a skeleton and some morph targets?

Popov72 (approved):

src/Meshes/abstractMesh.ts
deltakosh: Good catch!

src/Meshes/thinInstanceMesh.ts
deltakosh: I agree we should have the same set of parameters so the user can decide if they want to take skeleton and morph in account

src/Meshes/abstractMesh.ts
deltakosh: yeah you are right I missed the part where morph could be on different targets. 
I wanted to save the copy but I believe we cannot

src/Meshes/abstractMesh.ts
```diff
@@ -1371,6 +1377,11 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
                         }
                     }
                 }
+
+                if (this._positions) {
+                    const index = vertexCount * 3;
+                    this._positions[vertexCount].copyFromFloats(data[index], data[index + 1], data[index + 2]);
```
Popov72, line 1383: Shouldn't we take the `if` branch only if we modified `data[vertexCount]` starting line 1373? Else, it will overwrite the changes that the skeleton code could have made.

deltakosh:
Fixing the tests issues (the 'this._positions[vertexCount]' is obviously wrong)

PR #10671, opened 2021-07-15
Additional MergeMeshes optimizations

ryantrem:
This change is a follow up for PR #10639. The notable changes are:

- `Geometry.getIndices` can optionally (depending on args) make a copy of the indices. However, it always converts the indices to an array, even if the original indices was a TypedArray. I changed the way the copy is made to just call `slice()`, which exists on regular arrays and TypedArrays and ensures the type of the copy matches the type of the source. I think in general we want this behavior, but in my case specifically I definitely need it to be a TypedArray so we can efficiently operate on the data from native code (see bullet points below).
- I added back the `allow32BitsIndices` argument to `VertextData.merge`, which I accidentally lost in my previous PR (#10639). This wasn't evident though because of the above bullet point (TypedArrays for indices were always being converted to regular arrays, and so the indices were never constrained to 16 bits).
- I added an optional `createBoundingBox` parameter to `SubMesh.CreateFromIndices` which is passed through to the `SubMesh` constructor so a caller can avoid creating the bounding box. I made this change because when the bounding box is created, it calls `ComputeWorldMatrix` on the parent `Mesh`. This meant if you are merging say 250 meshes (as in my example), `ComputeWorldMatrix` is called on the same parent `Mesh` 250 times, of which 249 are a waste (as far as I can tell), and relatively expensive. So now I manually make the calls to compute the bounds for the submeshes all together with just a single call to `ComputeWorldMatrix` for the parent mesh. This improves performance of `MergeMeshes` by an additional 33%-50% beyond the improvements I made in #10639.
- For Babylon Native, 90% of the time spent in `MergeMeshes` is in six functions that iterate over vertices or indices, which is quite slow when we don't have JavaScript JIT compilation (as is the case on iOS and all platforms when using React Native with Hermes, which basically means most if not all of the time). Fortunately these six functions are quite simple and do not change frequently, [so I duplicated them in C++ (in Babylon Native)](https://github.com/BabylonJS/BabylonNative/pull/829) and the Babylon.js code will defer to those native helpers when they are available. These native functions are about 500x faster than the JavaScript version when running without JavaScript JIT, which reduces the total time of `MergeMeshes` in Babylon Native by 10x-12x.
- Added a `@nativeOverride` decorator to introduce a simple pattern for (optionally) redirecting functions to native implementations when available.


sebavan (commented):
@ryantrem I love a lot where it is going :-)

I am simply wondering how could we make those changes less intrusive I mean not having to check on each call and such.

Maybe the functions could be like 
```
static foo = () => {};
```
in babylon and in native we could have equivalent math.native,ts which would overwrite them ?

ryantrem:
> I am simply wondering how could we make those changes less intrusive I mean not having to check on each call and such.
> 
> Maybe the functions could be like
> 
> ```
> static foo = () => {};
> ```
> 
> in babylon and in native we could have equivalent math.native,ts which would overwrite them ?

I think the challenge is with overriding those strictly from a NAPI context since they are defined in ES6 modules (not as globals). It's the same issue we ran into with this change: https://github.com/BabylonJS/Babylon.js/commit/73916d62f314b42ca1e7c3d02432c149c2469716 (which at the time, we never came up with a better solution).

I believe the only way we could do an override is to initiate it from the JavaScript side, which we could do in Babylon React Native, but there is not "automatic" or "default" JavaScript that is executed as part of Babylon Native itself.

sebavan:
Yup I was exactly speaking of doing it from the JS side at the beginning of NativeEngine.

Basically we could call a initMathsOptimizations() function from a maths.vector.native.ts  which would switch the functions in maths.vector.ts. The implementations in maths.vector.native.ts would actually be exactly what you did. Do think it would not work ?

sebavan (approved):

src/Meshes/mesh.vertexData.ts
```diff
@@ -401,62 +402,76 @@ export class VertexData {
...
+            Vector4.TransformNormalToRef(normal, transformation, transformedNormal);
+            normals[index] = transformedNormal.x;
+            normals[index + 1] = transformedNormal.y;
+            normals[index + 2] = transformedNormal.z;
+            normals[index + 3] = transformedNormal.w;
+        }
+    }
+
+    @nativeOverride.filter((indices => !Array.isArray(indices)))
+    private static _FlipIndices(indices: IndicesArray) {
```
bghgary, line 446: nit: this name is maybe better as FlipFaces or FlipSideOrientation

bghgary (approved):

src/Meshes/geometry.ts
```diff
@@ -584,12 +584,7 @@ export class Geometry implements IGetSetVerticesData {
         if (!forceCopy && (!copyWhenShared || this._meshes.length === 1)) {
             return orig;
         } else {
-            var len = orig.length;
-            var copy = [];
-            for (var i = 0; i < len; i++) {
-                copy.push(orig[i]);
-            }
-            return copy;
+            return orig.slice();
```
Popov72, line 587: Should we use `Tools.Slice` instead?

src/Meshes/mesh.vertexData.ts
```diff
@@ -401,62 +402,76 @@ export class VertexData {
...
+            normals[index] = transformedNormal.x;
+            normals[index + 1] = transformedNormal.y;
+            normals[index + 2] = transformedNormal.z;
+        }
+    }
+
+    @nativeOverride.filter((normals) => !Array.isArray(normals))
+    private static _TransformVector4Normals(normals: FloatArray, transformation: DeepImmutable<Matrix>) {
+        var normal = TmpVectors.Vector4[0];
+        var transformedNormal = TmpVectors.Vector4[1];
```
Popov72, line 434: nit: `const` for both lines

Popov72 (approved):

src/Meshes/geometry.ts
ryantrem: It looks to me like `Tools.Slice` is useful if you don't know whether the object is an array, but in this case we do know it is an array. Is there another benefit?

Popov72: Looking at the doc, it says it is for IE support...

ryantrem: Note to self: read doc comments.

PR #10675, opened 2021-07-15
Fix GLB export with characters that don't fit into single UTF-16 code point

### LINKED ISSUE #10672: Exporting to .glb when node names contain UTF-8 characters is broken
See this forum thread for an example model that is broken: https://forum.babylonjs.com/t/how-to-import-a-babylon-file-in-blender/22210. To repro, import the provided `.babylon file` and then export to .glb. The exported .glb will fail to validate as it will contain invalid characters.

This is because there are nodes in the scene with UTF-8 characters, which are currently not encoded correctly when we convert string names to binary in `_generateGLBAsync` from gltfExporter.ts. I will open a PR shortly to fix this.

darraghjburke:
Fixes #10672. Uses TextEncoder when available to encode text correctly when converting from JSON to binary. If not available, simply places "_" characters when the true character doesn't fit into a single code point (ie the value is greater than 0xFFFF). See this MDN article for more details: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt


sebavan (changes requested):
Why not simply replacing JsonData by either the TextEncoder.encodeResult or the _ replaced one ?

on line 1079

```
let jsonData: Uint8Array;
if (typeof TextEncoder !== "undefined") { 
const encoder = ....
jsonData = ... encoder.encode();
}
else ...
```

darraghjburke:
@sebavan the array needs to have a [`buffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/buffer) and [`byteOffset`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/byteOffset) property but these can only be set when the array is [constructed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/Uint8Array)

sebavan:
ohhh Is textEncode.encode() not returning a uint8array containing this ?

sebavan (approved):

PR #10676, opened 2021-07-15
adding autoclear to PP when unlinking prepassRT

CraigFeldspar:
Fixes https://forum.babylonjs.com/t/disabling-ssao2-breaks-fxaa/22391

### DIFF
--- a/src/Materials/Textures/prePassRenderTarget.ts
+++ b/src/Materials/Textures/prePassRenderTarget.ts
@@ -131,5 +131,10 @@ export class PrePassRenderTarget extends MultiRenderTarget {
         if (this.renderTargetTexture) {
             this.renderTargetTexture._prePassRenderTarget = null;
         }
+
+        if (this._outputPostProcess) {
+            this._outputPostProcess.autoClear = true;
+            this._outputPostProcess.restoreDefaultInputTexture();
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Materials/Textures/prePassRenderTarget.ts
```diff
@@ -131,5 +131,10 @@ export class PrePassRenderTarget extends MultiRenderTarget {
         if (this.renderTargetTexture) {
             this.renderTargetTexture._prePassRenderTarget = null;
         }
+
+        if (this._outputPostProcess) {
+            this._outputPostProcess.autoClear = true;
```
sebavan, line 136: I wonder how we could keep track of the previous setup ? if need be

sebavan (approved):
Just wondering how safe it is to infer the previous state to be true ? probably more common I agree but not sure it is always the case ?

src/Materials/Textures/prePassRenderTarget.ts
sebavan: @CraigFeldspar do you think we should implement the part to remember of the previous status ?

CraigFeldspar: @sebavan honestly, I think this property should not have public access (I mean it should be public but hidden). It is already overriden in [DefaultRenderingPipeline](https://github.com/BabylonJS/Babylon.js/blob/34f4eda75b58b96f026a7d33c2c10de410fc6df7/src/PostProcesses/RenderPipeline/Pipelines/defaultRenderingPipeline.ts#L499), regardless of the previous status. I don't see any user use-case for that, but if there might be any, we should duplicate the property to have one for internal use.

sebavan: Ok so let s merge it like this once @Popov72 will validate :-)

Popov72 (approved):

PR #10677, opened 2021-07-15
WebGPU: fix float texture usages

Popov72:
(32 bits) float textures are not filterable in WebGPU (at least at the current time), so we must take some specific actions to handle them.

Also in the PR:
* added some more information in the Texture pane of the inspector (texture type, format, normalized?, compressed?, useSRGBBuffers?, etc)
* set some friendly names to the textures created by the `MultiRenderTarget` class so that it is easier to see what's going on in the Texture pane of the inspector
* moved `samplingMode` getter to `ThinTexture`
* trapped an error in texture preview (inspector) when the texture has no internal texture
* moved half float helpers from dds to `TextureTools`

[UPDATE] Also added support for 3D textures, as Chrome Canary now does support them.


inspector/src/components/actionTabs/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx
```diff
@@ -172,6 +238,14 @@ export class TexturePropertyGridComponent extends React.Component<ITextureProper
             { label: "Spherical", value: Texture.SPHERICAL_MODE },
         ];
 
+        const format = texture._texture?.format ?? -2;
```
sebavan, line 241: -2 sounds weird ? maybe adding a comment ?

inspector/src/components/actionTabs/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx
```diff
@@ -172,6 +238,14 @@ export class TexturePropertyGridComponent extends React.Component<ITextureProper
...
+        const type = texture._texture?.type ?? -2;
+
+        const oformat = this.findTextureFormat(format === -1 ? Constants.TEXTUREFORMAT_RGBA : format);
```
sebavan, line 244: even more as the check after is for -1

inspector/src/components/actionTabs/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx
```diff
@@ -238,13 +312,28 @@ export class TexturePropertyGridComponent extends React.Component<ITextureProper
                             });
                         }} />
                     }
+                    <TextLineComponent label="Format" value={oformat?.label ?? "unknown"} />
+                    {
+                        !oformat?.hideType && !oformat?.compressed &&
+                        <TextLineComponent label="Type" value={otype?.label ?? "unknown"} />
+                    }
+                    {
+                        !!oformat?.normalizable && !oformat?.compressed && !!otype?.normalizable &&
```
sebavan, line 321: It gets pretty hard to read with so many signs :-)

tests/validation/config.json
```diff
@@ -406,7 +406,8 @@
             "title": "Self shadowing",
             "scriptToRun": "/Demos/SelfShadowing/shadows.js",
             "functionToCall": "CreateShadowsTestScene",
-            "referenceImage": "selfShadowing.png"
+            "referenceImage": "selfShadowing.png",
+            "excludedEngines": ["webgpu"]
```
sebavan, line 410: Was this one not ok before ?

sebavan (approved):
Only couple of minor comments, this is insane to have to add all this :-)

inspector/src/components/actionTabs/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx
Popov72: It would be even worst if I remove the ? and add some null/undefined checks :)

tests/validation/config.json
Popov72: It was before the support for linear filtering is gone for float textures.

Now that test does not pass anymore because the ESM technic depends on the format of the texture (which is normally 32 bits float but is downscaled to 16 bits float in WebGPU).

sebavan (approved):

sebavan (approved):

PR #10689, opened 2021-07-19
Add sidebar to the performance graph

SahilTara:
This PR addresses adding a sidebar to the performance viewer. This sidebar allows the user to toggle what data sources are visible, and the color for each data source. #10566 

Currently we store the entire dataset as both metadata and data are merged into one object right now. This state storage will be a lot more light weight once we separate the objects as we will only store the metadata in the sidebar.

A change I had to make in an established component is a setting to specify if the colour picker should pop right or not.  This was done with the addition of an optional boolean variable. If it is unset, we pop left (which is the current default behaviour). If it is set to true we don't pop left. If false we also pop left which is default behaviour. Overall this should not have any behaviour changes to existing components.

Here is the video showing the behaviour:

https://user-images.githubusercontent.com/32103099/126194702-a6d1130d-7b9c-4e6e-bd60-df50ff68091a.mp4


sebavan (commented):
LGTM, adding @Popov72 for a quick review :-)

sebavan (approved):

Issue #10660, opened 2021-07-12
[GE] CSS Fixes


msDestiny14:
The following components have incorrect CSS
- [x] Slider 
- [x] Checkbox 
- [x] GUI alignments property

PR #10690, opened 2021-07-19
[GE] CSS and design fixes

msDestiny14:
https://github.com/BabylonJS/Babylon.js/issues/10660


sebavan (changes requested):
TS6133: 'OptionsLineComponent' is declared but its value is never read.

It does not build yet but the rest looks OK

PR #10693, opened 2021-07-19
Make window resizable for performance viewer 

SahilTara:
This PR makes it so that the performance viewer graph takes up the remaining window size, and allows the size to dynamically adjust to the window size (so resizing the graph will allow you to have a bigger or smaller graph!). #10566 

Here is a gif for what this feature adds (as video glitches out for some reason):
LiveDataResize


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -126,6 +126,23 @@ export class CanvasGraphService {
...
+            return;
+        }
+
+        this._width = width;
+        this._height = height;
+
+        ctx.canvas.width = width;
+        ctx.canvas.height = height;
+
+        this.draw();
```
SahilTara, line 143: Question: Should I be debouncing draw() inside this method? As it is expensive and resize can be called a lot LOL

sebavan (approved):

inspector/src/components/graph/canvasGraphService.ts
sebavan: Ohhhhh Yup good catch, I guess draw should be debouncing ??? so that no matter if called 100 times in a frame, it would handle it gracefully ?

SahilTara: Cool latest commit sets up the necessary steps debouncing the draw all the time! it will be called update, and eventually draw will become private (makes more sense conceptually too because external things shouldn't be telling the service when to draw). Going to see if I can make it private at this moment or not because performanceViewerComponent seems to have "draw" right now. Left a to do later note for myself though!

SahilTara: Okay replaced draw with _draw, and now we use update() everywhere instead 😄

sebavan (approved):

PR #10695, opened 2021-07-19
[GE] Parenting bug fixes

msDestiny14:
Bug fix for duplicated controls in the tree view.
Bug fix for found memory leaks
Bug fix for when inserting into child node

### DIFF
--- a/guiEditor/src/diagram/workbench.tsx
+++ b/guiEditor/src/diagram/workbench.tsx
@@ -291,32 +291,32 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         guiControl.highlightLineWidth = 5;
     }
 
-    private parent(control: Nullable<Control>) {
+    private parent(dropLocationControl: Nullable<Control>) {
         const draggedControl = this.props.globalState.draggedControl;
         const draggedControlParent = draggedControl?.parent;
-        
-        //checking to make sure the element is not being inserted into a child
-        if (this._isNotChildInsert(control, draggedControl)) {
-            if (draggedControlParent && draggedControl) {
-                draggedControlParent.removeControl(draggedControl);
 
-                if (control != null) {
-                    if (this.props.globalState.workbench.isContainer(control) &&
+        if (draggedControlParent && draggedControl) {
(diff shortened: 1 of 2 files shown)


guiEditor/src/diagram/workbench.tsx
```diff
@@ -295,33 +295,46 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
-
-            draggedControlParent.removeControl(draggedControl);
-        
-            if (control != null) {
-                if (this.props.globalState.workbench.isContainer(control)) {
-                    this.props.globalState.guiTexture.removeControl(draggedControl);
-                    (control as Container).addControl(draggedControl);
+        //checking to make sure the element is not being inserted into a child
+        if (this._isNotChildInsert(control, draggedControl)) {
+            if (draggedControlParent && draggedControl) {
```
sebavan, line 300: should be moved before as it is faster to test

guiEditor/src/diagram/workbench.tsx
```diff
@@ -295,33 +295,46 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+                draggedControlParent.removeControl(draggedControl);
+
+                if (control != null) {
```
sebavan, line 303: would be great to comment on the different cases here as it might be hard to catch why we have so many ?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -295,33 +295,46 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
                 }
-                if (control.parent) {
-                    let index = control.parent.children.indexOf(control);
-                    
-                    //adjusting index to be before or after based on where the control is over
-                    index = this._adjustParentingIndex(index); 
-                    control.parent.children.splice(index, 0, draggedControl);
-                    draggedControl.parent = control.parent;
+                else {
+                    draggedControlParent.children.splice(1, 0, draggedControl);
```
sebavan, line 320: why inserting at index 1 ? probably needs a comment too

sebavan (changes requested):

sebavan (approved):

PR #10703, opened 2021-07-20
Add a growable float32array for use in the performance viewer

SahilTara:
This PR is setup for collection of all sorts of data! We will simply push values into this "growable" array for each dataset! The class which will do the actual collection, and handle all of that logic will be done in a separate PR (to keep the PRs manageable). #10566 

No video this time as there is nothing graphical happening in this PR.


src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
...
+     * @param _capacity The initial capacity you would like to set for the array.
+     */
+    constructor(private _capacity: number) {
+        const buffer = new ArrayBuffer(_capacity * bytesPerItem);
+        this._view = new Float32Array(buffer);
+        this._length = 0;
+    }
+
+    /**
+     * The number of items currently in the array.
```
sebavan, line 22: Byte or Item would be great to be in the prop name :-)

src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
...
+        return this._view[index];
+    }
+
+    /**
+     * Gets a view of the original array from start to end (exclusive of end).
+     * @param start starting index.
+     * @param end ending index.
+     * @returns a subarray of the original array.
+     */
+    public subarray(start: number, end: number): GrowableFloat32Array {
```
sebavan, line 59: This creates a copy you can manipulate, if intended would be great to know it explicit from name and comment. 

Indeed, in the comment it states Gets a view which I thought would not copy and if the intent is to not copy, let s only return a Float32Array not a growable copy.

sebavan (changes requested):

src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
+const growthFactor = 1.5;
+const bytesPerItem = 32 / 8;
+
+/**
+ * A class acting as a growable float32array used in the performance viewer
+ */
+export class GrowableFloat32Array {
```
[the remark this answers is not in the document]
SahilTara: Done!

src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
+const growthFactor = 1.5;
+const bytesPerItem = 32 / 8;
```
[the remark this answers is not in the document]
SahilTara: Found out I could get rid of keeping track of the buffer, so the bytes per item is no longer necessary (removed it).

src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
+const growthFactor = 1.5;
+const bytesPerItem = 32 / 8;
+
+/**
+ * A class acting as a growable float32array used in the performance viewer
+ */
+export class GrowableFloat32Array {
+    private _view: Float32Array;
+    private _length: number;
```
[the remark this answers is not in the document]
SahilTara: changed to itemLength!

src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+export class GrowableFloat32Array {
+    private _view: Float32Array;
+    private _length: number;
+
+    /**
+     * Creates a new GrowableFloat32Array with the desired capacity.
+     * @param _capacity The initial capacity you would like to set for the array.
+     */
+    constructor(private _capacity: number) {
```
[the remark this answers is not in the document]
SahilTara: Changed to itemCapacity.

src/Misc/PerformanceViewer/growableFloat32Array.ts
SahilTara: changed to itemLength!

src/Misc/PerformanceViewer/growableFloat32Array.ts
```diff
@@ -0,0 +1,87 @@
...
+        const buffer = new ArrayBuffer(_capacity * bytesPerItem);
```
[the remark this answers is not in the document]
SahilTara: Done!

src/Misc/PerformanceViewer/growableFloat32Array.ts
SahilTara: Yep you're right, found out Float32Array gives me enough properties to just simply return it!

src/Misc/PerformanceViewer/dynamicFloat32Array.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export class DynamicFloat32Array {
+    private _view: Float32Array;
+    private _itemLength: number;
+
+    /**
+     * Creates a new DynamicFloat32Array with the desired item capacity.
+     * @param _itemCapacity The initial item capacity you would like to set for the array.
+     */
+     constructor(private _itemCapacity: number) {
```
bghgary, line 14: nit: why do you need to store itemCapacity? Is that not _view.length?

SahilTara: _view.length may contain elements we haven't initialized as _view.length will get the total capacity we set basically, but the number of actual elements will usually be a lot smaller than this! So this will let us know what the "Usable area" of our array is.

SahilTara: Okay yeah I've updated the code and just use the length now LOL, good catch.

sebavan (approved):

sebavan:
@bghgary I ll merge as soon as you approve this one :-)

sebavan (approved):

PR #10715, opened 2021-07-22
Fix shadows not displayed when using needDepthPrePass

Popov72:
See https://forum.babylonjs.com/t/possible-issue-with-shadowgenerator-on-needdepthprepass-meshes/22538


sebavan (approved):
Great Catch !!!!

PR #10719, opened 2021-07-22
Added PerfCollectionType and relevant strategies for the performance viewer.

SahilTara:
Added skeleton for predefined collection type strategies. Implemented strategy for fps and gpu frame time.

The PerfCollectionType values are set to support a bitfield for enabled collection types.

When we implement the PerformanceViewerCollection class we will use the record to add only the strategies we have enabled into a map for fast lookup when collecting.

Note: Currently we have a defaultGetDataImpl, this will be removed when I eventually implement getData for each and every strategy listed inside the record right now.

#10566


src/Misc/PerformanceViewer/perfCollectionType.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Defines the predefined collection types used in the performance viewer.
+ */
+export enum PerfCollectionType {
```
sebavan, line 4: This will probably not scale enough :-) and we ll be hard to use for custom inputs ?

sebavan (commented):
I kind of like it but I am afraid of the scalability and the lack of openness to externally referenced events ?

src/Misc/PerformanceViewer/perfCollectionType.ts
SahilTara: Well this mapping is only for the predefined collection types! Custom inputs will have their own strategies defined separately from the functions I've created here. The reason for the "closed-ness" is because there isn't really a need to check if we have enabled a custom input, because it wouldn't exist if we had not enabled it. 

we will probably create a separate record for "general strategies" if I find that the logic is shared enough for it (I suspect it will have some shared aspects), but populate the same "strategies" map if we find we have registered some custom events.

SahilTara: Discussed offline. The concern is more so if we expand our use cases. Because of this I have opted to use enum string instead. It has allowed the code to be cleaned up a bit as well! 😄

sebavan (commented):
Why not just exposing the various strategy containing the id + the callback  ???

So it is easy to add a new one ?

PR #10720, opened 2021-07-22
Depth peeling - WIP for review

MiiBond:


Popov72:
We already have #10659 regarding depth peeling, so I'm not quite sure how we should handle the current PR. @sebavan / @deltakosh ?

As I can see it, #10659 is more integrated into the core whereas the current PR is more lightweight but requires the user to do a bit of work.

MiiBond:
The intent of this PR was to provide a simple way to do depth peeling. The user is then free to use it how they wish. My Playground example shows how you could use it to do under-blending of many alpha layers. However, this is NOT how I'm most interested in using it and we should be careful that Babylon's implementation doesn't only provide this.

In PBR, transmission involves reflection, absorption, scattering, dispersion, etc. so simple alpha blending is not sufficient and doesn't have many real-world analogs (maybe porous, non-reflective materials like gauze or burlap). In the past, I had implemented what I called "deferred transparency" in Babylon ([this branch](https://github.com/MiiBond/Babylon.js/tree/transparency_v3)) where each depth-peeled layer was output in a g-buffer containing the relevant information needed to composite the scene back together and account for absorption, scattering, refraction, reflection, etc. My intent with this PR was to start the process of re-implementing this and getting the necessary pieces back into the core Babylon. As it is, I've had to abandon my old code in order to update our viewer to the latest Babylon and I want to avoid that in the future.

As for the rendering differences between the two PR's, I don't know for sure. My method renders 5 layers (you can change this in the Playground, of course) and renders from front to back, using "under" blending (instead of the usual "over" blending). I wouldn't risk much money betting that mine is more "correct" but it does look better, in my eyes.

Popov72:
If I compare with a rendering without OIT in a camera position where the green sphere is drawn before the red sphere (as the green sphere is inside the red sphere) I get:
|Without OIT (green before red)|#10659|this PR|
|-------------|---------|-------|
|image|image|image|

It's more in line with #10659 (it's not exactly the same rendering because of the back faces that are not drawn in the right order in the first screenshot).

In your PR, the green sphere does not look like being inside the red sphere to my eyes. The rendering of this green sphere is not unlike the one we get when not doing OIT but having the green sphere displayed after the red one:
|Without OIT (green after red)|this PR|
|-------------|-------|
|image|image|

MiiBond:
Ah, I misread your previous post and thought that the incorrect rendering was coming from #10659. I agree that the rendering from this PR looks wrong. I probably have the wrong blend equation setup.

However, it's not really that important for this PR since the playground project isn't part of the PR. What I'm trying to do is provide a way to do depth peeling and use the results of it in further rendering. Having a simple, integrated way to do OIT alpha blending as in #10659 is great but it's not what I need.

sebavan:
Hey @MiiBond, may I ask how you intent to use it ? is it mostly for OIT or do you have any other use ? I am wondering how to leverage @CraigFeldspar PR to provide you with the thing you need so that we do not have 2 Depth peeling features in the Framework ?

MiiBond:
I'm using it for OIT but I need control over the passes and blending. I use the depth peel layers to render out MRT's from the PBR uber shader for each layer (basically a series of g-buffers). I then composite those MRT's together, computing things like absorption, refraction, dispersion, etc.
I had implemented this all a couple of years ago but wanted to redo it in a newer version of Babylon and in a way that was more future proof as Babylon gets updated. I had so many hacks and changes in `pbr.fragment.fx` that it just wasn't going to be doable to merge.

sebavan:
@CraigFeldspar is it something we could somehow control from the pre pass ?

Issue #10748, opened 2021-07-24
[GE Enhancement] Clicking on an empty area of the tree view should deselect any selected controls


PatrickRyanMS:
If you have a control selected and click on another control or an empty area of the canvas, the original control will be deselected, but if you click in an empty area of the tree view, the control remains selected. We should add the empty areas of the tree view as another way to deselect the active control.

Issue #10727, opened 2021-07-23
[GE Enhancement] - Copying parented objects needs to respect parent relationships


PirateJC:
When you copy the a child of parent, the new version of the child should have the same parent as the original child. Currently it does not.

PR #10755, opened 2021-07-26
[GE[ Bug Fixes!

msDestiny14:
https://github.com/BabylonJS/Babylon.js/issues/10727
https://github.com/BabylonJS/Babylon.js/issues/10748
https://github.com/BabylonJS/Babylon.js/issues/10733

### DIFF
--- a/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
+++ b/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
@@ -46,6 +46,7 @@ import { OptionsLineComponent } from "../../sharedUiComponents/lines/optionsLine
 import { TextInputLineComponent } from "../../sharedUiComponents/lines/textInputLineComponent";
 import { FloatLineComponent } from "../../sharedUiComponents/lines/floatLineComponent";
 
+
 require("./propertyTab.scss");
 const adtIcon: string = require("../../../public/imgs/adtIcon.svg");
 const responsiveIcon: string = require("../../../public/imgs/responsiveIcon.svg");
@@ -126,15 +127,18 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
     }
 
     save() {
+        this.props.globalState.guiTexture.removeControl(this.props.globalState.workbench.artBoardBackground);
         try {
             const json = JSON.stringify(this.props.globalState.guiTexture.serializeContent());
             StringTools.DownloadAsFile(this.props.globalState.hostDocument, json, "guiTexture.json");
         } catch (error) {
             alert("Unable to save your GUI");
         }
+        this.props.globalState.guiTexture.getChildren()[0].children.unshift(this.props.globalState.workbench.artBoardBackground);
     }
 
     saveToSnippetServer() {
+        this.props.globalState.guiTexture.removeControl(this.props.globalState.workbench.artBoardBackground);
         const adt = this.props.globalState.guiTexture;
         const content = JSON.stringify(adt.serializeContent());
 
@@ -182,6 +186,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
         };
 
         xmlHttp.send(JSON.stringify(dataToSend));
+        adt.getChildren()[0].children.unshift(this.props.globalState.workbench.artBoardBackground);
     }
 
     loadFromSnippet() {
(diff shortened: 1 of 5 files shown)


guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -46,6 +46,7 @@ import { OptionsLineComponent } from "../../sharedUiComponents/lines/optionsLine
 import { TextInputLineComponent } from "../../sharedUiComponents/lines/textInputLineComponent";
 import { FloatLineComponent } from "../../sharedUiComponents/lines/floatLineComponent";
 
+
```
sebavan, line 49: should probably not be there

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -126,15 +127,18 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
 
     save() {
+        this.props.globalState.guiTexture.removeControl(this.props.globalState.workbench.artBoardBackground);
         try {
             const json = JSON.stringify(this.props.globalState.guiTexture.serializeContent());
             StringTools.DownloadAsFile(this.props.globalState.hostDocument, json, "guiTexture.json");
         } catch (error) {
             alert("Unable to save your GUI");
         }
+        this.props.globalState.guiTexture.getChildren()[0].children.unshift(this.props.globalState.workbench.artBoardBackground);
```
sebavan, line 137: This sounds unsafe if there are no children ?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -126,15 +127,18 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
     }
 
     save() {
+        this.props.globalState.guiTexture.removeControl(this.props.globalState.workbench.artBoardBackground);
```
sebavan, line 130: A comment would be nice here and on 137 as I am not sure why we need this and the unshift ?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -126,15 +127,18 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
             const json = JSON.stringify(this.props.globalState.guiTexture.serializeContent());
             StringTools.DownloadAsFile(this.props.globalState.hostDocument, json, "guiTexture.json");
         } catch (error) {
             alert("Unable to save your GUI");
         }
+        this.props.globalState.guiTexture.getChildren()[0].children.unshift(this.props.globalState.workbench.artBoardBackground);
     }
 
     saveToSnippetServer() {
+        this.props.globalState.guiTexture.removeControl(this.props.globalState.workbench.artBoardBackground);
```
sebavan, line 141: Same here and 189 ?

sebavan: Also this maybe could be factorize with a function taking a callback in parameter as you use the same pattern twice ?

sebavan (changes requested):
tiny comments

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Not sure what you mean by this. Could you explain?

sebavan:
```
function foo(doSmthgCallback: () => void) {
  this.props.globalState.guiTexture.removeControl(this.props.globalState.workbench.artBoardBackground);
  doSmthgCallback();
  adt.getChildren()[0].children.unshift(this.props.globalState.workbench.artBoardBackground);
}
```

smthg in this spirit ???

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -290,12 +295,15 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
 
-                                if (newControl) { //insert the new control into the adt
+                                if (newControl) { //insert the new control into the adt or parent container
                                     this.props.globalState.workbench.appendBlock(newControl);
+                                    this.props.globalState.guiTexture.removeControl(newControl);
+                                    this.state.currentNode.parent?.addControl(newControl);
+
                                     let index = 1;
-                                    while (this.props.globalState.workbench.nodes.filter(  //search if there are any copies
+                                    while ( this.props.globalState.guiTexture.getDescendants(false).filter(  //search if there are any copies
```
RaananW, line 304: why does this need to be iterated in a while loop? why don't you filter all duplicates and change their names?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -182,6 +186,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
         };
 
         xmlHttp.send(JSON.stringify(dataToSend));
+        adt.getChildren()[0].children.unshift(this.props.globalState.workbench.artBoardBackground);
```
RaananW, line 189: same comment @sebavan had - if no children are found this will throw

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Because people will only be copying one at a time. there may also be a "item copy 2" or and "item copy 5?" Now we could filter names out by searching for "item" but then we can get cases like "item item copy" or "item 2"

RaananW: so you are adding the item with the same name, iterate to check if there are 2 items with the same name, and then change the name, until it is no longer a duplicate?

Wouldn't iterating over `${this.state.currentNode.name} Copy ${index++}` until it is not found be enough?

PR #10756, opened 2021-07-26
Compile time bounds checking for TmpVectors

ryantrem:
I recently wrote some new code in Babylon.js that uses `TmpVectors`. I found this a little difficult to use because I needed to go to the definition to know how many temp objects were created for each array, and it felt very prone to human error. To address this, I added a `BuildTuple` function next to the `BuildArray` function that returns a strongly typed tuple (fixed length) and use the `BuildTuple` function in `TmpVectors`. `BuildArray` is used elsewhere in the code where it looks like it should also use `BuildTuple`, but in many (all?) of those cases, it also looks like it could just be using `TmpVectors` instead of creating new private temp storage. I'm happy to make changes there too, I just wasn't sure if there was any reason for these private versions other than being legacy.

With `BuildTuple`, attempting to index into the tuple/array out of range will cause a compile error, and intellisense makes it clear how many temp elements are available:
image

image

image

It's probably possible to define the tuple types more generically using some kind of recursive conditional type, but my understanding is that recursive types often significantly impact TypeScript compiler performance, and in this case we never create more than 13 elements so just manually defining the tuple types is perfectly manageable.

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -6318,20 +6318,20 @@ export class Matrix {
  * Same as Tmp but not exported to keep it only for math functions to avoid conflicts
  */
 class MathTmp {
-    public static Vector3: Vector3[] = ArrayTools.BuildArray(11, Vector3.Zero);
-    public static Matrix: Matrix[] = ArrayTools.BuildArray(2, Matrix.Identity);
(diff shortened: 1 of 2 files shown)


sebavan (approved):
I love it !!!!

PR #10758, opened 2021-07-27
WebGPU: Align maxVaryingVectors with the spec

Popov72:


sebavan:
How is the limit 15 :-) ? Is is stored on 4 bits :-)

PR #10762, opened 2021-07-28
Re-instate output render target fix with XR support

RaananW:
This PR brings back the changes made regarding output render target and active cameras, with a fix making XR work correctly.


sebavan (approved):
LGTM

bghgary:
It would be good if we can test this change in Babylon Native before we merge.

rgerd:
@bghgary / @RaananW I can take a look

RaananW:
Yep, that's what I wanted you to do before merging. Let me know if there are any issues

RaananW:
I have tested it on mixed reality (emulator and device), quest, oculus link and AR mode in android. all work as expected.

src/scene.ts
```diff
@@ -4282,6 +4299,18 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         }
 
         this._frameId++;
+        this._defaultFrameBufferCleared = false;
+        if (this._activeCamera?.outputRenderTarget && !this.activeCamera?.isRigCamera) {
+            this._activeCamera.outputRenderTarget._cleared = false;
+        }
```
rgerd, line 4305: Testing with the [default BabylonNative playground app](https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/experience.js), I was seeing broken clearing behavior due to the rig camera output render targets not getting their `_cleared` flags reset. Adding this block fixed the behavior:
```javascript
if (!!this._activeCamera?.rigCameras?.length) {
    for (let i = 0; i < this._activeCamera.rigCameras.length; ++i) {
        const rtt = this._activeCamera.rigCameras[i].outputRenderTarget;
        if (rtt) {
            rtt._cleared = false;
        }
    }
}
```

There might be a better solution, just throwing this one out there.

I don't have my VR headset in the office right now (testing on HoloLens), but this should be reproducible on the x64 win32 build of BabylonNative `mkdir Build; cd Build; cmake ..`, set [`vr = true`](https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/experience.js#L11) in experience.js.

src/scene.ts
```diff
@@ -3965,30 +3965,47 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
+        if (clear) {
+            this._clearFrameBuffer(camera);
+        }
+    }
+
+    private _clearFrameBuffer(camera: Nullable<Camera>) {
+        // we assume the framebuffer currently bound is the right one
+        if (camera && camera._multiviewTexture) {
+            // no clearing?
+        } if (camera && camera.outputRenderTarget) {
```
rgerd, line 3987: Should this be `else if`?

src/scene.ts
RaananW: Thanks a lot, I added the lines you suggested.

RaananW:
ready for review/merge, if everyone is fine with it.

sebavan:
I ll do tomorrow in the morning with a nightly to monitor during the day :-)

PR #10763, opened 2021-07-28
Add the performance viewer data collector

SahilTara:
This PR adds the performance viewer data collector, and new interfaces to support this collection of data. We use a data structure as follows to store the datasets.

```typescript
export interface IPerfDatasets {
    /**
     * The ids of our dataset.
     */
    ids: string[];

    /**
     * The data to be processed by the performance graph. Each slice will be of the form of [timestamp, numberOfPoints, value1, value2...]
     */
    data: DynamicFloat32Array;

    /**
     * A list of starting indices for each slice of data collected. Used for fast access of an arbitrary slice inside the data array.
     */
    startingIndices: DynamicFloat32Array;
}
```

We store an array of ids for cross reference for what offset belongs to which id. We do not allow removal of individual performance strategies because of this. 

We use the data property to store all of our data in the shape as described by the comment. This allows us to eliminate redundant timestamp data being stored, and allows us to only have a single array which stores all our data. 

We use the startingIndices array for fast lookup on where each slice begins.

This overall will decrease memory consumption from the previous method of collecting data as an array of IPerfPoints and having a separate array for each dataset, and it will also increase performance! Finishes the create data collector task in #10566

### DIFF
--- a/src/Misc/PerformanceViewer/performanceViewerCollector.ts
+++ b/src/Misc/PerformanceViewer/performanceViewerCollector.ts
@@ -37,7 +37,7 @@ export class PerformanceViewerCollector {
      * @param _scene the scene to collect on.
      * @param _enabledStrategyCallbacks the list of data to collect with callbacks for initialization purposes.
      */
-    constructor(private _scene: Scene, private _enabledStrategyCallbacks: PerfStrategyInitialization[]) {
+    constructor(private _scene: Scene, _enabledStrategyCallbacks?: PerfStrategyInitialization[]) {
         this.datasets = {
             ids: [],
             data: new DynamicFloat32Array(initialArraySize),
@@ -47,15 +47,22 @@ export class PerformanceViewerCollector {
         this._datasetMeta = new Map<string, IPerfMetadata>();
         this.datasetObservable = new Observable();
         this.metadataObservable = new Observable((observer) => observer.callback(this._datasetMeta, new EventState(0)));
-        this._initialize();
+        if (_enabledStrategyCallbacks) {
+            this.addCollectionStrategies(_enabledStrategyCallbacks);
+        }
     }
 
     /**
-     * Populates our internal structures and does appropriate initialization of strategies.
+     * This method adds additional collection strategies for data collection purposes.
+     * @param strategyCallbacks the list of data to collect with callbacks. 
      */
-    private _initialize() {
-        for (const strategyCallback of this._enabledStrategyCallbacks) {
+    public addCollectionStrategies(strategyCallbacks: PerfStrategyInitialization[]) {
+        for (const strategyCallback of strategyCallbacks) {
             const strategy = strategyCallback(this._scene);
+
+            if (this._strategies.has(strategy.id)) {
+                continue;
+            }
 
             this.datasets.ids.push(strategy.id);
 
--- a/src/Misc/PerformanceViewer/performanceViewerCollector.ts
+++ b/src/Misc/PerformanceViewer/performanceViewerCollector.ts
@@ -86,7 +86,7 @@ export class PerformanceViewerCollector {
             this.datasets.ids.push(strategy.id);
 
             this._datasetMeta.set(strategy.id, {
-                color: this._getHexFromId(strategy.id),
+                color: this._getHexColorFromId(strategy.id),
             });
 
             this._strategies.set(strategy.id, strategy);
@@ -96,11 +96,11 @@ export class PerformanceViewerCollector {
     }
 
     /**
-     * Gets a 6 character hexcode from a passed in string.
+     * Gets a 6 character hexcode representing the colour from a passed in string.
      * @param id the string to get a hex code for.
      * @returns a hexcode hashed from the id.
      */
-    private _getHexFromId(id: string) {
+    private _getHexColorFromId(id: string) {
         // this first bit is just a known way of hashing a string.
         let hash = 0;
         for (let i = 0; i < id.length; i++) {
--- a/src/Misc/PerformanceViewer/performanceViewerCollector.ts
+++ b/src/Misc/PerformanceViewer/performanceViewerCollector.ts
@@ -54,7 +54,7 @@ export class PerformanceViewerCollector {
 
     /**
      * This method adds additional collection strategies for data collection purposes.
-     * @param strategyCallbacks the list of data to collect with callbacks. 
+     * @param strategyCallbacks the list of data to collect with callbacks.
      */
     public addCollectionStrategies(strategyCallbacks: PerfStrategyInitialization[]) {
         for (const strategyCallback of strategyCallbacks) {
--- a/src/Misc/PerformanceViewer/performanceViewerCollector.ts
+++ b/src/Misc/PerformanceViewer/performanceViewerCollector.ts
@@ -1,5 +1,5 @@
 import { Scene } from "../../scene";
-import { IPerfDatasets, IPerfDataSliceSnapshot, IPerfMetadata } from "../interfaces/iPerfViewer";
+import { IPerfDatasets, IPerfMetadata } from "../interfaces/iPerfViewer";
 import { EventState, Observable } from "../observable";
 import { PrecisionDate } from "../precisionDate";
 import { DynamicFloat32Array } from "./dynamicFloat32Array";
@@ -30,7 +30,7 @@ export class PerformanceViewerCollector {
     /**
      * An observable you can attach to get deltas in the dataset. Subscribing to this will increase memory consumption slightly, and may hurt performance due to increased garbage collection needed.
      */
-    public readonly datasetObservable: Observable<IPerfDataSliceSnapshot>;
+    public readonly datasetObservable: Observable<number[]>;
     /**
      * An observable you can attach to get the most updated map of metadatas.
      */
@@ -158,7 +158,7 @@ export class PerformanceViewerCollector {
                 slice.push(this.datasets.data.at(startingIndex + PerformanceViewerCollector.SliceDataOffset + i));
             }
 
-            this.datasetObservable.notifyObservers({slice});
+            this.datasetObservable.notifyObservers(slice);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/Misc/interfaces/iPerfViewer.ts
```diff
@@ -13,6 +16,7 @@ export interface IPerfPoint {
     value: number;
 }
 
+// TODO: REMOVE ONCE Everything is connected!
```
SahilTara, line 19: To minimize the size of the PR I have marked these for removal and they will be removed in a later PR (When the canvasGraphService changes are in PR I will remove these interfaces).

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -0,0 +1,208 @@
...
+
+            this.datasetObservable.notifyObservers({slice});
+        }
+    }
+
+    /**
+     * Collects data for every dataset by using the appropriate strategy when the user wants.
+     * This method will then notify all observers with the latest slice of data.
+     */
+    public collectDataNow() {
```
SahilTara, line 151: I might be replacing this method with a single public collectData (which will just be _collectDataAtFrame renamed) if it makes sense to also update the datasets when the user calls this function.

SahilTara: What I have right now seems to be fine.

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -0,0 +1,225 @@
...
+        if (_enabledStrategyCallbacks) {
+            this.addCollectionStrategies(_enabledStrategyCallbacks);
+        }
+    }
+
+    /**
+     * This method adds additional collection strategies for data collection purposes.
+     * @param strategyCallbacks the list of data to collect with callbacks.
+     */
+    public addCollectionStrategies(strategyCallbacks: PerfStrategyInitialization[]) {
```
RaananW, line 77: if I want to add a single one, do I need to send an array?
Maybe you could use spread operator (https://howtodoinjava.com/typescript/spread-operator/), or make this addCollectionStrategy while leaving this as a help function that iterates over an array and uses the single add method?

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -0,0 +1,225 @@
...
+
+        this.metadataObservable.notifyObservers(this._datasetMeta);
+    }
+
+    /**
+     * Gets a 6 character hexcode from a passed in string.
+     * @param id the string to get a hex code for.
+     * @returns a hexcode hashed from the id.
+     */
+    private _getHexFromId(id: string) {
```
RaananW, line 102: is that just a way to hash the id, or is there a reason behind the 6 characters?

src/Misc/index.ts
```diff
@@ -55,3 +55,4 @@ export * from './timer';
 export * from "./copyTools";
 export * from "./reflector";
 export * from "./domManagement";
+export * from "./PerformanceViewer/index";
```
RaananW, line 58: no need for index

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            this.datasetObservable.notifyObservers({slice});
+        }
+    }
+
+    /**
+     * Collects data for every dataset by using the appropriate strategy when the user wants.
+     * This method will then notify all observers with the latest slice of data.
+     */
```
RaananW, line 168: Something in the name of the function seems off. getCurrentData? getCurrentSlice?

src/Misc/interfaces/iPerfViewer.ts
```diff
@@ -37,3 +41,48 @@ export interface IPerfDataset {
...
+     /**
+      * Specifies if data should be hidden, falsey by default.
+      */
+     hidden?: boolean;
+}
+
+/**
+ * Defines the shape of a snapshot sent to observers.
+ */
+export interface IPerfDataSliceSnapshot {
```
RaananW, line 83: any plan of extending it? (i.e. is it 100% needed, or is number[] enough?)

RaananW:
Just had a quick pass at the PR, I will run over it again tomorrow morning to see if I missed something. Just a few questions actually :-)

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: Oh yeah thanks for the suggestion, I forgot I could simply just use the spread operator 😄

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: Oh it gives you a 6 character hexcode representing the colour. Perhaps I should rename this to _getHexColorFromId?

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: I like getCurrentSlice!

src/Misc/interfaces/iPerfViewer.ts
SahilTara: Hmmm, I was thinking of possibly passing the ids with it, but I realized that I can just do datasets.ids for them! So I can just use number[].

src/Misc/index.ts
SahilTara: Oh huh, weird it wasn't working without index before but it is now (I had thought it wouldn't need the index either). Thanks for the suggestion!

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: Done!

src/Misc/index.ts
SahilTara: Updated!

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: Fixed it I think!

src/Misc/interfaces/iPerfViewer.ts
SahilTara: Okay I changed to number[]!

src/Misc/index.ts
SahilTara: [22:18:48] Line 58 Exports ./PerformanceViewer/ should be from the full path including index. /home/vsts/work/1/s/src/Misc/index.ts 

If I remove the index it seems to fail the lint check :(

sebavan (changes requested):
Is there any reasons to have some of the tools in src/misc and not as part of the perf profiler project ???

(this would help with the /index as by default we should never import full folders cause the lib can not be flagged as side effect false)

RaananW:
@sebavan - yes, we want the data collection process to be a part of core and not a part of an external project. this way we can have different views of the same data (and other devs will be able to implement their own perf analyzer without importing an external project).

src/Misc/PerformanceViewer/performanceViewerCollector.ts
RaananW: that would make it clearer. 
You could also assign a random color ;-) but it is nice that it is generated. this way every time a specific ID will be used it will have the same color.

src/Misc/index.ts
RaananW: Sorry! my fault. index is needed...

sebavan (approved):

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -0,0 +1,225 @@
...
+
+        meta[prop] = value;
+
+        this.metadataObservable.notifyObservers(this._datasetMeta);
+    }
+
+    /**
+     * Starts the realtime collection of data.
+     */
+    public start() {
```
RaananW, line 214: If the user calls start, stop and then start again it will have a huge gap in timestamps. Might be what the user wanted, but you could add a flag parameter to reset the datasets (maybe the default behavior?).
Maybe a future improvement would be to set an ID to this dataset, so every call to start it will save a new dataset so that the user can go back to older datasets but still get a new one when start is called again

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -0,0 +1,225 @@
...
+            for (let i = 0; i < numPoints; i++) {
+                slice.push(this.datasets.data.at(startingIndex + PerformanceViewerCollector.SliceDataOffset + i));
+            }
+
+            this.datasetObservable.notifyObservers(slice);
+        }
+    }
+
+    /**
+     * Collects and then sends the latest slice to any observers by using the appropriate strategy when the user wants.
```
RaananW, line 166: would be great to document the structure of the data slice. here and in the observable's codedoc would make sense

RaananW (approved):
LGTM! Added 2 suggestions, and waiting for the index fix

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: We now reset the indices and the data.

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: Documented the slices!

src/Misc/index.ts
SahilTara: Index is back!

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: Changed the name!

PR #10766, opened 2021-07-28
fix(RenderTargetTexture): Fix case where the scene is null when marking meshes as light dirty

zakhenry:
I'm still struggling to understand exactly how this scenario comes about, I believe it is something like calling mesh.dispose() after the scene has already been destroyed, or is being destroyed.

Either way this fixes the following error:

```
VM91032:1 Uncaught TypeError: Cannot read property 'meshes' of null
    at eval (eval at array.splice (renderTargetTexture.js:171), <anonymous>:1:17)
    at Array.array.splice (renderTargetTexture.js:171)
    at Array.array.splice (renderTargetTexture.js:169)
    at Array.array.splice (renderTargetTexture.js:169)
    at Array.array.splice (renderTargetTexture.js:169)
    at abstractMesh.js:1573
    at Array.forEach (<anonymous>)
    at Mesh.AbstractMesh.dispose (abstractMesh.js:1557)
    at Mesh.dispose (mesh.js:2060)
    at Subscription.<anonymous> (mesh-loader.ts:136)
```

### DIFF
--- a/src/Materials/Textures/renderTargetTexture.ts
+++ b/src/Materials/Textures/renderTargetTexture.ts
@@ -80,8 +80,8 @@ export class RenderTargetTexture extends Texture {
 
             var result = oldPush.apply(array, items);
 
-            if (wasEmpty && this.getScene()) {
-                this.getScene()!.meshes.forEach((mesh) => {
+            if (wasEmpty) {
+                this.getScene()?.meshes.forEach((mesh) => {
                     mesh._markSubMeshesAsLightDirty();
                 });
             }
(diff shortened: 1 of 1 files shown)


sebavan (changes requested):
Could you instead replace `this.getScene()!` by `this.getScene()?` :-)

zakhenry:
I'll give it a try (one sec gotta clone the repo etc) but I don't think typescript will like it because it probably can't safely infer that the value returned from getScene() hasn't changed between invocations.

I could instead assign scene to a const then the not-null assertion can go away

zakhenry:
wait never mind I didn't realise your `?` was not a question, but was a safe navigation operation 😆

zakhenry:
@sebavan ok fixed, and fixed the nearby one while I was at it

sebavan (approved):

PR #10768, opened 2021-07-29
[XR] WebXR spectator mode for desktop VR experiences

### LINKED ISSUE #10560: Using WebXR with a second canvas does not work
Specifying a canvas to use for the WebXR session, via outputCanvasOptions.canvasElement, results in a failure to render to the XR headset. 

This issue was identified as the underlying issue in two threads on the forum:
https://forum.babylonjs.com/t/webxr-render-canvas-and-streaming/20893 and
https://forum.babylonjs.com/t/strange-performance-with-webxr-ammojs-and-scene-animate/21675

Both of these threads started out investigating slightly different aspects and other factors were first thought to be an issue (such as Ammo) but the underlying issue, common to both, is that the WebXR session is somehow still dependant on the canvas that the engine was created with, regardless of the setting in outputCanvasOptions.

- Expected result: WebXR should only use the canvas specified in th eoptions and leave the engine's canvas untouched.
- Current result: Specifying an alternate canvas via the options breaks the rendering to the headset, ie it is still rendered to the engine's canvas

**Headset**
 - OS: tested on a Lenovo Mirage running android (daydream)
 - Browser: Chrome
 - Version: 84

**Additional context**
The frame buffers ARE correctly setup and it does seem to pass the canvas thru but somewhere in the render process it reverts to the renderTarget that the engine was created with/ In this sense, WebXR is ONLY working if the canvas used to setup the render buffers is the canvas that was used to create the engine.

I have setup a standalone project, outside of the playground, that can be used to reproduce this. This test has extra stuff in it as I  was isolating the issue but I have left them in in case it helps further testing. The test in its current state creates a physics playground and then attempts to start an XR session based on the value of canvasToUse, which can be changed on lines 58,59 to use either the canvas the engine was created with or an alternate canvas. Using the alternate canvas does not work, using the engine's canvas does. Other factors, such as streaming or physics engine mentioned in the threads was a red herring. Simply using a different canvas breaks WebXR. 
(Note: In my main project, as I am streaming the engine's canvas, I can see that the WebXR render IS going to the engine's canvas rather than using the frame buffers to render to the headset like it should be)

[test.zip](https://github.com/BabylonJS/Babylon.js/files/6705386/test.zip)

RaananW:
This PR enables spectator mode - cloning the first rig camera's view to the desktop canvas.
As written in the codedoc - this is expected to degrtade performance due to the extra render call of a camera (and technically the entire scene).

Fixes #10560

To use a different camera type or manipulate the camera on your own you can create the camera and set it to be the second in the activeCameras array:

```javascript
scene.activeCameras = [webXrcamera, yourNewCamera];
```

An important thing to remember is that the dimensions object will need to be reset in order for the new camera to render correctly:

```javascript
scene.onAfterRenderCameraObservable.add((camera) => {
  if (camera === webXrCamera) {
    // reset the dimensions object for correct resizing
    scene.getEngine().framebufferDimensionsObject = null;
  }
});
```

You can then update the camera's transformation (or not, if it is not needed) on each XRFrame

### DIFF
--- a/src/scene.ts
+++ b/src/scene.ts
@@ -4284,6 +4284,20 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         }
     }
 
+    private checkCameraRenderTarget(camera: Nullable<Camera>) {
+        if (camera?.outputRenderTarget && !camera?.isRigCamera) {
+            camera.outputRenderTarget._cleared = false;
+        }
+        if (camera?.rigCameras?.length) {
+            for (let i = 0; i < camera.rigCameras.length; ++i) {
+                const rtt = camera.rigCameras[i].outputRenderTarget;
+                if (rtt) {
+                    rtt._cleared = false;
+                }
+            }
+        }
+    }
+
     /**
      * Render the scene
      * @param updateCameras defines a boolean indicating if cameras must update according to their inputs (true by default)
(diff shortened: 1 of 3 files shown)


src/scene.ts
```diff
@@ -4300,24 +4300,22 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
-        if (this._activeCamera?.outputRenderTarget && !this.activeCamera?.isRigCamera) {
-            this._activeCamera.outputRenderTarget._cleared = false;
-        }
-        if (this._activeCamera?.rigCameras?.length) {
-            for (let i = 0; i < this._activeCamera.rigCameras.length; ++i) {
-                const rtt = this._activeCamera.rigCameras[i].outputRenderTarget;
-                if (rtt) {
-                    rtt._cleared = false;
-                }
+        const checkRtt = (camera: Nullable<Camera>) => {
```
sebavan, line 4303: This would create GC is it possible to move in it s own function instead ?

RaananW: sure!

src/scene.ts
```diff
@@ -4284,6 +4284,20 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         }
     }
 
+    private checkCameraRenderTarget(camera: Nullable<Camera>) {
```
sebavan, line 4287: should it be checkCameraRenderTarget = (camera: Nullable<Camera>) => {} to prevent issue if called from callbacks ?

sebavan (approved):

PR #10774, opened 2021-07-29
Implement the rest of the strategies for the performance viewer, except the CPU Strategy

SahilTara:
This PR Implements the rest of the strategies except for the cpu strategy within the default strategies we provide. This helps with the "Allow for the collection of draw calls, number of active meshes, frame time." and it helps complete the "Collect everything under count and frame steps duration" stretch goal #10566. 

Note: You will not be able to see all of these strategies on the graph by default, only a select portion that we will define, and then the user will be able to add more through the shared perfviewercollector.

### DIFF
--- a/src/Misc/PerformanceViewer/performanceViewerCollectionStrategies.ts
+++ b/src/Misc/PerformanceViewer/performanceViewerCollectionStrategies.ts
@@ -26,7 +26,7 @@ const defaultDisposeImpl = () => {};
 const defaultGetDataImpl = () => 0;
 
 /**
- * Initializer callback for a strategy, we allow sceneInstrumentation to save on allocations of this object as a lot of default strategies use it.
+ * Initializer callback for a strategy
  */
 export type PerfStrategyInitialization = (scene: Scene) => IPerfViewerCollectionStrategy;
 /**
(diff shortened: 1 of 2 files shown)


src/Misc/PerformanceViewer/performanceViewerCollectionStrategies.ts
```diff
@@ -201,10 +203,12 @@ export class PerfCollectionStrategy {
...
      */
     public static AbsoluteFpsStrategy(): PerfStrategyInitialization {
-        return () => {
+        return (scene) => {
+            const sceneInstrumentation = new SceneInstrumentation(scene);
+            sceneInstrumentation.captureFrameTime = true;
             return {
                 id: "absolute fps",
-                getData: defaultGetDataImpl,
+                getData: () => 1000.0 / sceneInstrumentation.frameTimeCounter.lastSecAverage,
```
SahilTara, line 211: I use lastSecondAverage a lot, I wonder if it would be better to use current instead of lastSecAverage given that my use case is real time data, vs the statistics tab being "every 500ms" data. Any thoughts?

RaananW (commented):
I'm wondering in general about the scene instrumental class. It seems like you are creating one for each strategy. Do you think there is a way of creating it once and reusing it?

src/Misc/PerformanceViewer/performanceViewerCollectionStrategies.ts
RaananW: yes, current value would be the right one for a realtime counter

SahilTara: Done!

src/Misc/PerformanceViewer/performanceViewerCollectionStrategies.ts
```diff
@@ -136,10 +151,17 @@ export class PerfCollectionStrategy {
      * @returns the initializer for the draw calls strategy
      */
     public static DrawCallsStrategy(): PerfStrategyInitialization {
-        return () => {
+        return (scene) => {
```
SahilTara, line 154: For some weird reason I seem to be getting a value of 0 for this strategy all the time :(. Going to see if i can figure out why

SahilTara: Oh wow I figured out why HAHA, had to do with me calling capture in onBeforeRender but the sceneInstrumentation clears this value onBeforeRender and it gets priority since it is added first! I now save the number of draw calls at the end of onAfterRender and return this value when needed!

PR #10775, opened 2021-07-30
Make the nativeOverrides.filter decorator factory (more) type safe

ryantrem:
I got help from the TypeScript team on this one, so I think it is the best we can do right now. If the number of arguments to filter predicate are incorrect, there will be a compile error. If the types of the arguments in the filter predicate are explicit and incorrect, there will be a compile error. However, without explicitly specifying the parameter types they are still not inferred (the compiler just thinks they are any). So additionally the usage pattern I've adopted here is to deduce the parameters from the function explicitly via `Parameters<typeof MathHelpers.extractMinAndMaxIndexed>` (for example) with a spread operator on the arg list and then destructure the arg list so we can refer to specific args. It's a little funky, but much more type safe than it was before. This extra type safety also revealed a bug in the original implementation, where the return value of the function was not actually returned on the first invocation (we are currently only overriding void returning functions).


sebavan (commented):
just wondering why you need ...[positions, indices] instead of just positions, indices when you call into it ?

PR #10776, opened 2021-07-30
Fix bakeTransformIntoVertices not working when using rotationQuaternion

Popov72:
See https://forum.babylonjs.com/t/baking-quaternions-deforms-meshes/22722


src/Meshes/transformNode.ts
```diff
@@ -1204,7 +1204,7 @@ export class TransformNode extends Node {
                     let tmpRotationQuaternion = TmpVectors.Quaternion[0];
                     bakedMatrix.decompose(child.scaling, tmpRotationQuaternion, child.position);
                     if (child.rotationQuaternion) {
-                        child.rotationQuaternion = tmpRotationQuaternion;
+                        child.rotationQuaternion = tmpRotationQuaternion.clone();
```
RaananW, line 1207: copyFrom can be used, if it already exists

RaananW (approved):
Just a small comment (one less object creation)

PR #10783, opened 2021-07-30
Apply changes to use perfCollector dataset and metadata in our canvas graph service.

SahilTara:
This PR finishes off the connection from the perf collector and the canvas a graph service. It also removes all the extra code that was unable to be changed in previous PRs because of the canvas graph service. 

Overall this PR also simplifies a lot of the logic inside the canvas graph service 😄.

Finishes the "Connect the graph to the collector." task inside #10566

### DIFF
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -1,6 +1,7 @@
 import { ICanvasGraphServiceSettings, IPerfMinMax, IGraphDrawableArea, IPerfMousePanningPosition, IPerfIndexBounds, IPerfTooltip, IPerfTextMeasureCache, IPerfLayoutSize } from "./graphSupportingTypes";
-import { IPerfDataset, IPerfPoint } from "babylonjs/Misc/interfaces/iPerfViewer";
+import { IPerfDatasets, IPerfMetadata } from "babylonjs/Misc/interfaces/iPerfViewer";
 import { Scalar } from "babylonjs/Maths/math.scalar";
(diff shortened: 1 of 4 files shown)


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -170,15 +171,10 @@ export class CanvasGraphService {
...
-        this.datasets.forEach((dataset: IPerfDataset) => {
-            // skip hidden and empty datasets!
-            if (dataset.data.length === 0 || !!dataset.hidden) {
-                return;
-            }
-
-            const pos = this._positions.get(dataset.id) ?? dataset.data.length - 1;
+            
+        // First we must get the end positions of our view port.
+        if (this._getNumberOfSlices() > 0) {
```
sebavan, line 176: could save this._getNumberOfSlices() ??? and reuse

sebavan (approved):
Only one neat for me but @RaananW might have more later today :-)

RaananW (approved):

inspector/src/components/graph/canvasGraphService.ts
SahilTara: Yeah looking at the code I could probably just not even save, but just check once and return early rather than every few lines, because theres nothing to do really if numberOfSlices === 0.

SahilTara: Okay yeah i still had to save because of some other uses like end - this._getNumberOfSlices and such, but I also  exit early.

PR #10786, opened 2021-07-31
Add new entry point to PBRCustomMaterial and a isEnabled property to Layer

Popov72:
Need these for the WebGPU demo!

I have also added the possibility to read the data from a texture "as is", meaning with no conversion. Currently, the data are read as an Uint8Array if the type is UNSIGNED_BYTE and as a Float32Array in all other cases. When using the new parameter, the buffer will be created as (U)Int8Array / (U)Int16Array / (U)Int32Array / Float32Array depending on the type of the texture.


src/Engines/thinEngine.ts
```diff
@@ -4700,6 +4700,83 @@ export class ThinEngine {
         return internalFormat;
     }
 
+    /**
```
sebavan, line 4703: Could this be moved from thinEngine to prevent growing it?

sebavan (changes requested):

src/Engines/thinEngine.ts
Popov72: I first put it in Misc/TextureTools.ts, but it means a dependency on this file from Extensions/engine.readTexture.ts: is that not a problem?

sebavan: so why not placing it in Extensions/engine.readTexture.ts ?

sebavan: but texture tools would be ok

Popov72: Moved to TextureTools.

sebavan (approved):

src/Engines/thinEngine.ts
Popov72: Ok, it does not work to put it in `TextureTools`, so I have moved the code to `engine.readTexture`.

sebavan:
One doc check to go and I ll merge @Popov72

PR #10789, opened 2021-08-02
navigation reaching observer

CedricGuillemet:
Following this discussion thread on Twitter : https://twitter.com/low_dencity/status/1418505253612400641

Added an observer that fires when agent reaches destination (at least, when he is within radius of destination)

### DIFF
--- a/src/Navigation/INavigationEngine.ts
+++ b/src/Navigation/INavigationEngine.ts
@@ -377,7 +377,7 @@ export interface IAgentParameters {
      * Observers will be notified when agent gets inside the virtual circle with this Radius around destination point.
      * Default is agent radius
      */
-     reachRadius: number;
+     reachRadius?: number;
 }
 
 /**
--- a/src/Navigation/INavigationEngine.ts
+++ b/src/Navigation/INavigationEngine.ts
@@ -372,6 +372,12 @@ export interface IAgentParameters {
      * How aggressive the agent manager should be at avoiding collisions with this agent. [Limit: >= 0]
      */
     separationWeight: number;
+
+    /**
+     * Observers will be notified when agent gets inside the virtual circle with this Radius around destination point.
+     * Default is agent radius
+     */
+     reachRadius: number;
 }
 
 /**
--- a/src/Navigation/Plugins/recastJSPlugin.ts
+++ b/src/Navigation/Plugins/recastJSPlugin.ts
@@ -484,7 +484,7 @@ export class RecastJSCrowd implements ICrowd {
      */
     public reachRadii: number[] = new Array<number>();
     /**
-     * true when a destination is active for an agent and notifier hasn't been noticed of reach
+     * true when a destination is active for an agent and notifier hasn't been notified of reach
      */
     private agentDestinationArmed: boolean[] = new Array<boolean>();
     /**
@@ -502,7 +502,7 @@ export class RecastJSCrowd implements ICrowd {
     private _onBeforeAnimationsObserver: Nullable<Observer<Scene>> = null;
 
     /**
-     *  Fires each time an agent is in reach radius of his destination
+     *  Fires each time an agent is in reach radius of its destination
      */
     public onReachTargetObservable = new Observable<{agentIndex: number, destination: Vector3}>();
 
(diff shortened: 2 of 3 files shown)


src/Navigation/INavigationEngine.ts
```diff
@@ -372,6 +372,12 @@ export interface IAgentParameters {
      * How aggressive the agent manager should be at avoiding collisions with this agent. [Limit: >= 0]
      */
     separationWeight: number;
+
+    /**
+     * Observers will be notified when agent gets inside the virtual circle with this Radius around destination point.
+     * Default is agent radius
+     */
+     reachRadius: number;
```
Popov72, line 380: If there's a default, maybe it should be declared this way:
```typescript
reachRadius?: number;
```

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -490,6 +479,18 @@ export class RecastJSCrowd implements ICrowd {
      * All agents created
      */
     public agents: number[] = new Array<number>();
+    /**
+     * agents reach radius
+     */
+    public reachRadii: number[] = new Array<number>();
+    /**
+     * true when a destination is active for an agent and notifier hasn't been noticed of reach
```
Popov72, line 487: nit: I think you meant "notified" instead of "noticed"

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -500,6 +501,11 @@ export class RecastJSCrowd implements ICrowd {
      */
     private _onBeforeAnimationsObserver: Nullable<Observer<Scene>> = null;
 
+    /**
+     *  Fires each time an agent is in reach radius of his destination
```
Popov72, line 505: nit: "its" instead of "his". Or maybe "their"?

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -632,6 +640,13 @@ export class RecastJSCrowd implements ICrowd {
      */
     agentGoto(index: number, destination: Vector3): void {
         this.recastCrowd.agentGoto(index, new this.bjsRECASTPlugin.bjsRECAST.Vec3(destination.x, destination.y, destination.z));
+
+        // arm observer
+        var item = this.agents.indexOf(index);
```
Popov72, line 645: nit: const

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -500,6 +501,11 @@ export class RecastJSCrowd implements ICrowd {
      */
     private _onBeforeAnimationsObserver: Nullable<Observer<Scene>> = null;
 
+    /**
+     *  Fires each time an agent is in reach radius of his destination
+     */
+    public onReachTargetObservable = new Observable<{agentIndex: number, destination: Vector3}>();
```
RaananW, line 507: Don't forget to clear the observable when disposed (otherwise it might not be garbage-collected)

RaananW (commented):
LGTM! Just need to clear the observable on dispose

src/Navigation/Plugins/recastJSPlugin.ts
CedricGuillemet: Good catch!

src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -727,7 +745,23 @@ export class RecastJSCrowd implements ICrowd {
...
+            this.transforms[index].position = agentPosition;
+            // check agent reach destination
+            if (this.agentDestinationArmed[index]) {
+                const dx = agentPosition.x - this.agentDestination[index].x;
+                const dz = agentPosition.z - this.agentDestination[index].z;
+                const radius = this.reachRadii[index];
+                const groundY = this.agentDestination[index].y - this.reachRadii[index];
+                const ceilingY = this.agentDestination[index].y + this.reachRadii[index];
+                const distanceXZSquared = dx * dx + dz * dz;
+                if (agentPosition.y > groundY && agentPosition.y < ceilingY && distanceXZSquared < (radius * radius)) {
```
Popov72, line 760: Why making a specific case for the y coordinate and not simply checking that `distance_squared(agent, agentDestination) < radius²`?

Popov72 (approved):

src/Navigation/Plugins/recastJSPlugin.ts
CedricGuillemet: Agents are cylinders.

sebavan (approved):

PR #10805, opened 2021-08-03
Add CSV importing and exporting of performance data!

SahilTara:
This PR adds the ability to do importing, exporting and recording of perf data. The recording can take place without having opened the window now which is super useful for mobile users, because opening the performance viewer for them will give inaccurate values for draw calls and such. The exporting and importing is done via csv to also be able to view it in something like excel if needed. 

Finishes CSV related items in #10566 

A gif of the added features:
LiveDataExportLoad

### DIFF
--- a/src/Misc/PerformanceViewer/performanceViewerCollector.ts
+++ b/src/Misc/PerformanceViewer/performanceViewerCollector.ts
@@ -211,6 +211,9 @@ export class PerformanceViewerCollector {
         this.metadataObservable.notifyObservers(this._datasetMeta);
     }
 
+    public loadFromFileData(data: string) {
+    }
+
     /**
      * Starts the realtime collection of data.
      * @param shouldPreserve optional boolean param, if set will preserve the dataset between calls of start.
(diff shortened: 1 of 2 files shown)


src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -211,6 +219,148 @@ export class PerformanceViewerCollector {
...
+     * Given a string containing file data, this function parses the file data into the datasets object.
+     * It returns a boolean to indicate if this object was successfully loaded with the data.
+     * @param data string content representing the file data.
+     * @returns true if the data was successfully loaded, false otherwise.
+     */
+    public loadFromFileData(data: string): boolean {
+        const lines =
+            data.split('\n')
+                .map((line) => (
+                    line.replace('\r', '')
```
RaananW, line 253: why not `replaceAll` (or replace with global regex) on the entire data? seems like we don't need \r's

RaananW (approved):
LGTM! just a small comment (which is not too important :-) )

Would be interesting to see if (string) template literals can be used instead of concatenating all of those small strings, but that's a whole different topic.

src/Misc/PerformanceViewer/performanceViewerCollector.ts
SahilTara: I use replace with a global regex now! replaceAll didn't seem to be available which is why I hadn't used it.

PR #10807, opened 2021-08-04
Fix create instance line mesh

sebavan:
https://forum.babylonjs.com/t/instance-color-alpha-for-linesmesh/22812

### DIFF
--- a/src/Meshes/linesMesh.ts
+++ b/src/Meshes/linesMesh.ts
@@ -263,12 +263,12 @@ export class LinesMesh extends Mesh {
      * @returns a new InstancedLinesMesh
      */
     public createInstance(name: string): InstancedLinesMesh {
-        let instance = new InstancedLinesMesh(name, this);
+        const instance = new InstancedLinesMesh(name, this);
 
         if (this.instancedBuffers) {
             instance.instancedBuffers = {};
 
-            for (var key in this.instancedBuffers) {
+            for (const key in this.instancedBuffers) {
                 instance.instancedBuffers[key] = this.instancedBuffers[key];
             }
         }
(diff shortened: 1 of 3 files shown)


src/Meshes/linesMesh.ts
```diff
@@ -263,7 +263,17 @@ export class LinesMesh extends Mesh {
...
      */
     public createInstance(name: string): InstancedLinesMesh {
-        return new InstancedLinesMesh(name, this);
+        let instance = new InstancedLinesMesh(name, this);
+
+        if (this.instancedBuffers) {
+            instance.instancedBuffers = {};
+
+            for (var key in this.instancedBuffers) {
+                instance.instancedBuffers[key] = this.instancedBuffers[key];
```
Popov72, line 272: I think you should set it to `null` instead, as it is done in `Mesh.registerInstancedBuffer`. Else the instances will use the same attributes than the master.

Popov72 (approved):

src/Meshes/linesMesh.ts
sebavan: It is actually a copy paste from instancedMeshFactory :-)

```
Mesh._instancedMeshFactory = (name: string, mesh: Mesh): InstancedMesh => {
    let instance = new InstancedMesh(name, mesh);

    if (mesh.instancedBuffers) {
        instance.instancedBuffers = {};

        for (var key in mesh.instancedBuffers) {
            instance.instancedBuffers[key] = mesh.instancedBuffers[key];
        }
    }

    return instance;
};
```

Popov72: Ok, my mistake, I thought `instancedBuffers` were vertex buffers but they are custom data instead, that can be overwritten afterward by the user!

PR #10814, opened 2021-08-05
Fix glow layer when using sRGB buffers + TextureBlock color space conversion

Popov72:
* Fix glow layer generation where the albedo/emissive texture may not be in gamma space (when using sRGB buffers)
* NME: the TextureBlock color conversions to linear/gamma space are only done if the texture is not already in the target space


src/Materials/Node/Blocks/Dual/textureBlock.ts
```diff
@@ -373,13 +373,17 @@ export class TextureBlock extends NodeMaterialBlock {
 
     private _generateConversionCode(state: NodeMaterialBuildState, output: NodeMaterialConnectionPoint, swizzle: string): void {
         if (swizzle !== 'a') { // no conversion if the output is "a" (alpha)
-            state.compilationString += `#ifdef ${this._linearDefineName}\r\n`;
-            state.compilationString += `${output.associatedVariableName} = toGammaSpace(${output.associatedVariableName});\r\n`;
-            state.compilationString += `#endif\r\n`;
+            if (!this.texture || !this.texture.gammaSpace) {
+                state.compilationString += `#ifdef ${this._linearDefineName}\r\n`;
```
RaananW, line 377: Template literals can come in handy here
```suggestion
                state.compilationString += `#ifdef ${this._linearDefineName}
${output.associatedVariableName} = toGammaSpace(${output.associatedVariableName});
#endif
`;
```

RaananW (approved):
LGTM! Just a small comment

RaananW:
Linting fails

sebavan (approved):

PR #10818, opened 2021-08-05
WebGPU: timestamp queries are not supported anymore

Popov72:
Hoping they will re-add support at some time...


sebavan (approved):
What happened ?

Popov72:
I guess it's this (from the spec):

```
Because timestamp query provides high-resolution GPU timestamp, we need to decide what constraints, if any, are on its availability.
```

PR #10819, opened 2021-08-05
Add scene extension for the performance collector.

SahilTara:
This PR Adds a scene extension for the PerformanceCollector 😄 it also adds additional defaults to the performance graph and fixes some bugs surrounding if max = min (now it shows at top of graph instead of not showing at all). Completes the "Add activemeshes, frame time and draw calls to the graph" and the "add a scene extension" tasks in #10566 

This scene extension is public hidden for now, until everything is ready for use by the user.


RaananW (approved):

RaananW:
There is a build issue

sebavan (approved):

SahilTara:
> There is a build issue

Yep there was a missing "./"

PR #10824, opened 2021-08-06
bug: no window object exists check in thin engine

nekochanoide:
This bug makes engine not compatible with workers


RaananW (approved):

nekochanoide:
Now I think that there might be more of this. So let me check first before merge :)

nekochanoide:
done

PR #10825, opened 2021-08-06
Fix offscreen canvas demo not working

Popov72:
See https://forum.babylonjs.com/t/unprotected-access-to-window-in-thin-engine/22880


Popov72:
Already fixed by https://github.com/BabylonJS/Babylon.js/pull/10824

PR #10833, opened 2021-08-09
Added some @since 5.0.0 tags

Popov72:
I have reviewed the what's new file and added some `@since` 5.0.0 tags wherever I could.

I have also fixed some glTF extension comments for extensions that are now merged into the Khronos repo.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -12,26 +12,26 @@
 
 ### General
 
-- Added zoomToMouseLocation on ArcRotateCamera ([lovettchris](https://github.com/lovettchris))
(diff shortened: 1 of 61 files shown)


RaananW:
tsdoc doesn't support the @since tag...

https://tsdoc.org/

RaananW:
Following up on this - we had this discussion (where i had the same answer) https://forum.babylonjs.com/t/since-tag-versioning/14393

I wonder why it passed back then. I'll have to investigate that.

Popov72:
That's strange because it works for this class WebXRDomOverlay: https://doc.babylonjs.com/typedoc/classes/babylon.webxrdomoverlay

Popov72:
Ah, I think I know: it's supported only for methods and classes, and it fails for properties.

RaananW (approved):
LGTM :-)

PR #10834, opened 2021-08-09
Fix error handling in basis texture

### LINKED ISSUE #10829: If basis transcoders can't be loaded, no error is surfaced
- Bug repro on [playground](https://playground.babylonjs.com): https://www.babylonjs-playground.com/#JTG6VF#2
- Note that you may have to open a new tab to get it to repro (if you have the transcoders cached, it'll work though you'll see an error in the console)
- Expected result:
I would expect the error handler I pass to the Texture object to be called. 
- Current result:
The error handler I pass to the Texture object is not called.

RaananW:
If the basis wasm and js files failed to load no error would have been triggered and the texture load would have thrown an exception. 

After merging, this PG - https://www.babylonjs-playground.com/#JTG6VF#2 will show the error window and not throw an exception in the console.

Fixes #10829


Popov72 (approved):
LGTM!

PR #10838, opened 2021-08-09
Added custom string events to performance viewer

SahilTara:
This PR adds the ability to register custom string events and let the collector know that these events have triggered.

The registerEvent will give back an object which in this case is just the id, we do this to be a bit less error prone to spelling mistakes and passing just any string. 

The code used to power it was as simple as this:
image

Here is a video of it in action!

https://user-images.githubusercontent.com/32103099/128747185-96d912d5-57ea-4d68-ae2a-2b7f502ccb5e.mp4


RaananW:
Is this just tracking if an event was triggered? there is no attached value to an event?
You have a click example - a mouse click is 0 or 1 (i.e. clicked or not clicked) but touch-sensitive controllers (like an xbox or xr controller) will provide a value while a click is trigered.

RaananW (approved):

PR #10839, opened 2021-08-09
Add tooltips to playground menu buttons

darraghjburke:
Adds more informative tooltips to menu buttons in the playground. Some of the button's actions are not immediately obvious so I tried to provide hints to make it more clear.

### DIFF
--- a/Playground/src/components/commandBarComponent.tsx
+++ b/Playground/src/components/commandBarComponent.tsx
@@ -75,7 +75,7 @@ export class CommandBarComponent extends React.Component<ICommandBarComponentPro
         var engineOptions = [
             {
                 label: "WebGL2",
-                tooltip: "Use WebGL 2 API",
+                tooltip: "Use WebGL 2 Renderer",
                 storeKey: "engineVersion",
                 isActive: activeEngineVersion === "WebGL2",
                 onClick: () => {
@@ -85,7 +85,7 @@ export class CommandBarComponent extends React.Component<ICommandBarComponentPro
             },
             {
                 label: "WebGL",
-                tooltip: "Use WebGL 1 API",
+                tooltip: "Use WebGL 1 Renderer",
                 storeKey: "engineVersion",
                 isActive: activeEngineVersion === "WebGL",
                 onClick: () => {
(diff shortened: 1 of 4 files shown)


Playground/src/components/commandBarComponent.tsx
```diff
@@ -74,6 +75,7 @@ export class CommandBarComponent extends React.Component<ICommandBarComponentPro
         var engineOptions = [
             {
                 label: "WebGL2",
+                tooltip: "Use WebGL 2 API",
```
deltakosh, line 78: I would say renderer instead of API maybe?

Playground/src/components/hamburgerMenu.tsx
```diff
@@ -95,6 +96,7 @@ export class HamburgerMenuComponent extends React.Component<IHamburgerMenuCompon
         var engineOptions = [
             {
                 label: "WebGL2",
+                tooltip: "WebGL 2.0",
```
deltakosh, line 99: maybe use the longer text here as well (like earlier)

deltakosh (changes requested):
super great..just a few tiny nit picking comments

Playground/src/components/commandBarComponent.tsx
darraghjburke: done :)

Playground/src/components/hamburgerMenu.tsx
darraghjburke: yes sorry! fixed

Playground/src/components/commandDropdownComponent.tsx
```diff
@@ -96,7 +97,7 @@ export class CommandDropdownComponent extends React.Component<ICommandDropdownCo
                                                     Utilities.StoreStringToStore(this.props.tooltip, m.label);
                                                     this.setState({isExpanded: false, activeState: m.label});
                                                 }
-                                            }} title={m.label}>
+                                            }} title={m.tooltip}>
```
RaananW, line 100: you can stay defensive here and do `title={m.tooltip || ''}`, but if everything has a tooltip it should be fine.

RaananW (approved):
LGTM! just a small note

Playground/src/components/commandDropdownComponent.tsx
darraghjburke: sure! I'll use the label as a fallback actually

deltakosh (approved):

PR #10117, opened 2021-04-01
Add support for cannon-es in plugin

frankieali:
closes #9810

Check for cannon-es method `addBody` for generating a physics body. Fallback to cannonjs method `add`.

PR #10844, opened 2021-08-10
Add support for cannon-es method removeBody()

Faber-smythe:
Solves a problem similar to the one fixed with #10117

The plugin should check for _cannon-es_ method **`world.removeBody()`** when getting rid of physics body. Falls back to _cannonjs_ method **`world.remove()`**.

Issue has been spotted and discussed on [babylon's forum](https://forum.babylonjs.com/t/cant-dispose-of-cloned-and-impostorized-meshes/22906/5).


RaananW (approved):
Thanks a lot!

Issue #10464, opened 2021-06-04
Leverage compute pressure


deltakosh:
Source: https://web.dev/compute-pressure/

The overall idea would be to expose these values in the inspector and to also uses them as targets for the SceneOptimizer

We should surface them in the API for the users to be able to simply consume them as well

I would also add a new tool in the inspector: the perf monitor. A graph displaying bars during the last x seconds (sliding window) of things like:
- cpu load
- gpu frame time
- fps
- etc...
These options can be turned on/off in the inspector

A bit like that:
![image](https://user-images.githubusercontent.com/1306056/121087047-cb052300-c798-11eb-9895-3a62afdc67d9.png)

but inside the inspector

deltakosh:
I think we could expose that info through 2 ways:
- a direct request..like engine.computePressure property
- an observable approach like: onComputePressureChanged (and we use internally 4 thresholds: 0.25, 0.50, 0.75 and 0.9)
Thoughts?

sebavan:
Yup I kind of like it and it should be doable to add it into native to in this way @bghgary  ?

sebavan:
Also it does not seem there is a direct query available so we might need a granular array of default values to convert an get callbacks often :-)

bghgary:
> Yup I kind of like it and it should be doable to add it into native to in this way @bghgary ?

Doesn't seem hard. Should be doable.

PR #10852, opened 2021-08-11
Add CPU Strategy to performance graph and onComputePressureChanged to scene

SahilTara:
This PR adds the ability to track cpu utilization in the performance graph, and the ability to subscribe to compute pressure changes by using the scene object.

Completes the "connect cpu usage to graph " in #10566 
Adds the "onComputePressureChanged" mentioned in #10464

### DIFF
--- a/src/scene.ts
+++ b/src/scene.ts
@@ -38,6 +38,7 @@ import { FileTools, LoadFileError, RequestFileError, ReadFileError } from './Mis
 import { IClipPlanesHolder } from './Misc/interfaces/iClipPlanesHolder';
 import { IPointerEvent } from "./Events/deviceInputEvents";
 import { LightConstants } from "./Lights/lightConstants";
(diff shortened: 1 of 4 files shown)


src/scene.ts
```diff
@@ -1504,6 +1505,17 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         if (!options || !options.virtual) {
             this._engine.onNewSceneAddedObservable.notifyObservers(this);
         }
+
+        if ('ComputePressureObserver' in window) {
```
deltakosh, line 1509: maybe having this check as a static on the ComputePressureObserver class?

SahilTara: Good idea!

RaananW:
I find it a bit confusing that the babylon class is named the same as the `window` object.

PR #10857, opened 2021-08-12
NME: Fix output names of PBRMetallicRoughness block

Popov72:
See https://forum.babylonjs.com/t/how-to-use-lightmaps-with-the-node-shader-and-the-pbrmetallicroughness-node/22990/7


RaananW (approved):
Just making sure -  no back compat issue here?

Popov72:
I don't think so, it was a bug, using the block by code would fail.

PR #10864, opened 2021-08-13
[GE] Grid Implementation

msDestiny14:
-Be able to add items into a grid using the SE drag and drop
-Be able to modify the grid control under the grid tab.
-Be able to modify the grid a GUI is inserted in on the object itself.

Also bug fix for SE
https://github.com/BabylonJS/ThePirateCove/issues/14


guiEditor/src/components/parentingPropertyGridComponent.tsx
```diff
@@ -1,75 +1,50 @@
...
+    control: Control,
+    lockObject: LockObject,
+    onPropertyChangedObservable?: Observable<PropertyChangedEvent>
 }
 
 export class ParentingPropertyGridComponent extends React.Component<IParentingPropertyGridComponentProps> {
     constructor(props: IParentingPropertyGridComponentProps) {
         super(props);
+        const cellInfo = (this.props.control.parent as Grid).getChildCellInfo(this.props.control);
+        this._rowNumber = parseInt(cellInfo.substring(0, cellInfo.search(":")));
```
RaananW, line 20: On certain cases this will be NaN (next line as well). Would be great to stay defensive and make sure it defaults to a value during construction.

guiEditor/src/components/parentingPropertyGridComponent.tsx
```diff
@@ -1,75 +1,50 @@
...
-    public removeChildGui(childNode: Control, parentNode: Control) {
-        (parentNode as Container).removeControl(childNode);
-        this.props.globalState.guiTexture.addControl(childNode);
+    _columnNumber: number;
+    _rowNumber: number;
+
+    updateGridPosition()
+    {
+        const grid = this.props.control.parent as Grid;
+        this.props.control.parent?.removeControl(this.props.control);
```
RaananW, line 29: this is the reference to grid, so if it doesn't exist it should fail. Why not use grid here?

RaananW (approved):
Just 2 comments, merge when resolved

PR #10868, opened 2021-08-15
Core engine: Abstract the framebuffers

Popov72:
I finally did not use the name `HardwareFramebufferWrapper` to abstract the feature but `RenderTargetWrapper` instead because `framebuffer` is specific to WebGL and does not exist in WebGPU: `RenderTargetWrapper` seems better to me (`RenderTarget` could also be right I think).

Those are breaking changes, but mostly deep in the engine, so most of the users should not be impacted: I only had to update 3 lines of code in the Babylon libraries (2 changes in **inspector/src/textureHelper.ts** and one in **nodeEditor/src/sharedComponents/textureLineComponent.tsx**) and no change in the validation test suite.

As acknowledged, I have moved `updateRenderTargetTextureSampleCount` from `Engine` to `engine.renderTarget.ts`.

I miss to add some docs in the new class + update the **what's new** file, but I'm waiting for your inputs before moving on.


deltakosh (approved):

RaananW:
Is this ready to be merged?

Popov72:
It's ok on my side but maybe we should wait for @sebavan review when he comes back from holidays?

deltakosh:
let's wait for seb

inspector/src/textureHelper.ts
```diff
@@ -68,7 +68,7 @@ export class TextureHelper {
         if (rttInternalTexture && internalTexture) {
             const samplingMode = internalTexture.samplingMode;
             texture.updateSamplingMode(Texture.NEAREST_NEAREST_MIPNEAREST);
-            scene.postProcessManager.directRender([lodPostProcess], rttInternalTexture);
+            scene.postProcessManager.directRender([lodPostProcess], rtt.renderTarget);
```
sebavan, line 71: in this case why is the code looking for rttInternalTexture in the previous if ? it is a bit confusing to me ?

sebavan (commented):
Would be great to check how bigger is thinengine ???

src/Materials/Textures/internalTexture.ts
```diff
@@ -215,10 +204,6 @@ export class InternalTexture {
     /** @hidden */
     public _compression: Nullable<string> = null;
     /** @hidden */
-    public _generateStencilBuffer: boolean = false;
-    /** @hidden */
-    public _generateDepthBuffer: boolean = false;
-    /** @hidden */
     public _comparisonFunction: number = 0;
```
sebavan, line 207: Should this still be here as well ?

sebavan:
And the latest point I can think off is the lost context scenario to double check ?

src/Materials/Textures/internalTexture.ts
Popov72: The `comparisonFunction` is not part of the RTT wrapper(?) We can have a texture which is not a RTT that has a comparison function.

Popov72:
> And the latest point I can think off is the lost context scenario to double check ?

I already did it, I checked 30 or 40 PGs for context lost/restore and they were all ok.

sebavan (approved):
@RaananW wait on the next commit changing imports to declare :-) from @Popov72

Popov72:
Imports removed!

Issue #10846, opened 2021-08-10
Update  MRTK-based controls to MRTK 2.0 specs


rickfromwork:
There's lots that still needs to be done, so this will be a bucket task to contain everything. Tasks are taken from MRTK's guide here:
https://docs.microsoft.com/en-us/windows/mixed-reality/mrtk-unity/?view=mrtkunity-2021-05#ux-building-blocks

- [ ] BoundingBoxGizmo: Introduce MRTK 2.0 visuals, behind a flag. Use MRTK 1 (current) by default
     - Replace each interaction joint with slender bars along the relevant edges
     - Hide non-relevant bars on interaction with a given bar
     - Draw faint bounding box when visuals are active
     - Hide most visuals when not active (ray/hand is not near)
- [ ] TouchHolographicMenu: Introduce ability to add more controls, like a backplated slider
- [ ] TouchHolographicMenu: Add manipulation visuals around edges
- [ ] TouchHolographicMenu: Add snap-to-world toggle functionality

Quality of life changes:
- [ ] TouchHolographicButton: Text scaling when sizing UI for AR/VR is too small
- [ ] HolographicSlate: Fix slate depth on browser (appears inside near clipping plane)
- [ ] Slider3D: Add text, description to slider thumb

Completed:
- [x] TouchHolographicButton: pressing button should compress front, not scale button
- [x] VolumeBasedPanel: Needs support for tall panels (just one column)
     - Already supported, just hidden. Explicitly setting VBP.rows will populate the panel one column at a time instead of one row at a time
- [x] VolumeBasedPanel: Expose a way to customize padding amounts
     - Padding is determined at TouchHolographicMenu level, change was made there
- [x] SixDofDragBehavior: Add support for near interaction rotation using wrist joint orientation
     - Pass orientation of controller/hand through via pick data, instead of just direction of raycast. This way we can also enable proper rotation manipulation for motion controllers.
- [x] HolographicSlate: Have corner resizers respect the slate's aspect ratio
- [x] HolographicSlate: Have scaling applied to slate be applied to the slate's bounding box as well
- [x] Slider3D: Add backplate support
     - MRTK appears to use touch instead of grab when a backplate is present
     - Visuals on the thumb also change to a circular cylinder
- [x] Slider3D: Change slider thumb visual to cylindrical oval
- [x] TouchHolographicMenu: Backplate visibility shouldn't be tied to all buttons that share resources
[...]

PR #10879, opened 2021-08-17
Updating TouchHolographicMenu to support changing margin size

rickfromwork:
TouchHolographicMenu supported changing the margin around buttons already, but the changes only took effect when adding or removing buttons to the menu. This change allows the margin's changes to take effect as soon as they're changed.

Done as part of #10846


RaananW (approved):
LGTM

PR #10880, opened 2021-08-17
Fixing resizing behavior of HolographicSlate

rickfromwork:
This fixes two issues:
- Resizing the HolographicSlate object after creation would not correctly resize the associated slate bounding box
- Resizing the slate using the corner manipulators would not maintain the aspect ratio, and would instead scale at a 1:1 pixel ratio (instead of a 1:1% ratio)

Also corrects two spots where the description for a Gizmo property getter had double spaces.

Done as part of #10846


RaananW:
Missing the `Scene` import, otherwise LGTM

RaananW (approved):

PR #10904, opened 2021-08-20
Improve bounding box management with a dirty state

deltakosh:


src/Meshes/abstractMesh.ts
```diff
@@ -1099,14 +1099,43 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
             return this._masterMesh.getBoundingInfo();
         }
 
-        if (!this._boundingInfo) {
-            // this._boundingInfo is being created here
+        if (!this._boundingInfo || this._boundingInfoIsDirty) {
```
bghgary, line 1102: would it work if you default _boundingInfoIsDirty to true so that you only need to check the flag?

src/Meshes/abstractMesh.ts
```diff
@@ -1099,14 +1099,43 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
         // cannot be null.
         return this._boundingInfo!;
     }
 
+    /**
+     * Overwrite the current bounding info
+     * @param boundingInfo defines the new bounding info
+     * @returns the current mesh
+     */
+    public setBoundingInfo(boundingInfo: BoundingInfo): AbstractMesh {
```
bghgary, line 1116: Not sure these should be public. They were public _ before right?

bghgary (approved):
Minor comments. LGTM

src/Meshes/abstractMesh.ts
deltakosh: yes but I use that change as an opportunity to open the API a bit

src/Meshes/abstractMesh.ts
deltakosh: good idea

src/Particles/solidParticle.ts
```diff
@@ -241,13 +246,13 @@ export class SolidParticle {
      * @returns true if it intersects
      */
     public intersectsMesh(target: Mesh | SolidParticle): boolean {
-        if (!this._boundingInfo || !target._boundingInfo) {
+        if (!this._boundingInfo || !target.getBoundingInfo()) {
```
Popov72, line 249: Should be:
```javascript
if (!this._boundingInfo || !target.hasBoundingInfo) {
```
instead, as `getBoundingInfo()` can't return `null` (I guess you will need to add a `hasBoundingInfo` to `SolidParticle`, however).

Popov72 (approved):

src/Particles/solidParticle.ts
deltakosh: I can do that

src/Meshes/abstractMesh.ts
```diff
@@ -663,8 +663,8 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
 
     /** @hidden */
     public _masterMesh: Nullable<AbstractMesh> = null;
-    /** @hidden */
-    public _boundingInfo: Nullable<BoundingInfo> = null;
+    private _boundingInfo: Nullable<BoundingInfo> = null;
```
barroij, line 666: Hi @deltakosh 
Would it be possible to turn these 2 props back to public ?

They are hidden and they are the only private props in this class. And migrating to BabylonJs 5 breaks my code :)
Cheers

barroij: (public or protected at least)

deltakosh: what broke your code? (You were using _features right? :))

barroij: I have my own derivation of `Mesh` that overwrites `_refreshBoundingInfo` and then accesses and sets `_boundingInfo`.

The error message is "Property '_boundingInfo' is private and only accessible within class 'AbstractMesh'."

barroij: (of course I can work-around that with some ugly `any`)

deltakosh: No worries to flag it as protected:) wanna do a PR?

PR #10906, opened 2021-08-20
Native resource handles + command buffer

ryantrem:
The Babylon Native half of this change is here: https://github.com/BabylonJS/BabylonNative/pull/873

Engine related resources created by Babylon Native (such as textures, vertex arrays, index buffers, etc.) are all treated as opaque objects. JavaScript doesn't do anything with them other than store them and pass them back to other Babylon Native functions. With this change, those opaque resources are now assumed to be numeric handles (see the associated Babylon Native changes). Since these handles are numeric, they can now easily be encoded into a "command buffer." The command buffer accumulates commands over the course of a single `scene.render` call, and then "submits" the command buffer to native at the end. For more complex scenes, this reduces the number of calls into native from thousands to just one. Additionally, since the command arguments are just read as numeric values out of typed arrays, the cost of retrieving args on the native side is pretty much eliminated. When profiling the JavaScript with Safari on iOS (for a scenario where I have a single mesh with ~250 submeshes), it claims that ~65% of CPU time is spent making those native calls. After this change, it claims 0.4% of CPU time is spent making those native calls. That said, with the profiler not attached, this change does not triple the frame rate (as the previous numbers might suggest), but does increase frame rate by about 50% which is still substantial.

There is a bunch of work to do to finish this change:
- [ ] Convert a bunch more functions to use the command buffer (not everything, but probably all functions that set state or delete resources (creating resources needs to continue to be synchronous, or the numeric resource handles need to be generated on the JS side and passed to the native side)).
- [ ] The command and argument buffers are just a fixed size that was big enough for my test scene. The buffers need to grow as needed, and we need a heuristic for shrinking them as well (e.g. at the end of a scene render, after calling `submitCommandBuffer`, maybe call a `resizeCommandBuffers` function that shrinks them down to the size needed for the last submit, or maybe the max of the last five submits, or something like that).
- [ ] Delete shader programs at the right time (maybe from `_releaseEffect`). Probably need input from @bghgary on this.
- [ ] We may want to have some kind of command buffer validation. The command and command arg encoding and decoding must happen in the same order or things will be totally broken, and this is hard to diagnose. Command validation could do something like record the order of encoding operations in another buffer (a validation buffer) which the native side could use to ensure decoding happens in the same order. We'd want this to be an explicit opt-in from JS, like a call to `nativeEngine.enableCommandBufferValidation()` that perhaps would replace the command buffer encoding functions to do extra work (so there is no overhead when not doing validation) and pass the validation buffer to native so it has it and knows to do the extra validation work.
- [ ] Remove all the commented out code (left there to make testing the old code path easy).
- [ ] Merge master.

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -127,6 +127,10 @@ interface INativeEngine {
     readonly COMMAND_SETTEXTURE: number;
     readonly COMMAND_BINDVERTEXARRAY: number;
     readonly COMMAND_SETSTATE: number;
+    readonly COMMAND_SETFLOAT: number;
+    readonly COMMAND_SETFLOAT2: number;
+    readonly COMMAND_SETFLOAT3: number;
+    readonly COMMAND_SETFLOAT4: number;
 
     dispose(): void;
 
@@ -1861,7 +1865,11 @@ export class NativeEngine extends Engine {
             return false;
         }
 
-        this._native.setFloat(uniform, value);
+        //this._native.setFloat(uniform, value);
+        this._commandBufferEncoder.beginEncodingCommand(this._native.COMMAND_SETFLOAT);
+        this._commandBufferEncoder.encodeCommandArgAsUInt32(uniform);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(value);
+        this._commandBufferEncoder.finishEncodingCommand();
         return true;
     }
 
@@ -1870,7 +1878,12 @@ export class NativeEngine extends Engine {
             return false;
         }
 
-        this._native.setFloat2(uniform, x, y);
+        //this._native.setFloat2(uniform, x, y);
+        this._commandBufferEncoder.beginEncodingCommand(this._native.COMMAND_SETFLOAT2);
+        this._commandBufferEncoder.encodeCommandArgAsUInt32(uniform);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(x);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(y);
+        this._commandBufferEncoder.finishEncodingCommand();
         return true;
     }
 
@@ -1879,7 +1892,13 @@ export class NativeEngine extends Engine {
             return false;
         }
 
-        this._native.setFloat3(uniform, x, y, z);
+        //this._native.setFloat3(uniform, x, y, z);
+        this._commandBufferEncoder.beginEncodingCommand(this._native.COMMAND_SETFLOAT3);
+        this._commandBufferEncoder.encodeCommandArgAsUInt32(uniform);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(x);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(y);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(z);
+        this._commandBufferEncoder.finishEncodingCommand();
         return true;
     }
 
@@ -1888,7 +1907,14 @@ export class NativeEngine extends Engine {
             return false;
         }
 
-        this._native.setFloat4(uniform, x, y, z, w);
+        //this._native.setFloat4(uniform, x, y, z, w);
+        this._commandBufferEncoder.beginEncodingCommand(this._native.COMMAND_SETFLOAT4);
+        this._commandBufferEncoder.encodeCommandArgAsUInt32(uniform);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(x);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(y);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(z);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(w);
+        this._commandBufferEncoder.finishEncodingCommand();
         return true;
     }
 
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -211,7 +211,10 @@ interface INativeEngine {
     setViewPort(x: number, y: number, width: number, height: number): void;
     setStencil(mask: number, stencilOpFail: number, depthOpFail: number, depthOpPass: number, func: number, ref: number): void;
 
-    submitCommandBuffer(commandCount: number, commandBuffer: Uint8Array, uint32ArgBuffer: Uint32Array, float32ArgBuffer: Float32Array): void;
+    setCommandBuffer(commandBuffer: Uint8Array): void;
+    setCommandUint32Buffer(commandBuffer: Uint32Array): void;
+    setCommandFloat32Buffer(commandBuffer: Float32Array): void;
+    submitCommandBuffer(commandCount: number): void;
 }
 
 class NativePipelineContext implements IPipelineContext {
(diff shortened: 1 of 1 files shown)


src/scene.ts
```diff
@@ -3993,6 +3993,8 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
             return;
         }
 
+        this.onBeforeCameraRenderObservable.notifyObservers(camera);
```
Popov72, line 3996: I don't think you can move the notification, it is a breaking change.

Current notified code expects that `_activeCamera`, viewport, `_renderId`, etc are setup when they are called. By moving the notification here it does not stand anymore.

src/scene.ts
```diff
@@ -4123,7 +4123,6 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
     private _processSubCameras(camera: Camera): void {
         if (camera.cameraRigMode === Constants.RIG_MODE_NONE || (camera.outputRenderTarget && camera.outputRenderTarget.getViewCount() > 1 && this.getEngine().getCaps().multiview)) {
             this._renderForCamera(camera);
-            this.onAfterRenderCameraObservable.notifyObservers(camera);
```
Popov72, line 4126: This is a breaking change, but probably a bearable one as it's only a matter of using `onAfterCameraRenderObservable` instead of `onAfterRenderCameraObservable`.

[EDIT] See comment below, it is in fact a breaking change in WebXR to remove the `onAfterRenderCameraObservable` observable.

src/scene.ts
```diff
@@ -4139,7 +4138,6 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         // Use _activeCamera instead of activeCamera to avoid onActiveCameraChanged
         this._activeCamera = camera;
         this.setTransformMatrix(this._activeCamera.getViewMatrix(), this._activeCamera.getProjectionMatrix());
-        this.onAfterRenderCameraObservable.notifyObservers(camera);
```
Popov72, line 4142: This one is a more problematic breaking change because the camera that is passed is not the camera passed to `_renderForCamera` in the loop above.

I think the code should be:
```javascript
if (!camera._useMultiviewToSingleView) {
    this.onAfterCameraRenderObservable.notifyObservers(camera);
}
```
The `if` is to avoid calling two times `onAfterCameraRenderObservable` in the `_useMultiviewToSingleView` case, as it seems it is what should be done. HOWEVER, it will break WebXR as the `onAfterRenderCameraObservable` observable is used by `WebXRExperienceHelper` (pinging @RaananW who will know better than me)...

So, for the time being, the existing code should be kept "as is" (        `this.onAfterRenderCameraObservable.notifyObservers(camera);`) to avoid breaking things.

src/scene.ts
RaananW: Though I totally agree with this change, this is in fact a breaking change. I have tried fixing that in an earlier PR (can't find the reference right now, sorry), but it has side effects that will influence different scenes. The main issue when I tested it was the setViewport, that until now was expected to be called before the observable is triggered, and in this case is executed after.

RaananW (commented):
Referring to the onAfter and onBefore camera render - 

It would be great to organize this. We have onAfterCameraRender and onAfterRenderCamera observables, which are confusing, and onBefore is in fact being triggered after a few changes were made in the engine/scene. The issue here, as I wrote in one of the comments, is that there are side effects to this change. 

Unifying both onAfter observables should be possible. The difference between them, for XR at least, is what camera triggers them. It is possible to solve this in XR, changing the observable, if it is guaranteed that all 3 cameras (parent and 2 rig cameras) are triggering notifying at the correct moment. As we are checking what camera triggered this event, it shouldn't break any of the implementations existing in the framework, but we need to check that older code will still run correctly.

A quick inspection of the playground search shows very little usage of `onAfterRenderCameraObservable`. 5 playgrounds are using it, none of them rig-camera related. They should still work if we unify the observables.
`onAfterCameraRenderObservable` (which is the one we should not change at all) is used in 55 playgrounds (https://doc.babylonjs.com/playground?q=onAfterCameraRenderObservable&type=code) and is used more often.

There is no good solution I see for that, apart from breaking the `onAfterRenderCameraObservable` behavior. Same goes to the onBefore observable, which will influence a lot of existing code. Most of it is in the framework, and from looking at playgrounds code using this observable, none of the PGs i could find expects the code that is currently running before the onBefore is triggered to run.

@deltakosh - would be great to hear your opinion as well. I believe most of the side effects are framework related and can be solved in the core code, but it is still breaking behavior for users who use this observable.

rgerd:
@Popov72 @RaananW Thank you for the context, I was just trying to get the observables to behave how I expected with stereo rendering and didn't spend enough time looking into the expected behavior engine-wide. I'm going to back out the change for this PR and find another way of approximating what we're trying to achieve for the command buffer scope without causing any breaking changes, since it seems the implications of the observables changes lie out of the scope of this already massive PR-- although I do think it's an important topic of discussion that I'd be happy to open in another PR.

src/Engines/nativeEngine.ts
```diff
@@ -1648,7 +1949,12 @@ export class NativeEngine extends Engine {
             return false;
         }
 
-        this._native.setFloatArray2(uniform, array);
+        //this._native.setFloatArray2(uniform, array);
+        this._commandBufferEncoder.beginEncodingCommand(this._native.COMMAND_SETFLOATARRAY2);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(uniform);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(array.length);
```
CedricGuillemet, line 1955: Why encoding an array length with a float instead of a uint32? This scares me as soon as the length is getting big.

ryantrem: Good eye @CedricGuillemet! This was actually an error. It is encoded into the float buffer and decoded from the uint buffer (which doesn't work). I already have a fix for this locally, but haven't pushed those changes yet. I will probably do so today.

src/Engines/nativeEngine.ts
```diff
@@ -1765,7 +2005,13 @@ export class NativeEngine extends Engine {
...
         }
 
-        this._native.setFloat4(uniform, x, y, z, w);
+        this._commandBufferEncoder.beginEncodingCommand(this._native.COMMAND_SETFLOAT4);
+        this._commandBufferEncoder.encodeCommandArgAsUInt32(uniform);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(x);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(y);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(z);
+        this._commandBufferEncoder.encodeCommandArgAsFloat32(w);
+        this._commandBufferEncoder.finishEncodingCommand();
```
RaananW, line 2014: (9This refers to all of the calls to `this._commandBufferEncoder`) can all of these calls be done by a single function accepting optional parameters? seems like those calls are constantly being repeated but with different parameters

RaananW (commented):
there's also a conflict to be solved. github's merge tool is not the best to say the least, would be better to merge locally instead of "resolving conflicts" on github.

src/scene.ts
ryantrem: Rolled back these changes.

src/scene.ts
ryantrem: Rolled back these changes.

src/scene.ts
ryantrem: Rolled back these changes.

src/Engines/nativeEngine.ts
ryantrem: If I understand, the idea would be to be able to pass multiple args to `encodeCommandArgAsFloat32` (for example), and the function would have to check if those values were actually passed in before trying to encode them into the buffer, is that right? If so, I think we could try it, we'd just have to see if it impacts perf at all.

PR #10911, opened 2021-08-23
Allow skipping the rtt initial clearing

RaananW:
This is important in XR scenarios.
See https://github.com/BabylonJS/BabylonReactNative/pull/278


Popov72:
The way to skip the initial clearing was to add an empty observer for the `onClearObservable`, but having a toggle for that seems nice to me.

Popov72 (approved):

PR #10921, opened 2021-08-24
add POINTERDOUBLETAP mask for BaseCameraPointerInput observer

nekochanoide:
Make observer of BaseCameraPointerInput check POINTERDOUBLETAP with respect of _pointerInput function that handles this type by call overrided onDoubleTap method


nekochanoide:
Also, I had problem with this file, because it has LF line endings 😕

RaananW (approved):

PR #10926, opened 2021-08-25
Modify ArcRotateCamera and FreeCamera to account for which button is pressed

PolygonalSun:
This PR contains a modification for the ArcRotateCamera and FreeCamera's mouse movement behavior to keep track of what button was pressed to initiate a drag/rotate camera movement.  Touch should behave the same as it works differently from Mouse.

### DIFF
--- a/src/Cameras/Inputs/BaseCameraPointersInput.ts
+++ b/src/Cameras/Inputs/BaseCameraPointersInput.ts
@@ -32,6 +32,8 @@ export abstract class BaseCameraPointersInput implements ICameraInput<Camera> {
      */
     protected _buttonsPressed: number;
 
+    private _currentActiveButton: number = -1;
+
     /**
      * Defines the buttons associated with the input to handle camera move.
      */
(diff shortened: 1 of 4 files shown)


src/Cameras/Inputs/BaseCameraPointersInput.ts
```diff
@@ -126,7 +131,7 @@ export abstract class BaseCameraPointersInput implements ICameraInput<Camera> {
                 }
             } else if (p.type === PointerEventTypes.POINTERDOUBLETAP) {
                 this.onDoubleTap(evt.pointerType);
-            } else if (p.type === PointerEventTypes.POINTERUP && srcElement) {
+            } else if ((p.type === PointerEventTypes.POINTERMOVE && srcElement && p.event.button === this._currentActiveButton && !isTouch) || (p.type === PointerEventTypes.POINTERUP && srcElement)) {
```
deltakosh, line 134: help understand why you have !touch?

PR #10954, opened 2021-08-30
Add Slider3d Backplate support and update visual design

tayomadein:
This PR does the following
- Adds backplate support for the slider3d class. Backplate visibility is false by default.
- Update visual design to MRTK v3.0

Preview:
image
With Backplate

image
Without Backplate

Slider Animation
slider3d-playground

Assets PR: https://github.com/BabylonJS/Assets/pull/51
Done as part of https://github.com/BabylonJS/Babylon.js/issues/10846


gui/src/3D/materials/mrdl/mrdlBackplateMaterial.ts
```diff
@@ -217,7 +217,7 @@ export class MRDLBackplateMaterial extends PushMaterial {
         this._activeEffect.setFloat("_Iridescence_Intensity_", 0.45);
         this._activeEffect.setFloat("_Iridescence_Edge_Intensity_", 1);
         this._activeEffect.setDirectColor4("_Iridescence_Tint_", new Color4(1, 1, 1, 1));
-        this._activeEffect.setTexture("_Iridescent_Map_", new Texture("https://github.com/rgerd/TestAssets/blob/master/PNG/BackplateIridescence.png", this.getScene()));
+        this._activeEffect.setTexture("_Iridescent_Map_", new Texture("https://raw.githubusercontent.com/rgerd/TestAssets/master/PNG/BackplateIridescence.png", this.getScene()));
```
sebavan, line 220: Could this be moved in the babylon assets repo ?

sebavan: also I wonder if it should not be embedded in the code in some ways or configurable through a static for instance ? @RaananW, any thoughts ?

sebavan (commented):
This sounds like a huge change, how big are the breaking changes introduced ?

gui/src/3D/controls/slider3D.ts
```diff
@@ -119,64 +135,108 @@ export class Slider3D extends Control3D {
...
+                if (!!sliderThumbModel) {
+                    sliderThumbModel.isPickable = true;
+                    sliderThumbModel.position.z = -0.115;
+                    sliderThumbModel.scaling = new Vector3(0.025, 0.3, 0.6);
+                    sliderThumbModel.position.x = this._convertToPosition(this.value);
+                    sliderThumbModel.addBehavior(this._createBehavior());
+                    if (!!this._sliderThumbMaterial) {
+                        sliderThumbModel.material = this._sliderThumbMaterial;
+                    }
+                    this._sliderThumb = sliderThumbModel;
```
RaananW, line 225: This is being set in line 231 already. It is then re-set here. Is there a need to set it in lines 229-231?
And if it needed, you will need to dispose the box that was created in 229.

gui/src/3D/materials/mrdl/mrdlBackplateMaterial.ts
```diff
@@ -0,0 +1,278 @@
...
+
+        this._activeEffect = effect;
+
+        // Matrices
+        this.bindOnlyWorldMatrix(world);
+        this._activeEffect.setMatrix("viewProjection", this.getScene().getTransformMatrix());
+        this._activeEffect.setVector3("cameraPosition", this.getScene().activeCamera!.position);
+
+        // "Round Rect"
+        this._activeEffect.setFloat("_Radius_", 0.3);
```
RaananW, line 195: can you extract those to (mutable?) constants outside of this function? Or is it not needed?

gui/src/3D/materials/mrdl/mrdlBackplateMaterial.ts
RaananW: Yes, to assets and have it configurable for local hosting.

RaananW (commented):
My main comments are about the fixed constants that exist in all materials.
Are they reusable (between the different materials maybe?)? should they be mutable?

rickfromwork:
Can you post some pictures of what the slider looks like now? I'm curious to see what the final visuals look like. Also can you add a pic of a slider with no backplate, if possible?

sebavan:
@tayomadein, any update on this PR or should I flag it as draft for the time being ?

gui/src/3D/materials/mrdl/mrdlBackplateMaterial.ts
tayomadein: The asset can be moved to repo, (Created this PR: https://github.com/BabylonJS/Assets/pull/51) and I can update this file once that is merged.

tayomadein:
@rickfromwork @RaananW I have updated the description.
@sebavan I am currently implementing the fix based on the comments. I should push my changes in a few hours.

sebavan:
@tayomadein it all LTGM. I guess it could be merged as soon as linting is ok ?

sebavan (approved):

sebavan:
ohhhhh and as @RaananW mentioned https://github.com/BabylonJS/Babylon.js/pull/10954/files/2dcd4da1278799b0518110575f206703a0b94837#r708161061

gui/src/3D/controls/slider3D.ts
RaananW:
```suggestion
                    if(this._sliderThumb) {
                        this._sliderThumb.dispose();
                    }
                    this._sliderThumb = sliderThumbModel;
```

tayomadein: @RaananW I like your suggestion. However, it doesn't solve the original reason on why I need this workaround. 

What I have noticed is that without this workaround, the models get rendered without materials. I suspect that this is associated with [`get mesh()` ](https://github.com/BabylonJS/Babylon.js/blob/81be3c37822d53c9cd38d50c29a1243791eae607/gui/src/3D/controls/slider3D.ts#L70) because when that is disabled, the materials exist. However, this creates a new bug where the `SliderThumb ` doesn't have the associated behavior.

Do you have any idea about this? 

image

gui/src/3D/materials/mrdl/mrdlBackplateMaterial.ts
```diff
@@ -0,0 +1,447 @@
...
+     * Gets or sets the edge width of the backplate.
+     */
+    @serialize()
+    public edgePower = 1;
+
+    /**
+     * Gets or sets the edge width of the backplate.
+     */
+    @serialize()
+    public edgeLineGradientBlend =0.5;
```
sebavan, line 210:
```suggestion
    public edgeLineGradientBlend = 0.5;
```

gui/src/3D/materials/mrdl/mrdlSliderBarMaterial.ts
```diff
@@ -0,0 +1,802 @@
...
+     * Gets or sets the position of the hover glow effect.
+     */
+    @serialize()
+    public blobPosition = new Vector3(0, 0, 0.1);
+
+    /**
+     * Gets or sets the intensity of the hover glow effect.
+     */
+    @serialize()
+    public blobIntensity = 0.5;
```
sebavan, line 290:
```suggestion
    public blobIntensity = 0.5;
```

gui/src/3D/materials/mrdl/mrdlSliderBarMaterial.ts
```diff
@@ -0,0 +1,802 @@
...
+     */
+    @serialize()
+    public blobIntensity = 0.5; 
+
+    /**
+     * Gets or sets the near size of the hover glow effect.
+     */
+    @serialize()
+    public blobNearSize = 0.01;
+
```
sebavan, line 297:
```suggestion
```

gui/src/3D/materials/mrdl/mrdlSliderBarMaterial.ts
```diff
@@ -0,0 +1,802 @@
...
+     * Gets or sets the finger position for right index middle position.
+     */
+    @serialize()
+    public rightIndexMiddlePosition = new Vector3(0, 0, 0);
+
+    /**
+     * Gets or sets the Decal Scle for XY.
+     */
+    @serialize()
+    public decalScaleXY = new Vector2(1.5,1.5);
```
sebavan, line 392:
```suggestion
    public decalScaleXY = new Vector2(1.5, 1.5);
```

gui/src/3D/materials/mrdl/mrdlSliderBarMaterial.ts
```diff
@@ -0,0 +1,802 @@
...
+
+    /**
+     * Gets or sets decalFrontOnly
+     * Default is true
+     */
+    @serialize()
+    public decalFrontOnly = true;
+
```
sebavan, line 400:
```suggestion
```

gui/src/3D/materials/mrdl/mrdlSliderThumbMaterial.ts
```diff
@@ -0,0 +1,802 @@
...
+    public decalScaleXY = new Vector2(1.5, 1.5);
+
+    /**
+     * Gets or sets decalFrontOnly
+     * Default is true
+     */
+    @serialize()
+    public decalFrontOnly = true;
+
+
```
sebavan, line 401:
```suggestion

```

sebavan (commented):
Lint

gui/src/3D/controls/slider3D.ts
RaananW: Would be great if you could show me the issue, maybe we can resolve it together. It's hard for me to understand the lifecycle of those meshes.
If I understand correctly this mesh should only render when the material is ready? you can hide it until then (isVisible = true), or only set it after the material is ready (i.e. don't set the first this._sliderMesh at all)

tayomadein: I was looking at the Control3D class https://github.com/BabylonJS/Babylon.js/blob/a32c04e6a9c50e17ca9bc9a1834aa7a546c23bb8/gui/src/3D/controls/control3D.ts#L257-L271 
What I understand is that when Line 266 is executed, the meshes from the [ slider3D's `_createNode`](https://github.com/BabylonJS/Babylon.js/blob/2dcd4da1278799b0518110575f206703a0b94837/gui/src/3D/controls/slider3D.ts#L181) are not yet loaded, `affectMaterial` on line 270 is never reached and the material doesn't get applied when the meshes eventually load.

Another approach I tried was to use the [Control3D's `get mesh()`](https://github.com/BabylonJS/Babylon.js/blob/a32c04e6a9c50e17ca9bc9a1834aa7a546c23bb8/gui/src/3D/controls/control3D.ts#L236-L242), while this approach reaches `affectMaterial` on line 270, it returns a `TransformNode` which is pickable on line 268 but makes the sliderMesh unpickable.

I will be glad to walk you through this issues. We can sync on this at your convenience.

RaananW: I see the issue now.
`_createNode` is expected to be synchronous, but is async when loading external meshes, which is not supported by the current construction architecture.
But it seems to be solved in the holographic button almost the same way as you do. Why is it important to initialize `this._sliderThumb` during construction? isn't it enough if it is available after the model has finished loading?
`_affectMaterial` is creating the materials. You can check if they are available in this function and then apply it to them, and if not, apply them after the model has loaded. 
this way _createNode returns a mesh (the collider box), so affectMaterial is being executed correctly. then, after the models were loaded, you can set the materials. Holographic backplate, though simpler than this class, can show you the way to solve this.

RaananW:
Hi @tayomadein - are you working on the issue we discussed? Any need to jump in a call to discuss this?

tayomadein:
> Hi @tayomadein - are you working on the issue we discussed? Any need to jump in a call to discuss this?

Hello @RaananW yes I am, working on it, the recommended approach disables the slider behavior so far and I would love to get on a call to discuss this.

RaananW (approved):

PR #10955, opened 2021-08-30
DeviceInputSystem: Reset movementX/Y at the end of each frame

PolygonalSun:
Since movement during a pointerlock only uses the movement deltas, we need to reset these values to zero at the end of each frame since the do not automatically reset and are relative values.  Thanks to @syntheticmagus for finding this issue.


sebavan (approved):

sebavan:
Ohhh wait was it not creating an issue you fixed by the past ???

sebavan:
Not sur it was the same though but I prefer to ask just in case :-)

PolygonalSun:
> Ohhh wait was it not creating an issue you fixed by the past ???

This issue was similar to the mouse wheel issue that was fixed a while back.  Fix is pretty much the same.

syntheticmagus (approved):

Issue #10952, opened 2021-08-30
Display grid has rendering issues 


msDestiny14:
This is an issue to the GUI system not the editor. I was able to recreate this in a playground. It seems to be an issue in the render function for display grid. The top and left line are not being cleared. 

![image](https://user-images.githubusercontent.com/67929198/131377378-eaf7bcb2-48e1-4df8-bf5b-5f20e384b54a.png)

Related issue tracking for the GUI editor.
https://github.com/BabylonJS/ThePirateCove/issues/56

PR #10956, opened 2021-08-30
Fix for display grid rendering

msDestiny14:
I increased the starting index by 1 so it wouldn't draw the left and top boarders Now it is balanced with the bottom and right side.

image

https://github.com/BabylonJS/Babylon.js/issues/10952

### DIFF
--- a/gui/src/2D/controls/displayGrid.ts
+++ b/gui/src/2D/controls/displayGrid.ts
@@ -183,7 +183,7 @@ export class DisplayGrid extends Control {
                 context.strokeStyle = this._minorLineColor;
                 context.lineWidth = this._minorLineTickness;
 
-                for (var x = -cellCountX / 2; x < cellCountX / 2; x++) {
+                for (var x = -cellCountX / 2 + 1; x < cellCountX / 2; x++) {
                     const cellX = left + x * this.cellWidth;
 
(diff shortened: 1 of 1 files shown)


gui/src/2D/controls/displayGrid.ts
```diff
@@ -183,7 +183,7 @@ export class DisplayGrid extends Control {
                 context.strokeStyle = this._minorLineColor;
                 context.lineWidth = this._minorLineTickness;
 
-                for (var x = -cellCountX / 2; x < cellCountX / 2; x++) {
+                for (var x = -cellCountX / 2 + 1; x < cellCountX / 2; x++) {
                     const cellX = left + x * this.cellWidth;
```
RaananW, line 187: doesn't that depend on the number of cells (i.e. on cellCountX and cellCountY)? or better - whether or not these numbers are odd/even?

msDestiny14: I can test with even different numbers too. I thought I was just increasing starting the iterations 1 forward.

msDestiny14: image
 
Tested out some wacky cell w/h and grid w/h and looks good to me.

RaananW (approved):

PR #10957, opened 2021-08-31
Added sleepBody support for ammojs

CedricGuillemet:
Test PG : https://playground.babylonjs.com/#PSNB9V


Popov72 (approved):

RaananW:
Will the body wake up on collision? Will setting something more than 0  "wake" it up?

CedricGuillemet:
Yes, it wakes up with collision. Other values for `setActivationState` will have different effects, not just waking up for >0 values.

PR #10968, opened 2021-08-31
GUI: Fix highlights not displayed because of children clipping

Popov72:
`Rectangle._clipForChildren` is drawing the rounded rectangle and call `context.clip()` to clip against this rounded rectangle.

As the highlights is drawn after `_clipForChildren` is called, it is also clipped by the rounded rectangle which isn't what we want.

That's why I have added a `context.save` / `context.restore` inside/at the end of `Container._draw` (the clipping is done after the call to `context.save()` so `context.restore()` will restore a non-clipped area (or at least the clipped area of the parent control, which is what we want anyway)). However, I don't know if saving/restoring the context where I do it has any side-effects or not...

PG: https://playground.babylonjs.com/#XCPP9Y#8737

### DIFF
--- a/gui/src/2D/controls/container.ts
+++ b/gui/src/2D/controls/container.ts
@@ -450,6 +450,8 @@ export class Container extends Control {
 
         this._localDraw(contextToDrawTo);
 
+        context.save();
+
         if (this.clipChildren) {
             this._clipForChildren(contextToDrawTo);
         }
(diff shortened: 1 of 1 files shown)


deltakosh:
Main question is how slow it is (if it is)

Popov72:
Calling save/restore is slow? Even if it is, I'm not sure it's noticeable, as you normally don't have hundreds of containers inside containers in typical applications(?) But I'm not knowledgeable enough with canvas/contexts to know for sure.

deltakosh:
> Calling save/restore is slow? 
I don't know ;) was just asking

deltakosh (approved):

sebavan (approved):

PR #10973, opened 2021-09-01
Updated recast js

CedricGuillemet:
No new functionalities. Only recastjs built with latest emscripten and recastnavigation.
Tested with every PG in documentation with and without worker thread.

incoming PR on extensions.


src/Navigation/Plugins/recastJSPlugin.ts
```diff
@@ -44,7 +44,7 @@ export class RecastJSPlugin implements INavigationEnginePlugin {
      */
     public constructor(recastInjection: any = Recast) {
         if (typeof recastInjection === "function") {
-            recastInjection(this.bjsRECAST);
```
RaananW, line 47: is `await Recast()` generate anything in the global namespace? no way to inject an import anymore?

CedricGuillemet: I'm not the best to answer these questions :) I build recastjs the same way as ammojs and load it the same way as well.

RaananW: Oh, just saw the ammo init, which has the same log. I assume that you can await this function instead, but i'm not not entirely sure without checking.

RaananW (approved):

Popov72 (approved):

sebavan (commented):
Would be great to add a breaking change due to the new initialization way of recast ???

CedricGuillemet:
@sebavan yes, preparing a new PR for that.

sebavan (approved):

PR #10976, opened 2021-09-01
Fix issue with Gizmo+Pivot

CedricGuillemet:
Tested with all gizmos + inspector, with and without pivot.
Thank you for testing @PirateJC I'm sure I missed something :)

### DIFF
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -232,6 +232,21 @@ export class Gizmo implements IDisposable {
     }
 
     /**
+     * Handle position/translation when using an attached node using pivot
+     */
+    protected _handlePivot() {
+        const attachedNodeTransform = this._attachedNode as any;
+        // check there is an active pivot for the TransformNode attached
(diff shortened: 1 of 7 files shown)


src/Gizmos/gizmo.ts
```diff
@@ -231,6 +231,21 @@ export class Gizmo implements IDisposable {
         }
     }
 
+    /**
+     * Handle position/translation when using an attached node using pivot
+     */
+    protected _handlePivot() {
+        const attachedNodeTransform = this._attachedNode as any;
```
sebavan, line 238: Is there any better type we could cast into ?

sebavan (approved):
Only a tiny comment on my side :-)

rickfromwork:
This might be a different kind of pivot, but I know some behaviors, like the baseSixDofDragBehavior, reference the mesh pivots (calling getAbsolutePivotPoint, etc.). Would anything need to change on their end?

CedricGuillemet:
@rickfromwork Tested ok with SixDofDragBehavior, it's not the same code path.

Popov72 (approved):

PR #10985, opened 2021-09-02
Exposing the tether distance for the NearMenu control

rickfromwork:
(Quality of life)
Exposing the tether distance allows developers to use the menu in the desktop browser as desired.

Previously, the tether distance was tailored to VR/AR, which placed it far too close to the camera in a browser (past the near clipping plane, it seems). By providing customization, it can now be tested with and tweaked from a browser window, without needing to enter XR just to see it as intended.

The value set determines the default distance, and the min/max distance are calculated based on the original ratios.

### DIFF
--- a/gui/src/3D/controls/nearMenu.ts
+++ b/gui/src/3D/controls/nearMenu.ts
@@ -112,6 +112,12 @@ export class NearMenu extends TouchHolographicMenu {
         return this;
     }
 
+    public setFollowDistance(distance: number) {
+        this._defaultBehavior.followBehavior.minimumDistance = distance * 3/4;
+        this._defaultBehavior.followBehavior.defaultDistance = distance;
+        this._defaultBehavior.followBehavior.maximumDistance = distance * 3/2;
+    }
+
     protected _createNode(scene: Scene): Nullable<TransformNode> {
         const node = super._createNode(scene)! as Mesh;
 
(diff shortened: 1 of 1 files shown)


RaananW (approved):

RaananW:
There is a doc and linting issue.

BTW - i know those numbers were fixed before, but are they arbitrary, or is there some form of specification for that?

sebavan (approved):

rickfromwork:
> There is a doc and linting issue.
> 
> BTW - i know those numbers were fixed before, but are they arbitrary, or is there some form of specification for that?

The MRTK-Unity docs say the following, which is where the 0.3-0.6 range likely came from:
"The menu follows you and stays within 30-60cm range from the user for the near interactions."
That said, since everything appears much larger when viewed in the browser compared to XR, even 60 centimeters is too close to the camera, and gets blocked by the camera's near clipping plane.

rickfromwork:
Moving this to draft while I think about the implications of changing the distance, as it trades XR usability for desktop usability

sebavan:
@rickfromwork any update on this one ?

rickfromwork:
> @rickfromwork any update on this one ?

No, I've been busy with XR pointer work. It'll be a week or two before I'm set to go back into this.

sebavan:
@rickfromwork is this still work in progress ?

PR #10999, opened 2021-09-03
Fix issue introduced with #10976

CedricGuillemet:
Follow up https://forum.babylonjs.com/t/setting-getter-only-property-usepivotmatrix/23531
Trying to clone a getter only produces an error.
I replaced the getter with a function. I changed the only place where it's used. I have to change the cast from TransformNode to any because some objects (camera/light) do not inherit from TransformNode.

### DIFF
--- a/src/Meshes/transformNode.ts
+++ b/src/Meshes/transformNode.ts
@@ -203,7 +203,7 @@ export class TransformNode extends Node {
     /**
      * return true if a pivot has been set
      */
-     public usePivotMatrix(): boolean {
+     public isUsingPivotMatrix(): boolean {
          return this._usePivotMatrix;
      }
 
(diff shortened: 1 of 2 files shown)


Popov72 (approved):

RaananW:
That's a breaking change though

RaananW:
Sorry, correction, it was just added yesterday, so it's all good :-) Just need to fix the documentation.

src/Meshes/transformNode.ts
```diff
@@ -203,7 +203,7 @@ export class TransformNode extends Node {
     /**
      * return true if a pivot has been set
      */
-     public get usePivotMatrix(): boolean {
+     public usePivotMatrix(): boolean {
```
sebavan, line 206: Could we change the name to isUsingPivotMatrix or simply add usePivotMatrix to the list of exception in the clone function instead ? :-)

Issue #10856, opened 2021-08-12
Rotation on ArcRotateCamera continues when button is released if another button is still active


PolygonalSun:
Forum post discussing this: https://forum.babylonjs.com/t/camera-rotation-is-stuck-because-of-pointerinput-fakemove/22171/5

We have a really active forum to help answering questions (https://forum.babylonjs.com/)

With cameras like the ArcRotateCamera, when a mouse button is pressed during a rotate-drag event, if the original button is release with this new button pressed, the rotation will still occur.

Fix for this will require implementing the following behavior:
The initial button press will initiate the behavior (rotate, pan, etc.) and any other buttons will be ignored.

sebavan:
@PolygonalSun, Any ETA on this one ?

PolygonalSun:
> @PolygonalSun, Any ETA on this one ?

Not yet.  I figured out that the reason it isn't working was because Safari is using different values for the PointerEvent's button (eg. 0 for both left click and no button clicked).  I'm working on a different approach that will work for Safari and Every other browser.

PR #11005, opened 2021-09-03
Fixed Context code to omit Safari

PolygonalSun:
This PR takes the original code from reverted PR https://github.com/BabylonJS/Babylon.js/pull/10942 and adds a check to only apply these checks to non-safari instances.


sebavan:
Should Fix #10856

sebavan (commented):
Only one thing would be to factorize the check code in base camera as it sounds highly similar and error prone for the next update if we forget one ?

sebavan:
Waiting on @RaananW feedback as well

PolygonalSun:
> Was it tested on all device types? mobile (android/ios) and desktop?

Hey @RaananW, I tested this on an android phone, ipad, and both mac and pc.

RaananW (approved):

Issue #11014, opened 2021-09-07
LinkWithMesh causes some flickering on button creation


msDestiny14:
https://forum.babylonjs.com/t/linkwithmesh-causes-some-flickering-on-button-creation/23614/2

PR #11015, opened 2021-09-07
Fix for LinkWithMesh causes some flickering on button creation

msDestiny14:
Since currentMeasure isn't calculated until the render and the update is happening before the render I call the computeLayout which will calculate the currentMeasures.

https://github.com/BabylonJS/Babylon.js/issues/11014


sebavan (commented):
Is this not really expensive on update ??? why was it working on the second frame ? we might be able to warm up the missing data earlier only when needed ?

msDestiny14:
The problem is these two lines: 

        var newLeft = ((projectedPosition.x + this._linkOffsetX.getValue(this._host)) - this._currentMeasure.width / 2);
        var newTop = ((projectedPosition.y + this._linkOffsetY.getValue(this._host)) - this._currentMeasure.height / 2);

_currentMeasure.width/height are not calculated until the first render frame. But the function that calls this is called before this. So the values are 0;

msDestiny14:
Another solution I thought about was just using w/h flat out without the _currentMeasure. But that might have other consequences.

deltakosh:
Yeah this is a very slow solution. We cannot render a text per update per mesh to get a measure

msDestiny14:
> Yeah this is a very slow solution. We cannot render a text per update per mesh to get a measure

What do you think of the second solution I wrote above?

sebavan (commented):
I do not think this will work :-( I guess we might have less precision than _currentMeasure.

This probably needs to be dealt with in another way

msDestiny14:
I'm gonna think more about it for now then.

PR #11025, opened 2021-09-08
[GE] Change webpage URL once saved to snippet server

msDestiny14:


RaananW (changes requested):
This will not allow to overwrite the entire save-to prodecude, but will use the existing one. 
NME duplicates the behavior allows you to change the entire procedure (or use the default one if that's the behavior you need). This will help the save process to be much more dynamic.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -156,66 +156,73 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
-                    this.props.globalState.customSave!.action(adt.snippetId).then(() => {
-                    }).catch((err) => {
+        if (this.props.globalState.customSave) {
+            this.props.globalState.customSave!.action(content).then((resolve) => {
+                adt.snippetId = (resolve as unknown as string);
+                this.forceUpdate();
+                if (navigator.clipboard) {
+                    navigator.clipboard.writeText(adt.snippetId);
+                }
+                alert("GUI saved with ID: " + adt.snippetId + " (please note that the id was also saved to your clipboard)");
```
RaananW, line 166: Do we want to use alert?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -156,66 +156,73 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+            const dataToSend = {
+                payload: JSON.stringify({
+                    gui: content,
+                }),
+                name: "",
+                description: "",
+                tags: "",
+            };
+    
+            xmlHttp.send(JSON.stringify(dataToSend));
```
RaananW, line 214: would be great to have a "loading" state, the disables the save button until the save is done (or fails). Otherwise it is possible to press the button many times

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -156,66 +156,73 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+    
+                        const windowAsAny = window as any;
+    
+                        if (windowAsAny.Playground && oldId) {
+                            windowAsAny.Playground.onRequestCodeChangeObservable.notifyObservers({
+                                regex: new RegExp(oldId, "g"),
+                                replace: `parseFromSnippetAsync("${adt.snippetId}`,
+                            });
+                        }
+                        alert("GUI saved with ID: " + adt.snippetId + " (please note that the id was also saved to your clipboard)");
```
RaananW, line 195: is it possible to unify both calls? either customSave, or the local save, and deal with error/success the same way (with the same function)?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -281,10 +281,12 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         this.globalState.onSelectionChangedObservable.notifyObservers(null);
         await this.globalState.guiTexture.parseFromSnippetAsync(snippedId);
         this.loadToEditor();
-        this.props.globalState.customSave!.action(this.globalState.guiTexture.snippetId).then(() => {
-        }).catch((err) => {
-            alert("Unable to load your GUI");
-        });
+        if (this.props.globalState.customLoad) {
+            this.props.globalState.customLoad!.action(this.globalState.guiTexture.snippetId).then(() => {
```
RaananW, line 285: the `then` function is empty. want to remove it?

RaananW (commented):
Small nits, and a general question - do we want to use alerts for the communication with the user?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Any reason why not?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Let me see what I can do. :)

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
RaananW: alert is blocking. But i guess that without some dialog mechanism there is no easy way of solving this. you can resolve this if there is no other way.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -198,13 +188,36 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+        if (this.props.globalState.customSave) {
+            await this.props.globalState.customSave!.action(content).then((resolve) => {
+                adt.snippetId = (resolve as unknown as string);
+                success = true;
+            }).catch((err) => {
+                alert(err);
+            });
+        }
+        else {
+            while (success === null) {
```
RaananW, line 207: This will block execution. why isn't `saveToSnippetServerHelper` return promise as well? 
then you do - 

```
const savePromise = this.props.globalState.customSave?.action || this.saveToSnippetServerHelper;
savePromise(content).then((resolve) => {
                adt.snippetId = (resolve as unknown as string);
                success = true;
            }).catch((err) => {
                alert(err);
            });
```
let's not use a while loop to control async events :-)

RaananW (changes requested):
don't control the save function with a while loop, this will block ui rendering.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -152,59 +152,66 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
-        const dataToSend = {
-            payload: JSON.stringify({
-                gui: content,
-            }),
-            name: "",
-            description: "",
-            tags: "",
-        };
+        const savePromise = this.props.globalState.customSave?.action || this.saveToSnippetServerHelper;
+        savePromise(content, adt).then((resolve: unknown) => {
```
RaananW, line 200: I am being nit-picky and I apologize for that :-)

resolve should always be a string, as both promises resolve with a string. If you define the return type of both customSave.action and saveToSnipperServerHelper to `Promise<string>` you will not need this casting in line 201 (as unknown as string)

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -173,7 +173,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
                         }
                         resolve(adt.snippetId);
                     } else {
-                        reject();
+                        reject("string");
```
sebavan, line 176: A real error message even generic might be better here ?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -191,14 +191,15 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+    
 
     saveToSnippetServer = async () => {
         const adt = this.props.globalState.guiTexture;
         const content = JSON.stringify(adt.serializeContent());
 
         const savePromise = this.props.globalState.customSave?.action || this.saveToSnippetServerHelper;
-        savePromise(content, adt).then((resolve: unknown) => {
-            adt.snippetId = (resolve as unknown as string);
+        savePromise(content, adt).then((resolve: string) => {
```
sebavan, line 201: could you rename resolve in result or snippetId ? resolve is confusing here

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -152,59 +152,67 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
-            description: "",
-            tags: "",
-        };
+        const savePromise = this.props.globalState.customSave?.action || this.saveToSnippetServerHelper;
+        savePromise(content, adt).then((resolve: string) => {
+            adt.snippetId = resolve;
+            if (navigator.clipboard) {
+                navigator.clipboard.writeText(adt.snippetId);
+            }
+            alert("GUI saved with ID: " + adt.snippetId + " (please note that the id was also saved to your clipboard)");
```
sebavan, line 206: Are we using alert in GE to communicate with the user ???

I guess a message panel like sandbox or playground could be better

sebavan (commented):
Some nits mainly but did you plan to move away from alert() later ?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
RaananW: what @msDestiny14 said was that NME uses the same alert. We had this conversation before, and i believe UX-wise alerts are bad, but we will need a form of dialog or local alert/toast system to avoid using alerts.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Whoops. Good find. ;)

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -152,59 +152,67 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
-                    const oldId = adt.snippetId;
-                    adt.snippetId = snippet.id;
-                    if (snippet.version && snippet.version != "0") {
-                        adt.snippetId += "#" + snippet.version;
-                    }
-                    this.forceUpdate();
-                    if (navigator.clipboard) {
-                        navigator.clipboard.writeText(adt.snippetId);
+    saveToSnippetServerHelper = (content: string, adt: AdvancedDynamicTexture) : Promise<string> => {
+        return new Promise((snippetId, reject) => {
```
sebavan, line 156: snippetId here should be named resolve :-)

PR #11028, opened 2021-09-08
Copy buffer if invalid typed array alignment for glTF loader

bghgary:
See https://forum.babylonjs.com/t/error-on-importing-a-simple-glb-exported-from-blender/23314.

This change loosens the requirement for typed array alignment such that it will copy the buffer when the typed array byte alignment requirement is not met. Note that glTF load performance will suffer a bit if this happens and the glTF validation will report an error.


sebavan (approved):

darraghjburke (approved):
lgtm! don't forget what's to add to what's new :)

Popov72 (approved):

PR #11034, opened 2021-09-08
Hand menu upgrades

rickfromwork:
Notable changes:
- Hand menu follows the left hand now by default, as most users are right-handed and this allows them to press buttons using their right hand
- Introduced visibility options, similar to MRTK's settings - The user should be looking at their hand, and the palm should be facing the user
- Moved the default position to the outer hand to make it easier to use when the palm is up (no need to overlap hands to interact with it)
- Near menu pin has been re-aligned after previous alignment changed to the TouchHolographicButton
- Padding on menus has been reduced
- Menu button scaling has been generalized to base TouchHolographicMenu class
- Fixed a failure that occurred when a menu tried updating its backplate but had no children

Done as part of #10846


gui/src/3D/controls/touchHolographicMenu.ts
```diff
@@ -38,11 +43,13 @@ export class TouchHolographicMenu extends VolumeBasedPanel {
     public set backPlateMargin(value: number) {
         this._backPlateMargin = value;
 
-        this.children.forEach((control) => {
-            this._updateCurrentMinMax(control.position);
-        });
+        if (this._children.length >= 1) {
```
CraigFeldspar, line 46: nit: `if (this._children.length)`

CraigFeldspar (approved):
LGTM !

RaananW (approved):

gui/src/3D/controls/touchHolographicMenu.ts
rickfromwork: Sorry, saw this too late!

sebavan: np I ll add it in the nightly

PR #11036, opened 2021-09-09
Geodesic and Goldberg Polyhedron

BabylonJSGuide:
Ref https://forum.babylonjs.com/t/pr-of-geodesic-and-goldberg-polyhedron-or-not/20707/4?u=johnk

For Goldberg poly mesh extended mesh to add properties so that individual faces can be colored, textured and built on.

Docs and examples to follow.


src/Maths/math.constants.ts
```diff
@@ -3,11 +3,32 @@
...
+ * Constant Golden Ratio value in Babylon.js
+ * @ignorenaming
+ */
+ export const PHI = (1 + Math.sqrt(5)) / 2;
+
+ /**
+ * Constant one third of the square root of 3
+ * @ignorenaming
+ */
+  export const THIRDR3 = Math.sqrt(3) / 3;
```
sebavan, line 23: I wonder if this should be in math.isovector only ? not sure this is reusable enough ?

src/Maths/math.constants.ts
```diff
@@ -3,11 +3,32 @@
...
+
+/**
+ * Constant used to define the isometric grid size in Babylon.js
+ * @ignorenaming
+ */
+ let IsoGridSize = 0.5;
+ export { IsoGridSize };
```
sebavan, line 30: Same comment here even more cause this file should only contain true constants. Also isoGridSize might be better off as an option with a default value in isoVector class ?

src/Maths/math.isovector.ts
```diff
@@ -0,0 +1,289 @@
...
+import { Vector3} from "./math.vector";
+import { THIRDR3, IsoGridSize} from './math.constants';
+import { DeepImmutable} from "../types";
+import { ArrayTools } from '../Misc/arrayTools';
+import { _TypeStore } from '../Misc/typeStore';
+
+/**
+ * Class representing an isovector containing 2 INTEGER coordinates
+ * x axis is horizontal
+ * y axis is 60 deg counter clockwise from positive y axis
```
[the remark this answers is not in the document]
sebavan: I am also wondering if it should only extend Vector2 as a lot of code is dupplicated and only a couple specific rotation and such have been added ?

sebavan (changes requested):
Looks pretty solid to me !!! thanks a lot @BabylonJSGuide !!!

BabylonJSGuide:
@sebavan  will work on your suggestions over next few days

src/Maths/math.constants.ts
BabylonJSGuide: Done

src/Maths/math.constants.ts
BabylonJSGuide: Done

src/Maths/math.isovector.ts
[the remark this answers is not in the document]
BabylonJSGuide: _Would be could if you have a reference like wiki_ As this is something I made up to expedite the project I doubt if you will find a corresponding definition anywhere.

 _if it should only extend Vector2_ Problem here is that other Vector2 properties do not produce anything sensible. 

Since the class is so specialized and only used for the purpose of creating the Geodesic and Goldberg polyhedra it probably makes more sense to remove unused properties and hide the class from the API doc?

src/Meshes/mesh.vertexData.ts
```diff
@@ -10,6 +10,7 @@ declare type Geometry = import("../Meshes/geometry").Geometry;
 declare type Mesh = import("../Meshes/mesh").Mesh;
 
 import { ICreateCapsuleOptions } from "./Builders/capsuleBuilder";
+import { PolyhedronData } from "./geoMesh";
```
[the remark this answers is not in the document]
BabylonJSGuide: Done

src/Maths/math.isovector.ts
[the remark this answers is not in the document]
sebavan: Sounds perfect to me if you hide it !!!

src/Meshes/geoMesh.ts
```diff
@@ -0,0 +1,790 @@
...
+import { IsoVector } from '../Maths/math.isovector';
+
+
+
+/**
+ * Class representing data for one face OAB of an equilateral icosahedron
+ * When O is the isovector (0, 0), A is isovector (m, n)
+ * @hidden
+ */
+export class Primary {
```
RaananW, line 13: This will be exported as BABYLON.Primary on our UMD package. it feels a bit too abstract for an exported class. If it is an internal class (that is not expected to be used by users) add an underscore and name it a bit differently so that we will know what Primary means.

src/Meshes/geoMesh.ts
```diff
@@ -0,0 +1,790 @@
...
+        public face: number[][]
+    ) {
+
+    }
+}
+
+/**
+ * This class Extends the PolyhedronData Class to provide measures for a Geodesic Polyhedron
+ */
+export class GeoData extends PolyhedronData{
```
RaananW, line 525: GeoData can be understood as geographic instead of Geodesic. would renaming this make sense?

RaananW (commented):
Thank you so much for that!!!

Just a few comments without diving into the code too much at the moment.
It's mainly about naming and whether or not the types should be exported.

src/Maths/math.isovector.ts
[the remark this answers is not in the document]
BabylonJSGuide: Unused properties removed, class hidden to API docs - done

src/Maths/math.isovector.ts
```diff
@@ -0,0 +1,289 @@
...
+}
+
+/**
+ * @hidden
+ */
+ export class TmpIsoVectors {
+    public static IsoVector: IsoVector[] = ArrayTools.BuildArray(6, IsoVector.Zero); // 6 temp IsoVectors at once should be enough
+}
+
+_TypeStore.RegisteredTypes["BABYLON.IsoVector"] = IsoVector;
```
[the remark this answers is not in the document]
BabylonJSGuide: Removed

src/Maths/math.isovector.ts
```diff
@@ -0,0 +1,289 @@
...
+    public static One(): IsoVector {
+        return new IsoVector(1, 1);
+    }
+}
+
+/**
+ * @hidden
+ */
+ export class TmpIsoVectors {
+    public static IsoVector: IsoVector[] = ArrayTools.BuildArray(6, IsoVector.Zero); // 6 temp IsoVectors at once should be enough
```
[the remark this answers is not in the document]
BabylonJSGuide: Removed as was never used

src/Meshes/geoMesh.ts
BabylonJSGuide: Done

BabylonJSGuide: Done

src/Meshes/geoMesh.ts
BabylonJSGuide: Done

BabylonJSGuide:
Any thoughts on the following would be helpful.

To be able to access the pentagon and hexagon faces of a Goldberg mesh (not the triangular facets creating the mesh) and for example alter their color or texture I wanted a user to have access to a range of data not normally available for a mesh. I set up a GDMesh to extend a Mesh and added the properties and methods I needed. 

To serialize this data I store it in the mesh metadata property. Having created a Goldberg mesh, exported it to a .babylon file all works well.  Having just tested it I found an issue (obvious in hindsight). On using this file to import the mesh it renders correctly and the data I want users to access appears in the metadata.The issue is that all the import methods assume a mesh is imported not a GDMesh and of course the extra properties and methods do not exist on the imported mesh. Changing all imports to import a GDMesh is unwarranted and so I see two solutions.

1. Leave things as they are and allow users creating a mesh to have access to the new properties and methods. Advise users importing such a mesh that they can access the additional data via the mesh metadata and provide the methods such as setFaceColors via an extension JavaScript file.
2. Remove the GDMesh class and make a Goldberg mesh just a mesh with the data in the mesh metadata property and provide a JavaScript extension to add in direct access to properties and methods.

Any help/ suggestions gratefully received.

src/Meshes/Builders/goldbergBuilder.ts
```diff
@@ -0,0 +1,369 @@
...
+            n === Math.floor(n);
+            Logger.Warn("n not an integer only floor(n) used");
+        }
+        if (n > m) {
+            const temp = n;
+            n = m;
+            m = temp;
+            Logger.Warn("n > m therefore m and n swapped");
+        }
+        const primTri: Primary = new Primary;
```
Popov72, line 351: Missing the parenthesis?
```javascript
const primTri: Primary = new Primary();
```

src/Meshes/geoMesh.ts
```diff
@@ -0,0 +1,790 @@
...
+
+
+/**
+ * Class representing data for one face OAB of an equilateral icosahedron
+ * When O is the isovector (0, 0), A is isovector (m, n)
+ * @hidden
+ */
+export class Primary {
+    //properties
+    public _m: number;
```
Popov72, line 15: I think you can remove all the leading underscores in property and function names, as the class is hidden anyway.

src/Meshes/Builders/goldbergBuilder.ts
```diff
@@ -0,0 +1,369 @@
...
+    public _faceZaxis: Vector3[] = [];
+    public _faceXaxis: Vector3[] = [];
+    public _faceYaxis: Vector3[] = [];
+    private _nbSharedFaces: number;
+    private _nbUnsharedFaces: number;
+    private _nbFaces: number;
+    private _nbFacesAtPole: number;
+    private _adjacentFaces: number[][];
+
+    private _setMetadata = () => {
```
Popov72, line 119: Why not declaring this function as usual?
```typescript
private _setMetadata() {
...
}
```

Same comment for all functions starting at `refreshFaceData` below.

The user can't overload those properties as the class is hidden, so I would advise to declare the functions as usual.

src/Meshes/geoMesh.ts
```diff
@@ -0,0 +1,790 @@
...
+        }
+        //order vertices by x and then y
+        vertices.sort((a, b) => {
+            return a.x - b.x
+        });
+    
+        vertices.sort((a, b) => {
+            return a.y - b.y
+        });
+
```
Popov72, line 396: Just to be sure, you didn't want to do this instead of the two sorts:
```typescript
        vertices.sort((a, b) => {
            return a.x === b.x ? a.y - b.y : a.x - b.x;
        });
```
?

src/Meshes/Builders/goldbergBuilder.ts
BabylonJSGuide: There is a bigger problem with refreshFaceData since it was intended to reset properties from the metadata after a Goldberg mesh was imported but ... see 
https://github.com/BabylonJS/Babylon.js/pull/11036#issuecomment-916865324

src/Meshes/geoMesh.ts
Popov72: The two sorts don't do the same thing than the single sort.

For eg, if you have (1,2), (0,3), (2,1):
* first method, sort the set by X then sort the set by Y:
  * sort by X: (0,3), (1,2), (2,1)
  * sort by Y: (2,1), (1,2), (0,3)
* second method, sort by X and for same X sort by Y:
  * (0,3), (1,2), (2,1)

It's not the same result.

Sorting the whole set by X then doing a second pass to sort by Y looked strange to me because I have never really saw this before and I don't understand what's the expected output (the X sort you do first will be in part wrecked by the Y sort depending on how the Y values are distributed - in my example above, the two methods lead to inverted sorting!), that's why I asked if what you wanted to do wasn't what I presented.

src/Meshes/geoMesh.ts
Popov72: I'm ok with that, but maybe to be compliant with Babylon code writing rules you should add a /** @hidden */ comment in front of each property / function that has a leading underscore and which is public, as it is not meant to be documented/seen by the user.

However, will let @sebavan / @deltakosh decide if that should be enforced or not.

src/Maths/math.isovector.ts
```diff
@@ -0,0 +1,289 @@
...
+    /**
+     * For an equilateral triangle OAB with O at isovector (0, 0) and A at isovector (m, n)
+     * Rotates one IsoVector 120 degrees counter clockwise about the center of the triangle
+     * Please note that this is an in place operation
+     * @param m integer a measure a Primary triangle of order (m, n) m > n
+     * @param n integer a measure for a Primary triangle of order (m, n)
+     * @returns the rotated IsoVector
+     */
+    public rotate120(m: number, n: number) { //m, n integers
+        if (m !== Math.floor(m)) {
```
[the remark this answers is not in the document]
Popov72: That's ok to me.

src/Meshes/geoMesh.ts
BabylonJSGuide: Asked about this earlier when I started https://forum.babylonjs.com/t/pr-of-geodesic-and-goldberg-polyhedron-or-not/20707/3?u=johnk

src/Meshes/geoMesh.ts
BabylonJSGuide: OK understand. Think rows and columns. The two sorts sort the columns in order then the rows. Then for each row you read in order you will read the columns in order.

src/Meshes/geoMesh.ts
BabylonJSGuide: I am of course happy to ensure consistency within Babylon.js and will amend anything as decided by the reviewers.

RaananW: Following the convention of `_Exporter` from GLTFExplorer and other classes (`_TimeToken`, `_InternalAbstractMeshDataInfo`) this is ok.
I agree that it will be fine (and probably cleaner) if there will be no `_` before all public elements in this hidden class, but it's fine to leave it this way.

RaananW (commented):
There are many lint issues (you can see them here - https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=15250&view=logs&j=32e2e1bb-a28f-5b18-6cfc-3f01273f5609&t=2bf39655-ae43-5ae7-fe00-6e76c9bf09b6).
Also - please don't commit package-lock.json, we update it during our nightlies / preview updates.

src/Meshes/geoMesh.ts
BabylonJSGuide: 2 Babylon.js Typescript experts to 1 student - will delete the _  in hidden classes. :)

BabylonJSGuide:
> 
> 
> There are many lint issues (you can see them here - https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=15250&view=logs&j=32e2e1bb-a28f-5b18-6cfc-3f01273f5609&t=2bf39655-ae43-5ae7-fe00-6e76c9bf09b6).
> Also - please don't commit package-lock.json, we update it during our nightlies / preview updates.

Forgot that pushing to my github updated the Babylon.js github. 

Am going to close this PR and when finished making suggested changes and updating comments will create a new PR

PR #11039, opened 2021-09-09
SahderMaterial - add #define if it's not there.

RaananW:
As per our documentation, it is possible to add defines in a shader material. The array is then being added as is to the defines array when compiling, but #define is not added if not set. This contradicts out documentation. Setting `defines: ["MyDefine"]`) will fail, as it wil add the line `MyDefine` to the shader.

### DIFF
--- a/src/Materials/shaderMaterial.ts
+++ b/src/Materials/shaderMaterial.ts
@@ -534,7 +534,7 @@ export class ShaderMaterial extends Material {
         for (var index = 0; index < this._options.defines.length; index++) {
             const defineToAdd = this._options.defines[index].indexOf("#define") === 0 ?
                 this._options.defines[index] :
-                `#define  + ${this._options.defines[index]}`;
+                `#define ${this._options.defines[index]}`;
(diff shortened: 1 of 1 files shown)


src/Materials/shaderMaterial.ts
```diff
@@ -532,7 +532,10 @@ export class ShaderMaterial extends Material {
         }
 
         for (var index = 0; index < this._options.defines.length; index++) {
-            defines.push(this._options.defines[index]);
+            const defineToAdd = this._options.defines[index].indexOf("#define") === 0 ?
+                this._options.defines[index] :
+                `#define  + ${this._options.defines[index]}`;
```
Popov72, line 537: You need to remove the " + " characters:
```javascript
   `#define ${this._options.defines[index]}`;
```

Popov72 (approved):

src/Materials/shaderMaterial.ts
RaananW: the downsides of converting to string literals... thanks!!!

PR #11045, opened 2021-09-09
Allow combineAction child condition check

sebavan:
https://forum.babylonjs.com/t/execute-code-action/23656

### DIFF
--- a/src/Actions/directActions.ts
+++ b/src/Actions/directActions.ts
@@ -410,14 +410,21 @@ export class CombineAction extends Action {
     public children: Action[];
 
     /**
+     * defines if the children actions conditions should be check before execution
+     */
+    public enableChildrenConditions: boolean;
+
+    /**
      * Instantiate the action
      * @param triggerOptions defines the trigger options
      * @param children defines the list of aggregated animations to run
      * @param condition defines the trigger related conditions
+     * @param enableChildrenConditions defines if the children actions conditions should be check before execution
      */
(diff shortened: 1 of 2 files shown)


deltakosh (approved):

src/Actions/directActions.ts
```diff
@@ -409,15 +409,22 @@ export class CombineAction extends Action {
...
+
     /**
      * Instantiate the action
      * @param triggerOptions defines the trigger options
      * @param children defines the list of aggregated animations to run
      * @param condition defines the trigger related conditions
+     * @param enableChildrenConditions defines if the children actions conditions should be check before execution
      */
-    constructor(triggerOptions: any, children: Action[], condition?: Condition) {
+    constructor(triggerOptions: any, children: Action[], condition?: Condition, enableChildrenConditions = false) {
```
RaananW, line 424: feels like this should be on per default?

RaananW (approved):

src/Actions/directActions.ts
sebavan: so basically back compat :-) but also there are 2 cases:
1. we consider that combined action is the one having the condition so we disregard children ones in this case which is the way we were seeing it before the forum post. Basically, the children should launch as long as the parent is ok. (handy to combine existing actions)
2. We want to also honor the children conditions as stated in the forum post

I am not sure of the ratio of both intents so in doubt I chose back compat lol

PR #11046, opened 2021-09-09
Fix for LinkWithMesh causes some flickering on button creation

msDestiny14:
https://github.com/BabylonJS/Babylon.js/issues/11014


sebavan (commented):
Looks way better, just need to do it for the line component as well :-)

Only question is when could parent._currentMeasure be unavailable ? to be sure it does not flicker in most cases ?

msDestiny14:
I'm trying to think of a case. Parent should always be available when that function is called because it will be inside the adt so _root will always be the parent worst case.

sebavan:
Yup, I was more focused on the _currentMeasure  of the parent :-)

gui/src/2D/controls/control.ts
```diff
@@ -1318,6 +1318,13 @@ export class Control {
         let oldLeft = this._left.getValue(this._host);
         let oldTop = this._top.getValue(this._host);
 
+        if (this._currentMeasure.width == 0 && this._currentMeasure.height == 0) {
```
RaananW, line 1321:
```suggestion
        if (this._currentMeasure.width === 0 && this._currentMeasure.height === 0) {
```

RaananW (approved):

PR #11056, opened 2021-09-13
Fix goToFrame when multiple calls are made

stevendelapena:
When in a paused state, multiple calls to "gotoFrame" create an incorrect _manualJumpDelay. This occurs because the currentFrame of the first RuntimeAnimation is used to determine the that offset. But the currentframe is set on each call regardless of whether the animatable is paused or not. So it uses the offset of the last call to gotoFrame rather than the currentFrame it is actually paused on. I added a local field to keep track of the true currentframe.

### DIFF
--- a/src/Animations/animatable.ts
+++ b/src/Animations/animatable.ts
@@ -22,6 +22,7 @@ export class Animatable {
     private _speedRatio = 1;
     private _weight = -1.0;
     private _syncRoot: Nullable<Animatable> = null;
+    private _frameToSyncFromJump: Nullable<number> = 0;
 
     /**
      * Gets or sets a boolean indicating if the animatable must be disposed and removed at the end of the animation.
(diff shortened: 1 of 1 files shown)


sebavan:
Adding @deltakosh to review :-)

sebavan (approved):

sebavan:
@stevendelapena Could you please provide a repro in the playground ? I merged but I am having doubts that it would actually fully works in all cases ?

stevendelapena:
Sure. See: https://playground.babylonjs.com/#7V0Y1I#960 . Note the long delay before the box comes into view. There shouldn't be any delay because the animation should be at frame 0 due to the nature of it being initially paused. But instead the manualJumpDelay is calculated from frame 20 during the last goToFrame.

The fix I added is to store the current frame (to frameToSyncFromJump) the first time manualJumpDelay is assigned. When manualJumpDelay is set back to null, which occurs in the first pass the animatable is resumed, frameToSyncFromJump is set to null as well.

PR #11057, opened 2021-09-13
Fix wrong name for a cloned camera

Popov72:
See https://forum.babylonjs.com/t/wrong-name-assigned-to-a-cloned-camera/23763


sebavan (commented):
Do you know why name is not assigned by the serialization helper ??? I am afraid of a regression there.

Popov72:
The new object is first created with the right name but the serialization helper then copies all properties from source to target and the name will be overwritten because of that. Maybe we can instruct the serialization helper to not take into account some properties while copying, but I found it easier to simply set the new name before returning.

PR #11063, opened 2021-09-13
NME Fixes

msDestiny14:
-Fix for inputLineComponents
-Fix for popup not appearing


nodeEditor/src/blockTools.ts
```diff
@@ -124,7 +124,7 @@ export class BlockTools {
                 return new TextureBlock("Texture", nodeMaterial.mode === NodeMaterialModes.Particle);
             case "ReflectionTextureBlock":
                 return new ReflectionTextureBlock("Reflection texture");
-            case "LightBlock":
+            case "LightsBlock":
```
deltakosh, line 127: why this change?

nodeEditor/src/sharedComponents/floatLineComponent.tsx
```diff
@@ -58,6 +58,15 @@ export class FloatLineComponent extends React.Component<IFloatLineComponentProps
         return false;
     }
 
+    componentWillUnmount() {
+        if (this.props.globalState.blockKeyboardEvents = true) {
```
deltakosh, line 62: ===

nodeEditor/src/sharedComponents/numericInputComponent.tsx
```diff
@@ -35,6 +35,12 @@ export class NumericInputComponent extends React.Component<INumericInputComponen
         return false;
     }
 
+    componentWillUnmount() {
+        if (this.props.globalState.blockKeyboardEvents = true) {
```
deltakosh, line 39: ===

nodeEditor/src/blockTools.ts
msDestiny14: @PirateJC asked me to change it.

nodeEditor/src/sharedComponents/textInputLineComponent.tsx
```diff
@@ -38,6 +38,13 @@ export class TextInputLineComponent extends React.Component<ITextInputLineCompon
         return false;
     }
 
+    componentWillUnmount() {
+        if (this.props.globalState.blockKeyboardEvents = true) {
```
deltakosh, line 42: ===

nodeEditor/src/sharedComponents/textInputLineComponent.tsx
```diff
@@ -61,7 +68,7 @@ export class TextInputLineComponent extends React.Component<ITextInputLineCompon
         this._localChange = true;
         const store = this.props.value !== undefined ? this.props.value : this.props.target[this.props.propertyName!];
 
-        if (this.props.validator && raisePropertyChanged) {
+        if (this.props.validator) {
```
deltakosh, line 71: why removing raisePropertyChanged?

deltakosh (changes requested):
.

nodeEditor/src/sharedComponents/textInputLineComponent.tsx
msDestiny14: to get the pop up showing again. I was actually wondering if I should remove the lines a bit lower that you added back in july since it will now do the same thing.

msDestiny14:
Updated comment to better overview the changes.

nodeEditor/src/blockTools.ts
deltakosh: This could be misleading when people will found out that the block name is not the same

nodeEditor/src/sharedComponents/textInputLineComponent.tsx
deltakosh: actually this should work

nodeEditor/src/blockTools.ts
Popov72: More importantly it's a breaking change, all node materials that are currently using `LightBlock` will fail loading.

nodeEditor/src/sharedComponents/textInputLineComponent.tsx
deltakosh: actually no: we do not want to validate on every change (only on blur) else we will prevent users to enter values (like the validator is that you cannot enter "var" but you want to type "variant". If you validate per stroke you will never be able to enter variant

nodeEditor/src/blockTools.ts
deltakosh: Totally. I should have mentioned that also. I'm getting old

msDestiny14: OH. That stinks I will remove it then.

nodeEditor/src/sharedComponents/textInputLineComponent.tsx
msDestiny14: the code you have below does similar should i remove that segment as well then?

deltakosh: I did not check but my ubber point is to make sure that your fix works with a variable that you want to name uv1 for instance. The system should block uv but not uv1

msDestiny14: Changed :)

deltakosh (approved):

PR #11065, opened 2021-09-13
Add option to specify MultiRenderTarget's depth texture format.

kircher1:
For WebGL 2, this change allows for choosing the depth texture format (16, 24, 24_S8, or 32F). Default is the same as before: 16 bits.

As an optimization, glRenderBuffers do not need to be created if a depth(/stencil) texture is being generated since it will be put on the framebuffer's depth/stencil attachment instead.

### DIFF
--- a/src/Engines/Extensions/engine.multiRender.ts
+++ b/src/Engines/Extensions/engine.multiRender.ts
@@ -267,6 +267,28 @@ ThinEngine.prototype.createMultipleRenderTarget = function (size: RenderTargetTe
         // Depth texture
         var depthTexture = new InternalTexture(this, InternalTextureSource.Depth);
 
+        var depthTextureFormat = Constants.TEXTUREFORMAT_DEPTH16;
+        if (this.webGLVersion > 1 &&
+            (options.depthTextueFormat == Constants.TEXTUREFORMAT_DEPTH32_FLOAT ||
+            options.depthTextueFormat == Constants.TEXTUREFORMAT_DEPTH24_STENCIL8)) {
+            depthTextureFormat = options.depthTextueFormat;
+        }
+
+        var depthTextureType = Constants.TEXTURETYPE_UNSIGNED_SHORT;
+        var glInternalDepthTextureFormat = gl.DEPTH_COMPONENT;
+        var glDepthTextureType = gl.UNSIGNED_SHORT;
+        if (this.webGLVersion > 1) {
+            if (depthTextureFormat == Constants.TEXTUREFORMAT_DEPTH32_FLOAT) {
+                depthTextureType = Constants.TEXTURETYPE_FLOAT;
+                glInternalDepthTextureFormat = gl.DEPTH_COMPONENT32F;
+                glDepthTextureType= gl.FLOAT;
+            } else if (depthTextureFormat == Constants.TEXTUREFORMAT_DEPTH24_STENCIL8) {
+                depthTextureType = Constants.TEXTURETYPE_UNSIGNED_INT_24_8;
+                glInternalDepthTextureFormat = gl.DEPTH24_STENCIL8;
+                glDepthTextureType= gl.UNSIGNED_INT;
+            }
+        }
+
         gl.activeTexture(gl.TEXTURE0);
         gl.bindTexture(gl.TEXTURE_2D, depthTexture._hardwareTexture!.underlyingResource);
         gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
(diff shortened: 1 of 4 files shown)


src/Engines/Extensions/engine.multiRender.ts
```diff
@@ -185,6 +186,12 @@ ThinEngine.prototype.createMultipleRenderTarget = function (size: RenderTargetTe
         if (options.samplingModes) {
             samplingModes = options.samplingModes;
         }
+        if (this.webGLVersion > 1 &&
+            (options.depthTextureFormat == Constants.TEXTUREFORMAT_DEPTH24_STENCIL8 ||
```
deltakosh, line 190: ===

deltakosh (approved):
looks good to me

src/Engines/Extensions/engine.multiRender.ts
```diff
@@ -198,12 +205,13 @@ ThinEngine.prototype.createMultipleRenderTarget = function (size: RenderTargetTe
...
-    var depthStencilBuffer = this._setupFramebufferDepthAttachments(generateStencilBuffer, generateDepthBuffer, width, height);
+    var useStencilTexture = this.webGLVersion > 1 && generateDepthTexture && options.depthTextureFormat == Constants.TEXTUREFORMAT_DEPTH24_STENCIL8;
+    var depthStencilBuffer = this._setupFramebufferDepthAttachments(!useStencilTexture && generateStencilBuffer, !generateDepthTexture && generateDepthBuffer, width, height);
 
     rtWrapper._framebuffer = framebuffer;
     rtWrapper._depthStencilBuffer = depthStencilBuffer;
-    rtWrapper._generateDepthBuffer = generateDepthBuffer;
-    rtWrapper._generateStencilBuffer = generateStencilBuffer;
+    rtWrapper._generateDepthBuffer = !generateDepthTexture && generateDepthBuffer;
+    rtWrapper._generateStencilBuffer = !generateDepthTexture && generateStencilBuffer;
```
kircher1, line 214: meant this to be: !useStencilTexture

src/Engines/constants.ts
```diff
@@ -164,6 +164,8 @@ export class Constants {
   public static readonly TEXTUREFORMAT_DEPTH32_FLOAT = 14;
   /** Depth 16 bits */
   public static readonly TEXTUREFORMAT_DEPTH16 = 15;
+  /** Depth 24 bits */
+  public static readonly TEXTUREFORMAT_DEPTH24 = 16;
```
Popov72, line 168: You need to also update the file components/tabs/propertyGrids/materials/texturePropertyGridComponent.tsx of the inspector to add support for this new type of format (see around line 63).

src/Engines/Extensions/engine.multiRender.ts
```diff
@@ -198,12 +205,13 @@ ThinEngine.prototype.createMultipleRenderTarget = function (size: RenderTargetTe
...
 
-    var depthStencilBuffer = this._setupFramebufferDepthAttachments(generateStencilBuffer, generateDepthBuffer, width, height);
+    var useStencilTexture = this.webGLVersion > 1 && generateDepthTexture && options.depthTextureFormat == Constants.TEXTUREFORMAT_DEPTH24_STENCIL8;
+    var depthStencilBuffer = this._setupFramebufferDepthAttachments(!useStencilTexture && generateStencilBuffer, !generateDepthTexture && generateDepthBuffer, width, height);
 
     rtWrapper._framebuffer = framebuffer;
     rtWrapper._depthStencilBuffer = depthStencilBuffer;
-    rtWrapper._generateDepthBuffer = generateDepthBuffer;
-    rtWrapper._generateStencilBuffer = generateStencilBuffer;
+    rtWrapper._generateDepthBuffer = !generateDepthTexture && generateDepthBuffer;
```
kircher1, line 213: If I'm interpreting the spec correctly, there's no point in generating a stencil buffer here either:

    FRAMEBUFFER_UNSUPPORTED if depth and stencil attachments, if present, are not the same image. 

3.7.4 https://www.khronos.org/registry/webgl/specs/2.0/

kircher1: i.e., both _generateDepthBuffer and _generateStencilBuffer should be gated on the same condition (!_generateDepthTexture)

Popov72 (approved):
LGTM.

However, you will need to remove the dist/preview release/babylon.module.d.ts file from the PR.

sebavan:
@kircher1 I ll merge as soon as you ll revert the changes to the babylon/module.d.ts file :-) Thanks a lot for the change

src/Engines/Extensions/engine.multiRender.ts
kircher1: @sebavan if you have any thoughts on this let me know. I'm actually going to leave it how it is in the PR. I realized if I disable the stencil render buffer when depth texture is enabled, that's a breaking change as previous code allowed for this too. And FWIW, I haven't been able to repro any GL errors by binding a stencil render buffer and a depth(/stencil) texture at the same time, so it's unclear to me if that is actually against the spec or not.

Issue #8863, opened 2020-09-03
Separate generating mip maps from using mip mapping for a texture


Popov72:
Currently there's a single property `generateMipMaps` on the texture that is used for both generating the mip maps and saying we want to use mip maps.

It prevents a use case which is to generate custom mip maps and use them for a specific texture.

Also, the other way around, for some reasons one may want to disable mip mapping for textures that have mip maps (for testing purpose or whatever).

Separating those two features would also avoid some custom code that currently is checking that the texture is a compressed texture to not generate the mipmaps, even if `generateMipMaps` is `true`.

DarraghBurkeMS:
This would be relevant for the texture inspector as well. We don't currently handle textures with custom mipmaps differently, but it's something that might be worth looking into. @PatrickRyanMS and I were thinking that we may want to _not_ regenerate mipmaps when the user edits a texture that has custom mipmaps.

thomlucc:
Closing. Please re-open if needed.

PR #11070, opened 2021-09-14
WebGPU: add suport to ShaderMaterial for native WGSL + misc updates

Popov72:
Sorry for the big PR!

A good chunk of it is specific to WebGPU, here's an overview:

## WGSL shader code
* There's a new property that indicates the language a shader is written in (GLSL or WGSL)
  * this property is used by `Effect` to select the right directories to retrieve the shaders from
* WGSL shaders are now in a separate ShadersWGSL/ directory (and ShadersWGSL/ShadersInclude/) and not in Shaders/
  * `ShaderStore`, `Effect` and some gulp scripts have been updated to support this change
* `ShaderMaterial` has been updated to support WGSL shaders
  * all features of WebGPU are supported: external textures, samplers, storage textures, storage buffers
    * a new `ExternalTexture` class has been created for external textures support
    * a new `Sampler` class has been added. The corresponding properties have been moved from `InternalTexture` to `Sampler` and `InternalTexture` now extends `Sampler`
      * adding this class also simplified the code of `WebGPUCacheBindGroup`
  * instances, morphs and bones are working in WGSL (by adding the appropriate `#include<...>` declarations as in GLSL)
  * I will make a doc page to explain how to use WGSL in `ShaderMaterial`
* WGSL shader code
    * I have reused in WGSL the same names for the built-ins than the ones used in GLSL
      * gl_VertexID, gl_InstanceID, gl_Position, gl_FragCoord, gl_FrontFacing, gl_FragDepth, gl_FragColor
      * I could use the names from the WGSL spec but as we can use the names we want it seemed better (for the end user) to use the same names than in GLSL
    * `attribute`, `varying`, `uniform` and `uniform buffers` are declared in the same way than in GLSL
      * in WGSL, the equivalent of attributes and varyings must be enclosed in a struct, meaning we can't easily dispatch these declarations in several files
      * I have reused the `attribute`/`varying`/`uniform` syntax from GLSL and added some wrapping code to comply with the WGSL syntax => I have copied the way GLSLang->Tint->WGSL is doing it
      * the `uniform` declarations are collected and added into the `LeftOver` uniform buffer which is the buffer that is updated when using the `ShaderMaterial.setXXX` methods

## Perf improvements
* there has been some perf improvements
  * the bind group caching mechanism has been improved to avoid querying the cache if no changes occurred in the material context since the last draw call: we just reuse the previous bind groups "as is"
  * the bind group descriptors are pre-created and reused when we need to create the bind groups (lower GC usage)
  * some internal reorg allowed to simplify and improve code in `WebGPUCacheRenderPipeline`
  * use only two bind groups, `#0` for the scene ubo and `#1` for everything else (which is possible because the max number of bindings in a group is 2^16)
    * this allows to issue only two `setBindGroup()` calls instead of 3 or 4.
    * in a future PR, we should handle the scene ubo "by hand" and re-set it only if it has changed. That way, in average, we will only issue a single `setBindGroup()` call to set the group `#1` as group `#0` will change infrequently
    * there's the `WebGPUShaderProcessingContext._SimplifiedKnownBindings` flag that can be set to `false` if we want to reverse to the previous way of handling binding groups (we used bind groups `#0`, `#1` and `#2` for specific known ubos)

## Miscellaneous
* the `webgpu.idl` file has been synchronized with the spec
* some parsing code from `ShaderCodeInliner` has been moved to a separate `Misc/CodeStringParsingTools` class as these functions are also used to parse the WGSL code
* global renaming
  * **set** has been renamed to **group** wherever possible in the WebGPU code because **group** is the named used in the spec for the bindings
  * **sampler** was used to mean **texture** in some internal variable names. As we now have separate sampler objects from texture objects, **sampler** has been renamed to **texture** when appropriate
* `webgpuShaderProcessor` has been split into 3 classes
  * base class `webgpuShaderProcessor`
  * `webgpuShaderProcessorGLSL` that processes shader code written in GLSL
  * `webgpuShaderProcessorWGSL` that processes shader code written in WGSL

[EDIT] I have also updated the WebGPU engine to implement #11065


src/Compute/computeEffect.ts
```diff
@@ -287,13 +289,13 @@ export class ComputeEffect {
         }
 
         // Is in local store ?
-        if (ShaderStore.ShadersStore[shader + key + "Shader"]) {
-            callback(ShaderStore.ShadersStore[shader + key + "Shader"]);
+        if (ShaderStore.GetShadersStore(this._shaderLanguage)[shader + key + "Shader"]) {
```
sebavan, line 292: should this be stored instead of calling every time ?

src/Materials/Textures/externalTexture.ts
```diff
@@ -0,0 +1,71 @@
...
+        this._video = video;
+        this.uniqueId = InternalTexture._Counter++;
+    }
+
+    /**
+     * Get if the texture is ready to be used (downloaded, converted, mip mapped...).
+     * @returns true if fully ready
+     */
+    public isReady(): boolean {
+        return true;
```
sebavan, line 63: Should this not check for the video readyness ?

sebavan (commented):
This is what we can call a PR ;-)

Let s check tomorrow so I get a better grasp at some of the changes before I ll do a second read.

It looks overall great and the only thing I am not too sure is the sampler split, and the fact that ShaderLanguage is an enum knowing we mostly do not rely on enums in the code, also the location in iProcessor might not be ideal but let s check tomorrow.

Playground/src/components/rendererComponent.tsx
```diff
@@ -143,6 +143,16 @@ export class RenderingComponent extends React.Component<IRenderingComponentProps
 
             if (useWebGPU) {
                 globalObject.createDefaultEngine = async function () {
+                    var glslangOptions = {
```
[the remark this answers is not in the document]
Popov72: Reverted

src/Compute/computeEffect.ts
Popov72: Done

src/Materials/Textures/externalTexture.ts
Popov72: Done

RaananW:
Wow! nice PR :-) I'll go over this later today

On a side-not, I believe we should squash this when merging to avoid the long history in future inspection of the code

src/Engines/WebGPU/webgpuShaderProcessorsGLSL.ts
```diff
@@ -0,0 +1,319 @@
...
+        this.webgpuProcessingContext = processingContext as WebGPUShaderProcessingContext;
+
+        this._missingVaryings.length = 0;
+        this._textureArrayProcessing.length = 0;
+    }
+
+    public varyingProcessor(varying: string, isFragment: boolean, preProcessors: { [key: string]: string }, processingContext: Nullable<ShaderProcessingContext>) {
+        this._preProcessors = preProcessors;
+
+        const varyingRegex = new RegExp(/\s*varying\s+(?:(?:highp)?|(?:lowp)?)\s*(\S+)\s+(\S+)\s*;/gm);
```
RaananW, line 45: no need to create a new RegExp if you are already using `/.../`
```suggestion
        const varyingRegex = /\s*varying\s+(?:(?:highp)?|(?:lowp)?)\s*(\S+)\s+(\S+)\s*;/gm;
```

src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
```diff
@@ -0,0 +1,484 @@
...
+        this._attributesWGSL = [];
+        this._attributesDeclWGSL = [];
+        this._attributeNamesWGSL = [];
+        this._varyingsWGSL = [];
+        this._varyingsDeclWGSL = [];
+        this._varyingNamesWGSL = [];
+    }
+
+    public varyingProcessor(varying: string, isFragment: boolean, preProcessors: { [key: string]: string }, processingContext: Nullable<ShaderProcessingContext>) {
+        const varyingRegex = new RegExp(/\s*varying\s+(?:(?:highp)?|(?:lowp)?)\s*(\S+)\s*:\s*(.+)\s*;/gm);
```
RaananW, line 104: can those regexps be reused over classes? (seems to be the same as in  WebGPUShaderProcessorGLSL with a slight modification - (.+) instead of (\S+), so I also wonder if the change is needed).

src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    public textureProcessor(texture: string, isFragment: boolean, preProcessors: { [key: string]: string }, processingContext: Nullable<ShaderProcessingContext>): string {
+        const match = this.textureRegexp.exec(texture);
+        if (match !== null) {
+            const name = match[1]; // name of the variable
+            const type = match[2]; // texture_2d<f32> or array<texture_2d_array<f32>, 5> for eg
+            const isArrayOfTexture = !!match[3];
+            const textureFunc = match[4]; // texture_2d, texture_depth_2d, etc
+            const isStorageTexture = textureFunc.indexOf("storage") > 0;
```
RaananW, line 168: > -1 ? (just wondering if it is intentional)

src/Materials/Textures/sampler.ts
```diff
@@ -0,0 +1,152 @@
+import { Constants } from "../../Engines/constants";
+import { Nullable } from "../../types";
+
+/**
+ * Class used to store a sampler data
+ */
+export class Sampler {
```
RaananW, line 7: As this will be exported as BABYLON.Sampler, could we call it TextureSampler (for example) to not have an abstract name in the main namespace?

src/Materials/shaderMaterial.ts
```diff
@@ -1053,6 +1170,21 @@ export class ShaderMaterial extends Material {
             result.setTexture(key, this._textures[key]);
         }
 
+        // TextureArray
+        for (var key in this._textureArrays) {
```
RaananW, line 1174: nit - `let` instead of `var`

RaananW (commented):
I only comment on the code itself, as i totally trust you when it comes to webgpu-related stuff :-)

And - mostly questions. otherwise LGTM!!

src/Engines/WebGPU/Extensions/engine.externalTexture.ts
```diff
@@ -0,0 +1,19 @@
...
+    return texture;
+};
+
+WebGPUEngine.prototype.setExternalTexture = function (name: string, texture: Nullable<ExternalTexture>): void {
+    if (this._currentEffect) {
+        if (!texture) {
+            this._currentMaterialContext.setTexture(name, null);
+            return;
+        }
+        this._setInternalTexture(name, texture);
```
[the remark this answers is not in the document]
Popov72: Yes, it can be called multiple times. But in any case it's up to the user to handle the lifetime of their external textures (knowing that in WebGPU you can't explicitly delete external textures anyway).

src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
Popov72: Yes, testing for `> 0` is intentional because **storage** can't start at index 0.

src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
Popov72: No, the regexp can be close to each other but the syntax in WGSL is different than the one in GLSL, so I prefer having separate regexp.

src/Materials/Textures/sampler.ts
Popov72: No problem but waiting for confirmation from @sebavan and/or @deltakosh as this will impact some external assets (PGs / Docs), so I prefer doing it a single time :)

src/Materials/shaderMaterial.ts
Popov72: I prefer not changing it because there are 20+ other similar loops in that function that are already existing and that are using `var` and updating the `var` to `const` is not really the goal of the PR (but that could probably be something we should do one day on all the Babylon code base). Only updating this loop to `const` would look strange among the others 20+, that's why I used `var` in the first place.

src/Materials/Textures/sampler.ts
sebavan: I guess TextureSampler is more along the Babylon.js Linguo, I am up for the change

src/Engines/thinEngine.ts
```diff
@@ -453,7 +456,7 @@ export class ThinEngine {
     /** @hidden */
     protected _boundTexturesCache: { [key: string]: Nullable<InternalTexture> } = {};
     /** @hidden */
-    protected _currentEffect: Nullable<Effect>;
+    public _currentEffect: Nullable<Effect>;
```
sebavan, line 459: Is there a way to avoid the public ?

src/Materials/Textures/internalTexture.ts
```diff
@@ -111,14 +112,20 @@ export class InternalTexture {
...
-     * Gets the sampling mode of the texture
-     */
-    public samplingMode: number = -1;
     /**
      * Gets a boolean indicating if the texture needs mipmaps generation
      */
     public generateMipMaps: boolean = false;
+    /**
+     * Gets a boolean indicating if the texture uses mipmaps
+     * TODO implements useMipMaps as a separate setting from generateMipMaps
```
sebavan, line 121: Could you remind me why we wanted the split ?

src/Materials/effect.ts
```diff
@@ -12,12 +12,17 @@ import { IMatrixLike, IVector2Like, IVector3Like, IVector4Like, IColor3Like, ICo
 import { ThinEngine } from '../Engines/thinEngine';
 import { IEffectFallbacks } from './iEffectFallbacks';
 import { ShaderStore as EngineShaderStore } from '../Engines/shaderStore';
+import { ShaderLanguage } from "./shaderLanguage";
+import { CodeStringParsingTools } from "../Misc/codeStringParsingTools";
```
sebavan, line 16: This should probably not be here as it is a pretty large dependency added in our min build ?

src/Materials/effect.ts
```diff
@@ -325,6 +338,9 @@ export class Effect implements IDisposable {
 
         };
         this._loadShader(vertexSource, "Vertex", "", (vertexCode) => {
+            if (processorOptions.processor?.removeCommentsBeforeProcessing) {
```
sebavan, line 341: Why adding this here and not in the processor ?

src/Materials/effect.ts
```diff
@@ -336,6 +352,9 @@ export class Effect implements IDisposable {
             }, this._engine);
         });
         this._loadShader(fragmentSource, "Fragment", "Pixel", (fragmentCode) => {
+            if (processorOptions.processor?.removeCommentsBeforeProcessing) {
```
sebavan, line 355: Why adding this here and not in the processor ?

src/Materials/effect.ts
```diff
@@ -869,6 +890,24 @@ export class Effect implements IDisposable {
         this._engine.setTextureArray(this._samplers[channel], this._uniforms[channel], textures, channel);
     }
 
+    /**
```
sebavan, line 893: This is so annoying to see WebGPU specifics here :-( I wonder what choice we could have ? I guess WebGPUEngine could extend effect ??? a bit like we do in engine extensions.

Always thinking to our min bundle

src/Materials/effect.ts
```diff
@@ -911,6 +950,15 @@ export class Effect implements IDisposable {
         this._engine.bindUniformBlock(this._pipelineContext!, blockName, index);
     }
 
+    /**
```
sebavan, line 953: Same here, could be great to move in a webgpu extension

src/Materials/shaderMaterial.ts
```diff
@@ -458,6 +506,51 @@ export class ShaderMaterial extends Material {
...
+        return this;
+    }
+
+    /**
+     * Set a storage buffer in the shader
+     * @param name Define the name of the storage buffer as defined in the shader
+     * @param buffer Define the value to give to the uniform
+     * @return the material itself allowing "fluent" like uniform updates
+     */
+     public setStorageBuffer(name: string, buffer: StorageBuffer): ShaderMaterial {
```
sebavan, line 545: Like for effects I wonder if it makes sense to move in a webgpu extension ?

sebavan: It probably makes less sense here :-)

sebavan (changes requested):
LGTM !!!

Mostly nit beside bundle size.

src/Engines/Processors/iShaderProcessor.ts
```diff
@@ -1,14 +1,24 @@
...
 import { Nullable } from "../../types";
 import { ShaderProcessingContext } from "./shaderProcessingOptions";
 
 declare type ThinEngine = import("../thinEngine").ThinEngine;
 
 /** @hidden */
 export interface IShaderProcessor {
+    shaderLanguage: ShaderLanguage;
+
+    uniformRegexp?: RegExp;
```
sebavan, line 11: last one, I wondering why we need dynamic regex at the interface level here

sebavan: I am actually not seeing where they are set but the PR is huge so I have an excuse ;-)

src/Engines/thinEngine.ts
Popov72: I reverted it back to `protected` as I checked the value of `_currentEffect` in `engine.externalTexture` but I really didn't have to do it.

src/Materials/Textures/internalTexture.ts
Popov72: See #8863

src/Materials/effect.ts
Popov72: This code was in `ShaderCodeInliner` previously, which is in the min build I guess?

I think I can move the dependency to `ShaderProcessor` but I don't know if it's better?

Popov72: Code moved to `ShaderProcessor`

src/Materials/effect.ts
Popov72: Code moved to `ShaderProcessor`

src/Materials/effect.ts
Popov72: Code moved to `ShaderProcessor`

src/Materials/effect.ts
Popov72: Code moved.

src/Materials/effect.ts
Popov72: Code moved.

src/Materials/shaderMaterial.ts
Popov72: I could move the `setXXX` functions but there are some other code related to these new objects (storage texture, storage buffer, sampler) that are embedded into existing functions that I can't move. So I think it's better to keep the functions here?

src/Engines/Processors/iShaderProcessor.ts
Popov72: Those regexp are used by both `ShaderCodeNode` and the webgpuShaderProcessorGLSL/webgpuShaderProcessorWGSL classes so I factored them in the interface. If `undefined` for one type of object, it means `ShaderCodeNode` should not called the corresponding interface method for this type of object (`uniformProcessor`, `textureProcessor`, etc).

src/Engines/Processors/shaderProcessor.ts
```diff
@@ -9,6 +9,8 @@ import { ShaderDefineExpression } from './Expressions/shaderDefineExpression';
 import { ShaderDefineArithmeticOperator } from './Expressions/Operators/shaderDefineArithmeticOperator';
 import { ProcessingOptions } from './shaderProcessingOptions';
 import { _DevTools } from '../../Misc/devTools';
+import { ShaderLanguage } from "../../Materials/shaderLanguage";
+import { CodeStringParsingTools } from "../../Misc/codeStringParsingTools";
```
sebavan, line 13: This as lots of WebGPUSpecific in growing the size of ThinEngine bundles, so either you split the file, or in the file, do not create a class and simply export the functions separately so tree shakers will only grab what they truly use :-)

sebavan:
Only 2 changes and I ll merge it in ! GG @Popov72

src/Materials/Textures/sampler.ts
Popov72: Done

src/Engines/Processors/shaderProcessor.ts
Popov72: I have done something for that, let me know if it's ok for you.

PR #11071, opened 2021-09-14
Prevent uncatchable OOB exception in GLTF loader

### LINKED ISSUE #11052: GLTFFileLoader.loadFile does not call onError callback for .GLB files with less than 20 characters
**Summary**
For corrupt GLB files with less than 20 bytes, `GLTFFileLoader.loadFile` throws an unhandled exception instead of calling the `onError` callback (see screenshots below).

Validated bug: https://forum.babylonjs.com/t/gltffileloader-loadfile-does-not-call-onerror-callback-for-files-with-less-than-20-characters/22669

**Repro**
- Bug repro on [sandbox](https://sandbox.babylonjs.com/).  Please note that this same issue exists in other, customer implementations with `@babylonjs/loaders` but the generic babylon sandbox is an easy repro.
- Expected result:
Upload a .GLB file with >= 20 characters.  Example contents:
```json
{
	"hello": "123abc"
}
```
In this scenario, the generic error message is displayed and the sandbox state is reset.
- Current result:
Upload a .GLB file with < 20 characters.  Example contents:
```json
{
	"hello": "123"
}
```
In this scenario, no error message is displayed and the sandbox stalls in the loading state.

**Screenshots**
When uploading a .GLB file with >= 20 characters.
image

When uploading a .GLB with < 20 characters and waiting a sufficient amount of time.
image

Browser console
image

**Desktop:**
 - OS: Windows, Mac
 - Browser: Chrome, Safari, Firefox
 - Version: Latest

**Additional context**
I was able to track the issue down to the construction of the Unt8Array in the following codeblock (starting on line 538). Fundamentally the issue is an out-of-bounds error resulting from the combination of byteOffset and byteLength being greater than the file length (in this case, 19).

`loaders/src/glTF/glTFFileLoader.ts`
```javascript
  return this._loadFile(scene, fileOrUrl, (data) => {
      const arrayBuffer = data as ArrayBuffer;
      this._unpackBinaryAsync(new DataReader({
          readAsync: (byteOffset, byteLength) => Promise.resolve(new Uint8Array(arrayBuffer, byteOffset, byteLength)),
          byteLength: arrayBuffer.byteLength
      })).then((loaderData) => {
          onSuccess(loaderData);
      }, onError);
  }, true, onError);
```

While most of the time this won’t be an issue, given the architecture of the loaders, it is nearly impossible to catch the error from the surface of the library so when a <= 19 character file is submitted, a crash is currently unpreventable.

sebavan:
https://forum.babylonjs.com/t/gltffileloader-loadfile-does-not-call-onerror-callback-for-files-with-less-than-20-characters/22669

### DIFF
--- a/loaders/src/glTF/glTFFileLoader.ts
+++ b/loaders/src/glTF/glTFFileLoader.ts
@@ -538,7 +538,14 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
             return this._loadFile(scene, fileOrUrl, (data) => {
                 const arrayBuffer = data as ArrayBuffer;
                 this._unpackBinaryAsync(new DataReader({
-                    readAsync: (byteOffset, byteLength) => Promise.resolve(new Uint8Array(arrayBuffer, byteOffset, byteLength)),
+                    readAsync: (byteOffset, byteLength) => {
+                        if (byteOffset + byteLength > arrayBuffer.byteLength) {
+                            return Promise.reject("ArrayBuffer not big enough to be read.");
+                        }
+
+                        const array = new Uint8Array(arrayBuffer, byteOffset, byteLength);
+                        return Promise.resolve(array);
+                    },
                     byteLength: arrayBuffer.byteLength
                 })).then((loaderData) => {
                     onSuccess(loaderData);


bghgary (changes requested):
I already have a local fix for this

sebavan:
Np I ll close it and wait on your PR

PR #11077, opened 2021-09-14
Adds a particle spritesheet loop flag

### LINKED ISSUE #11072: [ParticleSystem] Need parameter for spriteCellLoop
The particle sprite system has parameters for startSpriteCellID and endSpriteCellID but it would be useful to be able to control whether the animation of these cells loops or not. If we add a spriteCellLoop parameter this should be the behavior:

- True: the sprite animation loops from startSpriteCellID to endSpriteCellID and then starts again at startSpriteCellID
- False: the sprite animation loops from startSpriteCellID to endSpriteCellID and then holds on endSpriteCellID stopping the animation.

sebavan:
Fix #11072


Popov72 (approved):
LGTM!

PR #11080, opened 2021-09-15
Update to responsive design

msDestiny14:
https://github.com/BabylonJS/ThePirateCove/issues/73

### DIFF
--- a/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
+++ b/guiEditor/src/components/propertyTab/propertyTabComponent.tsx
@@ -406,7 +406,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
             { label: "Web (1920)", value: 0 },
             { label: "Phone (720)", value: 1 },
             { label: "Square (1200)", value: 2 },
-            { label: "Custom", value: 3 }];
+            ];
         const sizeValues = [
             new Vector2(1920, 1080),
             new Vector2(750, 1334),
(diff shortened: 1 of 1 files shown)


guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -446,19 +453,19 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
                         noDirectUpdate={true}
                         onSelect={(value: any) => {
                             this._sizeOption = value;
-                            if (this._sizeOption != (sizeOptions.length - 1)) {
+                            if (this._sizeOption != (sizeOptions.length)) {
```
RaananW, line 456:
```suggestion
                            if (this._sizeOption !== (sizeOptions.length)) {
```

Always use `===` or `!==`, unless it is really needed to compare different types. It's less of a performance issue and more of a runtime-error-handling thing. this should never work if one of those types is different than the other.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -446,19 +453,19 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
-                    />
-                    {this._sizeOption == (sizeOptions.length - 1) &&
+                    />}
+                    {this._sizeOption == (sizeOptions.length) &&
                         <div className="divider">
                             <FloatLineComponent
                                 icon={canvasSizeIcon}
                                 iconLabel="Canvas Size"
-                                label="X"
+                                label="W"
```
RaananW, line 468: was it intentional to change the label here ( i assume this is the keycode used?)

RaananW (commented):
small nits.

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Yes. This is to match the design. W for width and H for height.

Issue #11075, opened 2021-09-14
Gizmo Does Not Update Projection Texture


PirateJC:
Using this playground as an example:
https://playground.babylonjs.com/#7UHPQ0

The spotlight is positioned directly in front of the box, projecting a texture that looks like this:
![image](https://user-images.githubusercontent.com/1668182/133341564-5abb6210-39c8-49cb-8140-c2b0e6cbdfd2.png)

When manually changing the position of the spotlight via code, the texture projection updates with the light like this:
![image](https://user-images.githubusercontent.com/1668182/133341671-118f5d22-8b0f-4144-8e68-82137bfee0ec.png)

However, if you use the gizmo to change the light's position, the texture "position" does not update accordingly:
![image](https://user-images.githubusercontent.com/1668182/133341766-5ceeb272-da0a-4c0c-b1a0-f19b114fdd0c.png)

The texture position should move whenever the light position changes. Currently the behavior seems broken with the gizmo.

PR #11081, opened 2021-09-15
Fix light update gizmo

CedricGuillemet:
fixes https://github.com/BabylonJS/Babylon.js/issues/11075
Light projection matrix update is done thanks to a dirty flag. That flag is set to true in the setter. But copying values does call the setter and no update on the projection matrix. Calling setter instead.

### DIFF
--- a/src/Gizmos/gizmo.ts
+++ b/src/Gizmos/gizmo.ts
@@ -16,6 +16,7 @@ import { PointerEventTypes, PointerInfo } from '../Events/pointerEvents';
 import { LinesMesh } from '../Meshes/linesMesh';
 import { PointerDragBehavior } from "../Behaviors/Meshes/pointerDragBehavior";
 import { ShadowLight } from "../Lights/shadowLight";
+import { Light } from "../Lights/light";
 
 /**
  * Cache built by each axis. Used for managing state between all elements of gizmo for enhanced UI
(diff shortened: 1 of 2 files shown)


src/Gizmos/gizmo.ts
```diff
@@ -331,6 +332,23 @@ export class Gizmo implements IDisposable {
...
+                var invParent = this._tempMatrix1;
+                var nodeLocalMatrix = this._tempMatrix2;
+                parent.getWorldMatrix().invertToRef(invParent);
+                light.getWorldMatrix().multiplyToRef(invParent, nodeLocalMatrix);
+                nodeLocalMatrix.decompose(undefined, this._tempQuaternion, this._tempVector);
+            } else {
+                this._attachedNode._worldMatrix.decompose(undefined, this._tempQuaternion, this._tempVector);
+            }
+            // setter doesn't copy values. Need a new Vector3
+            light.position = new Vector3(this._tempVector.x, this._tempVector.y, this._tempVector.z);
```
RaananW, line 349: you can use .copyFrom instead of creating a new vector

src/Gizmos/gizmo.ts
```diff
@@ -331,6 +332,23 @@ export class Gizmo implements IDisposable {
                 lmat.copyFrom(bone.getWorldMatrix());
             }
             bone.markAsDirty();
+        } else if (this._attachedNode.getClassName() === "SpotLight" || this._attachedNode.getClassName() === "PointLight" || this._attachedNode.getClassName() === "DirectionalLight") {
```
RaananW, line 335: is it possible to use feature-detection here? You need a shadow-light, you can try detecting if the attachedNode has a function that only a shadow light has (like `position` or `direction`)

src/Gizmos/gizmo.ts
CedricGuillemet: Nop. That's what was done in LightGizmo and setter is not called...and dirty flag was unchanged. Or I missed something?

RaananW: We don't need the setter if position is set. You use the getter to get the vector and update it. Unless you need the setter to be called?

CedricGuillemet: yes, I need the setter to be called.

src/Gizmos/gizmo.ts
CedricGuillemet: I've just seen the GetTypeId. Using it instead.

src/Gizmos/gizmo.ts
RaananW: if there is no way around it it's ok, of course :-)
If it's called on each frame i would prefer adding a "setDirty" function to the light so that you can set it yourself (because of gc), but if it's not the case we can leave it this way

Popov72 (approved):

RaananW (approved):
Just a linting issue

src/Gizmos/gizmo.ts
CedricGuillemet: well, there are pros and cons for setDirty. As it's an internal state, it's awkward to force its state to recompute that way. That's a debate for another day :)

Issue #11091, opened 2021-09-15
Scaling a parent node (with no position set) can move a spatial audio position


Exolun:
**Repro**
- Bug repro on [Audio Positioning Test](https://www.babylonjs-playground.com/#0CSD1T#257):
- Expected result:  Audio plays when the scene loads
- Current result:  Can't hear audio - the scaling on the parent throws its position out of audible range.  (Despite the direct parent not having a position set)

**Desktop (please complete the following information):**
 - Device: PC
 - OS: Windows 10 Enterprise
 - Browser Chrome
 - Version 93.0.4577.63

RaananW:
The issue here is that you are creating a mesh without a geometry which is therefore not having any bounding information created upon construction.
This causes the bounding information that is generated after applying the parent to be incorrect. There are a few solutions here:

Use a transform node instead:

https://www.babylonjs-playground.com/#0CSD1T#258

Force a refresh on the bounding information before applying the parent:

https://www.babylonjs-playground.com/#0CSD1T#259

Use custom attenuation (check the sound options). This will use the mesh's absolute position instead of its bounding information

https://www.babylonjs-playground.com/#0CSD1T#260

This is a bug per se, but we might be able to solve this on our end as well. I will investigate for a future solution that won't break the current implementation.

RaananW:
To further explain the situation - when creating the bounding info the initial min and max values are taken using the absolute position of the mesh. This illustrates the issue:

https://playground.babylonjs.com/debug.html#YGAA79#3

You would expect both bounding information to be the same (at least the center and center world), but they are not due to the fact that the 1st mesh's bounding info is set after the parent was set so absolute position is the parent's.

Exolun:
Ah makes sense!  Should the sound attaching to the mesh force a refresh on the bounds to correct the position?

RaananW:
If you are constructing the bounding info before setting the parent the update will be taken care of.
The fix i pushed should solve this with no interaction on your part, but for now make sure to call `refreshBoundingInfo` after creating a mesh with no geometry

Exolun:
Yep in this case the audio is created/attached dynamically (on user action) long after the bounds are initially calculated.  

We're putting in a call to `refreshBoundingInfo` in the meantime.  Thanks for the fix!

PR #11093, opened 2021-09-16
initial bounding info should be based on the local position

RaananW:
Related to #11091

When creating a mesh without a geometry and setting its parent the next call to create bounding info will create it using the absolute position (9i.e. the parent position's as well).
This is not the case for meshes with a geometry.

I don't see a reason this will break any existing implementation. meshes created by MeshBuilders don't have a parent when initialized, and the min-max should anyhow be local, as the world matrix is the one in charge of transforming the world center/min/max to the right coordinates

The issue can be seen here - https://playground.babylonjs.com/debug.html#YGAA79#3

Check the console and see the local center of both meshes - it should be the same, but it isn't because of the usage of absolutePosition (as the first mesh's bounding info is created only after the parent is set while the sphere's is created when the geometry is created).


sebavan (approved):
GG

PR #11094, opened 2021-09-16
Updated Geodesic and Goldberg Polyhera

BabylonJSGuide:
@sebavan, @RaananW, @Popov72 this is a tidied up version. Have solved the import problem by switching from a Goldberg Mesh class to using a mixin.

Here are some very rough drafts of of it working (my workshop tests).
https://relaxed-varahamihira-799310.netlify.app/
https://relaxed-varahamihira-799310.netlify.app/test2.html  upto test7.html

Code https://github.com/BabylonJSGuide/Goldberg


RaananW:
Awesome!

The package-lock is still here, would you be able to revert this commit?
Also the linting warnings and errors are still there - https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=15370&view=logs&jobId=32e2e1bb-a28f-5b18-6cfc-3f01273f5609&j=32e2e1bb-a28f-5b18-6cfc-3f01273f5609&t=2bf39655-ae43-5ae7-fe00-6e76c9bf09b6

src/Meshes/Builders/goldbergBuilder.ts
```diff
@@ -0,0 +1,606 @@
...
+Mesh.CreateGoldberg = (name: string, options: { m?: number, n: number, size?: number, sizeX?: number, sizeY?: number, sizeZ?: number, updatable?: boolean, sideOrientation?: number }, scene: Scene): Goldberg => {
+    return GoldbergBuilder.CreateGoldberg(name, options, scene);
+}; */
+
+/**
+ * Class containing static functions to help procedurally build a Goldberg mesh
+ */
+ export class GoldbergBuilder {
+    /**
+     * Creates the Mesh for a Geodesic Polyhedron
```
sebavan, line 329: The comment sounds wrong -> should it not be about Goldberg also a quick link to what Goldberg is ? and I guess it should be renamed to GoldbergPolyhedron ?

src/Meshes/mesh.ts
```diff
@@ -4508,6 +4508,90 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
+        throw _DevTools.WarnImport("MeshBuilder");
+    }
+
+    /**
+     * Extends a mesh to a Goldberg mesh
+     * @param mesh the mesh to convert
+     * Warning  the mesh to convert MUST be an import of a peviously exported Goldberg mesh     
+     */
+    public static ToGoldberg(mesh: Mesh) {
+    }
```
sebavan, line 4593: Should probably have a warning here: throw _DevTools.WarnImport("MeshBuilder");

src/Meshes/Builders/geodesicBuilder.ts
```diff
@@ -0,0 +1,77 @@
...
+import { _PrimaryIsoTriangle, GeodesicData} from "../geodesicMesh"
+
+Mesh.CreateGeodesic = (name: string, options: { m?: number, n: number, size?: number, sizeX?: number, sizeY?: number, sizeZ?: number, custom?: any, faceUV?: Vector4[], faceColors?: Color4[], updatable?: boolean, sideOrientation?: number }, scene: Scene): Mesh => {
+    return GeoBuilder.CreateGeodesic(name, options, scene);
+};
+
+/**
+ * Class containing static functions to help procedurally build a geodesic mesh
+ */
+ export class GeoBuilder {
```
sebavan, line 17: Rename to GeodesicBuilder ?

sebavan (changes requested):
Could you also remove most of the `any` and replace with proper types ?

This looks better and better :-)

src/Meshes/Builders/goldbergBuilder.ts
```diff
@@ -0,0 +1,606 @@
...
+    vertexData.normals = normals;
+    vertexData.uvs = uvs;
+    return vertexData;
+};
+
+/**
+ * Class adding properties and methods to a Mesh to form a Goldberg Mesh
+ * Not intended to be used for the direct creation of a Goldberg Mesh
+ */
+//export class Goldberg extends Mesh {
```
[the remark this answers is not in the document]
BabylonJSGuide: Not needed I forgot to delete it. Too excited that I got the mix in to work.

BabylonJSGuide:
@RaananW The package-lock is still here, would you be able to revert this commit?

No idea how to do this. If I do find out how do I leave all subsequent commits. Help much appreciated here.

BabylonJSGuide:
@RaananW Also the linting warnings and errors are still there

Sorry about this they did not come up with gulp typedoc-check. It seems I have forgot a step because I now remember getting a listing like this for a previous PR a couple of years back.

BabylonJSGuide:
> Could you also remove most of the `any` and replace with proper types ?
> 
> This looks better and better :-)

Will have another look. Did try when I was doing it but stuck with  any when VSCode objected to my attempt at a type when using the variable. Any suggestions gratefully received.

src/Meshes/Builders/geodesicBuilder.ts
BabylonJSGuide: Done

src/Meshes/mesh.ts
BabylonJSGuide: MeshBuilder is not an alternative here. This takes a mesh that is an import of a previously exported Goldberg mixin extension to a mesh and adds back the mixin. As such I believe that 

BABYLON.Mesh.ToGoldberg(mesh)

makes more sense to the user than

BABYLON.MeshBuilder.ToGoldberg(mesh)

Popov72:
> Sorry about this they did not come up with gulp typedoc-check. It seems I have forgot a step because I now remember getting a listing like this for a previous PR a couple of years back.

You can check the linting by doing: `gulp core-tsLint`

src/Meshes/Builders/goldbergBuilder.ts
BabylonJSGuide: Done

src/Meshes/Builders/geodesicBuilder.ts
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Creates the Mesh for a Geodesic Polyhedron
```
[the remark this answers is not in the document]
BabylonJSGuide: Done

src/Meshes/mesh.ts
sebavan: Yup agree but still the user will need to import goldbergBuilder in order to have the function available which you should warn about here :-)

throw _DevTools.WarnImport("goldbergBuilder");

sebavan: I would actually even keep it completely outside of mesh and do not add a static here. It is not needed. We only did it in other case for back compat which is a pain for ES6 so far :-) @RaananW can confirm.

BabylonJSGuide:
@sebavan, @RaananW, @Popov72 thank you for your guidance. I have (hopefully) changed everything that needed changing and everything is now fit for use. My Goldberg repo now contains the latest version of babylon.max.js as compiled by this PR.

Please review and let me know id further work is needed.

BabylonJSGuide:
@sebavan re - I am still seeing a submodule being added into the repo, the babylon serializers js being modified and the ToGoldberg function on the mesh class.

No idea how the submodule or serialized change happened. I can delete the submodule and edit the serialized file back to what it should be. Is this the way to go?

Do not understand the problem with ToGoldberg. It is needed since if you export a Goldberg mesh and re-import it  it imports as a mesh without the added properties and methods needed to manipulate a Goldberg mesh should a user wish to use the extended version of the mesh.

sebavan:
about the first 2 comments yes removing the submodule and editing back the serializers should work.

About ToGoldberg let s check what @RaananW thinks ?

BabylonJSGuide:
About ToGoldberg let s check what @RaananW thinks ?

Could move it to Tools?

RaananW (commented):
Looks good to me!

A few comments:

1) There is no need to add the CreateFunctions to mesh. we are slowly moving out of this, and extending the mesh class is not needed.
2) You commited a built version of babylon.seralizers.js. is there a reason for that?

BabylonJSGuide:
> Looks good to me!
> 
> A few comments:
> 
> 1. There is no need to add the CreateFunctions to mesh. we are slowly moving out of this, and extending the mesh class is not needed.

I am more than happy to remove CreateGeodesic and CreateGoldberg from Mesh.ts
 
What I struggled with is a Goldberg mesh has added properties and methods in addition to the usual mesh ones. When exporting a Goldberg mesh the property values are stored in the mesh's metadata. When this is then imported it's imported as a mesh which of course does not have the additional properties and methods. Using JavaScript I could easily write a function that just adds the properties and methods on the fly. As far as I found out this is not possible with Typescript. Using a mixin I wrote a function ToGoldberg that takes a mesh and adds back the missing properties and methods. This function does  this by dynamically creating the Goldberg mesh from the mesh parameter and then disposing of the passed mesh. If I could find a way of just adding the extra properties and methods 


> 2. You commited a built version of babylon.seralizers.js. is there a reason for that?

As @sebavan pointed out, though I do not know how, but I had a commit that changed a serialized and the only way I could think of that I was submitting a correct version was to copy and paste the one on Babylon/babylon.js

RaananW:
> The code is from line 300 of https://github.com/BabylonJSGuide/Babylon.js/tree/master/src/Meshes.

Sorry - what file in this directory? couldn't find it in mesh. 
I assume the code is already in the framework itself? otherwise it should also be included in this PR. Or I might be misunderstanding (which is more than likely :-))

BabylonJSGuide:
>There is no need to add the CreateFunctions to mesh. we are slowly moving out of this, and extending the mesh class is not needed.

Removed

Also renamed ToGoldberg as ExtendToGoldberg

src/Meshes/Builders/geodesicBuilder.ts
```diff
@@ -0,0 +1,78 @@
+import { Scene } from "../../scene";
+import { Vector4 } from "../../Maths/math.vector";
+import { Color4 } from '../../Maths/math.color';
+import { Mesh } from "../../Meshes/mesh";
+import { PolyhedronBuilder } from "./polyhedronBuilder";
+import { Nullable } from '../../types';
+import { Logger } from "../../Misc/logger";
+import { _PrimaryIsoTriangle, GeodesicData } from "../geodesicMesh";
+
+Mesh.CreateGeodesic = (name: string, options: { m?: number, n: number, size?: number, sizeX?: number, sizeY?: number, sizeZ?: number, faceUV?: Vector4[], faceColors?: Color4[], updatable?: boolean, sideOrientation?: number }, scene: Scene): Mesh => {
```
RaananW, line 10: I guess this should be removed as well

src/Meshes/mesh.ts
RaananW: > @RaananW sorry copy and paste error
> 
> https://github.com/BabylonJSGuide/Babylon.js/blob/master/src/Meshes/Builders/goldbergBuilder.ts
> 
> Lines 150 to 328 for mixin
> 
> From 330 for ToGoldberg
> 
> Line 113 uses mixin for CreateGoldberg

Look good to me! It feels like it can be a part of the mesh prototype (as opposed to a static), but you can leave it this way. as long as it is separated and can be tree-shaken it is totally fine.

RaananW (approved):
This is awesome! 
Just a few build issues. i'll merge it right after.
Now comes the fun part - documenting the entire thing ;-)

sebavan:
I can not help help wondering if we should keep extending prototypes or classes in new code. Everything is accessible on the individual classes so I wonder if extending VertexData, MeshBuilder or Mesh is bringing a lot of value ?

We had to do it for back compat in the past but as this is brand new, I am wondering ?

BabylonJSGuide:
> 
> 
> I can not help help wondering if we should keep extending prototypes or classes in new code. Everything is accessible on the individual classes so I wonder if extending VertexData, MeshBuilder or Mesh is bringing a lot of value ?
> 
> We had to do it for back compat in the past but as this is brand new, I am wondering ?

In adding the Geodesic and the Goldberg I followed the pattern for creating the other meshes. If you are suggesting there is a different pattern because these are new classes I am totally lost as to where to go.

sebavan:
@BabylonJSGuide the comment was not about your current code :-) but more in general for me an @RaananW .

I merged it as is so you can document it and @RaananW will review the pattern later so that we could enhance it for 5.0. And then you ll have some new examples to follow :-)

Thanks a lot for your contribs.

BabylonJSGuide:
@sebavan, @RaananW, @Popov72 thank you all for your reviews and suggestions. I am working on the documentation.

PR #11102, opened 2021-09-17
Support WebP in environmentTextureTools

simonihmig:
This adds support for encoding and decoding WebP images in .env files. See https://forum.babylonjs.com/t/support-webp-for-env-maps/23644

The default type remains PNG, so the changes should be fully backwards compatible. I tested this by locally changing the arguments to [CreateEnvTextureAsync](https://github.com/BabylonJS/Babylon.js/blob/master/inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx#L254) in the Inspector, exporting some sample environments with different settings, and dropping them again into the sandbox. Everything seems to work fine. Not sure if there is some additional test coverage we can add here?

The second commit adds an optional `quality` argument to a bunch of Blob-related tools, so that it can be passed finally to `canvas.toBlob()`, which supports this argument for WebP images (at least in Chrome).

The results are quite satisfying. The playground's original [`environment.dds`](https://playground.babylonjs.com/textures/environment.dds) gave these results when converted to .env:
* using PNG: 271KB
* using WebP w/ default quality (0.8): 71KB
* using WebP w/ quality=0.5: 64KB
* using WebP w/ quality=0.1: 57KB

You can find these sample files [here](https://drive.google.com/drive/folders/1QKwSN0gl8EA5PzWRF-TUsHyCNLV4hrCR?usp=sharing)

I didn't update the Inspector to allow exporting .env files using WebP yet, but can do this in a separate PR when this is merged!

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -45,6 +45,7 @@
 - Modified behavior for FreeCamera and ArcRotateCamera so that default mouse dragging movements now account for what button was used to initiate it ([PolygonalSun](https://github.com/PolygonalSun))
 - Added coroutine capabilities to `Observable`s ([syntheticmagus](https://github.com/syntheticmagus))
 - Added a global OnTextureLoadErrorObservable to handle texture loading errors during model load ([RaananW](https://github.com/RaananW))
+- Add support to encode and decode .env environment textures using WebP instead of PNG ([simonihmig](https://github.com/simonihmig))
 
 ### Engine
 
--- a/src/Misc/environmentTextureTools.ts
+++ b/src/Misc/environmentTextureTools.ts
@@ -20,8 +20,7 @@ import "../Materials/Textures/baseTexture.polynomial";
 import "../Shaders/rgbdEncode.fragment";
 import "../Shaders/rgbdDecode.fragment";
 
-type EnvironmentTextureImageType = 'image/png' | 'image/webp';
-const defaultEnvironmentTextureImageType: EnvironmentTextureImageType = 'image/png';
+const defaultEnvironmentTextureImageType = 'image/png';
 
 /**
  * Raw texture data and descriptor sufficient for WebGL texture upload
@@ -48,9 +47,9 @@ export interface EnvironmentTextureInfo {
     specular: any;
 
     /**
-     * The image type used to encode the image data.
+     * The mime type used to encode the image data.
      */
-    imageType?: EnvironmentTextureImageType;
+    imageType?: string;
 }
 
 /**
@@ -104,11 +103,14 @@ interface EnvironmentTextureIrradianceInfoV1 {
     xy: Array<number>;
 }
 
+/**
+ * Options for creating environment textures
+ */
 export interface CreateEnvTextureOptions {
     /**
      * The mime type of encoded images.
      */
-    imageType?: EnvironmentTextureImageType;
+    imageType?: string;
 
     /**
      * the image quality of encoded WebP images.
@@ -445,7 +447,7 @@ export class EnvironmentTextureTools {
      * @param imageType the mime type of the image data
      * @returns a promise
      */
-    public static UploadLevelsAsync(texture: InternalTexture, imageData: ArrayBufferView[][], imageType: EnvironmentTextureImageType = defaultEnvironmentTextureImageType): Promise<void> {
+    public static UploadLevelsAsync(texture: InternalTexture, imageData: ArrayBufferView[][], imageType: string = defaultEnvironmentTextureImageType): Promise<void> {
         if (!Tools.IsExponentOfTwo(texture.width)) {
             throw new Error("Texture size must be a power of two");
         }
--- a/src/Misc/environmentTextureTools.ts
+++ b/src/Misc/environmentTextureTools.ts
@@ -105,7 +105,15 @@ interface EnvironmentTextureIrradianceInfoV1 {
 }
 
 export interface CreateEnvTextureOptions {
+    /**
+     * The mime type of encoded images.
+     */
     imageType?: EnvironmentTextureImageType;
+
+    /**
+     * the image quality of encoded WebP images.
+     */
+    imageQuality?: number;
 }
 
 /**
@@ -163,6 +171,7 @@ export class EnvironmentTextureTools {
      * @param texture defines the cube texture to convert in env file
      * @param options options for the conversion process
      * @param options.imageType the mime type for the encoded images, with support for "image/png" (default) and "image/webp"
+     * @param options.imageQuality the image quality of encoded WebP images.
      * @return a promise containing the environment data if successful.
      */
     public static async CreateEnvTextureAsync(texture: BaseTexture, options: CreateEnvTextureOptions = {}): Promise<ArrayBuffer> {
@@ -223,7 +232,7 @@ export class EnvironmentTextureTools {
 
                 const rgbdEncodedData = await engine._readTexturePixels(tempTexture, faceWidth, faceWidth);
 
-                const imageEncodedData = await Tools.DumpDataAsync(faceWidth, faceWidth, rgbdEncodedData, imageType, undefined, false, true);
+                const imageEncodedData = await Tools.DumpDataAsync(faceWidth, faceWidth, rgbdEncodedData, imageType, undefined, false, true, options.imageQuality);
 
                 specularTextures[i * 6 + face] = imageEncodedData as ArrayBuffer;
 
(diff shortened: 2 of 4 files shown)


sebavan:
@simonihmig This is huge !!! I ll review in a couple hours thanks a lot !!!

dist/preview release/what's new.md
```diff
@@ -1,328 +1,329 @@
-# 5.0.0
```
deltakosh, line 1: Can you check what is wrong with this file? We cannot compare them :(

src/Misc/environmentTextureTools.ts
```diff
@@ -113,7 +135,7 @@ export class EnvironmentTextureTools {
      * @param data The array buffer containing the .env bytes.
      * @returns the environment file info (the json header) if successfully parsed.
      */
-    public static GetEnvInfo(data: ArrayBufferView): Nullable<EnvironmentTextureInfo> {
+    public static GetEnvInfo(data: ArrayBufferView): Nullable<Required<EnvironmentTextureInfo>> {
```
deltakosh, line 138: Why adding Required?

src/Misc/environmentTextureTools.ts
```diff
@@ -218,6 +249,7 @@ export class EnvironmentTextureTools {
         let info: EnvironmentTextureInfo = {
             version: 1,
```
sebavan, line 250: Should we bump the version ? like 1.1 as it is not a breaking change ?

src/Misc/tools.ts
```diff
@@ -725,7 +727,7 @@ export class Tools {
         }
         canvas.toBlob(function (blob) {
             successCallback(blob);
-        }, mimeType);
+        }, mimeType, quality);
```
sebavan, line 730: is undefined all good here, or should there be a default value ? I guess it should but not sure what their implementation does and I am afraid it would be like using 0 in this case ?

src/Misc/tools.ts
```diff
@@ -734,10 +736,11 @@ export class Tools {
      * @param mimeType defines the mime type of the result
      * @param fileName defines he filename to download. If present, the result will automatically be downloaded
      * @param canvas canvas to get the data from. If not provided, use the default screenshot canvas
+     * @param quality defines the quality of the result
      */
-    static EncodeScreenshotCanvasData(successCallback?: (data: string) => void, mimeType: string = "image/png", fileName?: string, canvas?: HTMLCanvasElement): void {
+    static EncodeScreenshotCanvasData(successCallback?: (data: string) => void, mimeType: string = "image/png", fileName?: string, canvas?: HTMLCanvasElement, quality?: number): void {
         if (successCallback) {
-            var base64Image = (canvas ?? Tools._ScreenshotCanvas).toDataURL(mimeType);
+            var base64Image = (canvas ?? Tools._ScreenshotCanvas).toDataURL(mimeType, quality);
```
sebavan, line 743: Same remark for the both following lines.

sebavan (changes requested):
Would be great if you could fix the changes in the what s new and I ll then merge it all :-)

src/Misc/environmentTextureTools.ts
simonihmig: The type for `imageType` is optional [here](https://github.com/BabylonJS/Babylon.js/pull/11102/files#diff-0f4bb579a87a4d0c2f5f84ae1e50a6e14e275e51855ed3b3ad8ac3ab31937a7fR52), as any existing .env file will not have this set, so we should reflect that in the type. But the return value of `GetEnvInfo` is guaranteed to have it, as we are normalizing it by applying the default if not present [here](https://github.com/BabylonJS/Babylon.js/pull/11102/files#diff-0f4bb579a87a4d0c2f5f84ae1e50a6e14e275e51855ed3b3ad8ac3ab31937a7fR164-R166). So using `Required` was the easiest way to turn `imageType?: string` into `imageType: string`. Does that make sense?

src/Misc/environmentTextureTools.ts
simonihmig: Yeah, my thinking was we can keep it as is, as nothing has been removed from the JSON data. But after thinking about this again, I am beginning to think that actually it's not only about the json, but also about what kind of data is in these array buffers. 

Like when there is an .env files created with the changes here, using WebP, any application that still uses old BJS code without these changes will break reading it, as it will wrongly assume the array buffers contains png data, which they don't. And this will probably break in unexpected ways, basically with corrupted image data, instead of giving a meaningful error message.

So maybe we should bump that indeed to 2! What do you guys think?

dist/preview release/what's new.md
simonihmig: This is weird. The diff in my IDE looked all good. Will check...

src/Misc/tools.ts
simonihmig: [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) state `quality` is optional. So I think `canvas.toBlob(cb, type, undefined)` is equivalent to `canvas.toBlob(cb, type)`, right?

I wanted to preserve the defaults the browser chooses, instead of applying our own explicit defaults. And the default is *not* 0, at least when considering the MDN docs linked above!

src/Misc/environmentTextureTools.ts
deltakosh: makes sense thanks!

Popov72 (approved):
LGTM!

src/Misc/tools.ts
```diff
@@ -674,9 +675,9 @@ export class Tools {
                         }
                     };
                     fileReader.readAsArrayBuffer(blob!);
-                });
+                }, mimeType, quality);
```
simonihmig, line 678: Btw, actually there was a bug here, as the `mimeType` argument to `DumpData` was not passed to `toBlob` here! (it was passed in the else-block to `EncodeScreenshotCanvasData` though)

src/Misc/environmentTextureTools.ts
sebavan: Yup sounds like a good plan to me :-)

src/Misc/tools.ts
sebavan: Yup exactly my thoughts, I am just always scared of an internal implementation which relies on arguments.length vs grabbing default when undefined. Keep it like this it should be all good.

dist/preview release/what's new.md
simonihmig: Seems my IDE changed line endings to CRLF, no idea why, which made every line change. Fixed this now!

src/Misc/environmentTextureTools.ts
simonihmig: Did that now (bump version to 2)!

Changed the types a bit, `EnvironmentTextureInfo` is now a [discriminated union](https://basarat.gitbook.io/typescript/type-system/discriminated-unions) of the now separate v1 and v2 interfaces. So by checking the version, TS will narrow the type to the appropriate version. 

Added a normalize method, which takes any version and always returns the latest, so we don't have to care about differences between these versions in other parts of the codebase. This also removed the need for this `Required<>` type, as we now always return a `EnvironmentTextureInfoV2`, which is guaranteed to have `imageType` set. /cc @deltakosh 

Tested again locally, so I think this should be fine. But please re-review the latest changes! :)

src/Misc/environmentTextureTools.ts
```diff
@@ -616,9 +694,7 @@ export class EnvironmentTextureTools {
      * @param info defines the environment texture info retrieved through the GetEnvInfo method
      */
     public static UploadEnvSpherical(texture: InternalTexture, info: EnvironmentTextureInfo): void {
-        if (info.version !== 1) {
-            Logger.Warn('Unsupported babylon environment map version "' + info.version + '"');
-        }
+        info = EnvironmentTextureTools.normalizeEnvInfo(info);
```
simonihmig, line 697: Note that this *could* be seen as some kind of breaking change, as an unsupported version will now throw, while previously it only issued a warning. But I believe this was wrong, as the schema could have been totally different. At least it was inconsistent with the other places where an error was already thrown...

sebavan (approved):

sebavan:
Thanks a ton !!! I ll merge it right after the build !!!

sebavan:
@simonihmig were you able to validate we are not using values when alpha is high ? just to be sure :-)

simonihmig:
> were you able to validate we are not using values when alpha is high ?

Not sure I understand what you mean here? What exactly (and how) should I validate?

sebavan:
Basically I am wondering if for compression sake they are premultiplying alpha with the rest of the channels so when we would have high intensity in some places, we could lost the color info. like white + alpha of 0.1 would result in webp of 0.1, 0.1, 0.1, 0.1

This happens in IE with pngs for instance so I bet it is not the case but it is more of a sanity check

simonihmig:
I regularly contribute to other OSS projects for quite a while, but having a PR reviewed, re-reviewed and merged(!) by three maintainers within literally just a few hours, was a quite unique and enjoyable experience! Thanks folks! 😍

PR #11105, opened 2021-09-17
NME: Fix defines not refreshed when material rebuilt

Popov72:
See https://forum.babylonjs.com/t/nme-is-not-updating-linked-pg/23884/8

### DIFF
--- a/src/Materials/Node/Blocks/Dual/textureBlock.ts
+++ b/src/Materials/Node/Blocks/Dual/textureBlock.ts
@@ -274,36 +274,40 @@ export class TextureBlock extends NodeMaterialBlock {
         if (!defines._areTexturesDirty) {
             return;
         }
-
-        defines.setValue(this._mainUVDefineName, false);
+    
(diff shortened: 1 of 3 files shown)


src/Materials/Node/Blocks/Dual/textureBlock.ts
```diff
@@ -284,22 +286,24 @@ export class TextureBlock extends NodeMaterialBlock {
         }
 
         if (!this.texture || !this.texture.getTextureMatrix) {
-            defines.setValue(this._defineName, false);
-            defines.setValue(this._mainUVDefineName, true);
+            if (this._isMixed) {
```
deltakosh, line 289: Can you elaborate on why we need this test?

Popov72: Because else we can get a `#define undefined` entry in the list of defines because when `_isMixed == false` we don't execute `_injectVertexCode` which is where `_defineName` and `_mainUVDefineName` are set.

PR #11109, opened 2021-09-18
Properly invert ImageBitmaps

nmrugg:
This is primarily aimed at fixing issues with loading .babylon assets when using an OffscreenCanvas.

The issue is that according to the WebGL spec section [6.10](https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.10), `pixelStorei()` ignores `UNPACK_FLIP_Y_WEBGL` (and other parameters) for ImageBitmaps. Instead, `{imageOrientation: "flipY"}` should be passed to `createImageBitmap()`.

See https://forum.babylonjs.com/t/offscreencanvas-textures-are-inverted/23798

Unfortunately, both Firefox (92) and Safari (15) do not currently support OffscreenCanvas or passing options to createImageBitmap. Therefore, these changes are not possible to test on those platforms. So, I didn't add any tests to go along with this PR.

Here are some example playgrounds that should be fixed by this PR:
https://playground.babylonjs.com/#ZG8XE3#3
https://playground.babylonjs.com/#BES426#3
https://playground.babylonjs.com/#NMPB99#5


sebavan (changes requested):
~~could we instead pass only the invertY flag into the loadImage function ?

This would prevent the creation of the option in the general load function which creates garbage collection in case we do not need it. Also it abstracts a bit lower the underlying API we are using ?~~

Actually scratch that as there might be more options to pass in.

src/Engines/thinEngine.ts
```diff
@@ -3556,16 +3556,19 @@ export class ThinEngine {
 
                 prepareTexture(texture, extension, scene, img, texture.invertY, noMipmap, false, prepareTextureProcessFunction, samplingMode);
             };
+            // According to the WebGL spec section 6.10, ImageBitmaps must be inverted on creation.
+            // So, we pass imageOrientation to _FileToolsLoadImage() as it may create an ImageBitmap.
+            let imageBitmapOptions: ImageBitmapOptions = { imageOrientation: texture.invertY ? "flipY" : "none" };
```
sebavan, line 3561: Could we move the option creation near the _FileToolsLoadImage to prevent unnecessary object creation then GC when not needed ?

src/Misc/fileTools.ts
```diff
@@ -164,7 +165,7 @@ export class FileTools {
 
         if (typeof Image === "undefined" || (engine?._features.forceBitmapOverHTMLImageElement ?? false)) {
             FileTools.LoadFile(url, (data) => {
-                engine!.createImageBitmap(new Blob([data], { type: mimeType }), { premultiplyAlpha: "none" }).then((imgBmp) => {
+                engine!.createImageBitmap(new Blob([data], { type: mimeType }), { premultiplyAlpha: "none", ...imageBitmapOptions }).then((imgBmp) => {
```
sebavan, line 168: this would crash if imageBitmapOptions is undefined.

could you replace `...imageBitmapOptions` by `...(imageBitmapOptions || {})`

sebavan (changes requested):
Just 2 tiny changes and this is going in :-)

Thanks a lot !!!

nmrugg:
> Actually scratch that as there might be more options to pass in.

@sebavan,  I had the exact same thought processes.

> Could we move the option creation near the _FileToolsLoadImage to prevent unnecessary object creation then GC when not needed ?

I thought about that too. My initial conclusion was that this function is not called often enough to be a problem. But, I realized now that we can avoid unnecessarily creating objects when we don't need to invert the y. What do you think about it now?

I could go a step further and check if `Image` is `undefined` or `engine._features.forceBitmapOverHTMLImageElement` is `true` to see whether or not we are going to create an ImageBitmap or not. But that complicates the code more, so I'm not sure if that would be better or worse.

> this would crash if imageBitmapOptions is undefined.

I wondered about that too, so I looked it up. Actually, the spread operator allows for the source variable to be `null` or `undefined`. See <a href=https://tc39.es/proposal-object-rest-spread/#AbstractOperations-CopyDataProperties>3.1.3 in the spec</a>. It also <a href=https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md>lists examples of this usage</a>. In the compiled JS source code, the spread opperator is replaced by `Object.assign()` which has the same handling of `null` and `undefined`. See <a href=https://262.ecma-international.org/6.0/#sec-object.assign>19.1.2.1.5.a in the ECMA spec</a>.
So, there's nothing dangerous with this syntax; however, if you insist, I can certainly make the change.

sebavan:
This is all perfect then !!!

sebavan (approved):

nmrugg:
Glad you like it!

PR #11114, opened 2021-09-20
NME custom blocks second draft

BlakeOne:
Here’s the second draft of implementing NME custom blocks, based on user-defined json files.

Forum: https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961

### DIFF
--- a/nodeEditor/src/components/nodeList/nodeListComponent.tsx
+++ b/nodeEditor/src/components/nodeList/nodeListComponent.tsx
@@ -162,6 +162,7 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
     };
     
     private _customFrameList: {[key: string]: string};
+    private _customBlockList: {[key: string]: string};
 
     constructor(props: INodeListComponentProps) {
         super(props);
@@ -171,6 +172,11 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
         let frameJson = localStorage.getItem("Custom-Frame-List");
         if(frameJson) {
             this._customFrameList = JSON.parse(frameJson);
+        }
+
+        let blockJson = localStorage.getItem("Custom-Block-List");
+        if(blockJson) {
+            this._customBlockList = JSON.parse(blockJson);
         }
 
         this._onResetRequiredObserver = this.props.globalState.onResetRequiredObservable.add(() => {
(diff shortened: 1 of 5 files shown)


Popov72:
Seems nice to me!

However, I wonder if we should not do what [Unity does](https://docs.unity3d.com/Packages/com.unity.shadergraph@6.7/manual/Custom-Function-Node.html): define a function that wraps the glsl code. That way, there's no need for a special syntax to use the inputs/outputs (like `{XXX}` in your PR) and we can also use a function from an existing file (see Unity doc page), and that function will be able to use other functions defined in the file too.

Let's wait for @deltakosh input about this.

BlakeOne:
I think the issue is that the input uniform names are based on the input block's name, which the user of the block can change at runtime. If the uniform names instead were based on input.name, then the block creator wouldn't need the escape sequences, which are needed to replace the input name with the input block's runtime name (for example if one of the inputs is named "color" and another input named "color2", then the escape characters are needed to distinguish them when doing the replacements/mapping, I think).

BlakeOne:
It wont let me add it here, so I added the material I've been using for testing to the forum post below. It just multiplies two color blocks together and visualizes the output for comparing against the original, built in multiply block. 

The multiply custom block's json file is in the first post of the thread. Let me know, if you want I can make a few more simple custom blocks for more thorough testing of the features that are implemented so far...

https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961/2

BlakeOne:
Hmm, I just noticed that automated checks failed when building. It says: Cannot find module 'babylonjs/Materials'

The 5 files that I changed/created are working when I test locally, but I'm using "npm run start", not building it from scratch. I can't figure out what I did wrong that would cause that error on the build?

deltakosh:
Good question for @RaananW

RaananW:
Those two steps don't take the built version of the declaration files, but the already-existing module declaration (.d.ts). We usually don't do that, but - will you be able to build and submit the updated d.ts file (babylon.modules.d.ts from the dist/preview forlder) to see if the CI reacts correctly? there might be a need to submit the built source as well, but for now let's go with the .d.ts file and see if it doesn't fail

BlakeOne:
Here's a link to the repo I made, that just has copies of all the files that I created or changed. I think that's what you were asking for on the forum @delkatosh?

https://github.com/BlakeOne/NMECustomBlockFiles

BlakeOne:
Okay, I looked in the .t.ds file in /preview release/dist and found the error in the file. The return type for class CustomBlock's method findInputByName is this:

 import("babylonjs/Materials").NodeMaterialConnectionPoint | null;


Instead it should be like this I think, like the other functions:

NodeMaterialConnectionPoint | null;

IDK why it's being built that way thou, I don't explicitly provide the return type.

Popov72:
If we ask the user to write a function to wrap their glsl code, I think we don't have a problem with the names of the inputs/outputs.

For eg, the multiply could be written as:
```glsl
void myMultiplication(left: vec4, right: vec4, out result: vec4) {
    result = left * right;
}
```
Then, the `CustomBlock` class would emit something like this in the `_buildBlock` method (pseudo code):
```typescript
state.compilationString += <declare the output variables>;
state.compilationString += this._options.functionName + "(" + <here pass the inputs then the outputs separated by a comma> + ")";
```
We don't even have to require that the names used by the user to declare the inputs/outputs of their function match the names of the input/output ports, we just have to require they are declared in the same order than the ports.

If the function can work with a non static type (as the multiplication you wrote initially, which can work with float, vec2, etc), we can use some placeholder for the type of the input/output variables that will be replaced at runtime by `CustomBlock` when emitting the code.

BlakeOne:
Aha, I think I just need to import NodeMaterialConnectionPoint in the file customBlock.ts so that it has the return type for the findInputByName function. Welp, I'm still new to typescript and npm, and big projects in general. I'll submit another patch with the fixed file:)

BlakeOne:
Hmm, when I add this line to customBlock.ts, VSCode warns that NodeMaterialConnectionPoint is declared but never used, so maybe that's not the fix after all?

import {NodeMaterialConnectionPoint} from '../nodeMaterialBlockConnectionPoint';

deltakosh:
> ```ts
> <declare the output variables>
> ```

I like that idea a lot!!!

PR #11116, opened 2021-09-20
Improve Sound.setPosition performance

jure:
On a publicly deployed project I'm working on we've had numerous users report severe audio degradation over time on iOS and Safari. The project uses spatial audio. 

On both platforms the CPU usage was ticking upwards quickly. Digging in it turns out that setting a position of the `PannerNode` on every frame even when the position has not changed, as can happen via e.g. world matrix update observables, has adverse performance effects on WebKit.

Checking and returning early if the current and new positions are equal seems to have remedied this performance issue.

Happy to discuss further!

### DIFF
--- a/src/Audio/sound.ts
+++ b/src/Audio/sound.ts
@@ -625,6 +625,9 @@ export class Sound {
      * @param newPosition Defines the new position
      */
     public setPosition(newPosition: Vector3): void {
+        if (newPosition.equals(this._position)) {
+            return;
+        }
         this._position = newPosition;
 
         if (Engine.audioEngine?.canUseWebAudio && this.spatialSound && this._soundPanner && !isNaN(this._position.x) && !isNaN(this._position.y) && !isNaN(this._position.z)) {


src/Audio/sound.ts
```diff
@@ -625,6 +625,9 @@ export class Sound {
      * @param newPosition Defines the new position
      */
     public setPosition(newPosition: Vector3): void {
+        if(newPosition.equals(this._position)) {
```
Popov72, line 628: nit: space missing between `if` and `(`

Popov72 (approved):
LGTM, only a small linting issue!

jure:
Thank you for the 👍  @Popov72, I fixed the issue and then squashed the two commits.

sebavan:
Nice one @jure

jure:
Thanks for the merge, @sebavan! It should be said that this is likely a bandaid, and that the real culprit is hiding in WebKit: https://bugs.webkit.org/show_bug.cgi?id=227199

The bug comments include a reproduction, which results in crackling degraded audio on both iOS and Safari: https://brianpeiris.github.io/spatial-audio-test/

But at least if your positional audio nodes are not moving in any sort of animated frame by frame fashion, this early return will save you some time before the audio degradation occurs.

PR #11117, opened 2021-09-20
[XR] LightEstimation feature

RaananW:
Closing #8904


src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+                this._reflectionCubeMap._texture.dispose();
+            }
+            const internalTexture = new InternalTexture(this._xrSessionManager.scene.getEngine(), InternalTextureSource.Unknown);
+            internalTexture.isCube = true;
+            internalTexture.invertY = false;
+            internalTexture._useSRGBBuffer = this.options.reflectionFormat === 'srgba8';
+            internalTexture.format = Constants.TEXTUREFORMAT_RGBA;
+            internalTexture.generateMipMaps = false;
+            internalTexture.type = this.options.reflectionFormat !== 'srgba8' ? Constants.TEXTURETYPE_HALF_FLOAT : Constants.TEXTURETYPE_UNSIGNED_BYTE;
+            internalTexture.samplingMode = Constants.TEXTURE_NEAREST_SAMPLINGMODE;
```
sebavan, line 213: I am not sure we should use NEAREST here, actually an env map should have mip maps and filtering turned on

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+            internalTexture.width = this.ReflectionCubeMapTextureSize;
+            internalTexture.height = this.ReflectionCubeMapTextureSize;
+            internalTexture._cachedWrapU = Constants.TEXTURE_WRAP_ADDRESSMODE;
+            internalTexture._cachedWrapV = Constants.TEXTURE_WRAP_ADDRESSMODE;
+            internalTexture._hardwareTexture = new WebGLHardwareTexture(lp, this._getCanvasContext() as WebGLRenderingContext);
+            this._reflectionCubeMap._texture = internalTexture;
+            this._reflectionCubeMap._texture.isReady = true;
+            this._xrSessionManager.scene.markAllMaterialsAsDirty(Constants.MATERIAL_TextureDirtyFlag);
+            this._reflectionCubeMap.hasAlpha = true;
```
sebavan, line 222: Is it really having alpha ? wondering what it would represent

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+                    this._xrLightEstimate.primaryLightIntensity.z / this._intensity
+                );
+
+                // direction from instead of direction to
+                this._lightDirection.negateInPlace();
+                // set the values after calculating them
+                if (this.directionalLight) {
+                    this.directionalLight.direction.copyFrom(this._lightDirection);
+                    this.directionalLight.intensity = Math.min(this._intensity, 1.0);
+                    this.directionalLight.diffuse.copyFrom(this._lightColor);
```
sebavan, line 350: I am wondering if they provide a unit or decay we should follow so we could force the correct light mode as default will work differently in standard and PBR ?

sebavan (commented):
Loooks pretty coool to me :-)

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+                return;
+            }
+            this._cubeMapPollTime = now;
+        }
+        const lp = this._getXRGLBinding().getReflectionCubeMap(this._xrLightProbe);
+        if (lp && this._reflectionCubeMap) {
+            if (this._reflectionCubeMap._texture) {
+                this._reflectionCubeMap._texture.dispose();
+            }
+            const internalTexture = new InternalTexture(this._xrSessionManager.scene.getEngine(), InternalTextureSource.Unknown);
```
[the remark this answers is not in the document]
RaananW: yes! i wanted to ask you why it doesn't work.

I tried creating a new hardware texture, tried setting a new webgl texture in the existing hardware texture, tried releasing, tried setting all materials dirty. nothing worked other than this

src/XR/features/WebXRLightEstimation.ts
RaananW: no, they don't. they provide the following:

```
  readonly attribute Float32Array sphericalHarmonicsCoefficients;
  readonly attribute DOMPointReadOnly primaryLightDirection;
  readonly attribute DOMPointReadOnly primaryLightIntensity;
```
intensity represents RGB value.
The spherical harmonics is not being used here.

This light is optional. a dev could take the values and use them in a different way

src/XR/features/WebXRLightEstimation.ts
RaananW: this is a 16x16 texture generated by webxr itself. i didn't see the nead for mipmaps, but i can enable it.
Regarding sampling - i can remove it, of course.

src/XR/features/WebXRLightEstimation.ts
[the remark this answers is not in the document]
sebavan: You might need to invalidate the engine texture cache :-)

src/XR/features/WebXRLightEstimation.ts
sebavan: So lets force the mode to gltf which is probably the most used one :-) also sphericalHarmonicsCoefficients is AMAZING !!! as you could populate the ones in the environment texture with them !!!

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+    private _getXRGLBinding(): XRWebGLBinding {
+        if (this._xrWebGLBinding === null) {
+            let context = this._getCanvasContext();
+            this._xrWebGLBinding = new XRWebGLBinding(this._xrSessionManager.session, context);
+        }
+        return this._xrWebGLBinding;
+    }
+
+    /**
+     * Event Listener to for "reflectionchange" events.
```
Popov72, line 187: nit: remove "to"

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+     */
+    private _updateReflectionCubeMap = (): void => {
```
Popov72, line 189: nit: Why not using `private _updateReflectionCubeMap(): void {` as for all other methods of the class?

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+            if (now - this._cubeMapPollTime < this.options.cubeMapPollInterval) {
+                return;
+            }
+            this._cubeMapPollTime = now;
+        }
+        const lp = this._getXRGLBinding().getReflectionCubeMap(this._xrLightProbe);
+        if (lp && this._reflectionCubeMap) {
+            if (this._reflectionCubeMap._texture) {
+                this._reflectionCubeMap._texture.dispose();
+            }
```
Popov72, line 205: nit: you can save two lines by doing `this._reflectionCubeMap._texture?.dispose();` :)

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+     * detach this feature.
+     * Will usually be called by the features manager
+     *
+     * @returns true if successful.
+     */
+    public detach(): boolean {
+        const detached = super.detach();
+
+        if (this._xrLightProbe !== null) {
+            this._xrLightProbe.removeEventListener('reflectionchange', this._updateReflectionCubeMap);
```
Popov72, line 271: Shouldn't you test also for `!this.options.disableCubeMapReflection` before removing the event listener, as the attach code is testing for this condition before adding the event listener?

Popov72 (approved):
LGTM, only nits!

src/XR/features/WebXRLightEstimation.ts
```diff
@@ -0,0 +1,366 @@
...
+     * This number does not correspond to the WebXR specs version
+     */
+    public static readonly Version = 1;
+
+    /**
+     * ARCore's reflection cube map size is 16x16.
+     * Once other systems support this feature we will need to change this to be dynamic.
+     * see https://github.com/immersive-web/lighting-estimation/blob/main/lighting-estimation-explainer.md#cube-map-open-questions
+     */
+    private ReflectionCubeMapTextureSize: number = 16;
```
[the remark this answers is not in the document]
RaananW: originally it was public. But you are right, i'll add the _.

src/XR/features/WebXRLightEstimation.ts
RaananW: Good question! :-)
This is an event listener. to avoid binding this we force this to be the context using an arrow function

src/XR/features/WebXRLightEstimation.ts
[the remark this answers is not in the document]
RaananW: done, working.

src/XR/features/WebXRLightEstimation.ts
RaananW: Sure! removeEventListener is a no-op if it doesn't exist, but there is really no need to do that here, you are totally right.

src/XR/features/WebXRLightEstimation.ts
RaananW: Done :-)

sebavan (approved):
GG !!!!

PR #11118, opened 2021-09-20
NME Custom Blocks

BlakeOne:
Here’s the new patch for NME custom blocks, with the file customBlock.ts fixed to explicitly provide function findInputByName’s return type.

Forum: https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961

Link to the json test files (and src files): https://github.com/BlakeOne/NMECustomBlockFiles

### DIFF
--- a/nodeEditor/src/components/nodeList/nodeListComponent.tsx
+++ b/nodeEditor/src/components/nodeList/nodeListComponent.tsx
@@ -162,6 +162,7 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
     };
     
     private _customFrameList: {[key: string]: string};
+    private _customBlockList: {[key: string]: string};
 
     constructor(props: INodeListComponentProps) {
         super(props);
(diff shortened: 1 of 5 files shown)


BlakeOne:
Forum: https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961

Link to test files (and src files): https://github.com/BlakeOne/NMECustomBlockFiles

Popov72:
You should also read the Unity doc page I have linked in the other PR, I think it can lighten things a bit.

BlakeOne:
@deltakosh, @Popov72, it looks like the only keeping it from validating now is the missing "What's New Update", should I update that file to say that NME custom blocks were added and redo the patch?

deltakosh:
Please add it to the file..But I think we should merge it when completely done right? (meaning including the great idea from @Popov72 )

BlakeOne:
> Please add it to the file..But I think we should merge it when completely done right? (meaning including the great idea from @Popov72 )

Okay, I'll wait to make the patch until customBlock.ts is more finalized as well then.

And I just posted on the forum about implementing @Popov72's idea, to try to get a better handle on it before I start coding and testing it out:)

https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961/6

PR #11125, opened 2021-09-21
WebGPU: Add Nagasm as a transpiler module

Popov72:
There are still some problems with Nagasm (14 validation tests fail over 31 in the WebGPU test suite) but having the code integrated in the engine will simplify testing when new updates of the module are available.

TintWASM is still the default transpiler module.


sebavan (commented):
Let s discuss it back on Thursday with the Google folks and Mozilla one before merging ?

Popov72:
Discontinued for the time being.

PR #11129, opened 2021-09-21
recastjs with embedded WASM

CedricGuillemet:
follow up https://forum.babylonjs.com/t/significant-recast-js-performance-degrade-from-cdn-version-to-preview/23859/6

Latest emscripten produces poorly optimized JS...like really slow, way slower than non optimized code. I checked everything I could to see if it was because of an undefined behavior with LLVM or something else (C++ versions, checked different revisions of subrepo, emcc flag combination, ...) but no luck. JS is always very slow.

To not break everything, this PR provides a new recast.js that embeds WASM. Totally transparent for user, smaller file size and faster than previous JS version.


deltakosh (approved):

RaananW:
For future reference - recast should also have a modules version for es6-only users. The final few lines should be removed and should be replaced with

```javascript 
export { Recast };
export default Recast;
```

The recast package should also reference this module file in the `module` key (`main` should stay the same)

PR #11131, opened 2021-09-21
NME Custom Blocks

BlakeOne:
NME custom blocks incorporating @Popov72's updated CustomBlock class

https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961/14

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -118,6 +118,7 @@
 
 ### NME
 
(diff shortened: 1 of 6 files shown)


BlakeOne:
@Popov72 it looks your change of the import for NodeMaterialConnectionPoint to use ".."  caused the validation error, I'm going to change it back to importing from '../nodeMaterialBlockConnectionPoint', which validated yesterday, and resubmit the patch.

PR #11136, opened 2021-09-21
NME Custom Blocks

BlakeOne:
NME custom blocks incorporating @Popov72's updated CustomBlock class
Example custom block (JSON): https://github.com/BlakeOne/NMECustomBlockFiles/blob/main/multiplyBlock.json
Forum: https://forum.babylonjs.com/t/nme-custom-blocks-take-two/23961/16


nodeEditor/src/sharedComponents/draggableLineWithButtonComponent.tsx
```diff
@@ -21,7 +21,7 @@ export class DraggableLineWithButtonComponent extends React.Component<IDraggable
                 onDragStart={event => {
                     event.dataTransfer.setData("babylonjs-material-node", this.props.data);
                 }}>
-                {this.props.data.substr(0, this.props.data.length - 6)}
+                {this.props.data.substr(0, this.props.data.length - (this.props.data.indexOf("CustomBlock") > -1? 11 : 6))}
```
Popov72, line 24: I find it a bit hard-coded...

Maybe it's possible to pass some information (the number to subtract from `this.props.data.length`) through the `IDraggableLineWithButtonComponent` interface?

Popov72 (changes requested):
LGTM, only a small comment!

nodeEditor/src/sharedComponents/draggableLineWithButtonComponent.tsx
BlakeOne: Agreed, I thought about doing that initially but decided to prioritize changing as little as possible, so it’s still hard coded like it was when I got there.

TBH though, after the glsl changed from “{output} = {left} * {right};” to be much harder to read and right, with {} escapes all over the place (which was the issue with the original glsl that has far fewer {}’s), I’ve actually lost interest/motivation in this (which is fine if you’ve finished implementing it how you want). 

So I think it would make more sense if you want to make more changes, to merge and then make a patch to change that or just make a new patch based on the existing patch and close this one. Either way sounds good to me.

sebavan: You can simply do the changes and push back to your branch, no need to close and reopen a new PR, this will get updated.

Popov72: > TBH though, after the glsl changed from “{output} = {left} * {right};” to be much harder to read and right

I don't understand your comment...

There are some {} only around the **TYPE_XXX** names and because we allow to plug any type of input to this block. So, those appear only at the variable declarations. The glsl code inside the function does not use these markers. And for functions that take fixed types of variables (which are most the functions anyway - only a few functions allow for dynamic types of variables), you won't have any `{...}` syntax but the exact same code you would have in a `ShaderMaterial`, for eg.

sebavan:
@BlakeOne, any update on this one, or do you need some help with the feedbacks ?

sebavan:
Looks like @BlakeOne is not available @Popov72, would you mind giving it a shot ?

Popov72:
Yes, will do it as soon as possible.

PR #11141, opened 2021-09-22
SixDofDragBehavior update for XR devices

rickfromwork:
This PR updates the sixDofDrag behavior to be more intuitive when using XR controllers. I had initially planned on a larger rewrite, but as I worked through it I started to see why each part had to be the way it was, for non-XR at least.

Also in this is an update to plumb XR controller rotations through for endpoints to read from, and a small change in the legacy webVR code cleaning up an unused variable with a conflicting name.


RaananW (commented):
LGTM! There is a linting issue that needs to be resolved. 
The XR vs. 2D functions do seem very similar with slight modifications. is there a way to unify them? or is this where the old implementation failed?

RaananW:
Hi @rickfromwork - any update here?

rickfromwork:
> Hi @rickfromwork - any update here?

I just came back from vacation - Getting this finished and in is my main focus now

RaananW (approved):
LGTM!

PR #11147, opened 2021-09-22
Export .env environment textures using WebP from Inspector

simonihmig:
Following up on #11102.

### DIFF
--- a/inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
+++ b/inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
@@ -33,7 +33,12 @@ import { IndentedTextLineComponent } from "../../../sharedUiComponents/lines/ind
 import { TextInputLineComponent } from "../../../sharedUiComponents/lines/textInputLineComponent";
 import { LockObject } from "../../../sharedUiComponents/tabs/propertyGrids/lockObject";
 
-const GIF = require("gif.js.optimized")
+const GIF = require("gif.js.optimized");
+
+const envExportImageTypes = [
+    { label: "PNG", value: 0, imageType: "image/png" },
+    { label: "WebP", value: 1, imageType: "image/webp" },
+];
 
 export class ToolsTabComponent extends PaneComponent {
     private _lockObject = new LockObject();
@@ -49,6 +54,7 @@ export class ToolsTabComponent extends PaneComponent {
     private _reflectorHostname: string = "localhost";
     private _reflectorPort: number = 1234;
     private _reflector: Reflector;
+    private _envOptions = { imageType: 0, imageQuality: 0.8 };
 
     constructor(props: IPaneComponentProps) {
         super(props);
@@ -251,7 +257,12 @@ export class ToolsTabComponent extends PaneComponent {
 
     createEnvTexture() {
         const scene = this.props.scene;
-        EnvironmentTextureTools.CreateEnvTextureAsync(scene.environmentTexture as CubeTexture)
+        EnvironmentTextureTools.CreateEnvTextureAsync(
+            scene.environmentTexture as CubeTexture,
+            {
+                imageType: envExportImageTypes[this._envOptions.imageType].imageType,
+                imageQuality: this._envOptions.imageQuality
+            })
             .then((buffer: ArrayBuffer) => {
                 var blob = new Blob([buffer], { type: "octet/stream" });
                 Tools.Download(blob, "environment.env");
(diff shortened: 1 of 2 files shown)


inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
```diff
@@ -382,7 +393,16 @@ export class ToolsTabComponent extends PaneComponent {
                             <ButtonLineComponent label="Export to Babylon" onClick={() => this.exportBabylon()} />
                             {
                                 !scene.getEngine().premultipliedAlpha && scene.environmentTexture && scene.environmentTexture._prefiltered && scene.activeCamera &&
```
simonihmig, line 395: I didn't want to touch this line (yet), but wonder why this `scene.environmentTexture._prefiltered` is needed?

With this still in place, I can e.g. drop a gltf that already has an .env environment into the sandbox, and the "Generate .env texture" show up. Without that check, I can easily export the .env without problems. In this case I think it does make sense, as it allows you to convert an existing PNG-based .env to the new WebP format!

TL;dr I'd like to change this to:

```suggestion
                                !scene.getEngine().premultipliedAlpha && scene.environmentTexture && scene.activeCamera &&
```

Any reason not to do so?

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
```diff
@@ -49,6 +54,7 @@ export class ToolsTabComponent extends PaneComponent {
     private _reflectorHostname: string = "localhost";
     private _reflectorPort: number = 1234;
     private _reflector: Reflector;
+    private _envOptions = { imageType: 0, imageQuality: 0.8 };
```
sebavan, line 57: I guess imageType should be image/png by default :-)

sebavan: Also is imageQuality 0.8 by default on png ? I thought the quality was depending on the mime type for the default value ?

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
```diff
@@ -251,7 +257,12 @@ export class ToolsTabComponent extends PaneComponent {
 
     createEnvTexture() {
         const scene = this.props.scene;
-        EnvironmentTextureTools.CreateEnvTextureAsync(scene.environmentTexture as CubeTexture)
+        EnvironmentTextureTools.CreateEnvTextureAsync(
+            scene.environmentTexture as CubeTexture,
+            {
+                imageType: envExportImageTypes[this._envOptions.imageType].imageType,
+                imageQuality: this._envOptions.imageQuality
```
sebavan, line 264: Related to the first comment, the quality could be part of the envExportImageTypes array ?

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
sebavan: We should not be able to generate a texture from a none prefiltered cubemap, basically if people are relying on a lightprobe and such and use real time filtering, it is not smthg we should convert in an env file.

sebavan (commented):
Only a couple of nits :-) Thanks a lot

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
simonihmig: > I guess imageType should be image/png by default :-)

It is. I used a numeric value for `imageType`, as it seems `<OptionsLineComponent>` can only deal with numbers as value, right? So this numbers refers to the index of ´envExportImageTypes´, and ´envExportImageTypes[0]´ is the entry for PNG!

> Also is imageQuality 0.8 by default on png

I believe PNG will ignore the quality all together, as it only supports lossless compression.

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
simonihmig: I don't understand this... `envExportImageTypes` is a const/immutable array defining the options of `<OptionsLineComponent>`, and `imageQuality` is set by the user. So how should those be combined, or what exactly are you suggesting?

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
sebavan: I see :-) it is just naming then, could you rename imageType in _envOptions to imageTypeIndex or something similar to prevent any confusion ? 

Basically looking into it too quickly I tought it was used as the string here.

inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
sebavan: Yup all good as pngs do not rely on it, and also coming from my previous confusion between the index and the type.

sebavan (changes requested):
Basically mostly only one rename to go.

sebavan:
@simonihmig I did the rename after the merge nw and thanks again for the contrib !!!!

PR #11148, opened 2021-09-22
Fix texture caching in native engine when using srgb buffer

bghgary:


Popov72 (approved):

sebavan (approved):

bghgary:
Will be incorporated into https://github.com/BabylonJS/Babylon.js/pull/11149

PR #11156, opened 2021-09-24
[GE] Various Bug Fixes + Resizing Grid

msDestiny14:
-Fix for textinput not taking in negative numbers
-Grid editing mistake
-Property Tab UI bug
-Resize grid button

### DIFF
--- a/sharedUiComponents/lines/textInputLineComponent.tsx
+++ b/sharedUiComponents/lines/textInputLineComponent.tsx
@@ -72,7 +72,7 @@ export class TextInputLineComponent extends React.Component<ITextInputLineCompon
             }
 
             //Removing starting zero if there is a number of a minus after it. 
-            if(value.charAt(0) === '0' && value.length > 1 && (/[^0-9\-]/g.test(value.substr(1)) == false)) {
+            if(value.search(/0+[0-9\-]/g) === 0){
                 value = value.substr(1);
             }
         }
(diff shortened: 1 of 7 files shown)


sharedUiComponents/lines/textInputLineComponent.tsx
```diff
@@ -64,16 +64,15 @@ export class TextInputLineComponent extends React.Component<ITextInputLineCompon
...
-                checkValue = value.substr(1);
-            }
-            if (/[^0-9\.\p\x\%]/g.test(checkValue)) {
+            if (/[^0-9\.\p\x\%\-]/g.test(value)) {
                 return;
             }
             if (!value) {
                 value = "0";
             }
+            if(value.charAt(0) === '0' && value.length > 1 && (/[^0-9\-]/g.test(value.substr(1)) == false)) {
```
RaananW, line 73: what does this line do?
if value starts with 0, and is longer than 1, and the characters after poision 1 are not a number, cut the value to the value after position 1? what is the use case for that?

RaananW (commented):
LGTM, just a small question

sharedUiComponents/lines/textInputLineComponent.tsx
msDestiny14: this if for once a person types backspaces all the way it will be 0. and then the next character they type is a number like 4? so change 04 to just 4. Would a comment be good here?

sebavan: The test is confusing why not testing for what you want instead of a regex to not find it and == false ?

PR #11160, opened 2021-09-24
Remove unnecessary bind pose from bones

bghgary:


sebavan (approved):

src/Bones/bone.ts
```diff
@@ -71,8 +70,9 @@ export class Bone extends Node {
 
     /** @hidden */
     set _matrix(value: Matrix) {
```
bghgary, line 72: @deltakosh Is there a reason why this method didn't mark as dirty before?

src/Bones/bone.ts
```diff
@@ -71,8 +70,9 @@ export class Bone extends Node {
 
     /** @hidden */
     set _matrix(value: Matrix) {
+        this._needToCompose = false; // in case there was a pending compose
         this._localMatrix.copyFrom(value);
-        this._needToDecompose = true;
+        this._markAsDirtyAndDecompose();
```
Popov72, line 75: I think we should check that `value.updateFlag` !== `this._localMatrix.updateFlag` to avoid unnecessary decomposition of the matrix when it is the same than the existing one.

The `Skeleton.prepare` method is doing:
```typescript
if (this._numBonesWithLinkedTransformNode > 0) {
    for (const bone of this.bones) {
        if (bone._linkedTransformNode) {
            // Computing the world matrix also computes the local matrix.
            bone._linkedTransformNode.computeWorldMatrix();
            bone._matrix = bone._linkedTransformNode._localMatrix;
            bone.markAsDirty();
        }
    }
}
```
So it is always updating the bone local matrix. If an animation is stopped, for eg, the matrices won't change so we would not need to re-set/re-decompose them. Note that I think we should also remove the `bone.markAsDirty()` call as it is now already done in the `_matrix` setter.

Popov72 (approved):
LGTM!

src/Bones/bone.ts
bghgary: > I think we should check that value.updateFlag !== this._localMatrix.updateFlag to avoid unnecessary decomposition of the matrix when it is the same than the existing one.

If value isn't the same object as this._localMatrix.updateFlag, I'm not sure this check is correct. Is there a way to check if the matrix are the same object?

bghgary: Wait, never mind, the flag is from a static, so it should be unique.

bghgary: Hmm, I tried this, but the check doesn't work for glTF for some reason. Investigating...

bghgary: I fixed it by copying the updateFlag when copying the matrix. This works, but I kind of think the copyFrom function should copy the updateFlag also.

bghgary: Waiting on #11162

sebavan:
I guess you could remove this._localMatrix.updateFlag = value.updateFlag; to not forget :-)

bghgary:
Already did locally.

src/Bones/bone.ts
```diff
@@ -71,8 +70,15 @@ export class Bone extends Node {
 
     /** @hidden */
     set _matrix(value: Matrix) {
+        this._needToCompose = false; // in case there was a pending compose
```
Popov72, line 73: I would have moved this after the check on the `updateFlag`: if the `_needToCompose` flag is `true` when calling the `_matrix` setter it means the current local matrix needs to be composed and we should keep the flag `true` if we early return because the matrices are identical.

Popov72 (approved):
Only a small comment.

src/Bones/bone.ts
bghgary: No, that's not right. If this flag is true, that means someone set the bone transform using TRS. If that is still in progress and someone sets the matrix, then we want to ignore the TRS.

Issue #11100, opened 2021-09-16
Grid and ImageSlider not saving properly


msDestiny14:
Create a grid or an image slider using the gui editor. Save and try loading it in a playground or the editor again.

- [x] ImageSlider is not saving the images
- [x] Grid is not saving row/column info
- [x] Fonts on all controls

sebavan:
Closed by the previous PR

PR #11165, opened 2021-09-25
[GE] Control Audit, adt parseFromURLAsync

msDestiny14:
https://github.com/BabylonJS/ThePirateCove/issues/83
https://github.com/BabylonJS/Babylon.js/issues/11100


guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -66,23 +66,30 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
+        newValue = (negative ? '-' : '') + newValue;
         newValue += percentage ? '%' : 'px';
+
         (this.props.control as any)[propertyName] = newValue;
         this.forceUpdate();
     }
 
     render() {
         const control = this.props.control;
+        console.log(control);
```
sebavan, line 92: Should not be here

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -115,19 +122,32 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
                     <TextInputLineComponent numbersOnly={true} iconLabel={"Scale"} icon={sizeIcon} lockObject={this.props.lockObject} label="W" target={this} propertyName="_width" onPropertyChangedObservable={this.props.onPropertyChangedObservable}
                         onChange={(newValue) => {
-                            this._width = newValue;
+
                             if (control.typeName === "Image") {
                                 (control as Image).autoScale = false;
-                            };
+                            }
+                            else if (this.props.control.typeName === "ColorPicker") {
+                                console.log(newValue);
```
sebavan, line 130: Should not be here

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -115,19 +122,32 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
                         } />
                     <TextInputLineComponent numbersOnly={true} lockObject={this.props.lockObject} label="H" target={this} propertyName="_height" onPropertyChangedObservable={this.props.onPropertyChangedObservable}
                         onChange={(newValue) => {
-                            this._height = newValue;
                             if (control.typeName === "Image") {
                                 (control as Image).autoScale = false;
-                            };
+                            }
+                            else if (this.props.control.typeName === "ColorPicker") {
+                                console.log(newValue);
```
sebavan, line 145: Should be removed

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -66,23 +66,30 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
         if (value.charAt(value.length - 1) === '%') {
             percentage = true;
         }
-        else if(value.charAt(value.length - 1) === 'x' && value.charAt(value.length - 2) === 'p') {
+        else if (value.charAt(value.length - 1) === 'x' && value.charAt(value.length - 2) === 'p') {
             percentage = false;
         }
         let newValue = value.split('').filter(function (item) {
-            return (!isNaN(parseInt(item)));
+            if (item != 'p' && item != 'x' && item != '-' && item != '%') {
```
sebavan, line 77: Looks like there might be a safer  less expensive way ?

sebavan (changes requested):

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -70,15 +70,14 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
         }
-        else if(value.charAt(value.length - 1) === 'x' && value.charAt(value.length - 2) === 'p') {
+        else if (value.charAt(value.length - 1) === 'x' && value.charAt(value.length - 2) === 'p') {
             percentage = false;
         }
-        let newValue = value.split('').filter(function (item) {
-            return (!isNaN(parseInt(item)));
-        }).join('');
 
+        let newValue = value.replace(/[\p\x\%\-]/g, "");
```
RaananW, line 77: though valid, there is no reason breaking any of those characters . They are not special characters.
So this regex could be:

`[px%-]` instead.

RaananW: It also seems like you are trying to extract the number from the string. you can do something like this:

```javascript
"-123.543px".match(/([\d\.\,]+)/)[0] === '123.543';
```

This way you always make sure no other character is parsed.

Another question here - are you taking into account different number representations? for example a german (or french)  browser will represent the number as `123,543` (with a ',' instead of '.'). Is the browser parsing the number, or are you?

RaananW (commented):
Just a small comment about the regex :-)

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
msDestiny14: I'm doing an parse to get rip of P, X , % and - so I can add it back in myself.

RaananW: it is safer in that case to get what you need instead of removing what you don't.

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -70,15 +70,14 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
         }
-        else if(value.charAt(value.length - 1) === 'x' && value.charAt(value.length - 2) === 'p') {
+        else if (value.charAt(value.length - 1) === 'x' && value.charAt(value.length - 2) === 'p') {
             percentage = false;
         }
-        let newValue = value.split('').filter(function (item) {
-            return (!isNaN(parseInt(item)));
-        }).join('');
 
+        let newValue = value.match(/([\d\.\,]+)/g)?.toString();
```
RaananW, line 77: have you tested this?

msDestiny14: ya I've been teasing this. unless I missed something?

RaananW: just not sure about the toString call on an optional variable :-) Or why it is needed? Check if the matches array has a value and use it instead of calling toString on an array. Because:

``` javascript
"123.23px43234".match(/([\d\.\,]+)/g).toString() === '123.23,43234';
// but
"123.23px43234".match(/([\d\.\,]+)/g) === ['123.23', '43234'];
```


If it is optional (i.e. `?` at the end) it means it can be undefined. so newValue will be undefined, which means the next line will fail.

msDestiny14: how's this? :)

RaananW: works ;-)

PR #11178, opened 2021-09-27
Add return to rest button to skeleton inspector

bghgary:


inspector/src/components/actionTabs/tabs/propertyGrids/meshes/skeletonPropertyGridComponent.tsx
```diff
@@ -166,38 +166,38 @@ export class SkeletonPropertyGridComponent extends React.Component<ISkeletonProp
...
+                    }
                     <CheckBoxLineComponent label="Use texture to store matrices" target={skeleton} propertyName="useTextureToStoreBoneMatrices" onPropertyChangedObservable={this.props.onPropertyChangedObservable} />
-                    
-                    <LineContainerComponent title="DEBUG" selection={this.props.globalState}>                        
-                        <CheckBoxLineComponent label="Enabled" isSelected={() => this._skeletonViewersEnabled} onSelect={() => this.switchSkeletonViewers()} />
-                        <OptionsLineComponent label="displayMode" options={debugModeOptions} target={this._skeletonViewerDisplayOptions} propertyName="displayMode" onPropertyChangedObservable={this.props.onPropertyChangedObservable} onSelect={() => this.changeDisplayMode()} />
-                        {displayOptions}                   
-                    </LineContainerComponent>                    
+                    <ButtonLineComponent label="Return to rest" onClick={() => skeleton.returnToRest()} />
+                </LineContainerComponent>
```
sebavan, line 196: looks like it is now 2 lines instead of a line in a line

As I have no idea which one it should be, just raising here in case it was not intended even if I am pretty sure the new way is the right one  :-)

bghgary: Yeah, I noticed this, thought it was weird, and changed it. The display result is barely noticeable.

PR #11179, opened 2021-09-28
Add .js to imports in distES6

RaananW:
This PR adds another step to the gulp build process to add .js to local imports in all libraries that are not built using webpack.

This will allow removing fullySpecified from webpack configuration.
It should not hurt backwards compatibility - tested with rollup and webpack projects locally.


Tools/Gulp/tasks/gulpTasks-librariesES6.js
```diff
@@ -9,6 +9,7 @@ var symlinkDir = require('symlink-dir');
 // Gulp Helpers
 var rmDir = require("../../NodeHelpers/rmDir");
 var processImports = require("../helpers/gulp-processImportsToEs6");
+var convertImportsToJs = require("../helpers/gulp-importFromJsFiles");
```
sebavan, line 12: Could this be merged with the processImports step to prevent going over all the file twice assuming the steps is taking some times (if not, please fully disregard :-)) ?

sebavan (commented):
LGTM, only one quick comment

Tools/Gulp/tasks/gulpTasks-librariesES6.js
RaananW: that was my first attempt. However it will not work without changing the webpack configuration that builds inspector etc', since it changes the sources being used. This is why I wanted to do that on the dist files of projects not built with webpack.

PR #11184, opened 2021-09-29
GUI getControlByName

msDestiny14:
Adding @PirateJC for ux check. https://github.com/BabylonJS/ThePirateCove/issues/93


RaananW (commented):
Array find function does exactly what you want without creating unneeded arrays:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

It is not available in IE11, but as we are slowly removing support for this wonderful browser, I don't see a reason to support this browser with new functionality. Besides, it is already being used:

```javascript
private getControlByKey(key: string, value: string) : Nullable<Control> {
  return this._rootContainer.getDescendants().find(control => control[key] === value) || null;
}
```

msDestiny14:
like this?

msDestiny14:
wait actually what is @PirateJC and @PatrickRyanMS thoughts? While by key would be better technically is it too much to have users type an extra parameter?

PatrickRyanMS:
I'm not sure I follow the question. Are we needing to pass a key value instead of the name of the control? It would be a more straight path for the user to be able to get by the name they gave the control instead of an ID or key value that we assign since the user will have to find that value to pass. Is that what we are suggesting here?

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -272,6 +272,29 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+    /**
+    * Will return the first control with the given id
+    * @param uniqueId defines the id to search for
+    * @return the first control found or null
+    */
+    public getControlById(uniqueId: number): Nullable<Control> {
+        return this.getControlByKey("uniqueId", uniqueId);
+    }
+
+    private getControlByKey(key: string, value: any): Nullable<Control> {
```
RaananW, line 294:
```suggestion
    private _getControlByKey(key: string, value: any): Nullable<Control> {
```

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -272,6 +272,29 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+        return this.getControlByKey("name", name);
+    }
+
+    /**
+    * Will return the first control with the given id
+    * @param uniqueId defines the id to search for
+    * @return the first control found or null
+    */
+    public getControlById(uniqueId: number): Nullable<Control> {
+        return this.getControlByKey("uniqueId", uniqueId);
```
RaananW, line 291:
```suggestion
        return this._getControlByKey("uniqueId", uniqueId);
```

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -272,6 +272,29 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
         return this._rootContainer.getDescendants(directDescendantsOnly, predicate);
     }
+
+    /**
+    * Will return the first control with the given name
+    * @param name defines the name to search for
+    * @return the first control found or null
+    */
+    public getControlByName(name: string): Nullable<Control> {
+        return this.getControlByKey("name", name);
```
RaananW, line 282:
```suggestion
        return this._getControlByKey("name", name);
```

RaananW (approved):

PR #11187, opened 2021-09-29
[Modular Babylon] Moving static-only classes to modules

RaananW:
Babylon has a lot of classes that only contain static functions.
The issue with those classes is that if the user requires only one function they still need to import the entire class.

Example of those type of classes are the tools classes (like tools, filetools etc'). 

This PR moves some of those classes to be modules but keeps back-compat by exporting objects that contain the same objects as the classes.

This PR also intoroduces the `initSideEffects` function, which is executed when sideEffects is set to true, but being ignored when the package's sideEffect is false. This is a draft of eliminating sideeffects.

This is the first of more changes required. 

Setting it as draft to get feedback. I will continously push further changes based on the feedback.

(Most small changes are for the register class function, skip to the tools to see the bigger changes)


deltakosh (approved):
Perfect!

sebavan:
LGTM GG @RaananW !!!

Popov72:
That's quite a PR!

> This PR moves some of those classes to be modules but keeps back-compat by exporting objects that contain the same objects as the classes.

How are you doing this (the back compat)? I could not find something in the changes that would support it (but I certainly may have missed it as there are so many files changed!).

RaananW:
> That's quite a PR!
> 
> > This PR moves some of those classes to be modules but keeps back-compat by exporting objects that contain the same objects as the classes.
> 
> How are you doing this (the back compat)? I could not find something in the changes that would support it (but I certainly may have missed it as there are so many files changed!).

The idea is to have a single exported object that holds all of the (now) standalone functions and elements. A simple example can be found here:

https://github.com/BabylonJS/Babylon.js/pull/11187/files#diff-62506d07e85176c9b551b5aeff6c04b042937b30f8808089091343deeb1d3f20R114

This is the reason all tests do not fail and playgrounds will continue working as expected.

Those objects are only there for back-compat and should not be imported anywhere else. If they are imported, the unused functions will be be tree-shaken out.

A more extensive example can be seen in the filetools, which also has mutable variables (static getters and setters)

Popov72:
Thanks, I did miss these declarations!

src/Materials/Textures/texture.ts
```diff
@@ -764,9 +764,9 @@ export class Texture extends BaseTexture {
             if (typeof this._buffer === "string" && (this._buffer as string).substr(0, 5) === "data:") {
                 serializationObject.base64String = this._buffer;
                 serializationObject.name = serializationObject.name.replace("data:", "");
-            } else if (this.url && StartsWith(this.url, "data:") && this._buffer instanceof Uint8Array) {
+            } else if (this.url && StringTools.StartsWith(this.url, "data:") && this._buffer instanceof Uint8Array) {
                 serializationObject.base64String = "data:image/png;base64," + EncodeArrayBufferToBase64(this._buffer);
-            } else if (Texture.ForceSerializeBuffers) {
+            } else if (Texture.ForceSerializeBuffers || (this.url && StartsWith(this.url, "blob:"))) {
```
Popov72, line 769: Is it expected that you use `StringTools.StartsWith` two lines above and only `StartsWith` here?

RaananW: nope, this is wrong and should be fixed. Thanks!!
In the future we will use eslint to make sure those legacy imports are not used.

PR #11192, opened 2021-09-30
scene pointerMoveTrianglePredicate

lockphase:
Pls see: https://forum.babylonjs.com/t/scene-trianglepickingpredicate/24088

The default picking during pointer move is somewhat limiting in cases where the scene has many assets (with distinct action managers) which may be partially occluded to the user. Allowing a user to define custom triangle picking predicate extends the functionality of pointer move events while still preserving compatibility with action managers.


RaananW (approved):
Perfect! thank you so much :-)

PR #11194, opened 2021-09-30
[Modular Babylon] suggestion of keeping work out of internal texture

RaananW:
The side effect defined in Environment tools was not yet removed, but with this implementation there is little to no need for it, especially since it is a hidden function.

This is a standalone PR and is not related to the public-static restructure. It does complement it.

### DIFF
--- a/src/Materials/Textures/internalTexture.ts
+++ b/src/Materials/Textures/internalTexture.ts
@@ -143,7 +143,7 @@ export class InternalTexture extends TextureSampler {
      */
     public onLoadedObservable = new Observable<InternalTexture>();
     /**
-     * If this callback is defined is will be called instead of the default _rebuild function
+     * If this callback is defined it will be called instead of the default _rebuild function
      */
     public onRebuildCallback: Nullable<(internalTexture: InternalTexture) => {
         proxy: Nullable<InternalTexture | Promise<InternalTexture>>;
--- a/src/Materials/Textures/internalTexture.ts
+++ b/src/Materials/Textures/internalTexture.ts
@@ -82,7 +82,7 @@ export enum InternalTextureSource {
 export class InternalTexture extends TextureSampler {
 
     /** @hidden */
-    public static _UpdateRGBDAsync = (internalTexture: InternalTexture, data: ArrayBufferView[][], sphericalPolynomial: Nullable<SphericalPolynomial>, lodScale: number, lodOffset: number): Promise<void> => {
+    public static _UpdateRGBDAsync = (internalTexture: InternalTexture, data: ArrayBufferView[][], sphericalPolynomial: Nullable<SphericalPolynomial>, lodScale: number, lodOffset: number): Promise<InternalTexture> => {
         throw _DevTools.WarnImport("environmentTextureTools");
     }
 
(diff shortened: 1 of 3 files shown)


src/Materials/Textures/internalTexture.ts
```diff
@@ -142,6 +142,14 @@ export class InternalTexture extends TextureSampler {
      * Observable called when the texture is loaded
      */
     public onLoadedObservable = new Observable<InternalTexture>();
+    /**
+     * If this callback is defined is will be called instead of the default _rebuild function
```
Popov72, line 146: nit: "it" instead of the second "is"

src/Materials/Textures/rawCubeTexture.ts
```diff
@@ -81,6 +81,15 @@ export class RawCubeTexture extends CubeTexture {
...
     public static _UpdateRGBDAsync(internalTexture: InternalTexture, data: ArrayBufferView[][], sphericalPolynomial: Nullable<SphericalPolynomial>, lodScale: number, lodOffset: number): Promise<void> {
+        internalTexture.onRebuildCallback = (_internalTexture) => {
+            const proxy = internalTexture.getEngine().createRawCubeTexture(null, internalTexture.width, internalTexture.format, internalTexture.type,
+                internalTexture.generateMipMaps, internalTexture.invertY, internalTexture.samplingMode, internalTexture._compression);
+            return {
+                proxy: EnvironmentTextureTools._UpdateRGBDAsync(proxy, internalTexture._bufferViewArrayArray!, internalTexture._sphericalPolynomial, internalTexture._lodGenerationScale, internalTexture._lodGenerationOffset),
+                isReady: true,
+                isAsync: true
+            };
+        };
```
Popov72, line 92: Maybe this function (the code below) can be implemented by simply calling `EnvironmentTextureTools._UpdateRGBDAsync(...)`, as it is doing exactly the same thing?

Popov72 (approved):

src/Materials/Textures/rawCubeTexture.ts
RaananW: yes, if this is merged I will do the cleanup work in the other draft PR that organizes the public-static classes. Otherwise I will have many conflicts

RaananW: strike that. I changed the function :-)
Thanks!

src/Misc/environmentTextureTools.ts
```diff
@@ -715,7 +715,7 @@ export class EnvironmentTextureTools {
     }
 
     /** @hidden */
-    public static _UpdateRGBDAsync(internalTexture: InternalTexture, data: ArrayBufferView[][], sphericalPolynomial: Nullable<SphericalPolynomial>, lodScale: number, lodOffset: number): Promise<void> {
+    public static _UpdateRGBDAsync(internalTexture: InternalTexture, data: ArrayBufferView[][], sphericalPolynomial: Nullable<SphericalPolynomial>, lodScale: number, lodOffset: number): Promise<InternalTexture> {
         internalTexture._source = InternalTextureSource.CubeRawRGBD;
```
sebavan, line 719: Should this not add the callback as well for now ?

RaananW: well, this is the callback. It shouldn't register itself.

This PR is meant to decouple the rebuild from internal texture, this is just one specific case which I tried not to change too much.

RaananW: Whereas I agree that the current architecture doesn't make too much sense, I don't see where rebuild is being triggered

PR #11198, opened 2021-10-01
[GE] Saving W/H data, Adding New Screen Sizes, and Various Bug Fixes

msDestiny14:
BabylonJS/ThePirateCove#95
BabylonJS/ThePirateCove#96
https://github.com/BabylonJS/ThePirateCove/issues/90
https://github.com/BabylonJS/ThePirateCove/issues/80


gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -951,6 +954,11 @@ export class AdvancedDynamicTexture extends DynamicTexture {
      */
     public parseContent(serializedObject: any) {
         this._rootContainer = Control.Parse(serializedObject.root, this) as Container;
+        const width = serializedObject.width;
+        const height = serializedObject.height;
+        if (width && height) {
```
RaananW, line 959: this will only rescale if both are not 0. is that the expected behavior?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -385,9 +375,46 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
...
+            new Vector2(1125, 2436),
+            new Vector2(1170, 2532),
+            new Vector2(1284, 2778),
+            new Vector2(1080, 2220),
+            new Vector2(1080, 2340),
+            new Vector2(1024, 1024),
+            new Vector2(2048, 2048),
+        ];
+        const _sizeOptions = [
+            { label: "Web (1920)", value: 0 },
```
RaananW, line 396: so, i wonder - why reference the value as an index in the array when the values are not being reused? Why not keep the vector2 here as value?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -66,7 +66,9 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
     private _clipboard: Control[] = [];
     private _selectAll: boolean = false;
     private _camera: ArcRotateCamera;
-    private _cameraRadias = 1500;
+    private _cameraRadias: number;
+    private _cameraMaxRadiasFactor = 8192; // 2^13
```
RaananW, line 70: why did you select this number? shouldn't this be relative to the canvas size?

guiEditor/src/tools.ts
```diff
@@ -3,15 +3,15 @@ import { Grid } from "babylonjs-gui/2D/controls/grid";
 import { Vector2 } from "babylonjs/Maths/math";
 
 export class Tools {
-    public static LookForItem(item: any, selectedEntity: any): boolean {
-        if (item === selectedEntity) {
+    public static LookForItem(item: any, selectedEntity: any, firstIteration : boolean = true): boolean {
+        if (!firstIteration && item === selectedEntity) {
```
RaananW, line 7: is that the expected default behavior? always ignore the first group of children?

RaananW (commented):
LGTM! just a few questions i've got

guiEditor/src/diagram/workbench.tsx
msDestiny14: Camera limitations.  I am thinking of lowering canvas size now to match.

gui/src/2D/advancedDynamicTexture.ts
msDestiny14: Should i change it !== null instead. Want to cover the case for previous jsons where this is not defined.

guiEditor/src/tools.ts
msDestiny14: Otherwise the selected container will always be open (even if we don't want it to)

gui/src/2D/advancedDynamicTexture.ts
RaananW: you know how you want to deal with it :-)

You can check the typeof width and typeof height (should be be 'number') instead. And that they are >= 0

guiEditor/src/diagram/workbench.tsx
RaananW: is there a max radius to the camera?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: It's used lower? I'm not sure what this question means.

guiEditor/src/diagram/workbench.tsx
msDestiny14: apparently. Things glitch out after that. Just like adts stop working after 2^14

msDestiny14: apparently. Things glitch out after that. Just like adts stop working after 2^14

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
RaananW: wh is value not `Vector2(102, 1024)` ? Why do you need the other array if you already have a 1-to-1 map?

guiEditor/src/diagram/workbench.tsx
RaananW: can you show me a playground where that happens?

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
msDestiny14: Discussed offline. part of the structure of optionsLineComponent.

guiEditor/src/diagram/workbench.tsx
msDestiny14: did some updates :)

PR #11201, opened 2021-10-01
Add Observable.FromPromise

Symbitic:
Adds an `Observable.FromPromise` utility method to create an Observable from a Promise.

This will probably be used more by developers than within Babylon itself, but I still think it could be a useful utility because it will help people to link non-Babylon libraries into their applications.

I also created a pull request (BabylonJS/Documentation/pull/339) to add it to the documentation.


RaananW (approved):
LGTM!

It's worth mentioning in the docs (didn't read them yet) that is onErrorObservable is not defined the promise itself will need to catch the error, otherwise it will result in an unhandled promise rejection.
And that errors during the notify process will also be caught by the error observable.

PR #11202, opened 2021-10-01
[Modular Babylon] Moving deprecated functions to LTS version

RaananW:
This is a suggestion to the way we can deprecate functionality and not include it in different versions of the framework.

The suggestion is this:

Define two versions of babylon. an LTS version that will forever stay backwards compatible, and a "regular" version, with backwards compatibility based on major version. The lts version is being implemented using (filename).lts.ts files, that are only included in the LTS build.
This way we can continue releasing LTS backwards compatible version to whomever wishes to use it. This version might have small issues with tree shaking or bundle size, but will always work. 
This PR shows how we can "deprecate" both static methods and regular methos on both the mesh class and the scene class.

Note - the LTS version will have side effects, and this is something the users of the LTS version will have to live with.


sebavan (approved):

RaananW:
Some explanation to the current implementation - 

Imports are hoisted, so only importing the `.lts` files will not work as expected, as it will be executed before the code in the main module was executed. We need to therefore import a function that will be executed at the end of the main file. When building the esm version with no side effects this function will exist but will not be executed so it will be removed from the final build. 

The `_injectLTS` function can be injected post-build and doesn't have to be included in the main source file. An extra build step can be added (pseudo code!!)

```javascript
if_file_exists(filename.replace('.ts', '.lts.ts')) {
   add_to_source(`
import { _injectLTS } from "${filename.replace('.ts', '.lts.ts')}";
_injectLTS();
`);
}
```

Only adding `.lts` to all `index` files will not work, as the core libraries are importing from `@babylonjs/core/scene`, which, in this case, will need to do the internal importing for the lts version for it to work correctly.

A note about circular dependencies - as this function is executed after the main source was initialized we don't have an issue with cicular dependencies (even though these ARE circular dependencies). Another approach would be to pass the class  to the `_injectLTS` function. This has an issue with typescript that requires a full import (and not only a type declaration) in order to change the passed module (or class).

PR #11206, opened 2021-10-01
ShadowDepthWrapper: Fix crash when not passing a scene to the constructor

Popov72:


src/Materials/shadowDepthWrapper.ts
```diff
@@ -81,7 +81,7 @@ export class ShadowDepthWrapper {
      */
     constructor(baseMaterial: Material, scene: Scene, options?: IIOptionShadowDepthMaterial) {
         this._baseMaterial = baseMaterial;
-        this._scene = scene;
+        this._scene = scene ?? Engine.LastCreatedScene;
```
darraghjburke, line 84: Looks like this causes a Typescript error: TS2304: Cannot find name 'Engine'.

Popov72: Sorry, I missed adding the import. Fixed.

darraghjburke (approved):

PR #11217, opened 2021-10-05
Texture error when cached

sebavan:
https://forum.babylonjs.com/t/creation-of-texture-using-a-url-which-has-already-caused-failure-will-not-trigger-onerror-handler/24322/3

### DIFF
--- a/src/Materials/Textures/texture.ts
+++ b/src/Materials/Textures/texture.ts
@@ -457,7 +457,11 @@ export class Texture extends BaseTexture {
             if (this._texture.isReady) {
                 TimingTools.SetImmediate(() => load());
             } else {
-                this._texture.onLoadedObservable.add(load);
+                const loadObserver = this._texture.onLoadedObservable.add(load);
+                this._texture.onErrorObservable.add((e) => {
+                    errorHandler(e.message, e.exception);
+                    this._texture?.onLoadedObservable.remove(loadObserver);
+                });
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/Materials/Textures/texture.ts
```diff
@@ -457,7 +457,11 @@ export class Texture extends BaseTexture {
             if (this._texture.isReady) {
                 TimingTools.SetImmediate(() => load());
             } else {
-                this._texture.onLoadedObservable.add(load);
+                const loadObserver = this._texture.onLoadedObservable.add(load);
+                this._texture.onErrorObservable.add((e) => {
+                    errorHandler(e.message, e.exception);
```
deltakosh, line 462: smart

deltakosh (approved):

src/Materials/Textures/texture.ts
```diff
@@ -457,7 +457,11 @@ export class Texture extends BaseTexture {
             if (this._texture.isReady) {
                 TimingTools.SetImmediate(() => load());
             } else {
-                this._texture.onLoadedObservable.add(load);
+                const loadObserver = this._texture.onLoadedObservable.add(load);
+                this._texture.onErrorObservable.add((e) => {
+                    errorHandler(e.message, e.exception);
+                    this._texture?.onLoadedObservable.remove(loadObserver);
```
Popov72, line 463: If we update the url, does it trigger `onLoadedObservable`? If yes, I think you will have to remove the line above.

Popov72 (approved):

PR #11218, opened 2021-10-05
Fix for link with mesh and render scale

msDestiny14:
https://forum.babylonjs.com/t/control-linkwithmesh-breaks-when-advanceddynamictexture-renderscale-is-not-1/24301/4

### DIFF
--- a/gui/src/2D/advancedDynamicTexture.ts
+++ b/gui/src/2D/advancedDynamicTexture.ts
@@ -601,8 +601,7 @@ export class AdvancedDynamicTexture extends DynamicTexture {
                     continue;
                 }
                 control.notRenderable = false;
-                // Account for RenderScale.
-                projectedPosition.scaleInPlace(this.renderScale);
+
                 control._moveToProjectedPosition(projectedPosition);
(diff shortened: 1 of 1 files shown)


sebavan (changes requested):
Why not reusing getProjectedPositionWithZ from getProjectedPosition so you avoid the weird return line or the helper

sebavan:
Yup you could do this as it is not used every frame.

But another way is to mainly share only the:

var globalViewport = this._getGlobalViewport();
var projectedPosition = Vector3.Project(position, worldMatrix, scene.getTransformMatrix(), globalViewport);

with a projectToRef instead and so you could simply call this internal from the both functions + the checkupdate so next time we have only one place to maintain.

msDestiny14:
My only concern with the check update function is its a lot of calls on this._getGlobalViewport();

sebavan:
Good catch fully useless to this one !!!

msDestiny14:
Want me to just clean up the returns then?

sebavan (approved):

PR #11233, opened 2021-10-08
Fix Shader Material Effect Cache

sebavan:
https://forum.babylonjs.com/t/uncaught-typeerror-cannot-read-properties-of-undefined-reading-world/24328/10

For reference, what is happening is that a shared shader Material across meshes can have different defines, from instances, vertex color, moph and bones. Constructing the pseudo cache and the custom `_checkCache` due to the frequency of use is probably less efficient or breaks even with removing the `_checkCache`.


RaananW:
does this practically remove the cache from the shader material?

PR #11252, opened 2021-10-12
Fix Flat Shaded Generation for simple GLTF

sebavan:
https://forum.babylonjs.com/t/whats-wrong-with-my-lighting/24404/25


sebavan:
I ll change with material info if it pops on the forum :-)

PR #11266, opened 2021-10-13
Fix serialization of edited textures

darraghjburke:
Addresses this issue from the forum: https://forum.babylonjs.com/t/texture-swap-incorrectly-accounted-for-when-exporting-modified-babylon-scene-from-sandbox/24184?u=darraghburke

If you edit a texture, it no longer corresponds to the source URL, so we should set the URL to null. When we serialize the scene, we also want to make sure to generate base64 strings from the texture if the URL is null.


RaananW:
Just a small documentation issue and ready to merge :-)

PR #11289, opened 2021-10-17
Fix copying snippet id to clipboard for Chrome

BlakeOne:
Patch to fix the issue that Chrome can't copy to the clipboard unless the document has focus, by waiting until after writing to the clipboard has finished to create the alert window. Also the user is only told that the id has been copied to the clipboard if it's written successfully.

Forum post: https://forum.babylonjs.com/t/introducing-the-gui-editor-alpha/24578/35


BlakeOne:
Oops, I forgot to add a line to the What's New Update. I'd add it but I'm not sure if I should add it to the top of the GUI section or if maybe a new GUI Editor section is going to be created?

PR #11304, opened 2021-10-19
WebGPU: non compatibility mode implementation (aka "fast path")

Popov72:
The implementation is not finished yet but the PR is already quite big, so there will be another one (normally quite smaller!) to account for the missing bits (the biggest one being adding support for submeshes in `ShaderMaterial`).

Some quick numbers before describing the PR:
* even if it's a WebGPU PR, some changes also affect WebGL (like the render passes) that improve performances: the JS thread takes between 5% and 20% less time on the PGs I have tested compared to WebGL 4.2 (it's also better than the current WebGL 5.0 code in prod)
* WebGPU:
  * the JS thread takes between 10% and 16% **more** time than in WebGL in the *compatibility mode* (that is, with `engine.compatibilityMode = true`). It's expected, in this mode we have to use a number of caches that slow things down in WebGPU
  * the JS thread takes between 14% and 20% **less** time than in WebGL in the *non compatibility mode* (that is, with `engine.compatibilityMode = false`, the purpose of the PR)
  * note that the comparison with WebGL is done against the numbers measured **with** this PR for WebGL, not against the WebGL code currently in prod!

Here's an overview of the PR:
* I have added render pass ids, which are simply ids that are assigned to each "render pass". Basically, the main render pass (the one rendering the scene into the framebuffer) is always id #0 and each rendering by a `RenderTargetTexture` is using a specific id (or several, if the RTT is a cube or a 2D layer). The id(s) are allocated in the class constructor. This id is used to retrieve the right `DrawWrapper` on the submesh that is currently rendered (those wrappers are stored in the `SubMesh._drawWrappers` array: the index in this array is the render pass id). In WebGPU that allows to have a bundle cached (in `DrawWrapper.drawContext`) for each rendering and reusing it when in the _non compatibility mode_. However, it's also a perf gain in WebGL because the effects are not recreated/retrieved from the engine each time we execute `isReadyForSubMesh` but are retrieved from the `DrawWrapper`.
* I'm going to create a doc page about render pass ids and how they allow to use a different material depending on the pass (render target) a mesh is rendered into => see `RenderTargetTexture.setMaterialForRendering` in the current PR, but I have also added this method to the depth renderer (`DepthRender.setMaterialForRendering`). I will also update the `EffectLayer` class with this feature.
* Regarding post processes, I needed to know if one is going to overwrite the `textureSampler` texture of the shader. Indeed, the `PostProcess` class (in the `apply` method) is setting the value with the source input value, but some effects were overwriting it just afterward in a `postprocess.onApplyObserver` observer: in WebGPU it kills the bundle cache because it dirtyfies the material context. So, setting `postprocess.externalTextureSamplerBinding = true` instructs the `PostProcess` class to **not** update the value itself because it will be done by some external code. Note it's an optimization (for WebGPU): if you don't do it it still work, but you will have some bundles recreated each frame that could be avoided.
* When in fast mode, if using instances, we use an indirect draw to be able to update the instance count without recreating the cached bundle: we simply need to update the buffer holding the draw parameters (and more precisely the instance count parameter)


sebavan:
> I'm not sure what to look for regarding the error in the **Viewer Tests**...
> 
> The latest one is failing with this error:
> 
> ```
> Chrome Headless 94.0.4606.81 (Linux x86_64) scene manager should allow disabling and enabling ground texture FAILED
> 	Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
> ```
> 
> Locally everything is ok it seems:
> 
> ```
> ...
> LOG: 'BJS - [23:55:40]: Babylon.js v5.0.0-alpha.56 - Null engine'
> Chrome Headless 94.0.4606.81 (Windows 10): Executed 46 of 46 SUCCESS (2.325 secs / 2.117 secs)
> TOTAL: 46 SUCCESS
> [23:55:40] Finished '<anonymous>' after 4.76 s
> [23:55:40] Finished 'tests-viewer-unit' after 36 s
> ```

@RaananW can you have a look at this on ?

Popov72:
New update: added a `renderPassId` to `Camera`.

Instead of using the renderPassId = 0 to render the main scene, the `renderPassId` from the camera is used instead. It allows to optimize things even when multiple cameras render the scene into the main framebuffer, as in [this PG](https://playground.babylonjs.com/#D8AK3Z#35): the skybox is rendered by both cameras. Without the change, the render pass id 0 is used the two times which leads to a bundle recreation each time because the scene ubo is not the same when rendering through the first or the second camera.

PR #11306, opened 2021-10-19
Rename step in ScreenSpaceReflection Shader

darraghjburke:
The SSR post process shader has a uniform float named `step`, but `step` is a function in GLSL.
image
It still compiles fine on the web, but for Babylon Native this causes issues. See: https://github.com/BabylonJS/BabylonNative/issues/922

This PR changes only the name of the uniform, not the member variable on the class, so it will not break backward compatibility. (If we want to change the member variable as well, I can add getters + setters for the old name, but it didn't seem necessary to me.)


sebavan:
Awesome catch !!!

PR #11312, opened 2021-10-20
DeviceInputSystem: Fix Safari compatibility and remove "FakeMove"

PolygonalSun:
This PR contains a few changes that would remove FakeMove from the DeviceInputSystem and make it so that camera movement no longer gets stuck on Safari when the button is released while moving.

### DIFF
--- a/src/Cameras/Inputs/freeCameraMouseInput.ts
+++ b/src/Cameras/Inputs/freeCameraMouseInput.ts
@@ -44,7 +44,6 @@ export class FreeCameraMouseInput implements ICameraInput<FreeCamera> {
     public _allowCameraRotation = true;
 
     private _currentActiveButton: number = -1;
-    //private _usingSafari = Tools.IsSafari();
 
     /**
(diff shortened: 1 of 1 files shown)


PolygonalSun:
> I am always over scarred by input changes :-) Did you double check on all platforms/browsers as well as camera types ?

I checked both Chrome and Safari on MacOS.  On Windows, I checked Chrome/Edge and Firefox.  I tried to test against devices using `npm run start-public-ssl` but it wasn't reaching any of my other devices so I haven't had a chance to try mobile

PolygonalSun:
> > I am always over scarred by input changes :-) Did you double check on all platforms/browsers as well as camera types ?
> 
> I checked both Chrome and Safari on MacOS. On Windows, I checked Chrome/Edge and Firefox. I tried to test against devices using `npm run start-public-ssl` but it wasn't reaching any of my other devices so I haven't had a chance to try mobile

Checked on iOS and there are no negative changes

PR #11317, opened 2021-10-20
Baked vertex animation textures

brunobg:
This PR implements Vertex Animation Textures, according to the discussion at https://forum.babylonjs.com/t/vertex-animation-texture-module-implementation/24576. 

Suggestion of how to provide proper tests are welcome.

Code fragment to test this in the playground:

```
let baker = null, mesh = null;
    BABYLON.SceneLoader.ImportMeshAsync(
        "",
        "https://raw.githubusercontent.com/RaggarDK/Baby/baby/",
        "arr.babylon",
        scene,
        undefined
    ).then((importResult) => {
        mesh = importResult.meshes[0];
        baker = new BABYLON.VertexAnimationBaker(scene, mesh);
        return baker.bakeVertexData([{from: 1, to: 20, name: "aaa"}]);
    }).then((vertexData) => {
        const vertexTexture = baker.textureFromBakedVertexData(vertexData);
        mesh.material.bakedVertexAnimationMap.texture = vertexTexture;
        mesh.material.bakedVertexAnimationMap.animationParameters = new BABYLON.Vector4(
            0,
            20,
            0, 
            30
        );

        scene.registerBeforeRender(() => {
            mesh.material.bakedVertexAnimationMap.time += 1/60;
        });
    });
```
Specific points for code reviewers:

- [x] check if I implemented `time` correctly.
- [x] check `IMaterialBakedVertexAnimationDefines`. Does anybody see a define that we need? I don't see one.
- [x] `IMaterialBakedVertexAnimationDefines.setAnimationParameters`: I have no idea how to handle this. It's actually meant to be a set of parameters in an `instanceBuffer`, so I don't even think it belongs to that class. But I want to provide a helper function so it isn't an opaque call like `mesh.instancedBuffers.bakedVertexAnimationSettings = new Vector4(...)`. 
     - [x] Should it be part of `Mesh`?
     - [x] Is there a way to make it work for non-instance meshes as well? One may want to use this for a pure mesh.
     - [ ] Same for SPS, any way to make this work properly with SPS, or would `instancedBuffers` work for particles?
     - [x] same for `thinInstanceSetBuffer`.

Things left to implement, still open (feedback welcome):

- [x] change PBR material like StandardMaterial
- [x] Test this: `Maybe this._mesh.skeleton.prepare() will work instead of doing a full this._scene.render() call.`
- [x] automatic tests (suggestions here, please)
- [x] `In the instanced case, the animation data will be provided by the bakedVertexAnimationSettings instance buffer of the mesh (that the user will have to create themselves - maybe with the help of the baker class).` => see what the baker misc class can do to help with this.
- [x] add entries to the inspector:
    - [x] `inspector/src/components/actionTabs/tabs/propertyGrids/materials/pbrMaterialPropertyGridComponent.tsx`
    - [x] `inspector/src/components/actionTabs/tabs/propertyGrids/materials/standardMaterialPropertyGridComponent.tsx`
- [x] some way to serialize textures to files
- [x] some way to load textures from files
- [x] documentation with a playground example
- [x] credit @raggar
- [x] should I add `#include<bakedVertexAnimation>` and `#include<bakedVertexAnimationDeclaration>` in every `.fx` file that has `#include<bonesVertex>` and  `#include<bonesDeclaration>`?

Open questions:

- [ ] should this be integrated into the animation system? Perhaps `Mesh.beginAnimation()` or `Scene.beginAnimation()` could use VATs when they are available? If so VATs should be called by the animation system. If this is the case I'll need some guidelines as to how this should be done.


brunobg:
@sebavan thanks for the review, but could you give me a quite guideline on how to write a test for this? Some existing test that I could use as a guideline? I miss being able to test this code while writing it.

Popov72:
> @sebavan thanks for the review, but could you give me a quite guideline on how to write a test for this? Some existing test that I could use as a guideline? I miss being able to test this code while writing it.

We don't usually write unit tests (even if we have a number of them), but we generally prefer to write visual validation tests, to make sure the output is not corrupted when adding some new code to the codebase. We can deal with this at the very end when everything is working.

brunobg:
> I noticed the serialization/parsing code is broken because it converts the data to png, which does not work as we are dealing with floats. I changed it to simply dump/reload the values as an array of numbers.

I was trying to avoid naive serialization because it's so much larger than the float array size. But I missed the PNG conversion in that method. Are you okay with enabling '--downlevelIteration' on tsconfig? I could add a new `GenerateBase64StringFromArray` method using the base64 solution from https://stackoverflow.com/a/53108630/890870. 

I merged your commit but tests are not passing.

Popov72:
I will have a look tomorrow for the validation test.

Regarding the serialization, I don't know what `downlevelIteration` is used for but you can't use a spread construct like `...XXX` reliably with `String.fromCharCode` because it can lead to errors:
image

I'm all for using a transformation of the float array to binary then to base64 to save space but not using the spread operator.

Popov72:
Let's see if this change will fix the validation test problem:
https://github.com/Popov72/Babylon.js/commit/db84487848046049e58da2c8cb5b57e6a615ac34

brunobg:
> Regarding the serialization, I don't know what downlevelIteration is used for but you can't use a spread construct like ...XXX reliably with String.fromCharCode because it can lead to errors:

True, I should have provided a better example, sorry. Here's the sketch of `btoa/atob` version that works with large arrays:

```js
function encode(buffer) {
  return btoa(new Uint8Array(buffer).reduce((data, byte)=> {
     return data + String.fromCharCode(byte);
  }, ''));
}
function decode(str) {
   return new Float32Array(Uint8Array.from(atob(str), c => c.charCodeAt(0)));
}
```

Here's a decent library implementation that is not stack based and works properly: https://github.com/niklasvh/base64-arraybuffer/blob/master/src/index.ts 

There's also the [MDN implementation to avoid btoa/atob](https://developer.mozilla.org/en-US/docs/Glossary/Base64#appendix_decode_a_base64_string_to_uint8array_or_arraybuffer). 

Perhaps this could be also used in the `.babylon` files to compact them and load them more quickly.

Popov72:
Any implementation seems nice, you can pick any one you want (we should just give credits for it).

Regarding the validation test, it seems there's something with the Azure server which makes it fail. You should add `"excludeFromAutomaticTesting": true` in the config file:
```json
        {
            "title": "Baked Vertex Animation",
            "playgroundId": "#14WJWW#9",
            "referenceImage": "bakedVertexAnimation.png",
            "excludeFromAutomaticTesting": true
        }
```

brunobg:
Actually, you already have this serialization code in the repo. Changed the baker class to use it, tested in a PG and it all looks good. I have nothing else to add to this PR, so hopefully this is ready to be merged :)

Once again thanks for the huge amount of help.

sebavan:
@Popov72 let me know when you are all good and I ll make a final pass with @deltakosh

Popov72:
Good news!

Actually, the validation test is going to fail because you need to update the serialized data which is using the old format for the float array.

You can use this PG to generate the string:
https://playground.babylonjs.com/#14WJWW#10

=> look in the browser console for the string to set as the value for `vatJson` in the PG

brunobg:
> Actually, the validation test is going to fail because you need to update the serialized data which is using the old format for the float array.

Updated.

materialsLibrary/src/fire/fire.vertex.fx
```diff
@@ -13,6 +13,7 @@ attribute vec4 color;
 #endif
 
 #include<bonesDeclaration>
+#include<bakedVertexAnimationDeclaration>
```
Popov72, line 16: `#include<bakedVertexAnimation>` is missing

materialsLibrary/src/fur/fur.vertex.fx
```diff
@@ -15,6 +15,7 @@ attribute vec4 color;
 #endif
 
 #include<bonesDeclaration>
+#include<bakedVertexAnimationDeclaration>
```
Popov72, line 18: `#include<bakedVertexAnimation>` is missing

materialsLibrary/src/mix/mix.vertex.fx
```diff
@@ -16,6 +16,7 @@ attribute vec4 color;
 #endif
 
 #include<bonesDeclaration>
+#include<bakedVertexAnimationDeclaration>
```
Popov72, line 19: `#include<bakedVertexAnimation>` is missing

materialsLibrary/src/terrain/terrain.vertex.fx
```diff
@@ -16,6 +16,7 @@ attribute vec4 color;
 #endif
 
 #include<bonesDeclaration>
+#include<bakedVertexAnimationDeclaration>
```
Popov72, line 19: `#include<bakedVertexAnimation>` is missing

src/BakedVertexAnimation/bakedVertexAnimationManager.ts
```diff
@@ -0,0 +1,155 @@
...
+import { Scene } from "../scene";
+import { serialize, expandToProperty, serializeAsTexture, SerializationHelper } from "../Misc/decorators";
+import { BaseTexture } from '../Materials/Textures/baseTexture';
+import { Vector4 } from "../Maths/math.vector";
+import { Effect } from "../Materials/effect";
+
+/**
+ * This class is used to animate meshes using a baked vertex animation texture
+ * @see https://doc.babylonjs.com/divingDeeper/animation/baked_texture_animations
+ */
```
Popov72, line 11: Needs to add `@since 5.0` in the doc

src/BakedVertexAnimation/vertexAnimationBaker.ts
```diff
@@ -0,0 +1,158 @@
+import { AnimationRange } from "../Animations/animationRange";
+import { RawTexture } from "../Materials/Textures/rawTexture";
+import { Texture } from "../Materials/Textures/texture";
+import { Mesh } from "../Meshes/mesh";
+import { EncodeArrayBufferToBase64, DecodeBase64ToBinary } from "../Misc/stringTools";
+import { Scene } from "../scene";
+
+/**
+ * Class to bake vertex animation textures.
+ */
```
Popov72, line 10: Needs to add `@since 5.0` in the doc

src/BakedVertexAnimation/vertexAnimationBaker.ts
```diff
@@ -0,0 +1,158 @@
...
+        return vertexData;
+    }
+
+    /**
+     * Runs an animation frame and stores its vertex data
+     *
+     * @param vertexData The array to save data to.
+     * @param frameIndex Current frame in the skeleton animation to render.
+     * @param textureIndex Current index of the texture data.
+     * @returns
```
Popov72, line 71: nit: you can remove the line

src/BakedVertexAnimation/vertexAnimationBaker.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        const texture = RawTexture.CreateRGBATexture(
+            vertexData,
+            (boneCount + 1) * 4,
+            vertexData.length / ((boneCount + 1) * 4 * 4),
+            this._scene,
+            false,
+            false,
+            Texture.NEAREST_NEAREST,
+            1
```
Popov72, line 105: Please use `Constants.TEXTURETYPE_FLOAT` instead of 1.

src/BakedVertexAnimation/vertexAnimationBaker.ts
```diff
@@ -0,0 +1,158 @@
...
+    constructor(scene: Scene, mesh: Mesh) {
+        this._scene = scene;
+        this._mesh = mesh;
+    }
+
+    /**
+     * Bakes the animation into the texture. This should be called once, when the
+     * scene starts, so the VAT is generated and associated to the mesh.
+     * @param ranges Defines the ranges in the animation that will be baked.
+     * @returns Float32Array
```
Popov72, line 30: Need better doc for the `@returns` value => there are a number of others below. Note that you don't have to document the protected/private methods, only the public ones.

Popov72 (approved):
All good for me!

Congrats @brunobg for this PR which is not a small one!

src/Meshes/abstractMesh.ts
```diff
@@ -18,6 +18,7 @@ import { MaterialDefines } from "../Materials/materialDefines";
 import { Light } from "../Lights/light";
 import { Skeleton } from "../Bones/skeleton";
 import { MorphTargetManager } from "../Morph/morphTargetManager";
+import { BakedVertexAnimationManager } from "../BakedVertexAnimation/bakedVertexAnimationManager";
```
deltakosh, line 21: We should not import but declare it
declare type BakedVertexAnimationManager = import("../BakedVertexAnimation/bakedVertexAnimationManager").BakedVertexAnimationManager ;


cc @RaananW to make sure it will not be a problem for tree shaking and side effects

deltakosh: I would recommend to expose an interface actually and use that interface here

deltakosh: Morphtarget is part of what is problematic for treeshaking and side effects (same as skeleton)

brunobg: Added an interface. I was not sure what should be there, take a look at let me know if something is missing/too much.

brunobg:
> Congrats @brunobg for this PR which is not a small one!

Me? I just sent bad commits that you fixed :)

Popov72:
It seems there's a problem with `AbstractMesh` which now shows as being completely modified by the PR whereas it should be only a couple of lines. Maybe a problem with the line ending characters?

src/Meshes/abstractMesh.ts
RaananW: If it is only used as an interface typescript will remove it from the final .js file, so it should be fine.

RaananW:
> It seems there's a problem with `AbstractMesh` which now shows as being completely modified by the PR whereas it should be only a couple of lines. Maybe a problem with the line ending characters?

This sometimes happens if you edit the file using the github file editor. If possible you can revert those changes locally (while keeping the changes uncommited) and solve this to maintain our git hustory.

src/Materials/shaderMaterial.ts
```diff
@@ -718,6 +718,17 @@ export class ShaderMaterial extends Material {
             defines.push("#define NUM_MORPH_INFLUENCERS 0");
         }
 
+        // Baked Vertex Animation
+        if (mesh) {
+            const bvaManager = (<Mesh>mesh).bakedVertexAnimationManager;
+
+            if (bvaManager && bvaManager.isEnabled) {
+                defines.push("#define BAKED_VERTEX_ANIMATION_TEXTURE");
```
RaananW, line 726: small nit - no need to add "#define", it is now added automatically (if not present)

src/Meshes/abstractMesh.ts
sebavan: yup but the declare prevents any maintenance misuse of the imported file :-) The interface is great I agree

sebavan:
@brunobg this looks stuning !!! thanks @Popov72 for the support !!! I ll merge it and deploy as soon as the AbstractMesh file is fixed.

brunobg:
Sorry about the lined end issues, I should have checked that, but I don't know why VSCode mangled that file. Did a review and reverted the whitespace issues on blank lines (but left a couple on line ends. BTW, the project has an `.editorconfig` with `trim_trailing_whitespace = true`, so any future patches will remove those spaces anyway), as well as reverted a couple of files that didn't have changes in the end. I understand that the interface on MeshAbstract is okay the way I did, right, instead of the declare? 

Hopefully that's all! Thanks to all of you and I can't wait to see this merged. I'll use it right away.

deltakosh (approved):
Excellent job! thanks a lot

brunobg:
Any plans to make a new release soon? I could use this right away :)

Popov72:
It's already available on the Playground / preview CDN, this PG does work:
https://playground.babylonjs.com/#14WJWW#12

If you need a NPM package I can't tell, however.

brunobg:
Ah, yes, I'm looking for the NPM package. I'll be waiting!

PR #11333, opened 2021-10-23
GUI scaling for XR

rickfromwork:
This change exposes a new member on the GUI3DManager that toggles "realistic" scaling of 3D UI components, according to default MRTK button sizes. This is an opt-in feature that drastically scales down UI so that it is an appropriate size for viewing when in an XR experience.

Also included in the change are some smaller fixes I addressed during debugging. Of note, changing the eye tracking check for the hand menu behavior, and removing a per-frame getPose query in the hand tracker.


gui/src/3D/gui3DManager.ts
```diff
@@ -57,6 +58,37 @@ export class GUI3DManager implements IDisposable {
...
+                if (newScale !== 1) {
+                    control._isScaledByManager = true;
+                }
+            });
+        }
+    }
+
+    /** Gets if controls attached to this manager are realistically sized, based on the fact that 1 unit length is 1 meter */
+    public get useRealisticScaling() {
+        return this.controlScaling === 0.1;
```
RaananW, line 84: This value ( `0.1`) should be fixed in a constant in this class. It is also possible that this will return true if not set to use realistic scaling, but setting control scaling to 0.1 (0.1 / 1 === 0.1) so maybe a flag would make more sense? Not sure of the real usage here though

gui/src/3D/gui3DManager.ts
```diff
@@ -185,6 +217,8 @@ export class GUI3DManager implements IDisposable {
      */
     public addControl(control: Control3D): GUI3DManager {
         this._rootContainer.addControl(control);
+        control.scaling.scaleInPlace(this._customControlsScale);
+        control._isScaledByManager = true;
```
RaananW, line 221: This is set to true by default, but in the `controlScaling` function it is only true if newScale is not 1. Any side-effects for that?

RaananW (commented):
That's a great abstract solution.

I have a few nit-comments. Otherwise, LGTM. Is the 0.1 scaling enough to make it render correctly in XR?

rickfromwork:
> That's a great abstract solution.
> 
> I have a few nit-comments. Otherwise, LGTM. Is the 0.1 scaling enough to make it render correctly in XR?

That's the default scaling I've been applying to my local scenes. It sets the GUI buttons to a size of 10x10cm, I'll need to double-check what the MRTK guidelines say about button sizes though before locking the scale in.

RaananW:
@rickfromwork - Any update on this one?

gui/src/3D/gui3DManager.ts
```diff
@@ -57,6 +58,37 @@ export class GUI3DManager implements IDisposable {
...
+                if (newScale !== 1) {
+                    control._isScaledByManager = true;
+                }
+            });
+        }
+    }
+
+    /** Gets if controls attached to this manager are realistically sized, based on the fact that 1 unit length is 1 meter */
+    public get useRealisticScaling() {
+        return this.controlScaling === 0.032;
```
RaananW, line 84: nit - maybe make that a const to reuse it when needed?

gui/src/3D/controls/holographicSlate.ts
```diff
@@ -42,12 +42,12 @@ export class HolographicSlate extends ContentDisplay3D {
     /**
      * Dimensions of the slate
      */
-    public dimensions = new Vector3(0.7, 0.4, 0.001);
+    public dimensions = new Vector3(21.875, 12.5, 0.001);
```
RaananW, line 45: just wondering - is that a breaking change, or does it get adjusted correctly?

RaananW (commented):
LGTM! just have a question and a nit :-)

gui/src/3D/controls/holographicSlate.ts
rickfromwork: It gets adjusted correctly when the GUI3DManager flag is set. The slate will appear larger by default though, but this shouldn't be an issue since it was pre-scaled to an XR-appropriate size when it was first added, and should be larger without the flag to match the size of the other controls. Same goes for the NearMenu/HandMenu controls.

rickfromwork: I don't really consider it a breaking change since this class was introduced in 5.0

rickfromwork: All that being said, the default size is now much larger for both the HolographicSlate and Near/Hand Menu (since the internal scale-down on the menus has been removed).
 I don't expect developers to use these elements unless designing for an XR experience though, in which case they would want to use the new GUI3DManager flag, so I'm not too concerned with the defaults being very large.

RaananW (approved):

PR #11335, opened 2021-10-25
DeviceInputSystem: Enable passing of multiple event values to onInputChangedObservable

PolygonalSun:
This PR contains changes to DeviceInputSystem and InputManager so that multiple values are passed in a single call when using onInputChangedObservable.  Rather than call notifyObservers for each input value that's changed, we'll call it for each event once or twice (once for up/down, once for any movements, as applicable).

Potential contract change: rather than work with single value for inputIndex, currentState, and previousState, it would now work with arrays of each.


PolygonalSun:
I'm going to close this for the time being and address this after the Babylon Native integration is complete

Issue #11338, opened 2021-10-26
Issue on Alignment for Text


msDestiny14:
Determining if this is causes by the editor, the gui system, or serialization of gui. 

https://forum.babylonjs.com/t/gui-control-has-misaligned-text-after-changing-containers-font-size/24812/3

PR #11359, opened 2021-10-29
Fix for Font offset bug

msDestiny14:
Font offset was not getting recalculated once it was initially rendered. This would be problematic if there was a change to parents font.

https://github.com/BabylonJS/Babylon.js/issues/11338


sebavan (approved):
this._fontOffset || this.isDirty will be done more frequently but as the result is cached in controls it should be all good.

src/Engines/thinEngine.ts
```diff
@@ -5264,39 +5264,4 @@ export class ThinEngine {
         return document;
     }
 
-    /**
-     * Get Font size information
-     * @param font font name
-     * @return an object containing ascent, height and descent
-     */
-    public getFontOffset(font: string): { ascent: number, height: number, descent: number } {
```
RaananW, line 5272: @CedricGuillemet had a reason to move it here, would be great to get his feedback on removing it.
(note - no back-compat issue as it was added after 5.0.0 work started)

gui/src/2D/controls/control.ts
```diff
@@ -2070,6 +2070,11 @@ export class Control {
         }
 
         this._fontOffset = Control._GetFontOffset(this._font);
+
+        //children need to be refreshed
+        this.getDescendants().forEach((child) => {
+            child._markAllAsDirty();
+        });
```
RaananW, line 2077: nit
```suggestion
        this.getDescendants().forEach((child) => child._markAllAsDirty());
```

gui/src/2D/controls/control.ts
```diff
@@ -2196,12 +2201,43 @@ export class Control {
...
+        var fontHeight = 0;
+        try {
+            fontHeight = block.getBoundingClientRect().top - text.getBoundingClientRect().top;
+            block.style.verticalAlign = "baseline";
+            fontAscent = block.getBoundingClientRect().top - text.getBoundingClientRect().top;
+        } finally {
+            document.body.removeChild(div);
+        }
+        var result = { ascent: fontAscent, height: fontHeight, descent: fontHeight - fontAscent };
+        return result;
```
RaananW, line 2238:
```suggestion
        return { ascent: fontAscent, height: fontHeight, descent: fontHeight - fontAscent };
```

RaananW (commented):
nit-picking a little :-)

Waiting on @CedricGuillemet (unless I missed a conversation with him?)

src/Engines/thinEngine.ts
CedricGuillemet: It's been added here for native compatibility with GUI: https://github.com/BabylonJS/Babylon.js/blob/057143356f8261a286b4af7a0a4844257373d27a/src/Engines/nativeEngine.ts#L2853
If GUI controls do not use this function any more, it's good to be removed from thinEngine and nativeEngine.

RaananW: @msDestiny14  - this implementation will fail on native, can you make sure that your function is also dealing with native context as well?

RaananW (changes requested):
Just making sure this is not merged before fixing the native support.

src/Engines/thinEngine.ts
sebavan: Yup would be great if we can deal with it from GUI directly to not add extra code in Thin :-)

msDestiny14: I removed the function from the native side. The function inside of control.ts will now check if IsDocumentAvailable and will return 0's if not. (This was the native implementation)

@bghgary pinging you to keep you in the loop on the native side if you want to take a look.

bghgary: > I removed the function from the native side. The function inside of control.ts will now check if IsDocumentAvailable and will return 0's if not. (This was the native implementation)

I don't know if this will work. The native implementation is a placeholder. I won't guess that @CedricGuillemet intended on implementing it with some native call.

sebavan: In this case it might be moved on engine instead but not thinEngine ?

bghgary: I don't know this code that well, so let's get @CedricGuillemet's opinion first.

RaananW: I guess the main question is - does it need to be in the engine, or can it move back to the GUI and have a different implementation there? I doubt this is the only place where native and web differ, but I don't remember adding these kinds of functionalities to the engine.

CedricGuillemet: AFAIK, GUI implementation is agnostic between Native and web. Implementation differences only exist in Engine. It's possible to move this function in GUI and check if runtime is web or native. This means, a check and different code must exist there. It also means GUI knows about native implementation. Or only web version exists in GUI code and the function is overridden at runtime in native canvas. We can postpone this issue for now by moving the function in GUI, checking it's running on the web and return same object as that https://github.com/BabylonJS/Babylon.js/blob/ae488d1fc8d7599408a7e62ab877f19e32b8a0cb/src/Engines/nativeEngine.ts#L2855 otherwise. Canvas is not under active development for now so it won't be a problem for the weeks/months to come.

bghgary: We need to come up with a plan for platform abstraction. We have similar problems elsewhere and ideally we use the same pattern everywhere.

RaananW: I totally agree. Let's find time and discuss this.
Back to this PR - One of the two solutions will be enough:
1. @bghgary , @CedricGuillemet - is the current implementation enough for the current state in native? 
2. @msDestiny14 - is it possible to keep the current engine implementation and modify it there?

sebavan: Let's at least move it from ThinEngine to Engine :-)

bghgary: The check for document may not work in all cases since we sometimes fake a document to work around issues. Checking for `_native` should work though.

RaananW: So we are keeping it this way, but checking for _native instead of checking if document exists.

sebavan:
@msDestiny14 could you move _getFontOffsetHelper  to Engine (not ThinEngine) so we can merge the fix ?

sebavan (approved):

PR #11363, opened 2021-10-30
Fix native engine pending clearing order

bghgary:
Also, always add native data argument to the pending list since it's possible for object to be removed at any point.

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -660,8 +660,7 @@ class CommandBufferEncoder {
         }
 
         this._isCommandBufferScopeActive = false;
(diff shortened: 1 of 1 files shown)


ryantrem:
Do you know what the overhead is of adding these objects to the array for every operation, as well as clearing the array on submit? Does it impact perf?

ryantrem (approved):

RaananW (approved):

bghgary:
> Do you know what the overhead is of adding these objects to the array for every operation, as well as clearing the array on submit? Does it impact perf?

I don't know, but I was thinking of the same thing. I can measure.

PR #11366, opened 2021-11-01
Incorrectl format number in basis decoder

RaananW:
See https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_s3tc/

### DIFF
--- a/src/Misc/basis.ts
+++ b/src/Misc/basis.ts
@@ -106,7 +106,7 @@ export const BasisToolsOptions = {
  */
 export const GetInternalFormatFromBasisFormat = (basisFormat: number) => {
     // Corresponding internal formats
-    var COMPRESSED_RGB_S3TC_DXT1_EXT = 0x83F1;
+    var COMPRESSED_RGB_S3TC_DXT1_EXT = 0x83F0;
     var COMPRESSED_RGBA_S3TC_DXT5_EXT = 0x83F3;
     var RGB_ETC1_Format = 36196;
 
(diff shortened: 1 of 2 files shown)


src/Misc/basis.ts
```diff
@@ -106,7 +106,7 @@ export const BasisToolsOptions = {
  */
 export const GetInternalFormatFromBasisFormat = (basisFormat: number) => {
     // Corresponding internal formats
-    var COMPRESSED_RGB_S3TC_DXT1_EXT = 0x83F1;
+    var COMPRESSED_RGB_S3TC_DXT1_EXT = 0x83F0;
     var COMPRESSED_RGBA_S3TC_DXT5_EXT = 0x83F3;
```
bghgary, line 110: I think you can get these values from Engine._caps.s3tc?

src/Misc/basis.ts
```diff
@@ -106,7 +106,7 @@ export const BasisToolsOptions = {
  */
 export const GetInternalFormatFromBasisFormat = (basisFormat: number) => {
     // Corresponding internal formats
-    var COMPRESSED_RGB_S3TC_DXT1_EXT = 0x83F1;
+    var COMPRESSED_RGB_S3TC_DXT1_EXT = 0x83F0;
     var COMPRESSED_RGBA_S3TC_DXT5_EXT = 0x83F3;
     var RGB_ETC1_Format = 36196;
```
bghgary, line 111: I think this is engine._caps.etc1.COMPRESSED_RGB_ETC1_WEBGL

PR #11374, opened 2021-11-02
WebGPU: more changes for fast path support + misc fixes

Popov72:
New changes to avoid creating bundles in non compatibility mode.

Also:
* added `setMaterialForRendering` to `EffectLayer`
* improve perf when using the transmission helper
* improve perf of SPS by updating directly the vertex buffers and not calling `updateVerticesData` (which dirtify the material)
* fix occlusion queries not working anymore after the latest changes (`drawWrappers`)
* fix updating the previous matrices for thin instances


src/Engines/engine.ts
```diff
@@ -546,6 +546,8 @@ export class Engine extends ThinEngine {
 
         Engine.Instances.push(this);
 
+        (this.supportRenderPasses as any) = true;
```
sebavan, line 549: Could we avoid the any here ?

src/Engines/nativeEngine.ts
```diff
@@ -749,6 +749,7 @@ export class NativeEngine extends Engine {
 
         this._webGLVersion = 2;
         this.disableUniformBuffers = true;
+        (this.supportRenderPasses as any) = true;
```
sebavan, line 752: Same comment

src/Engines/nullEngine.ts
```diff
@@ -93,6 +93,8 @@ export class NullEngine extends Engine {
 
         Engine.Instances.push(this);
 
+        (this.supportRenderPasses as any) = true;
```
sebavan, line 96: Same, I am worry that a proper es version will prevent overriding readonly.

src/Engines/thinEngine.ts
```diff
@@ -428,6 +428,11 @@ export class ThinEngine {
      */
     public disableVertexArrayObjects = false;
 
+    /**
+     * Indicates that this engine supports render passes
+     */
+    public readonly supportRenderPasses = false;
```
sebavan, line 434: Maybe you only have a public get {} and a protected value ?

src/Engines/webgpuEngine.ts
```diff
@@ -536,6 +536,7 @@ export class WebGPUEngine extends Engine {
 
         (this.isNDCHalfZRange as any) = true;
         (this.hasOriginBottomLeft as any)  = false;
+        (this.supportRenderPasses as any) = true;
```
sebavan, line 539: same comment

src/Layers/effectLayer.ts
```diff
@@ -250,6 +251,10 @@ export abstract class EffectLayer {
      */
     public abstract serialize?(): any;
 
+    protected _numInternalDraws(): number {
```
sebavan, line 254: Could we add a quick comment ?

src/Engines/thinEngine.ts
sebavan: Also does it need to be on ThinEngine or Engine ???

sebavan (commented):
LGTM only a couple neats

src/Engines/engine.ts
Popov72: `supportRenderPasses` is declared as `readonly` as there's no reason for the user to change the value but I needed to create this property to ease the implementation of some other code. I don't know how to change this value by keeping the `readonly` modifier?

src/Engines/thinEngine.ts
Popov72: I need it in `ThinEngine` because in some code (I think the `ParticleSystem`) we can be passed a `ThinEngine` and I need to know if the engine supports render passes or not.

src/Engines/engine.ts
sebavan: maybe it could be in engineFeatures ?

Popov72: Good idea!

sebavan (approved):
LGTM !!!

PR #11380, opened 2021-11-03
ShaderMaterial: add support for storing effect on submeshes

Popov72:
This change is needed for the WebGPU non compatibility mode (fast path) to work.

Regarding `hardBindForSubMesh` (I can change the name of the method, did not really know what to use...), it's a bug fix.

If two meshes are displayed one after the other and they both have the same material, `mustRebind()` will return `false`. The problem is that the sub surface module is binding `vThicknessParam` which depends on the submesh. So I have added `hardBindForSubMesh` which is always called, independent from the `mustRebind()` value.


src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1943,7 +1943,10 @@ export abstract class PBRBaseMaterial extends PushMaterial {
         mesh.getMeshUniformBuffer().bindToEffect(effect, "Mesh");
         mesh.transferToEffect(world);
 
-        // PrePass
+        var engine = scene.getEngine();
```
sebavan, line 1946: nit: const

src/Materials/shaderMaterial.ts
```diff
@@ -147,9 +147,10 @@ export class ShaderMaterial extends Material {
      *  * object: { vertexSource: "vertex shader code string", fragmentSource: "fragment shader code string" } using with strings containing the shaders code
      *  * string: "./COMMON_NAME", used with external files COMMON_NAME.vertex.fx and COMMON_NAME.fragment.fx in index.html folder.
      * @param options Define the options used to create the shader
+     * @param storeEffectOnSubMeshes true to store effect on submeshes, false to store the effect directly in the material class.
      */
-    constructor(name: string, scene: Scene, shaderPath: any, options: Partial<IShaderMaterialOptions> = {}) {
-        super(name, scene);
+    constructor(name: string, scene: Scene, shaderPath: any, options: Partial<IShaderMaterialOptions> = {}, storeEffectOnSubMeshes = true) {
```
sebavan, line 152: Would it work if storeEffectOnSubMeshes is false with a pushMaterial ??? I always thought a push mat would need to store the effect on a submesh ?

src/Materials/shaderMaterial.ts
```diff
@@ -780,10 +789,13 @@ export class ShaderMaterial extends Material {
             shaderName = this.customShaderNameResolve(shaderName, uniforms, uniformBuffers, samplers, defines, attribs);
         }
 
-        var previousEffect = effect;
+        const drawWrapper = storeEffectOnSubMeshes ? subMesh._getDrawWrapper() : this._drawWrapper;
+        var previousEffect = drawWrapper?.effect ?? null;
```
sebavan, line 793: Nit: const

src/Materials/shaderMaterial.ts
```diff
@@ -780,10 +789,13 @@ export class ShaderMaterial extends Material {
             shaderName = this.customShaderNameResolve(shaderName, uniforms, uniformBuffers, samplers, defines, attribs);
         }
 
-        var previousEffect = effect;
+        const drawWrapper = storeEffectOnSubMeshes ? subMesh._getDrawWrapper() : this._drawWrapper;
+        var previousEffect = drawWrapper?.effect ?? null;
+        var previousDefines = drawWrapper?.defines ?? null;
```
sebavan, line 794: Nit: const

src/Materials/shaderMaterial.ts
```diff
@@ -1432,7 +1448,7 @@ export class ShaderMaterial extends Material {
      * @returns a new material
      */
     public static Parse(source: any, scene: Scene, rootUrl: string): ShaderMaterial {
-        var material = SerializationHelper.Parse(() => new ShaderMaterial(source.name, scene, source.shaderPath, source.options), source, scene, rootUrl);
+        var material = SerializationHelper.Parse(() => new ShaderMaterial(source.name, scene, source.shaderPath, source.options, source.storeEffectOnSubMeshes), source, scene, rootUrl);
```
sebavan, line 1451: Nit: const

sebavan: also would forcing storeEffectOnSubMesh to false works in the default case ?

sebavan (commented):
LGTM only a couple comments

src/Materials/shaderMaterial.ts
Popov72: Yes, a `ShaderMaterial` can work in both mode (that's the point of the PR actually), that's why it's a parameter of the constructor. The other materials (standard, PBR, etc) would probably not work but you can't change the value (`_storeEffectOnSubMeshes` is always `true`) as it is not exposed.

src/Materials/shaderMaterial.ts
Popov72: I'm not sure I understand.

`source.storeEffectOnSubMeshes` can be `undefined` if parsing a material that has been serialized before this PR, in which case the `storeEffectOnSubMeshes` parameter will take the default value defined in the constructor, which is `true`. That is the value we should set by default I think, else people writing code for WebGPU and using the non compat mode will experience lower perf when using `ShaderMaterial`. I think a lot of people won't be aware of it (even if we document it and explain what to do to fix it), so in my opinion it's better to set the value so that the perf are the best out of the box.

sebavan: Yup totally agree I was mostly wondering in case they were using in a way which would require false which would then be none back compat if it could exist ?

Popov72: Yes, it can be a breaking change in some cases: I have added some explanations in the "Breaking Changes" section of the what's new file. I expect that this will only affect a few people, though.

sebavan (approved):

sebavan:
@Popov72 , Looks like this breaks on Webgl1 :-)

PR #11381, opened 2021-11-03
Forward Tools.*Performance* calls to Babylon Native

ryantrem:
With this change calls to `Tools.PerformanceLoggingLevel`, `Tools.StartPerformanceCounter`, and `Tools.EndPerformanceCounter` are forwarded to Babylon Native as well. This accomplishes two things:
1. It lets us enable Babylon Native performance logging from JS
2. It lets us see what is happening in Babylon.js in the context of native performance profiles

For example, in the screenshot below (Xcode Instruments), the "Points of Interest" shows two native perf intervals (the top two), and a JS interval (the bottom one - "Scene render").

image

I also added an unrelated change to this PR, which is to add overloads for `Node.getChildren` (and other related `Node` functions) such that if the predicate you pass in is a type guard, the returned array will be strongly typed according to that type guard. For example:

```ts
// transformNodes will be a TransformNode[] rather than a Node[]
const transformNodes = node.getDescendants(false, (node): node is TransformNode => node instanceof TransformNode);
```

### DIFF
--- a/src/Misc/tools.ts
+++ b/src/Misc/tools.ts
@@ -20,6 +20,8 @@ import { SliceTools } from "./sliceTools";
 declare type Camera = import("../Cameras/camera").Camera;
 declare type Engine = import("../Engines/engine").Engine;
 
+declare const _native: any;
+
 interface IColor4Like {
     r: float;
     g: float;
@@ -1046,6 +1048,7 @@ export class Tools {
     public static readonly PerformanceConsoleLogLevel = 2;
 
     private static _performance: Performance;
+    private static _nativeCounters: Map<string, unknown>;
 
     /**
      * Sets the current performance log level
@@ -1065,6 +1068,19 @@ export class Tools {
 
         Tools.StartPerformanceCounter = Tools._StartPerformanceCounterDisabled;
         Tools.EndPerformanceCounter = Tools._EndPerformanceCounterDisabled;
+
+        let nativeEnablePerformanceLogging: (() => void) | undefined;
+        let nativeDisablePerformanceLogging: (() => void) | undefined;
+        if (typeof _native !== 'undefined') {
+            nativeEnablePerformanceLogging = _native.enablePerformanceLogging;
+            nativeDisablePerformanceLogging = _native.disablePerformanceLogging;
+
+            if (level === Tools.PerformanceNoneLogLevel && nativeDisablePerformanceLogging) {
+                nativeDisablePerformanceLogging();
+            } else if ((level & Tools.PerformanceUserMarkLogLevel) === Tools.PerformanceUserMarkLogLevel && nativeEnablePerformanceLogging) {
+                nativeEnablePerformanceLogging();
+            }
+        }
     }
 
     private static _StartPerformanceCounterDisabled(counterName: string, condition?: boolean): void { }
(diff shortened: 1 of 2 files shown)


src/Misc/tools.ts
```diff
@@ -20,6 +20,8 @@ import { SliceTools } from "./sliceTools";
 declare type Camera = import("../Cameras/camera").Camera;
 declare type Engine = import("../Engines/engine").Engine;
 
+declare const _native: any;
```
bghgary, line 23: I added an interface for this. Please use it and add your methods to the interface.

```suggestion
declare const _native: INative;
```

src/Misc/tools.ts
```diff
@@ -1046,11 +1048,33 @@ export class Tools {
...
+                    if (nativeDisablePerformanceLogging) {
+                        nativeDisablePerformanceLogging();
+                    }
+                } else {
+                    if (nativeEnablePerformanceLogging) {
+                        nativeEnablePerformanceLogging();
+                    }
+                }
+            }
+        }
```
bghgary, line 1076: I was thinking you would override this function in the constructor of nativeEngine.ts. Does that work?

bghgary (changes requested):

src/Misc/tools.ts
sebavan: Would be great I was about to propose the same :-)

src/node.ts
```diff
@@ -610,6 +626,22 @@ export class Node implements IBehaviorAware<Node> {
         return results;
     }
 
+    /**
+     * Get all child-meshes of this node
+     * @param directDescendantsOnly defines if true only direct descendants of 'this' will be considered, if false direct and also indirect (children of children, an so on in a recursive manner) descendants of 'this' will be considered (Default: false)
+     * @param predicate defines an optional predicate that will be called on every evaluated child, the predicate must return true for a given child to be part of the result, otherwise it will be ignored
+     * @returns an array of AbstractMesh
+     */
+    public getChildMeshes<T extends AbstractMesh>(directDescendantsOnly?: boolean, predicate?: (node: Node) => node is T): T[];
```
sebavan, line 635: I am really wondering if I could not use this in the wrong way let s say I call it with: 

```
getChildMeshes<Mesh>(false, (am) => true);
```

TS won t be able to detect I am not using a typeguard here and instancedMeshes would get mixed in return as Mesh type. I am really wondering if the overload is safe enough ?

sebavan: Should actually be totally fine in the latest TS version which is awesome :-) 

https://www.typescriptlang.org/play?#code/MYGwhgzhAEBCkFMByB7AJg6BvAUNf0ADgK4BGIAlsNAE4JhooB2IAntMABYUhp1MAuOIlQYA2gF1oAXmiSA3DjwES5KtADmCAC4BhbrwCyCCJxMAeACrQEAD20ImaGAEFSEbTTDBtx0wD4ACkI6NCowByFApnQEIXgIZFiAShl-aBiMaAoYS2ShSwUlAiIySmotPQM0PzMIYNDwyOho2PiRFLToUhQUEHomfOg3Dy8fWqLlfFVyzR19HhqTOoaEMOAIuJbMrYSkjFTpdJ6+gaGRz29fZclsKZKOZg9aE2IQbQghFxovVnMLsbXAIyOQSRQPAgAMxQNBawCe2g41WgKEh0G03AgADouIt+KlcBCIRQ0YFArjeNBIMN3JdxstkliAPpgdwAZUICGAFDAIAAYhQEJSAGTCoiNDYOcnVZIE+5Eh50CBvD5YkimaWLKmuWmA2rJcEKggAX3lJrN+Do2mINCYL2V7wghvwptNOFAkB1oyutRs9kczmEiVEmEJKjK6joDGYbGgLPZnO5vIFQrQIM8xAQijdHqg0F9dgcTi9dKBnDuJRmkfojBY7CZAFtlhyuTz+YLKbIM1mcDnwHmAJJMDxgJjANYF-3Fmne+mmCvhtTUKO12NMihN0wtpPt1PpmiZ7NKeHDxENkFMBAAd3zy0CBvdCOy59kl5vQ5HY4nd4fj9P3TtV9ryDfYEHvRRSCYHFqn4NViA1BsH0g6C8UcOCNQ3X8T2eJUVRBZDKgWIxlgsWogkCJlDnSbtfyAA

ryantrem: TS is able to detect you are not using a type guard and gives you an error:
`Signature '(node: Node): boolean' must be a type predicate.(2345)`
This is exactly the same pattern that is used for `Array.filter`.

sebavan:
@ryantrem should this be moved from draft now the hooks are available in native ?

ryantrem:
> @ryantrem should this be moved from draft now the hooks are available in native ?

No, I’m the end I just implemented this functionality directly in Babylon React Native. This is not great because it is not available to regular Babylon Native, but unfortunately I don’t have time right now to solve the larger problem of a general pattern for native only functionality in BJS that has zero impact on code size after tree shaking.

PR #11383, opened 2021-11-04
StructuredTextBlock (GUI 2D)

cronvel:
Following [this forum post](https://forum.babylonjs.com/t/multiple-colors-styles-in-one-textblock/24680/18), here is the PR with the *Structured Text* implementation for `TextBlock` (GUI 2D).
This is a working implementation but should be considered as a base to iterate over.

**TLDR:** this image This image was produced with a single `TextBlock` with word-wrapping and mixing color, bold, italic, underline and line-through.

**API:** instead of doing `textBlock.text = "My text"`...
... one would do `textBlock.structuredText = [ { text: "My text " } , { text: "with bold" , fontWeight: "bold" } ]`.
This is the simplest structure.

### Currently supported attributes:

* color (string | ICanvasGradient)
* underline (boolean)
* lineThrough (boolean)
* fontStyle (string)
* fontWeight (string)
* outlineWidth (number)
* outlineColor (string)
* shadowColor (string)
* shadowBlur (number)
* shadowOffsetX (number)
* shadowOffsetY (number)

### Few notes:

1. For instance, I coded it so it doesn't change the older `TextBlock` code. So there is 2 codepath ATM. BTW I think the non-structured-text code should be completely removed, except for the `.text` API.
2. Also I noted that the `TextBlock` is often fully recomputed for no reason.
E.g. simply opening dev-tools (which resize the screen but not the size of the `TextBlock`) have to compute word-wrapping again (which make a lot of `context.measureText()` calls).
Maybe it would be better to have more level of dirtiness, e.g. have a .`_innerContentIsDirty` property, and only recompute lines when the text have changed (or the `TextBlock` bounding box size).
3. I would also advocate for some alternate properties for structured text input, e.g. `bold: true` would convert to `fontWeight: bold`, for instance I've tried to keep the API close to the existing API and internal stuff. But I think it would be better to feed the `TextBlock` with more intuitive attribute.
4. The current code of this PR is not perfect with the `context.save()` and `context.restore()` things.
5. Font-size and font-family are not supported, and don't know if it makes sense, and it make things more complicated (e.g.: computing line-height. It's not that hard, but I don't think it's worth the trouble, it looks like a *niche* use-case to me)

It's my first time writing TypeScript, so maybe there are few things off.


cronvel:
@msDestiny14 Hello! Can you have a look, please?

msDestiny14:
Hey looks like you got some building errors.

Couple of comments and suggestions: 
1 like the last feature code PR  we just had for GUI I'm going to take my time to slowly review this and of course test. 
**2. If you have a playground example you can create with the new feature that would be fantastic!**

cronvel:
@msDestiny14

> Hey looks like you got some building errors.

That's strange, when I created the PR everything was green except 2 tests that were **pending**.
Also I don't have published build, just the `.ts` file to avoid polluting the diff. Should I publish the build too?

> Couple of comments and suggestions: 1 like the last feature code PR we just had for GUI I'm going to take my time to slowly review this and of course test. **2. If you have a playground example you can create with the new feature that would be fantastic!**

Can you tell me how to create a playground using my PR, I'm a bit confused! ;)

msDestiny14:
> @msDestiny14
> 
> > Hey looks like you got some building errors.
> 
> That's strange, when I created the PR everything was green except 2 tests that were **pending**. Also I don't have published build, just the `.ts` file to avoid polluting the diff. Should I publish the build too?
> 
> > Couple of comments and suggestions: 1 like the last feature code PR we just had for GUI I'm going to take my time to slowly review this and of course test. **2. If you have a playground example you can create with the new feature that would be fantastic!**
> 
> Can you tell me how to create a playground using my PR, I'm a bit confused! ;)

YA. you can create a index.js file to be able to test with Local debug mode. I'm assuming this is how you tested it yourself. What ever tests you used to validate just send an example so we can get a good feel on how it will be used in a end user scenario. 👍

gui/src/2D/controls/control.ts
```diff
@@ -37,19 +37,27 @@ export class Control {
     public parent: Nullable<Container>;
     /** @hidden */
     public _currentMeasure = Measure.Empty();
-    private _fontFamily = "Arial";
-    private _fontStyle = "";
-    private _fontWeight = "";
-    private _fontSize = new ValueAndUnit(18, ValueAndUnit.UNITMODE_PIXEL, false);
-    private _font: string;
+
+    //++CR temp: changing access from private to protected
```
deltakosh, line 41: Comments looks suspicious

gui/src/2D/controls/control.ts
```diff
@@ -1085,21 +1093,6 @@ export class Control {
         this._disabledColorItem = value;
         this._markAsDirty();
     }
-
-    /**
-     * Gets/sets the overlap group of the control.
```
deltakosh, line 1090: This should not be removed

deltakosh:
I like that PR a lot but I would recommend to create a new class StructuredTextBlock that inherits from TextBlock to avoid complexifing the TextBlock class and help with treeshacking

cronvel:
@deltakosh Thanks for your feedback.
I could do it on another class if you will, but note that the old class would be completely obsoleted by the new one, since the `StructuredTextBlock` would be a superset of the `TextBlock`. Is it OK?

Also I would like to add a *slow-typing* effect (typical effect found in video game dialog like RPG/JRPG, or Visual Novel, etc), this would mean code it twice or just code it for `StructuredTextBlock` class.

For `control.ts`, it seems I've done something wrong with `git`, in fact I have just changed some properties from `private` to `protected` (`textBlock.ts` needs access to those properties), I probably reverted few changes made during this week with a wrong *rebase* command.

msDestiny14:
> I like that PR a lot but I would recommend to create a new class StructuredTextBlock that inherits from TextBlock to avoid complexifing the TextBlock class and help with treeshacking

Now that I think about it more...taking this into account would make adding GUI editor support easier.

deltakosh:
> I could do it on another class if you will, but note that the old class would be completely obsoleted by the new one, since the StructuredTextBlock would be a superset of the TextBlock. Is it OK?

Not sure to understand. The textblock by itself is already functional right? and people may not be interested by all the added code (regarding perf or package size)

Regarding slow typing, I do not think this is something we want in the framework. I'm afraid it start to be bloated. This is definitely something we expect to happen on the developer side.

gui/src/2D/controls/control.ts
```diff
@@ -37,19 +37,19 @@ export class Control {
     public parent: Nullable<Container>;
     /** @hidden */
     public _currentMeasure = Measure.Empty();
-    private _fontFamily = "Arial";
-    private _fontStyle = "";
-    private _fontWeight = "";
-    private _fontSize = new ValueAndUnit(18, ValueAndUnit.UNITMODE_PIXEL, false);
-    private _font: string;
+    protected _fontFamily = "Arial";
```
sebavan, line 40: I am afraid this would bypass some of the font cache mecanism in place , is it intented ?

sebavan: same for all the protected changes

gui/src/2D/controls/textBlock.ts
```diff
@@ -5,9 +5,66 @@ import { Control } from "./control";
 import { RegisterClass } from "babylonjs/Misc/typeStore";
 import { Nullable } from "babylonjs/types";
 import { serialize } from 'babylonjs/Misc/decorators';
-import { ICanvasRenderingContext } from 'babylonjs/Engines/ICanvas';
+import { ICanvasRenderingContext , ICanvasGradient } from 'babylonjs/Engines/ICanvas';
 import { Engine } from 'babylonjs/Engines/engine';
 
+
+
+interface StructuredTextPart {
```
sebavan, line 13: this should be exported as well as anything used publicly

sebavan (changes requested):
This should be a different control to not over complexify the TextBlock ones. Might be inheritence if that helps but it should not be mixed in here for size and simplicity.

Most of the `var` should be let or const and comments sounds weird sometimes like auto generated.

I ll review in more details once it is a separate control

deltakosh:
I also wonder if we could use MD syntax ?

sebavan:
> I also wonder if we could use MD syntax ?

Would be amazing !!!

cronvel:
@deltakosh @sebavan 

> > I could do it on another class if you will, but note that the old class would be completely obsoleted by the new one, since the StructuredTextBlock would be a superset of the TextBlock. Is it OK?
> 
> Not sure to understand. The textblock by itself is already functional right? and people may not be interested by all the added code (regarding perf or package size)

About perf/package size/tree-shaking, it would not change much. For instance I have voluntarily kept 2 codepath (*structured* and *non-structured text*). This is temporary and solely because I wanted to get in touch with the team first. 

IMHO I would simply remove all *non-structured text* codepath (keep backward compatibility thanks to the`.text` getter/setter, setting `.text = "some text"` would be internally converted to `.structuredText = [{text:"some text"}]`), after doing that, the size of the file would be almost the same than the current one on *master* branch, maybe just ~100 more lines.

Also I would not insist if you prefer it to have its own class, just tell me.

> Regarding slow typing, I do not think this is something we want in the framework. I'm afraid it start to be bloated. This is definitely something we expect to happen on the developer side.

##### I would like to respectfully advocate for a framework integration:

* this would not be a big dev
* I think it's quite common for game to have such feature
* it's not possible to do that properly and efficiently on userland-side, e.g. simply updating the text **would not play well with word-wrap** (words at the end of line would start displaying at the right, then suddenly jump on the new line), and about the perf, it would require to compute lines (which need a lot of `context.measureText()`) again and again, everytime the user change something

Also I had in mind a very **simple and unopinionated implementation**:
* Add a property `.displayOnlyFirstNChars` (or whatever the name), when set to a number the `StructuredTextBlock` render function would simply stop after writing *n* characters.
* Updating `.displayOnlyFirstNChars` does not compute again the lines, in fact the lines parsing is totally unaware of this property.
* **No timer on the framework-side**, let the userland-side update `.displayOnlyFirstNChars` on its own. This is fine because this part is also the *opinionated* one. E.g. one may want to have a tiny pause after writing a letter, a moderate pause after writing a space, a long pause after writing a punctuation. This level of detail should not be part of the framework and could be easily implemented by developers.
* Have a `.getLastDisplayedChar()` utility method (or whatever the name) that return the last char (i.e. the *Nth* char where *n* is the number stored in the `.displayOnlyFirstNChars` property), this method help implementing the example given above, featuring variable pause.

I think it's super lightweight.

> I also wonder if we could use MD syntax ?

Could be really great, but this is on another level of complexity! It's out of the scope of this PR!
Just by looking at the current Github page, we can see than managing quotes, or highlighting codeblock needs extra-drawing.

Now that we are at it, adding some kind of hypertext-like features would be awesome too: some words in the text acting like a button.

cronvel:
@msDestiny14 @sebavan @deltakosh Sorry for the noob question, having only 1 day of TypeScript experience, but what linter are you using? (and where is its config file?)

sebavan:
About the linting part, you can find the issues by following the github links https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=16408&view=logs&j=32e2e1bb-a28f-5b18-6cfc-3f01273f5609&t=2bf39655-ae43-5ae7-fe00-6e76c9bf09b6

The config is here: https://github.com/BabylonJS/Babylon.js/blob/master/tslint.json and there is a gulp command to launch it locally: `gulp tsLint`

I would keep it a separate class mainly hoping in the future we could rely on https://github.com/WICG/canvas-formatted-text.

About the syntax, the thing is if we do not want to support different syntaxes for ever due to our back compat rules. we need to have our final one in as soon as possible. Maybe you could expose your data structure instead in this PR without parser and introduce a md parser in a subsequent PR.

gui/src/2D/controls/textBlock.ts
```diff
@@ -381,6 +468,124 @@ export class TextBlock extends Control {
         }
     }
 
+    //++CR
+    private _drawStructuredText(structuredText: StructuredText, textWidth: number, y: number, context: ICanvasRenderingContext): void {
+        var attr ,
+            x = 0 ,
+            width = this._currentMeasure.width ;
```
aWeirdo, line 475: Pattern using commas repeats throughout the code.
Variables should be declared each on their own line with let & const

>let x = 0;
const width = this._currentMeasure.width;

attr should be declared where it's used, e.g. line 498

deltakosh:
> Also I would not insist if you prefer it to have its own class, just tell me.
Yes please :)

deltakosh:
Regarding the slow text typing, I'm sorry but this is not a general ask enough. I would rather prefer to see a PR that makes it possible to be done in userland.

Please don't get me wrong here as I do not want you to be mad but we are the ultimate owner long term and we will maintain the code forever while keeping it backward compatible so we cannot "bloat" the code with too specific features

gui/src/2D/controls/textBlock.ts
deltakosh, line 13: Please make it a separated file as well

deltakosh:
> @msDestiny14 @sebavan @deltakosh Sorry for the noob question, having only 1 day of TypeScript experience, but what linter are you using? (and where is its config file?)

No worries at all :) we use tsling and the config is already in the repo

cronvel:
@msDestiny14 @deltakosh @sebavan @aWeirdo Hi there, I updated the code, creating the new class `StructuredTextBlock`, and taking into consideration your feedbacks.

The playground code to test it: [here](https://playground.babylonjs.com/#I4DDQH).

gui/src/2D/controls/structuredTextBlock.ts
```diff
@@ -0,0 +1,779 @@
...
+import { ValueAndUnit } from "../valueAndUnit";
+import { Control } from "./control";
+import { TextWrapping } from "./textBlock";
+import { RegisterClass } from "babylonjs/Misc/typeStore";
+import { Nullable } from "babylonjs/types";
+import { serialize } from 'babylonjs/Misc/decorators';
+import { ICanvasRenderingContext , ICanvasGradient } from 'babylonjs/Engines/ICanvas';
+import { Engine } from 'babylonjs/Engines/engine';
+
+export interface StructuredTextPart {
```
deltakosh, line 12: Cab you extract these interfaces into separate files?

cronvel: Ok, where should I put it? `gui/src/2D/controls` or `gui/src/2D`?

cronvel: Also it seems that you are prefixing interfaces with a `I`, so should I rename them to `IStructuredText` and `IStructuredTextPart` and name the file `iStructuredText.ts`?

aWeirdo: `I` prefix & putting them directly in the `gui/src/2D/controls` folder should be ok :)

deltakosh: that was what I tried to say :)

deltakosh (changes requested):
a few changes

cronvel:
Update:
* Created a separated file for the public interface
* Code clean-up
* Add missing getters/setters
* More flexibility for decoration (underline, line-through): customizable Y-offset and thickness (on the *control* level, not per text-parts)
* Support for *framed* style (see screenshot below), e.g. it could be used to render markdown backtick. The color, the outline color, outline thickness and corner radius of the rectangular frame are all configurable
* Optimization (avoid recomputing lines when not needed)

Screenshot
Code producing the screenshot: [PG](https://playground.babylonjs.com/#I4DDQH#1).

I also included in the update the `characterLimit` property, *just to demonstrate how trivial it is* (**only 3 extra-lines of code** + getter/setter).
I can help maintaining the whole structured text thing, if it helps the adoption of that feature ;)
FYI I'm the author of a terminal-based UI lib with [35k+ download per week (npm)](https://www.npmjs.com/package/terminal-kit) and [2.3k stars (Github)](https://github.com/cronvel/terminal-kit), so I'm an open-source veteran! ;)

About the `TextBlock`'s inheritance, there isn't much to inherit from, only one method is preserved from the `TextBlock` class and it may change too. It would get in the way more than it helps. Also there is no property `.text` on `StructuredTextBlock`. It would just create more confusion than it solves, especially if more feature are added in the future (Markdown parser, hypertext-like button/hover/tooltip, etc).

gui/src/2D/controls/control.ts
deltakosh: up @cronvel

gui/src/2D/controls/structuredTextBlock.ts
```diff
@@ -0,0 +1,993 @@
...
+    parts: StructuredText;
+
+    // Computed line width
+    width?: number;
+};
+type StructuredTextLines = Array<StructuredTextLine>;
+
+// Mostly the same than IStructuredTextPart, but nothing is optional here,
+// this is the attributes about to be sent to the canvas context.
+interface TextPartAttributes {
```
deltakosh, line 25: That should be a separated file as well ;)

gui/src/2D/controls/structuredTextBlock.ts
```diff
@@ -0,0 +1,779 @@
...
+  shadowBlur: number;
+  shadowOffsetX: number;
+  shadowOffsetY: number;
+}
+
+/**
+ * Class used to create structured text block control
+ */
+//export class StructuredTextBlock extends TextBlock {
+export class StructuredTextBlock extends Control {
```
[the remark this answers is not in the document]
deltakosh: up @cronvel

deltakosh:
I cc you on some remaining comments
Also could you please add your PG as a validation test? that will be super useful

gui/src/2D/controls/structuredTextBlock.ts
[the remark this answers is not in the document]
cronvel: @deltakosh 
> About the `TextBlock`'s inheritance, there isn't much to inherit from, only one method is preserved from the `TextBlock` class and it may change too. It would get in the way more than it helps. Also there is no property `.text` on `StructuredTextBlock`. It would just create more confusion than it solves, especially if more feature are added in the future (Markdown parser, hypertext-like button/hover/tooltip, etc).

gui/src/2D/controls/structuredTextBlock.ts
cronvel: @deltakosh Is it really useful since this interface is only used internally?

Also maybe I should have used the `type` keyword instead of `interface` to better express this? There are still some subtleties between *type alias* and *interface* that I don't understand very well and [stackoverflow doesn't helped me much on that matter](https://stackoverflow.com/questions/37233735/interfaces-vs-types-in-typescript).

cronvel:
@deltakosh 
> Also could you please add your PG as a validation test? that will be super useful
The PG code to test it is [here](https://playground.babylonjs.com/#I4DDQH#1), I don't know if there is a better place to put it.

I could put it here too:
```js
    // Basic setup
    light.intensity = 0.7;
    var sphere = BABYLON.MeshBuilder.CreateSphere("sphere", {diameter: 2, segments: 32}, scene);
    sphere.position.y = 1;
    var ground = BABYLON.MeshBuilder.CreateGround("ground", {width: 6, height: 6}, scene);

    // GUI
    var advancedTexture = BABYLON.GUI.AdvancedDynamicTexture.CreateFullscreenUI("UI");

    var rectangle = new BABYLON.GUI.Rectangle() ;
    rectangle.width = 0.5 ;
    rectangle.height = "160px" ;
    rectangle.thickness = 1 ;
    rectangle.cornerRadius = 5 ;
    rectangle.color = "black" ;
    rectangle.background = "#55aa55" ;
    advancedTexture.addControl( rectangle ) ;

    // Now the structured text!
    var structuredTextBlock = new BABYLON.GUI.StructuredTextBlock() ;
    structuredTextBlock.structuredText = [
        { text: "Lorem ipsum " },
        { text: "Lorem ipsum", color: "red", frame: true, 
            //frameColor: "#aa77cc", frameOutlineWidth: 2, frameOutlineColor: "#660088", frameCornerRadius: 5
        },
        { text: " " },
        { text: "lorem ipsum dolor", color: "#eeaa55", fontStyle: "italic", underline: true },
        { text: " " },
        { text: "lorem ipsum", color: "red", lineThrough: true },
        { text: " lorem ipsum dolor sit amet ", fontStyle: "italic", fontWeight: "bold" },
        { text: "lorem ipsum dolor sit amet, consectetur adipiscing elit", color: "#aa33cc", fontStyle: "italic", fontWeight: "bold", underline: true },
    ] ;

    structuredTextBlock.fontSize = "24px" ;
    structuredTextBlock.color = "white" ;
    structuredTextBlock.outlineWidth = 2 ;
    structuredTextBlock.outlineColor = "black" ;
    //structuredTextBlock.underlineRelativeY = 0.2;
    //structuredTextBlock.lineThroughRelativeY = -0.3;
    //structuredTextBlock.decorationRelativeThickness = 0.04;
    
    /*
    structuredTextBlock.frameColor = "#eea0a0";
    structuredTextBlock.frameOutlineWidth = 2;
    structuredTextBlock.frameOutlineColor = "#aa6060";
    //*/
    structuredTextBlock.frameCornerRadius = 5;
    structuredTextBlock.paddingLeft = structuredTextBlock.paddingRight = structuredTextBlock.paddingTop = structuredTextBlock.paddingBottom = "5px" ;
    structuredTextBlock.characterLimit = 0;

    //structuredTextBlock.textWrapping = BABYLON.GUI.TextWrapping.Clip ;
    //structuredTextBlock.textWrapping = BABYLON.GUI.TextWrapping.Ellipsis ;
    structuredTextBlock.textWrapping = BABYLON.GUI.TextWrapping.WordWrap ;

    //structuredTextBlock.resizeToFit = true ;
    rectangle.addControl( structuredTextBlock ) ;

    setInterval( () => structuredTextBlock.characterLimit += 1 , 20)

    return scene;
};
```

deltakosh:
For the visual tests: 
- Add it here: https://github.com/BabylonJS/Babylon.js/blob/master/tests/validation/config.json
- And drop the reference image there: https://github.com/BabylonJS/Babylon.js/tree/master/tests/validation/ReferenceImages

gui/src/2D/controls/structuredTextBlock.ts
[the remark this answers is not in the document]
deltakosh: I understand your point. That being said it will help users as they could have a function taking a TextBlock that cn also work with StructuredTExtBlock

Does it make sense?

cronvel:
@deltakosh @msDestiny14

> For the visual tests:
> 
>     * Add it here: https://github.com/BabylonJS/Babylon.js/blob/master/tests/validation/config.json
> 
>     * And drop the reference image there: https://github.com/BabylonJS/Babylon.js/tree/master/tests/validation/ReferenceImages
Sorry for yet another noob question, but is there a hidden button in the playground to generate the image with the correct size? Or should I painfully resize my browser's window until I got a 600x400 webgl rendering viewport to screenshot? ^^

sebavan:
> @deltakosh @msDestiny14
> 
> > For the visual tests:
> > ```
> > * Add it here: https://github.com/BabylonJS/Babylon.js/blob/master/tests/validation/config.json
> > 
> > * And drop the reference image there: https://github.com/BabylonJS/Babylon.js/tree/master/tests/validation/ReferenceImages
> > ```
> 
> Sorry for yet another noob question, but is there a hidden button in the playground to generate the image with the correct size? Or should I painfully resize my browser's window until I got a 600x400 webgl rendering viewport to screenshot? ^^

You can do it locally by first placing a dummy png (copy paste any of the existing one) in the reference images folder. Then launch the validation tests on this test: http://localhost:1338/tests/validation/?list=config&engine=webgl2&test=Prepass%20SSAO%20+%20particles Rename the test to yours.

Right click the generated image, save as... And done :-)

cronvel:
@sebavan Thanks!
Also I was not aware that the local playground was saving code to the official playground, it's very handy!
The visual test is added now!

sebavan:
@msDestiny14 could you do a quick pass of review ? I ll do mine once all good on your side

cronvel:
@sebavan @msDestiny14 I'm not exactly happy with how the business logic coming from the original `TextBlock` works. So I will do some code refactoring, so don't review now! ;)

I want to compute more things ahead of time, and prepare for future features.
Basically, when parsing the lines, I will store the exact coordinate of each part of text, so when rendering there will be nothing to compute, just draw.
Storing coordinate at line parsing could be useful for many things, e.g. for future hypertext-like features, or for advanced special effect/animation on part of text, and so on.

cronvel:
So, having moved positional computing ahead of time instead of at drawing time make it possible for developers to code nice effects with only few lines of code:
gif
To do it, just add a listener to the `.onLinesReadyObservable`, find the parts of text to animate, and with a timer move it the way you like!

cronvel:
News:
* Support for hover effect
* Support for clickable text part (hyperlink-like)

deltakosh:
We still have some unresolved comments:
image

gui/src/2D/controls/control.ts
cronvel: Ok it's now back to what it used to be on last push.

gui/src/2D/controls/structuredTextBlock.ts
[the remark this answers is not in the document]
cronvel: @deltakosh The problem is that there are already few incompatibilities between both classes, and it can only diverge more in the future.
There is even some **public properties that just doesn't exist anymore** on StructuredTextBlock (e.g. there is no `.text`).

cronvel:
@deltakosh @sebavan @msDestiny14 Hi there!

I pushed a new version, and I think we are now close to the definitive version for this PR.
This new control is really flexible, and I carefully made it so it's *not* bloated. It allows devs to make really cool things.

I also made it with **Markdown** (or BBCode or HTML subset) **in mind**.
Most basic Markdown features could be done (italic, bold, headers, inline backticks, and even links).
At the moment it lacks block-like features (decorations spanning over multiple lines, like block quote or block code), and image support.

Blocks could be easy to implement later, while image would require some asyncness tricks (because we need the size of the image about to be loaded, or need to compute again the lines once an image is loaded).

Also, I would like some feedbacks on those API names:
```js
    /**
     * An event triggered when there is a part having an href clicked
     */    public onClickHrefObservable = new Observable<HrefObservableData>();

    /**
     * An event triggered when the mouse enter a part having an href
     */
    public onEnterHrefObservable = new Observable<HrefObservableData>();

    /**
     * An event triggered when the mouse move out of a part having an href
     */
    public onOutHrefObservable = new Observable<HrefObservableData>();
```
**CONTEXT:** Whenever a text part has a `.href` property (and if the control is a pointer blocker), those events are fired respectively when the mouse click/enter/exit that text part. Hence the *Href* in the observer name. Also I'm not sure the names are 100% explicit, or if we should switch *Href* and the pointer even name (e.g. onHrefClickObservable onClickHrefObservable).
I thought about *Anchor* instead of *Href* to mimic HTML, but I don't like it much either, and found that *Href* is more straightforward due to the fact that that's the actual property name that trigger it.

Finally, I hope you will be happy with that brand new control, because I have spent way more time that is was reasonable to spend, considering my personal schedule! ;)

gui/src/2D/controls/structuredTextBlock.ts
```diff
@@ -0,0 +1,1295 @@
...
+    /**
+     * Function used to split a string into words. By default, a string is split at each space character found
+     */
+    public wordSplittingFunction: Nullable<(line: string) => string[]>;
+
+    /**
+    * Gets or sets structured text to display
+    */
+    @serialize()
+    public get structuredText(): StructuredText {
```
deltakosh, line 146: Why not calling it text? And thus it will be compatible with TextBlock

deltakosh: This is something I really want to achieve. We cannot have two "class" of text not being in the same inheritance branch

gui/src/2D/controls/structuredTextBlock.ts
```diff
@@ -0,0 +1,1295 @@
...
+        }
+
+        context.fillText(text, x, y);
+
+        if (attr.lineThrough) {
+            context.fillRect(x - decorationOverlap, y + lineThroughYOffset, width + 2 * decorationOverlap, thickness);
+        }
+    }
+
+    protected _drawFrame(attr: TextPartAttributes, x: number, y: number, width: number, height: number, context: ICanvasRenderingContext): void {
```
deltakosh, line 1207: This should not be part of the control. The control must only display text. The frame can be provided by a rectangle control

deltakosh:
> Finally, I hope you will be happy with that brand new control, because I have spent way more time than is was reasonable to spend, considering my personal schedule! ;)

We are:)! A lot actually. My only ask is to make it only about text (no frame) and make it inherit from Textblock.

deltakosh:
For your naming, I would probably prefer to refer to Link instead of Href. Like with `onLinkClickedObservable`

gui/src/2D/controls/structuredTextBlock.ts
cronvel: It was implemented to support this Markdown feature: --> `some inline code in a frame`
Is it not the *long term* goal to support Markdown anymore?

Also the frame **MUST** be drawn at the right time with the correct coordinates, if it's userland, at best it should be done with a *hook*. Because it's too late to draw it after: the rectangle **MUST** be drawn first on the 2D canvas or it would erase the text.

* without a hook: it's not possible for devs to implement it
* with a hook: it would be a quite complicated API requiring a lot of arguments for the hook (a 2D context, *x,y,width,height*, and attributes for the frame color, outline width/color, corner radius). I'm not sure it's worth the effort to externalize that with all the API glue-code and documentation that would require... and that probably nobody would read and use.

cronvel:
> For your naming, I would probably prefer to refer to Link instead of Href. Like with `onLinkClickedObservable`

Ok so `onLinkClickedObservable`, `onLinkEnteredObservable`, and for the last one? `onLinkExitedObservable`? (to follow the past participle trend?)

cronvel:
@deltakosh Also what is the release date for the next Babylon.js version? I don't want to miss the target ^^

gui/src/2D/controls/structuredTextBlock.ts
cronvel: @deltakosh (I forgot to @ you)

deltakosh: This is how the entire GUI system works. The parent is rendered first. Each control must only focus on one task. Rendering the frame is not expected from the text control:

https://playground.babylonjs.com/#XCPP9Y#10432

deltakosh:
> @deltakosh
> 
> > For your naming, I would probably prefer to refer to Link instead of Href. Like with `onLinkClickedObservable`
> 
> Ok so `onLinkClickedObservable`, `onLinkEnteredObservable`, and for the last one? `onLinkExitedObservable`? (to follow the past participle trend?)

Yes!

deltakosh:
> @deltakosh Also what is the release date for the next Babylon.js version? I don't want to miss the target ^^

No rush! It is April 28 2021 ;)

RaananW:
Wow, what a great PR! :-)

I'll try to summarize my personal thoughts about this in a single comment, and will probably do a code review later as well.

If I understand correctly, the idea here is to have a way to provide a sort of "markdown" support for a Babylon.js GUI element. meaning - paste your text (in a structured way), and we will display it correctly.

What we already have is a text block that can be added to any container available. What's missing from the current implementation is a way to have a flex-like container that can hold more than one text block. In other words, what's missing is a wrap-container that allows the dev to add elements to it, and wrap them if they reached its "end". Something like the stack panel but behaves more like a flex container with specific width and flex-wrap set. 

The text blocks can be styled (either individually, or using a style object), so we support this part as well. If there is some formatting missing in the text block, it can be added there.

A horizontal stackpanel is "roughly" that, but it requires the child to have a specific width and doesn't wrap. Something like this PR - https://playground.babylonjs.com/#XCPP9Y#10449. Since the width is set when creating the text (and not adjusted automatically) there is a gap between the text elements. 

As @deltakosh said, we are trying to maintain a one-task-one-element approach. If we already have a text block, let's (re)use it as much as we can. Reusing the text blocks in this form of container will eliminate some parts of the code - one of them is the pointer processing, which IMO is the one downside of this PR. the fact that we need to attach to different observables for the pointer clicks can be very confusing for the user. If we use many text blocks you can set one of them to be clickable using the existing pointer event system.  And if we have a container, we will be able to use it to style the frame. 

TL;dr

I think this PR should actually have two changes (unless impossible for some reason) - 

1. Make the existing TextBlock more dynamic (if it is missing any type of styling, add support to the text block
2. Add a (Text)Container that allows wrapping more than one (text) block. It should also allow wrapping the blocks internally (if they are too long). If an element has a width it should force this width, otherwise adjust accordingly.

----------------

This is the way I would implement those requirements. If it's impossible, or we do want to stay with the StructuredTextBlock, you should remove anything that does more than render the text (like the frame), and expect it to be used inside a container that is fully style-able.

cronvel:
Hello @RaananW ;)

> If I understand correctly, the idea here is to have a way to provide a sort of "markdown" support for a Babylon.js GUI element. meaning - paste your text (in a structured way), and we will display it correctly.

The idea of this PR is to pass *Structured Text* as input, an **agnostic** format consisting of array of objects, e.g. `[{text:"This is "},{text:"red",color:"#ff5050"}]`, achieving multi-styling inside the same block of text.
Someone wanting Markdown would have to use its favorite parser and transform it into *Structured Text*.
If someone prefer an HTML subset or BBCode, or whatever, no problem, again parse it with your favorite parser, and transform it into *Structured Text*.
Because the format is agnostic, it does not force any parser on you.
The format is really simple and straight-forward.

When coding it, I had video game dialog in mind (e.g. when a character talk in a RPG, or tutorial dialog), and also styling capabilities suitable for Markdown (because it looked like a popular request here).

Having built-in Markdown support could be great, but it would be better to keep this PR small and simple, and implement built-in Markdown later (while still supporting the agnostic *structured text* format).

> What we already have is a text block that can be added to any container available. What's missing from the current implementation is a way to have a flex-like container that can hold more than one text block. In other words, what's missing is a wrap-container that allows the dev to add elements to it, and wrap them if they reached its "end". Something like the stack panel but behaves more like a flex container with specific width and flex-wrap set.
> 
> The text blocks can be styled (either individually, or using a style object), so we support this part as well. If there is some formatting missing in the text block, it can be added there.
> 
> A horizontal stackpanel is "roughly" that, but it requires the child to have a specific width and doesn't wrap. Something like this PR - https://playground.babylonjs.com/#XCPP9Y#10449. Since the width is set when creating the text (and not adjusted automatically) there is a gap between the text elements.

I understand your point, and while it would be a **great idea**, this PR was not made to be a sort of inline/flex stackpanel container with wrapping.
It really is `TextBlock` on steroids (and my personal intent was to upgrade this class instead of creating a new one).

Also I basically ran out of free coding time, so if it will be the chosen path, I will probably hand over the whole thing to someone else (and I'm not even sure this PR would be a good starting point for that, there are 50% chance that it would be better to start again from scratch).

> As @deltakosh said, we are trying to maintain a one-task-one-element approach. If we already have a text block, let's (re)use it as much as we can. Reusing the text blocks in this form of container will eliminate some parts of the code - one of them is the pointer processing, which IMO is the one downside of this PR. the fact that we need to attach to different observables for the pointer clicks can be very confusing for the user.

I'm a bit confused with this. If you want hypertext link, only one observable is required: `onClickHrefObservable` (will be renamed `onLinkClickedObservable`).
One typical use-case for the 2 other observables would be for some hint/tooltip when hovering part of text.
That's two different use-cases.
By the way all of them are optional.

> This is the way I would implement those requirements. If it's impossible, or we do want to stay with the StructuredTextBlock, you should remove anything that does more than render the text (like the frame), and expect it to be used inside a container that is fully style-able.

To be honest, I don't see much difference between drawing frames or drawing underlines except a dozen of very trivial lines of code.
If this is a requirement, I would probably just drop that frame feature entirely, because there is no way at all to plug a *rectangle control* into it nicely (unless going the *flex container* way).

deltakosh:
> To be honest, I don't see much difference between drawing frames or drawing underlines except a dozen of very trivial lines of code.
If this is a requirement, I would probably just drop that frame feature entirely, because there is no way at all to plug a rectangle control into it nicely (unless going the flex container way).

I struggle to see why using a rectangle as the parent is not solving the frame question?

cronvel:
> > To be honest, I don't see much difference between drawing frames or drawing underlines except a dozen of very trivial lines of code.
> > If this is a requirement, I would probably just drop that frame feature entirely, because there is no way at all to plug a rectangle control into it nicely (unless going the flex container way).
> 
> I struggle to see why using a rectangle as the parent is not solving the frame question?

Sorry, I'm not a native english speaker and I suspect a misunderstanding here. The *frame* here is not something that surround the whole block of text, but just a/few word(s) within. That's why it's really similar to drawing the line of the underline text decoration. So that *frame* is susceptible to word-wrapping, reflow, etc, that makes its coordinates unpredictable ahead of time, and even changing (after a resize).

On this screenshot, this is the gray rounded frame under the second *Lorem Ispum* (in red):
Screenshot

For example if the whole *StructuredTextBlock* is shrinked (e.g. after a screen resize), the word-wrapping may even split this into 2 separated frames (one for *Lorem*, one for *Ipsum*).

I hope it clears the misunderstanding (if any).

deltakosh:
HAHA. ok yes there was a misunderstanding :) I thought it was the green rectangle!!

No problem for the grey one, it is part of the styling 

Sorry for the misunderstanding!

cronvel:
@deltakosh Yes! Indeed the green frame is **not** part of the *StructuredTextBlock*, it's just a regular *rectangle control*.
I'm glad we finally sort this out ;)

cronvel:
@deltakosh About the `TextBlock`'s inheritance, I'm struggling with many things.

Basically I'm cut out of `Control`'s methods, all `super.*()` calls produce bugs because the `TextBlock` counterpart get in the way and mess things up, when what I only need is access to the parent `Control`'s methods.
BTW Control.prototype.<method>.call() is not allowed, because those methods are not public (and it would be messy anyway to do that).

That's what I was forecasting, `TextBlock` and `StructuredTextBlock` diverge too much, and IMHO there are more sibling classes than parent-child classes.

I'm stuck.

RaananW:
@cronvel  - thank you so much for taking the time!

> The idea of this PR is to pass _Structured Text_ as input, an **agnostic** format consisting of array of objects, e.g. `[{text:"This is "},{text:"red",color:"#ff5050"}]`, achieving multi-styling inside the same block of text.

I totally get how much this is needed :-) I see it as a static "parse structure" function of the `TextContainer`. Which I believe your PR can represent. 

> It really is `TextBlock` on steroids (and my personal intent was to upgrade this class instead of creating a new one).

Yep! but it can still use the text block underneath. 

So, the way I see it - instead of rendering all of the structured text, you add children of type TextBlock and render them, with the same configuration you are currently using. I guess there are a few changes to be made (mainly for wrapping, i would assume), but it is still possible. Which leads me to the next point - 

> I'm a bit confused with this. If you want hypertext link, only one observable is required: `onClickHrefObservable` (will be renamed `onLinkClickedObservable`).

If you have text block children, and one of them is a link, it will notify the pointer observables when clicked on. I guess adding a "link" flag for the text block would be a nice option - when hovering above it, change the pointer, but otherwise, the pointer system on a specific element does support "link". Having those observables on the container level is confusing. They are also very specific to this component. No other component is so specific as to the type of event we are running. There is pointer down, pointer move, pointer up, over and out, but never something so specific. The dev should be able to use pointer events to react to that. It could be a part of your parse method (adding the observer to the text block)


> To be honest, I don't see much difference between drawing frames or drawing underlines except a dozen of very trivial lines of code.

I think we all misunderstood what this part does. Very much needed, very much part of the code. Thanks for clarifying!

cronvel:
@RaananW 

Here is the minimal working sample code for the link:
```js
var structuredTextBlock = new BABYLON.GUI.StructuredTextBlock() ;
structuredTextBlock.structuredText = [
    { text: "This is a proof of concept for " },
    { text: "hypertext button", color: "#3377aa", underline: true, href: "www.test.com" },
] ;

structuredTextBlock.isPointerBlocker = true;
structuredTextBlock.onLinkClickedObservable.add( data => alert( "href:" + data.href ) );
```
When clicking the link, an alert box will pop and display "www.test.com".

First, we can see how simple the setup is, this is really about **2 lines of code**, everything else being boiler-plate.

IMHO it would be really bad to move the callback inside the structure, the structure should not contain code, it should be *purely declarative*, and could come from any source (e.g. JSON over network, or whatever). Also if the text had multiple links, no need to pass multiple callbacks.

deltakosh:
> That's what I was forecasting, TextBlock and StructuredTextBlock diverge too much, and IMHO there are more sibling classes than parent-child classes.

It makes sense but then @RaananW idea to reuse code works well here

On the observable I'm less concerned than Raanan, I believe it is ok to have it on the top control (I will not die if it stays here)

RaananW:
> First, we can see how simple the setup is, this is really about **2 lines of code**, everything else being boiler-plate.

I am not saying it shouldn't be simple :-) I am just saying it should use already-existing elements

> IMHO it would be really bad to move the callback inside the structure, the structure should not contain code,

This gets the same answer as above :-) It's not about forcing the user to do something that is uncomfortable, it is about using an existing system. Every control already has pointer events registered to it. it's integrated in the entire system and being used by all parts of the framework. So when you add a link (same syntax as what you suggested), you just attach to the pointer down of the specific text block, and the user can control it using the pointer events of the container, or of this specific textblock, which is a child of the container. Also, I am not sure why would you not simply open the link automatically (it is a link after all), while providing an option to override the default behavior.

cronvel:
@RaananW
> Also, I am not sure why would you not simply open the link automatically (it is a link after all), while providing an option to override the default behavior.

It's called *link* because devs are familiar with HTML, but it is not meant to open a web-page. Ok my example had *"www.test.com"*, but it was just to illustrate. Beside `.href` have no type restriction and could be an object (it's pure userland data the framework doesn't care, it just transmit it).

Maybe the name *link* is not good if it confuses. It's just a clickable text that triggers some action.

RaananW:
> Maybe the name _link_ is not good if it confuses. It's just a clickable text that triggers some action.

Hence, a pointer event :-). This already exists and is part of the framework. Both names you mentioned had "link" and "href" in them so that might have made me think it's meant as a link, but if it is just a click event, any text block (any 2D control) has it already integrated. There is no issue in passing this data as part of the pointer event instead of creating a new observable... But the observables are not an issue. You could add "onChildClickedObservable", as long as it is easy and simple to use.

dad72:
Has this PR been dropped? nothing for 7 months.
I think he's an excellent PR, but he seems to be giving up after so much work and on such a good path.

RaananW:
if @cronvel is fine with it, maybe we can take over the PR and make slight changes? I will be really happy to merge this

dad72:
I think it can be taken back. the author no longer seems to be there to answer @RaananW .
This PR is such a cool addition for babylon.GUI, it would be a shame not to finish it.

cronvel:
@dad72 @RaananW Hi there,
Sorry for being late, I've got very limited time for open-source recently.

So, the question is how much work is necessary for this PR to be accepted?

* The last couple of suggestions (e.g. making all text-part reuse existing widgets) require almost a restart from scratch, I'm not even sure there is something to be taken over if we go this way. If that's the way to go, then someone else would have to make it, cause like I said, I've very limited time dedicated for open-source at the moment.
* If the *spirit of this PR* is accepted, and only few minor things should be changed, I could do it, but I should have a clear vision/path of the work to be done (and I want to avoid *change escalation* (<- I don't know if it makes sens, I have hard-time speaking english today!))

So it all comes down to how close this PR is to be finished. If there is too much work to do, then take it over ;)

cronvel:
So ultimately, what is the list of things to do for this PR to be accepted?

deltakosh:
Hey @cronvel! This is a great PR and after discussing quite extensively with the team we believe it qualifies more for an extension that to be inside the core repo

That will give you more latitude to evolve it

So my kind ask to you: do you mind reopening it in the extension repo? I'm closing it for now so our CI does not go crazy :D

dad72:
I have the impression that this functionality will not see the day. I was expecting it to use in my game chat, but I'll have to do without I guess.

deltakosh:
It is still here and it works :)

dad72:
You mean I could take the code to use the feature?

dad72:
yeah, I don't bet to make it work. I tried in the playground and I got an e.initFunction error that I don't understand.
I'm going to leave that to save myself a headache. Dommage.

cronvel:
@dad72 I'm using it on my side-project game, so it should work for you too ;)

@deltakosh Ok, let's do this. I will give it a try during the holiday, probably starting mid-July. (I don't really have holidays, since I'm freelance, but I usually have less work during the months of July and August).

Also I didn't found any documentation related to GUI extensions, any pointer? Is there an extension/plugin API?

deltakosh:
It is actually a separate repo where our community can maintain additional projects: https://github.com/BabylonJS/Extensions

PR #11386, opened 2021-11-04
Exclude unused material in glTFExporter

### LINKED ISSUE #10009: glTF2Exporter includes materials used by excluded meshes.
See:
https://forum.babylonjs.com/t/can-gltf-export-a-specific-mesh-from-scene/10641/3

Repro:
https://www.babylonjs-playground.com/#YBEKDT#4

Expected result:
Fan materials and textures should not be included in exported glTF from this playground.

daoshengmu:
Fixes #10009

- Excluding materials if they are not used by exported meshes.
- Collecting only exported nodes and materials beyond processing them.

### DIFF
--- a/serializers/src/glTF/2.0/glTFExporter.ts
+++ b/serializers/src/glTF/2.0/glTFExporter.ts
@@ -1543,7 +1543,7 @@ export class _Exporter {
                     for (const attribute of attributeData) {
                         const attributeKind = attribute.kind;
                         if ((attributeKind === VertexBuffer.UVKind || attributeKind === VertexBuffer.UV2Kind) && !this._options.exportUnusedUVs) {
-                            if (glTFMaterial && !this._glTFMaterialExporter._hasTexturesPresent(glTFMaterial)) {
+                            if (!glTFMaterial || !this._glTFMaterialExporter._hasTexturesPresent(glTFMaterial)) {
                                 continue;
                             }
                         }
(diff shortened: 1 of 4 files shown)


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1543,7 +1543,7 @@ export class _Exporter {
                     for (const attribute of attributeData) {
                         const attributeKind = attribute.kind;
                         if ((attributeKind === VertexBuffer.UVKind || attributeKind === VertexBuffer.UV2Kind) && !this._options.exportUnusedUVs) {
-                            if (glTFMaterial && !this._glTFMaterialExporter._hasTexturesPresent(glTFMaterial)) {
+                            if (!glTFMaterial || !this._glTFMaterialExporter._hasTexturesPresent(glTFMaterial)) {
```
sebavan, line 1546: Good catch

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1701,85 +1701,125 @@ export class _Exporter {
...
     }
 
+    /**
+     * Getting the nodes and materials that would be exported.
+     * @param nodes Babylon transform nodes
+     * @returns Array of nodes which would be exported.
+     * @returns Set of materials which would be exported.
+     */
+    private getExportNodes(nodes: Node[]): Promise<[Node[], Set<Material>]> {
+        let promiseChain = Promise.resolve();
```
sebavan, line 1784: most of the let in the function could be const

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1701,85 +1701,125 @@ export class _Exporter {
...
+        let exportNodes: Node[] = [];
+        let exportMaterials: Set<Material> = new Set<Material>();
+
+        for (let babylonNode of nodes) {
+            if (!this._options.shouldExportNode || this._options.shouldExportNode(babylonNode)) {
+                exportNodes.push(babylonNode);
+
+                if (babylonNode instanceof Mesh) {
```
sebavan, line 1792: Could we use getClassName instead ? instanceOf is pretty slow

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1701,85 +1701,125 @@ export class _Exporter {
...
         });
     }
 
+    /**
+     * Getting the nodes and materials that would be exported.
+     * @param nodes Babylon transform nodes
+     * @returns Array of nodes which would be exported.
+     * @returns Set of materials which would be exported.
+     */
+    private getExportNodes(nodes: Node[]): Promise<[Node[], Set<Material>]> {
```
sebavan, line 1783: I do not think this should be async so you could remove the promise chain

sebavan (commented):
Only a couple of minor comments, This looks like a great contrib !!!

sebavan (approved):

serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1772,14 +1773,50 @@ export class _Exporter {
         });
     }
 
+    /**
+     * Getting the nodes and materials that would be exported.
+     * @param nodes Babylon transform nodes
+     * @returns Array of nodes which would be exported.
+     * @returns Set of materials which would be exported.
+     */
+    private getExportNodes(nodes: Node[]): [Node[], Set<Material>] {
```
bghgary, line 1782: Super minor, but private functions should start with underscore. We can fix this later.

PR #11394, opened 2021-11-05
Add the missing setter for the BasisTools options

RaananW:
That was removed as part of previous module work. The setter is not setting the right object that is used throughout the framework.

### DIFF
--- a/src/Misc/basis.ts
+++ b/src/Misc/basis.ts
@@ -493,3 +493,21 @@ function workerFunc(): void {
         return dst;
     }
 }
+
+Object.defineProperty(BasisTools, "JSModuleURL", {
+    get: function (this: null) {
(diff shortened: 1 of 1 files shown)


Popov72 (approved):

src/Misc/basis.ts
```diff
@@ -493,3 +493,21 @@ function workerFunc(): void {
         return dst;
     }
 }
+
+Object.defineProperty(BasisTools, "JSModuleURL", {
```
sebavan, line 497: Could we move it near the type declaration ? I wonder if in maintenance we might not miss it when reading back the code ?

RaananW: This is moving to LTS in a month or two

PR #11400, opened 2021-11-07
Optimize scene UBOs

Popov72:
In this PR, new scene UBOs are created and used in subsystems where at least one data of the scene UBO (view/projection matrix, eye position) is modified compared to the regular (final) scene rendering.

It is the case for `MirrorTexture`, (CSM) `ShadowGenerator` and `ReflectionProbe`. In each case a new scene UBO (or 6 for reflection probe and n for CSM with n layers) is created and used in place of the regular scene ubo to perform the subsystem rendering. This way, the regular scene ubo (and each of the specifically created scene UBO) holds only a single GPU buffer which enables some optimizations inside the `UniformBuffer` class (we get rid of a comparison between two arrays of float + an array copy if the comparison is different). Also, having a single GPU buffer in the scene UB class is better for the fast path mode in WebGPU as it can lead to fewer recreations of the fast path bundle.

### DIFF
--- a/src/Engines/Extensions/engine.multiview.ts
+++ b/src/Engines/Extensions/engine.multiview.ts
@@ -131,15 +131,28 @@ declare module "../../scene" {
     }
 }
 
+function createMultiviewUbo(engine: Engine, name?: string) {
+    const ubo = new UniformBuffer(engine, undefined, true, name);
+    ubo.addUniform("viewProjection", 16);
+    ubo.addUniform("viewProjectionR", 16);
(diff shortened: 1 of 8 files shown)


sebavan (approved):

RaananW:
Is this ready to be merged?

Popov72:
I think we should wait for the review from @deltakosh.

Popov72:
@deltakosh is it ok for you?

deltakosh (approved):

PR #11403, opened 2021-11-08
Added optional preserveScalingSign to addChild and setParent to keep sign

CedricGuillemet:
follow up https://forum.babylonjs.com/t/mesh-addchild-with-negative-scaling-transforms-child-rotation-incorrectly/25052

reuse scaling sign coming from a node. Decompose recompute positive scaling and update rotation accordingly. Here, scaling sign is preserve so node scaling/rotation doesn't change after addChild/RemoveChild.

### DIFF
--- a/src/Maths/math.vector.ts
+++ b/src/Maths/math.vector.ts
@@ -4784,7 +4784,6 @@ export class Matrix {
 
         scale = scale || MathTmp.Vector3[0];
         if (preserveScalingNode) {
-            
             const signX = preserveScalingNode.scaling.x < 0 ? -1 : 1;
             const signY = preserveScalingNode.scaling.y < 0 ? -1 : 1;
             const signZ = preserveScalingNode.scaling.z < 0 ? -1 : 1;
@@ -4797,7 +4796,6 @@ export class Matrix {
             scale.y *= signY;
             scale.z *= signZ;
         } else {
-            scale = scale || MathTmp.Vector3[0];
             scale.x = Math.sqrt(m[0] * m[0] + m[1] * m[1] + m[2] * m[2]);
             scale.y = Math.sqrt(m[4] * m[4] + m[5] * m[5] + m[6] * m[6]);
             scale.z = Math.sqrt(m[8] * m[8] + m[9] * m[9] + m[10] * m[10]);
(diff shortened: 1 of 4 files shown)


src/Maths/math.vector.ts
```diff
@@ -4782,12 +4783,28 @@ export class Matrix {
...
-        scale.z = Math.sqrt(m[8] * m[8] + m[9] * m[9] + m[10] * m[10]);
+        if (preserveScalingNode) {
+            
+            const signX = preserveScalingNode.scaling.x < 0 ? -1 : 1;
+            const signY = preserveScalingNode.scaling.y < 0 ? -1 : 1;
+            const signZ = preserveScalingNode.scaling.z < 0 ? -1 : 1;
+
+            scale.x = Math.sqrt(m[0] * m[0] + m[1] * m[1] + m[2] * m[2]);
+            scale.y = Math.sqrt(m[4] * m[4] + m[5] * m[5] + m[6] * m[6]);
+            scale.z = Math.sqrt(m[8] * m[8] + m[9] * m[9] + m[10] * m[10]);
```
RaananW, line 4794: this calculation can come out of the if loop and be reused in both if and else

src/Maths/math.vector.ts
```diff
@@ -4782,12 +4783,28 @@ export class Matrix {
...
+
+            scale.x *= signX;
+            scale.y *= signY;
+            scale.z *= signZ;
+        } else {
+            scale = scale || MathTmp.Vector3[0];
```
RaananW, line 4800: no need for this line

RaananW (approved):

Popov72 (approved):

sebavan (approved):

src/Maths/math.vector.ts
```diff
@@ -4760,9 +4760,10 @@ export class Matrix {
      * @param scale defines the scale vector3 given as a reference to update
      * @param rotation defines the rotation quaternion given as a reference to update
      * @param translation defines the translation vector3 given as a reference to update
+     * @param preserveScalingNode Use scaling sign coming from this node. Otherwise scaling sign might change.
      * @returns true if operation was successful
      */
-    public decompose(scale?: Vector3, rotation?: Quaternion, translation?: Vector3): boolean {
+    public decompose(scale?: Vector3, rotation?: Quaternion, translation?: Vector3, preserveScalingNode?: TransformNode): boolean {
```
bghgary, line 4766: Would it be better to pass in a Vector3 for the scale? It is more generic and will be less dependent on TransformNode.

PR #11408, opened 2021-11-09
Add callback property to FramingBehavior

BlakeOne:
Add a callback property to FramingBehavior that gets passed to zoomOnMesh when the target is changed, so that  we can more easily know when the animation has ended.

Forum: https://forum.babylonjs.com/t/how-can-i-see-when-a-framingbehavior-animation-has-finished/25086/6

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -160,7 +160,7 @@
 
 ### Behaviors
 
-- Added `onTargetFramingAnimationEnd` callback to `FramingBehavior` ([BlakeOne](https://github.com/BlakeOne))
+- Added `onTargetFramingAnimationEndObservable` to `FramingBehavior` ([BlakeOne](https://github.com/BlakeOne))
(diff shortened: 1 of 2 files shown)


BlakeOne:
PS, if an observable would be preferable to the callback,  let me know and I can rework it that way. :)

RaananW:
We have moved towards using observables instead of fixed callbacks. This way you also don't have to check if it exists - just trigger notification and the observable will do the rest for you.

Thanks!!

BlakeOne:
Okay, I changed the new property to be an observable instead of a callback. zoomOnMesh takes a callback, so I used an anonymous callback which calls notifyObservers on the observable. :) 

PS, previously I was just passing the callback through, which is already being checked by onZoomMesh, so I think it just comes down to rather or not more than one callback will be needed by users and which is easier to set…

BlakeOne:
Hmm, also, on second thought, maybe onZoomToTargetEndObservable would be a better name than onTargetFramingAnimationEndObservable for this property?

sebavan (approved):

Issue #11334, opened 2021-10-23
Make the base material shader customizable


brunobg:
**Is your feature request related to a problem? Please describe.**

This is a proposal to make the base material shaders customizable. Essentially it'd be making `MaterialDefines` more like `CustomMaterial`, which would allow developers to patch materials at development time, making things that are only viable at this point as post-processing effects or multiple passes much easier to do and more efficient. This could also clean up some of the internals of BabylonJS, with fewer ad hoc `#includes<>` in the base shader files.

I'm splitting this from the discussion in https://github.com/BabylonJS/Babylon.js/pull/11317 so we can find a proposal that works, and I'm offering myself for implementing this.

Relevant comments:

_Originally posted by @brunobg in https://github.com/BabylonJS/Babylon.js/issues/11317#issuecomment-948678661_

>    May I not disregard it? :D When I was first thinking about how I could write this as a module the thought of monkey patching the material shader code came to me. Perhaps you could make the base glsl material shaders, like default.vertex.fx, more flexible. Perhaps making MaterialDefines more like CustomMaterial. This would make the material compilation slightly slower, but there would be an added benefit that all materials could be easily customized.
>    Effects that currently require postprocessing now could be done in one pass by patching the material. I can see a lot of cases for that. My own: I currently render caustics in postprocessing, which is a RTT. If I could do something like material.appendToLastFragmentPhase(myCausticsShaderCode) I could do everything in one pass.

_Originally posted by @Popov72 in https://github.com/BabylonJS/Babylon.js/issues/11317#issuecomment-949430908_ :

> I think it's possible to setup a mechanism on the js side to have material plugins, as our material extensions are all having the same structure:
> 
>     * they inject some additional defines in the material define list
> 
>     * they have a number of additional properties
> 
>     * they implement the same set of methods: isReadyForSubMesh, prepareDefines, bindForSubMesh, hasTexture, etc
> 
>     * [UPDATE] they add a number of additional uniforms in the material ubo
> 
> 
> However, I can't see how we can devise a generic way of injecting shader code in our standard/PBR materials... Depending on the material extension, the code must not be injected in the same location, and sometimes it is quite interlaced with the existing code (see the PBR extensions for eg)...
> 
[...]

PR #11410, opened 2021-11-09
Making the base material shader customizable

brunobg:
Proposal to port the `CustomMaterial` code into the base `Material` class, which would allow any material to have its shader customizable with external code fragments easily. This is proposed by https://github.com/BabylonJS/Babylon.js/issues/11334.

What is planned:

- a new `ShaderCustomization` class to store all the fragments and insert them into the actual shader code.
   - review special cases that rename `result` to a different variable. See if they are necessary and if there's a safer way to do them than a string replace.
- add `ShaderCustomization` as a new field in `Material`.
    - need to find a way to pass the vertex/fragment shader code to `ShaderCustomization`. 
    - review `customShaderNameResolve`. This seems like a hack to enable `CustomMaterial` to work, but it seems useless now. Check if this can be removed or it'd be an unacceptable breaking change.
    - call the `bind` callbacks on `ShaderCustomization` from `Material`.
- review `CustomMaterial` and `PBRCustomMaterial`. These classes will be useless now that `StandardMaterial` and `PBRMaterial` will have the same functionality, but they may be kept as pure façade classes to keep old code working just the same.
- see if new code entry points on the shaders are useful.


brunobg:
So this is my initial proposal to make materials customizable. Please take a look and tell me what you think. The initial commit is only the API and doesn't actually work, but should be enough to see if it makes sense and to refine the implementation tasks.

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+            for (let el in this._newSamplerInstances) {
+                const ea = el.toString().split('-');
+                if (ea[0] == 'sampler2D' && this._newSamplerInstances[el].isReady && this._newSamplerInstances[el].isReady()) {
+                    effect.setTexture(ea[1], this._newSamplerInstances[el]);
+                }
+            }
+        }
+    }
+
+    public ReviewUniform(name: string, arr: string[]): string[] {
```
deltakosh, line 104: just a note about caps: it should be `reviewUniform`

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+        if (name == "uniform" && this._newUniforms) {
```
deltakosh, line 105: ===

src/Materials/material.ts
```diff
@@ -194,6 +195,11 @@ export class Material implements IAnimatable {
      */
     public allowShaderHotSwapping = true;
 
+    /**
+     * Allows custom code pieces to me inserted in the shaders.
+     */
+    public shaderCustomization: ShaderCustomization;
```
deltakosh, line 201: please use an interface to help with tree shaking

src/Materials/material.ts
```diff
@@ -746,6 +752,8 @@ export class Material implements IAnimatable {
         if (this._scene.useMaterialMeshMap) {
             this.meshMap = {};
         }
+
+        this.shaderCustomization = new ShaderCustomization();
```
deltakosh, line 756: should not be by default

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
+import { Effect } from "./effect";
+import { Mesh } from "../Meshes/mesh";
+import { Texture } from "./Textures/texture";
+import { MaterialDefines } from "./materialDefines";
+
+/**
+ * Defines a code piece to customize the shader.
+ */
+export class ShaderCustomPiece {
```
deltakosh, line 9: should be in a separate folder with everything related to custom shader code (one file per entity)

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+     public code: string;
+    /**
+     * The insertion point.
+     */
+    public point: ShaderCodePoint;
+    /**
+     * The priority. Higher numbers are inserted first.
+     */
+    public priority: number;
+    constructor(code: string, point: ShaderCodePoint, priority: number = 0) {
```
deltakosh, line 22: Can be:

constructor(public code: string, public point: ShaderCodePoint, public priority: number = 0) {

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+
+/**
+ * All keys for ShaderCodePoint
+ */
+type ShaderCodePointKeys = keyof typeof ShaderCodePoint;
+
+export class ShaderCustomization {
+    public static ShaderIndexer = 1;
+    public CustomPieces: ShaderCustomPiece[];
+    _isCreatedShader: boolean;
```
deltakosh, line 59: must be flagged as private

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+    public AddUniform(name: string, kind: string, param: any): ShaderCustomization {
+        if (!this._customUniform) {
+            this._customUniform = new Array();
+            this._newUniforms = new Array();
+            this._newSamplerInstances = {};
+            this._newUniformInstances = {};
+        }
+        if (param) {
+            if (kind.indexOf("sampler") != -1) {
+                (<any>this._newSamplerInstances)[kind + "-" + name] = param;
```
deltakosh, line 259: we must avoid any as much as possible

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+            for (var ind = 0; ind < this._newUniforms.length; ind++) {
+                if (this._customUniform[ind].indexOf('sampler') != -1) {
+                    arr.push(this._newUniforms[ind]);
+                }
+            }
+        }
+        return arr;
+    }
+
+    public Builder(shaderName: string, uniforms: string[], uniformBuffers: string[], samplers: string[], defines: MaterialDefines | string[], attributes?: string[]): string {
```
deltakosh, line 122: I struggle to understand how the system will run it? Is it on me to call Builder?

src/Materials/customShaderCode.ts
deltakosh, line 105: can be tested initially and return:

```
if (!this._newUniforms) {
return;
}

File: src/Materials/customShaderCode.ts [hunk 10]
```diff
@@ -0,0 +1,295 @@
...
+                    effect.setTexture(ea[1], this._newSamplerInstances[el]);
+                }
+            }
+        }
+    }
+
+    public ReviewUniform(name: string, arr: string[]): string[] {
+        if (name == "uniform" && this._newUniforms) {
+            for (var ind = 0; ind < this._newUniforms.length; ind++) {
+                if (this._customUniform[ind].indexOf('sampler') == -1) {
```
deltakosh, line 107: I feel like this code can be factorized

sebavan:
@brunobg, please do not make further change yet and let s discuss the APIs :-)

I like how the entry points are defined and such but I am worried about how intrusive it is in all materials even if you are not using it.

I am wondering if it could be more like an helper working in pairs with some observables on the materials like onBeforeCompileEffectObservable or equivalent ? 

@deltakosh what do you think ?

src/Materials/customShaderCode.ts
```diff
@@ -0,0 +1,295 @@
...
+        //     Effect.ShadersStore[name + "VertexShader"] = Effect.ShadersStore[name + "VertexShader"].replace('#define CUSTOM_VERTEX_UPDATE_WORLDPOS', this.CustomParts.Vertex_After_WorldPosComputed);
+        // }
+        // if (this.CustomParts.Fragment_Before_Fog) {
+        //     Effect.ShadersStore[name + "PixelShader"] = Effect.ShadersStore[name + "PixelShader"].replace('#define CUSTOM_FRAGMENT_BEFORE_FOG', this.CustomParts.Fragment_Before_Fog);
+        // }
+
+        this._isCreatedShader = true;
+        this._createdShaderName = name;
+
+        return name;
```
deltakosh, line 240: We need a way to track the generated code better so we can remove it if required. Why not returning a fully functional ShaderMaterial?

deltakosh (changes requested):
I struggle to get the crux of it. I think this is supposed to work only with StandardMaterial right? Because we need the anchor points (like CUSTOM_FRAGMENT_...) right?

So why making it at Material level?

Also I think you expect the user to provide the vertex shader and the fragment shader right? Not sure to understand why this is useful for the end user.

Maybe you could provide an example of how it will look like in the PG?

brunobg:
BTW, the initial proposal comes from `CustomMaterial`, so a lot of the code review you guys made applies directly to it. I tried to write something as simple as possible just to validate the idea first, I don't even know if all the old methods would still be useful.

> I like how the entry points are defined and such but I am worried about how intrusive it is in all materials even if you are not using it.

I see. It's a NOOP in that case, and by using composition it could be easy to disable it by making it nullable. But an observable works too and might be more flexible. 

> I struggle to get the crux of it. I think this is supposed to work only with StandardMaterial right? Because we need the anchor points (like CUSTOM_FRAGMENT_...) right?

Standard and PBR mostly, which (and my opinions are all from my experience as an user) are the most common materials. But any material whose shaders supports this pre-defined entry points.

I think it'd be important to review the entry point list (I copied it straight from `CustomMaterial`). Some points would be useful in any script like `CUSTOM_FRAGMENT_BEFORE_FRAGCOLOR` (and I'd add a `_AFTER_` too).

> So why making it at Material level?

This enables one to change existing shaders in a way that is impossible currently. My personal example is [rendering caustics](https://doc.babylonjs.com/divingDeeper/postProcesses/renderTargetTextureMultiPass). It's currently done by a separate pass with a RTT, but it'd be trivial to add a small code at the end of any arbitrary material just multiplying the `fragColor` by the computed caustic light. Instead of two passes, the effect could be done in one pass. 

Pretty much all decal effects, anything that applies a change to the base shader color, as well as changing the vertex shader to compose with a geometry change effect, could be done this way. It also allows you to compose micro-shaders together. Want to change `fur` to add an electrical effect? No problem.

> Maybe you could provide an example of how it will look like in the PG?

Some rough ideas of what I'd like to to below.

Let's make a material greyscale, no matter what it is. Even fur.

```js
// makes this material greyscale
mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_After_FragColor,
   'color.rgb = dot(color.rgb, vec3(0.299, 0.587, 0.114));'
);
```

Adds some underwater effects to any material on the scene:

```js
// makes this 
mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_Definitions,
   `vec3 causticColor(vec3 color, vec3 position) { 
      vec3 newColor;
      // some magic code goes here
      return newColor;
   }`
);
// changes color according to distance compensating for water absorption
mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_Definitions,
   `vec3 underwaterColor(vec3 color, vec3 position) { 
      vec3 newColor;
      // some magic code goes here
      return newColor;
   }`
);

mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_After_FragColor,
   'color.rgb = causticColor(color, position) * underwatercolor(color, position);'
);
```

By having multiple code slots for each entry point it's possible to compose effects, which can become plugins and even be parametric. Think `PostProcess` but applied directly to individual materials.

```js
import { Grayscale } from 'ShaderPiece/grayscale';
import { Contrast } from 'ShaderPiece/contrast';
import { WaterDrops } from 'ShaderPiece/waterDrops';

mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_Before_FragColor,
   Contrast(1.5)
);
mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_Before_FragColor,
   WaterDrops(waterDropMinRadius, waterDropMaxRadius);
);
mesh.material.customizeShader.add(
   ShaderCodePoint.Fragment_Before_FragColor,
   Grayscale
);
```

Add a "beating" effect to the vertex shader, something like

```js
mesh.material.customizeShader.add(
   ShaderCodePoint.Vertex_Update_WorldPos
   'worldPos *= 1.0 + 0.2*cos(time);'
);
```

deltakosh:
> This enables one to change existing shaders in a way that is impossible currently.
But that will require you to have the entry point defines in the shader code and so far this is only for std and PBR?


Btw I like the overall API even though I would prefer that form for tree shaking reason:
```
let shaderCustomization = new ShaderCustomization(material);

shaderCustomization.add(
   ShaderCodePoint.Vertex_Update_WorldPos
   'worldPos *= 1.0 + 0.2*cos(time);'
);
```

brunobg:
> > This enables one to change existing shaders in a way that is impossible currently.
> > But that will require you to have the entry point defines in the shader code and so far this is only for std and PBR?

Yes, all the current shader files will need `#define` entry points. Adding those would be part of this PR. I don't think there's a way around that unfortunately, and perhaps some entry points won't make sense for certain files. But it's pretty much like having to add the `#include<>` currently, which is mandatory if you write your own shader to handle bones, instances etc.

> Btw I like the overall API even though I would prefer that form for tree shaking reason:
> 
> ```
> let shaderCustomization = new ShaderCustomization(material);
> 
> shaderCustomization.add(
>    ShaderCodePoint.Vertex_Update_WorldPos
>    'worldPos *= 1.0 + 0.2*cos(time);'
> );
> ```

Either would work well for me as an user, though in this case `Material` would need someway to call `ShaderCustomization`, probably like `customShaderNameResolve` now. 

But perhaps the API should allow some way to apply a single `ShaderCustomization` to all objects in a scene, or at least many of them? I can see requirements like `make all objects greyscale except the selected one`. What do you guys think, perhaps some way to apply it to the scene as a whole too? Or do you think that the `Observable` approach could handle this case well?

sebavan:
> But perhaps the API should allow some way to apply a single `ShaderCustomization` to all objects in a scene, or at least many of them? I can see requirements like `make all objects greyscale except the selected one`. What do you guys think, perhaps some way to apply it to the scene as a whole too? Or do you think that the `Observable` approach could handle this case well?

Let s not see too big at first but yes it could be a nice evolution depending how the feature is evolving.

I would say this look great:
```
const shaderCustomization = new ShaderCustomization(material);

shaderCustomization.add(
   ShaderCodePoint.Vertex_Update_WorldPos
   'worldPos *= 1.0 + 0.2*cos(time);'
);
```

Abstracting the dependency away is the most important part of this work. I agree the material will need to be able to extends on demand and Observables/Callbacks can definitely help to ensure Materials are not even aware ShaderCustomizations exists. 

To go one step further we could probably allow it to work with PostProcess and such as well later on or at the scene level as you mentionned but let s first abstract it enough from the objects they are targetting.

deltakosh:
> Abstracting the dependency away is the most important part of this work. I agree the material will need to be able to extends on demand and Observables/Callbacks can definitely help to ensure Materials are not even aware ShaderCustomizations exists.

YES!

Popov72:
I think we would also have liked to use this feature for our own code (detail map, sheen, clear coat, etc) so that we don't have to pay for them if we don't use these features. That's why we mentionned we could use a plugin mechanism to abstract this code.

Let me try to sketch something so it's easier to discuss as I fear it will take more time for me to describe it than to do a rough/partial implementation.

Popov72:
So, here's a proof of concept:

https://github.com/Popov72/Babylon.js/compare/master...Popov72:material-plugin?expand=1

I have updated the PBR sub surface configuration class to be a material plugin instead. This is the most complicated one among detail map, clear coat, sheen, etc so I thought it was a good test.

The main changes are:
* a `IMaterialPlugin` interface that the class must implement
* a `MaterialPluginManager` class that coordinates everything plugin related. Note that I think all the methods (except for `RegisterPlugin` and `InjectPlugins`) should be implemented as methods of `Material` instead. That way we would avoid to pass `material` as the first parameter for all the methods (=> let me know what you think). But having a separate class is probably more readable to evaluate the changes.

For sub surface (as well as clear coat, sheen, etc) the shader code is already available in the files in **Shaders/** and **ShadersInclude/** so there's no need to pass custom code. For user created plugin materials, however, one has to pass the code. It is done thanks to the `IMaterialPlugin.getCustomCode` method. See this PG for an example of a "colorify" material plugin: http://localhost:1338/Playground/index-local.html#P8B91Z#3

You have:
```javascript
    BABYLON.MaterialPluginManager.RegisterPlugin("colorify", (material) => {
        return new ColorifyPluginMaterial(material);
    });
```
to register the plugin, and:
```javascript
class ColorifyPluginMaterial {
    _isEnabled = true;
    color = new BABYLON.Color3(1, .6, 0.8);

    get isEnabled() {
        return this._isEnabled;
    }

    set isEnabled(enabled) {
        if (this._enabled === enabled) {
            return;
        }
        this._isEnabled = enabled;
        this.markAllAsDirty();
    }

    initialize(scene, dirtyCallbacks) {
        this.markAllAsDirty = dirtyCallbacks[BABYLON.Constants.MATERIAL_AllDirtyFlag];
    }

    collectDefineNames(names) {
        names.push("COLORIFY");
    }

    prepareDefines(defines, scene) {
        defines.COLORIFY = this._isEnabled;
    }

    addUniforms(uniforms) {
        uniforms.push("myColor");
    }

    bindForSubMesh(uniformBuffer, scene, engine, subMesh) {
        if (this._isEnabled) {
            uniformBuffer.getEffect().setColor3("myColor", this.color);
        }
    }

    getClassName() {
        return "ColorifyPluginMaterial";
    }

    getCustomCode(shaderType) {
        return shaderType === "vertex" ? null : {
            "CUSTOM_FRAGMENT_DEFINITIONS" : `uniform vec3 myColor;`,

            "CUSTOM_FRAGMENT_BEFORE_FRAGCOLOR": `
                #ifdef COLORIFY
                    finalColor.rgb *= myColor;
                #endif
            `
        };
    }
}
```
to implement the plugin.

In the material (only `PBRMaterial` in the example because I didn't update `StandardMaterial`), you can do `mat.colorify.isEnabled = false/true` or `mat.colorify.color.setAll(1, 0, 0)`: from the user standpoint, the plugin is integrated right inside the regular material (so, of course, you can also access `mat.subSurface` as before).

Let's discuss about this before going further.

sebavan:
This is already a tremendous improvment for the solution, thanks @Popov72  !!! 

I am still seeing some leftovers "intrusions" in the materials code and I wonder if it could be fully decoupled through observables/events/callbacks or similar so by default Material or PBR do not even know a pluginManager functionality is existing ?

Popov72:
Indeed, it is only a PoC and a number of things would have to be tackled (plugin order priority, selective injection of plugins, non intrusive way as @sebavan described, ...).

On my side I have a number of things to do/finish (mostly for WebGPU) before my (long) break at the end of the year, so I may not have time to move forward on the subject in the short term. Feel free to regain control if you so desire!

brunobg:
> Feel free to regain control if you so desire!

I'll gladly work on this again once a good non-intrusive is decided in this thread, since I don't know Babylon well enough to propose a good solution. Thanks @Popov72 :)

deltakosh:
> I'll gladly work on this again once a good non-intrusive is decided in this thread

Thanks a ton! I believe we simply need to add some observables where we need the injections and it should be beautiful

Popov72:
I don't think observables will work because we need to be able to retrieve some data from the observers: I don't think there's a way for the observers to return data to the observable that sent the notification (meaning when `notifyObservers` is called)?

Also, I added some specific properties to the `Material` class to be used by the `MaterialPluginManager`: if we don't want this plugin mechanism to be intrusive I guess we will need to move them somewhere else.

deltakosh:
The observables can notify the observers with an instance of an object where each observers can update data

brunobg:
> The observables can notify the observers with an instance of an object where each observers can update data

Any good references of observables in Babylon that I could follow for this implementation?

deltakosh:
Sure: https://github.com/BabylonJS/Babylon.js/blob/master/src/Materials/material.ts#L1077

Popov72:
See the `Material.onDisposeObservable` for example.

Basically, you declare your observable as a property of the class:
```typescript
    public onDisposeObservable = new Observable<Material>();
```
That way, external code can subscribe to it by doing something like `mymat.onDisposeObservable.add(() => { ... });`

Then in the `Material.dispose` method implementation we do `this.onDisposeObservable.notifyObservers()` to call the callbacks that have been registered for this observable.

So, we would need to add a number of `onXXXObservable` to the class so that the plugin manager can register to them and react accordingly. However, that would mean adding quite a lot of observables, like 10 or 15 of them... That seems overkill, maybe we could have a single observable and we would use the `mask` parameter to differentiate the events. Also, this observable would probably need to be static because the plugin manager needs to be notified from the `Material` constructor, so something like `public static OnEventObservable = new Observable<Material>();`

PoC:
https://github.com/Popov72/Babylon.js/compare/6efcaaf5dbc53125d897a928340fec49ef4b4039...32efbea4ed05f1f6e48ec58cc39954fa370ea3b0

I have only handled 3 events (creation / release / getting the `disableAlphaBlending` property) but it shows we can send data to the observer (**Disposed** event) or retrieve data (**GetDisableAlphaBlending** event).

Is that what you had in mind @deltakosh / @sebavan?

deltakosh:
Totally!!!

brunobg:
Thanks @Popov72. I think I'll have time to work on this next week. I think the general architecture is well defined and everyone approved it.

Popov72 (commented):
LGTM!

brunobg:
thanks for the review @Popov72! I'm trying my hand here at some changes to other PBR plugins to understand what you did. I'll also write the missing comments etc. I'm stuck debugging why my changes aren't working meanwhile.

src/Materials/PBR/pbrSheenConfiguration.ts
```diff
@@ -1,39 +1,59 @@
...
 }
 
+const modelDefines = new MaterialSheenDefines();
+
 /**
  * Define the code related to the Sheen parameters of the pbr material.
  */
-export class PBRSheenConfiguration {
+export class PBRSheenConfiguration implements IMaterialPlugin {
+    public priority: 0;
```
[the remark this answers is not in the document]
brunobg: Cool. I just added the field, I'll keep this open to handle once I get to the actual implementation of priority.

brunobg:
After adding the new plugins I'm getting this: `WebGL warning: drawElementsInstanced: Buffer for uniform block is smaller than UNIFORM_BLOCK_DATA_SIZE.` This happens at `thinEngine.drawElementsInstanced`. Any ideas of what I missed in the code?

brunobg:
> It probably won't work until you implement the priority thing: the plugin instances should be injected in the material in the priority order.

I see. So ideally `pbrUboDeclaration` should be built at runtime now, right? With a new method in `IMaterialPlugin` like `prepareUboDeclaration`?

I was also missing the imports, but everything works now, clear coat/anistropy/sheen/subsurface (detail map still to come). PGs working:

- PBR test from docs: http://localhost:1338/Playground/index-local.html#ASNS5H
- Updated the colorify demo: http://localhost:1338/Playground/index-local.html#P8B91Z#5

I guess this confirms the PoC idea works perfectly (thanks @Popov72!). How to do you want to proceed with this PR? Perhaps you could help me to draw a basic roadmap. Is this okay?

- finish PBRMaterial (only detail map?)
- see if anything needs to be made to `StandardMaterial` and `PushMaterial`
- add `#defines` injection points to all relevant `.fx`
- see if code like `pbrUboDeclaration` should be auto-generated
- check which other `MaterialEvent` items are necessary and implement them.
- implement `MaterialPluginManager.AddPluginToMaterial()`

Thanks!

sebavan:
One of the main remaining point would be to ensure MaterialPluginManager is not use in PBRBaseMaterial as we discussed earlier :-) I let you check with @Popov72 for this one.

Popov72:
Yes, we must remove all occurrences of `MaterialPluginManager` from `PBRBaseMaterial` and replace them by observer notifications.

`StandardMaterial` should also be updated to support plugins but there's no need to update `PushMaterial`.

Regarding auto-generating **pbrUboDeclaration.fx** I don't know: what do you think @sebavan? That would also allow an external material plugin to inject data inside the ubo instead of having its data alongside... Note that to support this we would need to update the `UniformBuffer.addUniform`method to pass the variable type.

One other thing is that `MaterialPluginManager` should not be written as a class with static methods but as regular exported functions instead, to match the code reorganization that @RaananW is doing.

brunobg:
> One other thing is that MaterialPluginManager should not be written as a class with static methods but as regular exported functions instead, to match the code reorganization that @RaananW is doing.

Any good reference of this in the code so I can follow the standard? Probably something like `groundBuilder.ts` with the `Create*` functions?

brunobg:
BTW, any specific reason to make `Material._EventInfo` a static private variable, instead of using a local variable? It's not reentrant.

brunobg:
Yet another question. Premature optimization is the root of all evil, but is it ok performance-wise to implement most of the observers directly on the `Material` class? Methods will look more or less like this now:

```ts
    /**
     * Specifies if the material uses a texture
     * @param texture defines the texture to check against the material
     * @returns a boolean specifying if the material uses the texture
     */
    public hasTexture(texture: BaseTexture): boolean {
        const userInfo: MaterialEventInfoHasTexture = {
            hasTexture: false
        };

        Material.OnEventObservable.notifyObservers(
            this,
            MaterialEvent.HasTexture,
            undefined,
            undefined,
            userInfo
        );

        return userInfo.hasTexture;
    }
```

This works perfectly well, but adds three extra function calls that in most cases will do nothing. Is this going to do a significant performance hit that should be avoided? Perhaps in that case `OnEventObservable` shouldn't be static.

Popov72:
> Any good reference of this in the code so I can follow the standard? Probably something like groundBuilder.ts with the Create* functions?

Yes, you can use this file as an example. Basically it's simply defining the functions as export **function XXX() { ... }** and not creating a class with static methods.

Note that you don't need to do something like:
```typescript
/**
 * Class containing static functions to help procedurally build meshes
 * @deprecated use the functions directly from the module
 */
export const GroundBuilder = {
    CreateGround,
    CreateGroundFromHeightMap,
    CreateTiledGround
};
```
This code is for backward compatibility (as the @deprecated tag shows).

> BTW, any specific reason to make Material._EventInfo a static private variable, instead of using a local variable? It's not reentrant.

I did it static to avoid creating a new object each time we need to notify the observers. Dealing with reetrancy would mean to create an object each time we notify and I fear it would be bad for perf/GC... I don't know what @sebavan things about this? Should we document the fact that the material events are not reetrant and one should not implement an observer that would in turn would lead to sending notifications to another/others observers or should we handle it in the proper way?

> Premature optimization is the root of all evil, but is it ok performance-wise to implement most of the observers directly on the Material class?

Yes, as much as possible the event notifications should be implemented in `Material` so that derived classes can inherit from those methods and don't reimplement the code (at least when the overriden methods call the super.XXX methods).

Regarding performances, I don't think we have a choice but I'm not sure it's a problem, the notification call will normally only be done a single time per frame per material per method that raises the notification (if the method is called each frame - not all methods raising the event is called each frame).

brunobg:
> Regarding auto-generating pbrUboDeclaration.fx I don't know: what do you think @sebavan? That would also allow an external material plugin to inject data inside the ubo instead of having its data alongside... Note that to support this we would need to update the UniformBuffer.addUniformmethod to pass the variable type.

Maybe I'm wrong, but I think this will be necessary if you look at `pbrBaseMaterial.buildUniformLayout()`. If someone adds a new uniform in their own plugin, it won't be reflected in `pbrUboDeclaration.fx` and it won't work, right?

brunobg:
Last commit implements the observable version. `RegisterMaterialPlugin` is now in the `BABYLON` namespace -- not sure if that is expected, but it's what I got from the GroundBuilder response.

Updated example: http://localhost:1338/Playground/index-local.html#P8B91Z#6

@Popov72 this is not the final code yet, so don't bother with a detailed code review. There's some stuff in PBR that should go to the base material class, detailmap as a plugin, etc. But I think it's good enough for a new overview, since the API and architecture won't change in future commits.

Popov72:
> If someone adds a new uniform in their own plugin, it won't be reflected in pbrUboDeclaration.fx and it won't work, right?

Yes it will work because you don't have to use the ubo to pass data: you can directly use the `Effect` object. That means however that we need to pass the effect to `bindForSubMesh` and `hardBindForSubMesh`.

brunobg:
Ok, the current code implements it all on PBR and Standard with observables. Waiting on feeback of previous questions to make a final cleanup and mark it ready for review.

> StandardMaterial should also be updated to support plugins but there's no need to update PushMaterial.

Well, lots of materials in the library extend `PushMaterial`. They are not customizable otherwise, and I see no reason why they shouldn't be.

Also, perhaps the `MaterialDefines` constructor could implement the signature from std/pbr? `(keysFromPlugins?: string[])`, and save the keys internally. Asking because it's a change in a base class.

Popov72:
> Well, lots of materials in the library extend PushMaterial

Yes but `PushMaterial` extends `Material` so inherits the event implementation from `Material`.

> Also, perhaps the MaterialDefines constructor could implement the signature from std/pbr? (keysFromPlugins?: string[]), and save the keys internally. Asking because it's a change in a base class.

I would lean toward doing it but I will let @sebavan / @deltakosh takes the decision. If we add it, I think we should name it `externalKeys` instead because `keysFromPlugins` is too specific from our current case.

I will make another pass of the code tomorrow.

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -248,85 +253,33 @@ export class PBRMaterialDefines extends MaterialDefines
...
     public DEBUGMODE = 0;
 
+    protected _keysFromPlugins?: string[];
+
     /**
      * Initializes the PBR Material defines.
+     * @param keysFromPlugins The plugin keys
      */
-    constructor() {
+    constructor(keysFromPlugins?: string[]) {
```
sebavan, line 270: I am not sure this signature can work as it does not take the type in account e.g. a number vs a boolean as well as the default values.

src/Materials/IMaterialPlugin.ts
```diff
@@ -0,0 +1,186 @@
...
+export interface IMaterialPlugin {
+    /**
+     * Defines the priority of the plugin. Lower numbers run first.
+     */
+    priority: number;
+
+    /**
+     * True if alpha blending should be disabled.
+     */
+     disableAlphaBlending?: boolean;
```
deltakosh, line 29: tiny space issue (nit)

src/Materials/PBR/pbrAnisotropicConfiguration.ts
```diff
@@ -1,33 +1,58 @@
...
     _needUVs: boolean;
 }
 
+const modelDefines = new MaterialAnisotropicDefines();
+
 /**
  * Define the code related to the anisotropic parameters of the pbr material.
  */
-export class PBRAnisotropicConfiguration {
+export class PBRAnisotropicConfiguration implements IMaterialPlugin {
```
deltakosh, line 48: Should we say `IMaterialPluggable` instead?

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -248,85 +253,33 @@ export class PBRMaterialDefines extends MaterialDefines
...
         super();
+        this._keysFromPlugins = keysFromPlugins;
         this.rebuild();
     }
 
+    public rebuild() {
+        super.rebuild();
+        if (this._keysFromPlugins) {
+            this._keys.push(...this._keysFromPlugins);
```
sebavan, line 279: Why not sharing the rebuild code from the base class and passing the array into it ?

src/Materials/PBR/pbrAnisotropicConfiguration.ts
```diff
@@ -70,19 +95,43 @@ export class PBRAnisotropicConfiguration {
...
+     * @param material The material implementing this plugin.
+     */
+    constructor(material: PBRBaseMaterial) {
+        this._material = material;
+    }
+
+    /**
+     * Initializes this plugin.
+     * @param scene The scene
+     * @param dirtyCallbacks
```
deltakosh, line 107: could be a bit more descriptive

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -248,85 +253,33 @@ export class PBRMaterialDefines extends MaterialDefines
...
-    public SS_USE_THICKNESS_AS_DEPTH = false;
-
-    public SS_MASK_FROM_THICKNESS_TEXTURE = false;
-    public SS_USE_GLTF_TEXTURES = false;
-
     public UNLIT = false;
 
     public DEBUGMODE = 0;
 
+    protected _keysFromPlugins?: string[];
```
sebavan, line 264: name as @Popov72 mentionned would be better as externalKeys

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -993,7 +951,20 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             return true;
         }
 
-        return this.subSurface.hasRenderTargetTextures();
+        // TODO: super.hasRenderTargetTextures; is not allowed. This copies the code.
```
deltakosh, line 954: need to remove the TODO

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -922,12 +875,12 @@ export abstract class PBRBaseMaterial extends PushMaterial {
     /**
      * Defines the clear coat layer parameters for the material.
      */
-    public readonly clearCoat = new PBRClearCoatConfiguration(this._markAllSubMeshesAsTexturesDirty.bind(this));
+    public readonly clearCoat: PBRClearCoatConfiguration;
```
sebavan, line 878: shouldn t those be nullable if uninitialized ? I am afraid of typings not highlighting the possibility it is not available but this would be a breaking change so it should always get imported for the current one.

src/Materials/material.ts
```diff
@@ -723,6 +758,12 @@ export class Material implements IAnimatable {
     constructor(name: string, scene: Scene, doNotAdd?: boolean) {
         this.name = name;
         this._scene = scene || EngineStore.LastCreatedScene;
+        this._dirtyCallbacks = {};
+
+        this._dirtyCallbacks[Constants.MATERIAL_TextureDirtyFlag] = this._markAllSubMeshesAsTexturesDirty.bind(this);
+        this._dirtyCallbacks[Constants.MATERIAL_PrePassDirtyFlag] = this._markAllSubMeshesAsPrePassDirty.bind(this);
+        this._dirtyCallbacks[Constants.MATERIAL_AllDirtyFlag] = this._markAllSubMeshesAsAllDirty.bind(this);
+        // TODO: need to add all dirty callbacks
```
deltakosh, line 766: need to fix that TODO as well

src/Materials/material.ts
```diff
@@ -746,6 +787,21 @@ export class Material implements IAnimatable {
         if (this._scene.useMaterialMeshMap) {
             this.meshMap = {};
         }
+
+        Material.OnEventObservable.notifyObservers(this, MaterialEvent.Created);
+
+        // TODO: _defineNamesFromPlugins needs to be updated with `MaterialPluginManager.AddPluginToMaterial` too
```
deltakosh, line 793: same

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -975,13 +928,18 @@ export abstract class PBRBaseMaterial extends PushMaterial {
                 this._renderTargets.push(<RenderTargetTexture>this._reflectionTexture);
             }
 
-            this.subSurface.fillRenderTargetTextures(this._renderTargets);
+            Material.OnEventObservable.notifyObservers(
+                this,
+                MaterialEvent.FillRenderTargetTextures,
+                undefined,
+                undefined,
+                { renderTargets: this._renderTargets } as MaterialEventInfoFillRenderTargetTextures
```
sebavan, line 936: This creates GC

src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1418,20 +1393,20 @@ export abstract class PBRBaseMaterial extends PushMaterial {
...
-
-        PBRClearCoatConfiguration.AddUniforms(uniforms);
-        PBRClearCoatConfiguration.AddSamplers(samplers);
-
-        PBRAnisotropicConfiguration.AddUniforms(uniforms);
-        PBRAnisotropicConfiguration.AddSamplers(samplers);
-
-        PBRSheenConfiguration.AddUniforms(uniforms);
-        PBRSheenConfiguration.AddSamplers(samplers);
+        Material.OnEventObservable.notifyObservers(
```
sebavan, line 1396: Should those be merged in only one event ?

deltakosh (changes requested):
Man! this is quite impressive. Excellent job! We ned to move the what's new entry to the major feature section

Really good job!

src/Materials/materialEvent.ts
```diff
@@ -0,0 +1,346 @@
...
+/**
+ * Type for material custom code generators.
+ */
+export type MaterialCustomCodeFunction = (shaderType: string, code: string) => string;
+
+/**
+ * Flags to filter observables in events for material plugins.
+ * @since 5.0.0
+ */
+ export enum MaterialEvent {
```
sebavan, line 24: Why do we need a flag type here for the enum ? I do not think a combination makes sense. as the type would be wrong. also there might be a clever way to prevent the ...as... we see everywhere to match an event and a type of data ?

sebavan (changes requested):
This looks much better GG !!!

I currently mostly looked in the pbr material changes, and I think the biggest points to solve are the define typings, the material events types and the GC generation. Let s focus on those first ?

sebavan:
Also on a side node I do not think all the registration should be static but more per material so event types and constants could be more targeted to the type of material.

we could use type inference to ensure of the typings and remove all the required cast

src/Materials/PBR/pbrBaseMaterial.ts
brunobg: @sebavan we were talking about this earlier:

> >  BTW, any specific reason to make Material._EventInfo a static private variable, instead of using a local variable? It's not reentrant.
> I did it static to avoid creating a new object each time we need to notify the observers. Dealing with reentrancy would mean to create an object each time we notify and I fear it would be bad for perf/GC... I don't know what @sebavan things about this? Should we document the fact that the material events are not reentrant and one should not implement an observer that would in turn would lead to sending notifications to another/others observers or should we handle it in the proper way?

What do you suggest? I think a static field in Material, shared by everything, is not a great idea. Perhaps a non-static field on the material instance, combining all the `MaterialEventInfo*` interfaces would be better? But it might keep some stale data that won't be GC'ed.

src/Materials/material.ts
brunobg: holding on the decision of whether registration will be static or not to finish this one.

src/Materials/PBR/pbrBaseMaterial.ts
Popov72: Maybe:
```
keysFromPlugins: Array<{ name: string, type: string, default: any }>
```
If we agree we should move this to the `MaterialDefines` class instead, replace `keysFromPlugins` by `externalProperties` (better than `externalKeys` as we provide the full description of additional properties).

src/Materials/PBR/pbrBaseMaterial.ts
brunobg: With the static plugin registration these are always initialized.

src/Materials/PBR/pbrBaseMaterial.ts
Popov72: Here's what I wrote about this above:
> I did it static to avoid creating a new object each time we need to notify the observers. Dealing with reentrancy would mean to create an object each time we notify and I fear it would be bad for perf/GC... I don't know what @sebavan things about this? Should we document the fact that the material events are not reentrant and one should not implement an observer that would in turn would lead to sending notifications to another/others observers or should we handle it in the proper way?

I think we should simply use a single event info object and reuse it each time we notifiy the observers as I did at first and document it is not reentrant and wait to see if it's a problem in the future.

src/Materials/materialEvent.ts
Popov72: Yes, combining flags make no sense but it's a flag because it's how it works at notification level. It's a mask you pass to `notifyObservers` as the second parameter (I kept only the relevant bits):

```typescript
public notifyObservers(eventData: T, mask: number = -1, target?: any, currentTarget?: any, userInfo?: any): boolean {
    ...
    for (var obs of this._observers) {
        if (obs.mask & mask) {
                state.lastReturnValue = obs.callback(eventData, state);
            ...
        }
    }
}
```
As there's a `AND` that is performed between the object mask and passed mask, we must define our events as a bit field.

src/Materials/PBR/pbrBaseMaterial.ts
Popov72: Not always when using ES6 if the corresponding material plugin files are not imported.

But I don't see a way to do it otherwise, it's up to the user to not use the property (`clearCoat`, `sheen`, `subSurface`, etc) if they have not imported the corresponding class.

src/Materials/PBR/pbrBaseMaterial.ts
Popov72: > Perhaps a non-static field on the material instance, combining all the MaterialEventInfo* interfaces would be better?

That won't help with the reentrancy problem. As javascript is mono-threaded, having a static field will work the same as having a specific instance per material, that's why I suggested using a static to save some memory. But using a single non-static property is fine by me too.

Popov72:
> we could use type inference to ensure of the typings and remove all the required cast
Do you mean we replace:
```typescript
        if (eventState.mask & MaterialEvent.GetDisableAlphaBlending) {
            (eventState.userInfo as MaterialEventInfoGetDisableAlphaBlending).disableAlphaBlending ||= DisableAlphaBlending(material);
        }
```
by something like:
```
if (isGetDisableAlphaBlendingEvent(eventState.userInfo)) {
    eventState.userInfo.disableAlphaBlending ||= DisableAlphaBlending(material);
}
```
and declare somewhere:
```typescript
function isGetDisableAlphaBlendingEvent(userInfo: any): userInfo is MaterialEventInfoGetDisableAlphaBlending {
    return (MaterialEventInfoGetDisableAlphaBlending).disableAlphaBlending !== undefined;
}
```
That would work but we are doing an additional function call that we don't otherwise have, which means we are doing 20 function calls or so if the event we are handling is the last one we check...

As we are at it, as we know we won't combine multiple events in a single notification call, we could replace all the `if (eventState.mask & XXX)` by a switch/case, allowing a constant time handling of an event (at least in the event observer).

src/Materials/PBR/pbrBaseMaterial.ts
brunobg: Ah, I see. But this means that the current code will break all existing apps that import modules directly for tree shaking, right? In the current code, if the `PBR/index.ts` file isn't imported, all these plugins won't be. Perhaps `pbrBaseMaterial` should import them directly?

Popov72: Yes, to avoid a breaking change I think `PBRBaseMaterial` should import them all, but it means it won't be possible to tree-shake finely with PBR (for eg, only import sheen config if you only use sheen).

As 5.0 is quite a breaking change release anyway, maybe we can document this one and let people know they should import the right config classes depending on the PBR features they are using. It's an anoyance but in return you save some space in download size (which is the first reason to use ES6 I think). @sebavan / @deltakosh ?

src/Materials/PBR/pbrBaseMaterial.ts
sebavan: happy with this :-)

src/Materials/PBR/pbrBaseMaterial.ts
sebavan: yup let s check with @RaananW how could we solve this with the LTS system ?

pbrBaseMaterial in one could have all of them without where the other one could be a nullable type ?

src/Materials/PBR/pbrBaseMaterial.ts
sebavan: I am not sure to understand. Let s say each material instance is having one observable instance instead of the global one.

Then the material keeps a private EventInfo field and use it on the events being fired. This way we should not have any problems. 

Any issues we can document would in any case be better than GC :-)

src/Materials/materialEvent.ts
sebavan: Makes total sense !!!!

sebavan:
> Do you mean we replace:
> 
> ```ts
> if (eventState.mask & MaterialEvent.GetDisableAlphaBlending) {
>     (eventState.userInfo as MaterialEventInfoGetDisableAlphaBlending).disableAlphaBlending ||= DisableAlphaBlending(material);
> }
> ```
> 
> by something like:
> 
> ```
> if (isGetDisableAlphaBlendingEvent(eventState.userInfo)) {
>     eventState.userInfo.disableAlphaBlending ||= DisableAlphaBlending(material);
> }
> ```
> 
> and declare somewhere:
> 
> ```ts
> function isGetDisableAlphaBlendingEvent(userInfo: any): userInfo is MaterialEventInfoGetDisableAlphaBlending {
>     return (userInfo as MaterialEventInfoGetDisableAlphaBlending).disableAlphaBlending !== undefined;
> }
> ```
> 
> That would work but we are doing an additional function call that we don't otherwise have, which means we are doing 20 function calls or so if the event we are handling is the last one we check...
> 
> As we are at it, as we know we won't combine multiple events in a single notification call, we could replace all the `if (eventState.mask & XXX)` by a switch/case, allowing a constant time handling of an event (or at least we enable this possibility at the javascript compiler).

No sorry, I was thinking if we add constants instead of flags, we could maybe rely on typescript discriminated union types to match the event info type and the registration we do. So it would be all typing only without runtime cost. Let me see if I can mock it around in a typescript playground.

sebavan:
I was thinking to smthg like this https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgZwgJwG4EM4BsIB4AqAfFALxQDeAUFDVOgCb0BcUAFAMbpZZzrsDWLVvXTB0LfAEpSxVAHsAlvQA0UALboEgqADsArmrgppJWYvoBuarWDIFAc3sohIsRNUatLfYeMyo8kpWAL6UlKCQUACiqBA6wAj44NBkAAxQAD5QAIyZUABMVuHJ0bHxAJI6AGZypBTWNOxyampyOixwcnI46DohYRHQMXHAlTW5ZMMV1bUAZPW0UBBlwNkdXT19lKHFkVOjM-l1+2NzC7TLI-nr3RC9-bvQGmBgCjr2dVSLqSwnM8oNHK-FanbIAxbXUojU75AE7dhYTQIKAABQAQgAlACyohQCi45xoYDsGGA0AA+hoyXYuPsAIQseBINCYHC4P41QhFQHEuRk9hk+hQck6PkKKogfYEJYADzJOnoyP4EBAciq6nQLze9lUAFVZfLFRqte8ANr4AC6UEIrEu0xqSUg7iWIJmLF1khYgSFX0WNgAFgoEAA6Sm4mlYfbB2wOJzIW2umqqO0HB3JSRWRY7RbICD2IPUgBicmQUvwBriRuVqvVz1e7z1FYVyLr2vNVptKdOjogzs43F4AiEXbdUA9-m9noC5kJfuAgZDYep+MjK2DDHoHC4PD4-DoyOEonEUPtchMZiUycTch7GcB2dogLgb030l9fqg86DodFwHFkpWVhUlUcgoCaFo2hYWw9GgYI70BRYv0XX9-32VgwQoMDmladpP2QGCoDgzMP0-BcfzFCU0Iw0DwJwqD8IgK8RjWPCCKIhCA2-EUKIAkZ0JArCINw6DGJdZj6JgpidEhETCPgkjSK4lDKMA2FMNoyDWNgu8FKQ8i-xUvi1Jo7DNJEqTVgk7TiI-PTuIM3j4lYYzBLorSLJY8yxPiGSGLkmzCO2MIESRKAcWXLgUSwPR8x0Wcmh0BBoIFEtWCpPEuBYdFsXDFc3w4mh0ojYNc3zJKUGLUtAOAthexPVMz38UD2IKjUIqwEq8wLCqSyo1RbWBaEZmyc9MPYhTWqKldOrKotesAjCBvqmFRua+SSKmrgZu65BKrQtSlo5OR8lW-z7yCoA

src/Materials/PBR/pbrBaseMaterial.ts
Popov72: If we are using a single private eventInfo field per material with have a possible reentrancy problem. Suppose an event is raised by a method of the material M and an observer of this event is calling a method of the same material M which itself raises an event: that will reuse the same eventInfo instance (because there's a single one for the material) than the one used for the first notification, which may be a problem (and which will be a problem if both events are the same).

Regarding not using a static observable but an instance per material is probably good: for the material plugin manager a static instance is enough because it attaches to all notifications of all materials. But if some user code want to attach to events of a single material instance it's probably better that the material has its own observable instance.

sebavan: Yup totally agree. 

Let s document the reentrancy and keep one instance of each eventInfo type limiting the reentrancy to the same event recursively firing ?

And lets make the handler an instance thingy :-)

Popov72:
It's a bit too much for my basic Typescript knowledge!

One thing I don't understand is: you reimplement your own observable/observer mechanism? Because I don't understand how your design can fit into the existing observable/observer implementation in Babylon.js which is based on masks: the `observable.add` method takes a mask as a parameter (which lets the observer indicates the events it is interested in) that is **ANDed** with the mask you pass to `observable.notifyObservers` to know if the observers should actually be notified. That's how we can use a single observable to dispatch multiple events. Else, we would need to create as many observables as the number of events.

sebavan:
I was mostly only thinking of the extra shortcut functions on the material which can prevent casting everywhere and match a const to a type through the type map. Let s discuss it during our sync on Monday

brunobg:
> Let s discuss it during our sync on Monday

I'm on hold here while you think about it.

src/Materials/materialPluginManager.ts
```diff
@@ -0,0 +1,475 @@
...
+/**
+ * Injects plugins on a material.
+ * @param material The material to inject plugins into.
+ */
+function InjectPlugins(material: Material): void {
+    for (const [propertyName, factory] of _Plugins) {
+        _AddPluginToMaterial(material, propertyName, factory);
+    }
+
+    material._plugins = material._plugins.sort((a, b) => a.priority - b.priority);
```
Popov72, line 187: `sort` is in place, you don't need to re-set the value.

src/Materials/materialPluginManager.ts
```diff
@@ -0,0 +1,475 @@
...
+/**
+ * Adds a plugin to a specific material. Use instead of RegisterMaterialPlugin if you don't
+ * want to make the plugin available to all materials.
+ * @param material The material to inject plugins into.
+ * @param propertyName The plugin name
+ * @param factory The factor function, which returns a IMaterialPlugin or null if it's not applicable.
+ */
+export function AddPluginToMaterial(material: Material, propertyName: string, factory: PluginMaterialFactory): void {
+    _AddPluginToMaterial(material, propertyName, factory);
+    material._plugins = material._plugins.sort((a, b) => a.priority - b.priority);
```
Popov72, line 199: `sort` is in place, you don't need to re-set the value.

src/Materials/standardMaterial.ts
```diff
@@ -193,11 +197,25 @@ export class StandardMaterialDefines extends MaterialDefines
     public IS_REFRACTION_LINEAR = false;
     public EXPOSURE = false;
 
-    constructor() {
+    protected _keysFromPlugins?: string[];
```
Popov72, line 200: Same comment for `_keysFromPlugins`, ctor and `rebuild` function than for the PBR base material.

Popov72 (changes requested):

Popov72:
Note there are a number of other points we are about to discuss, will let you know the outcome!

Popov72:
There will be a big refactoring, I will update my local repo tomorrow and will let you know when it's done!

Popov72:
Here are the changes:

https://github.com/BabylonJS/Babylon.js/compare/master...Popov72:Corollarium-custom-base-material (comparison between Master and my branch).

https://github.com/Corollarium/Babylon.js/compare/custom-base-material...Popov72:Corollarium-custom-base-material (comparison between your branch and mine, but you should sync your branch with master because else there are too many changes).

We have changed things a lot, happy to provide some explanations if needed!

We are still going to discuss with @sebavan, maybe there will be some other updates, but it should globally be ok.

Here's the updated PG that works with the latest changes:
* JS: http://localhost:1338/Playground/index-local.html#P8B91Z#27
* TS: http://localhost:1338/Playground/index-local.html#P8B91Z#29

sebavan:
Looks real good to me !!! :-)

brunobg:
I synced the PR with @Popov72's branch. 

> [17:01:59] New circular dependencies in core: 7
> [17:01:59]     From: 'src/Materials/material.detailMapConfiguration.ts' To: 'src/Materials/PBR/pbrBaseMaterial.ts'
> [17:01:59]     From: 'src/Materials/PBR/pbrBaseMaterial.ts' To: 'src/Materials/PBR/pbrAnisotropicConfiguration.ts'
> [17:01:59]     From: 'src/Materials/PBR/pbrBaseMaterial.ts' To: 'src/Materials/PBR/pbrBRDFConfiguration.ts'
> [17:01:59]     From: 'src/Materials/PBR/pbrBaseMaterial.ts' To: 'src/Materials/PBR/pbrClearCoatConfiguration.ts'
> [17:01:59]     From: 'src/Materials/PBR/pbrBaseMaterial.ts' To: 'src/Materials/PBR/pbrSheenConfiguration.ts'
> [17:01:59]     From: 'src/Materials/PBR/pbrBaseMaterial.ts' To: 'src/Materials/PBR/pbrSubSurfaceConfiguration.ts'
> [17:01:59]     From: 'src/Materials/standardMaterial.ts' To: 'src/Materials/material.detailMapConfiguration.ts'

Using `declare type` or `import type` won't fix this one:

> 'PBRBaseMaterial' cannot be used as a value because it was imported using 'import type'.

Happens here:

```ts
export function createDetailMapPlugin(material: Material): Nullable<MaterialPluginBase> {
    if (material instanceof PBRBaseMaterial || material instanceof StandardMaterial) {
```

Popov72:
I have fixed the circular dependency problem.

Good news on the performance front: I expected the code to be slightly slower or at best not significantly slower, but it is actually faster with the material manager (at least on the 2 PGs I have tested)! I don't really understand why as we essentially did some code reorganization of existing code, and we even added an observable in-between so we have some additional calls we did not have previously. It seems Chrome did like the refactoring...

Anyway, here's what I get:

Test             | prod      | local (this PR)      | Improvement over prod | local (single plugin)  | Improvement over prod |
-------------|----------|----------|-----------------------|----------------------------|--|
Wipeout     | 2180ms | 2050ms |  -6% | | |
Powerplant | 3600ms | 3150ms | -12.5% | | |
Powerplant PBR | 4389ms | 3728ms | -15% | 3398ms| -22%|

PGs:
* Wipeout: https://playground.babylonjs.com/#0HQJWK#23 (I tested after we press the "Demo" button)
* Powerplant: https://playground.babylonjs.com/#SYQW69#923 (there are 81 standard materials in this scene)
* Powerplant PBR: https://playground.babylonjs.com/#SYQW69#933. This is the same PG than the one above but using PBR materials instead of standard materials (there are 285 PBR materials in the scene)

I tooks those PGs because they are more typical of a real scene than a simple test with a single object at the center of the screen.

The time measured is the time spent in the javascript thread during a 5s snapshot in the **Performance** tab of Chrome (Version 96.0.4664.45 (Official Build) (64-bit)). So, lower values means better perf.

As you can see, the savings can be quite important!

I have also benchmarked the Powerplant PBR scene with a single plugin (the brdf): we can't currently get rid of the PBR material plugins but it will be possible in the future (for my tests, I simply commented out in the constructor the instancing of all plugins save for the brdf one). As you can see, there are some additional savings to have a single plugin instead of 6: 3398ms vs 3728ms (it's 9% faster).

sebavan:
This  is great and it might come from v8 optim cause the code paths are all the same maybe lol

Popov72:
Example of a Volumetric Fog plugin material (could go into the doc):
* with standard materials: http://localhost:1338/Playground/index-local.html#SYQW69#935
* with PBR materials (same plugin): http://localhost:1338/Playground/index-local.html#SYQW69#939

Popov72:
@sebavan **Module Tests** fail because the package for `minStandardMaterial` is now a bit too big:

```
[13:24:18] New size: 975861 bytes is bigger than baseline size : 965601 bytes on minStandardMaterial.
```

brunobg:
Branch synced again.

> Example of a Volumetric Fog plugin material (could go into the doc):

I'll write the docs, PR coming soon. 

Somethings on this PR that I can help with too, tell me if you agree with them:

- I can add `#define CUSTOM_xxx` to shaders that are missing them, such as the materials library. 
- add a `Fragment_After_FragColor`. 
- use the same variable names on certain shader points. Like `finalColor` and `color`, etc.

Popov72:
> I can add #define CUSTOM_xxx to shaders that are missing them, such as the materials library.

Ok for me. Note that some #defines won't be applicable, depending on the material.

> add a Fragment_After_FragColor.

I would rather add a `#define CUSTOM_FRAGMENT_MAIN_END` just before the end of the shader, in the same way we already have `#define CUSTOM_VERTEX_MAIN_END` for the vertex shader.

> use the same variable names on certain shader points. Like finalColor and color, etc.

I'm not sure. What do you think @sebavan? One important thing to note (and to add to the docs) is that we won't guarantee that using the shader point names / the regular expressions to update code with a material plugin will work accross Babylon.js versions. We reserve the possibility to update our shader code in a way that would break backward compatibility with those features if we have to.

sebavan:
> Example of a Volumetric Fog plugin material (could go into the doc):

This should definitely go in the doc :-)

> use the same variable names on certain shader points. Like finalColor and color, etc.

I do not think we should do that from what @Popov72 mentioned or at least not in the same PR. It is nevertheless a nice to have. But the doc will need to specify in BIG RED BOLD :-) that we will never guarantee shader back compat.

> @sebavan Module Tests fail because the package for minStandardMaterial is now a bit too big:

This sounds weird as we do not add much by default ??? but it would all go away with the none LTS version so it should be ok. Could you check with webpack analyser that we are not pulling unexpected items.

Also @Popov72 about the perf, were you running them against local dev as well in another branch ? as I am afraid it could be due to unbundled code vs bundled one ?

Popov72:
> This sounds weird as we do not add much by default ???

Yes, we do add a bit of code: the `MaterialPluginBase` / `MaterialPluginManager` classes, the type definitions for the events + structs, the export of the standard plugin factories. The 5 files involved amount for 28ko. After compression, I think it explains the growth by 7ko that we observe.

> Also @Popov72 about the perf, were you running them against local dev as well in another branch ? as I am afraid it could be due to unbundled code vs bundled one ?

You're right, it seems the prod was not updated with all the latest PRs yet! I have updated my post with the benchmarks.

sebavan:
I am actually now less afraid by the size than I am with the perfs :-( let s have a look offline on Monday as 6% sounds quite heavy.

brunobg:
Hey everyone, I'm curious, any news about this? I was afraid a lot of function calls for NOOPs with the observable approach would hit performance. Perhaps a simpler array of callbacks would solve the performance issues and keep essentially the same code architecture and not be too invasive on `Material`? 

I'm on hold to make those changes mentioned above. This PR would be such an awesome improvement :)

Popov72:
We are still working actively on this :)

There are indeed some perf problems, so we have replaced the observable by a callback and I'm trying to finalize the whole thing. I hope to push the latest changes tomorrow!

Popov72:
My branch has been updated:
https://github.com/BabylonJS/Babylon.js/compare/master...Popov72:Corollarium-custom-base-material

To be discussed on monday in our core meeting.

Popov72:
Ok, we discussed about this, I still have some small changes to do and we should be good!

I will do them before the end of the day.

Popov72:
I have updated the code, it should now be ok!

I will make another post with the perf results.

Popov72:
On the perf front (measured with the PowerPlant PG), compared to the code currently in the master branch (when perf difference is below 1% in one way or the other I assume we are in the error margin and consider we have the same perf):
* Standard material: same perf without the detail plugin, around 1.2% slower when using it
* PBR material:
  * no plugin (only BRDF), clearCoat only, SubSurface only, Sheen+Anisotropy+Detail map: between 1.2% and 2.7% faster, depending on the case (2.7% for the "no plugin" case, which is the one happening most often)
  * clearCoat + SubSurface, Sheen + SubSurface: same perf
  * all plugins (clearCoat, SubSurface, Anisotropy, Sheen, Detail map): 2% slower. Note however that no one in their right mind would ever use all the plugins in the same material...

brunobg:
> Once you have updated the PR you can switch it to "Ready for review" so that people can review it. Thanks!

I'll add the `CUSTOM_FRAGMENT_MAIN_END` point as well as the insertion points on existing `.fx` files and set it to review. Awesome work @Popov72 :)

brunobg:
I inserted the `BEGIN/END` #defines on all relevant fragment/vertex shaders, which IMHO are the most relevant. Is `CUSTOM_x_DEFINITIONS` important too? I can add it before every `void main(void)` if it's relevant. 

Other than that I can only thank you all for doing all the work for me :D I'll send a PR with documentation and examples over the next few days.

sebavan:
I will let @Popov72 do a first review pass and I ll do the second one with @deltakosh but it overall looks pretty amazing to me !!!

Popov72:
You need to add this to the constructor if you want that the plugin works whatever the default value of `_isDisabled` is:
```javascript
this._enable(this._isEnabled);
```
Going to review the PR now!

src/Materials/materialPluginEvent.ts
```diff
@@ -0,0 +1,112 @@
...
+    animatables: IAnimatable[];
+};
+
+/** @hidden */
+export type MaterialPluginGetActiveTextures = {
+    activeTextures: BaseTexture[];
+};
+
+/**
+ * Properties used by the FillRenderTargetTextures event
```
Popov72, line 76: Don't need to be visible, you can replace the doc by:
```
/** @hidden */
```

src/Materials/materialPluginEvent.ts
```diff
@@ -0,0 +1,112 @@
...
+    renderTargets: SmartArray<RenderTargetTexture>;
+};
+
+/** @hidden */
+export type MaterialPluginHasRenderTargetTextures = {
+    hasRenderTargetTextures: boolean;
+};
+
+/**
+ * Properties used by the HardBindForSubMesh event
```
Popov72, line 88: Don't need to be visible, you can replace the doc by:
```
/** @hidden */
```

materialsLibrary/src/custom/customMaterial.ts
```diff
@@ -21,6 +21,7 @@ export class ShaderSpecialParts {
     public Fragment_Begin: string;
     public Fragment_Definitions: string;
     public Fragment_MainBegin: string;
+    public Fragment_MainEnd: string;
```
Popov72, line 24: It seems `pbrCustomMaterial` has not been updated in the same way than `CustomMaterial` with `Fragment_MainEnd`?

Popov72 (approved):
Good for me, only nits or so. Great job!

src/Materials/material.ts
```diff
@@ -714,6 +725,44 @@ export class Material implements IAnimatable {
...
+            MaterialPluginBindForSubMesh &
+            MaterialPluginGetAnimatables &
+            MaterialPluginGetActiveTextures &
+            MaterialPluginFillRenderTargetTextures &
+            MaterialPluginHasRenderTargetTextures &
+            MaterialPluginHardBindForSubMesh
+    = { } as any; // will be initialized before each event notification
+
+    /** @hidden */
+    public _callbackPluginEvent: (id: number, info: MaterialPluginGetActiveTextures | MaterialPluginGetAnimatables | MaterialPluginHasTexture | MaterialPluginDisposed | MaterialPluginGetDefineNames | MaterialPluginPrepareEffect | MaterialPluginPrepareUniformBuffer) => void = () => void(0);
```
sebavan, line 752: totally unsure about naming here :-)

_pluginCallbackGeneric
_pluginCallbackIsReadyForSubMesh
...
 ?

Not sure event needs to be in but I am stupid with Naming :-)

sebavan (commented):
Amazing Job I only have only naming topic but it looks great !!!!!!!!

@deltakosh you can do your pass :-)

sebavan (approved):

deltakosh (approved):
Impressive update! Can't wait to read the doc

Can I recommend adding a dedicated unit test?

Popov72:
Note that we will need to do something for the "Module Tests" failing as it is a required step.

src/Materials/material.ts
brunobg: `_pluginCallbackGeneric` or `_pluginCallbackOther` make sense to me, vote your preferences and I'll change it

brunobg:
> Impressive update! Can't wait to read the doc

Life is a bit hectic, but PR will come some time next week.

src/Materials/material.ts
sebavan: @Popov72  ???

sebavan:
> Note that we will need to do something for the "Module Tests" failing as it is a required step.

Nah np I ll merge and make a nightly to make it go away :-)

Let s do the tiny rename first but this is huge !!!!!!

sebavan:
> Impressive update! Can't wait to read the doc
> 
> Can I recommend adding a dedicated unit test?

This is already fully tested by the Standard and PBR mat themselves who are relying on it :-)

brunobg:
As I'm writing the docs I realized CUSTOM_VERTEX_DEFINITIONS/CUSTOM_FRAGMENT_DEFINITIONS are important. I'll send a separate PR with these #defines everywhere once this one is accepted, only adding these definitions, to avoid another round of approvals in this PR.

PR #11414, opened 2021-11-09
Introduce color tangent based animations

deltakosh:


deltakosh:
sharing is not transparent as it will add code and tests where we want these functions to be super fast

Will change the const

sebavan (approved):

Issue #10983, opened 2021-09-01
[ACE] ACE Improvements


PirateJC:
This is a list of random suggestions for making ACE more user friendly!

Priority 1 list:
- [x] Pressing the key button in the curve editor, should always grab the value from that property in the 3D scene. For example, a sphere is on the groundplane on frame 10, you move the sphere up 3 units in position.y using the gizmo. With the position.y animation curve active, hitting the key button will create a new key at the playhead with a value of 3 units. New button needed for this behavior. It will not work on Animation Groups (@PatrickRyanMS for the button)
- [ ] When adding a new animation, the property box should change to a drop down selector that contains a list of ALL animatable properties for the selected object. The Dropdown list should allow you to type and filter the results. At the top of the dropdown list we should display the following items first at the top of the list: "Custom", "Position (Vector3)", "Rotation(Vector3)", "Scaling(Vector3)". After those 4 items a line visually separates all other properties. If we do not know the selected object (ie. Animation Group) we default to the "custom" behavior. Custom behavior displays a box to type in properties as well as a drop down box for selecting the "type." Any property where we already know the type, it's chosen for you.
- [x] Can we have live updates in the scene when we change a key's value? Currently the only way to see updates in the scene is by moving the playhead.
- [x] The 'jump to the next key' buttons seem to take ALL keys for all animations into account whether a curve is displayed in the curve editor or not. For example in the following image:
![image](https://user-images.githubusercontent.com/1668182/131592051-d4955aba-6f2c-4080-b6ec-67c84526d53f.png)
hitting this button:
![image](https://user-images.githubusercontent.com/1668182/131592064-af556666-878e-420a-858c-9febe6c0a6dd.png)
results in the following:
![image](https://user-images.githubusercontent.com/1668182/131592094-4a08dd86-5162-4a14-a764-94b532bf6212.png)
That's because there is a key at this point on a DIFFERENT animation:
![image](https://user-images.githubusercontent.com/1668182/131592131-91252c99-f8c7-4d37-a42f-798a97a7a6ac.png)
We should only be evaluation keys to navigate to (with those buttons) based on what keys are visible on the canvas.
- [x] When no key is selected, rather than hiding UI elements (frame and value box, tangent buttons, etc), let's just grey them out or make the boxes empty.
- [x] To deselect a key, you should be able to click on the canvas. Currently you have to CTRL+click+drag to deselect
[...]

PR #11428, opened 2021-11-10
Ace improvements

carolhmj:
This PR addresses a series of improvements from #10983. These are:

* We need new buttons that allow you to step one frame forward or backward.
* Left and Right arrow keys should also step the playhead forward or backward by one frame.
* We should have a 'step curve' option that holds a key value until the next key. This is useful for the traditional animation workflow of blocking poses first and then moving to splining the in-betweens.
* Hide tangents for Color animations, since they aren't interpolated yet
* Fix playhead showing NaN when animation has multiple keys on the same frame

There are a lot of changed files but most of them are SVGs, so don't be scared! 😄


dist/what's new.md
```diff
@@ -118,6 +118,7 @@
 - Added more functionality and options to the skeleton debug panel. ([Pryme8](https://github.com/Pryme8))
 - Along with bone index it is now possible to select a bone using its name when viewing bone weights ([#9117](https://github.com/BabylonJS/Babylon.js/issues/9117)) ([RaananW](https://github.com/RaananW))
 - Gradient nodes from NME can now be set to be visible in the inspector ([msDestiny14](https://github.com/msDestiny14))
+- ACE usability improvements and addition of new Step Tangent interpolation. ([carolhmj](https://github.com/carolhmj))
```
deltakosh, line 121: This is the wrong file:( the good one is in preview release

deltakosh (changes requested):
I will check tomorrow;)

thanks a ton!

dist/what's new.md
carolhmj: 😭
thanks for the heads up!

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/animationCurveEditorComponent.tsx
```diff
@@ -61,7 +61,7 @@ export class AnimationCurveEditorComponent extends React.Component<
     private _onKeyDown(evt: KeyboardEvent) {
         switch (evt.key) {
             case "Delete":
-                if (this.props.context.activeKeyPoints?.length) {
+                if (this.props.context.activeKeyPoints?.length && !this.props.context.focusedInput) {
```
deltakosh, line 64: can you explain why we need this test?

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/bottomBar/bottomBarComponent.tsx
```diff
@@ -60,7 +116,15 @@ IBottomBarComponentState
                 {
                     this.props.context.activeAnimations.length > 0 &&
                     <div id="bottom-bar-total">
-                        {this._renderMaxFrame()}
+                        <TextInputComponent 
+                        isNumber={true}
+                        value={this.state.clipLength}
+                        tooltip="Clip Length"
+                        id="clip-range"
+                        onValueAsNumberChanged={(newValue, isFocused) => {
```
deltakosh, line 124: I would recommend to also validate it when user hits enter

deltakosh (changes requested):
Excellent work.. Just a question and good to go! (please make sure next time to do smaller PR)

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/animationCurveEditorComponent.tsx
carolhmj: I noticed that if I selected a key and tried to use the Delete key for changing the value, the key was deleted, so the test is to avoid this situation.

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/bottomBar/bottomBarComponent.tsx
carolhmj: Added it!

deltakosh (approved):
excellent work!

sebavan:
Awesome one !!! thanks a lot

PR #11429, opened 2021-11-10
Update Tools.GetAbsoluteUrl to work in web workers

bghgary:
See https://forum.babylonjs.com/t/cannot-import-draco-compressed-mesh-in-a-web-worker/25125

### DIFF
--- a/src/Misc/tools.ts
+++ b/src/Misc/tools.ts
@@ -938,11 +938,10 @@ export class Tools {
      * @param url the input url
      * @return the absolute url
      */
-    public static GetAbsoluteUrl(url: string): string {
-        const a = document.createElement("a");
-        a.href = url;
-        return a.href;
-    }
+    public static GetAbsoluteUrl: (url: string) => string =
+        (typeof document === "object") ? (url) => { const a = document.createElement("a"); a.href = url; return a.href; } :
+        (typeof URL === "function") ? (url) => new URL(url, location.origin).href :
+        (url) => `${location.origin}/${url}`;
 
     // Logs
     /**
--- a/src/Misc/tools.ts
+++ b/src/Misc/tools.ts
@@ -934,9 +934,7 @@ export class Tools {
     }
 
     /**
-     * Gets the absolute url.
-     * @param url the input url
-     * @return the absolute url
+     * Function used to get the absolute url. Override for custom implementation.
      */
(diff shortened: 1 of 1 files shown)


src/Misc/tools.ts
```diff
@@ -938,11 +938,10 @@ export class Tools {
...
      */
-    public static GetAbsoluteUrl(url: string): string {
-        const a = document.createElement("a");
-        a.href = url;
-        return a.href;
-    }
+    public static GetAbsoluteUrl: (url: string) => string =
+        (typeof document === "object") ? (url) => { const a = document.createElement("a"); a.href = url; return a.href; } :
+        (typeof URL === "function") ? (url) => new URL(url, location.origin).href :
+        (url) => `${location.origin}/${url}`;
```
bghgary, line 944: This will still fail if `location` is not available (like in a node.js scenario), but people can simply override this function if it needs to work.

sebavan: let s return url by default instead of a crash ?

bghgary: I thought about that, but is that good? It may return a relative url when the function is called (GetAbsoluteUrl). Another thought I had is to add a static variable (e.g., BaseUrl that can be used to as a fallback or maybe override).

sebavan: Makes sense the static one could be prefixed if it does not already start with http... I kind of like it :-)

bghgary: You mean as a fallback or override? Also, what should the default of the BaseUrl be? :-)

RaananW: As you say, it is possible to override this function.  Maybe a warning would be good - if no location is available, warn that it is not there and that this function can be adjusted.
Otherwise, return `/` as the base URL, IMO it makes the most sense.

bghgary: I forgot that just using string concat is wrong. If the incoming URL is already absolute, this will result in a bad URL. Detecting whether a URL is absolute or relative is not trivial. I think we should perhaps just throw an error for the fallback case with a message saying you can override the function?

RaananW: I think this is fine. Just log the error

RaananW:
@bghgary - can you check why the doc check fails?

PR #11430, opened 2021-11-11
Fix TargetCamera rotation issue

BlakeOne:
This patches fixes an issue demonstrated in the below playground where the camera's rotation is misbehaving.

Playground: https://playground.babylonjs.com/#F68Z6N#4
Forum: https://forum.babylonjs.com/t/weird-behavior-with-universalcamera-and-rotation-property/25116/25

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -235,6 +235,7 @@
 
 ## Bugs
 
+- Fix an issue with `TargetCamera`'s rotation caching ([BlakeOne](https://github.com/BlakeOne))
 - Fix incorrect GUI.TextBlock width when resizeToFit is true & fontStyle is italic ([Kalkut](https://github.com/Kalkut))
 - Fix CubeTexture extension detection when rootUrl has a query string ([civa86](https://github.com/civa86))
 - Fix issue with the Promise polyfill where a return value was expected from resolve() ([Deltakosh](https://github.com/deltakosh))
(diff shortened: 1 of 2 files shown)


BlakeOne:
Previously _rotateUpVectorWithCameraRotationMatrix was called only when the z property of the camera's rotation vector changed, but with this patch it's called if any property of the rotation vector changed (when quaternion isn't set).

deltakosh:
Yep this fix was suspicious :) Thanks for your appreciated help :)

BlakeOne:
Happy to help. :) This one has me vexed because it seems that theoretically, if it works when the user converts the euler vector to quaternion manually  then it should be able to also work when BJS does that conversion for us, so I'll keep at it. :)

PS I updated the fix and got one of the validation fails corrected, the only one that seems different to me on visual inspection. :)

deltakosh:
Good job! Thanks a lot

deltakosh (approved):

0x53A:
Thank you for fixing this!

I can see that it was merged 8 days ago, and ``5.0.0-alpha.61`` was released 4 days ago, should this already be deployed to the playground?

I'm asking because on my system (Windows 10 x64 + Edge-Chrome v95.0.1020.53), it's still wrong:

babylon-5-a61

sebavan:
It was addressing only one part and Blake on the forum is sill looking into the others with @PolygonalSun

PR #11436, opened 2021-11-11
serialize isEnabled for lights and cameras

andreasplesch:
for discussion see https://forum.babylonjs.com/t/light-serialization/25036

https://www.babylonjs-playground.com/#PM0XL0 has a few simple tests for light.
I could generate similar tests for camera. `camera.setEnabled(false)` does not seem to have an effect.

I have build artefacts if useful.

Even with only a few lines, here are some questions:

- Remove/add comments ?
- Use of `in` operator ? It does not seem common in the code base but is supported by IE11.
- replace `if` with ternary ? 
```js
node.setEnabled(
    "isEnabled" in parseNode ? parsedNode.isEnabled : true
)
```

### DIFF
--- a/src/Cameras/camera.ts
+++ b/src/Cameras/camera.ts
@@ -1385,10 +1385,7 @@ export class Camera extends Node {
         if (parsedCamera.isEnabled !== undefined) {
             camera.setEnabled(parsedCamera.isEnabled);
         }
-        else {
-            camera.setEnabled(true);
-        }
-
+        
         return camera;
     }
 }
--- a/src/Cameras/camera.ts
+++ b/src/Cameras/camera.ts
@@ -1238,6 +1238,9 @@ export class Camera extends Node {
         SerializationHelper.AppendSerializedAnimations(this, serializationObject);
         serializationObject.ranges = this.serializeAnimationRanges();
 
+        // isEnabled
+        serializationObject.isEnabled = this.isEnabled();
+
         return serializationObject;
     }
 
--- a/src/Cameras/camera.ts
+++ b/src/Cameras/camera.ts
@@ -1385,7 +1385,7 @@ export class Camera extends Node {
         if (parsedCamera.isEnabled !== undefined) {
             camera.setEnabled(parsedCamera.isEnabled);
         }
-        
+
         return camera;
     }
 }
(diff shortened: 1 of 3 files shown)


sebavan:
I guess you could do smthg like this
```
if (parsedCamera.isEnabled !== undefined) {
    camera.setEnabled(parsedCamera.isEnabled);
}
```

You do not need comments in the serialized part as it is expressive enough, but might be good to add one in the parse section to let the next dev know the undefined check is for back compat.

andreasplesch:
> ```js
> // Check if isEnabled is defined to be back compatible with prior serialized versions.
> if (parsedCamera.isEnabled !== undefined) {
>     camera.setEnabled(parsedCamera.isEnabled);
> }
> ```
> 
> You do not need comments in the serialized part as it is expressive enough, but might be good to add one in the parse section to let the next dev know the undefined check is for back compat.

Makes sense. I adopted these and added a what's new line.

andreasplesch:
Here is a test scene: https://www.babylonjs-playground.com/#PM0XL0#2

It tests serialization and parsing, also for the back compat. case.

src/Cameras/camera.ts
```diff
@@ -1379,6 +1381,14 @@ export class Camera extends Node {
             scene.beginAnimation(camera, parsedCamera.autoAnimateFrom, parsedCamera.autoAnimateTo, parsedCamera.autoAnimateLoop, parsedCamera.autoAnimateSpeed || 1.0);
         }
 
+        // Check if isEnabled is defined to be back compatible with prior serialized versions.
+        if (parsedCamera.isEnabled !== undefined) {
+            camera.setEnabled(parsedCamera.isEnabled);
+        }
+        else {
+            camera.setEnabled(true);
```
deltakosh, line 1389: not required as the node will be enabled by default

src/Lights/light.ts
```diff
@@ -710,6 +712,14 @@ export abstract class Light extends Node implements ISortableLight {
             scene.beginAnimation(light, parsedLight.autoAnimateFrom, parsedLight.autoAnimateTo, parsedLight.autoAnimateLoop, parsedLight.autoAnimateSpeed || 1.0);
         }
 
+        // Check if isEnabled is defined to be back compatible with prior serialized versions.
+        if (parsedLight.isEnabled !== undefined) {
+            light.setEnabled(parsedLight.isEnabled);
+        }
+        else {
+            light.setEnabled(true);
```
deltakosh, line 720: same

deltakosh (changes requested):
all good. Simpy one line of code that could be removed

src/Lights/light.ts
andreasplesch: Right, the back compat. case would be covered by default.

PR #11439, opened 2021-11-12
Fix #11431

### LINKED ISSUE #11431: Reload an exported glTF model, its vertex winding and node transform are incorrect.
**Repro**
- Bug repro on [https://playground.babylonjs.com/#CZQA3A](https://playground.babylonjs.com/#CZQA3A):

vs
```

var delayCreateScene = function () {
    
    let fanDescendants;
    BABYLON.SceneLoader.ImportMesh(
        null,
        "/Playground/scenes/",
        "fileName.glb",
        scene,
        function (meshes) {          
            scene.createDefaultCameraOrLight(true, true, true);
            scene.createDefaultEnvironment();  
            console.log(meshes[0].position);
            fan = meshes[0];
            fan.position.x = 20;
            fanDescendants = fan.getDescendants(false);  
            console.log(fanDescendants);
    });
    
    return scene;
};
```
- Expected result:

The glTF model which is exported from glTFExported should match to the original glTF model.

- Current result:

The glTF model after exporting will have wrong vertex winding and the attached node transform is wrong.

**Screenshots**
The screenshot after loading from glTFLoader
Screen Shot 2021-11-10 at 11 52 30 PM

The screenshot after loading from an exported glTF model
Screen Shot 2021-11-10 at 11 53 51 PM

 - OS: [e.g. iOS] Mac OS
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] Version 95.0.4638.69

**Additional context**
Per my investigation, I notice BabylonJS uses left-handed coordinate system and glTF model is right-handed system, so we transform its root node at [glTFLoader.ts](https://github.com/BabylonJS/Babylon.js/blob/8403b1ea3f4ea9265de5d0f7ccbcc75023db300c/loaders/src/glTF/2.0/glTFLoader.ts#L470). However, in glTFExporter, we didn't invert the transform correctly. We are supposed to transform back from left to right handed system when exporting. After adding `glTFNode.scale = [1, 1, 1];` at [glTFExporter.ts](https://github.com/BabylonJS/Babylon.js/blob/8403b1ea3f4ea9265de5d0f7ccbcc75023db300c/serializers/src/glTF/2.0/glTFExporter.ts#L1740), the vertex winding issue is resolved at least. I need to dig into more about how to correct the root transform when applying attached nodes.

deltakosh:


serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -1632,14 +1632,15 @@ export class _Exporter {
...
+
             // Transform
             let matrix = node.getWorldMatrix();
             let matrixToLeftHanded = Matrix.Compose(this._convertToRightHandedSystem ? new Vector3(-1, 1, 1) : Vector3.One(), Quaternion.Identity(), Vector3.Zero());
             let matrixProduct = matrix.multiply(matrixToLeftHanded);
             let matrixIdentity = Matrix.IdentityReadOnly;
 
             for (let i = 0; i < 16; i++) {
-                if (Math.abs(matrixProduct.m[i] - matrixIdentity.m[i]) > Epsilon) {
+                if (matrixProduct.m[i] * matrixIdentity.m[i] < 0) { // checking if they have the same sign
```
bghgary, line 1643: I wonder if we can just check the matrix determinant to figure this out instead of all this math?

bghgary (approved):

serializers/src/glTF/2.0/glTFExporter.ts
deltakosh: I tried but it is not working :(

bghgary: really? that's surprising

deltakosh: actually it worked, I'm simply dumb

Popov72 (approved):

PR #11441, opened 2021-11-12
[GE] Transparency Texture!

msDestiny14:
image

https://github.com/BabylonJS/ThePirateCove/issues/28


RaananW:
All checks are failing :-)

Want to check that?

RaananW (commented):
The only thing that worries me here is the asynchronity in a synchronous function.
Would it be possible to change the function to an async function and wait with everything else until this finishes? it seems to be the initialization function and is probably important. 
This will also allow you to log errors and make sure everything is working as expected before the usr strarts working with the tool

guiEditor/src/diagram/workbench.tsx
```diff
@@ -747,21 +749,33 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
         this.artBoardBackground = new Rectangle("Art-Board-Background");
         this.artBoardBackground.width = "100%"
         this.artBoardBackground.height = "100%";
-        this.artBoardBackground.background = "white";
+        this.artBoardBackground.background = "transparent";
         this.artBoardBackground.thickness = 0;
 
         this.globalState.guiTexture.addControl(this.artBoardBackground);
+
+        let nodeMaterial = new NodeMaterial("NodeMaterial", this._scene);
```
sebavan, line 765: nit: const

guiEditor/src/diagram/workbench.tsx
```diff
@@ -747,21 +749,33 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+        await nodeMaterial.loadAsync("GUIEditorNodeMaterial.json").then(() => {
```
sebavan, line 766: why adding .then if you use the await syntax, this is kind of confusing . using await here will simplify readability between .then and the following code.

guiEditor/src/diagram/workbench.tsx
```diff
@@ -747,21 +749,33 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+            nodeMaterial.build(true);
+            this._textureMesh.material = nodeMaterial;
+            if (nodeMaterial) {
+                let block = nodeMaterial.getBlockByName("Texture") as TextureBlock;
```
sebavan, line 770: nit: const

guiEditor/src/guiEditor.ts
```diff
@@ -61,15 +61,16 @@ export class GUIEditor {
         ReactDOM.render(graphEditor, hostElement);
         // create the middle workbench canvas
         if (!globalState.guiTexture) {
-            globalState.workbench.createGUICanvas();
-            if (options.currentSnippetToken) {
-                try {
-                    globalState.workbench.loadFromSnippet(options.currentSnippetToken);
+            globalState.workbench.createGUICanvas().then(() => {
```
sebavan, line 64: try using await and mark Show(...) as async so that the code might be easier to understand and ppl relying on guiEditor would be sure when everything is ok

sebavan (changes requested):

guiEditor/src/guiEditor.ts
RaananW: This was my suggestion - not to use await in this function and not changing its signature while still allowing async operation to be logged correctly.

guiEditor/src/diagram/workbench.tsx
RaananW: Yep, totally agree on this one :-)

guiEditor/src/guiEditor.ts
sebavan: I wonder if we should do it while we do not have back compat issue to deal with so far ? mainly in case we have more async part during init in the future ?

msDestiny14: For clarification, Should I make the change? I do like the ability to be able to log should something go wrong.

sebavan: I let @RaananW chose for this one :-) not knowing enough of the rest of the project

RaananW: @sebavan is right - there is no issue with adding async to the show function.
Make the change - the code will be cleaner

msDestiny14: How's this. :)

guiEditor/src/guiEditor.ts
```diff
@@ -61,14 +61,15 @@ export class GUIEditor {
         ReactDOM.render(graphEditor, hostElement);
         // create the middle workbench canvas
         if (!globalState.guiTexture) {
-            globalState.workbench.createGUICanvas();
+            await globalState.workbench.createGUICanvas();
             if (options.currentSnippetToken) {
                 try {
                     globalState.workbench.loadFromSnippet(options.currentSnippetToken);
```
sebavan, line 67: this is async as well so it should probably be await ...

PR #11442, opened 2021-11-12
ThinEngine.MinimizeMemoryFootprint

syntheticmagus:
Sketch of mechanism for bouncing the texture loaders to release those resources when they're not being used.

Not sure the best way to characterize this without exposing implementation details, hence the generic "MinimizeMemoryFootprint" name. Philosophically, what this is trying to do is let developers inform the engine that a change in the engine's usage is about to happen (for example, we might be done loading the scene now and expect to go into a long period of rendering without loading) such that this is a good time to let go of certain resources that may have been cached. Right now, it's a generic name with a specific purpose, so this isn't a great fit; we should probably either decide to make specific and consequently expose (at least indirectly) the existence of the `_TextureLoaders` through the contract, or we should go generic and decide (even if we don't do it now) that this method may also release other kinds of resources in the future, too. Assuming we want this change at all, that is 😃; would love to discuss the preferred direction.

This came up while looking into a [forum question about KTX2](https://forum.babylonjs.com/t/how-to-release-ressources-for-ktx2-textures/25176), during which it became clear that the texture loaders can hold onto substantial resources statically. This isn't a fix for that issue -- there appears to be a memory leak in the KTX2 loader Web worker, and fixing that leak is the actual fix -- but this could be used as a workaround for that issue as well, especially if the leak turns out to be inside one of the KTX2 WASMs itself, which we don't produce ourselves and so can't fix directly.

### DIFF
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -175,8 +175,30 @@ export class ThinEngine {
         { key: "Mac OS.+Chrome\/72", capture: null, captureConstraint: null, targets: ["vao"] }
     ];
 
+    private static _textureLoaders: Nullable<IInternalTextureLoader[]>;
+
     /** @hidden */
-    public static _TextureLoaders: IInternalTextureLoader[] = [];
+    public static get _TextureLoaders(): IInternalTextureLoader[] {
+        if (!ThinEngine._textureLoaders) {
+            ThinEngine._textureLoaders = [];
+            ThinEngine._TextureLoaderFactories.forEach((factory) => {
+                ThinEngine._textureLoaders!.push(factory.create());
+            });
+        }
+        return ThinEngine._textureLoaders!;
+    }
+
+    /** @hidden */
+    public static _TextureLoaderFactories: { create: () => IInternalTextureLoader, dispose: (loader: IInternalTextureLoader) => void }[] = [];
+
+    public static MinimizeMemoryFootprint(): void {
+        if (ThinEngine._textureLoaders) {
+            for (let idx = 0; idx < ThinEngine._TextureLoaderFactories.length; ++idx) {
+                ThinEngine._TextureLoaderFactories[idx].dispose(ThinEngine._textureLoaders![idx]);
+            }
+            ThinEngine._textureLoaders = null;
+        }
+    }
 
     /**
      * Returns the current npm package of the sdk
--- a/src/Materials/Textures/Loaders/basisTextureLoader.ts
+++ b/src/Materials/Textures/Loaders/basisTextureLoader.ts
@@ -97,4 +97,7 @@ export class _BasisTextureLoader implements IInternalTextureLoader {
 }
 
 // Register the loader.
-Engine._TextureLoaders.push(new _BasisTextureLoader());
+Engine._TextureLoaderFactories.push({
+    create: () => { return new _BasisTextureLoader() },
+    dispose: (loader: IInternalTextureLoader) => {}
+});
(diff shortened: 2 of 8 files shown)


sebavan:
I do not think the lazy loading part is necessary (how much do we save ?), when building with ES6 you can consume only what you need already.

I guess the uninitialized part is the important one for the current workaround ?

syntheticmagus:
@sebavan I think the utility of the lazy loading part of it depends both on what you're using and when, because the loaders themselves vary in what they contain. The KTX2 loader in particular creates up to four (!) Web workers and holds onto them _statically_. Under the current architecture it does this as soon as Babylon.js loads, and there's no way to get rid of them. Thus, if you have an app that loads a single KTX2 texture in the first second of its execution, then goes on to render and operate for another ten minutes without loading any more KTX2s, the cost of the loader in memory and Web workers will impact the _entire runtime_, not just the load time, and there's nothing the consuming app can do about it because the KTX2 loader isn't exposed. I don't know if other loaders also hold onto heavy long-lasting resources, but if they do (or can), then the same problem applies.

So it really isn't the lazy loading that matters so much as the ability to enforce some kind of lifespan on these resources that otherwise live forever simply because they don't know when it's safe to die. That may have up to two benefits: (1) long-living heavy resources can be disposed so that they don't impact parts of the execution that don't need them, which becomes more useful if/as we add more resources like this to this capability; and (2) the fact that they _won't_ live forever eases some constraints on certain resources, allowing us to do heavier things if we need to because we now have a mechanism for developers to tell the `Engine`, "You can get rid of your heavy tools now."

sebavan:
sounds good to me, but let s ask @RaananW what pattern you should use on this one to prevent rebuilding it when he will move to es for those functions if it is early enough to built it only once ?

RaananW:
> sounds good to me, but let s ask @RaananW what pattern you should use on this one to prevent rebuilding it when he will move to es for those functions if it is early enough to built it only once ?

The general gist of future changes would be to move all of those static methods to be external functions in the Engine/ThinEngine module that receive an `engine` as the first variable (if needed). No more `private` or `public` static, but functions that can be called outside of the engine context. This way they will be included in the build if used, but will be tree-shaken out if not.

For example, this line:

```javascript
private static _textureLoaders: Nullable<IInternalTextureLoader[]>;
``` 

will be

```javascript
const _textureLoaders: Nullable<IInternalTextureLoader[]> = []
```
outside of the ThinEngine class, and the functions would look like this:

```javascript
export function MinimizeMemoryFootprint(): void {
        if (_textureLoaders) {
            for (let idx = 0; idx < _TextureLoaderFactories.length; ++idx) {
                _TextureLoaderFactories[idx].dispose(_textureLoaders![idx]);
            }
            _textureLoaders = null;
        }
    }
```

static functions don't need the actual engine, so it is simpler to pull them out of the class. I know that in this case we are talking about ca. 50 lines of code that is added, but all of those static functions accumulate.

In order to use them, a developer using esm would be able to:

```javascript
import { MinimizeMemoryFootprint } from "@outModulesPackage";

// or

import * as ThinEngineModule from "@outModulesPackage/Engines/ThinEngine";

// and then

ThinEngineModule.MinimizeMemoryFootprint(); // or the function, if imported directly
```

For back-compat, static functions that were added prior to the change will be added to the class (so that you could still use ThinEngine.MyWonderfulStaticMethod), but only in the LTS version. Side-note - this sentence is still TBD. I assume not all static methods will be thrown out, but I didn't dive into the engines yet.

deltakosh:
What is the current status here folks?

syntheticmagus:
@deltakosh I think the impetus to check this in has died down. I'm checking whether this will be valuable in the short term, in which case I'll push it forward, but most likely I'll just close it tomorrow as we can always reopen it later when it's more immediately relevant again.

syntheticmagus:
Closing this for now. I think the idea is still sound, but the need for it isn't currently clear and immediate. We can reopen it again later when/if it becomes tangibly relevant again.

PR #11446, opened 2021-11-12
Color4.FromHexString default alpha to 1

BlakeOne:
Patch to default alpha to 1 when a 7 char hex string is passed to Color4.FromHexString, to allow standard hex strings to be copied and pasted from color table websites, etc., more conveniently without having to add "ff" to the string each time.

When an empty or invalid string is passed, all 4 components still default to 0, in keeping with the original behavior.

Forum posts with more info on the use case for this are here: https://forum.babylonjs.com/t/default-alpha-to-1-for-color4-fromhexstring/25226/3

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -341,6 +341,7 @@
 
 ## Breaking changes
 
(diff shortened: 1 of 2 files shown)


BlakeOne:
Fixed the whitespace errors.

deltakosh (approved):

BlakeOne:
Reopened:)

PR #11450, opened 2021-11-12
Fix handedness issues with glTF serializer

bghgary:
There are 4 cases that this code now handles given a root node of the scene:
1. If the scene is right-handed and the root node matrix is identity, then remove the node.
2. If the scene is left-handed and converting the root node matrix to right-handed will result in identity, then remove the node.
3. If the scene is left-handed and the matrix determinant is negative, then change the matrix to right-handed.
4. Otherwise, convert the data to right-handed.


deltakosh (approved):

sebavan:
@bghgary I let you check for the build before I ll merge it in.

bghgary:
I'm abandoning this since I found some issues with my approach. This is a bit tricky to get right and the existing code already works for majority of cases.

PR #11452, opened 2021-11-12
Add a simple coroutine implementation and use it to make an async version of MergeMeshes.

ryantrem:
The primary motivation for this change is to introduce an asynchronous version of `MergeMeshes`. However, we can't break backward compatibility (so we still need a synchronous version of `MergeMeshes`), and `MergeMeshes` is too complicated to maintain two different versions of it. To solve this problem, I'm introducing some coroutine components. With these components, I can:
1. Turn an existing synchronous function into a coroutine by changing it from a `function` to a `function*`, changing the return type from `T` to `Coroutine<T>`, and adding `yield;` statements in parts of the function where we *potentially* want to yield control of the execution context so other stuff (like rendering) has a chance to execute.
2. Wrap the coroutine in a synchronous function (via helpers). This just steps the coroutine synchronously until it completes.
3. Wrap the coroutine in a new asynchronous function (via helpers). This steps the coroutine and occasionally actually yields control of the execution context by scheduling the next step of the coroutine via `setTimeout(step, 0)`.

For the asynchronous scheduling, I only implemented a very simple mechanism that yields control via `setTimeout` after 25ms of work. I kept this as an implementation detail for now, though in the future we could export more of the API to allow for customer schedulers. Also in the future, we could implement a much smarter scheduler, that uses heuristics like observing frame time (lets say 6ms) to try to determine how much time is left to do other work (e.g. 10ms to maintain 60fps), and then only use that upper limit for running coroutines. I don't want to try to implement such a system right now, but I think this is possible.

I tested these changes in the Playground and in Babylon React Native. In my test scenario, where I am using `MergeMeshes` on a relatively complex model, total execution time was the same (within some variance threshold from run to run) for the synchronous function, as well as the asynchronous version. As expected, the JS thread is much more responsive (but not 60fps) with the async version.

If people are happy with this addition, I'm willing to add unit tests as well.


src/Meshes/mesh.ts
```diff
@@ -4372,13 +4376,23 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
+        }
+
+        const vertexData: VertexData = yield* sourceVertexData._mergeCoroutine(meshVertexDatas, allow32BitsIndices);
 
         if (!meshSubclass) {
             meshSubclass = new Mesh(source.name + "_merged", source.getScene());
         }
 
-        (<VertexData>vertexData).applyToMesh(meshSubclass);
+        yield* vertexData._applyToCoroutine(meshSubclass);
```
deltakosh, line 4395: I'm afraid that could not work with IE and unfortunately this is pretty fundamental in the engine

Did you try it in IE?

deltakosh:
I'm really concerned about backward compat here. For current users this function will literally takes more frames and they may not want it at all. We must find another way to support your need without breaking backward compat and engine support on IE

Popov72:
I don't know if it's applicable, but @syntheticmagus has added coroutines lately (https://doc.babylonjs.com/divingDeeper/events/coroutines).

src/Meshes/mesh.ts
ryantrem: When you say it is fundamental to the engine, what do you mean? As far as I can tell, this is the only place `applyToMesh` is used, and this is in the context of `MergeMeshes`, which is only used in a few places like `hemisphereBuilder` and `sceneOptimizer`. These seem like not more minor aspects of BJS.

I did not try it in IE. Is it generator functions in general that don't work in IE, or just the `yield*` part to yield another generator?

deltakosh: yield and generator will not compile on IE
By "fundamental to the engine", I mean it is a core function that is used by several 1st party partners who (unfortunately) still needs to support IE

ryantrem:
> I'm really concerned about backward compat here. For current users this function will literally takes more frames and they may not want it at all. We must find another way to support your need without breaking backward compat and engine support on IE

What do you mean by "this function will literally take more frames?" The synchronous version is still synchronous and therefore runs in one frame, synchronously. The behavior should be identical to before the change, I think the only concern about backward compatibility is if `function*` breaks IE support.

ryantrem:
> I don't know if it's applicable, but @syntheticmagus has added coroutines lately (https://doc.babylonjs.com/divingDeeper/events/coroutines).

Hmm it does look like there is overlap. I will sync with @syntheticmagus.

src/Meshes/mesh.ts
ryantrem: I don't see any `yield` statements in the transpiled JS code, I guess because of the ES version being targeted by BJS, so presumably this won't break IE. I dev on a MacBook, but I can try to get a BJS dev environment setup on a Windows machine to test.

src/Misc/coroutine.ts
```diff
@@ -0,0 +1,133 @@
...
+
+// Coroutines are run with the runCoroutine function, which takes a Coroutine<T>, a CoroutineScheduler<T>, and a success and error callback.
+// A scheduler is responsible for scheduling the next step of a coroutine, either synchronously or asynchronously.
+
+/**
+ * A Coroutine<T> is the intersection of:
+ * 1. An Iterator that yields undefined, returns a T, and is not passed values with calls to next.
+ * 2. An IterableIterator of undefined (since it only yields undefined).
+ */
+export type Coroutine<T> = Iterator<undefined, T, unknown> & IterableIterator<undefined>;
```
syntheticmagus, line 26: Is this meaningfully different from the coroutines we already have?

https://github.com/BabylonJS/Babylon.js/blob/master/src/Misc/observableCoroutine.ts
https://doc.babylonjs.com/divingDeeper/events/coroutines

syntheticmagus: Never mind, just saw this had already been noted. Will sync separately.

src/Meshes/mesh.ts
sebavan: image

Some of the TS allowed constructs despite not emitting code relying on the feature might require other unsupported feature which is the case for instance with the async await syntax and the promise polyfill. I am not sure the required symbols are supported on IE.

sebavan:
converting to draft for now to prevent unintentional merges :-)

src/Misc/coroutine.ts
```diff
@@ -0,0 +1,133 @@
...
+        if (start === undefined) {
+            // If start is undefined, this is the first step of a coroutine.
+            start = performance.now();
+        } else {
+            // Otherwise check if the time interval has elapsed.
+            const end = performance.now();
+            if (end - start > yieldAfterMS) {
+                // If so, record a new start time, and schedule the coroutine step to happen later, effectively yielding control of the execution context.
+                start = end;
+                setTimeout(() => {
```
sebavan, line 57: queueMicrotask might be better here if available :-)

src/Misc/coroutine.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @param coroutineFactory A function that returns a Coroutine<T>.
+ * @returns A function that runs the coroutine asynchronously.
+ */
+export function makeAsyncFunction<TReturn, TCoroutineFactory extends (...params: any[]) => Coroutine<TReturn>>(coroutineFactory: TCoroutineFactory): (...params: Parameters<TCoroutineFactory>) => Promise<ExtractCoroutineReturnType<ReturnType<TCoroutineFactory>>> {
+    return (...params: Parameters<TCoroutineFactory>): Promise<ExtractCoroutineReturnType<ReturnType<TCoroutineFactory>>> => {
+        // Create the coroutine from the factory function.
+        const coroutine = coroutineFactory(...params);
+
+        // Run the coroutine with a yielding scheduler, resolving or rejecting the result promise when the coroutine finishes.
+        const result = new Deferred<TReturn>();
```
sebavan, line 126: Why not using a regular promise here ?

src/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "types": [],
         "baseUrl": "./",
         "rootDir": "./",
-        "outDir": "../.temp/manualBuildCore"
+        "outDir": "../.temp/manualBuildCore",
+        "downlevelIteration": true
```
sebavan, line 8: This might slow down iteration in general in babylon which I am afraid of.

src/Meshes/mesh.ts
ryantrem: I removed both the `yield*` and the `downlevelIteration` as I realized I could just manually iterate over the nested coroutine which is more verbose but functions identically.

src/Misc/coroutine.ts
ryantrem: It doesn't work, I think because we need to let control return to the browser to execute `requestAnimationFrame` so the next frame will actually render. If I yield with `queueMicrotask`, no frames are rendered while the coroutine is running.

src/Misc/coroutine.ts
ryantrem: I think I needed it in an older iteration of the code, but clearly that's no longer the case. Switched to Promise.

src/tsconfig.json
ryantrem: Removed.

ryantrem:
> > I don't know if it's applicable, but @syntheticmagus has added coroutines lately (https://doc.babylonjs.com/divingDeeper/events/coroutines).
> 
> Hmm it does look like there is overlap. I will sync with @syntheticmagus.

Looking closer at the `observableCoroutine`, it is basically coroutines coupled with `observables` as the sole scheduling mechanism. The coroutines I'm adding in this PR are more generalized, and so I was able to rework `observableCoroutine` to just build on top of these more general coroutine mechanisms. As far as I can tell, with these changes `observableCoroutine`:
- Has the same exact public contract.
- Has the same performance characteristics.
- Has the same memory allocation characteristics.
- Has the same lifetime semantics.

src/Misc/coroutine.ts
sebavan: Ohhh if it is to run on next frame, should we use the next render loop instead of a 0 timeout ?

sebavan:
This looks much better for old browser compat :-) I am happy it was only yield* having the issue not the full generators. We should have a quick chat today to see which approach is preferred in the team ?

src/Misc/coroutine.ts
bghgary: We can't rely on the render loop. There might not be a render loop running. If there isn't, this still needs to work.

src/Misc/coroutine.ts
```diff
@@ -0,0 +1,175 @@
...
+    never;
+
+/**
+ * Given a function that returns a Coroutine<T>, produce a function with the same parameters that returns a T.
+ * The returned function runs the coroutine synchronously.
+ * @param coroutineFactory A function that returns a Coroutine<T>.
+ * @returns A function that runs the coroutine synchronously.
+ * @hidden
+ */
+export function makeSyncFunction<TReturn, TCoroutineFactory extends (...params: any[]) => Coroutine<TReturn>>(coroutineFactory: TCoroutineFactory, abortSignal?: AbortSignal): (...params: Parameters<TCoroutineFactory>) => ExtractCoroutineReturnType<ReturnType<TCoroutineFactory>> {
```
syntheticmagus, line 156: It is a little weird to me that these are just sort of out in the global namespace. When imported under an alias (as is done in the Playground, for example), how will these show up? Will they just be `BABYLON.makeSyncFunction(...)`?

syntheticmagus: Namespacing or making these static functions on a type or something might also help clarify the connection between these and coroutines, I think.

syntheticmagus (commented):
Regarding the coroutines themselves, looks pretty good to me. My main concern with observable coroutines is that I don't want consumers of those (which are targeted toward business logic rather than infrastructure) to need a new concept of schedulers, and it looks like that simplicity is preserved here. This also lays the groundwork for adding capabilities such as value returning to that scenario as well, which is something we want but hadn't gotten to yet, so it's great from that perspective too.

src/Misc/coroutine.ts
bghgary: This is a good point. What happens to these functions in the documentation?

sebavan: @RaananW it sounds like a good topic for you ?

ryantrem: You are exactly right that these will show up as `BABYLON.makeSyncFunction(...)` for the UMD package (see the unit tests). That said, my understanding is that this is the direction BJS is actively moving to improve tree shaking (since otherwise if you just need one function in a class containing a bunch of static functions, all those functions get brought in). For ES6 modules, these functions are naturally "name spaced" by virtue of being in a module. I'm not sure if there is a way to group these functions beyond modules without hurting tree shaking. @RaananW can you confirm? Also, is there a naming scheme planned to make this more usable in UMD? For example, if we had something like `Mesh.Merge` and it was moved out to just be an exported function, would it just be named `Merge`, or is there some naming scheme we want to make it sort of name spaced within the context of UMD?

RaananW: Yes, the UMD package will contain those as a "global" member, because of the way we build it (anything exported is automatically a member of the BABYLON namespace).

What we can consider for the UMD LTS version is to remove the global declaration of functions assigned to static members, so that they at least don't appear in the typedoc of the BABYLON namespace. js-wise they will still be there though. I won't change the way we build. I don't see that as something that is needed for a few reasons. The main one is that UMD should slowly be a "legacy" package that is only used in very few cases, and the second is that almost any other framework (with UMD/Namespacing support" does the same - all functions are part of the global namespace.

Anyhow - we should not use static functions (especially not private static, which makes no sense when working with modules), but this is a topic I will first address in January (during the build system overhaul). They should be out of the class context to support better tree shaking.

Regarding naming - this is still an open question that I need to address soon. As the modules are still constructed based on directory, this will still be possible:

```javascript
import { Merge } from "@packageName/Meshes/mesh";
```
Instead of loading the Mesh class, you are loading from the Mesh module.

However, this would also be possible (while still supporting tree shaking, because there are no more side effects):

```javascript
import { Merge } from "@packageName";
```

This depends on whether or not we also provide an index file, but that's an internal decision I still need to make after experimenting with building the new esm package. This will eventually influence naming, as we can't have two functions named Merge (or - a better example is the `Parse` function that a lot of different components have)

TL;dr - I will run a pass at the static structure once I have all of these answers. For now - if it is assigned to a member of a class don't export it from the module and keep the class reference. This will take it out of the exported global namespace. If it is a new function export it and use it directly from the global namespace.

sebavan (approved):
LGTM !!! This is stunning work @ryantrem

syntheticmagus (approved):

PR #11463, opened 2021-11-15
allow disabling final color clamping

RaananW:
At the moment there is no way of disabling the final color clamp when using the PBR material (clamp to 30).
This sets clamping as the default value, but allows the dev to disable this using the image processing configuration.

### DIFF
--- a/src/Materials/Background/backgroundMaterial.ts
+++ b/src/Materials/Background/backgroundMaterial.ts
@@ -123,6 +123,7 @@ class BackgroundMaterialDefines extends MaterialDefines implements IImageProcess
     public SAMPLER3DGREENDEPTH = false;
     public SAMPLER3DBGRMAP = false;
     public IMAGEPROCESSINGPOSTPROCESS = false;
+    public CLAMPFINALCOLOR = true;
     public EXPOSURE = false;
     public MULTIVIEW = false;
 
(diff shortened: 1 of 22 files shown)


src/Materials/Background/backgroundMaterial.ts
```diff
@@ -123,6 +123,7 @@ class BackgroundMaterialDefines extends MaterialDefines implements IImageProcess
     public SAMPLER3DGREENDEPTH = false;
     public SAMPLER3DBGRMAP = false;
     public IMAGEPROCESSINGPOSTPROCESS = false;
+    public CLAMPFINALCOLOR = true;
```
sebavan, line 126: As true is not the default value in rebuild, I think you either need to rename to a negative name like UNCLAMPFINALCOLOR = false, or override the reset to force the value back to true in all the defines objects IIRC.

sebavan (commented):
Also this should be checked with the NodeMaterial as I can not remember how is it applied there ?

src/Materials/Background/backgroundMaterial.ts
RaananW: UNCLAMPFINALCOLOR sounded wrong...
But I agree that default false would be better.

Let's see what I can do

RaananW: CLAMPFINALCOLOR => SKIPFINALCOLORCLAMP

RaananW:
> Also this should be checked with the NodeMaterial as I can not remember how is it applied there ?

It is applied to the node material (has to be because it extends the image processing interface). Not sure how to use it there though.

sebavan (approved):

Popov72 (approved):

Popov72:
The `ImageProcessingBlock` does not clamp the value so the changes won't have any impact on it.

PR #11470, opened 2021-11-15
DeviceInputSystem: Modified Pointer Capture calls to improve IE 11 compatibility

PolygonalSun:
IE11 doesn't support `hasPointerCapture` so each call has been modified to include optional chaining to prevent erroneous calls when it doesn't exist.

### DIFF
--- a/src/DeviceInput/Implementations/webDeviceInputSystem.ts
+++ b/src/DeviceInput/Implementations/webDeviceInputSystem.ts
@@ -521,10 +521,10 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 deviceEvent.previousState = previousButton;
                 deviceEvent.currentState = pointer[evt.button + 2];
 
-                if (deviceType === DeviceType.Mouse && this._mouseId >= 0 && this._elementToAttachTo.hasPointerCapture(this._mouseId)) {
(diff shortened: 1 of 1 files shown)


deltakosh (approved):

sebavan (approved):

src/DeviceInput/Implementations/webDeviceInputSystem.ts
```diff
@@ -521,10 +521,10 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 deviceEvent.previousState = previousButton;
                 deviceEvent.currentState = pointer[evt.button + 2];
 
-                if (deviceType === DeviceType.Mouse && this._mouseId >= 0 && this._elementToAttachTo.hasPointerCapture(this._mouseId)) {
+                if (deviceType === DeviceType.Mouse && this._mouseId >= 0 && this._elementToAttachTo.hasPointerCapture?.(this._mouseId)) {
```
bghgary, line 524: Is it okay to not call releasePointerCapture when setPointerCapture is called?

PolygonalSun: It should automatically be released on a `pointerout` event but just in case there are issues with calling `setPointerCapture`, I've added a check to make sure that we only set it if `hadPointerCapture` is present.

bghgary: I'm not sure I understand. If it will autorelease on `pointerout`, why do we need to call this at all here?

PR #11474, opened 2021-11-16
Add a work quantization helper and use it for an async version of MergeMeshes

ryantrem:
This is a different approach to work quantization from what I tried in this PR: https://github.com/BabylonJS/Babylon.js/pull/11452

I've tested this in IE and it works as expected.
I also did not see any performance degradation in the existing synchronous `MergeMeshes` function.

This approach is not as good as the coroutine approach, but it's relatively simple and it works with IE.

### DIFF
--- a/src/Meshes/mesh.ts
+++ b/src/Meshes/mesh.ts
@@ -33,6 +33,7 @@ import { Path3D } from "../Maths/math.path";
 import { Plane } from "../Maths/math.plane";
 import { TransformNode } from "./transformNode";
 import { DrawWrapper } from "../Materials/drawWrapper";
+import { createWorkQuantizer } from '../Misc/workQuantizer';
 
 declare type InstancedMesh = import("./instancedMesh").InstancedMesh;
 declare type IPhysicsEnabledObject = import("../Physics/physicsImpostor").IPhysicsEnabledObject;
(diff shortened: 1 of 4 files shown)


src/Meshes/mesh.ts
```diff
@@ -4291,11 +4292,52 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
+        meshes: Array<Mesh>,
+        disposeSource = true,
+        allow32BitsIndices: boolean | undefined,
+        meshSubclass: Mesh | undefined,
+        subdivideWithSubMeshes: boolean | undefined,
+        multiMultiMaterials: boolean | undefined,
+        asynchronous: boolean,
+        onFinished: (mesh: Nullable<Mesh>) => void,
+    ): Promise<void> {
+        const quantizer = asynchronous ? createWorkQuantizer() : createWorkQuantizer(false);
```
deltakosh, line 4333:
```suggestion
        const quantizer = createWorkQuantizer(!asynchronous);

```

src/Meshes/mesh.vertexData.ts
```diff
@@ -274,70 +299,89 @@ export class VertexData {
         return this;
     }
 
-    private _applyTo(meshOrGeometry: IGetSetVerticesData, updatable: boolean = false): VertexData {
+    private async _applyTo(
+        meshOrGeometry: IGetSetVerticesData,
+        updatable = false,
+        asynchronous: boolean,
+    ): Promise<void> {
+        const quantizer = asynchronous ? createWorkQuantizer() : createWorkQuantizer(false);
```
deltakosh, line 307:
```suggestion
        const quantizer = createWorkQuantizer(!asynchronous);

```

deltakosh (approved):
I like it a lot

Only small nits

sebavan (commented):
I like the overall idea better :-)

In order to prevent the unawaited promise and their creation in the syncronous case, I wonder if mergeMeshes could instead return a list of task that we either run fully sequentially or that we deferred through microtask if the time has ellapsed ?

Not sure it is doable but smthg like this (Pseudo Code):

```
function getTasks() {
      const tasks = [];
       if (this.positions) {
            tasks.push(() => meshOrGeometry.setVerticesData(VertexBuffer.PositionKind, this.positions, updatable) }
        }
...
}

merge() {
  const mergeMeshTasks = _getTasks();
  for (const task of mergeMeshTasks) {
    task();
  }
}

mergeAsync() {
  const mergeMeshTasks = _getTasks();
  for (const task of mergeMeshTasks) {
    task();
    quantizer.shouldYield && await quantizer.yield();
  }
}
```

Let see what @RaananW thinks about it as well ?

src/Meshes/mesh.ts
ryantrem: As written, `true` is not a valid argument for `createWorkQuantizer`. Currently you have to either pass false (meaning don't quantize), or a `number` (meaning quantize into time chunks of the specified size), or `undefined` (meaning do the default thing, which is quantizing to the default time chunk). I did it this way to try to make it more obvious what the resulting behavior would be. I feel like making `true` a valid argument is less clear, but if others feel strongly otherwise I will change it.

ryantrem:
> I like the overall idea better :-)
> 
> In order to prevent the unawaited promise and their creation in the syncronous case, I wonder if mergeMeshes could instead return a list of task that we either run fully sequentially or that we deferred through microtask if the time has ellapsed ?
> 
> Not sure it is doable but smthg like this (Pseudo Code):
> 
> ```
> function getTasks() {
>       const tasks = [];
>        if (this.positions) {
>             tasks.push(() => meshOrGeometry.setVerticesData(VertexBuffer.PositionKind, this.positions, updatable) }
>         }
> ...
> }
> 
> merge() {
>   const mergeMeshTasks = _getTasks();
>   for (const task of mergeMeshTasks) {
>     task();
>   }
> }
> 
> mergeAsync() {
>   const mergeMeshTasks = _getTasks();
>   for (const task of mergeMeshTasks) {
>     task();
>     quantizer.shouldYield && await quantizer.yield();
>   }
> }
> ```
> 
> Let see what @RaananW thinks about it as well ?

This seems like a manual version of what generator functions already do. Given that, I kind of lean more towards generator functions / coroutines being the general solution, and only using the approach in this PR when we have to to maintain back compat with IE.

ryantrem:
As you can see in the latest iteration, getting the error handling correct with this approach is quite easy to mess up.

RaananW:
> This seems like a manual version of what generator functions already do. Given that, I kind of lean more towards generator functions / coroutines being the general solution, and only using the approach in this PR when we have to to maintain back compat with IE.

We are slowly removing IE support. a few new features are already not compatible with IE. if it's just for IE (and is a new feature), there is IMO no need to add it to the codebase. @deltakosh , any objection for that?

I am only referring to the quantizer - this seems to be a task scheduler that schedules the next task to the next tick (defaults to 25 ms). We have a few different implementations already that could help that - `Tools.DelayAsync` returns a delayed promise (setTimeout-based). Tools.AsyncLoop is an async implementation of a for loop (can also iterate also over an array of functions to be executed) (not mentioning generators and coroutines, as you already added that in the description)

I totally understand the need for this, I am just thinking if we should somehow unify all of those different scheduler implementations to a single accepted solution? What I like very much about your solution (that is missing in all others) is the fact that it is time-based. If no more than 25ms have passed, it will still run the next task.

sebavan:
> We are slowly removing IE support. a few new features are already not compatible with IE. if it's just for IE (and is a new feature), there is IMO no need to add it to the codebase. @deltakosh , any objection for that?

It is in use in mergeMesh which is unfortunately not a new feature and it was enforcing downleveliteration at the potential detriment of perfs in the overall app.

The other draft has addressed all those https://github.com/BabylonJS/Babylon.js/pull/11452 and I would mainly lean towards the other one cause unhandled promise will create us potential weird gotchas in the future. It is way to easy to introduce async code in now without knowing we broke something or that it won t be awaited. It is kind of what was happening with errors here.

ryantrem:
I think the error propagation was too error prone and hard to understand, so I switched things to still use the `WorkQuantizer`, but to use generator functions that yield instead. This is somewhat of a hybrid between the original solution in this PR, and the more complete coroutine solution in #11452. It works with IE as well, and does not have the error propagation gotchas we have when trying to use await in a function meant be able to run synchronously.

ryantrem:
Going back to the original idea: https://github.com/BabylonJS/Babylon.js/pull/11452

PR #11478, opened 2021-11-17
Code called twice

brunobg:
`PBRAnisotropicConfiguration.AddFallbacks` is being called twice here. I assume it's a bug.

### DIFF
--- a/src/Materials/PBR/pbrBaseMaterial.ts
+++ b/src/Materials/PBR/pbrBaseMaterial.ts
@@ -1303,6 +1303,7 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             fallbacks.addFallback(fallbackRank++, "PARALLAXOCCLUSION");
         }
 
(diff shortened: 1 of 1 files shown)


src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1303,7 +1303,6 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             fallbacks.addFallback(fallbackRank++, "PARALLAXOCCLUSION");
         }
 
-        fallbackRank = PBRAnisotropicConfiguration.AddFallbacks(defines, fallbacks, fallbackRank);
```
Popov72, line 1306: Actually it's a copy/paste problem: one of them (say the first one) should be `PBRClearCoatConfiguration.AddFallbacks(defines, fallbacks, fallbackRank);`

Popov72 (changes requested):

sebavan (approved):

PR #11480, opened 2021-11-17
Update key value ace

carolhmj:
When the user hits the "new key" button on ACE but a key already exists on that frame, update the key's value instead of creating a new one.


inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/graphComponent.tsx
```diff
@@ -250,6 +250,143 @@ export class GraphComponent extends React.Component<IGraphComponentProps, IGraph
...
             this.forceUpdate();
         });
+
+        // Update keypoint
+        this.props.context.onCreateOrUpdateKeyPointRequired.add(() => {
+            if (this.props.context.activeAnimations.length === 0) {
+                return;
+            }
+
+            for (const currentAnimation of this.props.context.activeAnimations) {
```
deltakosh, line 260: This seems to be a copy of the code earlier. Maybe we could factorize it?

sebavan (approved):

inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/graphComponent.tsx
carolhmj: Now that I'm thinking again about it, maybe we don't even need the new `onCreateOrUpdateKeyPointRequired`, because in all the places where we are creating a new key point (the new key button and the `onClipLengthIncreased` and `onClipLengthDecreased` listeners), we are already checking if the key exists. So perhaps this should be the default behavior of the `onNewKeyPointRequired` listener, what do you think?

deltakosh: If it works that would be even better yeah!

carolhmj: Done!

sebavan (approved):

PR #11482, opened 2021-11-17
Fix _getPositionData to apply morph before bones computation

Popov72:


deltakosh (approved):
Looks all good to me. If I may I would love to add a test to display the correct bounding box and to make sure we do not regress

PR #11495, opened 2021-11-19
[XR] multi-pointer screen coordinates projection

RaananW:
This PR will allow passing the projected screen coordinates of controllers in XR.

The data will be available in `event.screenX` and `event.screenY` (if available at all).

### DIFF
--- a/src/XR/features/WebXRControllerPointerSelection.ts
+++ b/src/XR/features/WebXRControllerPointerSelection.ts
@@ -828,7 +828,7 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
     private _augmentPointerInit(pointerEventInit: PointerEventInit, id: number, screenCoordinates?: { x: number; y: number; }): void {
         pointerEventInit.pointerId = id;
         pointerEventInit.pointerType = "xr";
-        if(screenCoordinates) {
+        if (screenCoordinates) {
             pointerEventInit.screenX = screenCoordinates.x;
             pointerEventInit.screenY = screenCoordinates.y;
         }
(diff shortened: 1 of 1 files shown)


src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -392,9 +393,16 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
                 if (camera) {
                     camera.viewport.toGlobalToRef(scene.getEngine().getRenderWidth(), scene.getEngine().getRenderHeight(), this._viewportRef);
                     Vector3.ProjectToRef(controllerGlobalPosition, this._identityMatrix, scene.getTransformMatrix(), this._viewportRef, this._screenCoordinatesRef);
-
-                    scene.pointerX = this._screenCoordinatesRef.x;
-                    scene.pointerY = this._screenCoordinatesRef.y;
+                    // stay safe
```
sebavan, line 412: What else than number could be in ?

src/XR/features/WebXRControllerPointerSelection.ts
```diff
@@ -487,6 +495,10 @@ export class WebXRControllerPointerSelection extends WebXRAbstractFeature {
             if (!controllerData.pick) {
                 return;
             }
+            if (controllerData.screenCoordinates) {
```
sebavan, line 514: could we factorize this block ? it looks similar in the 4 cases ?

sebavan (approved):
Only a couple Nits

src/XR/features/WebXRControllerPointerSelection.ts
RaananW: NaN :-)

src/XR/features/WebXRControllerPointerSelection.ts
RaananW: The pointerEvent object is reused on a lot of those, but is not reused between different types of ray modes.

RaananW: let me see if I can find a nice way to unify those.

PR #11496, opened 2021-11-19
Fix glTFLoader 2.0 when dealing with glTF files that contain no meshes

simonihmig:
AFAIK glTF may contain no meshes (and hence no nodes) at all (our use case are material-only assets). But in `_setupData` of the glTF 2.0 loader, `_rootBabylonMesh` is only initialized when nodes exist. If that is not the case, `_getMeshes()` returns `[undefined]`, which later throws e.g. in a `AssetContainer.dispose()` call.


loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -186,7 +186,7 @@ export class GLTFLoader implements IGLTFLoader {
      * The root Babylon mesh when loading the asset.
      */
     public get rootBabylonMesh(): Mesh {
-        return this._rootBabylonMesh;
+        return this._rootBabylonMesh!;
```
simonihmig, line 189: Note that `!` here is continuing the existing lie that this cannot be undefined, while the bug showed that it can be! I am not happy about this, but changing it would affect other pieces of code (e.g. extensions), and all TypeScript users that use this directly.

Happy to change that (to be nullable) if you agree!?

Note that the same "problem" applies to many other properties here and probably also in many other places, that are optimistically typed. A `strict` (or at least `strictPropertyInitialization: true`) TS config would catch all these cases (by complaining about `Property '...' has no initializer and is not definitely assigned in the constructor`), but this would probably affect most parts of the whole code base...

sebavan:
@bghgary will look into this one soon :-)

loaders/src/glTF/2.0/glTFLoader.ts
bghgary: Yeah, I wish we would turn this on, but it will take some effort.

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -506,7 +506,9 @@ export class GLTFLoader implements IGLTFLoader {
             for (let index of scene.nodes) {
                 const node = ArrayItem.Get(`${context}/nodes/${index}`, this._gltf.nodes, index);
                 promises.push(this.loadNodeAsync(`/nodes/${node.index}`, node, (babylonMesh) => {
-                    babylonMesh.parent = this._rootBabylonMesh;
+                    if (this._rootBabylonMesh) {
+                        babylonMesh.parent = this._rootBabylonMesh;
+                    }
```
bghgary, line 511: Is this necesary? If there is no rootBabylonMesh, then assigning the parent to undefined is fine right? Maybe make _rootBabylonMesh a Nullable<Mesh>?

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1111,7 +1115,9 @@ export class GLTFLoader implements IGLTFLoader {
         this._babylonScene._blockEntityCollection = false;
 
         // See https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#skins (second implementation note)
-        babylonSkeleton.overrideMesh = this._rootBabylonMesh;
+        if (this._rootBabylonMesh) {
+            babylonSkeleton.overrideMesh = this._rootBabylonMesh;
+        }
```
bghgary, line 1120: ditto for this

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -115,7 +115,7 @@ export class GLTFLoader implements IGLTFLoader {
     private _gltf: IGLTF;
     private _bin: Nullable<IDataBuffer>;
     private _babylonScene: Scene;
-    private _rootBabylonMesh: Mesh;
+    private _rootBabylonMesh?: Mesh;
```
bghgary, line 118: Based on the other comments below. I would recommend changing this to:
```suggestion
    private _rootBabylonMesh: Nullable<Mesh> = null;
```

bghgary (commented):
I have some minor comments, but it looks good otherwise.

loaders/src/glTF/2.0/glTFLoader.ts
bghgary: > Happy to change that (to be nullable) if you agree!?

I agree but it is unfortunately a breaking change. Maybe you can throw an exception instead?

loaders/src/glTF/2.0/glTFLoader.ts
simonihmig: I applied your suggestion!

loaders/src/glTF/2.0/glTFLoader.ts
simonihmig: Changed by making it null by default as suggested.

loaders/src/glTF/2.0/glTFLoader.ts
simonihmig: Changed by making it null by default as suggested.

loaders/src/glTF/2.0/glTFLoader.ts
simonihmig: Hm, does that make it less of a breaking change?

> it is unfortunately a breaking change

Yes, for TS-consumers at least, but wouldn't that be ok, as it will only be released as part of the next major, right?

sebavan: I would suggest to make it nullable and reference it in our what s new breaking change list. The impact is mostly internal as it does not seem to be used a lot outside; The playground does show no usage at all for instance: https://doc.babylonjs.com/playground?q=rootBabylonMesh&type=code

sebavan: @simonihmig, thinking about it more, lets make it nullable for sure :-)

sebavan:
@simonihmig , the last commit shows that the entire files have been change :-( 
image

I wonder if it is line ending ???

simonihmig:
> I wonder if it is line ending ???

Oh no, yet again! 🙈  Idk why my IDE freaks out like that only on this project. Will fix and force push in a minute!

loaders/src/glTF/2.0/glTFLoader.ts
simonihmig: @sebavan done! Had to update smaller parts of some glTF extensions. Although I don't know the code of those, the changes *should* be syntactically correct (i.e. don't change behavior), but please double check!

sebavan (approved):

sebavan:
Thanks a lot, I ll merge right after the build is done !!!

simonihmig:
> I wonder if it is line ending ???

So about this... this came when my IDE converted files that I touched to `CRLF` (i.e. Windows) line endings. When manually switching back to the "usual" `LF` (Linux/Mac) line endings, the diff was clean again. Which does suggest to me that line endings in this repo are supposed to be `LF`. *However*, I can now see why my IDE did this *rightly* , as this is specified as such in this repo's [`.editorconfig`](https://github.com/BabylonJS/Babylon.js/blob/master/.editorconfig#L7)! 🤔

Which *seems* to contradict what the committed files *actually* are. Or what do I miss?

Should I open a separate issue?

sebavan:
I do not think we should open a separate issue but let s keep the current file line ending as it is in PR and lets check with @RaananW if it should be part of its next big build system refacto ?

RaananW:
This will be unified as part of code refactoring along with lint integration.

PR #11498, opened 2021-11-19
Removing all properties that do not affecting a given control from the UI

msDestiny14:
image

For example, only containers will have font properties and color has been removed from image and color picker.

### DIFF
--- a/guiEditor/src/tools.ts
+++ b/guiEditor/src/tools.ts
@@ -78,4 +78,20 @@ export class Tools {
             grid.addControl(controls[i], tags[i].x, tags[i].y);
         }
     }
+
+    public static isContainer(guiControl: Control) {
+        switch (guiControl.typeName) {
+            case "Button":
+            case "StackPanel":
+            case "Rectangle":
+            case "Ellipse":
+            case "Grid":
+            case "ScrollViewer":
+            case "Container":
+            case "VirtualKeyboard":
+                return true;
+            default:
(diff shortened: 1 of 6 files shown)


guiEditor/src/tools.ts
```diff
@@ -78,4 +78,20 @@ export class Tools {
             grid.addControl(controls[i], tags[i].x, tags[i].y);
         }
     }
+
+    public static isContainer(guiControl: Control) {
```
sebavan, line 82: Why not adding a property to the controls to know if they are a container ? This would prevent to maintain this separately

sebavan: You could then simply do control.isContainer

sebavan (changes requested):

guiEditor/src/tools.ts
msDestiny14: Not a bad idea. I'm down to add it. :)

deltakosh: well even simpler: `return guiControl instanceof Container`

sebavan: Is it not used on a frequent code path ? cause instanceof is slow :-)

deltakosh: It is only for the editor

sebavan: so yes what he said :-)

PR #11501, opened 2021-11-19
NME: Add CustomBlock to node materials

Popov72:
This PR is adding the `CustomBlock` block to node materials. It allows to use straight glsl code as a node material block.

The custom blocks are located right below the custom frames:
image

The description of a custom block is a json file.

Ex:
* Perlin2D
```json
{
    "name": "Perlin2D",
    "comments": "Generates a Perlin noise single value given a vec2 and time",
    "target": "Neutral",
    "inParameters": [
        {
            "name": "p",
            "type": "Vector2"
        },
        {
            "name": "dim",
            "type": "Float"
        },
        {
            "name": "time",
            "type": "Float"
        }
    ],
    "outParameters": [
        {
            "name": "output",
            "type": "Float"
        }
    ],
    "functionName": "perlin",
    "code": [
        "float rand(vec2 co){return fract(sin(dot(co.xy ,vec2(12.9898,78.233))) * 43758.5453);}", 
        "float rand(vec2 co, float l) {return rand(vec2(rand(co), l));}", 
        "float rand(vec2 co, float l, float t) {return rand(vec2(rand(co, l), t));}", 
        "", 
        "void perlin(vec2 p, float dim, float time, out float result) {", 
        "   vec2 pos = floor(p * dim);", 
        "   vec2 posx = pos + vec2(1.0, 0.0);", 
        "   vec2 posy = pos + vec2(0.0, 1.0);", 
        "   vec2 posxy = pos + vec2(1.0);", 
        "   ", 
        "   float c = rand(pos, dim, time);", 
        "   float cx = rand(posx, dim, time);", 
        "   float cy = rand(posy, dim, time);", 
        "   float cxy = rand(posxy, dim, time);", 
        "   ", 
        "   vec2 d = fract(p * dim);", 
        "   d = -0.5 * cos(d * 3.14159265358979323846) + 0.5;", 
        "   ", 
        "   float ccx = mix(c, cx, d.x);", 
        "   float cycxy = mix(cy, cxy, d.x);", 
        "   float center = mix(ccx, cycxy, d.y);", 
        "   ", 
        "   result = center * 2.0 - 1.0;", 
        "}"
    ]    
}
```
* multiply
```json
{
    "name": "CustomMultiply",
    "comments": "Multiplies the left and right inputs of the same type together",
    "target": "Neutral",
    "inParameters": [
        {
            "name": "left",
            "type": "AutoDetect"
        },
        {
            "name": "right",
            "type": "AutoDetect"
        }
    ],
    "outParameters": [
        {
            "name": "output",
            "type": "BasedOnInput",
            "typeFromInput": "left"
        }
    ],
    "inLinkedConnectionTypes" : [
        {
            "input1": "left",
            "input2": "right",
            "looseCoupling": false
        }
    ],
    "functionName": "multiply_{TYPE_left}",
    "code": [
        "{TYPE_output} myHelper_{TYPE_left}({TYPE_left} l, {TYPE_right} r) { return l * r; }",
        "void multiply_{TYPE_left}({TYPE_left} l, {TYPE_right} r, out {TYPE_output} result) {",
        "   result = myHelper_{TYPE_left}(l, r);",
        "}"
    ]    
}
```
The `CustomMultiply` block does the same thing than the existing `Multiply` block but I have used it as an example to demonstrate using auto-detect / BasedOnInput types: when using these special types, you must use the `{TYPE_XXX}` syntax inside the glsl code because the types are known only at runtime when connecting inputs to the block in the NME.

We will probably need at some point a UI to create/modify custom blocks: for the time being you must handle the json file by hand (see how Unity is doing it: https://docs.unity3d.com/Packages/com.unity.shadergraph@6.7/manual/Custom-Function-Node.html).

Example of a node material using Perlin2D: https://nme.babylonjs.com/#3WEKUZ#1

### DIFF
--- a/nodeEditor/src/components/nodeList/nodeListComponent.tsx
+++ b/nodeEditor/src/components/nodeList/nodeListComponent.tsx
@@ -269,7 +269,7 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
         );
     }
 
-    removeBlock(value: string): void {
+    removeCustomBlock(value: string): void {
         let blockJson = localStorage.getItem("Custom-Block-List");
         if (blockJson) {
             let blockList = JSON.parse(blockJson);
@@ -477,7 +477,7 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
                                 tooltip={this._customBlockList[block] || ""}
                                 iconImage={deleteButton}
                                 iconTitle="Delete"
-                                onIconClick={(value) => this.removeBlock(value)}
+                                onIconClick={(value) => this.removeCustomBlock(value)}
                                 lenSuffixToRemove={11}
                             />
                         );
--- a/nodeEditor/src/components/nodeList/nodeListComponent.tsx
+++ b/nodeEditor/src/components/nodeList/nodeListComponent.tsx
@@ -163,6 +163,7 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
     };
 
     private _customFrameList: { [key: string]: string };
+    private _customBlockList: { [key: string]: string };
 
     constructor(props: INodeListComponentProps) {
         super(props);
@@ -172,6 +173,11 @@ export class NodeListComponent extends React.Component<INodeListComponentProps, 
         let frameJson = localStorage.getItem("Custom-Frame-List");
         if (frameJson) {
             this._customFrameList = JSON.parse(frameJson);
+        }
+
+        let blockJson = localStorage.getItem("Custom-Block-List");
+        if (blockJson) {
+            this._customBlockList = JSON.parse(blockJson);
         }
 
         this._onResetRequiredObserver = this.props.globalState.onResetRequiredObservable.add(() => {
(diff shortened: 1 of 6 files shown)


sebavan:
Why is the code an array ? vs a string ?

Popov72:
Because it's easier/more readable when doing the file by hand :)

In any case, you can always do `"code": ["all the code here"]` if you want to have the code in a single string. So I thought using an array was somewhat ok. We could also allow both a string or an array of string as two valid formats for "code".

sebavan:
> Because it's easier/more readable when doing the file by hand :)
> 
> In any case, you can always do `"code": ["all the code here"]` if you want to have the code in a single string. So I thought using an array was somewhat ok. We could also allow both a string or an array of string as two valid formats for "code".

it could also be a string a ppl use literals ? 

```js
cons shader = `void main() {
  gl...
};`
```

nodeEditor/src/components/nodeList/nodeListComponent.tsx
```diff
@@ -231,15 +237,65 @@ export class NodeListComponent extends React.Component<INodeListComponentProps,
...
+                localStorage.setItem("Custom-Block-List", JSON.stringify(blockList));
+                this._customBlockList = blockList;
+                this.forceUpdate();
+            },
+            undefined,
+            true
+        );
+    }
+
+    removeBlock(value: string): void {
```
deltakosh, line 272: maybe `removeCustomBlock` ?

deltakosh (approved):
All good for me (just a tiny nit). This will deserve its own doc for sure ;) with several examples like the ones you shared

Popov72:
> it could also be a string a ppl use literals ?

\`...\` syntax is not allowed in a json file.

> This will deserve its own doc for sure ;) with several examples like the ones you shared

Yes, documenting this is on my todo list.

sebavan:
> > it could also be a string a ppl use literals ?
> 
> `...` syntax is not allowed in a json file.

Mouahahahahah how did I miss this :-) This is all perfect then

sebavan (approved):

sebavan:
@Popov72 seems like something is not compiling ?

PR #11502, opened 2021-11-19
VAT fails with ShaderMaterial

brunobg:
VAT doesn't work with a ShaderMaterial, PG that reproduces the error: https://playground.babylonjs.com/#11GAIH#153

```
BJS - [17:55:14]: Error: VERTEX SHADER ERROR: 0:45: 'matricesIndices' : undeclared identifier
```

This is because `NUM_BONE_INFLUENCERS` is not defined. This patch defines it.

### DIFF
--- a/src/Materials/shaderMaterial.ts
+++ b/src/Materials/shaderMaterial.ts
@@ -667,7 +667,14 @@ export class ShaderMaterial extends PushMaterial {
                 }
             }
         } else {
-            defines.push("#define NUM_BONE_INFLUENCERS 0");
+            const bvaManager = (<Mesh>mesh).bakedVertexAnimationManager;
+
+            if (bvaManager && bvaManager.isEnabled) {
+                defines.push("#define NUM_BONE_INFLUENCERS " + mesh.numBoneInfluencers);
+            }
+            else {
+                defines.push("#define NUM_BONE_INFLUENCERS 0");
+            }
         }
 
         // Morph
--- a/src/Materials/shaderMaterial.ts
+++ b/src/Materials/shaderMaterial.ts
@@ -740,6 +740,19 @@ export class ShaderMaterial extends PushMaterial {
 
             if (bvaManager && bvaManager.isEnabled) {
                 defines.push("#define BAKED_VERTEX_ANIMATION_TEXTURE");
+                if (this._options.uniforms.indexOf("bakedVertexAnimationSettings") === -1) {
+                    this._options.uniforms.push("bakedVertexAnimationSettings");
+                }
+                if (this._options.uniforms.indexOf("bakedVertexAnimationTextureSizeInverted") === -1) {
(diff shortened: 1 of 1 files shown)


sebavan (approved):

Popov72:
In fact there is a bug in your PG, you must define `finalWorld` for the code injected by `#include<bonesVertex>` to work.

If you look at the console log at the first error log, you will see it is:
```
BJS - [22:50:35]: Error: VERTEX SHADER ERROR: 0:34: 'finalWorld' : undeclared identifier
```
The error with `matricesIndices` is not the true error: when a shader does not compile, Babylon is trying to remove some features   from the shader and recompile (it is the "fallback" mechanism). If it still does not work it still try to remove some other features and so on, a number of times. So, when you get a shader compilation error, you should always locate the very first error because the others can be different because of the code removed.

Also, you must use `finalWorld` matrix to transform the vertex. See https://playground.babylonjs.com/#11GAIH#155

Even after these corrections the PG still does not work because of a bug in `ShaderMaterial`: the uniforms/samplers used by BVA are not defined. The block at lines 731-739 should be replaced with:
```typescript
if (mesh) {
    const bvaManager = (<Mesh>mesh).bakedVertexAnimationManager;

    if (bvaManager && bvaManager.isEnabled) {
        defines.push("#define BAKED_VERTEX_ANIMATION_TEXTURE");
        if (this._options.uniforms.indexOf("bakedVertexAnimationSettings") === -1) {
            this._options.uniforms.push("bakedVertexAnimationSettings");
        }
        if (this._options.uniforms.indexOf("bakedVertexAnimationTextureSizeInverted") === -1) {
            this._options.uniforms.push("bakedVertexAnimationTextureSizeInverted");
        }
        if (this._options.uniforms.indexOf("bakedVertexAnimationTime") === -1) {
            this._options.uniforms.push("bakedVertexAnimationTime");
        }

        if (this._options.samplers.indexOf("bakedVertexAnimationTexture") === -1) {
            this._options.samplers.push("bakedVertexAnimationTexture");
        }
    }

    MaterialHelper.PrepareAttributesForBakedVertexAnimation(attribs, mesh, defines);
}
```

src/Materials/shaderMaterial.ts
```diff
@@ -667,7 +667,14 @@ export class ShaderMaterial extends PushMaterial {
                 }
             }
         } else {
-            defines.push("#define NUM_BONE_INFLUENCERS 0");
+            const bvaManager = (<Mesh>mesh).bakedVertexAnimationManager;
```
Popov72, line 670: See my answer in the comments, it is not the right fix.

Popov72 (changes requested):

sebavan:
@Popov72 could you have a final check ?

Popov72 (approved):

brunobg:
Thanks @Popov72, you are always awesome :)

PR #11503, opened 2021-11-19
Fix for GUI clipping when bug when parent container is scaled

msDestiny14:
Also tested with this playground as well as the GUI editor. 
https://www.babylonjs-playground.com/debug.html#F6XWSQ#2

### DIFF
--- a/gui/src/2D/controls/control.ts
+++ b/gui/src/2D/controls/control.ts
@@ -1617,24 +1617,32 @@ export class Control {
     }
 
     protected _evaluateClippingState(parentMeasure: Measure) {
+        let evaluatedMeasure = new Measure(0, 0, 0, 0);
+        this._currentMeasure.transformToRef(this._transformMatrix, evaluatedMeasure);
+
+        let evaluatedParentMeasure = new Measure(0, 0, 0, 0);
+        const parentMatrix = this.parent?._transformMatrix;
+        if (parentMatrix)
+            parentMeasure.transformToRef(parentMatrix, evaluatedParentMeasure);
+
         if (this.parent && this.parent.clipChildren) {
             // Early clip
-            if (this._currentMeasure.left > parentMeasure.left + parentMeasure.width) {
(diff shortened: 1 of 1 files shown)


gui/src/2D/controls/control.ts
```diff
@@ -1617,24 +1617,32 @@ export class Control {
     }
 
     protected _evaluateClippingState(parentMeasure: Measure) {
+        let evaluatedMeasure = new Measure(0, 0, 0, 0);
```
sebavan, line 1620: We should not allocate as much, you could keep the instance of measure in cache and reuse it. Also, Nit: it should be a const._evaluateClippingState

gui/src/2D/controls/control.ts
```diff
@@ -1617,24 +1617,32 @@ export class Control {
...
+        this._currentMeasure.transformToRef(this._transformMatrix, evaluatedMeasure);
+
+        let evaluatedParentMeasure = new Measure(0, 0, 0, 0);
```
sebavan, line 1623: We should not allocate as much, you could keep the instance of measure in cache and reuse it. Also, Nit: it should be a const._evaluateClippingState

gui/src/2D/controls/control.ts
```diff
@@ -1617,24 +1617,32 @@ export class Control {
...
+        const parentMatrix = this.parent?._transformMatrix;
```
sebavan, line 1624: You should move the 3 lines in the if block below as it already checks for parent nullity. preventing the need for the ? and parentMatrix test.

sebavan (changes requested):
I ll let @RaananW do the final pass on Monday as he was familiar with the issue.

RaananW:
I have no other comments - Seb nailed them all. 

Regarding updating the transformMatrix - if it is working without it then it's all fine. I didn't change the code, I only debugged using the console, so I might have missed something there.

gui/src/2D/controls/control.ts
msDestiny14: Not sure what you mean by this: const._evaluateClippingState?

PR #11512, opened 2021-11-22
Create structure for a persistent popup component that doesn't clos…

carolhmj:
…e when the inspector tab is changed.

Usage example:

* To open a popup: 
```javascript
globalState.onPopupContentRenderChangedObservable.notifyObservers(() => {
  return <div>Popup contents...</div>
});
```

* To close a popup:
```javascript
globalState.onPopupContentRenderChangedObservable.notifyObservers(null);
```

### DIFF
--- a/inspector/src/components/globalState.ts
+++ b/inspector/src/components/globalState.ts
@@ -12,6 +12,7 @@ import { CameraGizmo } from "babylonjs/Gizmos/cameraGizmo";
 import { PropertyChangedEvent } from "./propertyChangedEvent";
 import { ReplayRecorder } from "./replayRecorder";
 import { DataStorage } from "babylonjs/Misc/dataStorage";
+import { ReactNode } from "react";
 
 export class GlobalState {
     public onSelectionChangedObservable: Observable<any>;
@@ -22,6 +23,8 @@ export class GlobalState {
     public onPluginActivatedObserver: Nullable<Observer<ISceneLoaderPlugin | ISceneLoaderPluginAsync>>;
     public onNewSceneObservable = new Observable<Scene>();
     public sceneImportDefaults: { [key: string]: any } = {};
+
+    public onPopupRenderObservable = new Observable<Nullable<() => ReactNode>>();
 
     public validationResults: Nullable<IGLTFValidationResults> = null;
     public onValidationResultsUpdatedObservable = new Observable<Nullable<IGLTFValidationResults>>();
--- a/inspector/src/components/globalState.ts
+++ b/inspector/src/components/globalState.ts
@@ -13,6 +13,7 @@ import { PropertyChangedEvent } from "./propertyChangedEvent";
 import { ReplayRecorder } from "./replayRecorder";
 import { DataStorage } from "babylonjs/Misc/dataStorage";
 import { ReactNode } from "react";
+import { IPopupComponentProps } from "./popupComponent";
 
 export class GlobalState {
     public onSelectionChangedObservable: Observable<any>;
@@ -24,7 +25,8 @@ export class GlobalState {
     public onNewSceneObservable = new Observable<Scene>();
     public sceneImportDefaults: { [key: string]: any } = {};
 
-    public onPopupRenderObservable = new Observable<Nullable<() => ReactNode>>();
+    public onPopupPropsChangedObservable = new Observable<IPopupComponentProps>();
+    public onPopupContentRenderChangedObservable = new Observable<Nullable<() => ReactNode>>();
 
     public validationResults: Nullable<IGLTFValidationResults> = null;
     public onValidationResultsUpdatedObservable = new Observable<Nullable<IGLTFValidationResults>>();
--- a/inspector/src/components/persistentPopupHostComponent.tsx
+++ b/inspector/src/components/persistentPopupHostComponent.tsx
@@ -1,0 +1,45 @@
+import * as React from "react";
+import { PopupComponent } from "./popupComponent";
+import { Nullable } from "babylonjs/types";
+import { GlobalState } from "./globalState";
+import { Observer } from "babylonjs/Misc/observable";
+
+interface IPersistentPopupHostComponentProps {
+    globalState?: GlobalState;
+}
+
+interface IPersistentPopupHostComponentState {
+    renderFunction?: Nullable<() => React.ReactNode>;
+}
+
+/**
+ * This component is a persistent popup host that subscribes to any popup changes 
+ * @param props 
+ */
+export class PersistentPopupHostComponent extends React.Component<IPersistentPopupHostComponentProps, IPersistentPopupHostComponentState> {
+    private _renderFunctionChangeObserver?: Nullable<Observer<Nullable<() => React.ReactNode>>>;
+
+    constructor(props : IPersistentPopupHostComponentProps) {
+        super(props);
+
+        this.state = {};
+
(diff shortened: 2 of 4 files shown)


inspector/src/components/persistentPopupHostComponent.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    componentWillUnmount() {
+        if (this._renderFunctionChangeObserver) {
+            this.props.globalState?.onPopupRenderObservable.remove(this._renderFunctionChangeObserver!);
+        }
+    }
+
+    render() {
+        return (
+            this.state.renderFunction ? <PopupComponent id="test" title="test" size={{width: 100, height: 100}} onClose={() => {}}>
```
sebavan, line 40: not sure it should be called test test :-)

inspector/src/components/persistentPopupHostComponent.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Nullable } from "babylonjs/types";
+import { GlobalState } from "./globalState";
+import { Observer } from "babylonjs/Misc/observable";
+
+interface IPersistentPopupHostComponentProps {
+    globalState?: GlobalState;
+}
+
+interface IPersistentPopupHostComponentState {
+    renderFunction?: Nullable<() => React.ReactNode>;
```
RaananW, line 12: Why is this optional? Is there a case where we open a pop up and don't require to render anything there?

inspector/src/components/persistentPopupHostComponent.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from "react";
+import { PopupComponent } from "./popupComponent";
+import { Nullable } from "babylonjs/types";
+import { GlobalState } from "./globalState";
+import { Observer } from "babylonjs/Misc/observable";
+
+interface IPersistentPopupHostComponentProps {
+    globalState?: GlobalState;
```
RaananW, line 8: this should also not be optional. Otherwise this component cannot function. it is bound to the global state.

You can also use a part of the global state (i.e. the part you need), but since this is currently only used in the inspector and not shared, there is no need to decouple it from the inspector's global state.

inspector/src/components/persistentPopupHostComponent.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface IPersistentPopupHostComponentState {
+    renderFunction?: Nullable<() => React.ReactNode>;
+}
+
+/**
+ * This component is a persistent popup host that subscribes to any popup changes 
+ * @param props 
+ */
+export class PersistentPopupHostComponent extends React.Component<IPersistentPopupHostComponentProps, IPersistentPopupHostComponentState> {
+    private _renderFunctionChangeObserver?: Nullable<Observer<Nullable<() => React.ReactNode>>>;
```
RaananW, line 20: as this is initialized in the constructor it doesn't need to be optional and nullable. If it is needed, choose one of the two (right now its type is: `null | undefined | () => React.ReactNode`). Null is usually enough, unless there is explicit need for `undefined`.

inspector/src/components/persistentPopupHostComponent.tsx
RaananW: those should be configurable using the state, with default values.

inspector/src/inspector.ts
```diff
@@ -153,6 +160,8 @@ export class Inspector {
                 },
             });
             ReactDOM.render(sceneExplorerElement, this._SceneExplorerHost);
+            const popupElement = React.createElement(PersistentPopupHostComponent, {globalState: this._GlobalState});
+            ReactDOM.render(popupElement, this._PopupHost);
```
RaananW, line 164: is there a cleanup needed when a popup is closed? Or are all components being unmounted correctly (i.e. - is react mounting/unmounting process enough?).

inspector/src/components/globalState.ts
```diff
@@ -23,6 +24,8 @@ export class GlobalState {
     public onNewSceneObservable = new Observable<Scene>();
     public sceneImportDefaults: { [key: string]: any } = {};
 
+    public onPopupRenderObservable = new Observable<Nullable<() => ReactNode>>();
```
RaananW, line 27: Why is the type `Nullable`? Is there an architectural reason for that?

RaananW (commented):
Just a few questions!

The main ones are the optional and nullable variables. Feels like those are not optional but required for this component to work properly. In that case, they should not be optional or nullable.

inspector/src/components/globalState.ts
carolhmj: Yes, if the user calls the observable with a null function, it is saying they don't want to render anything anymore. That's why the PopupHost checks if renderFunction exists to render the popup. An alternative could be to have two separate observables, `onPopupStartRenderObservable` and `onPopupStopRenderObservable`.

RaananW: I see. This is a little confusing then. I would expect a function that is called `onPopupRenderObservable` to notify me when the popup is about to be rendered. It should be triggered in code, and not notified by the developer directly.

A better variant would be to have a `showPersistentPopup(options)` and `hidePersistentPopup()` functions. They in turn will notify those observing this observer.

inspector/src/components/persistentPopupHostComponent.tsx
carolhmj: For the wrapper, if it there isn't a render function it doesn't render the popup, but this could be just nullable, no need for the optional.

inspector/src/inspector.ts
carolhmj: I added a cleanup on the same function that cleans up the Tabs component, I think that will be enough

RaananW:
There is no need to use observables to change state. The observables can be used as events. i.e. - do something when the popup has opened. But they should not be used to pass content to the popup itself. In that case it would be better to have showPersistentPopup(options) and hidePersistentPopup() functions.

carolhmj:
> > Usage example:
> > 
> > * To open a popup:
> > 
> > ```js
> > globalState.onPopupContentRenderChangedObservable(() => {
> >   return <div>Popup contents...</div>
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * To close a popup:
> > 
> > ```js
> > globalState.onPopupContentRenderChangedObservable(null);
> > ```
> 
> I am not sure if you mean ".add" or ".notifyObservers", i assume you mean "notify".
> 
> There is no need to use observables to change state. The observables can be used as events. i.e. - do something when the popup has opened. But they should not be used to pass content to the popup itself. In that case it would be better to have showPersistentPopup(options) and hidePersistentPopup() functions. Those functions can internally call an observable if needed, but it would be better if the persistent pop up register those two functions in the global state and use them only when opening and closing.

Yes, sorry, it was supposed to be `notifyObservers`. Sorry, I don't think I completely understand this ideal popup structure yet. Let's say I have a React component that wants to display something in the persistent popup. Then, inside that component, should it do:

```javascript
globalState.showPersistentPopup()
```

and if that component wants to stop displaying the persistent popup, should it do:

```javascript
globalState.hidePersistentPopup()
```

then, those functions defined on the global state are the ones that communicate with the persistent popup, by notifying its observables, as an example. But how will the React component inform the popup of the content it was to render? Is this part of the options?

carolhmj:
Closing in favor of #11526

PR #11513, opened 2021-11-22
Modified onCanvasPointerOutObservable to check for bounds before notifying observers

PolygonalSun:
In Safari, there is a bug that causes a pointerout event to fire after every pointerup if pointer capture is set.  This PR changes the logic of the observable to only activate if the cursor is outside of the canvas.

### DIFF
--- a/src/Engines/engine.ts
+++ b/src/Engines/engine.ts
@@ -657,7 +657,16 @@ export class Engine extends ThinEngine {
         };
 
         this._onCanvasPointerOut = (ev) => {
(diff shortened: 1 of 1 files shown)


PolygonalSun:
Bug has been submitted to WebKit Bugzilla: https://bugs.webkit.org/show_bug.cgi?id=233432

sebavan (approved):

RaananW:
A bug in safari?!?! no way...

PR #11516, opened 2021-11-23
fix vertexdata serialize

Hypnosss:


sebavan:
The fix is not back compatible, I will push a new one in a sec. Thanks a lot for reporting it !!!

sebavan:
Here it is for reference #11517

PR #11522, opened 2021-11-24
GUI editor live connection

darraghjburke:
First pass at adding a live connection between the inspector and GUI editor. You can pop out a GUI editor window using the inspector, edit the texture, and see your changes reflected in real time.

There might definitely still be some bugs, so I would appreciate testing :)


gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -562,7 +593,41 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+            liveGuiTexture: this
+        });
+    }
+
+    /**
+     * Launch the node material editor
+     * @param config Define the configuration of the editor
+     * @return a promise fulfilled when the node editor is visible
+     */
+     public edit(config?: IGUIEditorOptions): Promise<void> {
```
sebavan, line 609: could we also allow passing in the options the GUIEditor type ? so that in case of es6 ppl would still have a way to use this function ?

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -562,7 +593,41 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+        return new Promise((resolve, reject) => {
+            this.BJSGUIEDITOR = this.BJSGUIEDITOR || this._getGlobalGUIEditor();
+            if (typeof this.BJSGUIEDITOR == 'undefined') {
```
sebavan, line 612: Nit: ===

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -562,7 +593,41 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+                const editorUrl = config && config.editorURL ? config.editorURL : AdvancedDynamicTexture.EditorURL;
```
sebavan, line 613: Nit: config?.editorURL || AdvancedDynamicTexture.EditorURL

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -562,7 +593,41 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+
+                // Load editor and add it to the DOM
+                Tools.LoadScript(editorUrl, () => {
```
sebavan, line 616: Might be good to reject on error here

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -562,7 +593,41 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+            } else {
+                // Otherwise creates the editor
+                this._createGUIEditor();
+                resolve();
+            }
+        });
+    }
+
     private _onResize(): void {
+        console.log("on resize");
```
sebavan, line 630: To be removed before exiting from draft

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -168,7 +168,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
             const json = JSON.stringify(this.props.globalState.guiTexture.serializeContent());
             StringTools.DownloadAsFile(this.props.globalState.hostDocument, json, "guiTexture.json");
         } catch (error) {
-            alert("Unable to save your GUI");
+            this.props.globalState.hostWindow.alert("Unable to save your GUI");
```
sebavan, line 171: could be interesting to log the error

guiEditor/src/diagram/workbench.tsx
```diff
@@ -26,6 +26,7 @@ import { Tools } from "../tools";
 import { CreateGround } from "babylonjs/Meshes/Builders/groundBuilder";
 import { NodeMaterial } from "babylonjs/Materials/Node/nodeMaterial";
 import { TextureBlock } from "babylonjs/Materials/Node/Blocks/Dual/textureBlock";
+import { Observer } from "babylonjs/Misc/index";
```
sebavan, line 29: should import from babylonjs/Misc/Observable

guiEditor/src/diagram/workbench.tsx
```diff
@@ -347,18 +352,45 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+            this.globalState.guiTexture.parseContent(serializationObject, true);
+        }
         this.loadToEditor();
     }
 
     async loadFromSnippet(snippedId: string) {
         this.globalState.onSelectionChangedObservable.notifyObservers(null);
-        await this.globalState.guiTexture.parseFromSnippetAsync(snippedId, true);
+        if (this.props.globalState.liveGuiTexture) {
+            await this.globalState.liveGuiTexture?.parseFromSnippetAsync(snippedId, true);
```
sebavan, line 389: shouldn t it be snippet and not snipped ?

sebavan (commented):
Just a few nits but on a general note, I am really wondering if we should add such kind of coupling between the Gui and the Editor.

The Edit function is already available on the GUIEditor package itself and easy to use. The edit function will alway be error prone, not type safe and have to deal with all the bundles possibilities. I know we have the same for inspector and such but I am really wondering if we should continue with this approach for new code ???

I ll let @RaananW decide on this and the final review

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -25,11 +25,42 @@ import { Color3 } from "babylonjs/Maths/math.color";
...
  * @see https://doc.babylonjs.com/how_to/gui
  */
 export class AdvancedDynamicTexture extends DynamicTexture {
+    /** Define the Url to load node editor script */
+    public static EditorURL = `https://unpkg.com/babylonjs-gui-editor@${Engine.Version}/babylon.guiEditor.js`;
+
+    private BJSGUIEDITOR = this._getGlobalGUIEditor();
+
+    /** Get the inspector from bundle or global */
+    private _getGlobalGUIEditor(): any {
```
RaananW, line 50: will this function work for the es6 package without declaring both `BABYLON` and `GUIEDITOR` on the global namespace? I know a similar approach is used in different areas of the framework, but we will slowly need to move out of global namespace changes.

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -25,11 +25,42 @@ import { Color3 } from "babylonjs/Maths/math.color";
...
+    editorURL?: string;
+}
+
 /**
  * Class used to create texture to support 2D GUI elements
  * @see https://doc.babylonjs.com/how_to/gui
  */
 export class AdvancedDynamicTexture extends DynamicTexture {
+    /** Define the Url to load node editor script */
+    public static EditorURL = `https://unpkg.com/babylonjs-gui-editor@${Engine.Version}/babylon.guiEditor.js`;
```
RaananW, line 45: the same comment about umd vs. es6 - this is a UMD version of the editor that might not work well if the gui package is es6

guiEditor/src/workbenchEditor.tsx
```diff
@@ -55,6 +55,9 @@ export class WorkbenchEditor extends React.Component<IGraphEditorProps, IGraphEd
         if (navigator.userAgent.indexOf("Mobile") !== -1) {
             ((this.props.globalState.hostDocument || document).querySelector(".blocker") as HTMLElement).style.visibility = "visible";
         }
+        this.props.globalState.onPopupClosedObservable.addOnce(() => {
+            this.componentWillUnmount();
```
RaananW, line 59: this is usually called by react itself, and should not be called by the developer. Is there a reason this is directly needed?
I assume you need to remove the event listener, the question is if you can do that in an outside function and let react do its cleanup on its own. Or maybe you need to make sure this component (which is the main one if I am not mistaken?) is also unmounted when the popup closes.

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -562,7 +593,41 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
+
+    /** Creates the GUI editor window. */
+    private _createGUIEditor() {
+        this.BJSGUIEDITOR.GUIEditor.Show({
+            liveGuiTexture: this
+        });
+    }
+
+    /**
+     * Launch the node material editor
```
RaananW, line 605: nit - gui editor

RaananW (commented):
Great work!

Just have a bit of concern about es6 vs. umd, and a small question regarding react.

Another main question I have is - does this code (the one to launch the gui editor) belong in the ADT code? Is the ADT the main entrance to 2D GUI or is it used for something else?

gui/src/2D/advancedDynamicTexture.ts
Popov72: Does that work?

I would rather think of `config?.editorURL ?? AdvancedDynamicTexture.EditorURL`

sebavan: it should and it would also bypass empty strings which are faulty where the "nullish" coallescing will only ever check null and undefined IIRC

guiEditor/src/workbenchEditor.tsx
darraghjburke: I'm not sure why this one is necessary, it's just copied from NME. I will check if we can just call the cleanup code directly

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,54 @@
...
+        return undefined;
+    }
+
+    public static async EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture) {
+        if (!this.guiEditor) {
+            if (typeof BABYLON !== 'undefined') {
+                // we are in UMD environment
+                this.guiEditor = this.guiEditor || this._getGlobalGUIEditor();
+                if (typeof this.guiEditor == 'undefined') {
+                    const editorUrl = `https://unpkg.com/babylonjs-gui-editor@${Engine.Version}/babylon.guiEditor.js`;;
```
RaananW, line 32: will be good to allow to override this url in case you want to host it yourself.

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,54 @@
...
+                        await Tools.LoadScriptAsync(editorUrl);
+                        this.guiEditor = this.guiEditor || this._getGlobalGUIEditor();
+                    } catch {
+                        Tools.Error(`Failed to load GUI editor from ${editorUrl}`);
+                    }
+                }
+            } else {
+                // we are in ES6 environment
+                try {
+                    this.guiEditor = await import("@babylonjs/gui-editor");
```
RaananW, line 45: nit - for future reference, we'll need to add the gui editor to the peer dependencies of this package.

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,54 @@
+import { AdvancedDynamicTexture } from "babylonjs-gui/2D/advancedDynamicTexture";
+import { Engine } from "babylonjs/Engines/engine";
+import { Tools } from "babylonjs/Misc/tools";
+
+
+declare var GUIEDITOR: any;
+declare var BABYLON: any;
+
+export class GUITools {
```
RaananW, line 9: in general, as this has only static members, there is no need to create the class for it. those can be module members:
roughly this:

```javascript
function _getGlobalGUIEditor(): any {
        // UMD Global name detection from Webpack Bundle UMD Name.
        if (typeof GUIEDITOR !== 'undefined') {
            return GUIEDITOR;
        }

        // In case of module let's check the global emitted from the editor entry point.
        if (typeof BABYLON !== 'undefined' && typeof BABYLON.GUIEditor !== 'undefined') {
            return BABYLON;
        }

        return undefined;
    }
}

export function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture) {
        if (!this.guiEditor) {
            if (typeof BABYLON !== 'undefined') {
                // we are in UMD environment
                this.guiEditor = this.guiEditor || this._getGlobalGUIEditor();
                if (typeof this.guiEditor == 'undefined') {
                    const editorUrl = `https://unpkg.com/babylonjs-gui-editor@${Engine.Version}/babylon.guiEditor.js`;;
                    // Load editor and add it to the DOM
                    try {
                        await Tools.LoadScriptAsync(editorUrl);
                        this.guiEditor = this.guiEditor || this._getGlobalGUIEditor();
                    } catch {
                        Tools.Error(`Failed to load GUI editor from ${editorUrl}`);
                    }
                }
            } else {
                // we are in ES6 environment
                try {
                    this.guiEditor = await import("@babylonjs/gui-editor");
                } catch {
                    Tools.Error("Failed to import @babylonjs/gui-editor. Please install the package if you want to use the GUI editor.");
                }
            }
        }
        this.guiEditor.GUIEditor.Show({liveGuiTexture: adt});
    }
```

RaananW (commented):
LGTM!

just a couple suggestions.

RaananW:
There seems to be an issue with the import, as the type is not present. this might be a bit more complicated when dealing with UMD, as the type will not be a part of the package. Let's talk about it offline.

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
darraghjburke: Makes sense to me. Should I put it in a namespace or just export the

RaananW: no need to namespace, just the functions on top level

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,56 @@
...
+            guiEditor = guiEditor || _getGlobalGUIEditor();
+            if (typeof guiEditor == 'undefined') {
+                const editorUrl = customEditorURL || `https://unpkg.com/babylonjs-gui-editor@${Engine.Version}/babylon.guiEditor.js`;;
+
+                // Load editor and add it to the DOM
+                try {
+                    await Tools.LoadScriptAsync(editorUrl);
+                    guiEditor = guiEditor || _getGlobalGUIEditor();
+                } catch {
+                    Tools.Error(`Failed to load GUI editor from ${editorUrl}`);
```
RaananW, line 47: You should throw this back, otherwise it will try running the `.Show` function even after it failed.

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,56 @@
...
+export function InjectGUIEditor(guiEditorPackage: any) {
+    guiEditor = guiEditorPackage;
+}
+
+/**
+ * Opens an ADT in the GUI editor
+ * if you are in an ES6 environment, you must first call InjectGUIEditor to provide the gui-editor package
+ * If you are in a UMD environment, you have the option to load the editor from a custom URL, or use the default (unpkg CDN) 
+*/
+export async function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture, customEditorURL?: string) {
```
RaananW, line 34: I think we should accept a package and not a custom URL. or add the ability to set the package URL (just like the inject function).

RaananW (commented):
Just two nits, and Seb's unresolved comment about logging.

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (!guiEditor) {
+        if (typeof BABYLON !== 'undefined') {
+            // we are in UMD environment
+            guiEditor = guiEditor || _getGlobalGUIEditor();
+            if (typeof guiEditor == 'undefined') {
```
RaananW, line 39:
```suggestion
            if (typeof guiEditor === 'undefined') {
```

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
darraghjburke: How's this?

RaananW (approved):
Amazing work!

Please merge with latest master to run the corrected CI again

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -0,0 +1,60 @@
...
+                try {
+                    await Tools.LoadScriptAsync(editorUrl);
+                    guiEditor = guiEditor || _getGlobalGUIEditor();
+                } catch {
+                    throw `Failed to load GUI editor from ${editorUrl}`;
+                }
+            }
+        } else {
+            // we are in ES6 environment
+            Tools.Error(`Tried to call EditAdvancedDynamicTexture without first injecting the GUI editor. You need to call InjectGUIEditor() with a reference to @babylonjs/gui-editor. It can be imported at runtime using await import("@babylonjs/gui-editor").`);
```
RaananW, line 56: Just one last final nit :-)

You should throw here as well. or make sure that guiEditor is defined before running line 59.

RaananW (approved):

inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
RaananW: great :-)

sebavan (approved):

PR #11525, opened 2021-11-24
[XR] fix an issue with firefox reality

RaananW:
Following up this - https://forum.babylonjs.com/t/a-vr-glasses-problem/25500
Even though this should never be undefined, it seems like in the case of firefox reality it is.
Make sure to use the predefined variable (in the outer scope( and wrap this in a try-catch to avoid any future errors.


sebavan (approved):
LGTM

PR #11527, opened 2021-11-24
Add tooltips to inspector items

darraghjburke:
It's currently very difficult to read lengthy names in the inspector. Your only option is to increase the width of the inspector, and there's a limit on how large it can grow to, so some names are just impossible to read. This is especially problematic because you'll often have multiple objects in a scene with names like `SomeReallyLongName_1` and `SomeReallyLongName_2`, making it difficult to distinguish between them.

This PR simply adds a tooltip so you can hover over an element and see the full name without needing to click on it.

image


sebavan (approved):
GG

PR #11528, opened 2021-11-24
Fix for link with mesh and hardware scaling

msDestiny14:
Have link with mesh now accounting for root container scaling.
 
In addition added an extra parameter that will auto scale the root to match the hardware scaling when they create their fullscreen adt right up front. 

https://forum.babylonjs.com/t/crisp-text-and-linkwithmesh-results-in-incorrect-gui-position/25340/


gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -1154,15 +1165,23 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
         var result = new AdvancedDynamicTexture(name, 0, 0, scene, false, sampling);
         // Display
         var layer = new Layer(name + "_layer", null, scene, !foreground);
         layer.texture = result;
         result._layerToDispose = layer;
         result._isFullscreen = true;
+
+        const resultScene = result.getScene();
+        if(adaptiveScaling && resultScene) {
+            result._rootContainer.scaleX = 1 / resultScene.getEngine().getHardwareScalingLevel();
```
sebavan, line 1181: 1 / ... could be computed only once

sebavan (commented):
Only one nit :-)

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -1154,15 +1165,23 @@ export class AdvancedDynamicTexture extends DynamicTexture {
      * @param foreground defines a boolean indicating if the texture must be rendered in foreground (default is true)
      * @param scene defines the hosting scene
      * @param sampling defines the texture sampling mode (Texture.BILINEAR_SAMPLINGMODE by default)
+     * @param adaptiveScaling defines whether to automatically scale root to match hardwarescaling (false by default)
      * @returns a new AdvancedDynamicTexture
      */
-    public static CreateFullscreenUI(name: string, foreground: boolean = true, scene: Nullable<Scene> = null, sampling = Texture.BILINEAR_SAMPLINGMODE): AdvancedDynamicTexture {
+    public static CreateFullscreenUI(name: string, foreground: boolean = true, scene: Nullable<Scene> = null, sampling = Texture.BILINEAR_SAMPLINGMODE, adaptiveScaling ?: boolean): AdvancedDynamicTexture {
```
sebavan, line 1171: Also put = false instead of ? in here as a default for consistency

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -594,7 +594,18 @@ export class AdvancedDynamicTexture extends DynamicTexture {
     /** @hidden */
     public _getGlobalViewport(): Viewport {
         var size = this.getSize();
-        return this._fullscreenViewport.toGlobal(size.width, size.height);
+        const globalViewPort = this._fullscreenViewport.toGlobal(size.width, size.height);
+
+        const targetX = globalViewPort.width * (1 / this.rootContainer.scaleX);
+        const targetY = globalViewPort.height * (1 / this.rootContainer.scaleY);
```
ecoinomist, line 600: Please consider Math.round() because of JS floating point issues (you may end up with `Float` width/height, but they should be `Integer`)

gui/src/2D/advancedDynamicTexture.ts
```diff
@@ -1154,15 +1165,24 @@ export class AdvancedDynamicTexture extends DynamicTexture {
...
-    public static CreateFullscreenUI(name: string, foreground: boolean = true, scene: Nullable<Scene> = null, sampling = Texture.BILINEAR_SAMPLINGMODE): AdvancedDynamicTexture {
+    public static CreateFullscreenUI(name: string, foreground: boolean = true, scene: Nullable<Scene> = null, sampling = Texture.BILINEAR_SAMPLINGMODE, adaptiveScaling : boolean = false): AdvancedDynamicTexture {
         var result = new AdvancedDynamicTexture(name, 0, 0, scene, false, sampling);
         // Display
         var layer = new Layer(name + "_layer", null, scene, !foreground);
         layer.texture = result;
         result._layerToDispose = layer;
         result._isFullscreen = true;
+
+        const resultScene = result.getScene();
```
RaananW, line 1179: scene is already defined, is that a different scene? can it be a different scene?

RaananW (commented):
One question regarding the scene, otherwise LGTM.

Please check the comment added (about rounding the value) - make sure it's needed and correct.

RaananW:
There is also a small linting issue :-)

gui/src/2D/advancedDynamicTexture.ts
msDestiny14: This is that the user can pass in and it can be null. If it is null the layer constructor takes care of it with. 
 this._scene = <Scene>(scene || EngineStore.LastCreatedScene);

RaananW: I understand that this will (probably) never fail, but it is prone to misunderstandings without diving into the code.

You pass a scene object that can be null, you use it to create an ADT, then you use it to create a layer, and then you are taking the scene from the ADT and use it (to get the engine).
You can either say - "the scene is `(scene || EngineStore.LastCreatedScene);`", or initially use the scene from the ADT to also create the layer. This way you are staying clear and consistent. Option 2 is the better one :-)

gui/src/2D/advancedDynamicTexture.ts
msDestiny14: Opening an issue to investigate at a later time. 
https://github.com/BabylonJS/Babylon.js/issues/11539

RaananW (approved):

Issue #10338, opened 2021-05-07
Have PrePassRenderer take `disableDepthWrite` flag into account


CraigFeldspar:
PrePassRenderer renders depth regardless of this flag.

CraigFeldspar:
on it

thomlucc:
fixed with #11533, thanks @CraigFeldspar!

PR #11533, opened 2021-11-26
prepass now takes depthwrite flag into account

CraigFeldspar:
Related to #10338

### DIFF
--- a/src/Rendering/prePassRenderer.ts
+++ b/src/Rendering/prePassRenderer.ts
@@ -265,17 +265,18 @@ export class PrePassRenderer {
      * @param subMesh Submesh on which the effect is applied
      */
     public bindAttachmentsForEffect(effect: Effect, subMesh: SubMesh) {
+        const material = subMesh.getMaterial();
+        const isPrePassCapable = material && material.isPrePassCapable;
+        const excluded = material && this.excludedMaterials.indexOf(material) !== -1;
+
         if (this.enabled && this._currentTarget.enabled) {
-            if (effect._multiTarget) {
+            if (effect._multiTarget && isPrePassCapable && !excluded) {
                 this._engine.bindAttachments(this._multiRenderAttachments);
             } else {
                 this._engine.bindAttachments(this._defaultAttachments);
 
-                if (this._geometryBuffer && this.currentRTisSceneRT) {
-                    const material = subMesh.getMaterial();
-                    if (material && !material.isPrePassCapable && this.excludedMaterials.indexOf(material) === -1) {
-                        this._geometryBuffer.renderList!.push(subMesh.getRenderingMesh());
-                    }
+                if (this._geometryBuffer && this.currentRTisSceneRT && !excluded) {
+                    this._geometryBuffer.renderList!.push(subMesh.getRenderingMesh());
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/Rendering/prePassRenderer.ts
```diff
@@ -265,17 +265,18 @@ export class PrePassRenderer {
...
             } else {
                 this._engine.bindAttachments(this._defaultAttachments);
 
-                if (this._geometryBuffer && this.currentRTisSceneRT) {
-                    const material = subMesh.getMaterial();
-                    if (material && !material.isPrePassCapable && this.excludedMaterials.indexOf(material) === -1) {
-                        this._geometryBuffer.renderList!.push(subMesh.getRenderingMesh());
-                    }
+                if (this._geometryBuffer && this.currentRTisSceneRT && !excluded) {
+                    this._geometryBuffer.renderList!.push(subMesh.getRenderingMesh());
```
Popov72, line 279: That will add the mesh into `renderList` each frame (in case the `if` condition is `true`)? `bindAttachmentsForEffect` is called by the `scene._beforeRenderingMeshStage` observable.

CraigFeldspar: yes indeed, but we clear the renderlist before each render pass here https://github.com/BabylonJS/Babylon.js/blob/b540dcdcff44fe30ced78241bb4a2f0d77ed7ba7/src/Rendering/prePassRenderer.ts#L389

Popov72 (approved):

src/Rendering/prePassRenderer.ts
sebavan: Is there a way to prevent GC ? maybe using renderList.length = 0 instead ???

RaananW:
will be discussed after merging, discussed internally with seb

Popov72:
The changes in `prePassRenderer.ts` make the validation test "Screen space reflections" fail: we don't see it because the test is disabled on the Azure server.

We will need to regenerate the picture, however there's still a problem somewhere because the rendering is not the same with the prepass renderer and with the geometry renderer:
* prepass: https://playground.babylonjs.com/#PIZ1GK#172
* geometry: https://playground.babylonjs.com/#PIZ1GK#308

If you open the PGs in two different tabs you will see there are some small differences in rendering but more importantly there are some artifacts when you move in the prepass case.

sebavan:
@CraigFeldspar could you have a look ?

CraigFeldspar:
> The changes in `prePassRenderer.ts` make the validation test "Screen space reflections" fail: we don't see it because the test is disabled on the Azure server.
> 
> We will need to regenerate the picture, however there's still a problem somewhere because the rendering is not the same with the prepass renderer and with the geometry renderer:
> 
> * prepass: https://playground.babylonjs.com/#PIZ1GK#172
> * geometry: https://playground.babylonjs.com/#PIZ1GK#308
> 
> If you open the PGs in two different tabs you will see there are some small differences in rendering but more importantly there are some artifacts when you move in the prepass case.

This is epic.
Actually it didn't introduce a new bug, it just revealed one that probably existed since the prepass implementation.
This PG happens to have `skybox.material.disableDepthWrite = true;` So now that it is correctly taken into account, test breaks, but you noticed that it is still different from GBR. Indeed GBR always took `disableDepthWrite` into account, but also reflectivity "jitter" which is stored in the reflectivitySampler alpha channel. Which was not written in prepass.

I'm making a PR and fixing the test shortly

PR #11537, opened 2021-11-29
Fix some .babylon files not loading anymore

Popov72:
See https://forum.babylonjs.com/t/upgrading-from-alpha-37-to-alpha-61-multiple-issues/25603/8

`isNaN("12")` is false, which was the problem with the file linked in the forum thread.

`if (!Number.isFinite(parentId))` would be a better fix but it is not supported by IE11... See https://stackabuse.com/javascript-check-if-variable-is-a-number/

Note that `typeof` does not catch the `Infinite` / `NaN` values (those are `number` according to `typeof`) values but it's not a problem here.

### DIFF
--- a/src/Loading/Plugins/babylonFileLoader.ts
+++ b/src/Loading/Plugins/babylonFileLoader.ts
@@ -118,7 +118,7 @@ var loadDetailLevels = (scene: Scene, mesh: AbstractMesh) => {
 };
 
 var findParent = (parentId: any, scene: Scene) => {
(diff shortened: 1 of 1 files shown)


RaananW (approved):
LGTM!

I wonder - why were we testing for NaN before?

Popov72:
I think it was a test meant to know if we are dealing with a number / not a number (string mostly) but unfortunately isNaN("XXX") where XXX is a number returns false.

PR #11547, opened 2021-11-29
Observable coroutine fix

syntheticmagus:
Fix from @ryantrem resolving an issue that was causing coroutines on observables to fail to execute in parallel, plus a minor name change to help distinguish per-step callbacks from coroutine success callbacks.


bghgary:
Should we consider adding some unit tests for this?

ryantrem (approved):

sebavan (approved):

sebavan:
I think @ryantrem added some ? and we should probably expand the list ?

ryantrem:
> I think @ryantrem added some ? and we should probably expand the list ?

I added unit tests for coroutines, but not for observableCoroutine.

ryantrem (approved):

PR #11561, opened 2021-12-01
Set the generateMipMaps property for basis textures

### LINKED ISSUE #11549: Loaded .basis textures don't actually have LINEAR_LINEAR_MIPLINEAR sampling initially
Issue stemming from forum: https://forum.babylonjs.com/t/no-mipmap-for-basis-universal-texture/25403/7

This .basis texture loads with a sampling mode of LINEAR_LINEAR_MIPLINEAR. However, if run updateSamplingMode(LINEAR_LINEAR_MIPLINEAR), the texture breaks. This can only be fixed by setting it to some sampling mode without a _MIP option. You can play around with the different sampling modes in the inspector to see which ones work.

There really might be two issues here:
- ~~why does the texture break under any sampling mode with a _MIP option?~~ Edit: according to forum thread, this is because the texture was not built with the -mipmaps flag on. This makes sense. However, the following issue is still here:
- why does the texture not have that sampling mode set initially, even though the texture object thinks it does?

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com/#E4VDDW#10):
- Expected result: the texture should not break when you set the sampling mode to LINEAR_LINEAR_MIPLINEAR, as that should be the sampling mode already (check console for evidence)
- Current result: the object becomes black.

 - OS: Windows
 - Browser Edge (repros in FF as well)
 - Version 96.0.1054.34

Popov72:
Fix #11549

### DIFF
--- a/src/Misc/basis.ts
+++ b/src/Misc/basis.ts
@@ -235,6 +235,7 @@ export const LoadTextureFromTranscodeResult = (texture: InternalTexture, transco
         } else {
             texture.width = rootImage.width;
             texture.height = rootImage.height;
+            texture.generateMipMaps = transcodeResult.fileInfo.images[i].levels.length > 1;
 
             // Upload all mip levels in the file
             transcodeResult.fileInfo.images[i].levels.forEach((level: any, index: number) => {


sebavan (approved):
You rock

PR #11562, opened 2021-12-01
Bug fix for if host is null

msDestiny14:
Host can be null if not assigned to an adt, for example in imageBasedSlider.


sebavan (approved):

sebavan:
@msDestiny14 can you update to master to see if the module tests are fixed ?

PR #11563, opened 2021-12-01
[GUI] Correct pointer events for rig cameras

### LINKED ISSUE #11544: Collision detection incorrect for GUI with StereoscopicCamera
Collision acts as if it was just a fullscreen viewport. 

https://forum.babylonjs.com/t/hey-guys-i-got-a-confusing-trouble-when-i-useing-gui-with-stereoscopiccamera/25589

RaananW:
Pointer events were using the current active camera, but in case of a camera with rig children it would take the parent camera and its viewport.

This PR sets the camera to be used with pointers for the rig camera that the mouse is in its viewport (and resets it afterwards).

Fixes #11544


sebavan (approved):

deltakosh (approved):
Smart!

PR #11572, opened 2021-12-02
Checkbox and color picker perf profiler design

carolhmj:
The current category values are placeholders for now.


sharedUiComponents/lines/checkBoxLineComponent.tsx
```diff
@@ -85,9 +88,11 @@ export class CheckBoxLineComponent extends React.Component<ICheckBoxLineComponen
...
                 {this.props.icon && <img src={this.props.icon} title={this.props.iconLabel} alt={this.props.iconLabel} className="icon" />}
-                <div className="label" title={this.props.iconLabel}>
-                    {this.props.label}
-                </div>
+                {this.props.label && 
+                    <div className="label" title={this.props.iconLabel}>
+                        {this.props.label}
+                    </div>}
+                {this.props.faIcons && <FontAwesomeIcon icon={this.state.isSelected ? this.props.faIcons.faIconEnabled : this.props.faIcons.faIconDisabled} onClick={() => !this.props.disabled && this.onChange()}/>}
                 <div className="checkBox">
```
RaananW, line 96: maybe instead of hiding it in css either show the icons of the actual checkbox. Would also be great to allow the disabled state even when icons are used.

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerSidebarComponent.tsx
```diff
@@ -54,17 +75,27 @@ export const PerformanceViewerSidebarComponent = (props: IPerformanceViewerSideb
...
                     {metadataCategoryId?.get(category)?.map((id) => {
                         const metadata = metadataMap?.get(id);
-                        return metadata && <div key={`perf-sidebar-item-${id}`} className="sidebar-item">
-                            <input type="checkbox"checked={!metadata.hidden} onChange={onCheckChange(id)} />
-                            <ColorPickerLineComponent value={Color3.FromHexString(metadata.color ?? "#000")} onColorChanged={onColorChange(id)} shouldPopRight />
-                            <span className="sidebar-item-label">{id}</span>
+                        return metadata && <div key={`perf-sidebar-item-${id}`} className="sidebar-item measure">
+                            {/* div with check box, color picker and category name */}
+                            <div className="category">
+                                <CheckBoxLineComponent isSelected={() => !metadata.hidden} onSelect={onCheckChange(id)} faIcons={{faIconEnabled: faCheckSquare, faIconDisabled: faSquare}} />
```
RaananW, line 91: i think there is no need for the `faIcon` in the keys, it is already defined in the name of the parameter.

RaananW (commented):
Awesome :-)

Just 2 nits

sharedUiComponents/lines/checkBoxLineComponent.tsx
carolhmj: I put a conditional to not render the checkbox if the faIcons property isn't undefined, but I've left the checkBox classname "hidden" since the other components that use the checkbox use that class.

PR #11575, opened 2021-12-02
Eliminate coroutine step recursion

ryantrem:
I tried profiling `MergeMeshesAsync` in Safari and it was basically un-profilable. :) The problem was that coroutines can be stepped synchronously (either in batches for the yielding scheduler, or entirely with the inline scheduler), and when they are stepped synchronously it was done recursively (the `resume` function calling the `resume` function). This resulted in very deep call stacks, with CPU time spread over many stack frames that are conceptually the same function. To fix this, I changed the `runCoroutine` function to loop as long as stepping is synchronous, and only exit the loop when a step is asynchronous (in which case we rely on completion of the async step to resume the coroutine and start the loop again). This makes profiling wayyyy better.

### DIFF
--- a/src/Misc/coroutine.ts
+++ b/src/Misc/coroutine.ts
@@ -85,42 +85,39 @@ export function createYieldingScheduler<T>(yieldAfterMS = 25) {
 /** @hidden */
 export function runCoroutine<T>(coroutine: AsyncCoroutine<T>, scheduler: CoroutineScheduler<T>, onSuccess: (result: T) => void, onError: (error: any) => void, abortSignal?: AbortSignal) {
     const resume = () => {
-        let shouldReschedule: boolean | undefined;
+        let reschedule: boolean | undefined;
+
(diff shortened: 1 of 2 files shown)


sebavan (approved):

src/Misc/coroutine.ts
```diff
@@ -84,26 +84,40 @@ export function createYieldingScheduler<T>(yieldAfterMS = 25) {
...
+            reschedule = undefined;
+
+            if (!abortSignal || !abortSignal.aborted) {
+                scheduler(coroutine, onStep, onError);
+            } else {
+                onError(new Error("Aborted"));
+            }
+
+            if (reschedule === undefined) {
+                // If shouldReschedule is defined at this point, then the coroutine must have stepped asynchronously, so stop looping and let the coroutine be resumed later.
```
RaananW, line 116: nit - did you mean undefined in the doc?

RaananW (approved):

Issue #10056, opened 2021-03-15
WebXRInput rapidly recreates Textures for hand tracking controllers whether or not hand tracking WebXR feature is enabled in BabylonReactNative


chrisfromwork:
**Repro**
- Run BabylonReactNative Playground app on HoloLens 2 and enter immersive xr mode: https://github.com/BabylonJS/BabylonReactNative/tree/master/Apps/Playground 

**Additional context**
- Texture creation occurs for WebXRHandTracking sources whether or not WebXRHandTracking is enabled
- The texture creation is specifically kicked off by WebXRInput when observing that controllers have been added/removed. For whatever reason, a WebXR hand seems to register as a controller
- If I understand correctly, there may be 5+ textures created for the input source. None of these textures are currently used to display hand tracking meshes
- When hands are brought in and out of the scene (find and lose tracking) we get texture recreation. This causes rendering hiccups/poor performance. It seems like these textures should at least be cached so that we don't constantly recreate them. It is common for hands to lose tracking during normal application life cycles

RaananW:
I believe this should be divided to a few issues, but is not a bug.

1) The textures are coming from a motion controller that is created when the hands are detected. A motion controller is initialized to each input source in tracked-pointer mode that has the `.gamepad` object in it. A hand object doesn't have a `.gamepad` defined, but a `.hand` with different definitions (see the IDL here - https://www.w3.org/TR/webxr-hand-input-1/#idl-index). Could that be an issue in Native? 
2) We should make sure that hand meshes can be reused (the tracked meshes)  and are not disposed, especially since we expect to never hand more than 2 hands in the scene, and have one left and one right hand, which can be cached.

chrisfromwork:
We will use this issue to track material caching/avoiding creating textures/meshes for input sources that declare a hand property.

I've opened the following issue to track BabylonNative changes:
https://github.com/BabylonJS/BabylonNative/issues/635

RaananW:
Can this be closed?

rickfromwork:
Related, this issue tracks updating Babylon to not create controller models when a hand object exists, and anything else that needs to change with the knowledge that webXR now sends hand input with both a hand object and gamepad object.
#10144

rickfromwork:
@RaananW Is this now resolved with #10188, or should there be an additional change along the lines of enabling the caching by default for hands?

RaananW:
[...]

PR #11583, opened 2021-12-03
[XR] cache hand meshes for faster loading

RaananW:
This PR does 2 things to enalbe faster loading of the hand meshes in XR:

1. It caches the hand meshes so that entering the session again will load the meshes instantly instead of downloading them again
2. Download the meshes in parallel and not wait between the hands.

### DIFF
--- a/src/XR/features/WebXRHandTracking.ts
+++ b/src/XR/features/WebXRHandTracking.ts
@@ -12,7 +12,7 @@ import { WebXRFeaturesManager } from "../webXRFeaturesManager";
 import { IDisposable, Scene } from "../../scene";
 import { Observable } from "../../Misc/observable";
 import { InstancedMesh } from "../../Meshes/instancedMesh";
-import { SceneLoader } from "../../Loading/sceneLoader";
+import { ISceneLoaderAsyncResult, SceneLoader } from "../../Loading/sceneLoader";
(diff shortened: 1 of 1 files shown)


RaananW:
Requested here - https://github.com/BabylonJS/Babylon.js/issues/10056

src/XR/features/WebXRHandTracking.ts
```diff
@@ -553,9 +556,22 @@ export class WebXRHandTracking extends WebXRAbstractFeature {
     private static _generateDefaultHandMeshesAsync(scene: Scene, options?: IWebXRHandTrackingOptions): Promise<{ left: AbstractMesh, right: AbstractMesh }> {
         return new Promise(async (resolve) => {
             const riggedMeshes: { [handedness: string]: AbstractMesh } = {};
+            // check the cache, defensive
```
sebavan, line 559: Looks all good to me but wondering if the cache should be per scene ???

RaananW: XR is scene-bound because of all of the resources. This can only happen if the scene was already disposed, including all of its resources (i.e. - running a playground again)

sebavan: Thanks

sebavan (approved):

PR #11588, opened 2021-12-04
Fix HolographicButton and Button3D when scene is right handed.

### LINKED ISSUE #11543: Userighthandedsystem + button text bug
When using scene.useRightHandedSystem = true; Appears that the text was flipped correctly but button isn't flipped.
Or should the text not flip at all? 

https://forum.babylonjs.com/t/userighthandedsystem-button-text-bug/25547

carolhmj:
Fix #11543
https://forum.babylonjs.com/t/userighthandedsystem-button-text-bug/25547

https://playground.babylonjs.com/#HB4C01#467
You can check on this playground, without the fix:
image

With the fix, since the button faces away from the viewer, you rotate the camera and the text has the proper orientation:
image


sebavan (approved):
Nicely DONE !!!!!!! GG

Popov72:
> With the fix, since the button faces away from the viewer, you rotate the camera and the text has the proper orientation:

I don't really understand this one? We should not have to move the camera to make the PG work when in right handed system. The buttons don't face away from the viewer when in rhs, it's only that everything is mirrored on the X axis.

Take for eg this PG (in lhs): https://playground.babylonjs.com/#HB4C01#470

Everything is ok, the sphere is on the left, we have the holographic button then the 3D button (from left to right) and the texts on the buttons are ok:
image

Now, uncomment line 7 to switch to rhs. Without the PR we get:
image

which is expected: everything is now mirrored on the X axis (the sphere is now on the right, the 3D/holographic buttons are also inverted on the X axis) and the texts are inverted (which is the bug).

The only change we expect with the PR (at least it's what I think and what I think someone making this PG would expect) is that the texts are now ok but instead we must rotate the camera 180° to get readable texts.

I think in the PR you should simply invert the textures, meaning keep what you did except the changes in `HolographicButton.tooltipText`, `HolographicButton._createNode` and `Button3D._createNode`. In `HolographicButton.tooltipText`, you should invert the tooltip texture:
```javasccript
// Create text texture for the tooltip
this._tooltipTexture = AdvancedDynamicTexture.CreateForMesh(this._tooltipMesh);
if (this._backPlate._scene.useRightHandedSystem) {
    this._tooltipTexture.wrapU = Texture.MIRROR_ADDRESSMODE;
    this._tooltipTexture.uOffset = 1;
}
```
That way we would have what I think is the right display from the PG above:
image

sebavan:
> > With the fix, since the button faces away from the viewer, you rotate the camera and the text has the proper orientation:
> 
> I don't really understand this one? We should not have to move the camera to make the PG work when in right handed system. The buttons don't face away from the viewer when in rhs, it's only that everything is mirrored on the X axis.
> 
> Take for eg this PG (in lhs): https://playground.babylonjs.com/#HB4C01#470
> 
> Everything is ok, the sphere is on the left, we have the holographic button then the 3D button (from left to right) and the texts on the buttons are ok: image
> 
> Now, uncomment line 7 to switch to rhs. Without the PR we get: image
> 
> which is expected: everything is now mirrored on the X axis (the sphere is now on the right, the 3D/holographic buttons are also inverted on the X axis) and the texts are inverted (which is the bug).
> 
> The only change we expect with the PR (at least it's what I think and what I think someone making this PG would expect) is that the texts are now ok but instead we must rotate the camera 180° to get readable texts.
> 
> I think in the PR you should simply invert the textures, meaning keep what you did except the changes in `HolographicButton.tooltipText`, `HolographicButton._createNode` and `Button3D._createNode`. In `HolographicButton.tooltipText`, you should invert the tooltip texture:
> 
> ```
> // Create text texture for the tooltip
> this._tooltipTexture = AdvancedDynamicTexture.CreateForMesh(this._tooltipMesh);
> if (this._backPlate._scene.useRightHandedSystem) {
>     this._tooltipTexture.wrapU = Texture.MIRROR_ADDRESSMODE;
>     this._tooltipTexture.uOffset = 1;
> }
> ```
> 
> That way we would have what I think is the right display from the PG above: image
> 
> But maybe I'm wrong in my expectations?

good catch, I ll let you check with @carolhmj before we merge.

carolhmj:
> I don't really understand this one? We should not have to move the camera to make the PG work when in right handed system. The buttons don't face away from the viewer when in rhs, it's only that everything is mirrored on the X axis.

Hmmm, now that I'm looking at this again, I think the desired result depends on how you expect the user to place the camera. I drew this to help visualize what's going on:
buttonandhandedness

On the LHS, the natural camera position is along the -Z axis, so the front of the button faces that axis. If we move to a RHS, but keep the camera positioned on the -Z axis, then we would still want the front of the buttons the same way, and just the X axis would be flipped. But isn't it more usual to place the camera along the +Z axis when using a RHS? (At least, that's what I used to see in OpenGL). Then in this case, we probably would want to have the front of the button facing the +Z axis instead.

@sebavan @Popov72 I'm unsure what the best option would be here. Maybe let the user chose with a boolean?

Popov72:
> But isn't it more usual to place the camera along the +Z axis when using a RHS?

No, for me camera position and rhs/lhs is independent. Take this PG for eg: https://playground.babylonjs.com/#HB4C01#468

It is in lhs mode. If you uncomment line 7 it is in rhs mode: the camera has not moved, the object has not moved and has not been flipped but the texts on the buttons are now reversed: I _think_ the user would expect the texts on the button not to be reversed instead of having to move the camera (or flip the buttons).

However, it's up to us to decide what is the right solution as both are valid: in both cases we can display the buttons the other way around by flipping them or the camera, so we are able to achieve the result we want.

rickfromwork:
Can you update this PR to also fix the TouchHolographicButton?

carolhmj:
> Can you update this PR to also fix the TouchHolographicButton?

Sure! Lemme just ping @deltakosh @RaananW first so we can talk about the camera question first. 

What do you think would be better on right handed mode: flip only the texture but leave the button facing the -Z axis, or flip the texture and the buttons so it faces the +Z?

RaananW:
> Sure! Lemme just ping @deltakosh @RaananW first so we can talk about the camera question first.

So far we are rendering the scene exactly the same on right handed and left handed systems. so the camera's z is negated, but the relevant meshes are also flipped. I would also flip the buttons, as "forward" is not backwards.

Popov72:
> So far we are rendering the scene exactly the same on right handed and left handed systems

Note sure what you mean here. If you are loading gltf/glb files then setting rhs or lhs won't change anything visually because we inject a root node with the proper transformation so that the display does not change. However, in all other cases (you load other types of files or you create meshes yourself), the scene will be mirrored on the X axis if you change from lhs to rhs - so the scene won't render the same way.

RaananW:
> > So far we are rendering the scene exactly the same on right handed and left handed systems
> 
> Note sure what you mean here. If you are loading gltf/glb files then setting rhs or lhs won't change anything visually because we inject a root node with the proper transformation so that the display does not change. However, in all other cases (you load other types of files or you create meshes yourself), the scene will be mirrored on the X axis if you change from lhs to rhs - so the scene won't render the same way.

Cleared that with carol offline. You are right, I should have elaborated more.  will be happy to explain, but it behaves as expected.

PR #11589, opened 2021-12-06
get attachedNode in PositionGizmo always return null

jtcheng:
get attachedNode in PositionGizmo always return null.

Playground:
https://playground.babylonjs.com/#31M2AP#243


sebavan (commented):
Not sure it is not by designed if the node are disabled. @CedricGuillemet could you have a look ?

CedricGuillemet:
This looks like it was a bug. Rotation/Scale gizmo are ok but positionGizmo was not. Thank you for finding and fixing it @jtcheng !

RaananW (approved):
(discussed with @CedricGuillemet)

PR #11591, opened 2021-12-06
[GE] Selection and Movement all in one!

msDestiny14:
- Now users can select and move together.
- To get the child element, double click to drill down the stack.


guiEditor/src/diagram/workbench.tsx
```diff
@@ -456,6 +480,20 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+        if (this._mainSelection === parent) {
+            return true;
+        }
+        while (parent.parent) {
+            parent = parent.parent;
+            if (this._mainSelection === parent) {
+                return true;
+            }
+        }
+        return false;
```
RaananW, line 494:
```suggestion
    do {
        if (this._mainSelection === control) {
            return true;
        };
        control = control.parent;
    } while(control);
```

guiEditor/src/diagram/workbench.tsx
```diff
@@ -464,7 +502,17 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         const onPointerDown = guiControl.onPointerDownObservable.add((evt) => {
             if (!this.isUp || evt.buttonIndex > 0) return;
             if (this._forceSelecting) {
-                this.isSelected(true, guiControl);
+                console.log(guiControl.name);
```
RaananW, line 505: you forgot this one here

guiEditor/src/diagram/workbench.tsx
```diff
@@ -464,7 +502,17 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         const onPointerDown = guiControl.onPointerDownObservable.add((evt) => {
             if (!this.isUp || evt.buttonIndex > 0) return;
             if (this._forceSelecting) {
-                this.isSelected(true, guiControl);
+                console.log(guiControl.name);
+                if (!this._doubleClick && this._isMainSelectionParent(guiControl)) {
+                    this._doubleClick = guiControl;
+                    window.setTimeout(() => {
+                        this._doubleClick = null;
+                    }, 1500);
```
RaananW, line 510: why are we waiting 1.5 seconds? is it configurable?

RaananW (commented):
A few nits, and this comment - 

Your code has no code doc at all. A few line-docs to explain what and why you are doing would have been nice to have. this help help you fix issues here later (instead of thinking "what am I doing here").

guiEditor/src/diagram/workbench.tsx
msDestiny14: This is a good point. Should this be up to the user?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -28,6 +28,8 @@ import { NodeMaterial } from "babylonjs/Materials/Node/nodeMaterial";
 import { TextureBlock } from "babylonjs/Materials/Node/Blocks/Dual/textureBlock";
 import { Observer } from "babylonjs/Misc/observable";
 import { GUIEditorNodeMaterial } from "./GUIEditorNodeMaterial";
+const DOUBLE_CLICK = 1500;
```
sebavan, line 31: Should this match inputManager DoubleClickDelay

guiEditor/src/diagram/workbench.tsx
```diff
@@ -90,6 +95,27 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         return this._selectedGuiNodes;
     }
 
+    // given a depth and a control gets the parent up the tree depth times
+    private _getParentWithDepth(control: Control, depth: number) {
+        let parent = control;
+        for (let i = 0; i < depth; ++i) {
+            if (parent.parent)
```
sebavan, line 102: Neat, the loop should break if no parent have been found

guiEditor/src/diagram/workbench.tsx
```diff
@@ -90,6 +95,27 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+                parent = parent.parent;
+        }
+        return parent;
+    }
+
+    //gets the higher parent of a given control.
+    private _getMaxParent(control: Control) {
+        let parent = control;
+        this._selectionDepth = 0;
```
sebavan, line 111: This is strange that _getMaxParent is changing selectionDepth when _getParentWithDepth does not ?

It looks like either both should and in this case I wonder if the should not also select and be renamed, or both should not ?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -214,6 +226,22 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         this.props.globalState.workbench = this;
     }
 
+    determineMouseSelection(selection: Nullable<Control>) {
+        if (selection && this._selectedGuiNodes.length <= 1) {
+            // if we're still on the same main selection, got down the tree.
+            if (selection === this._selectedGuiNodes[0] || selection === this._mainSelection) {
+                selection = this._getParentWithDepth(selection, --this._selectionDepth);
```
sebavan, line 233: `--this._selectionDepth` could the expression be placed on the line before ? 

This is a simple Neat preventing devs to think twice about does -- happens before or after :-)

sebavan: also it sounds like here selectionDepth will always end up -1 independently of the number of iteration in the getParentWithDepth unlinke the max one ?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -464,7 +503,17 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         const onPointerDown = guiControl.onPointerDownObservable.add((evt) => {
             if (!this.isUp || evt.buttonIndex > 0) return;
             if (this._forceSelecting) {
-                this.isSelected(true, guiControl);
+                // if this is our first click or the clicked control is a parent of the previous clicked control.
```
sebavan, line 506: is a parent should be is a child ?

sebavan: also you put or in the sentence but && in code and I can t see which one is right ?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -464,7 +503,17 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
-                this.isSelected(true, guiControl);
+                // if this is our first click or the clicked control is a parent of the previous clicked control.
+                if (!this._doubleClick && this._isMainSelectionParent(guiControl)) {
+                    this._doubleClick = guiControl;
+                    window.setTimeout(() => {
+                        this._doubleClick = null;
+                    }, DOUBLE_CLICK);
+                }
+                else { //select our new main control.
+                    this.determineMouseSelection(guiControl);
```
sebavan, line 514: should the second click control be checked against _doubleClick control to validate the action ? (wondering what happens if double click but not on the same position)

msDestiny14: If it is a different control, it will catch it in the next function. 
determineMouseSelection will either go down the tree or restart tree at new main control. Ill update the comment to reflect this,

guiEditor/src/diagram/workbench.tsx
msDestiny14: > 

Yes. Should should I move it inside the function?

guiEditor/src/diagram/workbench.tsx
RaananW: i guess the main question is why do you need the timeout at all?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -28,6 +28,9 @@ import { NodeMaterial } from "babylonjs/Materials/Node/nodeMaterial";
 import { TextureBlock } from "babylonjs/Materials/Node/Blocks/Dual/textureBlock";
 import { Observer } from "babylonjs/Misc/observable";
 import { GUIEditorNodeMaterial } from "./GUIEditorNodeMaterial";
+
+const DOUBLE_CLICK = Scene.DoubleClickDelay;
```
sebavan, line 32: Why a custom one here, maybe just directly always reuse the scene one so custo would apply accross the board ?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -90,6 +96,32 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         return this._selectedGuiNodes;
     }
 
+    // given a control gets the parent up the tree selectionDepth times. Selection depth is altered as we go down the tree.
+    private _getParentWithDepth(control: Control) {
+        --this._selectionDepth;
```
sebavan, line 101: should this be part of the loop ? as it looks like here it will always only remove 1 independently of the depth ?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -90,6 +96,32 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         return this._selectedGuiNodes;
     }
 
+    // given a control gets the parent up the tree selectionDepth times. Selection depth is altered as we go down the tree.
+    private _getParentWithDepth(control: Control) {
+        --this._selectionDepth;
+        let parent = control;
+        for (let i = 0; i < this._selectionDepth; ++i) {
+            if (parent.parent) {
```
sebavan, line 104:
```suggestion
            if (!parent.parent) {
              break;
            }
            parent = parent.parent;
```

guiEditor/src/diagram/workbench.tsx
msDestiny14: No. It should only remove once per functional call.

PR #11592, opened 2021-12-06
[GE] Adding descendentsOnlyPadding Checkbox

msDestiny14:
image


RaananW (approved):
A small note - the dist folder is updated during nightlies, so no need to commit it. It's here and updated so it's fine, but just for future reference :-)

PR #11599, opened 2021-12-07
Add all statistics to perf viewer

carolhmj:


sebavan (approved):

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerSidebarComponent.tsx
```diff
@@ -21,6 +24,18 @@ export const PerformanceViewerSidebarComponent = (props: IPerformanceViewerSideb
...
     // List of ordered categories
     const [metadataCategories, setMetadataCategories] = useState<string[]>();
+    // Min/Max/Current values of the ids
+    const [valueMap, setValueMap] = useState<Map<string, IPerfMinMax>>();
+
+    useEffect(() => {
+        const observer = (props: IVisibleRangeChangedObservableProps) => {
+            setValueMap(props.valueMap);
+        }
+        onVisibleRangeChangedObservable?.add(observer);
```
RaananW, line 34: nit - instead of using `?` (because it technically can be undefined, this is right), you can check at the beginning of the function if the `onVisibleRangeChangedObservable` is there, and return if it isn't.
This way you don't create an unneeded observer as well.

RaananW (commented):
LGTM! Just a small nit-picking :-)

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerSidebarComponent.tsx
carolhmj: Perfect, did that :D

PR #11609, opened 2021-12-08
DeviceSourceManager: Add mouse to list of already connected items

PolygonalSun:
This PR adds a line to automatically add the mouse to the DeviceSourceManager's list of items because the mouse is already connected in the DeviceInputSystem.

### DIFF
--- a/src/DeviceInput/deviceInputSystem.ts
+++ b/src/DeviceInput/deviceInputSystem.ts
@@ -58,12 +58,7 @@ export class DeviceInputSystem {
      */
     constructor(deviceInputSystem: IDeviceInputSystem) {
         this._deviceInputSystem = deviceInputSystem;
-        // Adds a callback that reassigns onDeviceConnected's callback to trigger the observable's observers for already connected devices
-        this.onDeviceConnectedObservable = new Observable((observer) => {
-            this._deviceInputSystem.onDeviceConnected = (deviceType, deviceSlot) => {
-                this.onDeviceConnectedObservable.notifyObservers({ deviceType, deviceSlot });
-            };
-        });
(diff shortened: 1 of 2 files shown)


src/DeviceInput/deviceInputSystem.ts
```diff
@@ -58,7 +58,12 @@ export class DeviceInputSystem {
      */
     constructor(deviceInputSystem: IDeviceInputSystem) {
         this._deviceInputSystem = deviceInputSystem;
-        this.onDeviceConnectedObservable = new Observable();
+        // Adds a callback that reassigns onDeviceConnected's callback to trigger the observable's observers for already connected devices
+        this.onDeviceConnectedObservable = new Observable((observer) => {
+            this._deviceInputSystem.onDeviceConnected = (deviceType, deviceSlot) => {
+                this.onDeviceConnectedObservable.notifyObservers({ deviceType, deviceSlot });
+            };
```
syntheticmagus, line 65: I think I don't have enough context for how this is set up to know whether/how this fixes the input system breakage. This looks like it's assigning a callback inside an observable's callback, which looks like it might be called every time a device is connected and consequently pave over the old callback every time. Is that what's supposed to happen?

PolygonalSun:
I just found a scenario in which this doesn't work so I'm closing this.

src/DeviceInput/deviceInputSystem.ts
PolygonalSun: Found an easier way to implement this fix

PolygonalSun:
Changed approach to fix

sebavan (approved):

PR #11611, opened 2021-12-09
[GE] Various Bug Fixes

msDestiny14:
-Fix with spacing empty value entered
-Fix for selection double clicking into children
-Can now drag and drop controls from the toolbar


guiEditor/src/workbenchEditor.tsx
```diff
@@ -10,6 +10,7 @@ import { MessageDialogComponent } from "./sharedComponents/messageDialog";
 import { SceneExplorerComponent } from "./components/sceneExplorer/sceneExplorerComponent";
 
 import { CommandBarComponent } from "./components/commandBarComponent";
+import { Nullable } from "babylonjs";
```
RaananW, line 13: nit - load from the actual file and not the main package.

RaananW (commented):
Please run a formatter to avoid linting issues, and use "!==" instead of "!="

PR #11612, opened 2021-12-10
Fix holographicSlate texture orientation when on right handed mode, and also improve on the other 3D button fixes

carolhmj:
Realized we can fix all 3D buttons just flipping the UVs on the parent class button so I thought this would look nicer than changing each individual one's applyFacade.

### DIFF
--- a/gui/src/3D/controls/holographicSlate.ts
+++ b/gui/src/3D/controls/holographicSlate.ts
@@ -267,9 +267,9 @@ export class HolographicSlate extends ContentDisplay3D {
         const node = new Mesh("slate" + this.name, scene);
 
         this._backPlate = CreateBox("backPlate" + this.name, { size: 1 }, scene);
-        const faceUV = new Array(6).fill(new Vector4(0,0,1,1));
+        const faceUV = new Array(6).fill(new Vector4(0, 0, 1, 1));
         if (scene.useRightHandedSystem) {
-            faceUV[0].copyFromFloats(0,1,1,0);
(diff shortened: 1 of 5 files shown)


gui/src/3D/controls/holographicSlate.ts
```diff
@@ -266,7 +267,11 @@ export class HolographicSlate extends ContentDisplay3D {
         const node = new Mesh("slate" + this.name, scene);
 
         this._backPlate = CreateBox("backPlate" + this.name, { size: 1 }, scene);
-        this._contentPlate = CreateBox("contentPlate" + this.name, { size: 1 }, scene);
+        const faceUV = new Array(6).fill(new Vector4(0,0,1,1));
+        if (scene.useRightHandedSystem) {
+            faceUV[0] = new Vector4(0,1,1,0);
```
RaananW, line 272: Nit - You can use .set or copyFromFloats instead of creating a new vector. Same things goes to the other vector assignment in the other file.

RaananW (approved):
LGTM! There is a linting issue and a small suggestion

Popov72 (approved):

gui/src/3D/controls/holographicSlate.ts
carolhmj: Done!

sebavan (approved):

sebavan:
@carolhmj looks like you have a tiny lint issue
image

PR #11615, opened 2021-12-10
Fixing lineMeshes + oit

CraigFeldspar:
Related to 
https://forum.babylonjs.com/t/transparencies-and-babylon-5-super-improvement/25536/6
Draft until tests are fixed

### DIFF
--- a/src/Meshes/linesMesh.ts
+++ b/src/Meshes/linesMesh.ts
@@ -112,7 +112,7 @@ export class LinesMesh extends Mesh {
         if (material) {
             this.material = material;
         } else {
-            this._lineMaterial = new ShaderMaterial("colorShader", this.getScene(), "color", options, false);
+            this.material = new ShaderMaterial("colorShader", this.getScene(), "color", options, false);
         }
     }
 
@@ -143,6 +143,7 @@ export class LinesMesh extends Mesh {
      */
     public set material(value: Material) {
         this._lineMaterial = value;
+        this._lineMaterial.fillMode = Material.LineListDrawMode;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Popov72 (approved):
LGTM!

deltakosh (approved):

sebavan (approved):

CraigFeldspar:
It doesn't come from this PR (got the same issue in master), but sometimes glowLayer and motionBlur tests are failing in local. When running the URL directly http://localhost:1338/tests/validation/index.html?Motion%20Blur#container#174, it always passes, but sometimes (not even always) in sequence using http://localhost:1338/tests/validation/index.html, they don't

Popov72:
Maybe it's some timing problems and we should raise the `renderCount` value of these tests?

Jean-MarieL:
**Many thanks for all your efforts**. I have tried the play ground with the current alpha (65 ?). It is definitely improved even if I still see some issues.
In the play Ground  @    https://playground.babylonjs.com/#M4DILD#4 
you will see **moiré effects**. Here I just have "zoomed in" in changing the focal distance:
PG Transp
Another issue is with our app, as soon I am displaying a system of lines the **display is completely perturbed**.
I am attaching below first the correct display, and then one with a system of lines turned to visible.
OK display
KO display
About the second issue I will  try to prepare a specific play ground !
Congrats to all of you, POUR LE **SUPER TRAVAIL** !!

sebavan:
Thanks a lot for the report, I am sure @CraigFeldspar will address those in not time :-)

Jean-MarieL:
HELLO Sebavan, Craig and others...
GOOD NEWS after an attempt to create a specific playground, where thinks want quite OK, everything turned to be fine, including in our APP !! (see below) THANKS A LOT AGAIN !!
TP_OK
It remains only the moiré effect appearing only in "extreme cases"

PR #11619, opened 2021-12-10
Keep performance recording values

carolhmj:


sebavan (approved):

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -478,8 +478,10 @@ export class PerformanceViewerCollector {
         if (!shouldPreserve) {
             this.datasets.data = new DynamicFloat32Array(initialArraySize);
             this.datasets.startingIndices = new DynamicFloat32Array(initialArraySize);
+            this._startingTimestamp = PrecisionDate.Now;
+        } else if (!this._startingTimestamp) {
```
RaananW, line 482: did you mean to check if it is undefined, or is any false-y value good here?

RaananW (approved):
LGTM! Just a small nit

src/Misc/PerformanceViewer/performanceViewerCollector.ts
carolhmj: Oh, it should be undefined, thanks for catching that!

PR #11624, opened 2021-12-12
Fix for ParticleSystem being disposed before _subEmitters is created

BlakeOne:
Fixes the issue where if a ParticleSystem is disposed before _subEmitters has been created & populated internally then the sub emitters are not disposed.

Here is the playground that demonstrates this issue: https://playground.babylonjs.com/#GXTHL9#50

Forum: https://forum.babylonjs.com/t/particle-system-dispose-leak-with-nested-subemitter/25952/2

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -251,6 +251,7 @@
 
 ## Bugs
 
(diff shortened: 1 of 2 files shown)


RaananW (approved):
Though it feels a bit off to create the array in order to dispose it, this is what we do throughout the file as well. LGTM!

Popov72 (approved):

PR #11627, opened 2021-12-13
Fix #11626

### LINKED ISSUE #11626: Save the handedness in the Babylon file
**Is your feature request related to a problem?**
Yes. Often the handedness is needed to properly display a scene after importing.

**Describe the solution you'd like**
Simply save the useRightHandedSystem flag when serializing a scene. Then set the  flag when loading a babylon file.

**Discussion**
https://forum.babylonjs.com/t/sandbox-with-rhs-babylon/26028

deltakosh:


sebavan (approved):

sebavan:
@deltakosh  I reran the test cause one seems failing.

andreasplesch:
I think it is this test (soft shadow right-handed) which is failing:
https://playground.babylonjs.com/#0YYQ3N#2

It may fail because the test appends a babylon file which expects to be used in a RHS but does not have the flag, and therefore the scene reverts to the default LHS, now.

Replacing scenes/SSAOcat.babylon with a new version with the flag may fix the test.

A way to keep backward compatibility may be to check if scene.useRightHandedSystem already had been set to true, and then not change it if serializationObject.useRightHandedSystem is not present (null or undefined?) ? Or actually do nothing if not present :

```js
if ( parsedData.useRightHandedSystem !== undefined && parsedData.useRightHandedSystem !== null ) {
    scene.useRightHandedSystem = !!parsedData.useRightHandedSystem;
}
```

Popov72 (approved):

PR #11634, opened 2021-12-14
Perf recording fixes

carolhmj:
Fix some issues when loading the performance profiler from a CSV then opening a live performance profiler. As part of this, I hid the buttons to open the profiler while the popup is open, so the user has to close it before opening another.

### DIFF
--- a/inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
+++ b/inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
@@ -70,24 +70,24 @@ export const PerformanceViewerComponent: React.FC<IPerformanceViewerComponentPro
         if (window) {
             window.close();
         }
-        if (isLoadedFromCsv) {
-            setIsLoadedFromCsv(false);
-            // If we load or unload a window from CSV data, we should clear the performanceCollector
-            // so that it can properly capture realtime performance
-            if (performanceCollector) {
-                performanceCollector.clear();
-                addStrategies(performanceCollector);
-            }
-        }
+        // if (isLoadedFromCsv) {
+        setIsLoadedFromCsv(false);
+        // }
         setIsOpen(false);
     };
 
+    useEffect(() => {
+        if (!isLoadedFromCsv) {
+            if (performanceCollector) {
+                performanceCollector.stop();
+                performanceCollector.clear(false, false);
+                addStrategies(performanceCollector);
+            }
+        }
+    }, [isLoadedFromCsv]);
+
     const startPerformanceViewerPopup = () => {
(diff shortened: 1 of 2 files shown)


inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
```diff
@@ -76,10 +70,20 @@ export const PerformanceViewerComponent: React.FC<IPerformanceViewerComponentPro
         if (window) {
             window.close();
         }
+        setIsLoadedFromCsv(false);
         setIsOpen(false);
-        setIsLoaded(false);
     };
 
+    useEffect(() => {
+        if (!isLoadedFromCsv) {
```
RaananW, line 78: what is the flow when setting this to true?
This seems to be a cleanup when setting it to false (the default behavior), but shouldn't be a cleanup when setting it to true?

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
```diff
@@ -129,52 +133,36 @@ export const PerformanceViewerComponent: React.FC<IPerformanceViewerComponentPro
...
-            performanceCollector?.start(true); //preserve data
-        } else {
-            performanceCollector?.stop();
-        }
-
-        return () => {
-            performanceCollector?.stop();
-        };
-    }, [recordingState]);
+        const perfCollector = scene.getPerfCollector();
```
RaananW, line 153: any cleanup needed here when the component is disposed?

RaananW (commented):
Great!

My main questions are regarding the setEffect calls and component cleanup when they are disposed or the state changes.

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
carolhmj: I wanted this useEffect to run only on the case where isLoadedToCsv's value changed from true to false (so, closing a window that was opened from a CSV file). When this happens, the performance collector should clear its data so it's ready to open either another CSV or a live recording.

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
carolhmj: I didn't want to call any stop or clear cleanups because when we open the Popup window and then change tabs, this component is disposed, and this would stop the data collection every time we changed tabs. If the user wants to stop the collector, they could go back to the Statistics tab, remount the component (which should get the same performance collector of the scene), and stop using the button.

PR #11637, opened 2021-12-14
Serialize and load animation interpolation values

carolhmj:


sebavan:
@carolhmj quite a lot of tests are failing :-) but your PR sounds rather good, I wonder if we did not parse values before that we are now accounting for :-)

carolhmj:
Nah that's on me! I wrote `.value` instead of `.values` in some places! I'm looking for those and fixing them now!

PR #11638, opened 2021-12-14
[XR] add a new function to layers feature

RaananW:
There was no way of tracking existing layers and adding a new one.


sebavan (approved):

src/XR/features/WebXRLayers.ts
```diff
@@ -242,6 +257,15 @@ export class WebXRLayers extends WebXRAbstractFeature {
         return new WebXRProjectionLayerWrapper(projLayer, this._xrWebGLBinding);
     }
 
+    /**
+     * Add a new layer to the already-existing list of layers
+     * @param wrappedLayer the new layer to add to the existing ones
+     */
+    public addXRSessionLayer(wrappedLayer: WebXRLayerWrapper) {
```
rgerd, line 264: Should there also be a `getXRSessionLayers`?

src/XR/features/WebXRLayers.ts
```diff
@@ -242,6 +257,15 @@ export class WebXRLayers extends WebXRAbstractFeature {
...
+        this._existingLayers.push(wrappedLayer);
+        this.setXRSessionLayers(this._existingLayers);
```
rgerd, line 266: Should `setXRSessionLayers` set `this._existingLayers`? 
This doesn't seem like a function that will be called very often so it feels like you could do something like `this.setXRSessionLayers([...this._existingLayers, wrappedLayer])` and then in `setXRSessionLayers`, `this._existingLayers = wrappedLayers.slice(0)`

src/XR/features/WebXRLayers.ts
```diff
@@ -118,6 +119,7 @@ class WebXRCompositionLayerRenderTargetTextureProvider extends WebXRLayerRenderT
  * @hidden
  */
 export class WebXRProjectionLayerWrapper extends WebXRCompositionLayerWrapper {
+    public babylonLayer: Nullable<Layer>;
```
rgerd, line 122: I didn't even realize we had a Layer type! 
Was this an intentional change? I don't see it being used elsewhere...

RaananW: oh! this is a preparation for a further change I was planning, good catch. I will remove it from here for now.

src/XR/features/WebXRLayers.ts
RaananW: I thought about it. At the beginning I exposed the existing layers, and then I thought the flow is - create a wrapper (and have a reference to it) and then add it to the session. I already have the wrapper (and am using it somewhere else), I don't know when I will need the entire array. Apart from when I want to remove the "base" layer, but that would stop rendering. What do you think?

src/XR/features/WebXRLayers.ts
RaananW:
```suggestion
```

src/XR/features/WebXRLayers.ts
RaananW: (BTW - I resolved the other conversation. I keep this one open for further discussion on this subject)

PR #11655, opened 2021-12-15
Add nativeXRFrame

rgerd:
This introduces NativeXRPlugin, which extends the native implementation to provide JS code leveraging more performant types for transferring data across the JS-native boundary (such as ArrayBuffers). I've also added the NativeXRFrame type, which will host JS wrappers around most of the functions required by XRFrame, but currently only wraps `getPoseData`, rendering `getPose` a much more performant function.

As an illustrative example, [`WebXRAnchorSystem._updateAnchorWithXRFrame`](https://github.com/BabylonJS/Babylon.js/blob/59588ebeb6dc45c3f2d25c206df6bf850124685c/src/XR/features/WebXRAnchorSystem.ts#L361-L379) was made 2.5x faster with this change, making the entire `WebXRAnchorSystem._onXRFrame` function almost 3x faster.

See https://github.com/BabylonJS/BabylonNative/pull/961 for context.


src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,98 @@
+import { NativeXRPlugin } from "./nativeXRPlugin";
+
+/** @hidden */
+interface INativeXRFrame extends XRFrame {
+    // Native supported functions
+    createAnchor(pose: XRRigidTransform, space: XRSpace): Promise<XRAnchor>;
+    fillPoses(spaces: XRSpace[], baseSpace: XRSpace, transforms: Float32Array): boolean;
+    getJointPose(joint: XRJointSpace, baseSpace: XRSpace): XRJointPose;
+    fillJointRadii(jointSpaces: XRJointSpace[], radii: Float32Array): boolean;
```
ryantrem, line 9: These are already part of the `XRFrame` interface right? I'm confused why they seem to be repeated here.

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,98 @@
...
+    getJointPose(joint: XRJointSpace, baseSpace: XRSpace): XRJointPose;
+    fillJointRadii(jointSpaces: XRJointSpace[], radii: Float32Array): boolean;
+
+    // Native-only helper functions
+    getPoseData: (space: XRSpace, baseSpace: XRReferenceSpace, vectorBuffer: ArrayBuffer, matrixBuffer: ArrayBuffer) => XRPose;
+}
+
+/** @hidden */
+export class NativeXRFrame implements XRFrame {
+    private _xrTransform = new XRRigidTransform();
```
ryantrem, line 17:
```suggestion
    private readonly _xrTransform = new XRRigidTransform();

```

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,98 @@
...
+    private _xrPose: XRPose = {
```
ryantrem, line 18:
```suggestion
    private readonly _xrPose: XRPose = {

```

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,98 @@
...
+        transform: this._xrTransform,
+        emulatedPosition: false
+    };
+    // Enough space for position, orientation
+    private _xrPoseVectorData = new Float32Array(4 * (4 + 4));
```
ryantrem, line 23:
```suggestion
    private readonly _xrPoseVectorData = new Float32Array(4 * (4 + 4));

```

src/XR/native/nativeXRFrame.ts
rgerd: This implies that they're implemented as opposed to optionally implemented (in webxr.d.ts they're listed with a `?`). I was just trying to avoid using `!` when calling the functions through `this._native`, but if it seems redundant I can take it out.

src/XR/webXRSessionManager.ts
```diff
@@ -190,6 +191,11 @@ export class WebXRSessionManager implements IDisposable, IWebXRRenderTargetTextu
         if (!this._xrNavigator.xr) {
             return Promise.reject("WebXR not available");
         }
+
+        if (this.isNative) {
+            NativeXRPlugin.ExtendNativeObject();
+        }
```
ryantrem, line 197: So is the pattern here that code can call `RegisterType` early (before `_native` exists in the global scope), and then by the time this `ExtendNativeObject` function is called `_native` will exist? If so, I wonder if instead it might make sense to more generically have JS side initialization code that either executes synchronously if `_native` already exists, and if not have it use `Object.defineProperty` in the global scope to define a `_native` property such that when the native side sets that global property, we could have a callback (as part of the `Object.defineProperty`) that could do the deferred initialization. It seems like that would be a generic enough mechanism to support most if not all of our JS side initialization upon `_native` being defined. Thoughts?

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    public get session(): XRSession {
+        return this._nativeImpl.session;
+    }
+
+    constructor(private _nativeImpl: INativeXRFrame) { }
+
+    public getPose(space: XRSpace, baseSpace: XRReferenceSpace): XRPose | undefined {
+        if (!this._nativeImpl.getPoseData(space, baseSpace, this._xrPoseVectorData.buffer, this._xrTransform.matrix.buffer)) {
```
ryantrem, line 32: Do these args need to be passed every time? Aren't they usually the same (especially `_xrPoseVectorData.buffer` and `_xrTransform.matrix.buffer`) every frame?

rgerd: This is a good point, although receiving the parameters in native doesn't seem to be significantly performance-heavy, especially when compared to pulling out the `Napi::External<xr::Space>` on the native side, so until we get that part much faster we won't see a difference if we ditch the extra parameters.

src/XR/webXRSessionManager.ts
rgerd: I changed the functions a bit to set the properties on _native directly if _native has already been defined, otherwise they're saved for later.
The `Object.defineProperty` solution is interesting, I just hesitate to define `_native` behind a setter/getter for performance reasons, and it feels like it warrants discussion as to whether we create an observable so that we could define multiple behaviors on definition of _native. If it's a blocker I can think about it some more

src/XR/native/nativeXRFrame.ts
ryantrem: Ah ok, got it.

sebavan (commented):
LGTM but I know nothing about XR or Native XR... so I won t be helpful.

As @RaananW is off I ll let @ryantrem and @bghgary validate the PR, let me know when you want me to merge it :-)

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/** @hidden */
+export class NativeXRFrame implements XRFrame {
+    private readonly _xrTransform = new XRRigidTransform();
+    private readonly _xrPose: XRPose = {
+        transform: this._xrTransform,
+        emulatedPosition: false
+    };
+    // Enough space for position, orientation
+    private readonly _xrPoseVectorData = new Float32Array(4 * (4 + 4));
```
bghgary, line 17: this should just be 8 right? this is the number of elements and not byteLength?

```suggestion
    private readonly _xrPoseVectorData = new Float32Array(4 + 4);
```

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,92 @@
...
+        const orientation = this._xrTransform.orientation as DOMPoint;
+        orientation.x = this._xrPoseVectorData[4];
+        orientation.y = this._xrPoseVectorData[5];
+        orientation.z = this._xrPoseVectorData[6];
+        orientation.w = this._xrPoseVectorData[7];
+        return this._xrPose;
+    }
+
+    public fillPoses(spaces: XRSpace[], baseSpace: XRSpace, transforms: Float32Array): boolean {
+        return this._nativeImpl.fillPoses!(spaces, baseSpace, transforms);
```
bghgary, line 44: Why can functions in XRFrame be undefined? 🤔

src/XR/native/nativeXRFrame.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public getViewerPose(referenceSpace: XRReferenceSpace): XRViewerPose | undefined {
+        return this._nativeImpl.getViewerPose(referenceSpace);
+    }
```
bghgary, line 49: I don't know if you intend on wrapping all of these functions, but if you just want to pass through, I believe you can just assign the function instead of forwarding. Saves one function call.

```suggestion
    public readonly getViewerPose = this._nativeImpl.getViewerPose;
```

bghgary (commented):
Minor comments.

src/XR/native/nativeXRFrame.ts
rgerd: It's the spec's way of conveying optional support.  
I originally extended the interface to mark the functions as supported (not optional), but it seemed confusing to have the functions redefined with the one character missing:
https://github.com/BabylonJS/Babylon.js/pull/11655/commits/d93e90ddc786db0b5548dd2d80644e855f7dc4b8#diff-ab4efb00fe84c8b2b0443bfbe2b09f90255034e04359051f5fb41e1432f5060fR6-R9
I personally don't have much of a preference either way.

bghgary (approved):

PR #11656, opened 2021-12-15
Fix issue when AssetContainer is added to Scene multiple times

BlakeOne:
Add check for _wasAddedToScene to method addAllToScene and return early if AssetContainer has already been added.

This fixes an issue which can be reproduced by opening this simple playground - which calls addAllToScene twice - and then pressing the run button, which causes the page to become unresponsive.

Playground: https://playground.babylonjs.com/#99WCUQ#12
Forum: https://forum.babylonjs.com/t/calling-assetcontainer-addalltoscene-multiple-times-freezes-page/26100

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -252,6 +252,7 @@
 
 ## Bugs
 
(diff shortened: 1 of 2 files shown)


BlakeOne:
Maybe a warning should be logged to the console too? Something like "Warning: addAllToScene should only be called once per AssetContainer"?

PR #11657, opened 2021-12-16
Fix playground when using older bjs versions

bghgary:
Using the playground against 4.2 or older is broken because of this. Needs testing before merge.


Popov72 (approved):
LGTM.

I don't know how to test locally as we can't change the version of the Playground but that seems ok to me.

sebavan (approved):

PR #11670, opened 2021-12-18
Added parameters closeShape and closePath to ExtrudeShape and ExtrudeShapeCustom

BabylonJSGuide:
Ref: https://forum.babylonjs.com/t/normal-discontinuity-in-extrusions/26091


src/Meshes/Builders/shapeBuilder.ts
```diff
@@ -27,7 +29,7 @@ import { Path3D } from "../../Maths/math.path";
  */
 export function ExtrudeShape(
     name: string,
-    options: { shape: Vector3[]; path: Vector3[]; scale?: number; rotation?: number; cap?: number; updatable?: boolean; sideOrientation?: number; frontUVs?: Vector4; backUVs?: Vector4; instance?: Mesh; invertUV?: boolean },
+    options: { shape: Vector3[]; path: Vector3[]; scale?: number; rotation?: number; closeShape?: boolean; closePath?: boolean; cap?: number; updatable?: boolean; sideOrientation?: number; frontUVs?: Vector4; backUVs?: Vector4; instance?: Mesh; invertUV?: boolean },
```
sebavan, line 32: This is unfortunately a breaking change, could both params be added at the end of the function ?

src/Meshes/Builders/shapeBuilder.ts
```diff
@@ -78,6 +84,8 @@ export function ExtrudeShapeCustom(
         rotationFunction?: Nullable<{ (i: number, distance: number): number }>;
         ribbonCloseArray?: boolean;
         ribbonClosePath?: boolean;
+        closeShape?: boolean;
```
sebavan, line 87: Same comment for this one ?

src/Meshes/Builders/shapeBuilder.ts
```diff
@@ -100,8 +108,8 @@ export function ExtrudeShapeCustom(
         (() => {
             return 0;
         });
-    const ribbonCloseArray = options.ribbonCloseArray || false;
-    const ribbonClosePath = options.ribbonClosePath || false;
+    const ribbonCloseArray = options.closePath || options.ribbonCloseArray || false;
```
sebavan, line 111: Why adding an extra param if it does identical than the existing one ?

sebavan (changes requested):

src/Meshes/Builders/shapeBuilder.ts
BabylonJSGuide: This change does not break existing code they only come into action when specifically set, defaults match existing code. Order does not matter as they are properties of the options parameter object not a parameter of the function.

src/Meshes/Builders/shapeBuilder.ts
BabylonJSGuide: As above

Popov72 (approved):

PR #11676, opened 2021-12-20
Updated Texture Compression batch script to use new PVRTC settings for PVRTexToolCLI.

Leolele99:
I noticed that the command line too for texture compression using PVRTexTool still works great for everything but PVRTC.

After their own manuals fell short, I asked in their forums and updated the values according to their recommendations.

The new documentation which supports this change will be available soon according to them.

Source: https://forums.imgtec.com/t/pvrtextoolcli-pvrtc-encoding-invalid-valid-values/3483


sebavan:
Thanks a lot !!!

PR #11684, opened 2021-12-21
Goldberg mesh update

sebavan:
Follow up of https://forum.babylonjs.com/t/trouble-accessing-methods-in-npm-typescript-project-goldberg-mesh/26209/6


sebavan:
@BabylonJSGuide , it would be amazing to have your feedback ? obviously i will update the doc and samples accordingly ?

This is to allow a better usage in typescript.

BabylonJSGuide:
@sebavan I really dislike using Typescript since I stuggle with its restrictions although I do see its worth for projects like Babylon.js. This being the case I only ever test the result of my changes with Javascript never Typescript, hence the issue that arose. I see the need for the change to 

```
export class GoldbergMesh extends Mesh ....
``` 
from using a mixin.

I will need some time to check the changes through properly, for example, I noticed that 

```
goldberg.refreshFaceData();
```

has been removed and I will need to find how you have replaced it. I will do it asap but with the holidays it might be several days.

BabylonJSGuide:
The reason for refreshFaceData is because of serialization. Since using the serializer for on a Goldberg poly only serializes as mesh and imports it as a mesh extra saved properties have to be restored from the metadata.  From what I now understand importing a Goldberg mesh will mean that the extra methods such as setFaceColors will not work in Typescript as the import will be of type class.

sebavan:
Good catch let me make it a real mesh before you try the new version.

sebavan:
Actually I automated the goldbergData to be stored during serialization and calling into ExtendMeshToGoldberg after deserialization as with the code before is all good.

I let you test the current code and let me know, thanks a lot @BabylonJSGuide !!! I ll update the doc + samples once all good from your point of view.

BabylonJSGuide:
OK but I do not understand if it is a just a mesh how the extension properties and methods that are just specific to  Goldberg Polyhedron fit?

Is there a way when importing a mesh to cast (if that is the  correct word) it to a Goldberg mesh if it is such a mesh?

sebavan:
Yup, exactly like before the PR you need to call into ExtendMeshToGoldberg(mesh) which I think was by design ? I can revamp this to make it a serialized and parsed type like other meshes and such if you prefer ?

sebavan:
Or we make all the extension functions not instances functions but helper functions only ? this is totally up to you. At the moment I did not touch the API (only renaming)

BabylonJSGuide:
Or we make all the extension functions not instances functions but helper functions only ?

Do not understand what this means. As I said I struggle with Typescript, this is because of some of the concepts.

If what you have now achieved is that 
```
goldberg.setFaceColors([0, 10, new BABYLON.Color4(1, 0, 0, 1)]);
```
and other functions and properties work in Typescript and you can export and then import it with all data intact and the functions and properties working in both Javascript and Typescript I will be happy. 

I will have a look tomorrow at your current changes and start testing.

I think to make it work

BabylonJSGuide:
@sebavan I have tested the PR using update code on the PGs from the [Goldberg docs page](https://doc.babylonjs.com/divingDeeper/mesh/creation/polyhedra/goldberg_poly) . Except for the last one they all work in JavaScript and given my limited knowledge probably in Typescript. To test in Typescript I created a blank file, did `Import {MeshBuilder} .... and pasted my test files in turn and all properties and functions were accepted.

The problem is with the last PG , https://playground.babylonjs.com/#A8VZGP#16, 'Import as a Goldberg Mesh' this does not even work in 5.0.0 alpha 65 which is after the MeshBuilder codes were updated. In my original PR commit e00959d136c02cca395eab99a5c6a11a9e8e6ceb Sept 22  `extendToGoldberg` was as a static method on a Mesh.  It is now the function `ExportMeshtoGoldberg` within GoldbergBuilder  however any of

BABYLON.Mesh.ExtendMeshToGoldberg(mesh);
BABYLON.MeshBuilder.ExtendMeshToGoldberg(mesh);
BABYLON.GoldbergBuilder.ExtendMeshToGoldberg(mesh);
BABYLON.Builders.ExtendMeshToGoldberg(mesh);

How do you use it?

sebavan:
It would be BABYLON.ExtendMeshToGoldberg directly on the namespace as it is a global export like before the PR I guess

BabylonJSGuide:
Unfortunately this still gives console errors in PG #A8VZGP#16,

BabylonJSGuide:
The error in version 5.0.0 alpha 65 seems to be solved by adding `ExtendMeshToGoldberg` in meshBuilder.ts to the list of imports and also to the list of static files of the MeshBuilder class and then using `BABYLON.MeshBuilder.ExtendMeshToGoldberg(mesh)`

In this case perhaps renaming `ExtendMeshToGoldberg` to something like `RecreateGoldbergMeshAfterImport` might make more sense.

When I get time I will export a Goldberg mesh and test its import using your PR plus the addition to meshBuilder.ts as above

BabylonJSGuide:
Have now done an export and import test after applying the changes suggested above and it works in Javascript. Only able to do very basic tests in Typescript but code is accepted in VSCode.

If you have better suggestions for dealing with imports of a GoldbergMesh they would be welcome.

Will test again after you have made updates to your PR.

When it comes to the docs I should have example playgrounds to update the existing ones.

Hope you are having a good holiday

sebavan:
Ok @BabylonJSGuide I did a new update making goldberg Mesh like the groundMesh or line Mesh, so you can simply serialize and parse it :-), no more need for extension.

Let me know what you think

BabylonJSGuide:
Thanks for all the work. Still having problems with importing. Build babylon.max.js using your latest PR.

My tests are to be found at https://github.com/BabylonJSGuide/Goldberg and can be run using https://relaxed-varahamihira-799310.netlify.app/gb1.html etc.

gbScene.babylon and gbMesh.babylon where exported using your latest PR and gbexportScene.html and gbexportMesh  and imported via importFromScene and importFromMesh respectively. Using Netlify you can see the errors.

I tried a file exported using you previous PR and a file importFromSceneOldFile using your current PR, in this case it loads but there is an error with goldbergData.

It may well be that I am doing something wrong with the export or import stage. Will have another look tomorrow.

Minor point in src/Meshes/mesh.ts line 3704 it should be `throw _WarnImport("GoldbergMesh");`

sebavan:
perfect let me have a look :-)

sebavan:
@BabylonJSGuide it should be all good now, I tried all your test pages and looked ok. Would be great if you could validate ?

BabylonJSGuide:
@sebavan brilliant, I am happy with the update. Would you like me to update the docs?

sebavan:
@BabylonJSGuide this would be a tremendous help as I am way more fluent in code than doc :-) I ll merge the code today for the nightly, thanks a lot.

RaananW (approved):

PR #11693, opened 2021-12-27
Add WordWrapEllipsis mode to textblock text wrapping.

mysunnytime:
Right now, in the wrapping mode of Ellipsis, the block is only going to keep one line and add ellipsis in the end.
image

If we choose WordWrap mode, then it shows all text even it overflows.
image

This PR added a WordWrapEllipsis mode so that the text can wrap and fill the textblock, followed by ellipsis if the text is too long.
image

Also tested in different text alighments
image

Validation tests all passed


sebavan:
Not sure it should be added, I can not remember a previous discussion with @deltakosh about this one, let s see what he thinks about it ?

deltakosh (approved):
All good to me! thanks a lot

deltakosh:
@mysunnytime please jsut fix the conflict and we can merge :)

PR #11697, opened 2021-12-29
fix freecamera touch input rotation sensibility and calculation error

m1911star:
- since *Sensibility is used as a divisor, so it should be `lower & faster`
-  when it's not a rotateCamera, it should not change `cameraRotation.y`


m1911star:
Just wonder why I made so many diffs. I just add one line to md file

`Fix FreeCameraTouchInput roation when moving ([m1911star](https://github.com/m1911star))`

sebavan:
Thanks

PolygonalSun:
The code from this PR had to be reverted due to an issue with the camera no longer rotating when the touch point was at the edge of the screen.

sebavan:
@m1911star you can find more info in this thread https://forum.babylonjs.com/t/change-in-universalcamera-touch-rotate-behavior-starting-with-beta-2/26681/13 if you are willing to resubmit a PR ? i guess a flag might be needed to not break back compat here.

m1911star:
@PolygonalSun @sebavan  Sorry for that, I will check later

PR #11704, opened 2022-01-03
Add function Animation.ParseFromACEFileAsync

BlakeOne:
PR to add the new function ParseFromACEFileAsync to the Animation class to let users load files that were saved from ACE.
Forum: https://forum.babylonjs.com/t/animation-from-json-loading/26421/8

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -124,6 +124,7 @@
 
 ### Inspector
 
+- Added static function `ParseFromACEFileAsync` to class Animation to handle files saved from ACE ([BlakeOne](https://github.com/BlakeOne))
 - Increased float precision to 4 ([msDestiny14](https://github.com/msDestiny14))
 - Added support for sounds in the inspector ([Deltakosh](https://github.com/deltakosh))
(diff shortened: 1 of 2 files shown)


sebavan (approved):
GG !!!

sebavan (changes requested):
Wait, I spoke too soon :-) Is there a way you can factorize with the code from ParseFromFileAsync as it is awfully similar but the .animations ???

deltakosh (changes requested):
Same comment as seb :)

sebavan:
maybe after parsing just check if animations exists and is an array, so we only have one function?

BlakeOne:
Just finished doing it by passing an optional parameter to the existing function and then saw your comment. I think would be better to check and handle it automatically like you suggested instead of using param, I'll make another commit for it that way. :)

BlakeOne:
Okay got it working with auto-detection of ACE-saved files by just checking for the animations array. Haha much simpler fix this way. :)

sebavan (approved):

deltakosh (approved):

PR #11705, opened 2022-01-03
`Geometry` should not use an hardcoded stride for the Position buffer.

barroij:
The `Geometry` class allows the user to specify a `stride` in `Geometry.setVerticesData`.

Still it is not possible to specify the stride when calliing `Geometry.setVerticesBuffer`, `Geometry.applyToMesh` or the setter `Geometry.boundingBias`, which leads to a call to 
`this._extend = extractMinAndMax(data, 0, this._totalVertices, this.boundingBias, 3);` with an hardcoded stride of 3.

This PR adds an optional `stride` parameter to both of these functions and a new setter-like function so that the extend could be properly computed if the stride is not 3.

### DIFF
--- a/src/Meshes/geometry.ts
+++ b/src/Meshes/geometry.ts
@@ -91,14 +91,24 @@ export class Geometry implements IGetSetVerticesData {
      *  Gets or sets the Bias Vector to apply on the bounding elements (box/sphere), the max extend is computed as v += v * bias.x + bias.y, the min is computed as v -= v * bias.x + bias.y
      */
     public set boundingBias(value: Vector2) {
-        if (this._boundingBias) {
-            this._boundingBias.copyFrom(value);
-        } else {
-            this._boundingBias = value.clone();
-        }
-
-        this._updateBoundingInfo(true, null);
+        this.setBoundingBias(value)
     }
+
+    /**
+     *  Sets the Bias Vector to apply on the bounding elements (box/sphere), the max extend is computed as v += v * bias.x + bias.y, the min is computed as v -= v * bias.x + bias.y
+     *  This function comes in addition to the `Geometry.boundingBias`setter in order to be able to specify a stride different from the default value
+     * @param value defines the new value of the boundingBias
+     * @param stride defines the stride size to use in the position buffer i.e. the distance between two positions in the positions array. (Default is 3)
+     */
+    public setBoundingBias(value: Vector2, stride?: number): void {
+       if (this._boundingBias) {
+           this._boundingBias.copyFrom(value);
+       } else {
+           this._boundingBias = value.clone();
+       }
+
+       this._updateBoundingInfo(true, null, stride);
+   }
 
     /**
      * Static function used to attach a new empty geometry to a mesh
@@ -239,7 +249,7 @@ export class Geometry implements IGetSetVerticesData {
      * @param kind defines the data kind (Position, normal, etc...)
      * @param data defines the vertex data to use
      * @param updatable defines if the vertex must be flagged as updatable (false as default)
-     * @param stride defines the stride to use (0 by default). This value is deduced from the kind value if not specified
+     * @param stride defines the stride size to use in the position buffer i.e. the distance between two positions in the positions array. (Default is 3)
      */
     public setVerticesData(kind: string, data: FloatArray, updatable: boolean = false, stride?: number): void {
         if (updatable && Array.isArray(data)) {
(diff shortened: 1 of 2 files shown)


barroij:
Additionnal notes :
- the stride is still 3 by default in `extractMinAndMax`
- The default stride is still used when calling `Geometry._applyToMesh` wihtin `Geometry._queueLoad`. It is not really possible as long as the stride is not store in the data.

barroij:
Actually closing it for now, it seems there are other places where the stride is hardcoded to 3 in the Geometry.

src/Meshes/geometry.ts
```diff
@@ -239,7 +249,7 @@ export class Geometry implements IGetSetVerticesData {
      * @param kind defines the data kind (Position, normal, etc...)
      * @param data defines the vertex data to use
      * @param updatable defines if the vertex must be flagged as updatable (false as default)
-     * @param stride defines the stride to use (0 by default). This value is deduced from the kind value if not specified
+     * @param stride defines the stride size to use in the position buffer i.e. the distance between two positions in the positions array. (Default is 3)
```
deltakosh, line 252: This is a breaking change..It should remain autodetect

barroij: You're right and the new comment is wrong, copy/paste is evil...

barroij: I did not change anything in this function, so I'll let the jsDoc unchanged

deltakosh: yep no worries :D

barroij: change reverted. And happy new year folks :)

deltakosh: Happy new year!!!

deltakosh (approved):

barroij:
@deltakosh there is some lint error, I'll fix that asap.

src/Meshes/geometry.ts
```diff
@@ -270,8 +280,9 @@ export class Geometry implements IGetSetVerticesData {
      * @param buffer defines the vertex buffer to use
      * @param totalVertices defines the total number of vertices for position kind (could be null)
      * @param disposeExistingBuffer disposes the existing buffer, if any (default: true)
+     * @param stride defines the stride size to use in the position buffer i.e. the distance between two positions in the positions array. (Default is 3)
      */
-    public setVerticesBuffer(buffer: VertexBuffer, totalVertices: Nullable<number> = null, disposeExistingBuffer = true): void {
+    public setVerticesBuffer(buffer: VertexBuffer, totalVertices: Nullable<number> = null, disposeExistingBuffer = true, stride?: number): void {
```
Popov72, line 285: You should not pass the `stride` value to this function: it is a data you can retrieve from `buffer`.`

In `VertexBuffer` you have a `byteStride` property. As it seems you need it as a number of components and not as a number of bytes, you can do `buffer.byteStride / VertexBuffer.GetTypeByteLength(buffer.type)`.

src/Meshes/geometry.ts
```diff
@@ -295,7 +306,7 @@ export class Geometry implements IGetSetVerticesData {
                 }
             }
 
-            this._updateExtend(data);
+            this._updateExtend(data, stride);
```
Popov72, line 309: See comment above, compute this value from `buffer.byteStride`.

src/Meshes/geometry.ts
```diff
@@ -704,7 +716,7 @@ export class Geometry implements IGetSetVerticesData {
 
             if (kind === VertexBuffer.PositionKind) {
                 if (!this._extend) {
-                    this._updateExtend();
+                    this._updateExtend(null, stride);
```
Popov72, line 719: Same comment than above, the `stride` value can be computed from `this._vertexBuffers[kind].byteStride`.

barroij:
@deltakosh @Popov72 Actually I think I'm gonna close it as I was a bit confused with my specific use case and what BabylonJs really does.
On my side I have some LineMesh2d and for them I have vertex buffer with a stride of 2 ([x, y, x, y, x, y, ...]. So it contains 2d points.
But BabylonJs is a 3d Engine and whathever the stride is, it assumes it will find 3d points in the vertex buffer to compute the extend and so on.
So I have no choice than having custom code to replace everything in BabylonJs that uses the vertex buffer as an input for 3d position.


Sorry for your time guys !

deltakosh:
No worries!

PR #11706, opened 2022-01-04
Support Custom Mesh Colliders

MackeyK24:
Add support for custom mesh colliders. (Different from custom shapes)

Add onCreateCustomMeshCollider handler to support creating custom mesh collider.

Example: Can use the hook to add support for Ammo.btSmoothTriangleMesh instead of Ammo.btTriangleMesh to help smooth out mesh collider contacts where the impostor type has to be a **MeshImposter** and not a **CustomImpostor**. Especially for high speed driving simulations


MackeyK24:
Example usage of both **onCreateCustomShape** and onCreateCustomMeshCollider** from my **babylon.manager.js** extension that i use for custom compound convex shapes and **Ammo.btSmoothTriangleMesh**

Screen Shot 2022-01-04 at 2 24 03 AM

sebavan (approved):

PR #11707, opened 2022-01-04
Fix camera collisions for RHS scenes with GLTF models

### LINKED ISSUE #11688: Incorrect camera collisions when scene is right handed
Forum discussion: https://forum.babylonjs.com/t/problem-when-checkcollisions-and-userighthandedsystem-are-used-together/26248

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com): https://www.babylonjs-playground.com/#I6J7UZ#47
- Expected result: Player should be able to walk on the ground
- Current result: Player clips through the mesh

**Additional context**
Doesn't seem to happen when using a plane ground or even a ground created from a height map: https://playground.babylonjs.com/#LQ4LI1#131

carolhmj:
Fix #11688

With the fix, collisions happen normally on the example playground: https://www.babylonjs-playground.com/#I6J7UZ#47

### DIFF
--- a/src/Collisions/collider.ts
+++ b/src/Collisions/collider.ts
@@ -394,14 +394,18 @@ export class Collider {
     }
 
     /** @hidden */
-    public _collide(trianglePlaneArray: Array<Plane>, pts: Vector3[], indices: IndicesArray, indexStart: number, indexEnd: number, decal: number, hasMaterial: boolean, hostMesh: AbstractMesh): void {
+    public _collide(trianglePlaneArray: Array<Plane>, pts: Vector3[], indices: IndicesArray, indexStart: number, indexEnd: number, decal: number, hasMaterial: boolean, hostMesh: AbstractMesh, invertTriangles?: boolean): void {
         if (!indices || indices.length === 0) {
             for (let i = 0; i < pts.length; i += 3) {
                 const p1 = pts[i];
                 const p2 = pts[i + 1];
                 const p3 = pts[i + 2];
-
-                this._testTriangle(i, trianglePlaneArray, p3, p2, p1, hasMaterial, hostMesh);
+                
+                if (invertTriangles) {
+                    this._testTriangle(i, trianglePlaneArray, p1, p2, p3, hasMaterial, hostMesh);
+                } else {
+                    this._testTriangle(i, trianglePlaneArray, p3, p2, p1, hasMaterial, hostMesh);
+                }
             }
         } else {
             for (let i = indexStart; i < indexEnd; i += 3) {
@@ -409,7 +413,11 @@ export class Collider {
                 const p2 = pts[indices[i + 1] - decal];
                 const p3 = pts[indices[i + 2] - decal];
 
-                this._testTriangle(i, trianglePlaneArray, p3, p2, p1, hasMaterial, hostMesh);
+                if (invertTriangles) {
+                    this._testTriangle(i, trianglePlaneArray, p1, p2, p3, hasMaterial, hostMesh);
+                } else {
+                    this._testTriangle(i, trianglePlaneArray, p3, p2, p1, hasMaterial, hostMesh);
+                }
             }
         }
     }
@@ -418,7 +426,7 @@ export class Collider {
     public _getResponse(pos: Vector3, vel: Vector3): void {
         pos.addToRef(vel, this._destinationPoint);
         vel.scaleInPlace((this._nearestDistance / vel.length()));
-
+    
         this._basePoint.addToRef(vel, pos);
         pos.subtractToRef(this.intersectionPoint, this._slidePlaneNormal);
         this._slidePlaneNormal.normalize();
@@ -426,7 +434,7 @@ export class Collider {
 
         pos.addInPlace(this._displacementVector);
         this.intersectionPoint.addInPlace(this._displacementVector);
-
+        
         this._slidePlaneNormal.scaleInPlace(Plane.SignedDistanceToPlaneFromPositionAndNormal(this.intersectionPoint, this._slidePlaneNormal, this._destinationPoint));
         this._destinationPoint.subtractInPlace(this._slidePlaneNormal);
 
--- a/src/Collisions/collider.ts
+++ b/src/Collisions/collider.ts
@@ -426,7 +426,7 @@ export class Collider {
     public _getResponse(pos: Vector3, vel: Vector3): void {
         pos.addToRef(vel, this._destinationPoint);
         vel.scaleInPlace((this._nearestDistance / vel.length()));
-    
+
         this._basePoint.addToRef(vel, pos);
         pos.subtractToRef(this.intersectionPoint, this._slidePlaneNormal);
         this._slidePlaneNormal.normalize();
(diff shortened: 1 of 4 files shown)


src/Collisions/collider.ts
```diff
@@ -426,7 +434,7 @@ export class Collider {
 
         pos.addInPlace(this._displacementVector);
         this.intersectionPoint.addInPlace(this._displacementVector);
-
+
```
sebavan, line 437: should not be here :-)

src/Meshes/abstractMesh.ts
```diff
@@ -1718,7 +1724,7 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
         // Transformation matrix
         const collisionsScalingMatrix = TmpVectors.Matrix[0];
         const collisionsTransformMatrix = TmpVectors.Matrix[1];
-        Matrix.ScalingToRef(1.0 / collider._radius.x, 1.0 / collider._radius.y, 1.0 / collider._radius.z, collisionsScalingMatrix);
+        Matrix.ScalingToRef(1.0 / collider._radius.x, 1.0 / collider._radius.y, 1.0 / collider._radius.z, collisionsScalingMatrix);
```
sebavan, line 1727: Trailing space

src/Meshes/mesh.ts
```diff
@@ -4501,6 +4501,11 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
             this.instances.pop();
         }
     }
+
+    /** @hidden */
+    public _shouldConvertRHS() {
+        return this.overrideMaterialSideOrientation === Material.CounterClockWiseSideOrientation;
```
sebavan, line 4507: I am wondering if it is always true when for instance scaling is -1 on one axis only, determinant of the matrix being negative ???

Popov72: The problem is with geometries that are described in rhs because the collision code expects it to be lhs.

When loading a glTF mesh with `scene.useRightHandedSystem=false` everything is ok because the root node the loader creates has a transformation that transforms the rhs geometry to lhs. However, when `scene.useRightHandedSystem=true` the loader sets an identity transformation and the rhs geometry stays rhs. That's the case we need to detect.

When a mesh is loaded by the glTF loader, we can detect this case by testing `this.overrideMaterialSideOrientation === Material.CounterClockWiseSideOrientation` (see https://github.com/BabylonJS/Babylon.js/issues/11688#issuecomment-1002219692), hence the test in @carolhmj code.

Apart from glTF, we normally don't handle rhs geometries so we don't have the problem. The problem can only arise if a user is creating "by hand" a rhs geometry and does not set a transformation that transforms the geometry from rhs to  lhs. But in that case the user would have to do the same thing than the glTF loader and set `overrideMaterialSideOrientation = Material.CounterClockWiseSideOrientation` else the geometry would be inside out.

deltakosh: > That's the case we need to detect

Yes I agree and then why not testing `scene.useRightHandedSystem=true`

Popov72: Because meshes not loded with the glTF loader will also have `scene.useRightHandedSystem==true` but they should not trigger the alternate collision code because their geometry is lhs.

deltakosh: well if you create the mesh in RH mode you will have to invert them to display correctly right? I get that you can override the materialSideOrientation but you can also create them with RH in mind.

Popov72: What do you mean by
> you will have to invert them to display correctly right?

and

> but you can also create them with RH in mind.

?

If the geometry is in rhs, you can either list the vertex indices backward (compared to creating a lhs geometry) or set the `sideOrientation` accordingly. However, I don't think the first solution is the right one because you should order the indices so that the triangle displays correctly in a rhs system when you create a rhs geometry. That's what happens in glTF: the vertex positions and indices are described in a rhs system and the glTF loader does not change them, it updates the `sideOrientation` property instead.

deltakosh: Yes that what I meant: you could create your geometry to work directly with RH (inverting indices). But in the nutshell, what I try to say is that testing scene.useRightHanded should suffice because I do not expect to have LH models in that mode

carolhmj: Thanks for bringing this to attention Seb, I think there are two cases here, depending if the vertex data from the mesh comes from a GLTF or not. If it comes from a GLTF, then the vertex data is RH and a negative determinant world would make it a LH and _not need conversion_. Whether if is any other kind of vertex data, then it is LH, and a negative determinant would convert to RH and need conversion. I made a diagram with two nodes to help me visualize this, since the example I work on has two nodes:
image

This looks tricky... do we need to add some kind of new property to save the vertex data's handedness? If so, whenever we bake a transformation in the vertices we also need to make sure the handedness is consistent.

carolhmj: Oh somehow Github didn't update and didn't show me @deltakosh and @Popov72 's messages O.O
We could assume that when `scene.useRightHandedSystem` is true then all of the scene's vertex data is right handed, then this could work... but I think MeshBuilder always creates vertex data in LHS, which would cause problems

deltakosh: > MeshBuilder always creates vertex data in LHS
Then then will be displayed wrong in RH mode

sebavan: I think it renders correctly cause the material is reversing the side orientation if no override exists.

sebavan: Basically I wonder if the check should be a combination of side orientation and work matrix determinant sign ???

Popov72: As @sebavan said, the lhs meshes are still rendered correctly in rhs mode because the side orientation is automatically reversed (see the constructor of `Material`).

I don't think the matrix determinant sign will help because it is not a way to know if a geometry is lhs or rhs: we can have a mesh with a lhs geometry that will either have a -1 or +1 determinant, and same thing with a rhs geometry, depending on the setup. I don't think factoring the side orientation will help. For eg, we can imagine we want to render a mesh inside-out (for whatever reason), so we would use the reverse side orientation we would have used to render the mesh normally. So, we can have any combination of matrix determinants and side orientations, even for a lhs mesh.

In my opinion, in the global case there's no way to know if a geometry is lhs or rhs, as it relies on the vertex coordinate values and the side orientation value. But the side orientation value itself is of no help because its value depends on the way you order the indices of the triangle: you can use an order and set the orientation to CCW (for eg), or order the other way around (CW) and set the orientation to CW.

We can only deal with what we know:
* all meshes created by the mesh builder are in lhs mode, so nothing to do
* all meshes loaded by glTF are rhs, and depending on `scene.useRightHandedSystem` there's a transformation that transforms them to lhs or not => `mesh.overrideMaterialSideOrientation` allows us to know when that happens

carolhmj: So @Popov72 should we keep the check the way it is now?

src/Meshes/abstractMesh.ts
carolhmj: Fixed!

src/Collisions/collider.ts
carolhmj: Fixed!

src/Meshes/mesh.ts
deltakosh: let's keep it like this and if we find issues we will fix them

deltakosh (approved):

PR #11710, opened 2022-01-05
Creation of InputTextArea component

Valerian-Perez-Wanadev:
Based on fneitzel prototype ([Added prototype of InputTextArea](https://github.com/BabylonJS/Babylon.js/pull/8275)), I improved the component to support the margin, the deletion (backspace/del.), the highlighting, deadkeys and copy/paste/cut shortcuts.

A lot of things remained to do :
- fix the CR on last line
- find why the text clip on single line during autostreching
- resolve todos
- clean/fix/update the code

My approach was highly naïve, so maybe a cleaner architecture will be necessery.


fneitzel:
Cool that you are finishing it, let me know if you need help.

gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1282 @@
...
+import { Measure } from '../measure';
+import { InputText } from "./inputText";
+import { ICanvasRenderingContext, ITextMetrics } from "babylonjs/Engines/ICanvas";
+import { PointerInfoBase } from "babylonjs/Events/pointerEvents";
+import { IKeyboardEvent } from "babylonjs/Events/deviceInputEvents";
+
+/**
+ * Enum that determines the text-wrapping mode to use.
+ */
+export enum TextWrapping_ {
```
RaananW, line 18: nit - why id you end with an underscore?

gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1282 @@
...
+
+    /**
+     * Wrap the text word-wise, i.e. try to add line-breaks at word boundary to fit within Control.width.
+     */
+    WordWrap = 1,
+
+    /**
+     * Ellipsize the text, i.e. shrink with trailing … when text is larger than Control.width.
+     */
+    Ellipsis,
```
RaananW, line 32: Is there a reason for these values? why use "-1"? and why doesn't ellipsis have any value?

gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1282 @@
...
+}
+
+/**
+ * Class used to create input text control
+ */
+export class InputTextArea extends InputText {
+
+    //private _textWrapping = TextWrapping_.Clip;
```
RaananW, line 40: would be great to cleanup any unneeded / commented line of code

gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1282 @@
...
+    /**
+    * An event triggered after the text was broken up into lines
+    */
+    public onLinesReadyObservable = new Observable<InputTextArea>();
+
+    private lastClickedCoordinateY = 0;
+    private _selectedLineIndex = 0;
+    /** @hidden */
+    public _connectedVirtualKeyboard: Nullable<VirtualKeyboard>;
+    public measure: ITextMetrics;
```
RaananW, line 63: nit - naming. Measure might be confusing (as it is probably thought to be the measure of the control). maybe "textMeasurement"?

gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1282 @@
...
+        if (evt && (evt.ctrlKey || evt.metaKey) && (keyCode === 67 || keyCode === 86 || keyCode === 88)) {
+            return;
+        }
+
+        //select all
+        if (evt && (evt.ctrlKey || evt.metaKey) && keyCode === 65) {
+            this._selectAllText();
+            evt.preventDefault();
+            return;
+        }
```
RaananW, line 138: why is this not a part of the switch?

gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1282 @@
...
+            }
+        }
+        context.restore();
+
+        clearTimeout(this._blinkTimeout);
+        this._blinkTimeout = <any>setTimeout(() => {
+            this._blinkIsEven = !this._blinkIsEven;
+            this._markAsDirty();
+        }, 500);
+        context.restore();
```
RaananW, line 1048: I understand why this is needed, but this could be solved within the render loop using timers - 
https://doc.babylonjs.com/divingDeeper/events/observables#observable-based-countdown-function
setTimeout could cause unexpected results, as it will surely run outside of the render loop

RaananW (commented):
Wow, thank you so much for the. Such a wonderful addition to the GUI!

I have a few comments. The code should be cleaned and formatted, but those are minor issues.

My bigger question is this - I noticed you implemented yourself backspace, delete and so on. This is, eventually, a 3d-controlled <textarea> element. Why not use a textarea as the backend of this component and allow the browser to do the heavy-lifting of text editing?

RaananW:
I also wait for @DarraghBurkeMS's review to test the actual functionality.

Popov72:
> Why not use a textarea as the backend of this component and allow the browser to do the heavy-lifting of text editing?

What about Native if using a textarea HTML element under the hood?

RaananW:
> What about Native if using a textarea HTML element under the hood?

That's a good point. @bghgary , is there some corresponding element to be used, or does the functionality needs to be (re) implemented?

darraghjburke:
I'm not Gary, but: I don't think we can depend on any external text area implementation in Babylon Native. One of the critical design choices is that it's "bring your own UI", so we can't rely on any particular one of the many cross-platform UI solutions out there. The goal AFAIK is to implement enough of the canvas API that we can support Babylon GUI, but Babylon Native has no plans to support any DOM features (at that point we'd be making a browser.)

From my POV, having a Babylon GUI text area brings a lot of value, but it's also worth keeping in mind that it will probably be significantly harder to maintain than our existing controls. That being said, I love that you've taken a stab at it, and I'll definitely give it a try later this week :)

deltakosh:
Totally aligned with @darraghjburke : no external dependencies beside just canvas API

bghgary:
> What about Native if using a textarea HTML element under the hood?

Maybe @CedricGuillemet can answer this part.

In general, I'm also curious if there are any localization concerns with this change.

Valerian-Perez-Wanadev:
Thanks a lot for all your feedbacks, I'll fix that later in the week!

CedricGuillemet:
> > What about Native if using a textarea HTML element under the hood?
> 
> Maybe @CedricGuillemet can answer this part.
> 

Hard to tell without testing. At the current state, (basic) text and shape rendering will work. Idk if any unimplemented method is needed here.

deltakosh:
@Valerian-Perez-Wanadev : @fneitzel is interested to contribute to your PR. Do you mind opening your branch for his PR?

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: It's an old implementation of first prototype. I removed it and reimported from the correct file.

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: It was certainly based on the enum in `textBlock.ts:14` which has the same behaviour. But I don't know why "-1" was used.

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: Removed!

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: Correct. It's changed to `textMetrics` to be complient with the type.

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: Oh, great!

Fneitzel uses the same method as in `inputText.ts`, but I can improve it today.

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: Oh, yes.

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: Currently, I'm facing some issues over AdvancedTimer class. 
- Is it correct to instantiate it in `_link` function to have access to the scene ?
- I can't stop/start in the same loop to reset the timer. 

I want to avoid SetAndStart to save some perf but it becomes to be the only solution...
Have you some ideas to bypass this ?

Valerian-Perez-Wanadev:
I've made some good improvments during the sart of the year but I'm in a dilemma...

The Component inherits from InputText, but it acts like a TextBlock. So for the textWrapping behaviour, I must follow the implementation of only Clip/WordWrap or Clip/WordWrap/Ellipsis/WordWrapEllipsis?

Furthermore, the ellipsis mode it's not very useful in InputTextArea... I choose to discard it for the moment. But I want to known your thoughts about this before merging my code.

deltakosh:
For things in common we can think about having a common ancestor?

Valerian-Perez-Wanadev:
> For things in common we can think about having a common ancestor?

Of course I will do thatif needed but actually my choice is on which behaviour to adopt. 

After seeing implementation of TextArea of Firefox and the InputText of BJS, I will keep their philosophy and let the user decides whether their add a line or let the component breaks the text alone. It's more natural to do it that way.

Valerian-Perez-Wanadev:
I have a question about the Enter key.

How can I override the blur by hitting Enter ? 

I want to add a new line with Enter but I can't figure how to do without blurring the input:/

deltakosh:
the onkeydown can be cancelled if you detect Enter (preventDefault)

dad72:
This a very useful feature. Thank you. I can't wait to use it to replace what I did in HTML

Valerian-Perez-Wanadev:
It's almost done. Some bugs are complicated to fix but I'm always working on it :)

jonathanandrewsuk:
@darraghjburke @RaananW just bumping this to register my appreciation for this PR!

packages/dev/gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1257 @@
...
+
+    /**
+     * Process the last keyboard input
+     *
+     * @param keyCode The ascii input number
+     * @param key The key string representation
+     * @param evt The keyboard event emits with input
+     * @hidden
+     */
+    public processKey(keyCode: number, key?: string, evt?: IKeyboardEvent) {
```
RaananW, line 180: keyCode was already deprecated and will slowly be removed from browsers. And we don't support IE11 anymore (bless whoever's in charge of that :-)) - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Would it be possible to work with `key` or `code` instead? Or was there any specific reason for choosing to work with keyCode?

I understand we use this internally quite often. And we need to change that (sooner rather than later). I am just trying to avoid commiting new code with deprecated APIs.

packages/dev/gui/src/2D/controls/inputTextArea.ts
```diff
@@ -0,0 +1,1257 @@
...
+    }
+
+    /**
+     * Select the word immediatly under the cursor on double click
+     *
+     * @param evt Pointer informations of double click
+     * @hidden
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    protected _processDblClick(evt: PointerInfo) {
```
RaananW, line 1210:
```suggestion
    protected _processDblClick(_evt: PointerInfo) {
```

(or remove the variable altogether :-) )

RaananW (commented):
This looks great!!

I just have a question about keyCode vs. key (and just a small suggestion that can be ignored).
Is it possible to see a working playground with this in the snapshot playground?

matthewpflueger:
Would love to see this merged! This is a perfect fit for our project...

packages/dev/gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: I prefer to keep the parameter, it will be more maintainable after the release (just in case).

packages/dev/gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: Hmm, I didn't notice that... I don't have any reason to keep this deprecated API. I will rewrite the processkey ASAP 👍

Valerian-Perez-Wanadev: @RaananW, unfortnaly I can't reproduce a correct behavior without the deprecation. The charCode filters the character and discard all modifier key like `Shift`, `Alt`, etc.

Actually, I can achieve almost the same things but I'm stuck to allow two duet characters: `^^` and ` `` `

gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: I can't restart the timer during the same frame which it's stop

packages/dev/gui/src/2D/controls/inputTextArea.ts
Valerian-Perez-Wanadev: @RaananW, Can I push this evol?

RaananW (approved):

packages/dev/gui/src/2D/controls/inputTextArea.ts
RaananW: Can this be resolved, or do we have some more to discuss?

Valerian-Perez-Wanadev: We can resolve of course! :)

Valerian-Perez-Wanadev:
@RaananW  Just one thing before merging. It's possible to assign @fneitzel as co-author on the changelog please? Originally it was his code and we work sometimes together on this final version.

RaananW:
> @RaananW Just one more thing before merging. It's possible to assign @fneitzel as co-author on the changelog please? Originally it was his code and we work sometimes together on this final version.

I will have to edit the changelog afterwards, as it is generated automatically, but i'd be happy to do that. Please remind me after the next version is merged :-)


Also a request regarding this PR - we will need to update the documentation with this new feature. Adding one or two working examples will also be wonderful. Will you be able to do that?

Valerian-Perez-Wanadev:
> > @RaananW Just one more thing before merging. It's possible to assign @fneitzel as co-author on the changelog please? Originally it was his code and we work sometimes together on this final version.
> 
> I will have to edit the changelog afterwards, as it is generated automatically, but i'd be happy to do that. Please remind me after the next version is merged :-)
> 
> Also a request regarding this PR - we will need to update the documentation with this new feature. Adding one or two working examples will also be wonderful. Will you be able to do that?

I'll do that 👍 

And my local PG has some beautiful examples!

jonathanandrewsuk:
@Valerian-Perez-Wanadev I'd be happy to help out with the docs (I've been using the component for a few weeks already)

Issue #11708, opened 2022-01-04
Setting bone rotation to (0,0,0) leads to bone flipping on every frame


carolhmj:
Related forum issue: https://forum.babylonjs.com/t/setting-rotation-quaternion-to-0-1-0-0-leads-to-the-bone-keep-flipping-180-degrees/25719

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com): https://playground.babylonjs.com/#BCU1XR#3784
- Expected result: Bone should keep a consistent orientation
- Current result: Bone flips orientation on each frame

PR #11712, opened 2022-01-05
Fix non initialized matrix when setting orientation for bones

CedricGuillemet:
Follow up https://forum.babylonjs.com/t/setting-rotation-quaternion-to-0-1-0-0-leads-to-the-bone-keep-flipping-180-degrees/25719

scaleMatrix is using a private matrix that is not initialized. Its value at index 0 is multiplied by -1 (determinant) giving a flip each frame. 

@carolhmj Can you please take some time to test this fix. Because I could not repro the issue reliably before the fix.


RaananW:
Fixing #11708

RaananW (approved):

carolhmj (approved):
Tested and approved!

CedricGuillemet:
Thank you @carolhmj !

PR #11713, opened 2022-01-05
[XR] set _framebuffer even when "falsy"

RaananW:
This prevented the WebXR Emulator from working.


sebavan (approved):
I can not get why.... :-)

RaananW:
> I can not get why.... :-)

You and me both. We either check somewhere for "null" instead of "false" (which i couldn't find), or the emulator itself expects something that I simply can't explain.
The emulator works locally when this is set. that's all I can say ;-)

PR #11716, opened 2022-01-05
Update texture dirtying

carolhmj:
So that we don't update all materials in a scene when just one texture is changed.


sebavan (approved):
Thanks a lot !!!

deltakosh (approved):

PR #11726, opened 2022-01-06
Hash code fix

deltakosh:


sebavan (approved):

src/Maths/math.vector.ts
```diff
@@ -10,6 +10,10 @@ import { PerformanceConfigurator } from '../Engines/performanceConfigurator';
 
 type TransformNode = import('../Meshes/transformNode').TransformNode;
 
+var _ExtractAsInt = (value: number) => {
+    return parseInt(value.toString().replace(/\W/g, ""));
```
Popov72, line 14: Don't you want to use `\D` instead of `\W`?

src/Maths/math.vector.ts
```diff
@@ -47,8 +51,11 @@ export class Vector2 {
      * @returns the Vector2 hash code as a number
      */
     public getHashCode(): number {
-        let hash = this.x | 0;
-        hash = (hash * 397) ^ (this.y | 0);
+        const x = _ExtractAsInt(this.x);
```
Popov72, line 54: Is this not a fix for a bug on the user side?

`x` and `y` are declared as numbers so they should not hold string data. If they are, it means the user put wrong (invalid) values there. I don't know at which frequency `getHashCode` can be called, but calling `_ExtractAsInt` two times is probably taking most of the time of the whole function time.

deltakosh: nope the problem was that this.x | 0 will convert to an int. So losing everything after the comma. 

The function is not meant to be fast anyway

src/Maths/math.vector.ts
deltakosh: nope I want to remove everything which is not a number

src/Maths/math.vector.ts
deltakosh: https://forum.babylonjs.com/t/gethashcode-of-vector3-class-does-not-work-on-some-x-y-z-values/26519/5

src/Maths/math.vector.ts
```diff
@@ -10,6 +10,10 @@ import { PerformanceConfigurator } from '../Engines/performanceConfigurator';
 
 type TransformNode = import('../Meshes/transformNode').TransformNode;
 
+var _ExtractAsInt = (value: number) => {
+    return parseInt(value.toString().replace(/\W/g, ""));
+}
```
RaananW, line 15:
```suggestion
};
```
(lint)

src/Maths/math.vector.ts
```diff
@@ -47,8 +51,11 @@ export class Vector2 {
      * @returns the Vector2 hash code as a number
      */
     public getHashCode(): number {
-        let hash = this.x | 0;
-        hash = (hash * 397) ^ (this.y | 0);
+        const x = _ExtractAsInt(this.x);
+        const y = _ExtractAsInt(this.y);
+
```
RaananW, line 56:
```suggestion

```
(lint)

src/Maths/math.vector.ts
RaananW: But then `123.123` will have the same result as `1231.23` or `123123`, no?

RaananW (commented):
The linting issues as suggestions

src/Maths/math.vector.ts
deltakosh: correct. The goal is not to have a perfect hash as it is impossible but at least to support more cases

RaananW (approved):

src/Maths/math.vector.ts
Popov72: > nope I want to remove everything which is not a number

That's what `\D` is for, no? `\W` won't remove a-z/A-Z but `\D` will.

`_ExtractAsInt("32a43")` will currently return 32 whereas it will return 3243 if using `\D`.

`_ExtractAsInt("a3243")` will currently return NaN whereas it will return 3243 if using `\D`.

src/Maths/math.vector.ts
Popov72: > So losing everything after the comma.

You will also loose everything after the comma with the current PR. Maybe you want ot use `parseFloat` instead of `parseInt`?

deltakosh: > You will also loose everything after the comma with the current PR 

How's that? The string get stripped of the "." and thus it becomes a long int.

> The fix could not be to simply remove | 0

The hash needs to be an int

src/Maths/math.vector.ts
deltakosh: wait ;) the vector3 contains float ;)

```
var temp = 0.255454685;
console.log(temp.toString().replace(/\W/g, ""));
```

will return: 255454685

deltakosh: This PR is not to support strings in Vector ;)

src/Maths/math.vector.ts
Popov72: Yes, too many holidays are not good for the brain, sorry for the noise :)

deltakosh: haha no worries! Enjoy your holidays!

PR #11727, opened 2022-01-06
[XR] fix incorrect flag

RaananW:


RaananW:
(an obvious bug, merged to fix it)

PR #11729, opened 2022-01-06
Enable performance profiler and save the categories when exporting.

carolhmj:


src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -197,6 +200,10 @@ export class PerformanceViewerCollector {
 
             this.datasets.ids.push(strategy.id);
 
+            if (category) {
+                category = category.replace(exportedDataSeparator, "");
```
RaananW, line 204: replace will only remove the first instance, unless you are using a regex. there is a replaceAll, or better - use a regex:
```suggestion
                category = category.replace(new RegExp(exportedDataSeparator, "g"), "");
```
if you wonder why I use the RegExp constructor instead of /@/g, it's because you can't use variables between the two `/ `

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
```diff
@@ -20,11 +20,10 @@ interface IPerformanceViewerComponentProps {
     scene: Scene;
 }
 
-// aribitrary window size
+// arbitrary window size
 const initialWindowSize = { width: 1024, height: 512 };
 
-// Note this should be false when committed until the feature is fully working.
-const isEnabled = false;
+const isEnabled = true;
```
sebavan, line 26: should we not remove the flag altogether ??? now that it is live

src/Misc/PerformanceViewer/performanceViewerCollector.ts
```diff
@@ -385,9 +392,13 @@ export class PerformanceViewerCollector {
             return false;
         }
 
+        let idCategoryMap : Map<string, string> = new Map<string, string>();
```
sebavan, line 395: Nit: const

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
RaananW: yes, that makes sense

deltakosh: Agreed

src/Misc/PerformanceViewer/performanceViewerCollector.ts
carolhmj: Updated!

inspector/src/components/actionTabs/tabs/performanceViewer/performanceViewerComponent.tsx
carolhmj: Removed the flag :D

src/Misc/PerformanceViewer/performanceViewerCollector.ts
carolhmj: Updated :D

RaananW (approved):

PR #11732, opened 2022-01-06
Replace input handling in InputManager with DeviceSourceManager

PolygonalSun:
This PR contains several changes
- All observables have been removed from the WebDeviceInputSystem to improve parity with NativeDeviceInputSystem.
- DeviceSourceManager is now the primary source of input handling for the Input Manager.
- The engine now keeps track of an implementation of the DeviceSourceManager, instead of the DeviceInputSystem.
- The DeviceSourceManager class acts as a sort of wrapper for the engine-only singleton of InternalDeviceSourceManager. (Implemented this way because of backwards-compatibility)
- On web, touch no longer creates and deletes touch devices on a touch event.  The new system now creates slots for the maximum number of touches (usually 10, if touchscreen exists) and maintains them as inputs come in.  Anything that exceeds the max number of touches is ignored.


src/Engines/engine.ts
```diff
@@ -391,9 +391,9 @@ export class Engine extends ThinEngine {
     public isPointerLock = false;
 
     /**
-     * Stores instance of DeviceInputSystem
+     * @hidden
      */
-    public deviceInputSystem: DeviceInputSystem;
+    public _deviceSourceManager: InternalDeviceSourceManager;
```
bghgary, line 396: This seems backwards to me. Why not augment the engine from the DeviceSourceManager side and then you won't have to expose InternalDeviceSourceManager at all. You can use engine.onDisposeObservable to dispose the instance.

src/DeviceInput/InputDevices/internalDeviceSourceManager.ts
```diff
@@ -0,0 +1,179 @@
...
+import { Observable, Observer } from '../../Misc/observable';
+import { IDeviceEvent, IDeviceInputSystem } from './inputInterfaces';
+import { NativeDeviceInputSystem } from './nativeDeviceInputSystem';
+import { WebDeviceInputSystem } from './webDeviceInputSystem';
+import { DeviceSource } from './deviceSource';
+import { INative } from '../../Engines/Native/nativeInterfaces';
+
+declare const _native: INative;
+
+/** @hidden */
```
bghgary, line 14: I believe once you put `@hidden` the top level class, you don't need to add `@hidden` for each child public member, but this won't matter if we can do what I said in the other comment since this class will then be private.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -89,7 +89,7 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
     /**
      * Configures events to work with an engine's active element
      */
-    public configureEvents() {
+    private configureEvents() {
```
bghgary, line 92:
```suggestion
    private _configureEvents() {
```

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -141,7 +141,7 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
      * @param deviceType Type of device to check for
      * @returns bool with status of device's existence
      */
-    public isDeviceAvailable(deviceType: DeviceType) {
+    public isDeviceAvailable(deviceType: DeviceType): boolean {
```
bghgary, line 144: nit: I would recommend adding return types to all of the functions not just this one.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -330,6 +330,15 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
      * Handle all actions that come from pointer interaction
      */
     private _handlePointerActions() {
+        this._maxTouchPoints = navigator.maxTouchPoints || 0;
```
bghgary, line 333: Is it safe to use navigator without checking that it is available? Maybe to be safe just add a check `DomManagement.IsNavigatorAvailable()`.

```suggestion
        this._maxTouchPoints = (DomManagement.IsNavigatorAvailable() && navigator.maxTouchPoints) || 0;
```

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -60,8 +60,8 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
     private _isUsingFirefox = navigator && navigator.userAgent && navigator.userAgent.indexOf("Firefox") !== -1;
```
bghgary, line 60: Not sure this is the right way to check for navigator existence.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -406,8 +415,16 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
             let deviceSlot = (deviceType === DeviceType.Mouse) ? 0 : evt.pointerId;
 
             if (deviceType === DeviceType.Touch) {
-                deviceSlot = this._rollingTouchId++;
-                this._activeTouchIds[deviceSlot] = evt.pointerId;
+                let idx = this._activeTouchIds.indexOf(-1);
```
bghgary, line 418: nit:
```suggestion
                const idx = this._activeTouchIds.indexOf(-1);
```

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -406,8 +415,16 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
...
-                deviceSlot = this._rollingTouchId++;
-                this._activeTouchIds[deviceSlot] = evt.pointerId;
+                let idx = this._activeTouchIds.indexOf(-1);
+
+                if (idx >= 0) {
+                    deviceSlot = idx;
+                    this._activeTouchIds[idx] = evt.pointerId;
+                }
+                else {
+                    // We can't find an open slot to store new pointer so just return (can only support max number of touches)
```
bghgary, line 425: Maybe log a warning?

sebavan (commented):
@bghgary said it all :-)

src/Engines/engine.ts
PolygonalSun: Discussed offline

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: Added warning

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: changed

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: changed

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: added

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: changed

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: changed

src/DeviceInput/InputDevices/internalDeviceSourceManager.ts
PolygonalSun: Talked offline

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -636,8 +636,6 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                         deviceEvent.previousState = 1;
                         this.onInputChanged(deviceEvent);
 
-                        this._unregisterDevice(DeviceType.Touch, deviceSlot);
```
bghgary, line 620: why was this removed?

PolygonalSun: So I had originally removed this because I had a rolling system of active ids and wanted to prevent the any unnecessary array deletions for touch.  After doing a couple of edge tests, I've found that I need to have something still trigger the onDeviceDisconnected functionality so I'm going to add a direct call to it in place of this.

sebavan:
I ll let @bghgary and @deltakosh check the final bits (tech details) this week but was happy with the overall approach on our Friday chat.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -41,7 +45,7 @@ export class WebDeviceInputSystemImpl implements IDeviceInputSystem {
     private _pointerActive: boolean = false;
     private _elementToAttachTo: HTMLElement;
     private _engine: Engine;
```
bghgary, line 47: nit: could be readonly. maybe can do a pass on this later.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
bghgary: Are we confident the array won't grow too big?

bghgary (commented):
Looks good. Just one last verification.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: The touch array should never grow any larger than the value of _maxTouchPoints which should be no greater than 10 (platform-dependent).

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: Easy enough to change

bghgary: There are more that can be readonly, but no need to do this now.

bghgary (approved):

Issue #11605, opened 2021-12-08
Advance Texture Bug, Babylon file appears to save incorrectly. 


msDestiny14:
Saving the scene as babylon file loads incorrectly in sandbox. The sandbox’s output is not matching with PG sample’s output.

https://forum.babylonjs.com/t/advanced-texture-bug/25859

PR #11734, opened 2022-01-06
Serialization: Use unique IDs for materials and multimaterials

darraghjburke:
Makes Serializer and Babylon Loader reference materials and multimaterials using unique IDs instead of IDs, which means that having two materials with the same ID/name will no longer cause flawed serialization.

Playground repro for multi materials: https://playground.babylonjs.com/#KAXWT5 If you save this file to .babylon and then open in the sandbox, the materials will be wrong. With these changes, it will retain the correct materials.

One question: is there a better way to set the uniqueId on the serialization object rather than doing it explicitly for every single Material class? I thought having @Serialize on the uniqueId property in Material.ts would be enough, but apparently it's not.

Addresses #11605


src/Materials/multiMaterial.ts
```diff
@@ -188,18 +188,20 @@ export class MultiMaterial extends Material {
 
         serializationObject.name = this.name;
         serializationObject.id = this.id;
+        serializationObject.uniqueId = this.uniqueId;
```
darraghjburke, line 191: Is there a way to avoid doing this for every material class?

deltakosh: what we could do is do it at Material level (with id, name and tags) and simply call it from each child. That will be painful only once ;)

src/Loading/Plugins/babylonFileLoader.ts
```diff
@@ -589,21 +590,35 @@ SceneLoader.RegisterPlugin({
...
-                                            var subMatId = parsedMultiMaterial.materials[matIndex];
-                                            loadedMaterialsIds.push(subMatId);
-                                            var mat = parseMaterialById(subMatId, parsedData, scene, rootUrl);
-                                            if (mat) {
-                                                log += "\n\tMaterial " + mat.toString(fullDetails);
+                                    const parsedMultiMaterial = parsedData.multiMaterials[multimatIndex];
+                                    if ((parsedMesh.materialUniqueId && parsedMultiMaterial.uniqueId === parsedMesh.materialUniqueId) || parsedMultiMaterial.id === parsedMesh.materialId) {
+                                        if (parsedMultiMaterial.materialsUniqueIds) { // if the materials inside the multimat are stored by unique id
+                                            for (var matIndex = 0, matCache = parsedMultiMaterial.materialsUniqueIds.length; matIndex < matCache; matIndex++) {
```
deltakosh, line 602: Is there a way to factorize this code a bit more?

src/Materials/multiMaterial.ts
```diff
@@ -245,21 +247,32 @@ export class MultiMaterial extends Material {
...
-            var subMatId = parsedMultiMaterial.materials[matIndex];
-
-            if (subMatId) {
-                // If the same multimaterial is loaded twice, the 2nd multimaterial needs to reference the latest material by that id which
-                // is why this lookup should use getLastMaterialByID instead of getMaterialByID
-                multiMaterial.subMaterials.push(scene.getLastMaterialById(subMatId));
-            } else {
-                multiMaterial.subMaterials.push(null);
-            }
+        if (parsedMultiMaterial.materialsUniqueIds) {
```
deltakosh, line 256: Again I wonder if we can refactor that code?

deltakosh (changes requested):
Overall all good! thanks a lot

src/Loading/Plugins/babylonFileLoader.ts
darraghjburke: Sure, I'll try to break it into a couple functions

src/Materials/multiMaterial.ts
darraghjburke: I can make it a function that we call twice but I am not sure that will improve readability

deltakosh: I feel bad when I see the same line multiple times ;D

src/Materials/multiMaterial.ts
darraghjburke: Done!

src/Loading/Plugins/babylonFileLoader.ts
darraghjburke: How's this?

src/Materials/multiMaterial.ts
darraghjburke: Oh, I found a way. How's that?

src/Loading/Plugins/babylonFileLoader.ts
```diff
@@ -589,21 +590,28 @@ SceneLoader.RegisterPlugin({
                         }
 
                         // Material ?
-                        if (parsedMesh.materialId) {
-                            var materialFound = (loadedMaterialsIds.indexOf(parsedMesh.materialId) !== -1);
+                        if (parsedMesh.materialUniqueId || parsedMesh.materialId) {
+                            // if we have a unique ID, look up and store in loadedMaterialsUniqueIds, else use laodedMaterialsIds
```
Popov72, line 594: nit: "loadedMaterialsIds"

src/Loading/Plugins/babylonFileLoader.ts
```diff
@@ -589,21 +590,28 @@ SceneLoader.RegisterPlugin({
                         }
 
                         // Material ?
-                        if (parsedMesh.materialId) {
-                            var materialFound = (loadedMaterialsIds.indexOf(parsedMesh.materialId) !== -1);
+                        if (parsedMesh.materialUniqueId || parsedMesh.materialId) {
```
Popov72, line 593: There are several locations where `materialUniqueId` is assumed to not be equal to 0, but it is indeed possible that `materialUniqueId=0` (`Material.uniqueId` (==`materialUniqueId`) is assigned with a call to `Scene.getUniqueId()` which is using `UniqueIdGenerator.UniqueId` which starts at 0).

Maybe the easiest fix would be to make `UniqueIdGenerator.UniqueId` starts at 1...

src/Meshes/mesh.ts
```diff
@@ -2561,6 +2561,34 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
 
+    /**
+     * Sets the mesh material by the material or multiMaterial `uniqueId` property
+     * @param uniqueId is a string identifying the material or the multiMaterial
+     * @returns the current mesh
+     */
+    public setMaterialByUniqueId(uniqueId: string): Mesh {
+        var materials = this.getScene().materials;
+        var index: number;
+        for (index = materials.length - 1; index > -1; index--) {
```
Popov72, line 2572: nit: const/let instead of var

src/scene.ts
```diff
@@ -2797,6 +2797,21 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         return null;
     }
 
+    /**
+     * Gets a the last added material using a given unique id
+     * @param uniqueId defines the material's unique Id
+     * @return the last material with the given uniqueId or null if none found.
+     */
+     public getLastMaterialByUniqueId(uniqueId: string): Nullable<Material> {
+        for (var index = this.materials.length - 1; index >= 0; index--) {
```
Popov72, line 2806: nit: let instead of var

src/Materials/standardMaterial.ts
```diff
@@ -1860,18 +1860,6 @@ export class StandardMaterial extends PushMaterial {
...
     }
 
-    /**
-     * Serializes this material in a JSON representation
-     * @returns the serialized material object
-     */
-    public serialize(): any {
-        const serializationObject = SerializationHelper.Serialize(this);
-
-        serializationObject.stencil = this.stencil.serialize();
```
sebavan, line 1870: Are we not losing this info ???

darraghjburke: We are just inheriting from Material.ts now, which serializes all the same info

sebavan: perfect

RaananW:
We already have the concept of "uniqueId", that is scene-based and could change on each load of the scene. It's mainly internal and is not being used by developers. Maybe extending this functionality (provide a reusable unuqueId for serialization) would be less confusing here? Nodes have unqiueId already, we just need to extend it to the materials, and maybe provide a "uniqueIdOffset" when serializing to be sure the ids don't overlap (naive, untested approach)

julien-moreau:
Hi all!
I agree with @RaananW. Also, dealing only with uniqueId (and consider `.id` property deprecated while serializing/parsing) will generate conflicts in future. Having 2 scenes being loaded will both have their unique ids starting from 0 and parenting may be corrupted. I experienced this using Babylon.JS v5 in the editor and I had to to force the legacy behavior of the babylonjs loader to take the property `.id` instead of `uniqueId` for the parents

deltakosh:
@julien-moreau I fixed that recently. It should work flawlessly as the system will use the uniqueId within the scene and not across all

julien-moreau:
Wooooooow i missed this commit! I’m testing asap!

src/Loading/Plugins/babylonFileLoader.ts
darraghjburke: Sure. Is this change okay @deltakosh ?

src/Loading/Plugins/babylonFileLoader.ts
darraghjburke: Fixed!

src/Meshes/mesh.ts
darraghjburke: Fixed :)

src/scene.ts
darraghjburke: Also fixed

src/Loading/Plugins/babylonFileLoader.ts
deltakosh: yes no problem

src/Materials/multiMaterial.ts
deltakosh: great

darraghjburke:
Tested and confirmed this works even with two scenes with flipped unique IDs

src/Meshes/mesh.ts
```diff
@@ -3486,11 +3514,11 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         // Material
         if (this.material) {
             if (!this.material.doNotSerialize) {
-                serializationObject.materialId = this.material.id;
+                serializationObject.materialUniqueId = this.material.uniqueId;
```
deltakosh, line 3517: we cannot change the current structure because old system will not be able to load the file anyway
We need to add the additional info (the uniqueId) and not remove the previous one

src/Meshes/mesh.ts
```diff
@@ -3870,8 +3898,12 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
         }
 
         // Material
-        if (parsedMesh.materialId) {
-            mesh.setMaterialById(parsedMesh.materialId);
+        const materialUniqueId = parsedMesh.materialUniqueId;
+        const materialId = parsedMesh.materialId;
+        if (materialUniqueId) {
+            mesh.setMaterialByUniqueId(materialUniqueId);
```
deltakosh, line 3904: This is where I disagree. Because if 2 scenes comes with the same uniqueId, this system will not work. What we need to do is to store the materialArray (the one containing the key (uniqueId)-> material) at a higher level so the mesh loader can use it to retrieve the current (for this file) mapping of uniqueIds

I hope it makes sense :)

deltakosh: Look what I did with tempIndexContainer in the babylon file loader

deltakosh (changes requested):
I believe we need to change a few how the system works

src/Meshes/mesh.ts
darraghjburke: Ah okay, just to confirm my understanding: we need to be able to load files created with Babylon 5.0 in earlier versions of Babylon?

deltakosh: as much as possible yes (and here the fix is easy to keep that downlevel support)

anand471:
> It's mainly internal and is not being used by developers

Oops....Just yesterday, I started using AbstractMesh.uniqueId to avoid some crazy conflicts in my application!! It's a public member listed in the documentation page & I hoped it could be used.

deltakosh (approved):

PR #11741, opened 2022-01-07
WebGPU: code cleaning

Popov72:


src/Engines/WebGPU/webgpuCacheSampler.ts
```diff
@@ -244,11 +244,34 @@ export class WebGPUCacheSampler {
...
             ...filterDescriptor,
             ...this._GetSamplerWrappingDescriptor(sampler),
-            compare: sampler._comparisonFunction ? WebGPUTextureHelper.GetCompareFunction(sampler._comparisonFunction) : undefined,
+            compare: sampler._comparisonFunction ? WebGPUCacheSampler.GetCompareFunction(sampler._comparisonFunction) : undefined,
             maxAnisotropy: filterDescriptor.anisotropyEnabled ? anisotropy : 1,
         };
     }
 
+    public static GetCompareFunction(compareFunction: Nullable<number>): GPUCompareFunction {
+        switch (compareFunction) {
```
sebavan, line 253: I always wonder if a map is faster ?

sebavan (approved):

src/Engines/WebGPU/webgpuCacheSampler.ts
Popov72: I don't know...

According to https://www.technicalfeeder.com/2021/06/map-object-makes-your-code-cleaner-than-if-else-and-switch/ (see **Performance comparison, Map, if-else, switch** in this page) a loop of 100M (in their test) is faster with map but a loop with 100K is faster with switch/case (I don't know which browser they used)...

I have tested with:
```javascript
const statusMap = new Map([
    [0, "Initializing"],
    [1, "Idle"],
    [2, "Running"],
    [3, "Stop"],
    [4, "Error"],
]);
function getStatusByMap(value) {
    return statusMap.get(value);
}
function getStatusBySwitch(value) {
    switch (value) {
        case 0: return "Initializing";
        case 1: return "Idle";
        case 2: return "Running";
        case 3: return "Stop";
        case 4: return "Error";
        default: return "Undefined";
    }
}
function measure(cb) {
    const start = new Date().getTime();
    for (let i = 0; i < 100000000; i++) {
        const val = i % 5;
        cb(val);
    }
    const end = new Date().getTime();
    console.log(`${cb.name}: ${end - start}`);
}
measure(getStatusByMap);
measure(getStatusBySwitch);
```
* in Chrome/Edge: map=469ms, switch=653ms
* in Firefox: map=7249ms, switch=892ms => quite a huge difference!

sebavan: nice !!! you tested it obviously :-)

PR #11744, opened 2022-01-10
Picking with MaxZ = 0

### LINKED ISSUE #11743: Setting infinite far plane breaks picking.
https://playground.babylonjs.com/#KNE0O#1156

You can see the NaN in the console log

zfar = 0 in the projection computation means infinite zfar

CedricGuillemet:
fixes #11743 

In picking  code, unprojecting with Z = 1 will break when camera.maxZ = 0. Instead use a value close to 1

### DIFF
--- a/src/Culling/ray.ts
+++ b/src/Culling/ray.ts
@@ -564,7 +564,8 @@ export class Ray {
         nearScreenSource.x = (sourceX / viewportWidth) * 2 - 1;
         nearScreenSource.y = -((sourceY / viewportHeight) * 2 - 1);
         nearScreenSource.z = -1.0;
-        var farScreenSource = TmpVectors.Vector3[1].copyFromFloats(nearScreenSource.x, nearScreenSource.y, 1.0);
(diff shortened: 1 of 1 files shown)


sebavan (approved):
You rock !!!

src/Culling/ray.ts
```diff
@@ -564,7 +564,8 @@ export class Ray {
         nearScreenSource.x = (sourceX / viewportWidth) * 2 - 1;
         nearScreenSource.y = -((sourceY / viewportHeight) * 2 - 1);
         nearScreenSource.z = -1.0;
```
Popov72, line 566: Not related to the PR but there's a problem in WebGPU (and Native I assume) where the Z coord in NDC has a `0..1` range and not `-1..1`. So `-1` should be replaced by `1` in this case.

However, there's currently no way in this function to know that the engine has a half Z range NDC (`engine.isNDCHalfZRange`)...

RaananW: Should we create an issue for this?

Popov72: Would probably be a good idea.

PR #11754, opened 2022-01-10
[GUI Editor] Major Gizmo Refactor

darraghjburke:
Significantly refactors the guiGizmo class, making the following changes:
* Uses React's render() function rather than direct DOM manipulation to display the selection box and scale points
* Supports drawing a selection box around multiple controls if you have more than one selected
* Uses DOM elements to render the selection box rather than GUI outlines. This gives us more flexibility over the appearance of the box, and allows us to use outlines for other things (displaying grid cells)
* Scale points are now represented using clearly named interfaces and enums, rather than just being an array. Removes the need for code like `index = index % 4`, so should be more readable now

Still unimplemented:
* Rotating using the gizmo
* Scaling multiple controls simultaneously

Wanted to open a PR though as this PR is already growing quite large. I will try to keep future PRs smaller, but I needed to significantly refactor this code to make the changes I wanted to make.

### DIFF
--- a/guiEditor/src/diagram/guiGizmo.tsx
+++ b/guiEditor/src/diagram/guiGizmo.tsx
@@ -17,36 +17,103 @@ export interface IGuiGizmoProps {
     globalState: GlobalState;
 }
 
+// which side of the bounding box are we on?
+enum ScalePointPosition {
+    Top = -1,
+    Left = -1,
+    Center = 0,
+    Right = 1,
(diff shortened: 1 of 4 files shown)


RaananW (approved):
LGTM!

I will be nit-picky and say - the only thing I want to be changed are the use of `==` and `!=` instead of `===` and `!==`.

PR #11756, opened 2022-01-11
[ACE] add Contex lock last Frame/Value properity and keyDataChanged Event

Moriyer:
lockLastFrameValue: make sure lastFrame's value couldn't be changed in ACE
lockLastFrameValue: make sure lastFrame's frame couldn't be changed in ACE
onActiveKeyDataChanged: event when keypoint value/frame/inTangent/outTangent changed

### DIFF
--- a/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/context.ts
+++ b/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/context.ts
@@ -84,6 +84,11 @@ export class Context {
 
     onSelectToActivated = new Observable<{ from: number; to: number }>();
 
+    lockLastFrameValue: boolean = false;
+    lockLastFrameFrame: boolean = false;
+
+    // value frame inTangent outTangent
+    onActiveKeyDataChanged = new Observable<number>();
     public prepare() {
         this.isPlaying = false;
         if (!this.animations || !this.animations.length) {
--- a/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/keyPoint.tsx
+++ b/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/keyPoint.tsx
@@ -533,7 +533,7 @@ export class KeyPointComponent extends React.Component<IKeyPointComponentProps, 
             if (nextX !== null) {
                 newX = Math.min(nextX - epsilon, newX);
             }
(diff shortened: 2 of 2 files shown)


inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/context.ts
```diff
@@ -84,6 +84,11 @@ export class Context {
 
     onSelectToActivated = new Observable<{ from: number; to: number }>();
 
+    lockLastFrameValue: boolean = false;
+    lockLastFrameFrame: boolean = false;
+
+    // value frame inTangent outTangent
+    onActiveKeyDataChanged = new Observable<number>();
```
deltakosh, line 91: Why this one is needed when we already have onKeyValueChanged ?

Moriyer: onActiveKeyDataChanged will notifies when inTangent,outTangent,frame/value changed. 
    The reason why I want to add onActiveKeyDataChanged is that BBL doesn't have inTangent and outTangent value changed event. I want to control some properities which is effected by globaltime (Procedual Sky Params/Scene fog/Light Intensity/etc). I used to create an AnimationGroup and many Animations for each properities,  but it seems that Animation cost too much. Then I create HermiteCurve Type(work like Animation but very lit) instead of Animation , I write a ReactComponent to make ACE to edit those curve, but I find that ACE didn't has inTangent and outTangent changed event, I can't update curve automatically when inTangent/outTangent changed.

deltakosh: Why not calling onActiveKeyDataChanged for inTangent and outTangent then?

Moriyer: Like I add to create a lit hermite curve, I can convert it to Animation to edit it with ACE,. I think I need 3 events to update hermite when there is any changed in ace: add key/delete key/keydata changed. if it is only for in/outTangent, it will.need 3 obsever to achieve the third event. I think this observable will make ACE more useful to editor other hermite type data with automatically update.

deltakosh: ok fair,, Not a big deal anyway :)

PR #11757, opened 2022-01-11
scene picking precision with huge values in world matrices

CedricGuillemet:
Reported by @sebavan 
with (really) high values in mesh world matrix, picking doesn't work.
https://playground.babylonjs.com/#KNE0O#1171

This is caused by float precisions. In this method:
https://github.com/BabylonJS/Babylon.js/blob/cc9d16b2b2867bd783f0d20d6cfac17e55e17e6e/src/Culling/ray.ts#L560

world matrix is multiplied by view and projection, then inverted. Matrix is using float and precision loss makes it impossible to pick mesh later in the code.

To improve precision, view ray is computed then transformed back into the mesh coordinates system. It's not bullet proof. With even higher values, picking will have 'holes' in resolution.

This decomposition is disabled by default. To enable it:
```
BABYLON.Ray.EnableDistantPicking = true;
```

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -47,6 +47,7 @@
 - Spelling of function/variables `xxxByID` renamed to `xxxById` to be consistent over the project. Old `xxxByID` reamain as deprecated that forward to the corresponding `xxxById` ([barroij](https://github.com/barroij))
 - Added new reflector tool that enable remote inspection of scenes. ([bghgary](https://github.com/bghgary))
 - Update `createPickingRay` and `createPickingRayToRef` matrix parameter to be nullable. ([jlivak](https://github.com/jlivak))
+- Improved scene picking precision with huge values in world matrices ([CedricGuillemet](https://github.com/CedricGuillemet)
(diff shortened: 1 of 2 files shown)


sebavan:
Sounds like this might impact perfs quite substantially, let me check if there are other ways around in the customer code.

sebavan (commented):
Should it also modify the intersectsMesh to not do it twice ? in case of distant ray enable ?

otherwise love it :-) just waiting on confirmation it fixes the project.

CedricGuillemet:
> Should it also modify the intersectsMesh to not do it twice ? in case of distant ray enable ?

Nop: https://github.com/BabylonJS/Babylon.js/blob/cc9d16b2b2867bd783f0d20d6cfac17e55e17e6e/src/Culling/ray.ts#L800
There is a predicate per mesh that update the ray, then, the returned ray is computed with identity matrix

https://github.com/BabylonJS/Babylon.js/blob/cc9d16b2b2867bd783f0d20d6cfac17e55e17e6e/src/Culling/ray.ts#L837

RaananW (approved):

PR #11758, opened 2022-01-11
Added option to force sRGB Buffer support state

### LINKED ISSUE #11747: Browser freezes when loading model
Related forum issue: https://forum.babylonjs.com/t/problem-when-checkcollisions-and-userighthandedsystem-are-used-together/
This behavior doesn't happen on 4.2.

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com): https://www.babylonjs-playground.com/#I6J7UZ#49

RaananW:
Turned on by default in tests and 3dcommerce sandbox

This is done due to an error in chrome (or angle?) :

https://bugs.chromium.org/p/chromium/issues/detail?id=1256340&q=SRGB8_ALPHA8&can=4

If set during construction (like in native or webgpu) this flag is ignored. 

Fixes #11747

Points for discussion:

- Should we also support this flag in native and webgpu (i.e. - if defined, set it to this value?)? Is there a reason for that?
- The way I see it - when the error is fixed we can still support this flag (back-compat) by making sure it is defined when setting it (in thinEngine) instead of just treating it as default = false. Any other views on that?
- Any other place this needs to be turned on by default? At the moment the sandbox and the visualization tests use this flag, setting it to true.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -75,6 +75,7 @@
 - Added `IAudioEngineOptions` interface to provide the audio engine with a pre-defined Audio Context and audio destination node. ([Vandy](https://github.com/svanderbeck11))
 - Added support for cannon-es method `world.removeBody()`. Falls back to cannon method `remove()`. ([Faber](https://https://github.com/Faber-smythe))
 - Added support for ZOffset Unit as we currently only supported factor. ([Sebavan](https://github.com/sebavan/)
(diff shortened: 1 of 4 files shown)


deltakosh (approved):
I like it. As you said when this problem will be resolved we could still protect backward compat so all good for me

sebavan (approved):

PR #11762, opened 2022-01-12
Improve import mesh

deltakosh:


src/Loading/Plugins/babylonFileLoader.ts
```diff
@@ -655,18 +674,36 @@ SceneLoader.RegisterPlugin({
...
+                        transformNode.parent = scene.getLastEntryById(transformNode._waitingParentId);
+                        transformNode._waitingParentId = null;
+                    }
+                }
                 var currentMesh: AbstractMesh;
                 for (index = 0, cache = scene.meshes.length; index < cache; index++) {
                     currentMesh = scene.meshes[index];
                     if (currentMesh._waitingParentId !== null) {
                         currentMesh.parent = scene.getLastEntryById(currentMesh._waitingParentId);
+                        if (currentMesh.parent instanceof TransformNode) {
```
RaananW, line 689: isn't any mesh is an instance of TransformNode? Might just be consmetic (as you are searching it in the transform node array), but maybe checking for `.className()` will save a few nano-seconds

RaananW (approved):
LGTM! just one question

src/Loading/Plugins/babylonFileLoader.ts
deltakosh: good one!

PR #11764, opened 2022-01-12
Improving TouchHolographicButton interaction visuals

rickfromwork:
This change improves usability by adding a semi-transparent mesh to the front of the `TouchHolographicButton` to better show the space taken up by the pressable part of the button (was previously a faint white border).
Also updated the color palate to MRTK 2.0 by using `FluentMaterial`. As a consequence, I had to rewrite some logic with the `NearMenu`'s usage of the button for it's toggle pinning button (it didn't support fluent materials, which may be why the TouchHolographicButton stopped using the fluent material for a backplate in the first place). Part of this rewrite included moving the toggle logic into the base `TouchButton` class, and removing the `TouchToggleButton` class.

tl;dr:
- Made `TouchHolographicButton`'s button visible (but faint). Should make 2D interactions much easier to see
- Moved toggle logic into `TouchButton` as a switch. Set `touchButton.isToggleButton = true` to use.
  - Added toggle effect on `TouchHolographicButton`, button lights up when toggled on
- Updated near menu visual test for new button translucency


RaananW (approved):

RaananW:
There is alinting issue, will merge right after it's fixed

rickfromwork:
> There is alinting issue, will merge right after it's fixed

Whoops, meant to fix linting before I went to bed.

PR #11768, opened 2022-01-12
Draw perf profiler lines on middle

carolhmj:


carolhmj:
Found another way to fix, will do it in another PR

RaananW (commented):
Would it be possible to add a buffer to all min and max, based on the delta between them?
This way you can continue drawing as you were, but no point will ever touch the edges.

carolhmj:
> Would it be possible to add a buffer to all min and max, based on the delta between them?
> This way you can continue drawing as you were, but no point will ever touch the edges.

Can you explain this idea a bit more?

PR #11770, opened 2022-01-12
 Improved prototype of InputTextArea

fneitzel:
Hi! I wanted to make a PR for "Improved prototype of InputTextArea #11710" based on the changes of Valerian-Perez-Wanadev, but it doesn't work, since I have no write access. Can you merge this?


deltakosh:
You need to ping him directly :) we cannot merge on his branch either

PR #11772, opened 2022-01-12
Moved EnableDistantPicking flag from Ray to abstractMesh

CedricGuillemet:
followup to https://github.com/BabylonJS/Babylon.js/pull/11757
Instead of a static flag in Ray, move the flag as a member of AbstractMesh for fine grained selection of mesh needing matrix decomposition for picking.


RaananW:
This might fail because the abstractMesh class is already crowded with variables. In that case you will need to add the flag to `_InternalAbstractMeshDataInfo` amd add a getter/setter

RaananW (approved):

PR #11775, opened 2022-01-13
Use Tools.LoadFile for sprites JSON BabylonNative compat.

CedricGuillemet:
following tests for https://github.com/BabylonJS/BabylonNative/pull/970
I ran into this issue https://github.com/BabylonJS/BabylonNative/issues/971
So, instead of changing Native to support onload/onerror, change sprites JSON loading to use Tools.


RaananW (approved):

RaananW:
Awesome. we should make sure we always use Tools and never xml requests directly

PR #11776, opened 2022-01-13
[Draco] normalize attributes if required

### LINKED ISSUE #11685: Draco decoding bug in Sandbox for GLB files
Four assets are attached:
- **no_draco_float_uvs.glb** - No draco compression, float UVs
- **draco_float_uvs.glb** - Draco compression, float UVs
- **no_draco_ushort_uvs.glb** - No draco compression, normalized unisigned short UVs
- **draco_ushort_uvs.glb** - Draco compression, normalized unisigned short UVs (**problem asset**)

While draco and the UV format vary, everything else in these assets is identical.
The asset which has draco applied to normalized unsigned short UVs doesn't render correctly in Sandbox, as if all UVs are (0, 0) (see screenshots). The problem asset renders correctly in three.js based viewers, Khronos' sample viewer and blender.

**Screenshots**

no_draco
draco

 - Windows 10
 - Chrome 96.0.4664.110

**Additional context**

Files to repro with:
[test.zip](https://github.com/BabylonJS/Babylon.js/files/7757387/test.zip)

RaananW:
Fixes #11685

Pass optional dividers in case the attributes are normalized, based on their types.

The chosen architecture is set this way because dracoCompression is part of Core and not the loaders and therefore has access only to primitives passed to it by the loader. Since dracoCompression is the component responsible for setting the vertex data there is no way out of normalizing the attributes there.

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -100,6 +100,7 @@
 - Added an observable for when loader state changed. ([bghgary](https://github.com/bghgary))
 - Fixed an issue where errors for loading certain assets (e.g. <20-byte GLBs) are not catchable. ([bghgary](https://github.com/bghgary))
 - Added support for `KHR_materials_emissive_strength` for glTF loader. ([sebavan](https://github.com/sebavan))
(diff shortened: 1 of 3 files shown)


RaananW:
Result of the compressed, normalized mesh:

image

deltakosh (approved):

PR #11779, opened 2022-01-14
Implementing near interaction for motion controllers

### LINKED ISSUE #10493: [XR] Make sure all hand-interactions are abstracted to support other interactions
Hand interactions like near interaction and other UI-oriented elements are tightly coupled to the hand API.
As they are mostly relay on the position of the finger for interaction, it seems like the interaction can be abstracted to support any type of mesh as the interaction-mesh.

This should allow XR controllers and other elements that are attached to them to be used as the near-interaction enablers.

(@CraigFeldspar @rickfromwork)

rickfromwork:
This change updates the near interaction logic to work with motion controllers in addition to the existing hand joint functionality. An orb is added to the front of the motion controller to dictate the touch point, and this orb comes into view only when near an object that can be interacted with.

Also added an enum to determine how near interaction is handled for controllers - Near interaction is either disabled, enabled at the center of the controller, or enabled in front of the controller along the raycast line for far interaction (enabled by default).

Resolves #10493

https://user-images.githubusercontent.com/70608380/151047413-6e60967a-c98f-4b50-b4e8-b7e01d09b9a0.mp4

### DIFF
--- a/src/XR/features/WebXRNearInteraction.ts
+++ b/src/XR/features/WebXRNearInteraction.ts
@@ -20,6 +20,7 @@ import { TransformNode } from "../../Meshes/transformNode";
 import { StandardMaterial } from "../../Materials/standardMaterial";
 import { Color3 } from "../../Maths/math.color";
 import { NodeMaterial } from "../../Materials/Node/nodeMaterial";
+import { Material } from "../../Materials/Material";
 import { Animation } from "../../Animations/animation";
 import { QuadraticEase, EasingFunction } from "../../Animations/easing";
 // side effects
@@ -120,6 +121,11 @@ export interface IWebXRNearInteractionOptions {
      * Near interaction mode for motion controllers
      */
     nearInteractionControllerMode?: WebXRNearControllerMode;
+
+    /**
+     * Optional material for the motion controller orb, if enabled
+     */
+    motionControllerOrbMaterial?: Material;
 }
 
 /**
@@ -347,6 +353,7 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
             return;
         }
 
+        // Don't always break to allow for animation fallthrough on rare cases of multi-transitions
         if (newState > controllerData.currentAnimationState) {
             switch(controllerData.currentAnimationState) {
                 case ControllerOrbAnimationState.DEHYDRATED: {
@@ -386,25 +393,15 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
     private readonly _hoverRadius = 0.1;
     private readonly _pickRadius = 0.02;
     private readonly _nearGrabLengthScale = 5;
-    private _touchPointQuaternion = new Quaternion();
-    private _touchPointOrientationVector = Vector3.Zero();
 
-    private _processTouchPoint(id: string, transform: XRRigidTransform) {
+    private _processTouchPoint(id: string, position: Vector3, orientation: Quaternion) {
         const controllerData = this._controllers[id];
 
-        let axisRHSMultiplier = this._scene.useRightHandedSystem ? 1 : -1;
-        let touchPointPosition = transform.position;
-        const touchPointOrientation = transform.orientation;
-        this._touchPointQuaternion.set(
-            touchPointOrientation.x,
-            touchPointOrientation.y,
-            touchPointOrientation.z * axisRHSMultiplier,
-            touchPointOrientation.w * axisRHSMultiplier
-        );
+        // Position and orientation could be temporary values, se we take care of them before calling any functions that use temporary vectors/quaternions
+        controllerData.grabRay.origin.copyFrom(position);
+        orientation.toEulerAnglesToRef(TmpVectors.Vector3[0]);
+        controllerData.grabRay.direction.copyFrom(TmpVectors.Vector3[0]);
 
-        // set near interaction grab ray parameters
-        const nearGrabRayLength = this._nearGrabLengthScale * this._hoverRadius;
-        controllerData.grabRay.origin.set(touchPointPosition.x, touchPointPosition.y, touchPointPosition.z * axisRHSMultiplier);
 
         if (this._options.nearInteractionControllerMode === WebXRNearControllerMode.CENTERED_IN_FRONT && !(controllerData.xrController?.inputSource.hand)) {
             // offset the touch point in the direction the transform is facing
@@ -412,11 +409,7 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
             controllerData.grabRay.origin.addInPlace(this._tmpRay.direction.scale(0.05));
         }
 
-        this._touchPointQuaternion.toEulerAnglesToRef(this._touchPointOrientationVector);
-        controllerData.grabRay.direction.set(this._touchPointOrientationVector.x, this._touchPointOrientationVector.y, this._touchPointOrientationVector.z);
-        controllerData.grabRay.length = nearGrabRayLength;
-
-        // set positions for near pick and hover
+        controllerData.grabRay.length = this._nearGrabLengthScale * this._hoverRadius;
         controllerData.touchCollisionMesh.position.copyFrom(controllerData.grabRay.origin);
     }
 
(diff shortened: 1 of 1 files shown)


RaananW:
LGTM, will wait for @PatrickRyanMS to comment as well

PatrickRyanMS:
I think the disconnect I have here is that the sphere looks like a physical rubber ball hovering in front of the controller mesh here, which does not feel aligned with the mental model of the controller. I think if it felt more like an energy sphere, the mental leap to associating it with the controller mesh wouldn't be as large. Something that is not lit by the scene lighting which is part of what is making it feel like a physical object hovering in front of the controller mesh. 

To that end, I put together this node material and simple animation to explain what I was thinking. https://playground.babylonjs.com/#NY7TA0#2 The animation for colliding with an interactive object is for the sphere to grow with a slight bounce at the end. The animation in the above example showing the sphere scaling to 0 is a little hard to understand as it feels like touching the interactive element either deflates or destroys the sphere. The slight enlargement is meant to mimic what happens to a balloon when it is pressed against a solid object... it gets slightly larger while the side touching the object and the opposite side collapse a little. We don't have this type of compression here because we don't know what angle the user is touching the interactive element from so the slight enlargement will make the user feel as though they are pressing the sphere into the object. 

The shader does require alpha blending, but also fakes in a glow to make it feel like light/energy. I did experiment with using an add blend mode rather than the combine blend mode, but when it interacts with a very bright object, you won't be able to see the sphere. I tried to use a brighter value blending to a darker value in color to help with readability over various values of color.

rickfromwork:
> I think the disconnect I have here is that the sphere looks like a physical rubber ball hovering in front of the controller mesh here, which does not feel aligned with the mental model of the controller. I think if it felt more like an energy sphere, the mental leap to associating it with the controller mesh wouldn't be as large. Something that is not lit by the scene lighting which is part of what is making it feel like a physical object hovering in front of the controller mesh.
> 
> To that end, I put together this node material and simple animation to explain what I was thinking. https://playground.babylonjs.com/#NY7TA0#2 The animation for colliding with an interactive object is for the sphere to grow with a slight bounce at the end. The animation in the above example showing the sphere scaling to 0 is a little hard to understand as it feels like touching the interactive element either deflates or destroys the sphere. The slight enlargement is meant to mimic what happens to a balloon when it is pressed against a solid object... it gets slightly larger while the side touching the object and the opposite side collapse a little. We don't have this type of compression here because we don't know what angle the user is touching the interactive element from so the slight enlargement will make the user feel as though they are pressing the sphere into the object.
> 
> The shader does require alpha blending, but also fakes in a glow to make it feel like light/energy. I did experiment with using an add blend mode rather than the combine blend mode, but when it interacts with a very bright object, you won't be able to see the sphere. I tried to use a brighter value blending to a darker value in color to help with readability over various values of color.

I can go and implement something like this, but I do have a concern about resizing the orb (though this also applies to the current version). Since the orb represents the interaction point, I'm a little worried that there will be users that think the surface of the orb is the contact point for near touch interactions, and so a resize on contact might go against what they expect.

The alternative is to not update the size ever though, which makes it harder to distinguish between an active interaction and no interaction. What are your thoughts on having a color change instead of a scaling change based on the interaction? Though that also has the downside of not being as accessible to those that are colorblind.

PatrickRyanMS:
The accessibility aspect of only a color change is the main concern here. I think we can make the size change do a couple of things. If you need to know about the proximity of the touch point to the interactable, we can either fade in or scale up the sphere from invisible. Then on a true "hover" as the touch point is about to press the interactable, we scale up a bit more. Then on press, we get the larger scale. So there may be three sizes, "invisible", "hover", and "press". We can supplement the size interaction with a color change as well to reinforce the states, but we definitely need a silhouette change to signify the states in an accessible way.

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -636,30 +757,84 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
-            resultMesh = CreateSphere(name, { diameter: 1 }, sceneToUse);
-            resultMesh.scaling.set(scale, scale, scale);
-            resultMesh.isVisible = false;
+        let touchCollisionMesh = CreateSphere("PickSphere", { diameter: 1 }, meshCreationScene);
+        touchCollisionMesh.scaling.set(this._pickRadius, this._pickRadius, this._pickRadius);
+        touchCollisionMesh.isVisible = false;
 
-            return resultMesh;
+        // Generate the material for the touch mesh visuals
+        NodeMaterial.ParseFromSnippetAsync("8RUNKL#3", meshCreationScene).then(nodeMaterial => {
```
RaananW: We need to allow overriding this, as this should also work offline as well.

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -636,30 +757,84 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
+        let dehydrateKeys = [
+            {frame:  0, value: new Vector3(0.03, 0.03, 0.03)},
+            {frame: 10, value: new Vector3(0.0, 0.0, 0.0)},
+            {frame: 15, value: new Vector3(0.0, 0.0, 0.0)}
+        ];
+
+        let touchAction = new Animation("touch", "scaling", 60, Animation.ANIMATIONTYPE_VECTOR3, Animation.ANIMATIONLOOPMODE_CONSTANT);
+        let releaseAction = new Animation("release", "scaling", 60, Animation.ANIMATIONTYPE_VECTOR3, Animation.ANIMATIONLOOPMODE_CONSTANT);
+        let hydrateAction = new Animation("hydrate", "scaling", 60, Animation.ANIMATIONTYPE_VECTOR3, Animation.ANIMATIONLOOPMODE_CONSTANT);
+        let dehydrateAction = new Animation("dehydrate", "scaling", 60, Animation.ANIMATIONTYPE_VECTOR3, Animation.ANIMATIONLOOPMODE_CONSTANT);
```
RaananW, line 811: nit - could all be consts

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -435,6 +545,17 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
+            // Update the interaction animation. Only updates if the visible touch mesh is active
+            if (controllerData.grabInteraction || controllerData.nearInteraction) {
+                this._handleTransitionAnimation(controllerData, ControllerOrbAnimationState.TOUCH);
+            }
+            else if (controllerData.hoverInteraction) {
+                this._handleTransitionAnimation(controllerData, ControllerOrbAnimationState.HYDRATED);
+            }
+            else {
+                this._handleTransitionAnimation(controllerData, ControllerOrbAnimationState.DEHYDRATED);
+            }
```
RaananW:
```suggestion
            let state = ControllerOrbAnimationState.DEHYDRATED;
            if (controllerData.grabInteraction || controllerData.nearInteraction) {
                state  = ControllerOrbAnimationState.TOUCH;
            } else if (controllerData.hoverInteraction) {
                state  = ControllerOrbAnimationState.HYDRATED;
            }
            this._handleTransitionAnimation(controllerData, state);
```

RaananW: just a suggestion, of course :-)

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -317,27 +431,23 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
-                            controllerData.grabRay.length = nearGrabRayLength;
+                            this._processTouchPoint(id, indexTipPose.transform);
+                        }
+                    }
+                }
+                else if (controllerData.xrController.inputSource.gamepad) {
+                    const controller = controllerData.xrController.inputSource;
+                    if (this._options.nearInteractionControllerMode === WebXRNearControllerMode.CENTERED_ON_CONTROLLER) {
+                        // Try to get the grip space, with a target ray space fallback
+                        let controllerPose = _xrFrame.getPose(controller.gripSpace || controller.targetRaySpace, this._xrSessionManager.referenceSpace);
```
RaananW, line 453: don't we have the pose in the grip/pointer abstract meshes?

src/XR/features/WebXRNearInteraction.ts
```diff
@@ -317,27 +431,23 @@ export class WebXRNearInteraction extends WebXRAbstractFeature {
...
+                        if (controllerPose && controllerPose.transform) {
+                            this._processTouchPoint(id, controllerPose.transform);
+                        }
+                    }
+                    else if (this._options.nearInteractionControllerMode === WebXRNearControllerMode.CENTERED_IN_FRONT) {
+                        let controllerPose = _xrFrame.getPose(controller.targetRaySpace, this._xrSessionManager.referenceSpace);
+                        if (controllerPose && controllerPose.transform) {
+                            this._processTouchPoint(id, controllerPose.transform);
```
RaananW, line 461: The only difference here is the xr space used, you could save a few lines removing the repeated code. Unless i totally missed something :-)

RaananW (commented):
It feels like we already calculating certain things (like the grip and pointer locations) that you are recalculating in the feature. can we save the grab-ray calculation as well?

RaananW (approved):

PR #11781, opened 2022-01-14
Low pass filter perf profiler

carolhmj:


inspector/src/components/graph/canvasGraphService.ts
```diff
@@ -356,25 +366,36 @@ export class CanvasGraphService {
                 const timestamp = this.datasets.data.at(this.datasets.startingIndices.at(pointIndex));
                 const value = this.datasets.data.at(valueIndex);
 
+                if (!prevValue) {
```
RaananW, line 369: can it be 0? is 0 a valid value?

RaananW (commented):
LGTM! just a small question

inspector/src/components/graph/canvasGraphService.ts
carolhmj: It is a valid value, thanks for catching that!

carolhmj: Updated!

RaananW (approved):

PR #11783, opened 2022-01-14
PG: Start render loop before creating scene

BlakeOne:
PR for playground to start the default render loop before calling the scene creation function, so that we can stop it and start a new one without having to use setTimeout() to wait for the default to be started.

Basically the current code to start the render loop is moved into a hidden public method which is called after creating the engine, right before calling the user function that creates the scene.

Forum discussion: https://forum.babylonjs.com/t/make-it-easier-to-stop-default-render-loop-on-playground/26749/2

### DIFF
--- a/Playground/src/components/rendererComponent.tsx
+++ b/Playground/src/components/rendererComponent.tsx
@@ -100,6 +100,32 @@ export class RenderingComponent extends React.Component<IRenderingComponentProps
         });
     }
 
+    /** @hidden */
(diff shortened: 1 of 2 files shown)


deltakosh (approved):

deltakosh:
Just making sure: can you test with the JS and the TS mode?

BlakeOne:
Yep I should have mentioned I tested it locally with JS and TS, using a simple animation, and it seems to be working well. :)

PR #11784, opened 2022-01-14
[XR] Ability to add a QuadLayer

RaananW:
This is step one in supporting fullscreen UI in XR.

It's now possible to create a quad layer, get its texture and render anything needed to it. It's technically the same as adding a plane with a texture, but accelerated by the underlying system. It also supports XR's reference spaces which make it more usable in different XR scenarios.

### DIFF
--- a/src/XR/features/WebXRLayers.ts
+++ b/src/XR/features/WebXRLayers.ts
@@ -69,7 +69,10 @@ class WebXRCompositionLayerRenderTargetTextureProvider extends WebXRLayerRenderT
         }
 
         this._lastSubImages.set(eye, subImage);
-
+        // don't skip clear for the first rtt. skip for all others
+        if (subImageIndex === 0) {
+            this._renderTargetTextures[subImageIndex].skipInitialClear = false;
+        }
         return this._renderTargetTextures[subImageIndex];
     }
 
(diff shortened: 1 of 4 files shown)


src/XR/webXRRenderTargetTextureProvider.ts
```diff
@@ -63,19 +63,22 @@ export abstract class WebXRLayerRenderTargetTextureProvider implements IWebXRRen
         // Create render target texture from the internal texture
         const renderTargetTexture = new RenderTargetTexture("XR renderTargetTexture", { width, height }, this._scene);
         const renderTargetWrapper = renderTargetTexture.renderTarget!;
-        (renderTargetWrapper as WebGLRenderTargetWrapper)._framebuffer = framebuffer;
+        // Set the framebuffer, make sure it works in all scenarios - emulator, no layers and layers
+        if (framebuffer || !colorTexture) {
```
rgerd, line 67: nit: Just curious, why is this necessary?

src/XR/features/WebXRLayers.ts
```diff
@@ -69,7 +69,10 @@ class WebXRCompositionLayerRenderTargetTextureProvider extends WebXRLayerRenderT
         }
 
         this._lastSubImages.set(eye, subImage);
-
+        // don't skip clear for the first rtt. skip for all others
+        if (subImageIndex === 0) {
```
rgerd, line 73: nit: Does the WebXR driver not already clear the texture during the call to `getSubImage`?
https://immersive-web.github.io/layers/#xropaquetextures

rgerd (approved):

src/XR/webXRRenderTargetTextureProvider.ts
RaananW: This adds support to non-layers support.

I made different adjustments to the clear state of certain textures, I will see if it is still needed.

src/XR/features/WebXRLayers.ts
RaananW: i'll test it again and write back here. If i remember correctly without that only one eye was rendering

deltakosh:
Any news on this little one? ;)

RaananW:
CLosing for now, will address in a future release.

PR #11809, opened 2022-01-20
Inspector autopicking is in sync with GimoManager autopicking

CedricGuillemet:
Follow up to a discussion with @PatrickRyanMS 
In the inspector, when a gizmo is active, autopicking is on in the gizmomanger even if not active in Inspector.
So, clicking to rotate camera will select a mesh. To fix that, gizmomanger autopicking (on by default) is in sync with Inspector.

### DIFF
--- a/src/Gizmos/gizmoManager.ts
+++ b/src/Gizmos/gizmoManager.ts
@@ -26,6 +26,9 @@ export class GizmoManager implements IDisposable {
     /** When true, the gizmo will be detached from the current object when a pointer down occurs with an empty picked mesh */
     public clearGizmoOnEmptyPointerEvent = false;
 
+    /** When true (default), picking to attach a new mesh is enabled. This works in sync with inspector autopicking. */
+    public enableAutoPicking = true;
+
     /** Fires an event when the manager is attached to a mesh */
     public onAttachedToMeshObservable = new Observable<Nullable<AbstractMesh>>();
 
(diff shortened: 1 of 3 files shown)


src/Gizmos/gizmoManager.ts
```diff
@@ -136,32 +139,34 @@ export class GizmoManager implements IDisposable {
...
-                    } else {
-                        // Attach to the parent node that is an attachableMesh
-                        var found = false;
-                        this.attachableMeshes.forEach((mesh) => {
-                            if (node && (node == mesh || node.isDescendantOf(mesh))) {
-                                node = mesh;
-                                found = true;
+                    if (this.enableAutoPicking) {
+                        var node: Nullable<Node> = pointerInfo.pickInfo.pickedMesh;
+                        if (this.attachableMeshes == null) {
```
RaananW, line 144:
```suggestion
                        if (this.attachableMeshes === null) { // or if (!this.attachableMeshes) {
```

RaananW: I assume you will also need to check if the array's length is 0?

src/Gizmos/gizmoManager.ts
```diff
@@ -136,32 +139,34 @@ export class GizmoManager implements IDisposable {
...
+                            // Attach to the most parent node
+                            while (node && node.parent != null) {
+                                node = node.parent;
+                            }
+                        } else {
+                            // Attach to the parent node that is an attachableMesh
+                            var found = false;
+                            this.attachableMeshes.forEach((mesh) => {
```
RaananW, line 152: you can use `.some` instead of forEach, and break when you found it. the variable `found` will get the result of the `.some` call.

src/Gizmos/gizmoManager.ts
```diff
@@ -136,32 +139,34 @@ export class GizmoManager implements IDisposable {
...
+                    if (this.enableAutoPicking) {
+                        var node: Nullable<Node> = pointerInfo.pickInfo.pickedMesh;
+                        if (this.attachableMeshes == null) {
+                            // Attach to the most parent node
+                            while (node && node.parent != null) {
+                                node = node.parent;
+                            }
+                        } else {
+                            // Attach to the parent node that is an attachableMesh
+                            var found = false;
```
RaananW, line 151: nit `let`

src/Gizmos/gizmoManager.ts
CedricGuillemet: This code has only been indented for the if. I don't want to do any change that may cause more problem.

RaananW (approved):

PR #11812, opened 2022-01-20
DeviceInputSystem: Adjust order of operations on pointerdown

PolygonalSun:
When a `pointerdown` event was being addressed, changes to X and Y were being handled before the actual button press so move events were being fired before the button events.  This PR changes that order to handle the button first.  This is to address an issue with the ArcRotateCamera that was causing an unnecessary rotation before allowing for pan/pinch actions.

### DIFF
--- a/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
+++ b/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
@@ -482,9 +482,12 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 deviceEvent.deviceType = deviceType;
                 deviceEvent.deviceSlot = deviceSlot;
 
+                // NOTE: The +2 used here to is because PointerInput has the same value progression for its mouse buttons as PointerEvent.button
+                // However, we have our X and Y values front-loaded to group together the touch inputs but not break this progression
+                // EG. ([X, Y, Left-click], Middle-click, etc...)
                 deviceEvent.inputIndex = evt.button + 2;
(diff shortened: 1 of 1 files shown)


src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -482,6 +482,11 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 deviceEvent.deviceType = deviceType;
                 deviceEvent.deviceSlot = deviceSlot;
 
+                deviceEvent.inputIndex = evt.button + 2;
```
RaananW, line 485: please add a comment explaining why the +2 reference

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -482,6 +482,11 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
                 deviceEvent.deviceType = deviceType;
                 deviceEvent.deviceSlot = deviceSlot;
 
+                deviceEvent.inputIndex = evt.button + 2;
+                deviceEvent.previousState = previousButton;
+                deviceEvent.currentState = pointer[evt.button + 2];
```
RaananW, line 487: you could use `deviceEvent.inputIndex` instead of adding 2 again

PolygonalSun: Added for Up and Down

RaananW (approved):

PR #11837, opened 2022-01-24
Added load texture2D array capability to native engine

SergioRZMasson:
Added the capability for the native engine to handle texture2D arrays in order to enable morph targets to use texture arrays on Babylon Native.

### DIFF
--- a/src/Engines/Native/nativeInterfaces.ts
+++ b/src/Engines/Native/nativeInterfaces.ts
@@ -27,7 +27,7 @@ export interface INativeEngine {
     createTexture(): WebGLTexture;
     loadTexture(texture: WebGLTexture, data: ArrayBufferView, generateMips: boolean, invertY: boolean, srgb: boolean, onSuccess: () => void, onError: () => void): void;
     loadRawTexture(texture: WebGLTexture, data: ArrayBufferView, width: number, height: number, format: number, generateMips: boolean, invertY: boolean): void;
-    loadRawTexture2DArray(texture: WebGLTexture, data: Nullable<ArrayBufferView>, width: number, height: number, depth: number, format: number, generateMipMaps: boolean, invertY: boolean, samplingMode: number, compression: Nullable<string>, textureType : number) : void;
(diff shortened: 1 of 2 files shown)


RaananW (approved):

RaananW:
A small linting issue needs to be fixed

src/Engines/Native/nativeInterfaces.ts
```diff
@@ -27,6 +27,7 @@ export interface INativeEngine {
     createTexture(): WebGLTexture;
     loadTexture(texture: WebGLTexture, data: ArrayBufferView, generateMips: boolean, invertY: boolean, srgb: boolean, onSuccess: () => void, onError: () => void): void;
     loadRawTexture(texture: WebGLTexture, data: ArrayBufferView, width: number, height: number, format: number, generateMips: boolean, invertY: boolean): void;
+    loadRawTexture2DArray(texture: WebGLTexture, data: Nullable<ArrayBufferView>, width: number, height: number, depth: number, format: number, generateMipMaps: boolean, invertY: boolean, samplingMode: number, compression: Nullable<string>, textureType : number) : void;
```
RaananW, line 30: did you mean the name this function `loadRawTexture2DArray` and the implemented `createRawTexture2DArray` ?

SergioRZMasson: The native engine interface method is called loadRawTexture2DArray because it is only loading data to the object that is create on the JS side. The NativeEngine method is called createRawTexture2DArray because it needs to override the method from ThinEngine and in will actually create the texture object.

src/Engines/nativeEngine.ts
```diff
@@ -2029,6 +2029,35 @@ export class NativeEngine extends Engine {
...
+        texture.width = texture.baseWidth;
+        texture.height = texture.baseHeight;
+        texture._compression = compression;
+        texture.type = textureType;
+        texture.is2DArray = true;
+        texture.is3D = false;
+        texture.depth = depth;
+
+        if (texture._hardwareTexture) 
+        {
```
RaananW, line 2050:
```suggestion
        if (texture._hardwareTexture) {
```

PR #11862, opened 2022-01-27
Users/jypandey/user errors

jp833:
Throw custom errors for some end user scenarios we are seeing on our React native App.

1. Invalid texture format
2. Missing positions for mesh
3. Unexpected magic number found when loading file
4. Adds a custom SceneLoaderError with an inner exception with the root cause


RaananW:
LGTM! 
@bghgary - the BaseError is marked as error (by you, according to github :-)). Is that ok to export it per default now?

bghgary:
> LGTM! @bghgary - the BaseError is marked as "ignore" (by you, according to github :-)). Is that ok to export it per default now?

Yes, I don't see why not, but I think I copied and pasted the comment for it from the FileError and forgot to update it. We should probably update the comment so that the docs are good.

loaders/src/glTF/errors.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseError } from "babylonjs/Misc/baseError";
+
+/**
+ * Application error when unexpected magic number is found when loading file.
+ */
+export class GlTFLoaderUnexpectedMagicError extends BaseError {
```
bghgary, line 6: Do we really want such a specific error for this? Can it be more generic like `GLTFLoaderError`?

Also, small nit
```suggestion
export class GLTFLoaderUnexpectedMagicError extends BaseError {
```

jp833: The intention here was to get the specific error, localize it and show to end user (of our app) so that they can self-recover. I believe this unexpected magic error is fixable by user.

bghgary: Hmm, okay. There are lots of error from the glTF loader that can be fixed by the user. Are we going to eventually add a custom error to each one? Maybe we can use an error code instead?

jp833: Yes error codes make more sense. I believe we'd have to eventually officially document those. What is the guidance there? Check in the code first and followup with doc team?

bghgary: Ideally merge the code then follow up with an update to the docs. We can start with just one error code for now if this is all you need.

src/Engines/nativeEngine.ts
```diff
@@ -30,6 +30,7 @@ import { IStencilState } from "../States/IStencilState";
 import { RenderTargetWrapper } from "./renderTargetWrapper";
 import { NativeData, NativeDataStream } from "./Native/nativeDataStream";
 import { INative, INativeCamera, INativeEngine } from "./Native/nativeInterfaces";
+import { BaseError } from "..";
```
Alex-MSFT, line 33: Should this be ../Misc/baseError?

src/Loading/errors.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Application error to indicate load scene failures.
+ */
+export class SceneLoaderError extends BaseError {
+    /**
+     * Creates a new SceneLoaderError
+     * @param message defines the message of the error
+     * @param innerException the exception that caused the outer exception
+     */
+    constructor(message: string, innerException?: Error) {
+        super(message, undefined, innerException);
```
Alex-MSFT, line 13: Why no error code for this category of errors?

jp833: This is supposed to be an outer error for what I have currently added. But on second thought, it would be good to have code for this.

src/Engines/nativeEngine.ts
RaananW: Yes, this needs to be changed

src/Misc/errorCodes.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    // Material validation
+    UnsupportedTextureError: "UnsupportedTextureError",
+
+    // GLTF Loader errors
+    GLTFLoaderUnexpectedMagicError: "GLTFLoaderUnexpectedMagicError",
+
+    // Scene Load errors
+    SceneLoaderError: "SceneLoaderError",
+});
```
bghgary, line 16: Why use strings for error codes?

jp833: Either is fine with me. String error codes are more flexible and explanatory so I used it. But if there is a best practice we should be following here I can make the change here. Or if there is any preferred way in open source projects for user-facing error codes.

bghgary: I don't know if there is a best practice. @RaananW @sebavan @deltakosh any thoughts?

RaananW: I would rather use a const enum, which is technically almost the same solution. I don't see a reason to use strings, as the variable name is of more important when handling these errors. You can also use a "as const" object:

```
const ErrorCodes = {
  FirstError: "FirstError",
  SecondError: "SecondError"
} as const
```
which is typescript's recommendation when complex enum features are not needed - https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

src/Misc/baseError.ts
```diff
@@ -1,12 +1,46 @@
+import { ErrorCodesString } from "./errorCodes";
+
 /**
+ * BabylonJS error internal class wrapper. Extending Error for a class with docs
+ * fails typedoc validation, because lib.es5.d.ts does not have any docs for Error.
  * @ignore
- * Application error to support additional information when loading a file
  */
-export abstract class BaseError extends Error {
+class BabylonJSErrorWrapper extends Error {}
```
jp833, line 8: typedoc-validate fails if the class extending error is not marked @ignore. Added an internal error wrapper class as workaround.

bghgary:
@jp833 I broke the build yesterday. Make sure you merge master to get the CI working again. Sorry about this.

RaananW:
@bghgary - do you want to review and approve this?

src/Misc/baseError.ts
bghgary: I'm not sure I understand this. Error is an interface. Typedoc is complaining about the interface not having doc comments, but if we add a class inheritance layer then it works?

Also, isn't this going to introduce another prototype chain? Does that cause any issues?

src/Loading/errors.ts
```diff
@@ -0,0 +1,19 @@
+import { BaseError } from "../Misc/baseError";
+import { ErrorCodes } from "../Misc/errorCodes";
+
+/**
+ * Application error to indicate load scene failures.
+ */
+export class SceneLoaderError extends BaseError {
```
bghgary, line 7: This seems inconsistent with other usages. Why does only SceneLoaderError have a special derived error but other classes (like glTF loader) don't? I see it auto sets the SceneLoaderError code, but I'm not sure on the error hierarchy.

src/Misc/baseError.ts
```diff
@@ -1,12 +1,46 @@
...
  * @ignore
- * Application error to support additional information when loading a file
  */
-export abstract class BaseError extends Error {
+class BabylonJSErrorWrapper extends Error {}
+
+/**
+ * Base Error class for babylonjs errors
+ */
+export class BaseError extends BabylonJSErrorWrapper {
```
bghgary, line 13: I find it odd that we are exposing/using BaseError for errors. This class is intended to be an abstract base class.

bghgary (commented):
Sorry for all the comments, but this is exposing things that we will not be able to change easily later due to backcompat and thus we should try to get it right the first time.

RaananW (approved):

src/Loading/errors.ts
jp833: SceneLoader would also wrap an inner exception unlike the gltf loader error. The innerEx would have the actual error code. The error hierarchy does get confusing :/. Totally makes sense to remove this class since we would be also checking for the inner exception. In which case, BaseError would wrap a BaseError/Error.

bghgary: Having an additional layer here would allow people to do something like `if (error instanceof SceneLoaderError)` but because there is already an error code with categories, it's not really useful anymore. I think the general pattern should be to introduce a new hierarchy only if there are additional properties.

src/Misc/baseError.ts
```diff
@@ -1,12 +1,46 @@
...
+
+    /**
+     * The error code
+     */
+    public errorCode?: ErrorCodesType;
+
+    /**
+     * The exception that caused the outer exception
+     */
+    public innerException?: Error;
```
bghgary, line 29: Adding these to `baseError` is also a bit weird for `LoadFileError` and `RequestFileError` since they both derive from this class but don't have `errorCode` or `innerException`.

jp833:
@[bghgary](https://github.com/bghgary) bghgary [4 days ago](https://github.com/BabylonJS/Babylon.js/pull/11862#discussion_r804213879)
I'm not sure I understand this. Error is an interface. Typedoc is complaining about the interface not having doc comments, but if we add a class inheritance layer then it works?

Also, isn't this going to introduce another prototype chain? Does that cause any issues?


The reason introduced this was because we wanted to not @ignore BaseError. This is causing typedoc to fail. I am inclining towards keeping BaseError as is and adding a non-abstract class like BabylonJSError extends BaseError and use it everywhere instead.

src/Misc/baseError.ts
```diff
@@ -1,46 +1,66 @@
...
+ * @ignore
  */
-export abstract class BaseError extends BabylonJSErrorWrapper {
-    // See https://stackoverflow.com/questions/12915412/how-do-i-extend-a-host-object-e-g-error-in-typescript
-    // and https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
-
-    // Polyfill for Object.setPrototypeOf if necessary.
-    protected static _setPrototypeOf: (o: any, proto: object | null) => any =
-        (Object as any).setPrototypeOf || ((o, proto) => { o.__proto__ = proto; return o; });
+abstract class BaseError extends Error {}
```
bghgary, line 18: If this is staying, might as well put the `setPrototypeOf` back here.

src/Misc/baseError.ts
```diff
@@ -1,46 +1,66 @@
...
+        this.errorCode = errorCode;
+        this.name = "RuntimeError";
+        setPrototypeOf(this, RuntimeError.prototype);
+    }
+}
+
+/**
+ * Application runtime outer error that wraps the cause/innerError
+ */
+export class RuntimeWrapperError extends RuntimeError {
```
bghgary, line 46: I know we went back and forth a bit on this. What do you think about putting `innerError` as an optional parameter of `RuntimeError` and then you can remove `RuntimeWrapperError`?

src/Misc/baseError.ts
```diff
@@ -1,46 +1,66 @@
 import { ErrorCodesType } from "./errorCodes";
```
bghgary, line 1: baseError.ts should probably be renamed to runtimeError.ts or maybe just error.ts.

Does `errorCodes.ts` still need to be a separate file? Could that just go here?

src/Misc/baseError.ts
```diff
@@ -1,46 +1,66 @@
...
 
+/**
+ * Application runtime error
+ */
+ export class RuntimeError extends BaseError {
     /**
      * The error code
      */
-    public errorCode: ErrorCodesType;
+     public errorCode: ErrorCodesType;
```
bghgary, line 27: nit: extra space
```suggestion
    public errorCode: ErrorCodesType;
```

dist/preview release/what's new.md
```diff
@@ -72,6 +72,7 @@
 - Modified touch in `WebDeviceInputSystem` to no longer delete touch points after pointer up. ([PolygonalSun](https://github.com/PolygonalSun))
 - Added support for DualSense controllers to DeviceInputSystem. ([PolygonalSun](https://github.com/PolygonalSun))
 - Added `getCreationOptions` on `ThinEngine`. ([carolhmj](https://github.com/carolhmj))
+- Added errorCodes for BaseError ([jp833](https://github.com/jp833))
```
bghgary, line 75: Comment probably should be updated :-)

src/Misc/baseError.ts
jp833: I put it out so we can remove BaseError to be exposed alltogether. So that when we fix the Error documentation issue we can remove it cleanly. Lemme know if you feel strongly about this.

src/Misc/baseError.ts
jp833: Makes sense.

src/Misc/baseError.ts
jp833: Reverted it back. When we resolve the Error documentation issue may be do this at that time.

bghgary (approved):
LGTM except for one nit. Not for this PR necessarily, but do we want any unit tests for these?

src/Misc/error.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Error code type
+ */
+export type ErrorCodesType = typeof ErrorCodes[keyof typeof ErrorCodes];
+
+/**
+ * Application runtime error
+ */
+ export class RuntimeError extends BaseError {
```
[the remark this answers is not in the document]
jp833: oh. Anytime I paste it is inserting a space. :-o
Fixed this.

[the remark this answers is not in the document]
bghgary: Yeah, I notice this too. It's annoying.

PR #11864, opened 2022-01-27
Weaken restriction on Texture Types

jtwigg:
The texture allowed for layers was constrained to only allow `Texture` assignments. However `BaseTexture` would have sufficed. 

Motivation was driven by getting the  `export class AnimatedGifTexture extends BaseTexture` to work with a Layer.

Forum Reference: https://forum.babylonjs.com/t/make-a-background-with-basetexture/27116/3?u=asdf989

Testing I've done:...
Not amazing. I force cast my `AnimatedGifTexture` into the `layer.texture = animatedGif as unknown as Texture` and it worked pretty well thank you very much.

https://github.com/sebavan/BabylonjsAnimatedGifSample


RaananW (approved):

RaananW:
Thanks a lot! makes sense :-)

PR #11872, opened 2022-01-27
[GE] Rendering rearchitecture

darraghjburke:
Significant refactor of the GUI editor's rendering architecture. In the old setup, we place a plane with a GUI texture and a camera, and then zoomed/panned using the camera.

In this new architecture, we render the GUI in full screen. To zoom in, we increase or decrease the scale of the root container. To pan, we move the root container. This unlocks a powerful new ability: controls can now be dragging outside the bounds of the GUI.

The main changes in this PR:
- Replaced textured mesh with fullscreen GUI
- Rewrote pan and zoom functionality
- Moved coordinate projection code into a separate coordinateHelper
- Created an artboard class that draws the artboard, both checkerboard and stroke, as HTML element rather than using a shader

One thing that's a bit unfinished: we currently show all the (invisible) containers in the tree view. This will be removed in the beta, but it's there for now for debugging purposes. (maybe we can hide it behind a debug flag later).

This should be the last giant PR for GUI editor😅 Going forward, everything else should be small feature additions and improvements. (but no promises)


gui/src/2D/controls/container.ts
```diff
@@ -516,6 +517,10 @@ export class Container extends Control {
             }
         }
 
+        if (!super.contains(x, y)) {
```
RaananW, line 520: instead of calling "contains" twice, store it in a variable and use the same variable twice. save a few CPU cycles :-)

guiEditor/src/components/propertyTab/propertyTabComponent.tsx
```diff
@@ -123,7 +122,7 @@ export class PropertyTabComponent extends React.Component<IPropertyTabComponentP
             }
         });
         this.props.globalState.onResizeObservable.add((newSize) => {
-            this.props.globalState.workbench.artBoardBackground._markAsDirty(true);
+            // this.props.globalState.workbench.artBoardBackground._markAsDirty(true);
```
RaananW, line 125: nit - cleanup

guiEditor/src/components/sceneExplorer/treeItemSelectableComponent.tsx
```diff
@@ -122,6 +122,9 @@ export class TreeItemSelectableComponent extends React.Component<
     }
 
     render() {
+        // if (this.props.entity === this.props.globalState.workbench.trueRootContainer || this.props.entity === this.props.globalState.workbench.visibleRegionContainer) {
+        //     return this.renderChildren(true, false);
+        // }
```
RaananW, line 127: nit - cleanup

RaananW (commented):
LGTM!

Just a few nit-picky comments and 👍

RaananW:
There is also a build issue regarding the `trueRootContainer` variable.

RaananW (approved):

PR #11874, opened 2022-01-28
Parent find by (empty) id

CedricGuillemet:
Follow up on https://forum.babylonjs.com/t/bonelookcontroller-and-bone-orientation/27042/10

When loading a .babylon scene, there is a parent attachment per id. But when the scene already has a mesh with an empty name, the first mesh in the .babylon is attached to it. I fixed it by checking parent name is not empty and is not zero length.

### DIFF
--- a/src/Loading/Plugins/babylonFileLoader.ts
+++ b/src/Loading/Plugins/babylonFileLoader.ts
@@ -736,7 +736,7 @@ SceneLoader.RegisterPlugin({
                 var currentMesh: AbstractMesh;
                 for (index = 0, cache = scene.meshes.length; index < cache; index++) {
                     currentMesh = scene.meshes[index];
-                    if (currentMesh._waitingParentId !== null && currentMesh._waitingParentId.length) {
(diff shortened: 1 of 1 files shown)


src/Loading/Plugins/babylonFileLoader.ts
```diff
@@ -736,7 +736,7 @@ SceneLoader.RegisterPlugin({
                 var currentMesh: AbstractMesh;
                 for (index = 0, cache = scene.meshes.length; index < cache; index++) {
                     currentMesh = scene.meshes[index];
-                    if (currentMesh._waitingParentId !== null) {
+                    if (currentMesh._waitingParentId !== null && currentMesh._waitingParentId.length) {
```
RaananW, line 739: if it is a string, why not simply do:

```suggestion
                    if (currentMesh._waitingParentId) {
```
Will return false for empty strings

CedricGuillemet: Tested and approved!

RaananW (approved):

PR #11883, opened 2022-01-31
fix matrix caching updte

deltakosh:


src/Maths/math.vector.ts
```diff
@@ -4613,7 +4613,7 @@ export class Matrix {
     public copyFrom(other: DeepImmutable<Matrix>): Matrix {
         other.copyToArray(this._m);
         const o = (other as Matrix);
-        this.updateFlag = o.updateFlag;
+        this.updateFlag++;
```
sebavan, line 4616: this seems much safer !!!

sebavan (approved):

src/Maths/math.vector.ts
bghgary: Hmm, I'm not sure. At a minimum, it should be this, no?
```suggestion
        this.updateFlag = Matrix._updateFlagSeed++;;
```

But then, this will cause all the caching to not work. Isn't the whole point of copying the flag to avoid updating if they are the same?

deltakosh: yeah this is a good point.

deltakosh: Then we need to understand why there is an issue on the forum: https://forum.babylonjs.com/t/matrix-copyfrom-stopped-to-work-starting-alpha-48/27232/18

bghgary: I'll reply on the forum.

PR #11884, opened 2022-01-31
Fix spherical harmonics solid angle computation

Meakk:


src/Misc/HighDynamicRange/cubemapToSphericalPolynomial.ts
```diff
@@ -116,6 +116,10 @@ export class CubeMapToSphericalPolynomialTools {
         // The (u,v) of the first texel is half a texel from the corner (-1,-1).
         var minUV = du * 0.5 - 1.0;
 
+        // Compute the area on the unit sphere of the rectangle defined by (x,y) and the origin
+        // See https://www.rorydriscoll.com/2012/01/15/cubemap-texel-solid-angle/
+        var AreaElement = (x : number, y : number) => { return Math.atan2(x * y, Math.sqrt(x * x + y * y + 1)); };
```
sebavan, line 121: Would be great to move this into its own private function to prevent over allocation and garbage despite not being called frequently ?

src/Misc/HighDynamicRange/cubemapToSphericalPolynomial.ts
```diff
@@ -136,7 +140,7 @@ export class CubeMapToSphericalPolynomialTools {
                                 fileFace.worldAxisForNormal);
                     worldDirection.normalize();
 
-                    var deltaSolidAngle = Math.pow(1.0 + u * u + v * v, -3.0 / 2.0);
+                    var deltaSolidAngle = AreaElement(u - 0.5 * du, v - 0.5 * dv) - AreaElement(u - 0.5 * du, v + 0.5 * dv) - AreaElement(u + 0.5 * du, v - 0.5 * dv) + AreaElement(u + 0.5 * du, v + 0.5 * dv);
```
sebavan, line 143: could we extract u-0.5du and such to reuse the result as they appear several times in here ?

sebavan (changes requested):
Just a couple nits but it looks good to me !!!

src/Misc/HighDynamicRange/cubemapToSphericalPolynomial.ts
```diff
@@ -98,6 +98,14 @@ export class CubeMapToSphericalPolynomialTools {
         });
     }
 
+    /**
+     * Compute the area on the unit sphere of the rectangle defined by (x,y) and the origin
+     * See https://www.rorydriscoll.com/2012/01/15/cubemap-texel-solid-angle/
+     */
+    private static AreaElement(x : number, y : number): number {
```
Popov72, line 105: nit: should be _AreaElement because the function is private

PR #11899, opened 2022-02-02
update hoverRadius for fluentMaterial

tayomadein:
This PR updates the hoverRadius for fluent material so that the hover glow doesn't extend past the bounds of the active element.

Before:
flutentMaterialHoverRadiusBefore

After:
flutentMaterialHoverRadiusAfter


RaananW:
Will it also work correctly when you scale (down) a component?

PR #11903, opened 2022-02-02
Small fixes

RaananW:
Tests correction (strict mode)
Moving Engine.Last??? to EngineStore.Last???
Fixing two memory leaks when disposing a scene.


deltakosh (approved):

Popov72:
Shouldn't this be added in the "Breaking Change" section of the what's new file?

deltakosh:
We still have the Engine.LastXXX right?

RaananW:
> We still have the Engine.LastXXX right?

Yep, i haven't changed that

PR #11907, opened 2022-02-03
Fix DDS and KTX mipmap loading

Meakk:
See [this thread](https://forum.babylonjs.com/t/loading-a-cubetexture-with-cubemaps-produces-black-skybox-specular/27299) for more details.  

When a DDS or KTX file contains mipmaps levels, it is required to set the max level to ensure texture completeness, otherwise sampling will return 0 and Firefox emits a warning.  
See https://www.khronos.org/opengl/wiki/Texture#Mipmap_completeness:
> The [mipmap range](https://www.khronos.org/opengl/wiki/Texture#Mipmap_range) texture parameters are given values where GL_TEXTURE_BASE_LEVEL <= GL_TEXTURE_MAX_LEVEL.
The base and max levels must only specify mipmap levels that have been allocated, in accord with the above.


deltakosh (approved):

deltakosh:
@sebavan will push a PR to flag native and webgpu with false on the new parameter
@meakk do you mind doing an additional PR to store the maxLevel on the texture (this will be useful for the native and webgpu engine)

sebavan:
Follow up available here: https://github.com/BabylonJS/Babylon.js/pull/11912

PR #11910, opened 2022-02-03
Updated ThinEngine resize to optionally take device pixel ratio into account

travisvroman:
When creating an `Engine` object, one of the constructor's parameters is `adaptToDeviceRatio`, which properly sets `_hardwareScalingLevel` during creation. This part works well.

The issue arises if the user then zooms their browser. This triggers a call to `resize()` (as it should), but `resize()` no longer takes the device pixel ratio into account. This presents in poor aliasing the more you zoom out. Here is an example, zoomed in 200%:
image

One such workaround is to call `setHardwareScalingLevel` and pass the inverse of the window.devicePixelRatio (so something like `1.0 / (window.devicePixelRatio || 1)`, which then calls resize and accomplishes the same thing, but this feels somewhat clunky.

This solution adds a new private `_adaptToDeviceRatio` in order to maintain this setting. Then, in `resize()`, if this property is set to true, it performs the same calculations as in the constructor to re-query the device pixel ratio at that time and automatically assign it for that resize. This way the user can use the same API to resize as normal, and the `adaptToDeviceRatio` actually persists (as one might assume it does on creation).

Here is the same scene (still zoomed in to 200%), using this change:
image

### DIFF
--- a/src/Engines/thinEngine.ts
+++ b/src/Engines/thinEngine.ts
@@ -1663,7 +1663,7 @@ export class ThinEngine {
         let height: number;
 
         // Requery hardware scaling level to handle zoomed-in resizing.
-        if(this._adaptToDeviceRatio) {
+        if (this._adaptToDeviceRatio) {
             const devicePixelRatio = IsWindowObjectExist() ? (window.devicePixelRatio || 1.0) : 1.0;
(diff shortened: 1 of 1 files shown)


deltakosh (approved):
Excellent! thanks a lot for the fix!

PR #11912, opened 2022-02-03
Fix maxLodLevel WebGPU

sebavan:


sebavan:
Follow up on https://github.com/BabylonJS/Babylon.js/pull/11907/files 

@bghgary would be great to look for native support.

@Popov72 would be great if you could double check I am not doing anything stupid.

deltakosh (approved):

Popov72:
That seems ok to me, except you will also need to perform a modification in `WebGPUEngine._generateMipmaps`.

src/Engines/Extensions/engine.cubeTexture.ts
```diff
@@ -220,16 +220,17 @@ ThinEngine.prototype._partialLoadImg = function (url: string, index: number, loa
...
+ThinEngine.prototype._setCubeMapTextureParams = function (texture: InternalTexture, loadMipmap: boolean, maxLevel?: number): void {
     var gl = this._gl;
     gl.texParameteri(gl.TEXTURE_CUBE_MAP, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
     gl.texParameteri(gl.TEXTURE_CUBE_MAP, gl.TEXTURE_MIN_FILTER, loadMipmap ? gl.LINEAR_MIPMAP_LINEAR : gl.LINEAR);
     gl.texParameteri(gl.TEXTURE_CUBE_MAP, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
     gl.texParameteri(gl.TEXTURE_CUBE_MAP, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
     texture.samplingMode = loadMipmap ? Constants.TEXTURE_TRILINEAR_SAMPLINGMODE : Constants.TEXTURE_LINEAR_LINEAR;
 
-    if (loadMipmap && this.getCaps().textureMaxLevel && maxLevel > 0) {
+    if (loadMipmap && this.getCaps().textureMaxLevel && maxLevel !== undefined && maxLevel > 0) {
```
Popov72, line 231: I think maxLevel == 0 is a valid value (and in effect disable mipmapping).

src/Engines/WebGPU/webgpuTextureHelper.ts
```diff
@@ -1369,6 +1369,13 @@ export class WebGPUTextureHelper {
 
         const hasMipMaps = texture.generateMipMaps;
         const layerCount = depth || 1;
+        let mipmapCount;
+        if (texture._maxLodLevel !== null) {
+            mipmapCount = texture._maxLodLevel;
```
Popov72, line 1374: I think we may need to use `texture._maxLodLevel+1` here, but I'm not sure, I could not find a clear doc on TEXTURE_MAX_LEVEL... But I do think the number of mipmaps equals maxLodLevel+1 as I understand it 0 is a valid value for TEXTURE_MAX_LEVEL.

src/Engines/Extensions/engine.cubeTexture.ts
sebavan: I should rename maxLodLevel... I took it from the previous PR to ensure the PG with incomplete DDS was ok in WEBGPU as well :-( maxLodLevel is containing the value + 1

src/Engines/WebGPU/webgpuTextureHelper.ts
sebavan: I should rename maxLodLevel... I took it from the previous PR to ensure the PG with incomplete DDS was ok in WEBGPU as well :-( maxLodLevel is containing the value + 1

sebavan: @Popov72 / @deltakosh I wonder if I should rename it or pass the value - 1 in param and add 1 when calling the function.

About generateMipMaps, I ll do it in the same PR. It is currently only ever used in DDS and is hidden so we are all good in term of time to fix it.

deltakosh: It is the max number of levels: so in my mind 0 means NO mip right. So I guess the name is fine  (hence this is the name in webGL world)

I'm not sure about the +1 though. This is something we need to confirm / test

src/Engines/Extensions/engine.cubeTexture.ts
deltakosh: I agree that maxLevel should be set if set to 0 but I'm not so sure for the name change

deltakosh: We need to test to be sure 0 is ok :) I found no doc on that topic

src/Engines/WebGPU/webgpuTextureHelper.ts
deltakosh: why +1 ? does 0 work here?

sebavan: Actually only comes frome me being dumb there is well the -1 https://github.com/BabylonJS/Babylon.js/pull/11907/files#diff-5a4cb9f6543e06e9c4821b90e96560872069e5bcb8eae7d70cadf058ee20933bR59 and tests from the previous pg are working so we are all set on this front…

just have to look at the generate mip map now in case this is used outside of the dds

Meakk: Yes, the number of mip maps is the `max_level - min_level + 1`.  
If you set min level 0, and max level 4, you have 5 mipmap levels

Popov72: Ok, so max_level=0 is indeed valid (I think min_level == base_level?).

Meakk: Yes, max_level=0 should be valid in theory. I've never checked though.

sebavan: I did kept > 1 as it should not change anything for us :-) basically in case of 0 we fallback on the general case where hasMipmap would be false. I ll at least tonight change the generation of mipmAp in WebGpu to account for the new param.

PR #11914, opened 2022-02-03
[GE] Make inputs work with multi-select

darraghjburke:
Makes the parameter inputs in the right panel work with multiselect. If you select multiple controls of the same type, you can edit their properties simultaneously. If you select controls of a different type, you can still edit the common properties such as color and background. When they have conflicting values, a — will be shown as a placeholder. A couple of complex features, like grids, are disabled in multiselect.

This is a big PR, but it's mostly the same change repeated across tons of different control types.

Also, some of the changes from my last PR ended up in this PR, I think because I started off from the wrong branch. Sorry about that :( you should be safe to ignore anything outside of guiEditor/src/components/ or /sharedComponents/


sharedUiComponents/lines/targetsProxy.tsx
```diff
@@ -0,0 +1,43 @@
...
+            if (targets.length === 0) return conflictingValuesPlaceholder;
+            const firstValue = targets[0][name];
+            for (const target of targets) {
+                if (target[name] !== firstValue) {
+                    return conflictingValuesPlaceholder;
+                }
+            }
+            return firstValue;
+        },
+        set(_, name, value, __) {
```
RaananW, line 26: nit - unneeded last variable in the function

guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -41,75 +42,99 @@ const vAlignBottomIcon: string = require("../../../../sharedUiComponents/imgs/vA
...
+    controls: Control[];
     lockObject: LockObject;
     onPropertyChangedObservable?: Observable<PropertyChangedEvent>;
 }
 
 export class CommonControlPropertyGridComponent extends React.Component<ICommonControlPropertyGridComponentProps> {
-    private _width = this.props.control.width;
-    private _height = this.props.control.height;
+    private _width : String | Number = 0;
+    private _height : String | Number = 0;
```
RaananW, line 52: did you really mean String | Number, or string | number?

RaananW (commented):
great use of a Proxy! No need to change the existing components.

I have a question regarding conflicting states. the checkBoxLineComponent has an addition of `isConflict` in its internal state. but the others don't.  I noticed the proxy takes care of conflicts, returning a specific value if the targets have different values. is there no need for `isCOnflict` state in others? or - why is it needed in the checkbox component?

guiEditor/src/components/propertyTab/propertyGrids/gui/inputTextPropertyGridComponent.tsx
```diff
@@ -107,15 +103,15 @@ export class InputTextPropertyGridComponent extends React.Component<IInputTextPr
...
+                    target={makeTargetsProxy(inputTexts, this.props.onPropertyChangedObservable)}
                     propertyName="margin"
                     onPropertyChangedObservable={this.props.onPropertyChangedObservable}
                 />
                 <CheckBoxLineComponent
                     iconLabel={"Auto stretch width"}
                     icon={sizeIcon}
                     label="AUTO STRETCH"
-                    target={inputText}
+                    target={makeTargetsProxy(inputTexts, this.props.onPropertyChangedObservable, false)}
```
RaananW, line 114: the compiler complains about the last variable here. is it needed?

RaananW (approved):

PR #11919, opened 2022-02-04
Fix issue with glTF skeletons

### LINKED ISSUE #11838: Add support for creating instances of skinned meshes from glTF when the skinned mesh has non-identity scale
See forum: https://forum.babylonjs.com/t/boundingvectors-of-instanced-glb-mesh-hierarchy/26069

I haven't looked at this deeply yet, but I believe the problem comes from the `overrideMesh` parameter on the `skin`. This was added for glTF to handle ignoring transforms for the skinned mesh as per spec, but it doesn't work with instancing.

bghgary:
Fixes #11838

- Change the glTF loader behavior to place the skinned mesh as a sibling of the skeleton root node without loading the transform. This replaces the overrideMesh behavior.
- Removes overrideMesh from the Skeleton class
- Clone the skeleton property when cloning a mesh
- Update bounding info with skin and morph applied for clone and new instance


src/Meshes/mesh.ts
```diff
@@ -2021,7 +2021,7 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
             if (batch.renderSelf[subMesh._id]) {
                 // Draw
                 if (onBeforeDraw) {
-                    onBeforeDraw(false, renderingMesh._effectiveMesh.getWorldMatrix(), effectiveMaterial, renderingMesh._effectiveMesh);
+                    onBeforeDraw(false, renderingMesh.getWorldMatrix(), effectiveMaterial);
```
bghgary, line 2024: I'm not confident this code is right. Can someone look over this and see if it makes sense? Also, what's a good way to test this code?

deltakosh (approved):

src/Lights/Shadows/shadowGenerator.ts
```diff
@@ -1220,10 +1220,10 @@ export class ShadowGenerator implements IShadowGenerator {
 
             // Draw
             renderingMesh._processRendering(effectiveMesh, subMesh, effect, material.fillMode, batch, hardwareInstancedRendering,
-                (isInstance, worldOverride, effectiveMaterial, effectiveMeshOverride) => {
-                    if (effectiveMeshOverride && effectiveMesh !== effectiveMeshOverride) {
-                        effectiveMeshOverride.getMeshUniformBuffer().bindToEffect(effect, "Mesh");
-                        effectiveMeshOverride.transferToEffect(worldOverride);
+                (isInstance, worldOverride) => {
+                    if (effectiveMesh !== renderingMesh) {
```
Popov72, line 1224: I think the condition should be `if (effectiveMesh !== renderingMesh && !isInstance) {` to be iso with the previous code.

Indeed, in the case where hardware instance rendering is not supported, `Mesh._processRendering` was calling the callback with `isInstance==true` and `effectiveMeshOverride==undefined` with the old code. In that case, we executed the `else` branch, so we should also do it with the new code.

Also, I think there's a bug in the old code, I think `effectiveMesh.transferToEffect(world);` should be `effectiveMesh.transferToEffect(isInstance ? worldOverride : world);` so that we use the world matrix passed to the function (which is the world matrix of the instance) when hardware instance rendering is not supported...

The relevant bit of code from `_processRendering` is:
```typescript
var world = instance.getWorldMatrix();
if (onBeforeDraw) {
    onBeforeDraw(true, world, effectiveMaterial);
}
```

src/Meshes/instancedMesh.ts
```diff
@@ -70,7 +70,7 @@ export class InstancedMesh extends AbstractMesh {
 
         this.setPivotMatrix(source.getPivotMatrix());
 
-        this.refreshBoundingInfo();
+        this.refreshBoundingInfo(true, true);
```
Popov72, line 73: This one is a breaking change but I don't know if it should be noted in the what's new file...

src/Meshes/mesh.ts
```diff
@@ -1332,7 +1332,7 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
      * @param applyMorph  defines whether to apply the morph target before computing the bounding info
      * @returns the current mesh
      */
-    public refreshBoundingInfo(applySkeleton: boolean = false, applyMorph: boolean = true): Mesh {
+    public refreshBoundingInfo(applySkeleton: boolean = false, applyMorph: boolean = false): Mesh {
```
Popov72, line 1335: Also a breaking change for `applyMorph`, but probably more a bug as I don't see why it was `true` by default previously...

src/Meshes/mesh.ts
Popov72: This code path is the normal code path when rendering a mesh without instances, so it can be tested with the default playground. Now, I'm not sure if it is `renderingMesh` that should be used to get the world matrix...

Popov72 (commented):
Nice PR, it simplifies a bit the code and make a same skeleton usable by multiple meshes!

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -204,15 +213,15 @@ export class GLTFLoader implements IGLTFLoader {
 
         this._completePromises.length = 0;
 
-        for (const name in this._extensions) {
-            const extension = this._extensions[name];
-            extension.dispose && extension.dispose();
-            delete this._extensions[name];
-        }
+        this._extensions.forEach((extension) => extension && extension.dispose());
+        this._extensions.length = 0;
```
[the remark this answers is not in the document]
bghgary: Oops, that's a mistake.

src/Meshes/mesh.ts
deltakosh: renderingMesh._effectiveMesh will return renderingMesh so I guess we are fine

src/Meshes/instancedMesh.ts
deltakosh: We considered it a bug fix :) The bb were wrong before

src/Meshes/mesh.ts
deltakosh: Yep same

loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -204,15 +213,15 @@ export class GLTFLoader implements IGLTFLoader {
...
 
-        (this._gltf as any) = null;
-        (this._babylonScene as any) = null;
+        (this._gltf as Nullable<IGLTF>) = null; // TODO
```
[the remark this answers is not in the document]
bghgary: Yes, fixing this will require touch a lot of lines to put `!` in many places since it is guaranteed to not be null in many places.

src/Lights/Shadows/shadowGenerator.ts
bghgary: I'll trust your judgement on this. I have no idea what this code is supposed to do. I believe what I did is identical to what it did before. There were no checks for `isInstance` before.

src/Meshes/instancedMesh.ts
bghgary: I'll add it to what's new. I will also note that skeletons are now being set when cloned.

src/Meshes/mesh.ts
bghgary: I believe this was added in 5.0 so should be okay.

src/Lights/Shadows/shadowGenerator.ts
bghgary: That said, I would prefer to do this change in a separate PR.

Popov72: The new check on `!isInstance` is to replace the old `effectiveMeshOverride` check. Previously we had:
```typescript
if (effectiveMeshOverride && effectiveMesh !== effectiveMeshOverride) {
```
`effectiveMesh !== effectiveMeshOverride` is now `effectiveMesh !== renderingMesh`  because `effectiveMeshOverride` is set to `renderingMesh` by the `Mesh._processRendering` calling code:
```typescript
if (onBeforeDraw) {
    onBeforeDraw(false, renderingMesh._effectiveMesh.getWorldMatrix(), effectiveMaterial, renderingMesh._effectiveMesh);
}
```
However, there's a second call to the callback in `Mesh._processRendering`:
```typescript
if (onBeforeDraw) {
    onBeforeDraw(true, world, effectiveMaterial);
}
```
With the old code in `ShadowGenerator`, we would take the `else` branch of the `if (effectiveMeshOverride && effectiveMesh !== effectiveMeshOverride) {` because `effectiveMeshOverride` is not set. In the new code, to differentiate the two `onBeforeDraw` calls, we can't use `effectiveMeshOverride` anymore but we can  use `isInstance` instead because the first call set it to `false` and the second to `true`. So, if `isInstance==true` we should take the `else` branch.

bghgary: I'll fix it in a separate PR.

Popov72: I think by disabling the hardware support for instances: `engine.getCaps().instancedArrays = false`.

Popov72:
This PG does not work with the new code: http://localhost:1338/Playground/index-local.html#8RU8Q3#111 (the PR is not deployed to the Playground yet, you will have to test it locally).

We should get:

image

We get instead:

image

Even when setting `m2.parent = m.parent;` and `m3.parent = m.parent;` everything is not ok, the eyes are not placed correctly ( http://localhost:1338/Playground/index-local.html#8RU8Q3#112):

image

(for the left alien, the eyes are inside the head)

bghgary:
This is an expected break because this PG is creating instances based on meshes with "Mesh" in the name. Since the left and right eye are no longer rooted in the same place, creating instances won't work without the parents.

What is the intention of this PG? If this is PG is trying to create a new instance of the model, call `instantiateHierarchy` from the root.

Popov72:
The intention is to create 2 instances of the original mesh. Even when using `instantiateHeirarchy` it does not work, the eyes are misplaced: 

http://localhost:1338/Playground/index-local.html#8RU8Q3#114

I guess my question is: how to make the PG work when this PR is in effect?

> this PG is creating instances based on meshes with "Mesh" in the name

It is instancing meshes that have "Mesh" in the class name, not in the name. Anyway, this code is useless so I have removed it.

Popov72:
Ok, I have found a way by applying `instantiateHierarchy ` on `__root__`:

http://localhost:1338/Playground/index-local.html#8RU8Q3#116

However, it's a breaking change compared to the previous PG, are we expected to do it that way now?

bghgary:
Yes, this is what I said earlier. 😄 

> `instantiateHierarchy` from the root

Instantiating from the middle of the skeleton shouldn't be doing what it was doing before. It just happens to work out in your original PG because the previous code was walking up the hierarchy and creating fake joints.

PR #11928, opened 2022-02-07
Add public version of AbstractMesh's _getPositionData function

BlakeOne:
PR to add a public version of the _getPositionData function to the AbstractMesh class to allow users to get positions data with skeleton and/or morph applied.
https://forum.babylonjs.com/t/how-do-i-extract-mesh-vertex-data-properly/27381/9

### DIFF
--- a/inspector/src/components/graph/canvasGraphService.ts
+++ b/inspector/src/components/graph/canvasGraphService.ts
@@ -92,6 +92,9 @@ const maxXDistancePercBetweenLinePoints = 0.1;
 // Color used to draw the rectangle that indicates no collection of data
 const noDataRectangleColor = "#aaaaaa";
 
+const smoothingFactor = 0.2; // factor to smooth the graph with
+const rangeMargin = 0.1; // extra margin to expand the min/max range on the graph
+
 /**
  * This function will debounce calls to functions.
  *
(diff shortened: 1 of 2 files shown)


deltakosh (approved):
Thanks :)

deltakosh:
Some tests are failing :(

BlakeOne:
I'm going to trace through the called functions and see if I can find the problem, it seems like it should work thou

BlakeOne:
Okay I rearranged the order of the different parts to match the original exactly and it’s validating locally now. :)

And I ended up adding an optional parameter to the function so that the positions data can optionally be passed in. That seems to be the most straightforward way to match the original behavior exactly, at least that I can find.

deltakosh:
Seems great like that!

BlakeOne:
It looks like the checks got cancelled... The only error is that it was cancelled thou, so hopefully the checks just needs to be restarted.
`##[error]The build was canceled by Babylon 4Ever.`

BlakeOne:
Hmm now that it cancelled and restarted the checks a few times they finally finished but there’s another failed visual check, “Chibi Rex FAILED”. It’s too late now to get into it tonight but tomorrow I can give it another try.

PR #11934, opened 2022-02-07
[XR] Fix updateTransformMatrix for multiview

### LINKED ISSUE #11933: [XR] WebXR controllers too big when rendering a skybox in multiview
[Forum user Charge posted a about some bugs they found while testing out the new multiview rendering feature](https://forum.babylonjs.com/t/webxr-layers-feature-broken-on-quest-2-with-skybox-and-environment/27258), including an issue they were seeing where the xr controllers were rendering too big when using a skybox.

This is the playground that repros the issue:
https://playground.babylonjs.com/#9K3MRA#741

rgerd:
A forum user found that WebXR controllers looked too big when rendering a skybox in multiview. This was due to the scene's projection matrices not being updated properly when rendering multiview.

This is because the skybox uses the `ignoreCameraMaxZ` flag, which causes the scene to update the transform matrix uniforms for the current `activeCamera`, but since the old code didn't update the transform matrix uniforms for the per-eye rig cameras when rendering multiview, the projection matrix was being set to the projection for the parent camera, causing the controllers to look too large after rendering everything else in the scene.

fix #11933


rgerd:
Should I still update the what's new for a bug that wasn't published in an official version?

RaananW:
> @RaananW / @deltakosh Should I still update the what's new for a bug that wasn't published in an official version?

No need.  to update what's new in that case.

There are validation tests failing, will wait for you to merge

RaananW (approved):

PR #11936, opened 2022-02-08
Fixing 2D control interactions on the HolographicSlate

rickfromwork:
The HolographicSlate is a 3D control with an embedded AdvancedDynamicTexture on it, but any 2D controls placed on it could not be interacted with. This boiled down to it missing the attachToMesh call connecting the texture to the backplate.

Playground used to test: https://www.babylonjs-playground.com/#XHZWI1#9

### DIFF
--- a/gui/src/3D/controls/holographicSlate.ts
+++ b/gui/src/3D/controls/holographicSlate.ts
@@ -148,6 +148,11 @@ export class HolographicSlate extends ContentDisplay3D {
      */
     protected _applyFacade(facadeTexture: AdvancedDynamicTexture) {
         this._contentMaterial.albedoTexture = facadeTexture;
+
+        // We should have a content plate by this point, but check for safety
+        if (this._contentPlate) {
+            facadeTexture.attachToMesh(this._contentPlate, true);
+        }
     }
 
     private _rebuildContent(): void {


rickfromwork:
Looking at the code, since we can place anything onto the AdvancedDynamicTexture, I don't think there's a need to specifically hold data about a potential image in the HolographicSlate class that overwrites the stored content (since it can easily be added separately and alongside 2D controls, as I've done in the example).
@RaananW Do you have any thoughts on that? If we decide not to have the image data included within the HolographicSlate class, I can make a new PR to pull it out (and clean up some other commented sections I noticed along the way).

RaananW:
> Looking at the code, since we can place anything onto the AdvancedDynamicTexture, I don't think there's a need to specifically hold data about a potential image in the HolographicSlate class that overwrites the stored content (since it can easily be added separately and alongside 2D controls, as I've done in the example).
> 
> @RaananW Do you have any thoughts on that? If we decide not to have the image data included within the HolographicSlate class, I can make a new PR to pull it out (and clean up some other commented sections I noticed along the way).

I don't see a reason to keep the image data if the referenced ADT is enough.

RaananW (approved):

PR #11937, opened 2022-02-08
Add public version of AbstractMesh's _getPositionData function

BlakeOne:
PR to add a public version of the _getPositionData function to the AbstractMesh class to allow users to get positions data with skeleton and/or morph applied.

Forum: https://forum.babylonjs.com/t/how-do-i-extract-mesh-vertex-data-properly/27381/9

NOTE: Undid previous commits, started with fresh, up-to-date repo and re-applied the changes. Hopefully it will build and validate successfully, which it is locally now...

Original closed PR for reference: https://github.com/BabylonJS/Babylon.js/pull/11928

### DIFF
--- a/dist/preview release/what's new.md
+++ b/dist/preview release/what's new.md
@@ -130,6 +130,7 @@
 
 ### Meshes
 
+- Added public version of `AbstractMesh` function _getPositionData. ([BlakeOne](https://github.com/BlakeOne))
 - Added default options parameter to Create functions. ([BlakeOne](https://github.com/BlakeOne))
 - `LineMesh` now allows assigning custom material via `material` setter. ([FullStackForger](https://github.com/FullStackForger)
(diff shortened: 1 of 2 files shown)


BlakeOne:
Hurray it passed all the checks now. :)

RaananW:
Thanks!

Just wondering, why do we need both? And there seems to be a different between both functions - the hidden public (`_get...`) calls the public function, but shouldn't it happen the other way?

BlakeOne:
The existing hidden function handles internals that @deltakosh doesn't want exposed. The exposed function has just the part that gets the vertex positions and applies the skeleton and/or morhp. The hidden internal function needs to do that as well as updating internal stuff as it did before, so it calls the exposed function for code reuse...

PS: Unfortunately when I undid the commits the old PR errorred and auto-closed so I had to make this new one.

deltakosh (approved):

PR #11940, opened 2022-02-08
[GUI] make sure host exists before running code on host

RaananW:
Fixing https://forum.babylonjs.com/t/setting-isenabled-on-a-checkbox-now-errors-out-crashing-our-gui/27448


darraghjburke (approved):
Looks good to me! Nice catch

PR #11944, opened 2022-02-08
Fix #11943

### LINKED ISSUE #11943: Mesh instances broken without hardware instancing support
**Repro**
- Bug repro: https://playground.babylonjs.com/#MSAHKR#33
- Expected result: Three meshes showing
- Current result: Only one mesh is showing

Note that in LH mode, the same problem doesn't happen.

deltakosh:


RaananW (approved):

src/Meshes/mesh.ts
```diff
@@ -515,6 +515,16 @@ export class Mesh extends AbstractMesh implements IGetSetVerticesData {
...
 
+        this._onBeforeDraw = ((isInstance: boolean, world: Matrix, effectiveMaterial?: Material) => {
+            if (isInstance && effectiveMaterial) {
+                if (this._uniformBuffer) {
+                    this.transferToEffect(world);
+                } else {
+                    effectiveMaterial.bindOnlyWorldMatrix(world);
+                }
+            }
+        }).bind(this);
```
RaananW, line 526: probably no need for the bind, but it works with it as well :-)

deltakosh: It is required because of the way we call it

deltakosh: As I want to avoid using an arrow function each time it is called

PR #11945, opened 2022-02-08
Load GLTF animations at 60 FPS by default instead of 1 FPS

carolhmj:
Added a static `TargetFps` variable in the GLTFFileLoader to control to which FPS we will target the animations. This will make them show better in the ACE, since it's not usual to see frames with decimal points.


RaananW:
Two glTF-related visualization tests fail - is there no change in running animations? want to check those demos and see what went wrong?

carolhmj:
> Two glTF-related visualization tests fail - is there no change in running animations? want to check those demos and see what went wrong?

Just a FYI I'll probably take a bit to tackle this because I'm setting up a new machine

loaders/src/glTF/glTFFileLoader.ts
```diff
@@ -150,6 +150,11 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
 
     private _onParsedObserver: Nullable<Observer<IGLTFLoaderData>>;
 
+    /**
+     * When loading GLTF Animations, which are defined in seconds, target them to this FPS.
+     */
+    public static TargetFps = 60;
+
```
bghgary, line 157: This is the old way of setting properties. It would be better to follow the pattern from the V2 options below.

carolhmj: Updated to a non static variable!

RaananW:
oh, you commited the build files as well...

That's all good. try to avoid commiting anything in the dist folder next time.

loaders/src/glTF/glTFFileLoader.ts
```diff
@@ -242,6 +242,11 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
      */
     public useSRGBBuffers = true;
 
+    /**
+     * When loading GLTF Animations, which are defined in seconds, target them to this FPS.
```
bghgary, line 246: nit: 
```suggestion
     * When loading glTF animations, which are defined in seconds, target them to this FPS.
```

PR #11950, opened 2022-02-08
Fix incorrectly named property on KeyboardInfoPre

bghgary:


RaananW:
That's a breaking change. Should we add getter/setter (deprecated) for the old property?

bghgary:
> That's a breaking change. Should we add getter/setter (deprecated) for the old property?

I added it in case you didn't see.

RaananW (approved):

PR #11953, opened 2022-02-09
[Native]Enable linear sampling for float textures

CedricGuillemet:
Used by Trail in the demo. IIRC , @bghgary , you had issues with linear sampling & float tex a while ago. Is it related?

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -828,7 +828,7 @@ export class NativeEngine extends Engine {
             highPrecisionShaderSupported: true,
             colorBufferFloat: false,
             textureFloat: true,
-            textureFloatLinearFiltering: false,
+            textureFloatLinearFiltering: true,
             textureFloatRender: false,
             textureHalfFloat: false,
             textureHalfFloatLinearFiltering: false,


Popov72:
Is this always available in Native? In WebGL/WebGL2 it's an extension, so it's not guaranteed to be available everywhere.

bghgary:
> @bghgary , you had issues with linear sampling & float tex a while ago. Is it related?

This sounds familiar but I don't think it was this capability. What does this capability translate to in terms of Babylon Native calls?

CedricGuillemet:
https://github.com/BabylonJS/Babylon.js/blob/88b68cece5dde2fb3e078011b449d64a4d3453af/src/Materials/Textures/rawTexture.ts#L43
Only point sampling for raw float texture.

CedricGuillemet:
> Is this always available in Native? In WebGL/WebGL2 it's an extension, so it's not guaranteed to be available everywhere.

I assume so as it's not a cap in bgfx. Moreover, there some latency between native and web. For example, compute shaders are available for more than 10 years now where on the web ...
https://developer.download.nvidia.com/compute/DevZone/docs/html/DirectCompute/doc/DirectCompute_Programming_Guide.pdf

bghgary:
When I was debugging skeleton issues I saw that linear filtering of float textures didn't work on many Android devices. We should make sure this works everywhere.

https://github.com/babylonjs/babylonreactnative/issues/256#issuecomment-921364716

PR #11954, opened 2022-02-09
[GE] Input Units

darraghjburke:
Adds unit buttons to positional properties (top, left, width, height, all padding). Rather than using text to specify px or %, you now have a button which you can click to toggle. (Typing the unit still works).

Also includes many UI changes for the right panel, although there is still a lot left to do. I decided to open a PR here before it grew too big.


guiEditor/src/diagram/coordinateHelper.tsx
```diff
@@ -213,4 +219,107 @@ export class CoordinateHelper {
...
+                const thickness = (guiControl.parent as Rectangle).thickness * 2;
+                ratioX = guiControl.parent._currentMeasure.width - thickness;
+                ratioY = guiControl.parent._currentMeasure.height - thickness;
+            } else {
+                ratioX = guiControl.parent._currentMeasure.width;
+                ratioY = guiControl.parent._currentMeasure.height;
+            }
+        }
+        let old = {} as any;
+        if (!properties || properties.includes("left")) {
```
RaananW, line 245: maybe instead of this, set a default value for "properties" to be the whole list. Something like:

properties = properties || ["width", "left" , "height" , "top" , "paddingLeft" ,"paddingRight" , "paddingTop" ,"paddingBottom"]
This way you don't have to check if properties is undefined

guiEditor/src/diagram/coordinateHelper.tsx
```diff
@@ -213,4 +219,107 @@ export class CoordinateHelper {
...
+            old.paddingRight = guiControl.paddingRight;
+            const paddingRight = (guiControl.paddingRightInPixels * 100) / ratioX;
+            guiControl.paddingRight = `${paddingRight.toFixed(2)}%`;
+        }
+        if (!properties || properties.includes("paddingBottom")) {
+            old.paddingBottom = guiControl.paddingBottom;
+            const paddingBottom = (guiControl.paddingBottomInPixels * 100) / ratioY;
+            guiControl.paddingBottom = `${paddingBottom.toFixed(2)}%`;
+        }
+        for(const property of properties || ["left", "top", "width", "height"]) {
```
RaananW, line 285: maybe this is the default value?

guiEditor/src/components/propertyTab/propertyGrids/gui/ellipsePropertyGridComponent.tsx
```diff
@@ -47,6 +40,17 @@ export class EllipsePropertyGridComponent extends React.Component<IEllipseProper
                     target={makeTargetsProxy(ellipses, this.props.onPropertyChangedObservable)}
                     propertyName="thickness"
                     onPropertyChangedObservable={this.props.onPropertyChangedObservable}
+                    unit={"PX"}
```
RaananW, line 43: is `unit: 'PX'` and arbitrary value? or is there some sort of mapping between PX => functionality and % => functionality?

guiEditor/src/diagram/coordinateHelper.tsx
RaananW, line 245: feels like this entire section can be abstracted. Don't invest too much time for that, but see if you can prevent code repetition here. If it's not simple though, there is no need :-)

darraghjburke: You're totally right! Went through a few different iterations so I didn't stop to think about how simple it could ultimately be :)

guiEditor/src/components/propertyTab/propertyGrids/gui/ellipsePropertyGridComponent.tsx
darraghjburke: For now, the unit string and the functionality of the unit button is totally controlled outside of the component itself. So in the common control component, we handle all the logic of swapping units between PX and %.

I could make some kind of more formal unit class or enum, but honestly, I didn't see much added value in it over just passing in a string. Especially since most controls will just use a fixed PX unit.

RaananW: That's a fair point. The only reason I see making an enum in that case is to make sure formatting (px vs PX vs Px) will be preserved. But that's fine like that.

RaananW (approved):
Just a small linting issue

PR #11959, opened 2022-02-09
DeviceInputSystem: Add configure/removeEvents to be called by DeviceSourceManager

PolygonalSun:
This PR contains a few fixes.  First, the contract for DeviceInputSystem has been changed to add `configureEvents` and `removeEvents` as public functions.  These functions will now also be called by the DeviceSourceManager, in order to enable and disable event handling.  The DeviceSourceManager will also have `enableEvents` and `disableEvents`.  These functions will call configure and remove respectively.  The enable and disableEvents functions should only be called by the InputManager but can tehcnically be called by other classes or the user.  Finally, the default max number of touch points if `navigator.maxTouchPoints` doesn't exist is now 2 to support the usage of device that may be interpreted as touch but aren't touch.  2 was also picked because we only really support up to 2 finger gestures.


src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -121,33 +122,31 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
...
-            window.removeEventListener("gamepadconnected", this._gamepadConnectedEvent);
-            window.removeEventListener("gamepaddisconnected", this._gamepadDisconnectedEvent);
-        }
-    }
-
     /**
      * Configures events to work with an engine's active element
      */
-    private _configureEvents(): void {
+     public configureEvents(): void {
```
bghgary, line 128: These names are asymmetrical. Either configure/unconfigure or match the naming from DeviceSourceManager?

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -121,33 +122,31 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
...
+                            const deviceSlot = +deviceSlotKey;
+                            const device = this._inputs[deviceType]?.[deviceSlot];
+                            if (device) {
+                                for (let inputIndex = 0; inputIndex < device.length; inputIndex++) {
+                                    device[inputIndex] = 0;
+                                }
+                            }
+                        }
+                    }
+                }
```
bghgary, line 149: nit:
```suggestion
                for (const inputs of this._inputs) {
                    if (inputs) {
                        for (const deviceSlotKey in inputs) {
                            const deviceSlot = +deviceSlotKey;
                            const device = inputs[deviceSlot];
                            if (device) {
                                for (let inputIndex = 0; inputIndex < device.length; inputIndex++) {
                                    device[inputIndex] = 0;
                                }
                            }
                        }
                    }
                }

```

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -156,12 +155,58 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
...
+        this.onDeviceConnected = () => { };
+        this.onDeviceDisconnected = () => { };
+        this.onInputChanged = () => { };
+
+        if (this._elementToAttachTo) {
+            this.removeEvents();
+
+            // Gamepad Events
+            window.removeEventListener("gamepadconnected", this._gamepadConnectedEvent);
+            window.removeEventListener("gamepaddisconnected", this._gamepadDisconnectedEvent);
```
bghgary, line 206: why is this not part of removeEvents?

src/Inputs/scene.inputManager.ts
```diff
@@ -899,6 +902,7 @@ export class InputManager {
         if (this._alreadyAttached) {
 
             this._deviceSourceManager.onInputChangedObservable.remove(this._onInputObserver);
```
bghgary, line 904: Would it be better to dispose the deviceSourceManager here and recreate on attachControl? This will avoid all the interface changes.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -43,6 +43,23 @@ export class DeviceSourceManager {
...
+    /**
+     * Disable handling of input events
+     * @hidden
+     */
+    public disableEvents: () => void;
+
+    /**
+     * Dispose of DeviceSourceManager
+     */
+    public dispose: () => void;
```
bghgary, line 61: Probably makes more sense to extend `IDisposable`

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -56,5 +73,8 @@ export class DeviceSourceManager {
         this.getDeviceSource = this._deviceSourceManager.getDeviceSource;
         this.getDeviceSources = this._deviceSourceManager.getDeviceSources;
         this.getDevices = this._deviceSourceManager.getDevices;
+        this.enableEvents = this._deviceSourceManager.enableEvents;
+        this.disableEvents = this._deviceSourceManager.disableEvents;
+        this.dispose = this._deviceSourceManager.dispose;
```
bghgary, line 78: Does this work? `InternalDeviceSourceManager` is basically a singleton right now. If dispose is called, it will dispose the singleton, which isn't what you want? I think you may want to use a ref counting scheme or do something else entirely.

bghgary (changes requested):

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: Changing to configure/unconfigure

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: I assume this is an old carryover from all of the changes.  I moved this to unconfigure

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I can implement IDisposable but that'd also require having this method because it's an interface (unless I'm misunderstanding this comment)

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I see what you mean.  Lemme implement a ref count to avoid it getting disposed of before it needs to.

src/DeviceInput/InputDevices/deviceSourceManager.ts
bghgary: Yes, you are right. Extending from IDisposable will allow people to see quickly that you can dispose without searching. It's not a big deal.

src/Inputs/scene.inputManager.ts
PolygonalSun: I originally thought about that but I figured that recreating another DeviceSourceManager would mean that we'd lose any observers that a user created for connected, disconnected, and changed inputs.  I think that just affecting the events would be good enough.

bghgary: Hmm, interesting. I think you are right in principle. This does bring up an interesting problem with the way DeviceSourceManager is currently implemented. DeviceSourceManager is pointing to a singleton InternalDeviceSourceManager, but all the events are currently shared. If this continues, then one instance of DSM can unregister an event from another instance and vice versa. Similarly, dispose should probably dispose the events that have been register for a particular instance of DSM. Ugh, I'm not sure this DSM trick we did is a good idea anymore.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: Changed

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: I ended up writing dispose to just dispose of the DSM instance and left the InternalDSM to handle the proper disposal when the engine calls dispose

PolygonalSun: The dispose on the DeviceSourceManager will now just remove any observers and remove any references to parts of the InternalDeviceSourceManager, preventing any changes to other instances of the DSM.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -43,18 +44,63 @@ export class DeviceSourceManager {
...
-        this.onDeviceConnectedObservable = this._deviceSourceManager.onDeviceConnectedObservable;
-        this.onInputChangedObservable = this._deviceSourceManager.onInputChangedObservable;
-        this.onDeviceDisconnectedObservable = this._deviceSourceManager.onDeviceDisconnectedObservable;
+        // Observables
+        this.onDeviceConnectedObservable = new Observable<DeviceSource<DeviceType>>((observer) => {
+            this.getDevices().forEach((device) => {
+                if (device) {
+                    this.onDeviceConnectedObservable.notifyObserver(observer, device);
+                }
+            });
```
bghgary, line 77:
```suggestion
            for (const device of this.getDevices()) {
                this.onDeviceConnectedObservable.notifyObserver(observer, device);
            }
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -43,18 +44,63 @@ export class DeviceSourceManager {
...
+    /**
+     * Disable handling of input events
+     * @hidden
+     */
+    public disableEvents: () => void;
+
+    /**
+     * Dispose of DeviceSourceManager
+     */
+    public dispose: () => void;
```
bghgary, line 62: Why is this not a normal function?

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -25,7 +26,7 @@ export class DeviceSourceManager {
     public readonly onDeviceDisconnectedObservable: Observable<DeviceSource<DeviceType>>;
 
     // Private Members
-    private _deviceSourceManager: InternalDeviceSourceManager;
+    private _deviceSourceManager: Nullable<InternalDeviceSourceManager>;
```
bghgary, line 29: This isn't necessary. You can get away with it without making this nullable.

src/Inputs/scene.inputManager.ts
bghgary: Talked offline and we believe we don't need enable/disable events anymore and we can dispose the DSM now. This should simplify the code a bunch.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -43,18 +44,63 @@ export class DeviceSourceManager {
...
+        this.disableEvents = this._deviceSourceManager.disableEvents;
+
+        this.dispose = (): void => {
+            // Null out observable refs
+            this._deviceSourceManager?.unregisterManager(this);
+            this.onDeviceConnectedObservable.clear();
+            this.onInputChangedObservable.clear();
+            this.onDeviceDisconnectedObservable.clear();
+            // Null out function refs
+            this.getDeviceSource = <T extends DeviceType>(deviceType: T, deviceSlot?: number) => { return null; };
```
bghgary, line 97:
```suggestion
            this.getDeviceSource = () => null;
```

bghgary: Similarly for the following lines.

src/DeviceInput/InputDevices/internalDeviceSourceManager.ts
```diff
@@ -115,9 +110,33 @@ export class InternalDeviceSourceManager implements IDisposable {
...
+
+    public enableEvents = (): void => {
+        this._deviceInputSystem.configureEvents();
+    }
+
+    public disableEvents = (): void => {
+        this._deviceInputSystem.unconfigureEvents();
+    }
+
+    public dispose = (): void => {
```
bghgary, line 133: These should be normal functions.

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -25,7 +26,7 @@ export class DeviceSourceManager {
     public readonly onDeviceDisconnectedObservable: Observable<DeviceSource<DeviceType>>;
 
     // Private Members
-    private _deviceSourceManager: InternalDeviceSourceManager;
+    private _deviceSourceManager: Nullable<InternalDeviceSourceManager>;
 
     // Public Functions
```
bghgary, line 31: These function lambdas must be readonly.

src/DeviceInput/InputDevices/internalDeviceSourceManager.ts
```diff
@@ -115,9 +110,33 @@ export class InternalDeviceSourceManager implements IDisposable {
...
+        // If we have managers that weren't properly disposed of, dispose of their observables.
```
bghgary, line 134: I don't think this is necessary. The DSM's dispose will do this already.

bghgary (changes requested):
Reviewed offline live. Comments below.

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Removed Nullable

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Converted to normal function

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Removed if

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: simplified signature as per we discussed offline

src/DeviceInput/InputDevices/internalDeviceSourceManager.ts
PolygonalSun: dispose made normal, remove enable/disable

src/DeviceInput/InputDevices/internalDeviceSourceManager.ts
PolygonalSun: Removed and set up DSMs to handle their own dispose calls

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: function lambdas have been made readonly at the internalDSM level

RaananW:
@bghgary - I leave it to you to resolve the conversations and approve

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: Update: modified to follow public get/private set pattern

src/Inputs/scene.inputManager.ts
PolygonalSun: enable/disable has been removed

src/Engines/Extensions/engine.views.ts
```diff
@@ -22,12 +22,20 @@ export class EngineView {
...
         /**
          * Gets or sets the  HTML element to use for attaching events
          */
         inputElement: Nullable<HTMLElement>;
 
+        /**
+         * Observable to handle when a change to inputElement occurs
+         * @hidden
+         */
+        onEngineViewChanged: () => void;
```
bghgary, line 37:
```suggestion
        _onEngineViewChanged?: () => void;
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -25,36 +27,84 @@ export class DeviceSourceManager {
...
+    private _getDevices: () => ReadonlyArray<DeviceSource<DeviceType>>;
 
     // Public Functions
     /**
      * Gets a DeviceSource, given a type and slot
      */
-    public getDeviceSource: <T extends DeviceType>(deviceType: T, deviceSlot?: number) => Nullable<DeviceSource<T>>;
+    public get getDeviceSource(): <T extends DeviceType>(deviceType: T, deviceSlot?: number) => Nullable<DeviceSource<T>> {
+        return this._getDeviceSource;
+    }
```
bghgary, line 43:
```suggestion
    public getDeviceSource<T extends DeviceType>(deviceType: T, deviceSlot?: number): Nullable<DeviceSource<T>> {
        return this._getDeviceSource(deviceType, deviceSlot);
    }
```

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -82,7 +83,18 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
...
+
+        // Set callback to enable event handler switching when inputElement changes
+        if (!this._engine.onEngineViewChanged) {
+            this._engine.onEngineViewChanged = () => {
+                this._enableEvents();
+            };
+
+            this._engine.onDisposeObservable.add(() => {
+                this._engine.onEngineViewChanged = () => { };
+            });
```
bghgary, line 96: not necessary

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -125,29 +137,41 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
      * Dispose of all the eventlisteners
      */
     public dispose(): void {
-        // Observables
+        // Callbacks
         this.onDeviceConnected = () => { };
         this.onDeviceDisconnected = () => { };
         this.onInputChanged = () => { };
+        this._engine.onEngineViewChanged = () => { };
```
bghgary, line 144:
```suggestion
        delete this._engine.onEngineViewChanged;
```

src/DeviceInput/InputDevices/deviceSourceManager.ts
```diff
@@ -25,36 +27,84 @@ export class DeviceSourceManager {
...
+        this.onDeviceDisconnectedObservable.clear();
+        // Null out function refs
+        this._getDeviceSource = () => { return null; };
+        this._getDeviceSources = () => { return []; };
+        this._getDevices = () => { return []; };
+
+        if (this._engine._deviceSourceManager && --this._engine._deviceSourceManager._refCount < 1) {
+            this._engine._deviceSourceManager.dispose();
+            delete this._engine._deviceSourceManager;
+        }
```
bghgary, line 107: nit:
```suggestion
        this.onDeviceConnectedObservable.clear();
        this.onInputChangedObservable.clear();
        this.onDeviceDisconnectedObservable.clear();
        // Null out function refs
        this._getDeviceSource = () => { return null; };
        this._getDeviceSources = () => { return []; };
        this._getDevices = () => { return []; };

        if (this._engine._deviceSourceManager) {
            this._engine._deviceSourceManager.unregisterManager(this);

            if (--this._engine._deviceSourceManager._refCount < 1) {
                this._engine._deviceSourceManager.dispose();
                delete this._engine._deviceSourceManager;
            }
        }
```

src/Inputs/scene.inputManager.ts
```diff
@@ -897,8 +890,8 @@ export class InputManager {
      */
     public detachControl() {
         if (this._alreadyAttached) {
-
-            this._deviceSourceManager.onInputChangedObservable.remove(this._onInputObserver);
+            this._deviceSourceManager?.dispose();
```
bghgary, line 893: nit:
```suggestion
            this._deviceSourceManager!.dispose();
```

bghgary (approved):
two small nits

src/Inputs/scene.inputManager.ts
PolygonalSun: fixed

src/DeviceInput/InputDevices/deviceSourceManager.ts
PolygonalSun: fixed

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: fixed

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: removed

PR #11960, opened 2022-02-10
[GE] Animation Sheet Improvements

darraghjburke:
Makes the animation sheet less confusing. Effectively, makes it a toggle rather than using magic values of cell id. Prevents invalid cell IDs, widths, and heights.

Also swaps out the default grass image for this one:
imageControlDefault

Full list of changes here: https://github.com/BabylonJS/ThePirateCove/issues/141


guiEditor/src/guiNodeTools.ts
```diff
@@ -18,7 +18,7 @@ import { RadioButton } from "babylonjs-gui/2D/controls/radioButton";
 import { ImageBasedSlider } from "babylonjs-gui/2D/controls/sliders/imageBasedSlider";
 
 export class GUINodeTools {
-    public static ImageControlDefaultUrl = "https://playground.babylonjs.com/textures/grass.png";
+    public static ImageControlDefaultUrl = "https://playground.babylonjs.com/textures/imageControlDefault.jpg";
```
RaananW, line 21: maybe the playground is not the best place for this. This should be in the assets library instead

RaananW: awesome :-)

RaananW (approved):

PR #11962, opened 2022-02-10
Add a flag to force Babylon to use OpenGL convention for UVs

deltakosh:


src/Meshes/Builders/capsuleBuilder.ts
```diff
@@ -131,7 +132,7 @@ export function CreateCapsuleVertexData(
             normal.set(sinTheta, slope, cosTheta).normalize();
             normals.push(normal.x, normal.y, normal.z);
             // uv
-            uvs.push(u, 1 - v / vl);
+            uvs.push(u, v / vl ? v / vl : 1 - v / vl);
```
Popov72, line 135: Should be `CompatibilityOptions.UseOpenGLOrientationForUV ? v / vl : 1 - v / vl`

src/Meshes/geometry.ts
```diff
@@ -1208,31 +1209,61 @@ export class Geometry implements IGetSetVerticesData {
 
             if (binaryInfo.uvsAttrDesc && binaryInfo.uvsAttrDesc.count > 0) {
                 var uvsData = new Float32Array(parsedGeometry, binaryInfo.uvsAttrDesc.offset, binaryInfo.uvsAttrDesc.count);
+                if (CompatibilityOptions.UseOpenGLOrientationForUV) {
+                    for (var index = 1; index < uvsData.length; index += 2) {
+                        uvsData[index] *= -1;
```
Popov72, line 1214: Why not `1-uvsData[index]`? I think it's not the same thing than `*= -1` if the address mode is CLAMP.

For eg, if v=0.3, the current code will compute -0.3 and in CLAMP mode you will get 0. With `1-v` you get 0.7 and still get 0.7 in CLAMP mode.

Same comment for all the loops below.

src/Meshes/Builders/capsuleBuilder.ts
deltakosh: lol yes!! bad copy paste

RaananW (approved):

RaananW:
There is a case issue (first letter should be a capital C) with the file you have added. Not noticeable in windows, failing in Unix

PR #11979, opened 2022-02-15
HolographicSlate clean-up

rickfromwork:
This change updates the HolographicSlate to not center around holding an image. Instead, an image can be set by setting the .content variable. This variable is also used to assign other controls/content (unchanged). Previously, setting the image URL variable would overwrite any content added via the aforementioned variable.

Other changes include cleaning up the public vectors to avoid mismatch issues (updating the slate size without updating the title bar size), and renaming public variables for more clarity (backplate -> titleBar).

### DIFF
--- a/gui/src/3D/controls/holographicSlate.ts
+++ b/gui/src/3D/controls/holographicSlate.ts
@@ -11,7 +11,6 @@ import { Matrix, Quaternion, Vector2, Vector3 } from "babylonjs/Maths/math.vecto
 import { Control3D } from "./control3D";
 import { ContentDisplay3D } from "./contentDisplay3D";
 import { AdvancedDynamicTexture } from "../../2D/advancedDynamicTexture";
-import { Image } from "../../2D/controls/image";
 import { SlateGizmo } from "../gizmos/slateGizmo";
 import { DefaultBehavior } from "../behaviors/defaultBehavior";
 import { Viewport } from "babylonjs/Maths/math.viewport";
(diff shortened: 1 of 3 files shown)


gui/src/3D/controls/holographicSlate.ts
```diff
@@ -366,24 +330,24 @@ export class HolographicSlate extends ContentDisplay3D {
...
             // if (pickedPoint) {
-            //     this._backPlateMaterial. = pickedPoint;
-            //     this._backPlateMaterial.hoverColor.a = 1.0;
+            //     this._titleBarMaterial.globalLeftIndexTipPosition = pickedPoint;
+            //     this._titleBarMaterial.hoverColor.a = 1.0;
             // } else {
-            //     this._backPlateMaterial.hoverColor.a = 0;
+            //     this._titleBarMaterial.hoverColor.a = 0;
             // }
         });
```
RaananW, line 342: if nothing is actually running here, maybe you can either comment the entire `add` function or remove it altogether.

gui/src/3D/gizmos/slateGizmo.ts
```diff
@@ -236,7 +235,7 @@ export class SlateGizmo extends Gizmo {
 
         const dragStart = (event: { position: Vector3 }) => {
             if (this.attachedSlate && this.attachedMesh) {
-                dimensionsStart.copyFrom(this.attachedSlate.dimensions);
+                dimensionsStart.set(this.attachedSlate.dimensions.x, this.attachedSlate.dimensions.y, 0.001);
```
RaananW, line 238: maybe use Epsilon from math.constants instead of setting 0.001? If it makes sense, of course.

RaananW (commented):
LGTM! just two questions

gui/src/3D/gizmos/slateGizmo.ts
rickfromwork: I used 0.001 since it was what was used before, but honestly I think epsilon would be fine.

Side note, I noticed that if the depth is 0 then you can't interact with the contents at all (as if the ray hit passes right through the slate). Do you know anything about that?

RaananW (approved):

PR #11994, opened 2022-02-17
[GE] Highlight all selected controls

darraghjburke:
Correctly shows you all selected controls in the tree view, rather than just the most recent selection


RaananW (approved):
LGTM 😊

PR #11995, opened 2022-02-17
DeviceInputSystem: Add handling for pointercancel events

PolygonalSun:
This PR contains additional event handling for `pointercancel` events.  The event factory was also modified to now assign a value for pointerType.

### DIFF
--- a/src/DeviceInput/Helpers/eventFactory.ts
+++ b/src/DeviceInput/Helpers/eventFactory.ts
@@ -50,6 +50,8 @@ export class DeviceEventFactory {
 
         evt.pointerId = deviceType === DeviceType.Mouse ? 1 : deviceSlot;
 
+        evt.pointerType = deviceType === DeviceType.Mouse ? "mouse" : "touch";
+
         if (inputIndex === PointerInput.Move) {
             evt.type = "pointermove";
         }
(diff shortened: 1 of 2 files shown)


RaananW:
When is pointercancel triggered in the case of babylon's element? by pointerout scenarios?

src/DeviceInput/Helpers/eventFactory.ts
```diff
@@ -50,6 +50,8 @@ export class DeviceEventFactory {
 
         evt.pointerId = deviceType === DeviceType.Mouse ? 1 : deviceSlot;
 
+        evt.pointerType = deviceType === DeviceType.Mouse ? "mouse" : "touch";
```
sebavan, line 53: nit: can we enclose in a if with the previous line ?

RaananW:
> pointercancel events are triggered when the OS or some process interrupts what the pointer is doing and takes over the pointer, with respect to context. For example, on mobile, if you are trying to move things in the canvas and initiate a "Back" gesture, a pointercancel event will fire. On Desktop, it could be something like clicking and dragging an image on the page. It should be noted that with Mouse, these events are incredibly rare.
> 
> If a pointercancel event is fired, a pointerout and pointerleave will also be fired as a result afterwards but a pointerout will should never trigger a pointercancel.

So the general idea is to release all locked pointers and trigger a pointer up if the pointer state was "on", right?

LGTM, seb left a comment.

src/DeviceInput/Helpers/eventFactory.ts
PolygonalSun: Fixed

sebavan (approved):
LGTM, let s address @RaananW comments :-)

PolygonalSun:
> > pointercancel events are triggered when the OS or some process interrupts what the pointer is doing and takes over the pointer, with respect to context. For example, on mobile, if you are trying to move things in the canvas and initiate a "Back" gesture, a pointercancel event will fire. On Desktop, it could be something like clicking and dragging an image on the page. It should be noted that with Mouse, these events are incredibly rare.
> > If a pointercancel event is fired, a pointerout and pointerleave will also be fired as a result afterwards but a pointerout will should never trigger a pointercancel.
> 
> So the general idea is to release all locked pointers and trigger a pointer up if the pointer state was "on", right?
> 
> LGTM, seb left a comment.

Yup, that's pretty much the idea.  It's kind of like a mini-blur.

PR #12008, opened 2022-02-20
SubSurfaceConfiguration: Fix name in exception message

Popov72:
Also, why not having:
```javascript
import "./SubSurfaceSceneComponent";
```
directly in `SubSurfaceConfiguration` instead of relying on the user not forgetting to add it themself? This import is always required for  the `SubSurfaceConfiguration` object to work as expected, so it would be easier to just have the import done by this class.

Same thing for the pre pass renderer, the depth renderer, and more generally all the scene components that are all built the same way.

[...] Ah, it seems there's a problem with the order of the module initializations when doing this. There's an error on:
```typescript
SubSurfaceConfiguration._SceneComponentInitialization = (scene: Scene) => {
    // Register the G Buffer component to the scene.
    let component = scene._getComponent(SceneComponentConstants.NAME_SUBSURFACE) as SubSurfaceSceneComponent;
    if (!component) {
        component = new SubSurfaceSceneComponent(scene);
        scene._addComponent(component);
    }
};
```
which says "Cannot access 'SubSurfaceConfiguration' before initialization"...

### DIFF
--- a/src/Rendering/subSurfaceConfiguration.ts
+++ b/src/Rendering/subSurfaceConfiguration.ts
@@ -14,7 +14,7 @@ import { Constants } from "../Engines/constants";
 export class SubSurfaceConfiguration implements PrePassEffectConfiguration {
     /** @hidden */
     public static _SceneComponentInitialization: (scene: Scene) => void = (_) => {
-        throw _WarnImport("PrePassRendererSceneComponent");
+        throw _WarnImport("SubSurfaceSceneComponent");
     }
 
     private _ssDiffusionS: number[] = [];


sebavan:
I am all for adding the missing `import "./SubSurfaceSceneComponent";` if you can in the PR ?

Popov72:
In fact it does not work, I have updated my post above afterwards.

There's a problem with the order of initializations I guess, and I don't know if it can be fixed (I'm not savvy enough with those things to know).

sebavan:
So let s use your fix for now and we ll find a way to address it with @RaananW

sebavan (approved):

PR #12009, opened 2022-02-21
Fix type of param clearColor for drawText

BlakeOne:
When drawing text to a DynamicTexture, null is passed for the clearColor to not clear it. However in Typescript null is not allowed do to incorrect type of this parameter. This PR  just changes the type from "string" to "string | null" (like the param "color").

Forum issue: https://forum.babylonjs.com/t/writing-text-on-top-of-a-cylinder/27591/11


RaananW (approved):
LGTM!

This is not really a bug, so no need to update the readme in that case.

Also, if possible, please add the line at the end of the what's-new to maintain chronologic order. Thanks!!

PR #12025, opened 2022-02-22
Fix env intensity perf + axes vectors

sebavan:
- Fixes https://forum.babylonjs.com/t/450-worse-performance-when-setting-environmentintensity/27867/8
- Address @bghgary Feedback on transfornode GC


src/Meshes/transformNode.ts
```diff
@@ -254,30 +252,39 @@ export class TransformNode extends Node {
      * The forward direction of that transform in world space.
      */
     public get forward(): Vector3 {
-        return Vector3.Normalize(Vector3.TransformNormal(
-            this.getScene().useRightHandedSystem ? this._forwardInverted : this._forward,
-            this.getWorldMatrix()
-        ));
+        Vector3.TransformNormalToRef(
```
bghgary, line 255: I probably would have used `TransformNormalFromFloatsToRef` but this works too.

src/Meshes/transformNode.ts
```diff
@@ -254,30 +252,39 @@ export class TransformNode extends Node {
...
-            this.getScene().useRightHandedSystem ? this._forwardInverted : this._forward,
-            this.getWorldMatrix()
-        ));
+        Vector3.TransformNormalToRef(
+            this.getScene().useRightHandedSystem ? Vector3.RightHandedForwardReadOnly : Vector3.LeftHandedForwardReadOnly,
+            this.getWorldMatrix(),
+            this._forward
+        );
+        this._forward.normalize();
+        return this._forward;
```
bghgary, line 261: nit: can some some chars :-)
```suggestion
        return this._forward.normalize();
```

bghgary (approved):
LTGM. Some really small nits.

src/Meshes/transformNode.ts
```diff
@@ -252,39 +252,36 @@ export class TransformNode extends Node {
...
     }
 
     /**
      * The right direction of that transform in world space.
      */
     public get right(): Vector3 {
-        Vector3.TransformNormalToRef(
-            this.getScene().useRightHandedSystem ? Vector3.LeftReadOnly : Vector3.RightReadOnly,
+        Vector3.TransformNormalFromFloatsToRef(
+            this.getScene().useRightHandedSystem ? -1.0 : 1.0, 0, 0,
```
bghgary, line 280: This isn't your change, but is this right? We flip both forward and right vectors for right-handed? Doesn't that make it left-handed again?

sebavan: yup I wondered the same but I did not dare changing it due to back compat :-(

PR #12026, opened 2022-02-22
[GUI] Make touch inputs open the prompt on inputText and inputPassword

darraghjburke:
Forum issue: https://forum.babylonjs.com/t/inputpassword-gui-bug-on-ipad/27819

Previously, we were deciding whether or not to open an input prompt based on the user agent. This would fail to detect tablet devices which don't have "mobile" in the user agent. I have switched it to open the prompt if the last pointerDown event came from a "touch" device.

As a consequence of this, touching the input even from a touch-enabled laptop will open the prompt. I think this is the right behavior, but not everyone may agree. As far as I can tell, there is no way to detect the presence of a physical keyboard.

We may want to consider doing the same if the input is from a "pen" device? I am not sure.


sebavan (approved):

sebavan (approved):

PolygonalSun:
LGTM

PR #12030, opened 2022-02-23
Early return from InstancedMesh.getLOD if there are no LODs

jure:
`Mesh.getLOD` has this nice thing where it returns early in case there are no LODs. `InstancedMesh.getLOD` uses `Mesh.getLOD` underneath, but not before it calls off to a `AbstractMesh.getBoundingInfo`: https://github.com/BabylonJS/Babylon.js/blob/087deb584b40fc5fe8543d54b3e3eed60d1cc121/src/Meshes/instancedMesh.ts#L390

So even if there are no LODs, a whole lot happens to figure out which one would be the right one:
Screenshot 2022-02-23 at 01 05 02

I assume LODs are not used very often, especially in combination with `InstancedMesh`es, so this PR includes a significant early return optimisation for that common case.

### DIFF
--- a/src/Meshes/instancedMesh.ts
+++ b/src/Meshes/instancedMesh.ts
@@ -390,15 +390,9 @@ export class InstancedMesh extends AbstractMesh {
         const sourceMeshLODLevels = this.sourceMesh.getLODLevels();
         if(!sourceMeshLODLevels || sourceMeshLODLevels.length === 0) {
             this._currentLOD = this.sourceMesh;
-            return this.sourceMesh;
-        }
-
-        let boundingInfo = this.getBoundingInfo();
-
-        this._currentLOD = <Mesh>this.sourceMesh.getLOD(camera, boundingInfo.boundingSphere);
-
-        if (this._currentLOD === this.sourceMesh) {
-            return this.sourceMesh;
+        } else {
+            let boundingInfo = this.getBoundingInfo();
+            this._currentLOD = <Mesh>this.sourceMesh.getLOD(camera, boundingInfo.boundingSphere);
         }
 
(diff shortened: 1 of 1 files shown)


sebavan (approved):

sebavan:
I wonder if you could change the code from 

```
        const sourceMeshLODLevels = this.sourceMesh.getLODLevels();
        if(!sourceMeshLODLevels || sourceMeshLODLevels.length === 0) {
            this._currentLOD = this.sourceMesh;
            return this.sourceMesh;
        }

        let boundingInfo = this.getBoundingInfo();

        this._currentLOD = <Mesh>this.sourceMesh.getLOD(camera, boundingInfo.boundingSphere);
        if (this._currentLOD === this.sourceMesh) {
            return this.sourceMesh;
        }
        return this._currentLOD;
```

to 

```
        const sourceMeshLODLevels = this.sourceMesh.getLODLevels();
        if(!sourceMeshLODLevels || sourceMeshLODLevels.length === 0) {
            this._currentLOD = this.sourceMesh;
        } else {
            let boundingInfo = this.getBoundingInfo();
            this._currentLOD = <Mesh>this.sourceMesh.getLOD(camera, boundingInfo.boundingSphere);
        }
        
        return this._currentLOD;
```

sebavan:
if not, nw I ll do it after the pr

jure:
> I wonder if you could change the code from

Yes, cleaner this way. Thank you!

RaananW:
Ready to be merged! just waiting for the linting issues (formatting, in that case)

sebavan (approved):
Thanks a lot !!!

PR #12036, opened 2022-02-23
ParticleSystem: Add some accessors

Popov72:
Added accessors for the vertex / index buffers as some external code could be interested in using what is generated by the particle system...

### DIFF
--- a/src/Particles/particleSystem.ts
+++ b/src/Particles/particleSystem.ts
@@ -276,6 +276,20 @@ export class ParticleSystem extends BaseParticleSystem implements IDisposable, I
     }
 
     /**
+     * Gets the vertex buffers used by the particle system
+     */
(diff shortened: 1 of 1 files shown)


sebavan (commented):
This sounds a bit scary and I wonder if it should not stay hidden by placing the underscore as public cause touching this might have plenty of side effects :-)

Popov72:
The use case is reusing the vertex buffers (like the position) in other contexts, for example to achieve some special rendering for a set of particles.

It's on the responsibility of the user to use these accessors and mess with them or not: you can always crash anything if you do something bad :) My use case only implies reading the vertex buffers not modifying them => I have added an **Immutable** modifier on the returned object.

deltakosh (approved):

sebavan (approved):

PR #12037, opened 2022-02-23
Add arrow keys to move controls on GUI Editor

carolhmj:


guiEditor/src/diagram/workbench.tsx
```diff
@@ -975,6 +975,25 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
                 case "F":
                     this.globalState.onFitToWindowObservable.notifyObservers();
                     break;
+                case "ArrowUp": // move up
+                case "W":
+                case "w":
+                    this.moveControls(false, k.event.shiftKey ? -5 : -1);
```
darraghjburke, line 981: Can we make this a constant?

guiEditor/src/diagram/workbench.tsx
```diff
@@ -1015,6 +1034,34 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         this.props.globalState.onGizmoUpdateRequireObservable.notifyObservers();
     }
 
+    // Move the selected controls. Can be either on horizontal (leftInPixels) or 
+    // vertical (topInPixels) direction
+    moveControls(horizontal: boolean, amount: number) {
```
darraghjburke, line 1039: should second parameter be named vertical?

darraghjburke (commented):
love it!! just a couple nitpicks

guiEditor/src/diagram/workbench.tsx
RaananW: seems like horizontal is the mode (i.e. horizontal when true, vertical when false?).

RaananW (approved):
LGTM! I agree with @darraghjburke - the numbers can be constants (to avoid repeating them), but apart from that - all good :-)

Approved, waiting for resolution of the comments already provided.

guiEditor/src/diagram/workbench.tsx
carolhmj: Yeah, you're right, it's the mode! I think `moveHorizontal` could work, what do you think?

carolhmj: Updated!

guiEditor/src/diagram/workbench.tsx
carolhmj: Did it!

guiEditor/src/diagram/workbench.tsx
darraghjburke: Ah got it, I misread it, thanks for the explanation!

PR #12043, opened 2022-02-24
[GE] Selection Improvements

darraghjburke:
Makes the list of selected controls a property of the global state, rather than of the workbench individually. Uses functions to control the selected controls, so the observable can just notify listeners when the selection changes rather than handling selection logic.

Also, prevents user from selecting two controls with different parents.


RaananW (approved):

RaananW:
Some conflict(s) need to be resolves

PR #12046, opened 2022-02-24
Avoid extra vertex data copies during Mesh.MergeMeshes

ryantrem:
Previously, `MergeMeshes` would call `VertexData.ExtractFromMesh(mesh, true, true);` which would make a copies of the vertex buffers so they could be transformed prior to merge. With these changes, `MergeMeshes` calls `VertexData.ExtractFromMesh(mesh, false, false);` which does not make copies of the vertex buffers, and instead passes the transform matrix down to lower layers of the merge so the transform can be done "in place" on the final buffers. For the large model I'm testing with (7.5M verts), this reduces peak memory during the merge process by about 140mb. It also makes the merge operation faster, particularly in Babylon Native where for the same model this change reduces the merge time from 1470ms to 845ms.

### DIFF
--- a/src/Engines/nativeEngine.ts
+++ b/src/Engines/nativeEngine.ts
@@ -763,7 +763,7 @@ class CommandBufferEncoder {
 /** @hidden */
 export class NativeEngine extends Engine {
     // This must match the protocol version in NativeEngine.cpp
-    private static readonly PROTOCOL_VERSION = 2;
+    private static readonly PROTOCOL_VERSION = 3;
 
     private readonly _engine: INativeEngine = new _native.Engine();
     private readonly _camera: Nullable<INativeCamera> = _native.Camera ? new _native.Camera() : null;
--- a/src/Meshes/mesh.vertexData.ts
+++ b/src/Meshes/mesh.vertexData.ts
@@ -11,6 +11,7 @@ declare type Geometry = import("../Meshes/geometry").Geometry;
 declare type Mesh = import("../Meshes/mesh").Mesh;
 
 import { ICreateCapsuleOptions } from "./Builders/capsuleBuilder";
+import { Tools } from "../Misc/tools";
 declare type PolyhedronData= import("./geodesicMesh").PolyhedronData;
 
 /**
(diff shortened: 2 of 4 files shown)


src/Meshes/mesh.vertexData.ts
```diff
@@ -504,14 +504,15 @@ export class VertexData {
      * @returns the modified VertexData
      */
     public merge(others: VertexData | VertexData[], use32BitsIndices = false) {
-        return runCoroutineSync(this._mergeCoroutine(others, use32BitsIndices, false));
+        const vertexDatas: [vertexData: VertexData, transform?: Matrix][] = Array.isArray(others) ? others.map((other) => [other, undefined]) : [[others, undefined]];
```
sebavan, line 507: Why not keeping the complexity in _mergeCoroutine ? as it would be share with the async version ???

sebavan (commented):
only one tiny comment :-)

src/Meshes/mesh.vertexData.ts
ryantrem: All this is doing is adding undefined transform matrices to the args. I have to do this to not create a breaking change in this public API.

sebavan: My bad I was thinking of the Mesh static function did not even notice it was in vertex data ... :-)

sebavan (approved):

src/Engines/nativeEngine.ts
```diff
@@ -763,7 +763,7 @@ class CommandBufferEncoder {
 /** @hidden */
 export class NativeEngine extends Engine {
     // This must match the protocol version in NativeEngine.cpp
-    private static readonly PROTOCOL_VERSION = 2;
+    private static readonly PROTOCOL_VERSION = 3;
```
bghgary, line 766: Was there a change to the protocol?

src/Meshes/mesh.vertexData.ts
```diff
@@ -568,58 +577,73 @@ export class VertexData {
...
+            if (matrix) {
+                if (kind === VertexBuffer.PositionKind) {
+                    VertexData._TransformVector3Coordinates(element, matrix, offset, length);
+                } else if (kind === VertexBuffer.NormalKind) {
+                    VertexData._TransformVector3Normals(element, matrix, offset, length);
+                } else if (kind === VertexBuffer.TangentKind) {
+                    VertexData._TransformVector4Normals(element, matrix, offset, length);
+                }
+            }
+        };
```
bghgary, line 634: nit: not a big deal, but you can do the if checks outside instead of inside the function to improve perf slightly.

bghgary (commented):
LGTM except one nit

src/Engines/nativeEngine.ts
ryantrem: Yes, I added `offset` and `length` params to several transform functions that have native implementations.

src/Meshes/mesh.vertexData.ts
bghgary:
```suggestion
        const transformRange = matrix ? () => {} :
            kind === VertexBuffer.PositionKind ? VertexData._TransformVector3Coordinates :
            kind === VertexBuffer.NormalKind ? VertexData._TransformVector3Normals :
            kind === VertexBuffer.TangentKind ? VertexData._TransformVector4Normals;
```

bghgary (approved):

PR #12048, opened 2022-02-24
Turn picking on when gizmo is enabled (Inspector)

BlakeOne:
For the Inspector automatically turn picking on when gizmo is turned on so that we don't have to click an extra button to start using the gizmo.

https://forum.babylonjs.com/t/can-inspector-turn-on-picking-when-gizmo-is-enabled/27939/3

### DIFF
--- a/dist/what's new.md
+++ b/dist/what's new.md
@@ -121,6 +121,8 @@
 - Added more functionality and options to the skeleton debug panel. ([Pryme8](https://github.com/Pryme8))
 - Along with bone index it is now possible to select a bone using its name when viewing bone weights ([#9117](https://github.com/BabylonJS/Babylon.js/issues/9117)) ([RaananW](https://github.com/RaananW))
 - Gradient nodes from NME can now be set to be visible in the inspector ([msDestiny14](https://github.com/msDestiny14))
+- Automatically turn picking on when gizmo is enabled (Inspector) (BlakeOne](https://github.com/BlakeOne))
+
 
 ### Cameras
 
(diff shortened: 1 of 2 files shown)


sebavan:
actually lets ask @PatrickRyanMS what he thinks in term of UX cause I could see the current option being more convenient in some cases ?

BlakeOne:
Well if you already have a mesh selected and turn picking off then I could see maybe wanting to switch gizmos without turning the picking back on I suppose...

But it would (more?) often be more convenient to be able to open the Inspector, turn on a gizmo and go straight to clicking on a mesh to transform it, without needing to click an extra button that will always need to be clicked the first time since an enabled  gizmo can't be used without also turning on picking.

Also just for next time it would be better to ask him before I'm asked to make a PR to implement my proposed change... :)

sebavan:
> Also just for next time it would be much better to ask him before I'm asked to make a PR to implement my proposed change, just to save time... :)

Yup I would have but I thought it was a regression from the post you created :-) Now that it is about a new feature, I d prefer to be sure.

I totally agree turning it on sounds reasonable in probably the majority of case.

PatrickRyanMS:
I agree that when enabling the gizmo, it's likely that the user will want to be able to pick meshes by default. However, there is one complication in this in that the skybox, if present in the scene, is also a pick candidate. If the user clicks on their mesh to get the gizmo and then clicks off into "empty space" if there isn't another mesh under the mouse and there is a skybox in the scene, the skybox will be selected. And it seems that the gizmo is rendering at the position of the camera, so it does not render on screen. If you use the transform box gizmo, it will render on the skybox, but you can't manipulate it. 

I would like to ask for one more change with this PR if we are automatically enabling picking when the gizmo is active and that would be to exclude meshes that have [infiniteDistance](https://doc.babylonjs.com/typedoc/classes/babylon.mesh#infinitedistance) enabled from being picked. This will make the scenes a little easier to manage when picking is enabled. Would you be able to add this extra bit to the PR, @BlakeOne?

PR #12050, opened 2022-02-24
fixing pointer exit bug and exposing more info about near touch state

rickfromwork:
I noticed while testing controls that the pointer up event isn't being fired for near interaction, due to the pick data being null (as the near interactor is no longer colliding with the target. This change saves the previous pick data for use one frame past the end of the collision so that a pointer up can be sent to the correct target.

Also I expose two new components of the TouchButton3D class for developer access, namely isActiveNearInteraction and getPressDepth(position).
- isActiveNearInteraction returns true if there is an active near interaction on that button
- getPressDepth(position) returns how far into the button the near interaction is
These two additions allow developers to easily perform different actions based on the interaction type used to press the button, and tie in components to react based on press depth.

### DIFF
--- a/gui/src/3D/controls/touchButton3D.ts
+++ b/gui/src/3D/controls/touchButton3D.ts
@@ -168,6 +168,7 @@ export class TouchButton3D extends Button3D {
     /**
      * Get the height of the touchPoint from the collidable part of the button
      * @param touchPoint the point to compare to the button, in absolute position
+     * @returns the depth of the touch point into the front of the button
      */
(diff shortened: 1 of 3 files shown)


rickfromwork:
These changes were needed in order to make a TouchMeshButton3D behave like below:

https://user-images.githubusercontent.com/70608380/155621883-7692c59f-0d03-42de-aec6-2b0c4097fc3a.mp4

RaananW (approved):

PR #12056, opened 2022-02-26
[GUI] Add checkPointerEveryFrame property

### LINKED ISSUE #12027: Feature request: fire pointerOut events on GUI controls when the camera moves
Forum issue: https://forum.babylonjs.com/t/mesh-button-3d-doesnt-update-onpointeroutobservable-if-only-camera-moves/27716

Moving the pointer triggers pointerOut events, but moving the camera such that the pointer is no longer over a control does not currently trigger them. This behavior creates a strange result for the user where the control still thinks it's hovered even though the pointer is no longer on top of it.

The solution here will be call processPicking on the ADT whenever the camera's view matrix changes.

darraghjburke:
Adds a "checkPointerEveryFrame" property to AdvancedDynamicTexture. If true, and if the ADT is linked to a mesh or there are any controls linked to a mesh, every scene render will call doPicking, which has the potential to trigger POINTERMOVE events.

Test playground: https://playground.babylonjs.com/#ZI9AK7#2027

Makes PointerInfo nullable, as ew don't have any pointerinfo if there was no real pointer signal.

Closes #12027.

### DIFF
--- a/gui/src/2D/advancedDynamicTexture.ts
+++ b/gui/src/2D/advancedDynamicTexture.ts
@@ -45,9 +45,6 @@ export class AdvancedDynamicTexture extends DynamicTexture {
     private _resizeObserver: Nullable<Observer<Engine>>;
     private _preKeyboardObserver: Nullable<Observer<KeyboardInfoPre>>;
     private _pointerMoveObserver: Nullable<Observer<PointerInfoPre>>;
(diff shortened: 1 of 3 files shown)


RaananW (approved):

RaananW:
just a linting issue

PR #12065, opened 2022-02-28
Added check to ArcRotateCamera for non-y axis panning

PolygonalSun:
This PR contains a simple fix for a regression with the ArcRotateCamera's behavior.  This fix was provided by forum user Oscar and adds a check to see if the panning axis includes the y-axis.  If not, the transformedDirection information is then ignored (in the same manner as if mapPanning was true.  This should put the current behavior inline with the previous behavior from 4.2


PolygonalSun:
This is in reference to forum post: https://forum.babylonjs.com/t/panningaxis-behavior-changed-between-4-2-and-5-alpha/21116

sebavan (approved):

PR #12070, opened 2022-03-01
Fix native shaders OpenGL

CedricGuillemet:
In native, every uniform is converted to a vec4 and usage in shader is updated accordingly. This is done because bgfx only support vec4/mat4, for every rendering backend.
For example, 
`uniform bool alphaTest;` is converted to `uniform vec4 alphaTest;` and usage is changed:
`if (alphaTest)` to `if (alphaTest.x)`
A problem rised with OpenGL (at least it doesn't with Metal and HLSL, don't know for Vulkan): Conversion from a float value to a boolean is not authorized and compiling the shader crashes the app.

The fix here casts alphaTest to a float and compares it with 0.
I fixed it for the DOF postprocess as well. IIRC it crashed on Native ( ping @DarraghBurkeMS )

Tested with D3D11 and OpenGL.

### DIFF
--- a/src/Shaders/depthOfField.fragment.fx
+++ b/src/Shaders/depthOfField.fragment.fx
@@ -172,7 +172,7 @@ void main(void)
 	float coc = abs(aperture * (screen_distance * (inverse_focal_length - 1.0 / distance) - 1.0));
 
 	// disable blur
-	if (dof_enabled == false || coc < 0.07) { coc = 0.0; }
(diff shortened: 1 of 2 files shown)


bghgary:
This should be done in the shader compiler on the native side.

bghgary (changes requested):
This should be done in the shader compiler on the native side.

CedricGuillemet:
Fix will be done in Shader traversers/compilation instead. Closing.

PR #12074, opened 2022-03-02
Build Accessibility Tree from scene

mysunnytime:
(A draft PR hoping to get early feedback.)

1) build a DOM tree for: Node that are marked as salient, and GUI Controls. In this way, the screen reader can read the page content.
2) make the DOM elements interactive for actionable nodes. So the user can use "tab" key to focus on interactive nodes or controls, and "enter" key to trigger the interactive nodes.
3) when the scene content changes (e.g. show/hide new salient mesh), the DOM tree updates too.

@RaananW @Exolun 

image
image

Testing:

Testing code 1:
Given a scene that has objects that cover conditions of: 

- some are salient, some are not not-salient 
- some objects (spheres) are interactive, and could be left and right clicked
- objects have parent-children hierarchy

```
let createScene = function () {
    let scene = new BABYLON.Scene(engine);
    let camera = new BABYLON.ArcRotateCamera("Camera", Math.PI/2, Math.PI/4, 10, new BABYLON.Vector3(0, 0, 0), scene);
    let light = new BABYLON.HemisphericLight("light", new BABYLON.Vector3(0, 1, 0), scene);
    light.intensity = 0.7;

    // add some objects

    // not salient objects
    const parent = new BABYLON.TransformNode('parent');
    parent.accessibilityTag = {
        isSalient: true,
        description: "A parent of all, of the root",
    }

    const boxDecs = new BABYLON.TransformNode('boxDecs');
    boxDecs.parent = parent;
    boxDecs.accessibilityTag = {
        isSalient: true,
        description: "A parent without salient children",
    }
    let boxDec1 = BABYLON.MeshBuilder.CreateBox("boxDec1", {size: 0.3}, scene);
    boxDec1.parent = boxDecs;
    boxDec1.position.x = -3;
    boxDec1.position.z = -4;
    let boxDec2 = BABYLON.MeshBuilder.CreateBox("boxDec2", {size: 0.3}, scene);
    boxDec2.parent = boxDecs;
    boxDec2.position.x = 3;
    boxDec2.position.z = -4;

    // salient objects, static
    let boxes = new BABYLON.TransformNode("boxes");
    boxes.parent = parent;
    boxes.accessibilityTag = {
        isSalient: true,
        description: "A group of boxes",
    }
    let box0 = BABYLON.MeshBuilder.CreateBox("box3", {size: 0.5}, scene);
    box0.parent = boxes;
    box0.position.z = -1;
    box0.position.y = 0.6;
    box0.accessibilityTag = {
        isSalient: true,
        description: "A small box in the middle of the scene",
    }
    let box1 = BABYLON.MeshBuilder.CreateBox("box1", {}, scene);
    box1.parent = boxes;
    box1.position.x = 1;
    box1.accessibilityTag = {
        isSalient: true,
        description: "A big box on the left of the small box",
    }
    let box2 = BABYLON.MeshBuilder.CreateBox("box2", {}, scene);
    box2.parent = boxes;
    box2.position.x = -1;
    box2.accessibilityTag = {
        isSalient: true,
        description: "A big box on the right of the small box",
    }

    // salient objects, interactive
    let sphereWrapper = new BABYLON.TransformNode('sphereWrapper');
    sphereWrapper.accessibilityTag = {
        isSalient: true,
        description: 'A group of spheres',
    };
    // sphereWrapper.parent = parent;
    let mat = new BABYLON.StandardMaterial("Gray", scene);
    mat.diffuseColor = BABYLON.Color3.Gray();
    let spheresCount = 6;
    let alpha = 0;
    for (let index = 0; index < spheresCount; index++) {
        const sphere = BABYLON.Mesh.CreateSphere("Sphere " + index, 32, 1, scene);
        sphere.parent = sphereWrapper;
        sphere.position.x = 3 * Math.cos(alpha);
        sphere.position.z = 3 * Math.sin(alpha);
        sphere.material = mat;
        const sphereOnClicked = () => {
            alert(`You just clicked ${sphere.name}!`)
        }
        const sphereOnLeftClicked2 = () => {
            alert(`You just LEFT clicked ${sphere.name}!`)
        }
        const sphereOnRightClicked = () => {
            alert(`You just RIGHT clicked ${sphere.name}!`)
        }
        sphere.accessibilityTag = {
            isSalient: true,
            description: sphere.name,
        };
        sphere.actionManager = new BABYLON.ActionManager(scene);
        sphere.actionManager.registerAction(new BABYLON.ExecuteCodeAction(BABYLON.ActionManager.OnPickTrigger, sphereOnClicked));
        sphere.actionManager.registerAction(new BABYLON.ExecuteCodeAction(BABYLON.ActionManager.OnLeftPickTrigger, sphereOnLeftClicked2));
        sphere.actionManager.registerAction(new BABYLON.ExecuteCodeAction(BABYLON.ActionManager.OnRightPickTrigger, sphereOnRightClicked));
        alpha += (2 * Math.PI) / spheresCount;
    }

    console.log('Start the show!');
    AccessibilityRenderer.renderAccessibilityTree(scene);
    return scene;
};
```

Testing code 2:

```
/**
 * A simple GUI scene. Shows a card GUI for a easter event, with buttons to 'Join' event, and 'close' card.
 */
export let createScene = function ()
{
    let scene = new BABYLON.Scene(engine);
    let camera = new BABYLON.ArcRotateCamera('Camera', -Math.PI/2, Math.PI/2, 5, new BABYLON.Vector3(0, 0, 0), scene);
    let light = new BABYLON.HemisphericLight('light', new BABYLON.Vector3(0, 1, 0), scene);
    light.intensity = 0.7;

    let wrapper = new BABYLON.TransformNode("Wrapper");
    wrapper.setEnabled(true);

    let egg = BABYLON.MeshBuilder.CreateSphere("Egg", {diameterX: 0.62, diameterY: 0.8, diameterZ: 0.6}, scene);
    egg.parent = wrapper;
    egg.accessibilityTag = {
        isSalient: true,
        description: "An easter egg"
    }
    egg.actionManager = new BABYLON.ActionManager(scene);
    egg.actionManager.registerAction(new BABYLON.ExecuteCodeAction(
        BABYLON.ActionManager.OnPickTrigger,
        () => {
            wrapper.setEnabled(false);
            card.setEnabled(true);
        })
    );

    let box1 = BABYLON.MeshBuilder.CreateBox("box1", {size: 0.3}, scene);
    box1.parent = wrapper;
    box1.position.x = 1;
    box1.position.z = 0.2;
    box1.accessibilityTag = {
        isSalient: true,
        description: "A small box on the left of the egg",
        isWholeObject: true,
    }
    let box2 = BABYLON.MeshBuilder.CreateBox("box2", {size: 0.3}, scene);
    box2.parent = wrapper;
    box2.position.x = -1;
    box2.position.z = 0.2;
    box2.accessibilityTag = {
        isSalient: true,
        description: "A small box on the right of the egg",
        isWholeObject: true,
    }

    let box = BABYLON.MeshBuilder.CreateBox("box", {size: 0.5}, scene);
    box.position.y = -0.65;
    box.parent = egg;

    // GUI
    let card = BABYLON.MeshBuilder.CreatePlane('card', {size: 3});
    card.setEnabled(false);
    card.accessibilityTag = {
        isSalient: true,
        description: "Easter Event Card"
    }
    card.position.z = 0.5;
    let adt = GUI.AdvancedDynamicTexture.CreateForMesh(card);

    let wrapFront = new GUI.Rectangle('TeamsCardUI_wrapFront');
    wrapFront.width = '80%';
    wrapFront.background = 'white';
    adt.addControl(wrapFront);

    let thumbnailBg = new GUI.Rectangle('TeamsCardUI_ThumbnailBg');
    thumbnailBg.width = '100%';
    thumbnailBg.height = '40%';
    thumbnailBg.background = 'gray';
    thumbnailBg.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    thumbnailBg.top = 0;
    wrapFront.addControl(thumbnailBg);

    let url = 'https://raw.githubusercontent.com/TNyawara/EasterBunnyGroupProject/master/Assets/Easter_UI/Backgrounds/background_10.png';
    let thumbnailCustomized = new GUI.Image('TeamsCardUI_ThumbnailImage', url);
    thumbnailCustomized.alt = 'Background image';
    thumbnailCustomized.width = '100%';
    thumbnailCustomized.height = '100%';
    thumbnailBg.addControl(thumbnailCustomized);

    const titleFront = new GUI.TextBlock(
    'TeamsCardUIText_Title',
    'Event: Happy Hoppy'
    );
    titleFront.fontSize = 60;
    titleFront.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    titleFront.textHorizontalAlignment = GUI.Control.HORIZONTAL_ALIGNMENT_LEFT;
    titleFront.paddingLeft = '7.5%';
    titleFront.top = '40%';
    titleFront.height = '10%';
    wrapFront.addControl(titleFront);

    let dateFront = new GUI.TextBlock('TeamsCardUIText_Date', 'Every day');
    wrapFront.addControl(dateFront);
    dateFront.fontSize = 40;
    dateFront.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    dateFront.textHorizontalAlignment = GUI.Control.HORIZONTAL_ALIGNMENT_LEFT;
    dateFront.paddingLeft = '7.5%';
    dateFront.top = '50%';
    dateFront.height = '5%';
    dateFront.isEnabled = false;

    const timeFront = new GUI.TextBlock('TeamsCardUIText_Time', '00:00 - 23:59');
    wrapFront.addControl(timeFront);
    timeFront.fontSize = 40;
    timeFront.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    timeFront.textHorizontalAlignment = GUI.Control.HORIZONTAL_ALIGNMENT_LEFT;
    timeFront.paddingLeft = '35%';
    timeFront.top = '50%';
    timeFront.height = '5%';

    const meetingDetail = new GUI.TextBlock(
    'TeamsCardUIText_GroupName',
    "Help the little bunny rabbits get ready for Easter! Look at all the different colors to decorate Easter eggs with and pick out the shapes you'd like to wear in the parade. "
    );
    wrapFront.addControl(meetingDetail);
    meetingDetail.fontSize = 40;
    meetingDetail.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    meetingDetail.textHorizontalAlignment = GUI.Control.HORIZONTAL_ALIGNMENT_LEFT;
    meetingDetail.paddingLeft = '7.5%';
    meetingDetail.top = '55%';
    meetingDetail.height = '30%';
    meetingDetail.width = '100%';
    meetingDetail.textWrapping = GUI.TextWrapping.WordWrapEllipsis;

    let joinButton = GUI.Button.CreateSimpleButton('TeamsCardUIButton_Join', 'Join');
    joinButton.background = 'black';
    joinButton.color = 'white';
    joinButton.fontSize = 40;
    joinButton.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    joinButton.textBlock.textHorizontalAlignment = GUI.Control.VERTICAL_ALIGNMENT_CENTER;
    joinButton.top = '85%';
    joinButton.height = '10%';
    joinButton.width = '40%';
    wrapFront.addControl(joinButton);
    joinButton.onPointerClickObservable.add(() => {
        alert('💐🌼Happy Easter! 🐰🥚');
    });

    let closeButton = GUI.Button.CreateSimpleButton('TeamsCardUIButton_Close', 'X');
    closeButton.background = 'black';
    closeButton.color = 'white';
    closeButton.fontSize = 40;
    closeButton.verticalAlignment = GUI.Control.VERTICAL_ALIGNMENT_TOP;
    closeButton.horizontalAlignment = GUI.Control.HORIZONTAL_ALIGNMENT_RIGHT;
    closeButton.textBlock.textHorizontalAlignment = GUI.Control.VERTICAL_ALIGNMENT_CENTER;
    closeButton.top = '0';
    closeButton.height = '12%';
    closeButton.width = '15%';
    wrapFront.addControl(closeButton);
    closeButton.onPointerClickObservable.add(() => {
        card.setEnabled(false);
        wrapper.setEnabled(true);
    });

    console.log('Start the show!');
    AccessibilityRenderer.RenderAccessibilityTree(scene);
    return scene;
}
```


carolhmj:
+1 for AWESOME FEATURE! I'll love to see it in next release 😄

sebavan:
I love it a lot !!!! the main issue I am seeing is that it requires the full inspector which will turn out to force loading all of Babylon creating a huge dependency. I bet @RaananW can help with this part as this component should be easily embeddable without pulling all of the framework.

RaananW:
Sorry, didn't find time for review today, I'll go over this tomorrow.

deltakosh:
No rush, we will not merge until after the release

deltakosh:
I guess we can think about continuing this thread

mysunnytime:
I updated it so it's now in its own package, and it support GUI (TextBlock and Button only for now). It can now also update when scene updates (add/remove node and GUI control, enable/disable node, visible/invisible GUI control). I think it's in a good shape to be reviewed :) @RaananW, @sebavan, @deltakosh

deltakosh:
Please bear with us as @RaananW is OOF and I'm expecting him to give us the green light :)

deltakosh:
Some questions on my side:
- Why using GUI? I believe the goal is to generate DOM elements right?
- DO you see this as part of the framework ? (I see it is decorrelated here). THen it will be an independent NPM package?

packages/dev/gui/src/2D/controls/container.ts
```diff
@@ -208,6 +209,9 @@ export class Container extends Control {
         this._reOrderControl(control);
 
         this._markAsDirty();
+
+        this.onControlAddedOrRemovedObservable.notifyObservers(control);
```
deltakosh, line 213: I would prefer having OnAdd and another OnRemoved

packages/dev/core/src/node.ts
```diff
@@ -974,3 +981,18 @@ export class Node implements IBehaviorAware<Node> {
         };
     }
 }
+
+// TODO: refactor this so the change of this tag can be observed.
+/**
+ * Define an interface for a node to indicate it's info for accessibility.
+ */
+ export interface IAccessibilityTag {
```
deltakosh, line 989: Please make it an independent file :)

packages/dev/core/src/node.ts
```diff
@@ -588,10 +596,9 @@ export class Node implements IBehaviorAware<Node> {
             return;
         }
         this._nodeDataStorage._isEnabled = value;
+        this._syncParentEnabledState();
```
deltakosh, line 599: Why this move?

mysunnytime: Oh, this is because when some object is enable/disabled, but that hasn't sync up to its children at the time _onEnabledStateChangedObservable notify observers, the accessibility updates (which observes it) won't get the correct tree.

packages/dev/core/src/node.ts
mysunnytime: Sure!

packages/dev/core/src/node.ts
deltakosh: Makes sense!

packages/dev/gui/src/2D/controls/container.ts
mysunnytime: Changed to OnAdd and OnRemoved!

packages/dev/core/src/node.ts
```diff
@@ -125,6 +125,14 @@ export class Node implements IBehaviorAware<Node> {
      */
     public inspectableCustomProperties: IInspectable[];
 
+    /**
+     * Gets or sets the accessibility tag to describe the mesh for accessibility purpose.
+     */
+    public accessibilityTag?: IAccessibilityTag = {
```
RaananW, line 131: instead of initializing it with a false value, maybe this should be Nullable and set to null? this way you don't need the isSalient flag at all

packages/dev/core/src/node.ts
RaananW, line 989: I believe this interface should look a little different.

As I mentioned in a different comment - the isSalient flag is redundant - if there is no accessibility tag this node should not be considered "accessible".
It should also extend to more than description. For example, it could have an `aria` object that will support the ARIA standard and will allow defining labels, live, ans so on. Something like this:

```javascript
{
  description?: string;
  role?: string;
  aria?: {
    label?: string;
    live?: string;
    relevant?: string;
    // further aria that we want to support
  }
}
```

this object fits 1-to-1 with an HTML object, so you could technically extend an HTML element with these details. You could argue that instead of an `aria` object we should have the actual tag names (like `aria-label` and so on), and that would be fine as well.
This way you provide the developer with everything they need to make the scene fully accessible according to their standards.

This could also be a simple `{[key:string]: string}` map that has its keys automatically added to the corresponding HTML element. So setting:

```javascript
{
  "aria-label": "This is a label",
  "aria-live": "polite"
}
```

will generate the HTML:

```html
<div aria-label="This is a label" aria-live="polite"/> 
```

packages/dev/core/src/node.ts
RaananW, line 131: A node should also notify when this element has changed so that the implementation knows it needs to update the HTML.
TBH I am not sure if the node itself is the right place or the actual implementation should cache this and check for changes, but my gut feeling is telling me it should happen in the node, similar to the way the node is caching other values and checks them. This way the HTML implementation could also know if there is an already-existing node that has its accessibility tag set.

RaananW (commented):
Oh, this is awesome!!! Can't wait to have it as part of the framework :-)

I have a few general questions/suggestions (and will be happy to discuss them all)

1) I think the better place for this would be the `tools` directory and not the `dev` directory, as this is a tool using the core libs.
2) The role definition of the component holding the data can be a bit confusing. Both Control and Node should have the accessibility tag added to them, allowing the dev to extend the corresponding (HTML) elements. Those tags can be cached and updated in the React implementation when changed. I wrote in one of the comments that I think the cache and "changed" functionality should be a part of the node/control (or maybe even its own "accessibility class" that will take care of that for both classes as part of core?). This way the dev could change the a11y data when the app is already running and will always see the changes propagate to the react implementation (or any other implementation that will use this API).
3) You are using a lot of instanceof, which can be a little unreliable (and a bit slow, though the differences are not that extreme). If possible, try using feature-detection or use the getClassName function (if possible, of course). 

A bit philosophical, but I think that, in general, screen readers don't know what 3D content is. They understand 2D very well, and this is a kind of 3D-to-2D converter. My question here is whether the order (and "level") should be controlled by something other than the tree-structure? it makes sense in GUI (which is "2D" in its nature), but might not fit the node implementation. TBH - I don't have a good answer as to what is better here, but will be happy to talk about it :-)

And in general - I believe the React implementation should use the provided observables in a more dynamic way instead of reading the entire scene structure on every change. So when a node (or control) is added, a new component should be created, and this component should attach to the corresponding node/control's observables to decide whether or not its a11y tag should be rendered. In this case it would also make sense to have a "onA11YTagUpdatedObservable` (and maybe added/removed as well? though "changed" might be enough) and attach to it. This way changes in the scene graph will automatically propagate to the react implementation in a more performant way (instead of checking all nodes for changes on each frame).

packages/dev/core/src/node.ts
mysunnytime: Interesting! That's a good call! I'll look into that more and see how to incorporate it into the code.

packages/dev/core/src/node.ts
mysunnytime: Good idea! Let me change it.

mysunnytime:
> 1. I think the better place for this would be the `tools` directory and not the `dev` directory, as this is a tool using the core libs.

Sure I can do that!

mysunnytime:
> 3. You are using a lot of instanceof, which can be a little unreliable (and a bit slow, though the differences are not that extreme). If possible, try using feature-detection or use the getClassName function (if possible, of course).

Initially I was using 'getClassName', but it failed to get correct result, especially with polymorphism - but instancof always works, so I changed to instanceof. What is feature-detection?

mysunnytime:
@RaananW 
> A bit philosophical, but I think that, in general, screen readers don't know what 3D content is. They understand 2D very well, and this is a kind of 3D-to-2D converter. My question here is whether the order (and "level") should be controlled by something other than the tree-structure? it makes sense in GUI (which is "2D" in its nature), but might not fit the node implementation. TBH - I don't have a good answer as to what is better here, but will be happy to talk about it :-)

Good question! Like what I commented for answering David's question, I think there can be multiple renderers to render. The one we have is the most basic, like you said, 3D->2D flattening renderer. Later we can expand it to renderer that only render objects that the user can 'see' at current moment, like a scene scanner. And it can order things align to the expected viewing order (from center to peripheral).

mysunnytime:
@RaananW 
> And in general - I believe the React implementation should use the provided observables in a more dynamic way instead of reading the entire scene structure on every change. So when a node (or control) is added, a new component should be created, and this component should attach to the corresponding node/control's observables to decide whether or not its a11y tag should be rendered. In this case it would also make sense to have a "onA11YTagUpdatedObservable` (and maybe added/removed as well? though "changed" might be enough) and attach to it. This way changes in the scene graph will automatically propagate to the react implementation in a more performant way (instead of checking all nodes for changes on each frame).

I just added the 'onA11YTagUpdatedObservable' to node (named 'onAccessibilityTagChangedObservable'). Feeling changed observable is enough, no need add/remove tag observable, if it's just assigning 'Null' to the tag.

RaananW:
> Yea, I do feel it's not consistent that node has a a11y tag, but the controls don't. I chose to do that, because the controls are already very close to DOM element. Like button, image and textblock, themself already strongly imply how they should appear in the a11y tree, so having another a11y tag feels redundant. But indeed, we can also add the a11y tag to controls, and let them automatically get value from controls existing fields - how does that sound? Although, I don't understand when you say "tags can be cached". What is it to do with cache?
> 

I should have explained better about caching 😊. If the dev changes a value inside the a11y tag the observable should notify of the change. The cache needs to check if any value has changed inside this tag. The tag itself can be a more complex object and do the notification by itself (for example by making the tag a class and using getters and setters for the values), but that might be too much for an object that should be simple.

RaananW:
> 
> Initially I was using 'getClassName', but it failed to get correct result, especially with polymorphism - but instancof always works, so I changed to instanceof. What is feature-detection?
> 
> 
If instanceof is the only thing that works it's fine. Feature detection is, for example, checking if the object has a function defined that only an abstract mesh and its children has. This way you can be due that this object is an abstract mesh without checking the instance itself. It might not be possible in different cases, so instanceof is safer.

packages/tools/accessibility/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "@dev/accessibility",
+    "name": "@tools/accessibility",
     "description": "Render accessibility tree for 3D content in DOM, thus the 3D content is compatible with screen readers and tab selection",
     "private": false,
```
RaananW, line 4: should probably stay private, as this will not be released to npm.

RaananW: resurfacing this - the PR will not be merged before this is set to true, otherwise npm will try releasing this during the publish process.

packages/tools/accessibility/src/accessibilityGUIItem.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * If this entity is actionable (can be clicked).
+     */
+    public override get isActionable(): boolean {
+        if (this._isActionable) {
+            return this._isActionable;
+        }
+
+        if (this.entity instanceof Button) {
```
carolhmj, line 55: Could this be set in the class constructor instead of the get?

packages/tools/accessibility/src/accessibilityGUIItem.ts
```diff
@@ -0,0 +1,113 @@
...
+     * If this entity is focusable (can be focused by tab key pressing).
+     */
+    public override get isFocusable(): boolean {
+        // TODO: change back
+        // return true;
+        if (this._isFocusable) {
+            return this._isFocusable;
+        }
+
+        if (this.entity instanceof Button) {
```
carolhmj, line 74: Same idea as _isActionable

packages/tools/accessibility/src/accessibilityNodeItem.ts
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        let actions: IAction[] = [];
+        actions.push(...this._getTriggerActions(this.entity, Constants.ACTION_OnRightPickTrigger));
+        actions.push(...this._getTriggerActions(this.entity, Constants.ACTION_OnPickTrigger));
+
+        actions.forEach((action) => {
+            action._executeCurrent();
+        });
+    }
```
RaananW, line 131: Feels like this is repeating code that can be moved to a single private function with different parameters.

packages/tools/accessibility/src/accessibilityTreeComponent.tsx
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    private _isGUI(node: Node) {
+        let isGUI = false;
+        if (node instanceof AbstractMesh) {
+            let curMesh = node as AbstractMesh;
+            let textures = curMesh.material?.getActiveTextures();
+            if (textures && textures[0] instanceof AdvancedDynamicTexture) {
+                isGUI = true;
+            }
```
RaananW, line 228: nit
```suggestion
            isGUI = !!textures && textures[0] instanceof AdvancedDynamicTexture;
```

packages/tools/accessibility/src/accessibilityGUIItem.ts
```diff
@@ -0,0 +1,147 @@
...
+    public override blur(): void {
+        // If defined eventHandler, override default.
+        let eventHandler = (this.entity as Control).accessibilityTag?.eventHandler;
+        if (eventHandler?.onblur) {
+            eventHandler.onblur();
+            return;
+        }
+
+        const control = this.entity as Control;
+        control.isHighlighted = false;
```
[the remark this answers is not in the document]
mysunnytime: Yea, if onblur is defined (customized), we do customized version and return.

packages/tools/accessibility/src/accessibilityGUIItem.ts
```diff
@@ -0,0 +1,147 @@
...
+    public override click(): void {
+        // If defined eventHandler, override default.
+        let eventHandler = (this.entity as Control).accessibilityTag?.eventHandler;
+        if (eventHandler?.onclick) {
+            eventHandler.onclick();
+            return;
+        }
+
+        if (!this.isActionable) return;
+        this.entity.onPointerClickObservable.notifyObservers(new Vector2WithInfo(new Vector2()));
```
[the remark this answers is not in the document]
mysunnytime: Good question! That's also a question I have. the notifyOvservers require this parameter, but I checked the places that used this and seems it needs "coordinates" and "buttonIndex", but actually don't know how to get them in my case. How should I fill in this?

packages/dev/gui/src/2D/controls/container.ts
```diff
@@ -223,6 +227,7 @@ export class Container extends Control {
             this.removeControl(child);
         }
 
+        this.onControlRemovedObservable.notifyObservers(null);
```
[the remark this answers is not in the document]
mysunnytime: Nice catch!

packages/tools/accessibility/src/accessibilityGUIItem.ts
mysunnytime: It's doing a lazy instanciation.

package-lock.json
```diff
@@ -4510,7 +4514,7 @@
             }
         },
         "node_modules/@storybook/builder-webpack4/node_modules/html-minifier-terser/node_modules/terser": {
-            "version": "4.8.1",
+            "version": "4.8.0",
```
carolhmj, line 4517: Should package-lock be merged?

packages/public/@babylonjs/html-twin-renderer/webpack.config.js
```diff
@@ -0,0 +1,13 @@
+const commonConfigGenerator = require("@dev/build-tools").webpackTools.commonUMDWebpackConfiguration;
+
+module.exports = (env) => {
+    const commonConfig = commonConfigGenerator({
+        mode: env.production ? "production" : "development",
+        devPackageName: "html-twin-renderer",
+        devPackageAliasPath: `../../../tools/accessibility/dist`,
+        // namespace: "HTMLTWINRENDERER",
```
carolhmj, line 8: Can the comment be removed?

packages/tools/accessibility/src/accessibilityNodeItem.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        return this._isActionable;
+    }
+
+    /**
+     * If this entity is focusable (can be focused by tab key pressing).
+     */
+    public override get isFocusable(): boolean {
+        return true;
```
carolhmj, line 49: Why are we always returning true here?

packages/tools/accessibility/src/accessibilityTreeComponent.tsx
```diff
@@ -0,0 +1,242 @@
...
+    constructor(props: IAccessibilityTreeComponentProps) {
+        super(props);
+        const a11yTreeItems = this._updateAccessibilityTreeItems();
+        this.state = {
+            a11yTreeItems: a11yTreeItems,
+        };
+    }
+
+    componentDidMount() {
+        console.log("componentDidMount");
```
carolhmj, line 35: Can we remove the log?

packages/tools/accessibility/src/accessibilityTreeComponent.tsx
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        // observe node enabled changed
+        scene.getNodes().forEach((node) => {
+            addNodeObservers(node);
+        });
+    }
+
+    componentWillUnmount() {
+        console.log("componentWillUnmount");
```
carolhmj, line 142: Can we remove the log here?

packages/tools/accessibility/src/accessibilityTreeComponent.tsx
```diff
@@ -0,0 +1,242 @@
...
+            </div>
+        );
+    }
+
+    private _updateAccessibilityTreeItems(): AccessibilityItem[] {
+        // Nodes
+        const rootNodes = this.props.scene.rootNodes.slice(0);
+        for (const mesh of this.props.scene.meshes) {
+            // Adding nodes that are parented to a bone
+            if (mesh.parent && mesh.parent.getClassName() === "Bone") {
```
carolhmj, line 170: Why this special case?

carolhmj (changes requested):

packages/tools/accessibility/src/accessibilityTreeComponent.tsx
mysunnytime: Nice catch!

packages/tools/accessibility/src/accessibilityNodeItem.ts
mysunnytime: Right! Need to revert it.

packages/tools/accessibility/src/accessibilityTreeComponent.tsx
mysunnytime: Learned this from Inspector https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/inspector/src/components/sceneExplorer/sceneExplorerComponent.tsx#L434

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
+This package renders accessibility tree (html twins of the scene objects) for 3D content in DOM, thus the 3D content is compatible with screen readers and keyboard navigation.
+
+## What is Screen Reader?
+
+[Screen reader](https://en.wikipedia.org/wiki/Screen_reader) is an assistive technology for blind or low vision people to interact with digital content. For users with blindness or low vision, they can't rely 100% on visual interface. Depending on the degree of visual disabilities, some of they don't even use a monitor display or mouse when they use computers. So, for this kind of user, they use screen reader, a software, to read what are displayed visually on the screen, through speech, or through braille output. They use keyboard to tell the screen reader what to read and interact with the computer. Here's an example video of how a user uses screen reader: [Screen Reader Demo - YouTube](https://www.youtube.com/watch?v=q_ATY9gimOM&ab_channel=SLCCUniversalAccess).
```
PatrickRyanMS, line 5: "...visual disabilities, some of they don't..." should be "...visual disabilities, some of **them** don't..."

PatrickRyanMS, line 5: "So, for this kind of user, they use screen reader, a software, to read what are displayed visually on the screen, through speech, or through braille output." Simplifying sentence to reduce commas:

"This user would use screen reading software to translate what is displayed on screen into speech or braille output."

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## What is Screen Reader?
+
+[Screen reader](https://en.wikipedia.org/wiki/Screen_reader) is an assistive technology for blind or low vision people to interact with digital content. For users with blindness or low vision, they can't rely 100% on visual interface. Depending on the degree of visual disabilities, some of they don't even use a monitor display or mouse when they use computers. So, for this kind of user, they use screen reader, a software, to read what are displayed visually on the screen, through speech, or through braille output. They use keyboard to tell the screen reader what to read and interact with the computer. Here's an example video of how a user uses screen reader: [Screen Reader Demo - YouTube](https://www.youtube.com/watch?v=q_ATY9gimOM&ab_channel=SLCCUniversalAccess).
+
+There are different screen reader applications that a user can choose, including OS built-in screen readers like Windows's "Narrator", MacOS and iOS's "VoiceOver", and 3-rd party screen readers like JAWS, NVDA.
```
PatrickRyanMS, line 7: "3-rd party" should be "third-party"

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+There are different screen reader applications that a user can choose, including OS built-in screen readers like Windows's "Narrator", MacOS and iOS's "VoiceOver", and 3-rd party screen readers like JAWS, NVDA.
+
+Nowadays 2D pages on web generally have good accessibility support to screen reader users, because it can understand different HTML elements. But 3D applications like webGL applications is not. This is because objects in webGL applications are rendered in a <canvas> element. When screen reader read the page, it will only read: "*Image*", but can't interpret any objects inside the scene. If we do not deal with it correctly, users who are blind or low vision will have difficulty using the application.
```
PatrickRyanMS, line 9: "But 3D applications like webGL applications is not." should be:
"But 3D applications like webGL applications **do** not." for subject/predicate agreement

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+### RenderAccessibilityTree()
+
+The accessibility package basically generates HTML elements for your scene, based on the metadata (IAccessibilityTag) you added to your scene content. The screen reader then can read the HTML elements. To generate it for your scene:
+
+```javascript
+ACCESSIBILITY.AccessibilityRenderer.RenderAccessibilityTree(scene);
+```
+
+This will generate a `<div id="accessibility-tree">` HTML element right after the babylonJS scene's canvas element. Inside this div element, the renderer generate an HTML twin for each of your accessible contents in the scene. These HTML twin elements are internally connected with BabylonJS objects, so the screen reader user can also interact with the BabylonJS objects through their HTML twin.
```
PatrickRyanMS, line 41: "...the renderer generate an HTML twin..." the verb should be for a singular subject:

"...the renderer **generates** an HTML twin..."

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Interaction
+
+Some of your contents in the scene might be interactable (e.g. clickable). For Node type objects, if you use BabylonJS's [ActionManager](https://doc.babylonjs.com/divingDeeper/events/actions) to define the interaction, the Accessibility package can automatically detects it and apply on the generated HTML twin elements, so that the user can use keyboard to trigger events like click or right click on the HTML twin elements, thus trigger the correspond action on the BabylonJS objects. Only ACTION_OnPickTrigger, ACTION_OnLeftPickTrigger, ACTION_OnRightPickTrigger are supported. For Control type objects, if you defined observer for onPointerClickObservable, the Accessibility package can also automatically detects it and apply on the generated HTML twin element.
```
PatrickRyanMS, line 45: "BabylonJS's" should be "Babyon.js's", anywhere you use "BabylonJS" we should change to "Babylon.js" for branding consistency.
and
"...Accessibility package can automatically detects it and apply..." should be 
"...Accessibility package can automatically **detect** it and apply..."
and 
"...onPointerClickObservable, the Accessibility package can also automatically detects it and apply..." should be 
"...onPointerClickObservable, the Accessibility package can also automatically **detect** it and apply..."

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+Your scene might be not static, and you may want to update the accessibility tree when your scene is changed. The accessibility tree will automatically update when:
+
+- A Node (Mesh or TransformNode) is added/removed in a scene;
+- A Node (Mesh or TransformNode)'s enabled status is changed;
+- A Control is added/removed from a Container;
+- A Control's isVisible status is changed;
+- A Node or Control's IAccessibilityTag is assigned or re-assigned;
+
+Example:
+https://playground.babylonjs.com/?snapshot=refs/pull/12074/merge#C9GZTF#5
```
PatrickRyanMS, line 71: this will need to be converted to a real playground link once the PG is updated the accessibility is live.

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Customize HTML Twin with ARIA Attributes
+If you are a pro in Web Accessibility, and know what you are doing, you can use IAccessibilityTag.role and IAccessibilityTag.aria to assign different [Role and ARIA attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) to the HTML Twin of this object to do whatevery you want.
```
PatrickRyanMS, line 74: "...object to do whatevery you want." has a typo and should be:
"...object to do **whatever** you want."

packages/tools/accessibility/README.md
mysunnytime: Great! Succinct!

carolhmj (approved):
🥳

sebavan:
❤️‍🔥

packages/tools/accessibility/README.md
```diff
@@ -0,0 +1,88 @@
...
+let egg = BABYLON.MeshBuilder.CreateSphere("Egg", {diameterX: 0.62, diameterY: 0.8, diameterZ: 0.6}, scene);
+egg.accessibilityTag = {
+    description: "An easter egg",
+    eventHandler: {
+      "onclick": yourFunction
+    }
+}
+```
+Example:
+https://playground.babylonjs.com/?snapshot=refs/pull/12074/merge#C9GZTF#6
```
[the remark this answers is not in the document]
RaananW: TODO

PR #12091, opened 2022-03-04
[GE] Move GUI and Artboard settings to command bar

darraghjburke:
Moves the GUI size and artboard color settings to the command bar to match with the design.

Old layout:
image

New layout:
image

Also fixes some font issues with the properties panel.


RaananW:
Waiting for the conflicts and will merge right after

PR #12110, opened 2022-03-07
Simple easy fix for simplex math singularity at origin

syntheticmagus:
The simplex math that drives one of the NME noise nodes has a singularity at the origin. The easiest way to solve this is to just add a tiny offset if the input point is detected to be exactly zero.

### DIFF
--- a/src/Materials/Node/Blocks/simplexPerlin3DBlock.ts
+++ b/src/Materials/Node/Blocks/simplexPerlin3DBlock.ts
@@ -85,6 +85,7 @@ export class SimplexPerlin3DBlock extends NodeMaterialBlock {
         functionString += `const float SIMPLEX_CORNER_POS = 0.5;\r\n`;
         functionString += `const float SIMPLEX_TETRAHADRON_HEIGHT = 0.70710678118654752440084436210485;\r\n`;
         functionString += `float SimplexPerlin3D( vec3 P ){\r\n`;
+        functionString += `    P.x = P.x == 0. ? 0.00001 : P.x;\r\n`;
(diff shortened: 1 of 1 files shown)


src/Materials/Node/Blocks/simplexPerlin3DBlock.ts
```diff
@@ -85,6 +85,7 @@ export class SimplexPerlin3DBlock extends NodeMaterialBlock {
         functionString += `const float SIMPLEX_CORNER_POS = 0.5;\r\n`;
         functionString += `const float SIMPLEX_TETRAHADRON_HEIGHT = 0.70710678118654752440084436210485;\r\n`;
         functionString += `float SimplexPerlin3D( vec3 P ){\r\n`;
+        functionString += `    P.x = P == vec3(0., 0., 0.) ? 0.00001 : P.x;\r\n`;
```
sebavan, line 88: is there a way to clamp or use something else than a comparison ?

Popov72: Maybe:
```glsl
P.x = mix(0.00001, P.x, step(0.0, dot(P, P)));
```

deltakosh (approved):

src/Materials/Node/Blocks/simplexPerlin3DBlock.ts
sebavan: Branching or what @Popov72 proposed will have the same cost more or less has one side is a constant and the other a simple value so in this case branching might even be faster than dot and mix :-) I looked deeper and I do not think there is a way to simply clamp, thought P was a color for some reasons so always positive....

sebavan (approved):

PR #12121, opened 2022-03-08
Fix WebGPU Compute on Meshes

sebavan:
Trying to fix https://playground.babylonjs.com/#3URR7V#142 I noticed data may not be available for storage buffers


carolhmj (approved):
AWESOME! Thanks so much Seb!

PR #12124, opened 2022-03-08
Fix gltf extras with new skinning pattern

bghgary:
See: https://forum.babylonjs.com/t/extrasasmetadata-bug/28097


sebavan (approved):
Nice one !!!

PR #12125, opened 2022-03-09
Added Load Animation From Snippet Template

PirateJC:


PirateJC:
Did not update the what's new since it's just a new playground template. Let me know if we think it's needed.

Playground/templates.json
```diff
@@ -16,9 +16,14 @@
     },
     {
         "label": "Load GPU Particle System from Snippet Server",
-        "documentation": "https://doc.babylonjs.com/babylon101/particles#snippet-server",
+        "documentation": "https://doc.babylonjs.com/toolsAndResources/tools/animationCurveEditor#load-from-the-snippet-server",
```
RaananW, line 19: is that the right link? it documents loading from the server, but not a particle system

PR #12127, opened 2022-03-09
Fix VideoDome Dispose

sebavan:
https://forum.babylonjs.com/t/videodome-clicktoplay-and-dispose-maybe-not-as-expected/28271/2


kaliatech:
👍

deltakosh (approved):

PR #12128, opened 2022-03-09
Allow collapsed frames to be selected and dragged in NME

carolhmj:
Without the change
nme-frame-selection-2
With the change
nme-frame-selection-1


deltakosh (approved):
Excellent!

sebavan (approved):
Nice one !!!!

PR #12132, opened 2022-03-09
Fix exception when cloning a skinned mesh with skeleton.needInitialSkinMatrix is true

bghgary:
See: https://forum.babylonjs.com/t/meshes-with-skeletons-that-have-needinitialskinmatrix-true-crash-when-cloned/28247


sebavan (approved):
Only cause I trust you as I totally have no clue :-)

bghgary:
I'm not super confident on this one, so maybe we can go over it. Or maybe @deltakosh can check it also.

deltakosh (approved):
looks good to me

PR #12144, opened 2022-03-11
Simple Update in AxesViewer to accept line thickness arg

mrlooi:


sebavan (approved):
missing the comments in the @param list but looks good to me

RaananW (approved):

RaananW:
it does feel like the constructor is geeing bigger than we would want it to be, but there is currently no way out of this.

mrlooi:
> @mrlooi - there are docs issues and linting issues. care to check those? Thanks :-)

Oh I edited this directly via github, how do I check the linting stuff? Might need some help here

RaananW:
It is just missing the documentation for the parameter lineThickness

src/Debug/axesViewer.ts
```diff
@@ -52,7 +52,7 @@ export class AxesViewer {
      * @param yAxis defines the node hierarchy used to render the y-axis
      * @param zAxis defines the node hierarchy used to render the z-axis
      */
```
RaananW, line 54:
```suggestion
     * @param lineThickness The line thickness to use when creating the arrow. defaults to 1.
     */
```

RaananW: @mrlooi - You can merge this, unless you want to write your own documentation line :-)

PR #12149, opened 2022-03-11
Gltf materials

sebavan:
Add a way to either only download the gltf materials or to skip all of them.


deltakosh (approved):

loaders/src/glTF/glTFFileLoader.ts
```diff
@@ -238,6 +238,16 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
      */
     public loadAllMaterials = false;
 
+    /**
+     * If true, load only all the materials defined in the file. Defaults to false.
```
bghgary, line 242:
```suggestion
     * If true, load only the materials defined in the file. Defaults to false.
```

loaders/src/glTF/glTFFileLoader.ts
```diff
@@ -238,6 +238,16 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
...
 
+    /**
+     * If true, load only all the materials defined in the file. Defaults to false.
+     */
+    public loadOnlyMaterials = false;
+
+    /**
+     * If true, do not load any materials defined in the file. Defaults to false.
+     */
+     public skipMaterials = false;
```
bghgary, line 249:
```suggestion
    public skipMaterials = false;
```

bghgary (approved):
A couple of nits. I'm a bit concerned we are adding more and more options without doing to work to make setting options easier, but it should be okay for now.

loaders/src/glTF/glTFFileLoader.ts
```diff
@@ -238,6 +238,22 @@ export class GLTFFileLoader implements IDisposable, ISceneLoaderPluginAsync, ISc
...
+    /**
+     * If true, do not load any materials defined in the file. Defaults to false.
+     */
+    public skipMaterials = false;
+
+    /**
+     * If defined, provides a callback to override a specific material from a GLTF file.
+     * If the callback returns null, the requested gltf material will be loaded.
+     */
+    public customMaterialLoad: Nullable<(material: GLTF2.IMaterial) => Nullable<Material>> = null;
```
bghgary, line 255: This is the same as adding a glTF loader extension?

bghgary (changes requested):

PR #12158, opened 2022-03-14
Add warnings and early return when trying to create a mesh impostor i…

carolhmj:
…n an object without vertices. Currently, when doing just this, we end up with an undefined use exception, which isn't informative to users:
image

This will make easier to understand the usage error:
image


sebavan (approved):

RaananW (approved):
@CedricGuillemet , should this be added to ammo as well?

CedricGuillemet:
Ammo and Cannon behave differently. Ammo will find all children.

PR #12162, opened 2022-03-14
Bone look controller GLTF

sebavan:
https://forum.babylonjs.com/t/babylon-boneikcontroller-with-gltf-mesh/23391/55


RaananW (approved):
LGTM! Just wonder why the `const bone = this.bone;` is needed :-)

sebavan:
5 "this." saved lol but yes agree it does not do much...

PR #12164, opened 2022-03-15
DeviceInputSystem: Fix callbacks for NativeDeviceInputSystem

PolygonalSun:
This PR contains changes to how the `onDeviceConnected`, `onDeviceDisconnected`, and `onInputChanged` are initialized.  Now, these callbacks are passed from the DeviceSourceManager to the DeviceInputSystem via the constructor.  This change was necessary because there was an issue where a device was being connected _before_ `onDeviceConnected` was properly updated.  This was causing already connected devices to never get properly registered with all necessary Observables.  Now, with these changes, all callbacks are required to be set up as part of the constructor, making it impossible to connect a device before the callbacks are set.

Also, the PROTOCOL_VERSION has been updated to 5.


src/DeviceInput/InputDevices/deviceTypes.ts
```diff
@@ -10,4 +10,9 @@ export type DeviceInput<T extends DeviceType> =
...
     T extends DeviceType.Switch ? SwitchInput :
     T extends DeviceType.DualSense ? DualSenseInput :
-    never;
\ No newline at end of file
+    never;
+
+    /**
+     * Type to define functions that handle connect/disconnect fo devices
+     */
+export type DeviceStatusChangedCallback = (deviceType: DeviceType, deviceSlot: number) => void;
```
bghgary, line 18: I know I suggested this when we talked, but I'm not sure it makes sense to do this if we can't do it to all 3 callbacks. Maybe it would make more sense to add this once the native contact catches up.

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -14,27 +15,7 @@ const MAX_POINTER_INPUTS = Object.keys(PointerInput).length / 2;
...
-        }
-    }
-
-    public get onDeviceConnected(): (deviceType: DeviceType, deviceSlot: number) => void {
-        return this._onDeviceConnected;
-    }
-
+    public onDeviceConnected: (deviceType: DeviceType, deviceSlot: number) => void;
     public onDeviceDisconnected: (deviceType: DeviceType, deviceSlot: number) => void;
     public onInputChanged: (deviceType: DeviceType, deviceSlot: number, eventData: IUIEvent) => void;
```
bghgary, line 20: Does this need to be public anymore?

src/DeviceInput/InputDevices/nativeDeviceInputSystem.ts
```diff
@@ -1,34 +1,23 @@
...
-        this._nativeInput.onInputChanged = (deviceType, deviceSlot, inputIndex, currentState) => {
+    public constructor(onDeviceConnected: (deviceType: DeviceType, deviceSlot: number) => void, onDeviceDisconnected: (deviceType: DeviceType, deviceSlot: number) => void, onInputChanged: (deviceType: DeviceType, deviceSlot: number, eventData: IUIEvent) => void) {
+        this._nativeInput = (_native.DeviceInputSystem) ? new _native.DeviceInputSystem(onDeviceConnected, onDeviceDisconnected, (deviceType, deviceSlot, inputIndex, currentState) => {
             const idx = (inputIndex === NativePointerInput.Horizontal || inputIndex === NativePointerInput.Vertical || inputIndex === NativePointerInput.DeltaHorizontal || inputIndex === NativePointerInput.DeltaVertical) ? PointerInput.Move : inputIndex;
             const evt = DeviceEventFactory.CreateDeviceEvent(deviceType, deviceSlot, idx, currentState, this);
 
-            this.onInputChanged(deviceType, deviceSlot, evt);
-        };
+            onInputChanged(deviceType, deviceSlot, evt);
+        }) : this._createDummyNativeInput();
```
bghgary, line 20: Missed this earlier. `_createDummyNativeInput` should be updated to not include the removed events.

bghgary (approved):
LGTM

src/DeviceInput/InputDevices/deviceTypes.ts
PolygonalSun: Removed and will address at a later time

src/DeviceInput/InputDevices/webDeviceInputSystem.ts
PolygonalSun: Changed to private

src/DeviceInput/InputDevices/nativeDeviceInputSystem.ts
PolygonalSun: Removed from function and removed INativeInput

PR #12166, opened 2022-03-15
Fit crop viewport around selected controls

carolhmj:


sebavan (approved):

darraghjburke (changes requested):
This does not seem to work as expected if the control is not positioned at 0, 0.
fit to window

carolhmj:
Thanks for catching that @darraghjburke ! I tried a fix with non-centralized positions

carolhmj:
@darraghjburke branch algorithm updated to work on rotated controls (finally got the matrices!) and for the new build system :D

darraghjburke (approved):
excellent!

PR #12176, opened 2022-03-16
Additional WebXR image tracking changes for native integration

Alex-MSFT:
This PR includes some more changes for WebXR Image Tracking integration for BabylonNative. and some cleanup from the previous change: https://github.com/BabylonJS/Babylon.js/pull/12097

The primary change is to  update nativeXRFrame to pull the results array directly from the frame object rather than marshalling back to native code every time.  This should improve per-frame CPU load.

See: https://github.com/BabylonJS/BabylonNative/pull/1032 for the consuming BabylonNative change.


packages/dev/core/src/XR/features/WebXRImageTracking.ts
```diff
@@ -132,14 +130,6 @@ export class WebXRImageTracking extends WebXRAbstractFeature {
         return super.detach();
     }
 
-    /**
-     * Check if the needed objects are defined.
-     * This does not mean that the feature is enabled, but that the objects needed are well defined.
-     */
-    public isCompatible(): boolean {
```
ryantrem, line 139: Was this something you previously added that is no longer needed or something?

Alex-MSFT: This was added in the original implementation from Raanan, however this is just an interface which doesn't actually need to be implemented as a hardened JS class.

We also already check for whether getImageTrackingResults and getTrackedImageScores are implemented by the WebXR side before attempting to use the feature, so this is kind of just an extra safety check that isn't actually explicitly needed.

For these reasons I decided to remove it.  If we want to keep this here to prevent attempting to initialize the feature then I'd need to just define a dummy class somewhere for the native side since we don't actually use this anywhere natively.

Alex-MSFT: For other WebXR features we don't explicitly check the classes when attaching like we are doing in this case.

RaananW: This was added when it was in beta mode and was not available in most ar-supporting browsers. This has changed since then. There is also the extra check of "required features".

This is just an initial test to see if it is even viable to enable this feature. Is there a different object you can check instead?

ryantrem (approved):

RaananW:
Is this ready to be merged now?

Alex-MSFT:
> Is this ready to be merged now?

I believe so, as long as you are okay with taking out the IsCompatible check for turning this feature on.  Basically the only cost there is we will load the image for cases where it might not be necessary.

packages/dev/core/src/Engines/engine.ts
```diff
@@ -298,7 +298,7 @@ export class Engine extends ThinEngine {
      * @param options An object that sets options for the image's extraction.
      * @returns ImageBitmap.
      */
-    public createImageBitmapFromSource(imageSource: string, options?: ImageBitmapOptions): Promise<ImageBitmap> {
+    public _createImageBitmapFromSource(imageSource: string, options?: ImageBitmapOptions): Promise<ImageBitmap> {
```
RaananW, line 301: This breaks backwards compatibility. Any reason for the change in the function name?

RaananW (commented):
Just a question about back compat, otherwise LGTM!

packages/dev/core/src/Engines/engine.ts
Alex-MSFT: Is this actually intoducing a Backwards compat issue, the only reference to this is in WebXRImage, which is also shipped in the @babylonjs\core package so it is only changing the interface for internal use.

I only updated this to include the underscore to match the naming pattern for other hidden methods, so happy to change it back if this does break backwards compat.

RaananW (approved):

packages/dev/core/src/Engines/engine.ts
RaananW: discussed internally. resolving :-)

bghgary:
@Alex-MSFT Just double checking, we don't need to update the protocol version for this right?

Alex-MSFT:
> @Alex-MSFT Just double checking, we don't need to update the protocol version for this right?

No we should not need to, this change should be forwards compatible with the current BabylonNative without any changes.

PR #12178, opened 2022-03-16
Fix bug in NativeEngine.setAlphaMode

bghgary:


ryantrem:
What was the effect of this bug?

ryantrem (approved):

bghgary:
Anything that calls getAlphaMode() will get a wilding incorrect value, then when setAlphaMode is called with that value, _getNativeAlphaMode will throw an exception.

The specific features that I'm aware of which use these function are: EffectLayer and OutlineRenderer.

syntheticmagus (approved):

RaananW:
@bghgary - I will let you merge this one.

PR #12185, opened 2022-03-17
[GUI Editor] UI Update

darraghjburke:
Improves the UI across the GUI editor in various places:

* Adds arrows to canvas resolution and shadow offsets
* Fixes the spacing of the properties panel - units now form a vertical line
* Updates the font file: font faces are now all Acumin
* Left grab and right grab are now invisible and exist at the edge of the left and right panels
* Fixes some linting issues


RaananW:
There are build issues in the gui-editor project

packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -668,7 +668,7 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         return true;
     }
 
-    onMove(evt: React.PointerEvent) {
+    onMove() {
```
carolhmj, line 671: evt is used just below on line 684, so it shouldn't be removed.

darraghjburke: Nice catch, thank you!

carolhmj (approved):

RaananW:
Just making sure - is the color picker working correctly int the other packages?

PR #12201, opened 2022-03-18
[GUI Editor] Space + Click to Pan

darraghjburke:
Holding space in selection mode will now show the pan icon selected and will let you click and drag.

Under the hood, I've created a KeyboardManager class to track all modifier keys.

There is some major technical debt existing in all our tools; the lack of a state management system. We are relying on GlobalState and observables, but this makes it really easy to introduce bugs related to state changes. We may want to consider something like Redux in the future.


darraghjburke:
Marking as draft as this may cause issues with the node editor due to the removal of keyboardBlocker. Going to investigate the right fix for this

darraghjburke:
Will have to pick this up on Monday

darraghjburke:
Ok, for now I have left the blockKeyboardEvents property in, in the interests of time. In the long run we will want to remove this variable from NME and GUI Editor, and use event.target instead to determine if keyboard shortcuts should act on the editor.

RaananW:
@darraghjburke - just some conflicts (from the big PR before)

packages/tools/guiEditor/src/globalState.ts
```diff
@@ -19,6 +20,12 @@ export enum DragOverLocation {
     NONE = 3,
 }
 
+export enum Tool {
```
RaananW, line 23: I think "Tool" is a very abstract name for that. Maybe EditorTool, or GUIEditorControlTool, or something more specific?

RaananW (commented):
Just a small comment about naming. otherwise LGTM!!

packages/tools/guiEditor/src/globalState.ts
carolhmj: +1

carolhmj (approved):
Agree with Raanan about the Tool naming, it's great otherwise! :)

RaananW:
@darraghjburke - any update on this one?

packages/tools/guiEditor/src/globalState.ts
darraghjburke: Sure, we can do that

PR #12202, opened 2022-03-18
[GUI Editor] Fix copy/cut/paste inside text inputs

darraghjburke:
Fixes an issue where you copy/paste events where interpreted as control copy/paste events even when they originated from textboxes. Simple capitalization mismatch.


RaananW (approved):
Just missing formatting and good to go

PR #12205, opened 2022-03-18
[GUI Editor] Fix control icons in tree view

darraghjburke:
Two icons were incorrect in tree view:
* TextBlock
* Button

This PR changes the tree view to use the same icons as the creation panel, and makes the names consistent with the values return by getClassName() in the control classes.


carolhmj (approved):

RaananW (approved):
Assuming those icons are defined in the controls, LGTM!

darraghjburke:
> Assuming those icons are defined in the controls, LGTM!

They are defined in ControlTypes.ts :)

PR #12207, opened 2022-03-18
Update TextureDome after setting halfMode or crossEye

kaliatech:
Changing `TextureDome.halfDome` property after setting `videoMode` does not work. It does work if the sequence is reversed. Here’s a PG showing the issue:

 - [BJS Playground - VideoDome.halfMode issue](https://playground.babylonjs.com/#5I9YRP#1)

The fix is to recalculate the texture coordinates after setting the property.  Same for `crossEye`. This was already being being done when setting the `textureMode`.

Related forum post:
 - https://forum.babylonjs.com/t/videodome-halfdome-and-videomode-order-shouldnt-matter/28551


sebavan (approved):
Thanks a lot

PR #12208, opened 2022-03-19
Perf fix

deltakosh:


deltakosh:
@CedricGuillemet / @syntheticmagus: please check if I did not break anything for Native (I do not believe so but I want to be sure )

PR #12222, opened 2022-03-21
Fix WebGPU visualization test + update .gitignore

Popov72:
For some reason the visualization test fix in https://github.com/BabylonJS/Babylon.js/pull/11961 was not in anymore.

I have also removed the `tools/tests/webgpu.json` file as it seems the right one is in `tools/tests/visualization/webgpu.json`.

Finally, I have added two directories in the `.gitignore` file, that are created when you debug.


RaananW (approved):
LGTM!

I assume you used a different browser path to test? this is probably the reason for the creation of those directories. We should of course support it :-)

Thanks!

PR #12228, opened 2022-03-21
[NME] fix popout issues

darraghjburke:
Fixes the following issues in NME when opened as a popout from the inspector:
* Copy color picker color did not work (switched to Clipboard API)
* Save to snippet server dialog did not open
* Help page did not open

### DIFF
--- a/packages/tools/nodeEditor/src/sharedComponents/color3LineComponent.tsx
+++ b/packages/tools/nodeEditor/src/sharedComponents/color3LineComponent.tsx
@@ -75,6 +75,7 @@ export class Color3LineComponent extends React.Component<IColor3LineComponentPro
     }
 
     copyToClipboard() {
+        const { navigator } = this.props.globalState.hostWindow;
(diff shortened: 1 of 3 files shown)


packages/tools/nodeEditor/src/sharedComponents/color3LineComponent.tsx
```diff
@@ -75,19 +75,7 @@ export class Color3LineComponent extends React.Component<IColor3LineComponentPro
...
-        if (window.getSelection) {
-            const range = document.createRange();
-            range.selectNode(element);
-            window.getSelection()!.removeAllRanges();
-            window.getSelection()!.addRange(range);
-        }
-
-        document.execCommand("copy");
-        element.remove();
+        navigator.clipboard.writeText(this.getCurrentColor().toHexString());
```
RaananW, line 78: if you can use feature detection it will be the safest. i.e. check that `navigator.clipboard` exists and that it has the writeText function.

deltakosh (approved):

PR #12232, opened 2022-03-22
Linting imports

RaananW:
This PR introduces:

1) No import from index (with safe-check during es6 build)
2) no Circular dependencies
3) Bug fix for Legacy in the es6 packages
4) Changing a lot of the imports to import-type (done using auto-fix of eslint) to avoid circular dependencies for types in typescript.


sebavan (approved):
did not check all cause github crashes

deltakosh:
We should merge this one to avoid conflicts :)

RaananW:
This PR now also solves an issue with the legacy in es6, I will merge it right after the CI is done

PR #12240, opened 2022-03-22
Fix caching of parented node

carolhmj:
When setting a node's parent to null/undefined, its cached parent wouldn't be properly updated, causing some issues, such as instanced meshes not having reduced draw calls. 
A playground example can be seen in:
https://playground.babylonjs.com/#6JCJHZ

Without the fix, we have 11 draw calls:
image

With it, we have 1 draw call:
image


sebavan (approved):
GG :-)

PR #12242, opened 2022-03-23
Safeguarding derivates in bump shader

CraigFeldspar:


RaananW:
There seems to be a shader compilation error. you can see the log here - https://dev.azure.com/babylonjs/243fb099-542d-4655-b246-4dfd67131bd4/_apis/build/builds/19143/logs/64

If you want to check the failing playground with your changes:

https://playground.babylonjs.com/?snapshot=refs/pull/12242/merge#LSAUH2#2

I will put a task for me to make sure and log these kinds of errors better.

CraigFeldspar:
> There seems to be a shader compilation error. you can see the log here - https://dev.azure.com/babylonjs/243fb099-542d-4655-b246-4dfd67131bd4/_apis/build/builds/19143/logs/64
> 
> If you want to check the failing playground with your changes:
> 
> https://playground.babylonjs.com/?snapshot=refs/pull/12242/merge#LSAUH2#2
> 
> I will put a task for me to make sure and log these kinds of errors better.

I forgot 1 commit indeed, this should be fixed now

CraigFeldspar:
> it looks like the change is completely breaking some of the renders with bump.

Yea totally, I have no idea why though, this check seemed okay

deltakosh:
Should we close this PR?

CraigFeldspar:
> Should we close this PR?

oh shoo, I forgot about this one. This can be closed as it is breaking bump, but the issue with OIT + transparent + bump still stands. Maybe we can put the safeguard to some lower value like 1e-7 ? Seems hacky though

deltakosh:
I would say: let's try to understand why it is breaking the bump

Is it a sign issue?

CraigFeldspar:
Unlikely, dot(x,x) is always positive. My first guess is that 1e-5 is too high of a value

Popov72:
You can try with the smallest number greater than 0 that can be represented in 32 bits float: ~1.1755e-38.

packages/dev/core/src/Materials/Textures/prePassRenderTarget.ts
```diff
@@ -61,6 +62,7 @@ export class PrePassRenderTarget extends MultiRenderTarget {
         super(name, size, count, scene, options);
 
         this.renderTargetTexture = renderTargetTexture;
+        this._renderPassIds = this.renderTargetTexture ? this.renderTargetTexture.renderPassIds as number[] : [Constants.RENDERPASS_MAIN];
```
Popov72, line 65: Actually, `RENDERPASS_MAIN` is a default value only used as a fallback value.

For eg, `Scene._renderForCamera` is doing this:

```typescript
this._engine.currentRenderPassId = camera.outputRenderTarget?.renderPassId ?? camera.renderPassId ?? Constants.RENDERPASS_MAIN;
```

So, it uses the `renderPassId` of the camera `outputRenderTarget` (if any), then the `renderPassId` from the camera (if any - normally we should always have one and stop here) and uses `RENDERPASS_MAIN` as the fallback just in case.

So, in your case using `RENDERPASS_MAIN` is probably wrong and you should instead use the one from the `camera.outputRenderTarget`/`camera` in effect when rendering into this `PrePassRenderTarget`.

CraigFeldspar: oh thanks @Popov72, this code doesn't belong in this PR this is a branching error.
But your comment stands, I'll make sure to change that in the other PR

deltakosh:
Should we close it? I see no activity since a while

CraigFeldspar:
> Should we close it? I see no activity since a while

Yes you can close it, this breaks bump and I also don't like the fact that we add a shader instruction in all cases, just for OIT not to break with bump.
We'll find another way to have derivatives computed

PR #12247, opened 2022-03-23
[GUI Editor] Overhaul the parameters for each control type

darraghjburke:
* Brings the parameters for each control type up to date with the new design. Spacing is fixed and new icons have been added.
* Adds custom arrows to the float input component.
* Deletes shared components that were not used anywhere

A few controls are still not finished, these are controls that need more substantial changes:
1. Grid
2. Display grid
5. Line

NME and inspector have been tested :)


RaananW (approved):

packages/tools/guiEditor/src/components/propertyTab/propertyGrids/gui/checkboxPropertyGridComponent.tsx
```diff
@@ -4,13 +4,15 @@ import { PropertyChangedEvent } from "shared-ui-components/propertyChangedEvent"
 import { CommonControlPropertyGridComponent } from "../gui/commonControlPropertyGridComponent";
 import { LockObject } from "shared-ui-components/tabs/propertyGrids/lockObject";
 import { Checkbox } from "gui/2D/controls/checkbox";
-import { FloatLineComponent } from "shared-ui-components/lines/floatLineComponent";
 import { CheckBoxLineComponent } from "shared-ui-components/lines/checkBoxLineComponent";
 import { TextLineComponent } from "shared-ui-components/lines/textLineComponent";
 import { makeTargetsProxy } from "shared-ui-components/lines/targetsProxy";
 
-import fillColorIcon from "shared-ui-components/imgs/fillColorIcon.svg";
+import checkBoxIconDark from "../../../../imgs/checkBoxIconDark.svg";
```
RaananW, line 11: should probably be 3 levels down

darraghjburke: Fixed! Thanks

carolhmj (commented):
Hey there! Not sure if this is intentional, but the rotate slider is quite short compared to the available space:
image
It also doesn't accept negative values, while you can set negative angles in code and it works.

darraghjburke:
> Hey there! Not sure if this is intentional, but the rotate slider is quite short compared to the available space: image It also doesn't accept negative values, while you can set negative angles in code and it works.

Yes, the size of the slider needs to be fixed. This is one layout issue that's a little tricky, I decided to save it for a future time since there are higher priority things to go after right now.

As far as negative angles, technically you can set the rotation in code to any value and it will work, but we don't really have a great way to represent that in the GUI editor. What we could do is let you input any value and then compute the final result between 0 and 2*Pi, but this may require some changes to the slider component. I'm going to save this for a later date as well

Issue #12255, opened 2022-03-24
[GUI] Virtual keyboard encodings are broken in GUI editor


darraghjburke:
Repro:
1. Open gui.babylonjs.com
2. Add a virtual keyboard

Several keys (ctrl, shift, etc.) are broken. Somewhere in the build step, the encoding gets messed up. Playground works fine, as does localhost GUI editor.

PR #12259, opened 2022-03-25
[Bauild] Tools revamp

RaananW:
1) Tools now all have a fallback in case the preview CDN is offline
2) Tools encode correctly (utf-8)
3) Tools all support snapshots now - add ?snapshot=refs/heads/master to the URL of any tool (or change the ID to any of the other pull requests snapshots
4) Playground - declarations fallbacks implemented. Also - tool will not stop working if declaration could not be loaded.
5) General formatting and structure changes.

### DIFF
--- a/packages/tools/guiEditor/package.json
+++ b/packages/tools/guiEditor/package.json
@@ -5,9 +5,9 @@
     "files": [
         "dist"
     ],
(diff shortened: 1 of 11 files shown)


RaananW:
Also addressing #12255 (encoding was missing)

PR #12260, opened 2022-03-25
chore: update main and types paths in UMD babylonjs-materials package

julien-moreau:
Paths to main & types missed "js" in the package.json of the "babylonjs-materials" UMD package.

Here is a preview of what the package contains where setting "babylonjs.materials.js" instead of "babylon.materials.js" fixed the links.

Setting these real paths avoids errors when requiring the package and compilation errors such as "babylonjs-material is not a module".

Capture d’écran 2022-03-25 à 13 03 58

Do you confirm that this is the only file that need to be changed ?

### DIFF
--- a/packages/public/umd/babylonjs-materials/package.json
+++ b/packages/public/umd/babylonjs-materials/package.json
@@ -1,8 +1,8 @@
 {
     "name": "babylonjs-materials",
     "version": "5.0.0-rc.10",
-    "main": "babylon.materials.js",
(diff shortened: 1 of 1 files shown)


RaananW (approved):

RaananW:
thanks! merging this now and will build rc.11 very soon with a few other small fixes.

julien-moreau:
Thanks @RaananW you rox!
On my side I fixed the package.json manually so I can continue working and everything works perfect!

PR #12262, opened 2022-03-25
[Build] Local Server changes

RaananW:
* Added npm start which will run the dev watcher and the babylon server in parallel.
* The server now creates two untracked files - createScene and createEngine. Use them to generate your scene when working locally.

Note - declarations and auto complete will work once you run the dev-watch at least once.


sebavan:
Can not wait to try it :-)

PR #12266, opened 2022-03-26
Fix double opposite in PBR NME Right Handed

sebavan:
https://forum.babylonjs.com/t/nme-pbrmetallicroughnessblock-reflections-are-incorrect-with-right-handed-coordinate-system/28753/3


packages/dev/core/src/Materials/Node/Blocks/PBR/reflectionBlock.ts
```diff
@@ -346,7 +346,7 @@ export class ReflectionBlock extends ReflectionTextureBaseBlock {
 
         const computeReflectionCoordsFunc = `
             vec3 computeReflectionCoordsPBR(vec4 worldPos, vec3 worldNormal) {
-                ${this.handleFragmentSideCodeReflectionCoords("worldNormal", "worldPos", true)}
+                ${this.handleFragmentSideCodeReflectionCoords("worldNormal", "worldPos", true, true)}
```
Popov72, line 349: Just to be sure: the last `true` should not be `scene.useRightHandedSystem`?

sebavan: I do not think so but I might be fully wrong. the main issue is that the invertz code is already in the createReflectionCoords code from the pbrReflectionBlock where we already remap:

```
{ search: /REFLECTIONMAP_OPPOSITEZ/g, replace: reflectionBlock?._defineOppositeZ ?? "REFLECTIONMAP_OPPOSITEZ" },
```

Do not hesitate if I miss anything :-)

RaananW (approved):
@sebavan - I will let you resolve the conversation and merge

PR #12271, opened 2022-03-28
Fix deletion of descendant controls in the gui editor

carolhmj:
Without this change, deleting a button, for example, would only delete it on the live texture but not on the linked GUI editor


packages/tools/guiEditor/src/globalState.ts
```diff
@@ -163,10 +164,37 @@ export class GlobalState {
...
+
     public deleteSelectedNodes() {
         for (const control of this.selectedControls) {
-            this.guiTexture.removeControl(control);
-            this.liveGuiTexture?.removeControl(control);
+            const guiTextureParent = this._findParentControlInTexture(this.guiTexture, control);
+            guiTextureParent?.removeControl(control);
+            if (this.liveGuiTexture) {
+                const allDescendants = control.getDescendants();
+                for (const descendant of allDescendants) {
```
darraghjburke, line 190: Is it necessary to remove all descendants? If we remove all direct children of the control, shouldn't that take all the descendants with them?

darraghjburke (commented):
Looks good, just one question!

packages/tools/guiEditor/src/globalState.ts
carolhmj: Removing the direct descendants of a control does work on the Editor GUI Texture, but I think the Live GUI texture is organized in a different way, because I tried that and it didn't work :(

darraghjburke: Weird...so what happens to the descendants? Do they become children of the root?

carolhmj: Yeah, they're children of the root on the Live texture

darraghjburke: Strange...we should definitely look into this more later on. Either way, this PR LGTM!

RaananW (approved):

PR #12275, opened 2022-03-28
[GUI Editor] Fix unit switching when value is 0

darraghjburke:
Lets float and text inputs update when the unit changes. Fixes an issue where converting a 0px value to a 0% value would not show the updated unit.


RaananW (approved):

carolhmj (dismissed):
I still can't switch units if I follow these steps:
1 - Add a rectangle to the GUI through the editor
2 - Add a text block to the GUI through the editor
3 - Parent the text block to the rectangle and try to change its width to %. It works when clicking on the "%" button, but once you click on the height, the width changes back to px.

darraghjburke:
> I still can't switch units if I follow these steps: 1 - Add a rectangle to the GUI through the editor 2 - Add a text block to the GUI through the editor 3 - Parent the text block to the rectangle and try to change its width to %. It works when clicking on the "%" button, but once you click on the height, the width changes back to px.

This is the resizeToFit issue, where a textblock with resizeToFit on will not allow you to change its width/height. We just need to update the UI to reflect that. I will do a separate PR later today for that :)

sebavan (approved):

PR #12276, opened 2022-03-28
[GUI Editor] Line Control Updates

darraghjburke:
Primarily, this PR refactors the gizmo to support both regular controls and line controls, which have a different kind of gizmo (points on the ends of the line.)

There is still work remaining on line controls: with this PR, the gizmo is not actually draggable, it's just a visual indicator.

Also makes some other improvements:
* Rounds values of x1,y1,x2,y2 to 2 decimal places
* Hides padding from lines parameters (not relevant)


packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -656,17 +657,15 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
 
     onMove(evt: React.PointerEvent) {
         const pos = this.getScaledPointerPosition();
         // Move or guiNodes
         if (this._mouseStartPoint != null && !this._panning) {
-            let selected = false;
             this.props.globalState.selectedControls.forEach((element) => {
                 if (pos) {
-                    selected = this._onMove(element, new Vector2(pos.x, pos.y), this._mouseStartPoint!) || selected;
+                    this._onMove(element, new Vector2(pos.x, pos.y), this._mouseStartPoint!);
```
RaananW, line 668: a little nit-picky, but - is it possible to reuse a vector instead of creating a new one? onPointerDown or Up wouldn't matter so much, but pointermove happens quite often.

RaananW (commented):
Just a small comment. otherwise - LGTM!

carolhmj (approved):

packages/tools/guiEditor/src/diagram/workbench.tsx
darraghjburke: good point!

PR #12287, opened 2022-03-29
[GUI Editor] Units refactor

darraghjburke:
Refactors units inside text/float inputs to be a separate component. Now you simply pass a UnitButton into the unit prop of the input component. This reduces code duplication and made it easy to add units to the slider component.

Also:
* makes units use "disabled" HTML attribute rather than a "disabled" class, for increasing accesibility
* fixes minor CSS issues with the slider component.


RaananW (commented):
Love this change! Preparing for a global unit button 😊
Just making sure it doesn't influence any other tool?

darraghjburke:
> Love this change! Preparing for a global unit button 😊 Just making sure it doesn't influence any other tool?

Yep, tested on playground and NME. All good! Only GUI editor is using units right now

RaananW (approved):

PR #12291, opened 2022-03-30
Fix recursive dispose

sebavan:


packages/dev/core/src/scene.ts
```diff
@@ -4727,6 +4703,15 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         this._isDisposed = true;
     }
 
+    private _disposeList<T extends IDisposable>(items: T[], callback?: (item: T) => void): void {
+        const itemsCopy = items.slice(0);
```
bghgary, line 4707: Should we use SliceTools.Slice here?

deltakosh (approved):

PR #12297, opened 2022-03-30
avoid duplication

deltakosh:


sebavan (approved):

packages/dev/core/src/scene.ts
```diff
@@ -2669,6 +2674,10 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
             return;
         }
 
+        if (newMaterial._indexInSceneMaterialArray !== -1) { // Already there?? 
+            return;
```
ryantrem, line 2678: If the material was added to a different scene, it will be hard to diagnose the problem. Maybe check the scene in this case and if it is different then throw an error?

deltakosh: I can but adding a material from another scene will crash anyway. This is not a supported use case (and if it works it is purely by chance)

PR #12298, opened 2022-03-30
Revert "[GUI Editor] Fix unit switching when value is 0"

RaananW:
Reverts BabylonJS/Babylon.js#12275


RaananW:
@darraghjburke - this has to be reverted, please check again afterwards.

darraghjburke (approved):

PR #12301, opened 2022-03-30
[GUI Editor] Changes to saving and cleanup

darraghjburke:
Fixes a bug where saving the GUI multiple times would trigger a crash.

Also modifies the behavior around saving and restoring GUI properties when opening a live GUI scene. Now we store a list of all observers attached to the control, remove them all, and then restore them when the editor is closed. This seems like the best way to handle observers - they should not trigger while the GUI editor is open.


carolhmj (approved):

packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -321,33 +321,116 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         this.props.globalState.onPointerUpObservable.notifyObservers(null);
     };
 
+    /**
+     * Adds editor behavior to control
+     * @param guiControl
+     */
+    addEditorBehavior(guiControl: Control) {
+        const onPointerUpStored = guiControl.onPointerUpObservable.observers;
+        guiControl.onPointerUpObservable.clear();
```
RaananW, line 330: This takes an assumption that when you call clear you create a new array. otherwise the clear call will also clear the array you are storing.
To stay safe (just in case someone changes the `clear` function of the observable, make sure to slice the array (i.e. - make a shallow copy) instead of just referencing it.

RaananW: Same goes to the rest of the references :-)

RaananW (changes requested):
Just a small issue with array reference vs. array cloning.
otherwise LGTM!!

packages/tools/guiEditor/src/diagram/workbench.tsx
darraghjburke: Ok done!

PR #12302, opened 2022-03-30
Cameras: Reset current active button when controls are detached

PolygonalSun:
This PR contains a simple fix that resets _currentActiveButton to -1 in our cameras.  This addresses an a potential scenario where the controls are detached before a pointerup is handled.


deltakosh (approved):

bghgary (approved):
LGTM

PR #12312, opened 2022-03-31
WebGPU Update

sebavan:


packages/dev/core/src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
```diff
@@ -494,4 +508,11 @@ export class WebGPUShaderProcessorWGSL extends WebGPUShaderProcessor {
 
         return code;
     }
+
+    private _processStridedUniformArrays(code: string): string {
+        for (const uniformArrayName of this._stridedUniformArrays) {
+            code = code.replace(new RegExp(`${uniformArrayName}\\[(.*)\\]`, "g"), `${uniformArrayName}[$1].el`);
```
Popov72, line 514: I think syntax like `myarray [23]` is allowed in wgsl, so a `\s*` should be added in the regexp?

Something like: `${uniformArrayName}\\s*\\[(.*)\\]`

Popov72 (approved):
LGTM!

What a mess we have to do because they removed `@stride`...

packages/dev/core/src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
```diff
@@ -365,23 +369,33 @@ export class WebGPUShaderProcessorWGSL extends WebGPUShaderProcessor {
...
         let ubo = `struct ${name} {\n`;
         for (const leftOverUniform of this._webgpuProcessingContext.leftOverUniforms) {
             const type = leftOverUniform.type.replace(/^(.*?)(<.*>)?$/, "$1");
             const size = WebGPUShaderProcessor.UniformSizes[type];
 
             if (leftOverUniform.length > 0) {
                 if (size <= 2) {
-                    ubo += ` @align(16) ${leftOverUniform.name} : @stride(16) array<${leftOverUniform.type}, ${leftOverUniform.length}>;\n`;
+                    const stridedArrayType = `${name}_${stridedArraysLength}_strided_arr`;
+                    stridedArrays += `struct ${name}_${stridedArraysLength}_strided_arr {
```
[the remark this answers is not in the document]
sebavan: Good catch

packages/dev/core/src/Engines/WebGPU/webgpuShaderProcessorsWGSL.ts
sebavan: yup

sebavan:
> LGTM!
> 
> What a mess we have to do because they removed `@stride`...

Yes, this is insane... and what they now do in tint.

deltakosh (approved):

PR #12318, opened 2022-04-01
Added a new flag for checking double sided mesh in collision

CedricGuillemet:
Added a new flag for checking double sided mesh in collision detection (disable by default)

follow up https://forum.babylonjs.com/t/movewithcollisions-cant-handle-doublesided-meshes/28928
double sided face are evaluated 2 times, by design. With this change, they are only evaluated once.
To not break back compat, it's hidden behind a flag.
To enable it :
```
BABYLON.Collider.DoubleSidedCheck = true;
```


RaananW (approved):

RaananW:
we'll need to document that as well :-)

PR #12320, opened 2022-04-01
empty loadingUIText setter for null engine

nekochanoide:
using null engine and setting loading ui text to it causes crash...
added 'do nothing' setter for `loadingUIText`


RaananW:
The setter is being inherited from the Engine class, and is located in NullEngine as well. I gues the issue is that loadingScreen is gone?

nekochanoide:
yes. no loadingScreen in null engine but `NullEngine` type says that it's okey to use `loadingUIText`, so code in `Engine` throws 'can't access `loadingUIText` of undefined'.

nekochanoide:
to break current version execute this
```typescript
let e = new BABYLON.NullEngine();
e.loadingUIText = "hello";
```

deltakosh (approved):

PR #12327, opened 2022-04-01
[NME] Adjust behavior of NME selection to only select a frame if there aren…

carolhmj:
…'t nodes selected. Also fix an error that showed up on compile.


PatrickRyanMS:
I tested the functionality and marquee select is working well with nodes in frames. I still think it's worth looking into multi-select for frames either through ctrl + click or marquee selecting frames only. But that can come after this issue as currently there is still a way to move frames one at a time and it does not block functionality like not being able to marquee select nodes in a frame does.

PatrickRyanMS (approved):

deltakosh (approved):

PR #12331, opened 2022-04-03
PBR thin instance color attribute needs to be pushed

BlakeOne:
PBRBaseMaterial is missing the code to push the attribute for thin instance color. This PR just copies the same lines from StandardMaterial.

Here's a [PG](https://playground.babylonjs.com/#JGBXWZ#3) that shows the issue that's fixed by the PR, where the thin instances are black when they should be red.


sebavan:
for reference https://github.com/BabylonJS/Babylon.js/pull/12333

PR #12333, opened 2022-04-04
Fix instance color forwarding for PBR Instances

CedricGuillemet:
Followup https://forum.babylonjs.com/t/vertex-error-and-no-colors-on-thin-instances-after-upgrade/28492/13
Instance colors were not properly forwarded for PBR materials

### DIFF
--- a/packages/dev/core/src/Materials/PBR/pbrBaseMaterial.ts
+++ b/packages/dev/core/src/Materials/PBR/pbrBaseMaterial.ts
@@ -1326,6 +1326,10 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             attribs.push(VertexBuffer.ColorKind);
         }
 
+        if (defines.INSTANCESCOLOR) {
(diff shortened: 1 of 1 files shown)


sebavan (approved):

packages/dev/core/src/Materials/PBR/pbrBaseMaterial.ts
```diff
@@ -1326,6 +1326,10 @@ export abstract class PBRBaseMaterial extends PushMaterial {
             attribs.push(VertexBuffer.ColorKind);
         }
 
+        if (defines.INSTANCESCOLOR) {
```
Popov72, line 1329: This block of code is also missing in all materials of the material library (and maybe in some other materials, I know the fluent controls in 3D GUI are also using custom materials).

I think it would be better to update the `MaterialHelper.PrepareAttributesForInstances` method to push `VertexBuffer.ColorInstanceKind` in the attribute list because this method is called each time instances/thin instances are supported by materials. This way, you won't have to update individually each material.

PR #12337, opened 2022-04-05
fix physics regression

CedricGuillemet:
followup https://forum.babylonjs.com/t/compound-to-empty-mesh-demo-works-differently-for-cannon-omio-vs-ammo/29062

When computing bounding info from scratch, before this change, the min and max position was `mesh.position`. So, in this case:
```
var compoundBody = new BABYLON.Mesh("compoundBody", scene);
compoundBody.position.y=3;

compoundBody.computeWorldMatrix(true);
console.log(compoundBody.getBoundingInfo().boundingSphere.centerWorld.y);
console.log(compoundBody.position.y);
```
min and max are (0,3,0), bouding sphere local center is (0,3,0).
world bounding sphere center is local center transformed by world matrix and its value is (0,6, 0) which is not correct. It breaks the physics for oimo and cannon. Computation is different with ammo and was not touched by this issue.

@RaananW Git blame indicates you changed it last. Do you remember why/what it fixed?

### DIFF
--- a/packages/dev/core/src/Meshes/abstractMesh.ts
+++ b/packages/dev/core/src/Meshes/abstractMesh.ts
@@ -1573,7 +1573,7 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
         if (this._boundingInfo) {
             this._boundingInfo.update(this.worldMatrixFromCache);
         } else {
-            this._boundingInfo = new BoundingInfo(this.position, this.position, this.worldMatrixFromCache);
+            this._boundingInfo = new BoundingInfo(Vector3.Zero(), Vector3.Zero(), this.worldMatrixFromCache);
         }
         this._updateSubMeshesBoundingInfo(this.worldMatrixFromCache);
         return this;


RaananW:
The reason you see me last is the build system changes. Going back before the changes (there is a branch called legacyMaster that can show you the structure pre-changes), the blame is this:

image

@bghgary should review that.

deltakosh (approved):

bghgary:
> The reason you see me last is the build system changes. Going back before the changes (there is a branch called legacyMaster that can show you the structure pre-changes), the blame is this:
> 
> image
> 
> @bghgary should review that.

Actually, I just renamed a variable. It's actually this one: https://github.com/BabylonJS/Babylon.js/commit/37281d494b26e2402be9c112f31bb9cad833cc1f

RaananW:
The plot thickens! It used to be absolute position, i just changed it to position

https://github.com/BabylonJS/Babylon.js/pull/11093

I'll trace this, but this has been this way for quite some time now. Not sure what are the consequences of changing it

PR #12342, opened 2022-04-05
Update the GLTF breaking change in What's new with a workaround.

carolhmj:


deltakosh (approved):

deltakosh:
You also need to change it on the doc page (as it was published when we shipped 5.0)

PR #12345, opened 2022-04-05
[GUI] use ignoreAdaptiveScaling on containers width adaptWidthToChildren or adaptHeightToChildren

darraghjburke:
See forum thread: https://forum.babylonjs.com/t/gui-adaptwidthtochildren-not-working-with-adaptive-scaling/28911/7

Containers with adaptWidthToChildren or adaptHeightToChildren currently have adaptative scaling applied twice, once when their children's dimensions are computed and once when their own dimensions are computed. This PR simply turns off adaptive scaling on the relevant property when it's being adapted to the children's dimensions.


RaananW (approved):

RaananW:
I'm very happy to have merged pr number #12345 😁

PR #12348, opened 2022-04-05
[GUI] On parse, reconnects the keys of the virtual keyboard

darraghjburke:
Forum thread: https://forum.babylonjs.com/t/webxr-gui-editor-and-virtualkeyboard/27826

When parsing a virtual keyboard, we can scan the children for any key buttons and reconnect them to the keyboard. This PR uses the name of the button to do this - this will work unless they have renamed the control. We could also consider using the text of the button instead.

@RaananW curious about your thoughts on this. I know we debated whether or not this was worth supporting but I think it's a small enough change that it might be worth having.


RaananW (approved):
I'm perfectly fine with this, but we should document that to explain the meaning behind changing the names of the controls (Or prevent them from being changed, but that might be harder to implement). 
This could also allow creating a different keyboard layout.

PR #12351, opened 2022-04-06
[Build] Allow editors packages to be included in a single project

RaananW:
Due to the way the shared ui components were declared it was not possible to include both editors in a single typescript project.
This generates a different namespace to the shared ui components so they are not declared twice.

### DIFF
--- a/packages/dev/buildTools/src/generateDeclaration.ts
+++ b/packages/dev/buildTools/src/generateDeclaration.ts
@@ -3,7 +3,7 @@ import * as fs from "fs";
 import * as path from "path";
 import * as chokidar from "chokidar";
 
-import { camelize, checkArgs, checkDirectorySync, debounce, findRootDirectory, getHashOfContent, getHashOfFile } from "./utils";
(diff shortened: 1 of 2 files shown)


packages/dev/buildTools/src/generateDeclaration.ts
```diff
@@ -3,7 +3,7 @@ import * as fs from "fs";
 import * as path from "path";
 import * as chokidar from "chokidar";
 
-import { camelize, checkArgs, checkDirectorySync, debounce, findRootDirectory, getHashOfContent, getHashOfFile } from "./utils";
+import { camelize, checkArgs, checkDirectorySync, debounce, findRootDirectory, getHashOfContent, getHashOfFile, kebabize } from "./utils";
```
deltakosh, line 6: And the nominees for the best variable name are... :)

RaananW: make it kebab case FTW! :-)

RaananW: This - https://stackoverflow.com/questions/63116039/camelcase-to-kebab-case ;-)

PR #12360, opened 2022-04-07
Removed flip faces for MergeMeshes

deltakosh:
I tested this fix with:
- https://playground.babylonjs.com/#E99FHV (glb merging)
- https://playground.babylonjs.com/#YIU90M#782 (glb merging)
- https://playground.babylonjs.com/#KBS9I5#76 (babylon file merging)

We do not need to flip the faces as the initial world matrix is merged as well so it is actually important to keep the indices the way they are.

Related forum discussion: https://forum.babylonjs.com/t/babylon-5-0-2-mesh-mergemeshes-flips-backface-of-glb-models/29043/11


RaananW:
As reference - fixed versions should be:

https://playground.babylonjs.com/?snapshot=refs/pull/12360/merge#E99FHV (glb merging)
https://playground.babylonjs.com/?snapshot=refs/pull/12360/merge#YIU90M#782 (glb merging)
https://playground.babylonjs.com/?snapshot=refs/pull/12360/merge#KBS9I5#76 (babylon file merging)

ryantrem:
How was this not totally broken if it was flipping faces but shouldn't be?

deltakosh:
Wait..something is really wrong with the build or something

I need to double check with @RaananW

deltakosh:
> How was this not totally broken if it was flipping faces but shouldn't be?

This is a very interesting case where the mix could work on some situation thanks to the overrideMaterialOrientation but like you I'm amazed that we did not see it earlier

deltakosh:
Ok so I confirmed the fix now :)

https://playground.babylonjs.com/?snapshot=refs/pull/12360/merge#E99FHV (glb merging)
https://playground.babylonjs.com/?snapshot=refs/pull/12360/merge#YIU90M#782 (glb merging)
https://playground.babylonjs.com/?snapshot=refs/pull/12360/merge#KBS9I5#76 (babylon file merging)

bghgary:
I feel like we have had a few regressions with merge mesh. Would it be good to add some unit tests?

deltakosh:
oh yes! I will

sebavan:
@ryantrem would be amazing if you could test in your app as well as adding one of your "crazy tests" in the test list as well :-)

RaananW (approved):
LGTM, but i would wait for @ryantrem and @bghgary.

ryantrem:
I should be able to test these changes out later today.

bghgary:
There is no need to wait for me.

ryantrem:
I tested these changes (by commenting out the three calls to `_FlipFaces`), and it definitely breaks merging with some of our models. Shall we sync directly?

ryantrem (changes requested):
As mentioned in my comment, this breaks merging for some of my team's use cases, so I'm hoping these issues can be figured out before we move forward with this PR.

deltakosh:
Ideally a repro in the PG would be useful so I can compare with other (which are working)

deltakosh:
Moving to Draft until we figure out why it is not working for @ryantrem models

deltakosh:
Ok this was a very wrong fix..The code was actually correct and the problem was coming from the PG we used :(

PR #12362, opened 2022-04-07
ktxTextureLoader: Mark _useSRGBBuffers when loading an SRGB-enabled texture format

magcius:
Also set texture.format to the correct value,
since it seems the code elsewhere, like in the WebGPU engine, would
like a "base" texture format.

Completely untested patch to try and fix https://forum.babylonjs.com/t/compressed-ktx-texture-appears-too-dark/29060/4

### DIFF
--- a/packages/dev/core/src/Engines/constants.ts
+++ b/packages/dev/core/src/Engines/constants.ts
@@ -175,22 +175,46 @@ export class Constants {
 
     /** Compressed BC7 */
     public static readonly TEXTUREFORMAT_COMPRESSED_RGBA_BPTC_UNORM = 36492;
+    /** Compressed BC7 (SRGB) */
+    public static readonly TEXTUREFORMAT_COMPRESSED_SRGB_ALPHA_BPTC_UNORM = 36493;
     /** Compressed BC6 unsigned float */
     public static readonly TEXTUREFORMAT_COMPRESSED_RGB_BPTC_UNSIGNED_FLOAT = 36495;
     /** Compressed BC6 signed float */
     public static readonly TEXTUREFORMAT_COMPRESSED_RGB_BPTC_SIGNED_FLOAT = 36494;
     /** Compressed BC3 */
     public static readonly TEXTUREFORMAT_COMPRESSED_RGBA_S3TC_DXT5 = 33779;
+    /** Compressed BC3 (SRGB) */
+    public static readonly TEXTUREFORMAT_COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT = 35919;
     /** Compressed BC2 */
     public static readonly TEXTUREFORMAT_COMPRESSED_RGBA_S3TC_DXT3 = 33778;
+    /** Compressed BC2 (SRGB) */
+    public static readonly TEXTUREFORMAT_COMPRESSED_SRGB_ALPHA_S3TC_DXT3_EXT = 35918;
     /** Compressed BC1 (RGBA) */
     public static readonly TEXTUREFORMAT_COMPRESSED_RGBA_S3TC_DXT1 = 33777;
     /** Compressed BC1 (RGB) */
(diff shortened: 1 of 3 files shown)


magcius:
I don't believe I can select labels. Did I miss my chance?

bghgary:
@magcius This code doesn't work because `_useSRGBBuffers` get reset back to false here: https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Engines/thinEngine.ts#L4485-L4486

bghgary:
I have a fix.

magcius:
Ah, nice catch, thanks. I believe this would still fix the issue we were seeing with dark textures (the issue there is that the SRGB format *and* the shader fallback are both used), but ideally we'd be using the native SRGB hardware support instead of the shader fallback.

There is a similar bug in the KTX2 texture loader where it won't properly use SRGB buffers either, because it only sets texture.gammaTexture and not _useSRGBBuffer (it should set both, I believe).

My preferred fix for this would be to drop the glInternalFormat from the compressed texture upload function, and replace it with code that reconstructs the right texture format based on texture.format and _useSRGBBuffer, similar to how the webgpuEngine works. Does that sound good to you?

bghgary:
My local fix isn't quite right. There are still problems with it. I'm still checking.

bghgary:
@sebavan @deltakosh I've made some changes that makes this work for KTX 1. KTX 2 works but it doesn't use srgb buffers even if it's available. Other texture loaders don't seem to support srgb formats. Should we try to fix srgb formats more broadly?

magcius:
KTX2 doesn't use SRGB buffers because it only sets gammaTexture, not _useSRGBBuffers. I think a proper fix for that would be to decide the SRGB buffer-ness at texture creation time, like how the WebGPU backend works.

bghgary:
> I think a proper fix for that would be to decide the SRGB buffer-ness at texture creation time, like how the WebGPU backend works.

Can you explain what you mean by this? I don't see any code that decides this flag at texture creation that is different than WebGL or when using texture loaders (e.g. KTX).

bghgary:
> The idea is we'd do something similar in the WebGL backend as well -- any time someone creates a texture with BC1 and _useSRGBBuffers is set to true, we'd use BC1_SRGB.

There is similar code for WebGL here for the compressed texture part: https://github.com/BabylonJS/Babylon.js/blob/81fd1dc6750d241b570132c39f6ca563917f46b5/packages/dev/core/src/Engines/thinEngine.ts#L4444-L4492

> Currently, _useSRGBBuffers just forces the format the other way -- if it's set to false, it will strip SRGB from the format.

Unless I missed something, that's not what happens in the function you referenced for WebGPU.

> KTX1 has glInternalFormat = ETC2_SRGB. The KTX1 loader would split this into texture.format = ETC2 and _useSRGBBuffers = true

This is what is happening in your PR. I just fixed it so that it works. I think we should be good to go for KTX1.

> KTX2 has transcoded texture format (e.g. ETC2) and SRGB comes from the DTF flag. More straightforward: texture.format = ETC2 and _useSRGBBuffers = true

Right. We can additionally fix this, but it should be a separate fix.

> The texture creation code will merge both of these so even though texture.format is ETC2, the texture's internal format is ETC2_SRGB, and marked at texture creation time.

Just for a bit of context, the main reason this _useSRGBBuffers flag exists is because, at runtime, we may or may not have srgb buffers support. This support can be disabled in the engine capabilities or for WebGL1 when [EXT_sRGB](https://developer.mozilla.org/en-US/docs/Web/API/EXT_sRGB?msclkid=07d2c6ebbb4211ec9b8bafe380926cd0) is not present. When srgb buffers are not supported, the engine will fall back to converting the color in the shader which is controlled by the `_gammaSpace` flag on the internal texture.

magcius:
Sounds fine to me. Let's get this fix merged now and deal with the rest later.

bghgary (approved):

PR #12367, opened 2022-04-09
Revert the missing fix and add some logs

deltakosh:


ryantrem (approved):

deltakosh:
Cc @sebavan as it seems the iridescence tests are failing

sebavan:
> Cc @sebavan as it seems the iridescence tests are failing

Can you try to merge master in ? I do not understand why master is not failing :-(

sebavan:
I updated it to see, still no explanations so far...

sebavan (approved):

sebavan:
@RaananW would be great if you could have a look, I wonder why the config was from master but none of the iridescence code was available ? but at least it is all in :-)

RaananW:
@sebavan - I will look into that. The config is taken from the branch it's in, nothing should be taken from master.

PR #12369, opened 2022-04-09
Fix minor shader compiler issues

kaliatech:
Changes:
 - Allow semicolons in shader single line comments
 - ~~Allow indented multiline macros~~
 
Related post in forum showing reproduction for the single line comment issue:
- https://forum.babylonjs.com/t/possible-shader-compiler-bug-with-single-line-comments/29182

~~There are no reports for the indented multiline macro issue, but I verified locally that indenting a multiline macro would break compiler, and with this fix, it works.~~

### DIFF
--- a/packages/dev/core/src/Engines/Processors/shaderCodeCursor.ts
+++ b/packages/dev/core/src/Engines/Processors/shaderCodeCursor.ts
@@ -21,6 +21,12 @@ export class ShaderCodeCursor {
                 continue;
             }
 
+            // Do not split single line comments
+            if (line.trim().startsWith("//")) {
+                this._lines.push(line);
+                continue;
+            }
+
             const split = line.split(";");
 
             for (let index = 0; index < split.length; index++) {


sebavan (approved):
Nice :-)

sebavan:
This loooks like a great start but could you merge back master in so that we keep only the issues related to the PBR ? I am seeing some iridescence issues related to master in the build.

kaliatech:
> could you merge back master in so that we keep only the issues related to the PBR

I rebased back to `master` before your recent PR.  The tests are still failing and it does seem to be due to changes I made in this PR.  Looking at some of the failing test PGs, the underlying reason is not obvious to me though. My guess is that's due to my 2nd commit to allow for multiline defines, but that's just a guess.

If not obvious to anyone else, I will try running/debugging the failing tests next week.

BlakeOne:
Hey @kaliatech, it seems pushing the trimmed line instead of the original line fixes the 3 PG's that were failing validation, although I'm not sure why necessary... Like below for example. :)

`const trimmedLine = line.trim();
if (trimmedLine[0] === "#" || trimmedLine.startsWith("//")) {
    this._lines.push(trimmedLine);
    continue;
}`

PR #12374, opened 2022-04-11
Physics,scene Bug fixes

CedricGuillemet:
- Bug fix https://forum.babylonjs.com/t/inspector-bug-with-wireframe-not-being-aligned-on-mesh/29021/10
Clone for wireframe mesh was also cloning physics impostor. And the impostor having a delta rotation, it was applied twice (parent's delta + delta) causing invalid rendering
- Bug fix https://forum.babylonjs.com/t/physicsimpostor-ignoreparent-causes-crash/29194/4
- scene `skipPointerDownPicking` : do not pick ray in scene when not appropriate. This is causing massive performance drop on native.

### DIFF
--- a/packages/dev/core/src/Inputs/scene.inputManager.ts
+++ b/packages/dev/core/src/Inputs/scene.inputManager.ts
@@ -745,7 +745,12 @@ export class InputManager {
 
             // Meshes
             this._pickedDownMesh = null;
-            const pickResult = scene.pick(this._unTranslatedPointerX, this._unTranslatedPointerY, scene.pointerDownPredicate, false, scene.cameraToUseForPointers);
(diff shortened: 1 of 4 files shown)


deltakosh:
@CedricGuillemet When is skipPointerDownPicking supposed to be true?

RaananW (approved):

CedricGuillemet:
@deltakosh any time you don't need (CPU intensive) picking : game, viewer, demo, ...
For example, on native I drop from 60FPS to 8FPS with space demo. picking in non-JIT is *very* expensive. It's also expensive on the web.
another case here : https://github.com/BabylonJS/BabylonReactNative/issues/387

PR #12378, opened 2022-04-11
add overrideCloneAlphaBetaRadius to ArcRotateCamera

BlakeOne:
Adds new property overrideCloneAlphaBetaRadius to the class ArcRotateCamera. This allows the parameter to setTarget to be specified when animating the camera's target property.

For example, when using FramingBehavior the override can be set to true, which then allows the framingTime (i.e. speed) of the framing animation to be changed independently, without altering anything else about the animation.

Forum discussion: https://forum.babylonjs.com/t/zoomonmesheshierarchy-inconsistent-camera-rotation/29100/8


deltakosh (approved):

PolygonalSun (approved):
LGTM

PR #12387, opened 2022-04-12
switch sandbox to double (and fix a tiny GC issue with audio)

deltakosh:


packages/tools/sandbox/src/components/renderingZone.tsx
```diff
@@ -80,6 +80,7 @@ export class RenderingZone extends React.Component<IRenderingZoneProps> {
             await (this._engine as WebGPUEngine).initAsync();
         } else {
             this._engine = new Engine(this._canvas, antialias, {
+                useHighPrecisionMatrix: true,
```
Popov72, line 83: I think you should also add this option to the WebGPU engine creation?

deltakosh: Yes totally!

Wanna do it ? (I can do it else)

Popov72: I think it will be easier if you do it as you are already doing the PR, it's only copying the `useHighPrecisionMatrix: true,` line 4 or 5 lines above.

[...] Ah, you will also need to add the `useHighPrecisionMatrix` option in the `WebGPUEngineOptions` interface and call `PerformanceConfigurator.SetMatrixPrecision(!!options.useHighPrecisionMatrix);` in the constructor as this option was not supported yet!

deltakosh: yeah that was my point ;D

Adding the line is ok of course.

I'll do it ;D

RaananW (approved):

PR #12389, opened 2022-04-12
Export KHR extensions: IOR, Transmission and Volume

pandaGaume:
GLTF 2.0 Serializer is missing the 

- KHR_materials_ior 
- KHR_materials_volume 
- KHR_materials_transmissions

extensions

This PR add support for these extensions to the GLTF2.0 serializer.
Also, some minor modification has been made into the GLTF Loader extensions to be consistant with the GLTF definitions.
Test has been conducted using [this](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/StainedGlassLamp) standard sample model
Materials has been loaded and serialized succesfully without any diff.


deltakosh:
The package.json and package.lock should not be in the PR

pandaGaume:
@bghgary I guess the visualisation test should not pass because adding these extensions will definitively change the final webgl result..

packages/dev/serializers/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
```diff
@@ -0,0 +1,91 @@
...
+                const subs = babylonMaterial.subSurface;
+                const transmissionFactor = subs.refractionIntensity == 0 ? undefined : subs.refractionIntensity;
+                const transmissionTexture = this._exporter._glTFMaterialExporter._getTextureInfo(subs.refractionIntensityTexture) ?? undefined;
+
+                const volumeInfo: IKHRMaterialsTransmission = {
+                    transmissionFactor: transmissionFactor,
+                    transmissionTexture: transmissionTexture,
+                    hasTextures: () => {
+                        return this._hasTexturesExtension(babylonMaterial);
+                    },
```
bghgary, line 81: Ugh, I didn't realize this was added to the gltf2interfaces. This shouldn't be there. It's an existing problem, so we can address later.

bghgary:
> @bghgary I guess the visualisation test should not pass because adding these extensions will definitively change the final webgl result..

I don't think any of these tests use the extensions you added, so something is going on.

packages/dev/serializers/src/glTF/2.0/Extensions/KHR_materials_transmission.ts
pandaGaume: Well, i just duplicate and implements the interface
```
    /** @hidden */
    interface IMaterialExtension {
        hasTextures?(): boolean;
    } 
```

bghgary: Yes, I see that. This is an implementation detail of serializer and shouldn't be in the interfaces.

pandaGaume:
@bghgary may i have the model used for the test (which is related to the GLTF serializer) ?. The only reason i imagine it's it may export an extension (which was not possible before) because of the material, or because of a wrong selecting test into the new extensions.
I need the src to reproduce, otherwise i'm blind...

bghgary:
You can figure this out by looking at https://github.com/BabylonJS/Babylon.js/blob/579cb2ee652b179b18dbb0605664a99712e803e1/legacy/tests/validation/config.json and the searching for the title of the test. This will give you a playground to test.

pandaGaume:
Already solve some visualisation tests fails, but still stoked on somes such [this one](https://www.babylonjs-playground.com/#LRFB2D#114) where no serializer involved. So far the result in my local playground is the same as the online playground.

bghgary:
@RaananW Can you take a look when you get back and help us figure out why this is failing? Thanks.

RaananW:
Seems like a Netlify issue, as some tests use assets from the main page and not the cdn.

You can see the error in the logs:

ERR Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://www.babylonjs.com/Assets/NeonPipe/glTF/NeonPipe.gltf. (Reason: CORS request did not succeed). Status code: (null). 


The solution would probably be to move all tests to use the assets cdn. But that's not related to this PR

deltakosh:
Tests are green. Good to go @bghgary ?

bghgary (approved):

PR #12391, opened 2022-04-12
Fix #12174

### LINKED ISSUE #12174: Unicode/emoji characters in Playground are lost upon Save
https://forum.babylonjs.com/t/unicode-emoji-in-playground/28458/2

Unicode/emoji characters in the code of the Playground are lost when the Playground is saved. Not sure if it is the same issue, but it also seems not possible to type emoji characters into the Playground (though you can paste them).

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com): N/A (the bug is present as soon as you save the Playground)
- Expected result: Unicode/emoji characters can be saved with the Playground.
- Current result: Unicode/emoji characters after a Save show up as `??`. If they are used in something like a glb URI, then the URI is broken and the model wont' load.

**Screenshots**
BJS-Playground-Emoji

 - OS: Windows 11
 - Browser: Edge
 - Version: 99.0.1150.39

deltakosh:


deltakosh:
Still need to do it only when required

deltakosh:
Better! Now it is searchable and usable :D

carolhmj (approved):

PR #12397, opened 2022-04-14
DeviceInputSystem: Remove pollInput calls for MouseWheel from EventFactory

PolygonalSun:
This PR contains two changes.  First, all pollInput calls that reference MouseWheelX/Y/Z as it's not necessary for Native to get that value.  Additionally, the Move conditional has been removed from NativeDeviceInputSystem as it's not needed.

### DIFF
--- a/packages/dev/core/src/DeviceInput/InputDevices/deviceTypes.ts
+++ b/packages/dev/core/src/DeviceInput/InputDevices/deviceTypes.ts
@@ -6,7 +6,7 @@ import type { DeviceType, PointerInput, DualShockInput, XboxInput, SwitchInput, 
 export type DeviceInput<T extends DeviceType> = T extends DeviceType.Keyboard | DeviceType.Generic
     ? number
     : T extends DeviceType.Mouse | DeviceType.Touch
-    ? Exclude<PointerInput, PointerInput.Move>
+    ? Exclude<PointerInput, PointerInput.Move | PointerInput.MouseWheelX | PointerInput.MouseWheelY | PointerInput.MouseWheelZ>
     : T extends DeviceType.DualShock
     ? DualShockInput
     : T extends DeviceType.Xbox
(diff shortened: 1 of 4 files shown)


packages/dev/core/src/DeviceInput/InputDevices/deviceTypes.ts
```diff
@@ -6,7 +6,7 @@ import type { DeviceType, PointerInput, DualShockInput, XboxInput, SwitchInput,
 export type DeviceInput<T extends DeviceType> = T extends DeviceType.Keyboard | DeviceType.Generic
     ? number
     : T extends DeviceType.Mouse | DeviceType.Touch
-    ? Exclude<PointerInput, PointerInput.Move>
+    ? Exclude<PointerInput, PointerInput.Move | PointerInput.MouseWheelX | PointerInput.MouseWheelY | PointerInput.MouseWheelZ>
```
bghgary, line 9: Probably should add this to what's new under breaking changes.

PolygonalSun: Updated

what's new.md
```diff
@@ -465,4 +465,5 @@ BABYLON.SceneLoader.OnPluginActivatedObservable.add(function(plugin) {
 ```
 - `currentState` and `previousState` have been removed from use in `onInputChangedObservable` in the `DeviceSourceManager` ([PolygonalSun](https://github.com/PolygonalSun))
 - `PointerInput` movement enums are no longer being used in any movement event handling in the `DeviceInputSystem` and `InputManager` ([PolygonalSun](https://github.com/PolygonalSun))
-- Shadow generators now use the `Material.alphaCutOff` value instead of a hard-coded 0.4 value. ([Popov72](https://github.com/Popov72))
\ No newline at end of file
+- Shadow generators now use the `Material.alphaCutOff` value instead of a hard-coded 0.4 value. ([Popov72](https://github.com/Popov72))
+- `PointerInput` mouse enums are no longer usable with the `DeviceSource`'s `getInput` function for Mouse and Touch ([PolygonalSun](https://github.com/PolygonalSun))
```
bghgary, line 469: I don't think we have a new one for future releases yet. Maybe need to check with @RaananW.

RaananW: https://github.com/BabylonJS/Babylon.js/pull/12397#issuecomment-1099698950

No need to update the what's new file, just make sure the PR is tagged correctly.

RaananW: The change log can be found at CHANGELOG.md on the root directory

PolygonalSun: File reverted.

bghgary: Okay, we should make sure the PR titles are good then since it appears that ends up in the change log.

bghgary (approved):

deltakosh (approved):

PR #12407, opened 2022-04-19
Ext mesh gpu instancing

pandaGaume:
Add support to EXT_mesh_gpu_instancing to Babylon GLTF Serializer.
Thin instances are used to save transformation into GLTF buffer.  
This extension is NOT official and not supported yet by many viewer. This may be used with caution while terminology between Local Transformation and World transformation are not consistent beetwen this Extension and the babylon Thin instances.
Test has been conducted sucessfully with playground and sandbox, using respectively coding samples and imported gltf sample.
The main sample used was the "official" [gltf sample](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/samples/teapots_galore)


deltakosh:
Do you mind adding a visual test so we can keep track of it?

pandaGaume:
@sebavan or @RaananW could you give me direction to add visualtion test ?? already know where and how to update the config.json but do not know how to capture correct image ref and where to store GLTF samples (Assets ??).
G.

RaananW:
Hi, of course.

I was sure i documented it but i can't find it at the moment (and am currently away on my mobile) did this time here and next time well documented:

Run your test on its own with -u as a parameter. This will generate the image for you automatically. You can run next directly, but simpler would be:

npm run test:visualization -- -t "name of test" -i "webgl2" -u

This will run your test using webgl2 and update the snapshot.

Retracing the model - we use our assets cdn for test models. But every model uploaded there has a creative commons license so only submit the model there of its ok for it to be used all over. Submit a PR to the assets library and once merged (10 minutes or so after) it will be available on assets.babylonjs.com
Otherwise you can use any hosting provider to host the file, just make sure CORS is enabled.

pandaGaume:
@RaananW I do not have write access to Asset. Could you save the following glb into the meshes/gpu_instancing dir
[teapotgalore.zip](https://github.com/BabylonJS/Babylon.js/files/8631176/teapotgalore.zip)
thanks.

bghgary:
Actually, the assets repo already has this asset: https://github.com/BabylonJS/Assets/tree/master/meshes/TeapotsGalore

You can access it like this: https://assets.babylonjs.com/meshes/TeapotsGalore/TeapotsGalore.gltf

PR #12415, opened 2022-04-20
Skeleton._sortBones() crash for bones with parents from other skeletons

EvgenyRodygin:
It is possible for a bone to have a parent bone from a skeleton different from the original one.
Currently, the method crashes when this is the case.
This simple check fixes it. Notice that "index" might be -1 when the parent is from another skeleton in the previous iteration.

### DIFF
--- a/packages/dev/core/src/Bones/skeleton.ts
+++ b/packages/dev/core/src/Bones/skeleton.ts
@@ -878,6 +878,8 @@ export class Skeleton implements IAnimatable {
         visited[index] = true;
 
         const bone = this.bones[index];
+        if (!bone) return;
+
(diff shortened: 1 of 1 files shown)


EvgenyRodygin:
> I'm deeply surprise you could get a bone with a different skeleton than the rest of the bones. This is clearly not supported by design
> 
> How do you end up with that?
> 
> Anyway this fix is great

Yeah. Shouldn't be the case I know. You won't get something like this importing regular files I think (might be wrong though).
In our app we create a separate IK skeleton and the root bone of this IK skeleton is attached to the bone from the FK-skeleton. Might be not the best design though that what I encountered.

PR #12419, opened 2022-04-21
ShaderMaterial: keep uniqueId in serialize()

lxq100:
Like Material here: https://github.com/BabylonJS/Babylon.js/blob/c654c7303965a43f340f8c1e0eab3fc804c566f1/packages/dev/core/src/Materials/material.ts#L1719
Related commit: https://github.com/BabylonJS/Babylon.js/commit/738236a7578d1c6dfcae4f93cfffeb768ea5a826


aWeirdo:
I think this is a bug in the material class, uniqueId is a session-based unique id, (at least it used to be), so should not be serialized at all

deltakosh:
They are not session based now :) so this PR is good to go

PR #12421, opened 2022-04-21
Remove some unnecessary computations

deltakosh:


sebavan (approved):

packages/dev/core/src/Rendering/renderingGroup.ts
```diff
@@ -249,9 +249,14 @@ export class RenderingGroup {
             sortedArray.sort(sortCompareFn);
         }
 
+        const scene = sortedArray[0].getMesh().getScene();
```
Popov72, line 252: I think it's possible that `sortedArray.length==0` at that point, so `sortedArray[0].getMesh()` would fail.

packages/dev/core/src/Rendering/renderingGroup.ts
```diff
@@ -273,8 +278,14 @@ export class RenderingGroup {
      * @param subMeshes The submeshes to render
      */
     private static _RenderUnsorted(subMeshes: SmartArray<SubMesh>): void {
+        const scene = subMeshes.data[0].getMesh().getScene();
```
Popov72, line 281: Same comment than above.

packages/dev/core/src/Rendering/renderingGroup.ts
deltakosh: I don't think so as all the call to it check if the array has a length > 0:
https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Rendering/renderingGroup.ts#L136
https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Rendering/renderingGroup.ts#L143
https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Rendering/renderingGroup.ts#L148
https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Rendering/renderingGroup.ts#L170

(Unless I missed one entry point?)

PR #12422, opened 2022-04-21
Exposes framebufferScaleFactor to EngineOptions

imerso:
This change allows the engine user to adjust framebufferScaleFactor on engine creation. This option will allow to reduce resolution in XR mode. For devices like the Oculus Quest, that is imho a very important option.


deltakosh (approved):

RaananW (changes requested):
This is already possible using the canvas options provided to the default experience helper (if you are using it), or providing the option directly when creating the experience yourself.

For example:

```javascript
await scene.createDefaultXRExperienceAsync({
  outputCanvasOptions: {
    canvasOptions: {
      framebufferScaleFactor: aValueOtherThan1
    }
  }
});
```

If that doesn't work it's a bug, but there is no need to add this (XR specific) feature to the thin engine.

imerso:
That works perfectly, thank you!

PR #12426, opened 2022-04-22
Input value accumulation

syntheticmagus:
Draft PR put up to be part of a conversation about input aggregation and sampling.

### DIFF
--- a/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
+++ b/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
@@ -13,10 +13,37 @@ const MAX_KEYCODES = 255;
 // eslint-disable-next-line @typescript-eslint/naming-convention
 const MAX_POINTER_INPUTS = Object.keys(PointerInput).length / 2;
 
+class InputValue {
+    private _currentValue: number;
+    private _accumulatingValue: number;
+
+    public constructor(defaultValue: number = 0) {
+        this._currentValue = defaultValue;
+        this._accumulatingValue = defaultValue;
+    }
+
+    public set(value: number): void {
+        this._currentValue = value;
+    }
+
+    public get value(): number {
+        return this._currentValue;
+    }
+
+    public accumulate(value: number): void {
+        this._accumulatingValue += value;
+    }
+
+    public tick(defaultValue: number = 0): void {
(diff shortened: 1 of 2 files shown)


packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -13,10 +13,37 @@ const MAX_KEYCODES = 255;
...
+    private _currentValue: number;
+    private _accumulatingValue: number;
+
+    public constructor(defaultValue: number = 0) {
+        this._currentValue = defaultValue;
+        this._accumulatingValue = defaultValue;
+    }
+
+    public set(value: number): void {
+        this._currentValue = value;
```
RaananW, line 26: Shouldn't the `accumulatingValue` be set here as well? otherwise it can store incorrect values?

packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -13,10 +13,37 @@ const MAX_KEYCODES = 255;
...
+
+    public get value(): number {
+        return this._currentValue;
+    }
+
+    public accumulate(value: number): void {
+        this._accumulatingValue += value;
+    }
+
+    public tick(defaultValue: number = 0): void {
```
RaananW, line 37: Would be great to have those functions documented. What is the use of this function?

packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -13,10 +13,37 @@ const MAX_KEYCODES = 255;
...
+class InputValue {
+    private _currentValue: number;
+    private _accumulatingValue: number;
+
+    public constructor(defaultValue: number = 0) {
+        this._currentValue = defaultValue;
+        this._accumulatingValue = defaultValue;
+    }
+
+    public set(value: number): void {
```
RaananW, line 25: (Personal opinion :-)) I find a "set" function a bit confusing. It would be better to generate a setter for the `_currentValue` value (especially since it already has a getter).

syntheticmagus:
Discussed in a meeting, confirmed that this kind of change isn't really the way we want to be going with this.

Issue #12204, opened 2022-03-18
Allow arc rotate auto rotation to rotate to a certain alpha value


RaananW:
Add an optional target alpha value to the auto rotate behavior. The behavior will stop when reaching this beta value.

TheCrowd:
@RaananW  could you clarify this issue with more details and can I work on this issue?

RaananW:
@TheCrowd - Of course! Thank you :-)

The arc rotate camera has an auto-rotate behavior. the auto rotate behavior rotates constantly until stopped (for example by a mouse input like a click or a drag).

This feature should allow setting a target alpha that will act as a goal for the auto-rotate behavior. the auto-rotate will stop rotating when reaching this alpha value.

(Note that I originally wrote beta, which actually needed to be alpha. corrected all over).

TheCrowd:
@RaananW Not sure if I understand this correctly, so the goal is achieved, further input won't trigger any rotate behavior?

RaananW:
no - if you drag and change the alpha, the alpha should be animated back to the predefined alpha. It's quite straightforward - if alpha is not the predefined alpha (within an epsilon), run the auto-rotate algorithm. If it's === defined alpha, don't rotate until user input was provided.

TheCrowd:
@RaananW  Thanks for explaining, I will have a try

TheCrowd:
@RaananW  please have a look at this [PR](https://github.com/BabylonJS/Babylon.js/pull/12429)

carolhmj:
Closed by https://github.com/BabylonJS/Babylon.js/pull/12429

PR #12429, opened 2022-04-23
#12204 Add limit to auto rotate

TheCrowd:
add `targetAlpha` to autoRotateBehavior to serve as a limit. When the limit is reached, the autoRotateBehavoir would stop.

### DIFF
--- a/packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
+++ b/packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
@@ -1,5 +1,5 @@
 import type { Behavior } from "../../Behaviors/behavior";
-import type { Camera } from "../../Cameras/camera";
+import { Camera } from "../../Cameras/camera";
 import type { ArcRotateCamera } from "../../Cameras/arcRotateCamera";
 import type { Nullable } from "../../types";
 import type { Observer } from "../../Misc/observable";
@@ -23,7 +23,21 @@ export class AutoRotationBehavior implements Behavior<ArcRotateCamera> {
     private _idleRotationSpeed = 0.05;
     private _idleRotationWaitTime = 2000;
     private _idleRotationSpinupTime = 2000;
+    private _targetAlpha:number;
 
+    /**
+    * Sets the target aplha for Camera to rotate to
+    * @param targetAlpha
+    */
+    public setTargetAlpha(targetAlpha: number): void {
+        this._targetAlpha = targetAlpha;
+    }
+    /**
+     * Gets the target alpha for Camera to rotate to 
+    */    
+    public get targetAlpha(): number {
+        return this._targetAlpha;
+    }
     /**
      * Sets the flag that indicates if user zooming should stop animation.
      */
@@ -123,6 +137,9 @@ export class AutoRotationBehavior implements Behavior<ArcRotateCamera> {
         });
 
         this._onAfterCheckInputsObserver = camera.onAfterCheckInputsObservable.add(() => {
+            if(this._reachTargetAlpha()){
+                return;
+            }
             const now = PrecisionDate.Now;
             let dt = 0;
             if (this._lastFrameTime != null) {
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
```diff
@@ -169,6 +186,17 @@ export class AutoRotationBehavior implements Behavior<ArcRotateCamera> {
         this._lastInteractionTime = customTime ?? PrecisionDate.Now;
     }
 
+    /**
+     * Returns true if camera alpha reaches the target alpha
+     * @returns true if camera alpha reaches the target alpha
+     */
+    private _reachTargetAlpha(): boolean {
+        if (this._attachedCamera && this._targetAlpha) {
+            return this._attachedCamera.alpha <= this._targetAlpha;
```
RaananW, line 195: the alpha inspection should check if the alpha is within an epsilon of the _targetAlpha value. In this case, any time the camera's alpha is higher than the target alpha it will stop. But the user should be able to rotate the camera. if rotated a bit in the rotation direction it should start auto-rotating again.

You can use the Constants.Epsilon value to check the value is within this range.

packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
```diff
@@ -23,7 +23,21 @@ export class AutoRotationBehavior implements Behavior<ArcRotateCamera> {
     private _idleRotationSpeed = 0.05;
     private _idleRotationWaitTime = 2000;
     private _idleRotationSpinupTime = 2000;
+    private _targetAlpha: number;
 
+    /**
+     * Sets the target aplha for Camera to rotate to
+     * @param targetAlpha
+     */
+    public setTargetAlpha(targetAlpha: number): void {
```
RaananW, line 32: instead of adding the setTargetAlpha function, you can use `set targetAlpha(...`, or simply make `targetAlpha` a public Nullable<number> instead of having both a getter and a setter.

RaananW (changes requested):
Thank you so much for this PR!

There are a few comments I added. You also need to allow the user to disable this feature, so make sure the `targetAlpha` is not only a number, but optionally `null` (which should be the default value)

packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
TheCrowd: Thanks for the suggestion. I have implement this in the new commit @RaananW

packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
```diff
@@ -23,6 +24,7 @@ export class AutoRotationBehavior implements Behavior<ArcRotateCamera> {
     private _idleRotationSpeed = 0.05;
     private _idleRotationWaitTime = 2000;
     private _idleRotationSpinupTime = 2000;
+    public _targetAlpha: Nullable<number> = null;
```
RaananW, line 27: public variable should not start with an underscore. you can change it to `targetAlpha`

packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
```diff
@@ -6,6 +6,7 @@ import type { Observer } from "../../Misc/observable";
 import type { PointerInfoPre } from "../../Events/pointerEvents";
 import { PointerEventTypes } from "../../Events/pointerEvents";
 import { PrecisionDate } from "../../Misc/precisionDate";
+import { Epsilon } from "core/Maths/math.constants";
```
RaananW, line 9: you should also load the file directly from its path and not from `core`

packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
RaananW: thanks a lot :-)

packages/dev/core/src/Behaviors/Cameras/autoRotationBehavior.ts
TheCrowd: thanks. see the latest commit

deltakosh (approved):

RaananW (approved):

PR #12433, opened 2022-04-25
Add backward compatibability for param kind to be "color" for thin instance methods

BlakeOne:
If kind "color" is passed to the function thinInstanceSetAttributeAt, it will be converted to "instanceColor", for back compat.
Forum: https://forum.babylonjs.com/t/vertex-error-and-no-colors-on-thin-instances-after-upgrade/28492/19

(Update: added this back compat to all thin instance methods that accept param kind)

### DIFF
--- a/packages/dev/core/src/Meshes/thinInstanceMesh.ts
+++ b/packages/dev/core/src/Meshes/thinInstanceMesh.ts
@@ -168,6 +168,10 @@ Mesh.prototype.thinInstanceSetMatrixAt = function (index: number, matrix: DeepIm
 };
 
 Mesh.prototype.thinInstanceSetAttributeAt = function (kind: string, index: number, value: Array<number>, refresh: boolean = true): boolean {
+    if (kind === VertexBuffer.ColorKind) {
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Meshes/thinInstanceMesh.ts
```diff
@@ -168,6 +168,10 @@ Mesh.prototype.thinInstanceSetMatrixAt = function (index: number, matrix: DeepIm
 };
 
 Mesh.prototype.thinInstanceSetAttributeAt = function (kind: string, index: number, value: Array<number>, refresh: boolean = true): boolean {
+    if (kind === VertexBuffer.ColorKind) {
```
Popov72, line 171: I think this block of code should be added in all the thin instance methods that are taking a "kind" as a parameter?

BlakeOne: OK sounds good, to help finish the back compat…  Pretty sure I got all of them - it was already being done for thinInstanceSetBuffer. :)

deltakosh (approved):

PR #12434, opened 2022-04-25
Fix thinInstanceCount setter to work with mesh clone

BlakeOne:
After cloning a mesh that has thin instances, setting thinInstanceCount on the clone currently doesn't work (unless it is set to 0 the setter will return without updating the internal instanceCount).

With this PR, if a mesh doesn't have the thin instance matrixData, then it checks to see if the mesh's source has it before defaulting the max instance count to 0.

PG Repro: https://playground.babylonjs.com/#217750#76
Forum: https://forum.babylonjs.com/t/cant-set-thininstancecount-for-clone/29674

### DIFF
--- a/packages/dev/core/src/Meshes/thinInstanceMesh.ts
+++ b/packages/dev/core/src/Meshes/thinInstanceMesh.ts
@@ -188,7 +188,7 @@ Object.defineProperty(Mesh.prototype, "thinInstanceCount", {
         return this._thinInstanceDataStorage.instancesCount;
     },
     set: function (this: Mesh, value: number) {
-        const numMaxInstances = (this._thinInstanceDataStorage.matrixData?.length ?? 0) / 16;
+        const numMaxInstances = (this._thinInstanceDataStorage.matrixData?.length ?? (this.source?._thinInstanceDataStorage.matrixData?.length ?? 0)) / 16;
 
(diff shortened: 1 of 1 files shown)


BlakeOne:
Hmm, the format check is failing but the output doesn't say what the issue is, just that it failed... And I can't see what the formatting issue would be?

Popov72:
This is a breaking change, as now the cloned mesh will be displayed with the same thin instances than the source mesh.

But it does not seem unreasonable to me to expect this behaviour when cloning a mesh with thin instances. And if we don't want the thin instances in the cloned mesh, we can call `makeGeometryUnique` on the mesh.

Let's see what the Babylon.js team think about this.

BlakeOne:
> This is a breaking change, as now the cloned mesh will be displayed with the same thin instances than the source mesh.

It won't be thou unless you explicitly set thinInstanceCount to the number of thin instances that you want displayed. Otherwise the behavior should be the same as it is now AFAIK.

Because the only change after the PR should be that E.G. when you set it to 3, you get the first 3 thin instances displayed as expected, if you set it to 2 you get just the first 2 as expected, etc. Currently nothing would happen at all (the setter just returns), which seems pretty unexpected IMO...

sebavan (approved):

PR #12458, opened 2022-04-29
Remove BABYLON namespace references in the Inspector

### LINKED ISSUE #12343: Remove BABYLON namespace references in the Inspector
An example of this is: https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx#L88

If we use the Inspector in a ES6 environment, this leads to errors like:
image
when opening the Tools tab. 

We should look for all of the instances of the namespace and remove them if possible.

RaananW:
Fixes #12343

The build system now allows to set dependencies to both the es6 and the umd packages, so importing from gui-editor is enough

### DIFF
--- a/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/meshes/meshPropertyGridComponent.tsx
+++ b/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/meshes/meshPropertyGridComponent.tsx
@@ -1,7 +1,6 @@
 import * as React from "react";
 
 import type { Observable } from "core/Misc/observable";
-import { Tools } from "core/Misc/tools";
 import { Vector3, TmpVectors } from "core/Maths/math.vector";
 import { Color3 } from "core/Maths/math.color";
 import type { Mesh } from "core/Meshes/mesh";
@@ -36,6 +35,7 @@ import { HexLineComponent } from "shared-ui-components/lines/hexLineComponent";
 import { SkeletonViewer } from "core/Debug/skeletonViewer";
 import type { ShaderMaterial } from "core/Materials/shaderMaterial";
 import type { IInspectableOptions } from "core/Misc/iInspectable";
+import { NormalMaterial } from "materials/normal/normalMaterial";
 
 import "core/Physics/physicsEngineComponent";
 import { ParentPropertyGridComponent } from "../parentPropertyGridComponent";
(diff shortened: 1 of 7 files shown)


packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/meshes/meshPropertyGridComponent.tsx
```diff
@@ -166,11 +166,12 @@ export class MeshPropertyGridComponent extends React.Component<
             mesh.reservedDataStore.originalMaterial = null;
             this.setState({ displayNormals: false });
         } else {
-            if (!(BABYLON as any).NormalMaterial) {
-                this.setState({ displayNormals: true });
-                Tools.LoadScript("https://preview.babylonjs.com/materialsLibrary/babylonjs.materials.js", () => {
-                    this.displayNormals();
-                });
+            if (!NormalMaterial) {
+                // this.setState({ displayNormals: true });
```
carolhmj, line 170: Can't this be removed instead of just commented?

packages/dev/inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx
```diff
@@ -84,8 +85,9 @@ export class ToolsTabComponent extends PaneComponent {
     }
 
     componentDidMount() {
-        if (!(BABYLON as any).GLTF2Export) {
-            Tools.LoadScript("https://preview.babylonjs.com/serializers/babylonjs.serializers.min.js", () => {});
+        if (!GLTF2Export) {
+            console.log("GLTF2Export is not available");
+            // Tools.LoadScript("https://preview.babylonjs.com/serializers/babylonjs.serializers.min.js", () => {});
```
carolhmj, line 90: Same for here

carolhmj (changes requested):

packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/meshes/meshPropertyGridComponent.tsx
RaananW: I added a warning instead of logging. It is needed, as it is possible that the inspector will be used without loading the needed libs

carolhmj (approved):

PR #12461, opened 2022-04-30
Create a TBN NME node to handle tangent space orientation

sebavan:
https://forum.babylonjs.com/t/nodematerial-renders-incorrectly-when-using-mesh-tangents/29784/7


RaananW (approved):

deltakosh (approved):

deltakosh:
Please do not merge before @Popov72 can validate as well as we are touching his PBR nodes

packages/dev/core/src/Materials/Node/Blocks/Dual/TBNBlock.ts
```diff
@@ -0,0 +1,171 @@
...
+            }
+            tangentInput.output.connectTo(this.tangent);
+        }
+    }
+
+    public prepareDefines(mesh: AbstractMesh, nodeMaterial: NodeMaterial, defines: NodeMaterialDefines) {
+        const normal = this.normal;
+        const tangent = this.tangent;
+
+        let normalAvailable = normal.isConnected && normal.isConnectedToInputBlock;
```
Popov72, line 126: Why testing that the normal is coming only from an `InputBlock`? Shouldn't we simply use `let normalAvailable = normal.isConnected;`?

packages/dev/core/src/Materials/Node/Blocks/Dual/TBNBlock.ts
```diff
@@ -0,0 +1,171 @@
...
+        const world = this.world;
+        const TBN = this.TBN;
+
+        if (state.target !== NodeMaterialBlockTargets.Fragment) {
+            // Vertex
+            state._emitVaryingFromString(TBN.associatedVariableName, "mat3");
+
+            state.compilationString += `
+                // ${this.name}
+                vec3 tbnNormal = normalize(${normal.associatedVariableName});
```
Popov72, line 155: You should use `${normal.associatedVariableName}.xyz` has we can input a vec4 for the normal.

packages/dev/core/src/Materials/Node/Blocks/Input/inputBlock.ts
```diff
@@ -133,6 +132,7 @@ export class InputBlock extends NodeMaterialBlock {
                     case "world1":
                     case "world2":
                     case "world3":
+                    case "tangent":
```
Popov72, line 135: Is this change not a breaking change (switching the tangent from vec3 to vec4)? I guess some node materials can break because of that?

Popov72 (commented):
LGTM, only some small comments!

packages/dev/core/src/Materials/Node/Blocks/Dual/TBNBlock.ts
sebavan: Good catch

packages/dev/core/src/Materials/Node/Blocks/Input/inputBlock.ts
sebavan: I added the allow vector3 on the other ones and ensures deserialization was working for the ones saved with vec3

Popov72: Yes, but what if a node material is using a tangent attribute which is not directly connected to an input block but is doing some intermediate computing? It's probably not very common, but maybe adding a line in the "Breaking changes" section that now **tangent** is a vec4 would be appropriate.

PR #12464, opened 2022-05-02
Nine patch overlap fix

Pryme8:
https://forum.babylonjs.com/t/nine-patch-showing-gaps-with-alpha-png/29365/7

image
The edges look weird because I edited it in photoshop before converting it to a blob for testing inline.  I have tested on firefox and brave that the interior artifacts are gone.


deltakosh (approved):

sebavan (approved):

Pryme8:
image
I see this but its saying the PR is still open?

And then there are some azure pipeline errors, what is this asking of me?

PR #12472, opened 2022-05-04
Instance color mixing

CedricGuillemet:
Follow up https://forum.babylonjs.com/t/vertex-error-and-no-colors-on-thin-instances-after-upgrade/28492

Mixing color instances from thin instances and vertex colors from mesh.


packages/dev/core/src/Shaders/color.vertex.fx
```diff
@@ -47,10 +47,14 @@ void main(void) {
...
-#elif INSTANCESCOLOR
-	vColor = instanceColor;
+#if defined(VERTEXCOLOR) || defined(INSTANCESCOLOR)
+    vColor = vec4(1.0);
+    #ifdef VERTEXCOLOR
+        vColor *= color;
+    #endif
+    #ifdef INSTANCESCOLOR
+        vColor *= instanceColor;
+    #endif
```
Popov72, line 57: What about putting this block in an external file that you can #include?

It is reused a lot of time, and if some changes should occur they will have to be done for all the blocks.

RaananW (approved):

PR #12475, opened 2022-05-04
reduce the call to dirty

deltakosh:


packages/dev/core/src/Materials/material.ts
```diff
@@ -304,8 +304,14 @@ export class Material implements IAnimatable {
         if (this._alpha === value) {
             return;
         }
+
+        const oldValue = this._alpha;
         this._alpha = value;
-        this.markAsDirty(Material.MiscDirtyFlag);
+
+        // Only call dirty when there is a state change (no alpha / alpha)
+        if ((oldValue === 1 || value === 1) && oldValue !== value) {
```
RaananW, line 312:
```suggestion
        if (oldValue === 1 || value === 1) {
```

The two values are different for sure (line 304)

PR #12477, opened 2022-05-05
Optimize rendering to reuse material as much as possible

deltakosh:


packages/dev/core/src/Rendering/renderingGroup.ts
```diff
@@ -354,6 +337,25 @@ export class RenderingGroup {
...
+     */
+    public static PainterSortCompare(a: SubMesh, b: SubMesh): number {
+        const meshA = a.getMesh();
+        const meshB = b.getMesh();
+
+        if (meshA.material && meshB.material) {
+            return meshA.material.uniqueId - meshB.material.uniqueId;
+        }        
+
+        return meshA.uniqueId - meshB.uniqueId;
```
RaananW, line 356: very nitty, but - there is always a chance (very low one?)  that the delta between two meshes' uniqueId will be the same as two different meshes with materials. Would that be an issue here?

deltakosh: Nope at the very worse we will miss the cache

PR #12478, opened 2022-05-05
Make AbstractMesh props protected instead of private

barroij:


barroij:
@deltakosh Here you are :)

RaananW (approved):

deltakosh (approved):

PR #12482, opened 2022-05-05
warning and checks for noimpostor children

### LINKED ISSUE #12446: Unable to show Mesh Impostors in Cannon/Oimo
**Repro**
- Bug repro on [playground](https://playground.babylonjs.com): https://playground.babylonjs.com/#66PS52#164
- Expected result: We should be able to see the impostors of the meshes added
- Current result: We see an error
image

CedricGuillemet:
fixes #12446 

2 issues :
- when creating impostor for children mesh, their impostor will effectively be created with the NoImpostor. So shapes are not available and you get the crash. I've added a check and log a warning
- The correct way to debugdisplay impostor in this case is to do the call on the NoImpostor parent. But you have to provide a 2nd parameter or nothing appears. I've also added a warning in that particular case.

### DIFF
--- a/packages/dev/core/src/Debug/physicsViewer.ts
+++ b/packages/dev/core/src/Debug/physicsViewer.ts
@@ -248,7 +248,8 @@ export class PhysicsViewer {
         let mesh: Nullable<AbstractMesh> = null;
         const utilityLayerScene = this._utilityLayer.utilityLayerScene;
         if (!impostor.physicsBody) {
-            Logger.Warn("Unable to get physicsBody of impostor. It might be initialized later with by its parent's impostor.");
+            Logger.Warn("Unable to get physicsBody of impostor. It might be initialized later by its parent's impostor.");
+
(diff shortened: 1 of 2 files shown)


packages/dev/core/src/Debug/physicsViewer.ts
```diff
@@ -246,7 +247,10 @@ export class PhysicsViewer {
 
         let mesh: Nullable<AbstractMesh> = null;
         const utilityLayerScene = this._utilityLayer.utilityLayerScene;
-
+        if (!impostor.physicsBody) {
+            Logger.Warn("Unable to get physicsBody of impostor. It might be initialized later with by its parent's impostor.");
```
deltakosh, line 251:
```suggestion
            Logger.Warn("Unable to get physicsBody of impostor. It might be initialized later by its parent's impostor.");

```

deltakosh (approved):

PR #12488, opened 2022-05-06
Add support for COMPRESSED_SRGB_S3TC_DXT1_EXT

kircher1:
If requesting sRGB for an ST3C **RGB** texture, the previous logic was using an sRGB + **ALPHA** format. This change will use RGB (no alpha) instead.


Popov72 (approved):
LGTM!

PR #12497, opened 2022-05-06
A couple of minor fixes

bghgary:
- Fix the PG link for Draco Compression doc comment as the path to the bin path changed in the glTF-Sample-Models repo.
- Allow empty morph targets in the glTF loader (see [here](https://github.com/KhronosGroup/glTF/issues/2154#issuecomment-1116701277))

### DIFF
--- a/packages/dev/core/src/Meshes/Compression/dracoCompression.ts
+++ b/packages/dev/core/src/Meshes/Compression/dracoCompression.ts
@@ -237,7 +237,7 @@ export interface IDracoCompressionConfiguration {
  *     var vertexData = await DracoCompression.Default.decodeMeshAsync(data);
  * ```
  *
- * @see https://www.babylonjs-playground.com/#N3EK4B#0
+ * @see https://playground.babylonjs.com/#DMZIBD#0
  */
 export class DracoCompression implements IDisposable {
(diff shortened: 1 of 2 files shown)


sebavan:
I wonder if the PR name should change for the readme generation @RaananW  ?

bghgary:
Do we generate a readme entry for every single PR? This doesn't seem like it should have a changelog entry.

RaananW:
Until the next release you can change the title and tag as much as you like. I'm not filtering any PRs in the change log. Any change made will be linked there.
We can add a filter, but isn't it better to list every change made?

bghgary:
> We can add a filter, but isn't it better to list every change made?

This hasn't been the case before. I would think some changes are not useful to list in the changelog. If we have to update the changelog, I will break this PR into two separate ones.

sebavan:
Also when fixing a fix in the same release aka "Shooot!!!! I broke it again" I guess having a way to tag as do not add to the change log might be cool.

RaananW:
> > We can add a filter, but isn't it better to list every change made?
> 
> This hasn't been the case before. I would think some changes are not useful to list in the changelog. If we have to update the changelog, I will break this PR into two separate ones.

I added a label that will not include this PR in the change log.

RaananW (approved):

PR #12498, opened 2022-05-06
Add getNormalsData to AbstractMesh and refactor common part with getP…

carolhmj:
…ositionData

Related forum issue: https://forum.babylonjs.com/t/how-to-get-vertex-normals-of-mesh-with-skeleton-morph/29959/12

Playground: playground.babylonjs.com/#SYQW69#1061

### DIFF
--- a/packages/dev/core/src/Meshes/abstractMesh.ts
+++ b/packages/dev/core/src/Meshes/abstractMesh.ts
@@ -1456,6 +1456,92 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
     }
 
     /**
+     * Iterate through the morph target data and apply influences to data
+     * @param data the data to apply influences to.
+     * @param morphTargetManager the morph target.
+     * @param fn an optional function to execute each loop
+     */
+    private _iterateMorphData(data: FloatArray, morphTargetManager: MorphTargetManager, fn?: () => void) {
+        for (let vertexCount = 0; vertexCount < data.length; vertexCount++) {
+            for (let targetCount = 0; targetCount < morphTargetManager.numTargets; targetCount++) {
+                const targetMorph = morphTargetManager.getTarget(targetCount);
+                const influence = targetMorph.influence;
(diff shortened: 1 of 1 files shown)


sebavan (commented):
Just wondering if a flag in the parameter of the other function would be enough  to only change the transform function ?

RaananW (commented):
It seems like this is mostly the same code as getPositionData, but for a different data type. do you think it'll be possible to reuse the code?

carolhmj:
@sebavan @RaananW code changed to use a single function!

packages/dev/core/src/Meshes/abstractMesh.ts
```diff
@@ -1456,14 +1456,14 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
-     * @returns the position data
+     * Internal function to get buffer data and possibly apply morphs and normals
+     * @param applySkeleton
+     * @param applyMorph
+     * @param data
+     * @param isNormalData
      */
-    public getPositionData(applySkeleton: boolean = false, applyMorph: boolean = false, data?: Nullable<FloatArray>): Nullable<FloatArray> {
-        data = data ?? Tools.Slice(this.getVerticesData(VertexBuffer.PositionKind));
+    private _getData(applySkeleton: boolean = false, applyMorph: boolean = false, data?: Nullable<FloatArray>, isNormalData: boolean = false): Nullable<FloatArray> {
```
RaananW, line 1465: You can accept "kind" and pass the right kind to the function (normal or position)

RaananW (commented):
Just me being nitpicky 😏

I'll let you decide if it makes sense, add i see you are using this flag in the function as well (not only to determine the kind)

packages/dev/core/src/Meshes/abstractMesh.ts
carolhmj: Thanks, it does look cleaner this way :D

RaananW (approved):

PR #12499, opened 2022-05-06
Fix PointerDragBehavior to account for button context, add button filter

PolygonalSun:
This PR contains two changes to `PointerDragBehavior`.
- When a mouse button is pressed to initiate a drag, the button is stored and only a `pointerup` event from that button can release the drag (except for detach and other non-input behaviors).
- There is now an array of mouse buttons that are used to define what buttons are eligible to start a drag behavior.  This can be changed by the user (LMB, MMB, and RMB are set as defaults).

Tested on Chrome, Firefox, Edge, and Safari (MacOS only).


packages/dev/core/src/Behaviors/Meshes/pointerDragBehavior.ts
```diff
@@ -30,10 +30,15 @@ export class PointerDragBehavior implements Behavior<AbstractMesh> {
...
     private _useAlternatePickedPointAboveMaxDragAngleDragSpeed = -1.1;
+    private _activeDragButton: number = -1;
     /**
      * The maximum tolerated angle between the drag plane and dragging pointer rays to trigger pointer events. Set to 0 to allow any angle (default: 0)
      */
     public maxDragAngle = 0;
+    /**
+     * Butttons that can be used to initiate a drag
+     */
+    public dragButtons = [0, 1, 2];
```
RaananW, line 41: shouldn't we limit this per default (while allowing other buttons to be added?) or will it be a breaking change?

RaananW (commented):
LGTM! Just a small question

packages/dev/core/src/Behaviors/Meshes/pointerDragBehavior.ts
RaananW: I am fine with whatever you decide :-) You can resolve this whenever you want.

RaananW (approved):

PR #12502, opened 2022-05-09
Fix ubo not of the right size when using material plugins

Popov72:
See https://forum.babylonjs.com/t/pluginmaterial-bug/30098

To fix the problem I simply generate a specific #define for each plugin and use this define in the vertex/fragment shaders when the plugin is used.

I have also added a visualization test to test this case.


deltakosh:
you rock!

deltakosh (approved):

sebavan (approved):

PR #12505, opened 2022-05-09
Fix BoundingBoxGizmo rotate incorrectly when using RightHand system.

gongsiyi123:
This PR fix a bug of BoundingBoxGizmo. when scene.useRightHandedSystem = true, BoundingBoxGizmo rotate to opposite direction.
Related issue: [https://forum.babylonjs.com/t/bounding-box-gizmo-rotate-incorrectly-when-scene-is-using-right-hand-system/30206](url)
Playground: [https://playground.babylonjs.com/#SG9ZZB#15](url)

### DIFF
--- a/packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
+++ b/packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
@@ -358,6 +358,11 @@ export class BoundingBoxGizmo extends Gizmo {
 
                         // Rotate around center of bounding box
                         this._anchorMesh.addChild(this.attachedMesh, Gizmo.PreserveScaling);
+                        if (this._anchorMesh.getScene().useRightHandedSystem) {
+                            this._tmpQuaternion.x *= -1;
+                            this._tmpQuaternion.y *= -1;
+                            this._tmpQuaternion.z *= -1;
+                        }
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
```diff
@@ -358,6 +358,11 @@ export class BoundingBoxGizmo extends Gizmo {
 
                         // Rotate around center of bounding box
                         this._anchorMesh.addChild(this.attachedMesh, Gizmo.PreserveScaling);
+                        if (this._anchorMesh.getScene().useRightHandedSystem) {
+                            this._tmpQuaternion.x *= -1;
```
CedricGuillemet, line 362: Please use https://github.com/BabylonJS/Babylon.js/blob/5406aca3f55afdda974600032c21af71aae4190a/packages/dev/core/src/Maths/math.vector.ts#L3688

CedricGuillemet (changes requested):

packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
gongsiyi123: Thanks for you guidance, I replaced it with conjugateInPlace() method :)

CedricGuillemet (approved):

PR #12509, opened 2022-05-10
Add a warning when using ignored options with Dashed Lines update.

carolhmj:
From: https://forum.babylonjs.com/t/dashed-line-not-correctly-updated/29250 this is to make the behavior clearer to users.

### DIFF
--- a/packages/dev/core/src/Meshes/Builders/linesBuilder.ts
+++ b/packages/dev/core/src/Meshes/Builders/linesBuilder.ts
@@ -7,7 +7,7 @@ import type { FloatArray, Nullable } from "../../types";
 import { LinesMesh } from "../../Meshes/linesMesh";
 import type { Scene } from "../../scene";
 import { VertexBuffer } from "../../Buffers/buffer";
-import { Tools } from "../../Misc";
+import { Tools } from "../../Misc/tools";
 
 declare type Material = import("../../Materials/material").Material;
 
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Meshes/Builders/linesBuilder.ts
```diff
@@ -7,6 +7,7 @@ import type { FloatArray, Nullable } from "../../types";
 import { LinesMesh } from "../../Meshes/linesMesh";
 import type { Scene } from "../../scene";
 import { VertexBuffer } from "../../Buffers/buffer";
+import { Tools } from "../../Misc";
```
darraghjburke, line 10: Looks like there is a build error relating to this import

carolhmj: Thanks for catching that, fixed it!

packages/dev/core/src/Meshes/Builders/linesBuilder.ts
```diff
@@ -7,6 +7,7 @@ import type { FloatArray, Nullable } from "../../types";
 import { LinesMesh } from "../../Meshes/linesMesh";
 import type { Scene } from "../../scene";
 import { VertexBuffer } from "../../Buffers/buffer";
+import { Tools } from "../../Misc/tools";
```
sebavan, line 10: Can we use logger instead of tools as an import ? to keep tree shaking happy :-)

sebavan (changes requested):
only one tiny comment

packages/dev/core/src/Meshes/Builders/linesBuilder.ts
carolhmj: Done!

sebavan (approved):

PR #12510, opened 2022-05-10
Make NME TBNBlock fragment compatible

MiikaH:
Allows using fragment as well as vertex inputs with the TBNBlock node. This can for example be used to flip backface normals.

This change still works with vertex normal/tangent inputs, like this playground example: https://playground.babylonjs.com/#VL1FPT#5

But now it also works with fragment inputs, such as this Playground that flips backface normals and tangents and therefore rendering the backside correctly: https://playground.babylonjs.com/#VL1FPT#7
This one currently fails with `Uncaught (in promise) FrontFacingBlock must only be used in a fragment shader` error in the master branch but works with this PR change.

### DIFF
--- a/packages/dev/core/src/Materials/Node/Blocks/Dual/TBNBlock.ts
+++ b/packages/dev/core/src/Materials/Node/Blocks/Dual/TBNBlock.ts
@@ -21,9 +21,9 @@ export class TBNBlock extends NodeMaterialBlock {
     public constructor(name: string) {
         super(name, NodeMaterialBlockTargets.VertexAndFragment, true);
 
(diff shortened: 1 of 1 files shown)


sebavan:
I guess the varying registration should also be removed ?

MiikaH:
> I guess the varying registration should also be removed ?

Hmm, where is that? It already removes the `_emitVaryingFromString` if that is what you meant.

sebavan:
Yup I meant that one.... forget me lol :-)

sebavan:
I added @Popov72 the hero of nme PBR nodes and plenty of other things for review.

Popov72 (approved):
LGTM!

PR #12511, opened 2022-05-11
Plumb through useSRGBBuffer for _createInternalTexture.

kircher1:
This is a follow up change to #12486; A little more plumbing is required to get the useSRGBBuffer value applied to the texture format. With current logic, `fullOptions.useSRGBBuffer` is always undefined.


kircher1:
Note, I plumbed this into the WebGPU implementation as well.

Popov72 (approved):

PR #12512, opened 2022-05-11
Add custom word divider for TextBlock

miudit:
Some languages such as Japanese and Chinese has no "word divider" (the character between words).
Currently, when rendering Japanese in TextBlock, an unnecessary space is inserted between words.
jsfiddle: [https://jsfiddle.net/3ph9m0cx/](https://jsfiddle.net/3ph9m0cx/)

This PR makes it customizable.
In case of Japanese, setting `wordDivider = ""` resolves it.

### DIFF
--- a/packages/dev/gui/src/2D/controls/textBlock.ts
+++ b/packages/dev/gui/src/2D/controls/textBlock.ts
@@ -49,7 +49,7 @@ export class TextBlock extends Control {
     private _outlineColor: string = "white";
     private _underline: boolean = false;
     private _lineThrough: boolean = false;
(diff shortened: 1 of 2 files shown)


sebavan:
@darraghjburke could you double check this one ? adding @RaananW as well

RaananW (approved):
Great idea! I'll wait for @darraghjburke for the final approval

darraghjburke (approved):
LGTM! For future reference, could you provide a playground link showcasing the behavior rather than a JSFiddle? That would make it easier for us to test on the snapshot

PR #12515, opened 2022-05-11
Exposed Native Engine capabilities to Babylon.js, fixes morph targets on Babylon Native Android

SergioRZMasson:
## Overview

Morph targets are currently failing on Babylon Native when running on Android since bgfx does not support Texture2D arrays for Android API BGFX_CONFIG_RENDERER_OPENGLES lower than 31 (currently set to 30). The appropriate way of handling this type of problem would be to check bgfx capabilities and handle it appropriately on the JS side by setting ```canUseTextureForTargets``` property at ```MorphTargetManager```. This PR exposes all capabilities info provided by the Native Engine so they can be accessed by the js NativeEngine and appropriately handled by Babylon.js 

## Aditions

- Exposed Native engine caps to javascript. 
- After this PR gets merged the following [Babylon Native PR](https://github.com/BabylonJS/BabylonNative/pull/1075) can be merged.
- Fixes [Android Morph Targets](https://github.com/BabylonJS/BabylonNative/issues/1049) issue on Babylon Native.


RaananW (approved):
Waiting for @sebavan review as well before merging

packages/dev/core/src/Engines/webgpuEngine.ts
```diff
@@ -810,6 +810,7 @@ export class WebGPUEngine extends Engine {
             supportSRGBBuffers: true,
             supportTransformFeedbacks: false,
             textureMaxLevel: true,
+            texture2DArrayMaxLayerCount: this._webGLVersion > 1 ? 256 : 128,
```
deltakosh, line 813: not sure what it means here to have a webglversion (for webgpu)

packages/dev/core/src/Engines/nullEngine.ts
```diff
@@ -152,6 +152,7 @@ export class NullEngine extends Engine {
             supportSRGBBuffers: false,
             supportTransformFeedbacks: false,
             textureMaxLevel: false,
+            texture2DArrayMaxLayerCount: this._webGLVersion > 1 ? 256 : 128,
```
deltakosh, line 155: not sure what it means here to have a webglversion (for nullengine we do't care)

packages/dev/core/src/Engines/webgpuEngine.ts
SergioRZMasson: You are right @deltakosh, this value should be set to 2048 for webgpu. Updated in the latest commit.

packages/dev/core/src/Engines/Native/nativeInterfaces.ts
```diff
@@ -87,6 +87,31 @@ interface INativeEngineConstructor {
...
+    readonly CAPS_LIMITS_MAX_INDEX_BUFFERS: number;
+    readonly CAPS_LIMITS_MAX_VERTEX_BUFFERS: number;
+    readonly CAPS_LIMITS_MAX_DYNAMIC_INDEX_BUFFERS: number;
+    readonly CAPS_LIMITS_MAX_DYNAMIC_VERTEX_BUFFERS: number;
+    readonly CAPS_LIMITS_MAX_UNIFORMS: number;
+    readonly CAPS_LIMITS_MAX_OCCLUSION_QUERIES: number;
+    readonly CAPS_LIMITS_MAX_ENCODERS: number;
+    readonly CAPS_LIMITS_MIN_RESOURCE_CB_SIZE: number;
+    readonly CAPS_LIMITS_TRANSIENT_VB_SIZE: number;
+    readonly CAPS_LIMITS_TRANSIENT_IB_SIZE: number;
```
bghgary, line 113: Are these all being used?

packages/dev/core/src/Engines/nativeEngine.ts
```diff
@@ -888,6 +888,7 @@ export class NativeEngine extends Engine {
             supportSRGBBuffers: true,
             supportTransformFeedbacks: false,
             textureMaxLevel: false,
+            texture2DArrayMaxLayerCount: _native.Engine.CAPS_LIMITS_MAX_TEXTURE_LAYERS,
```
bghgary, line 891: We need to increment the PROTOCOL_VERSION for this change.

packages/dev/core/src/Engines/Native/nativeInterfaces.ts
SergioRZMasson: No, we currently only use CAPS_LIMITS_MAX_TEXTURE_LAYERS, but I'm exposing all limits provided by bgfx in case the native engine needs any of them in the future. This way we don't have to change the PROTOCOL_VERSION again.

bghgary: I'm not sure about this. My guess is that you will have to update the protocol version anyways as this list may not translate correctly.

SergioRZMasson: I've updated the exposed caps to only have the ones we currently require or have a direct equivalent on the js caps.

bghgary (approved):

PR #12523, opened 2022-05-13
Add new prepass to compute the specularity-glossiness map whatever the material type

Mannns:
This is the PR following this issue:
https://forum.babylonjs.com/t/prepass-get-accurate-reflectance-information/30351

The aim is to compute reflectance and glossiness information whatever the material type (standard, PBR, PBR Specular-Glossiness, PBR Metallic-Roughness). The architecture is based on the other PrePasses. The translation model from Metallic-Roughness to Specular-Glossiness is based on: https://marmoset.co/posts/pbr-texture-conversion.

### DIFF
--- a/packages/dev/core/src/Rendering/geometryBufferRenderer.ts
+++ b/packages/dev/core/src/Rendering/geometryBufferRenderer.ts
@@ -17,6 +17,10 @@ import type { Observer } from "../Misc/observable";
 import type { Engine } from "../Engines/engine";
 import type { Nullable } from "../types";
 import { Material } from "../Materials/material";
+import { StandardMaterial } from "../Materials/standardMaterial";
+import { PBRMaterial } from "../Materials/PBR/pbrMaterial";
+import { PBRMetallicRoughnessMaterial } from "../Materials/PBR/pbrMetallicRoughnessMaterial";
+import { PBRSpecularGlossinessMaterial } from "../Materials/PBR/pbrSpecularGlossinessMaterial";
 
 import "../Shaders/geometry.fragment";
 import "../Shaders/geometry.vertex";
@@ -270,14 +274,14 @@ export class GeometryBufferRenderer {
     }
 
     /**
-     * Gets a boolean indicating if objects roughness are enabled in the G buffer.
+     * Gets a boolean indicating if objects reflectivity are enabled in the G buffer.
      */
     public get enableReflectivity(): boolean {
         return this._enableReflectivity;
     }
 
     /**
-     * Sets whether or not objects roughness are enabled for the G buffer.
+     * Sets whether or not objects reflectivity are enabled for the G buffer.
      */
     public set enableReflectivity(enable: boolean) {
         this._enableReflectivity = enable;
@@ -360,14 +364,102 @@ export class GeometryBufferRenderer {
             }
 
             if (this._enableReflectivity) {
-                if (material.specularTexture) {
-                    defines.push("#define HAS_SPECULAR");
-                    defines.push(`#define REFLECTIVITY_UV${material.specularTexture.coordinatesIndex + 1}`);
-                    needUv = true;
-                } else if (material.reflectivityTexture) {
-                    defines.push("#define HAS_REFLECTIVITY");
-                    defines.push(`#define REFLECTIVITY_UV${material.reflectivityTexture.coordinatesIndex + 1}`);
-                    needUv = true;
+                let metallicWorkflow = false;
+                // for PBR materials: cf. https://doc.babylonjs.com/divingDeeper/materials/using/masterPBR
+                if (material instanceof PBRMetallicRoughnessMaterial) {
+                    // if it is a PBR material in MetallicRoughness Mode:
+                    if (material.metallicRoughnessTexture != null) {
+                        defines.push("#define ORMTEXTURE");
+                        defines.push(`#define REFLECTIVITY_UV${material.metallicRoughnessTexture.coordinatesIndex + 1}`);
+                        defines.push("#define METALLICWORKFLOW");
+                        needUv = true;
+                        metallicWorkflow = true;
+                    }
+                    if (material.metallic != null) {
+                        defines.push("#define METALLIC");
+                        defines.push("#define METALLICWORKFLOW");
+                        metallicWorkflow = true;
+                    }
+                    if (material.roughness != null) {
+                        defines.push("#define ROUGHNESS");
+                        defines.push("#define METALLICWORKFLOW");
+                        metallicWorkflow = true;
+                    }
+                    if (metallicWorkflow) {
+                        if (material.baseTexture != null) {
+                            defines.push("#define ALBEDOTEXTURE");
+                            defines.push(`#define ALBEDO_UV${material.baseTexture.coordinatesIndex + 1}`);
+                            needUv = true;
+                        } else if (material.baseColor != null) {
+                            defines.push("#define ALBEDOCOLOR");
+                        }
+                    }
+                } else if (material instanceof PBRSpecularGlossinessMaterial) {
+                    // if it is a PBR material in Specular/Glossiness Mode:
+                    if (material.specularGlossinessTexture != null) {
+                        defines.push("#define SPECULARGLOSSINESSTEXTURE");
+                        defines.push(`#define REFLECTIVITY_UV${material.specularGlossinessTexture.coordinatesIndex + 1}`);
+                        needUv = true;
+                    } else {
+                        if (material.specularColor != null) {
+                            defines.push("#define REFLECTIVITYCOLOR");
+                        }
+                    }
+                    if (material.glossiness != null) {
+                        defines.push("#define GLOSSINESSS");
+                    }
+                } else if (material instanceof PBRMaterial) {
+                    // if it is the bigger PBRMaterial
+                    if (material.metallicTexture != null) {
+                        defines.push("#define ORMTEXTURE");
+                        defines.push(`#define REFLECTIVITY_UV${material.metallicTexture.coordinatesIndex + 1}`);
+                        defines.push("#define METALLICWORKFLOW");
+                        needUv = true;
+                        metallicWorkflow = true;
+                    }
+                    if (material.metallic != null) {
+                        defines.push("#define METALLIC");
+                        defines.push("#define METALLICWORKFLOW");
+                        metallicWorkflow = true;
+                    }
+
+                    if (material.roughness != null) {
+                        defines.push("#define ROUGHNESS");
+                        defines.push("#define METALLICWORKFLOW");
+                        metallicWorkflow = true;
+                    }
+
+                    if (metallicWorkflow) {
+                        if (material.albedoTexture != null) {
+                            defines.push("#define ALBEDOTEXTURE");
+                            defines.push(`#define ALBEDO_UV${material.albedoTexture.coordinatesIndex + 1}`);
+                            needUv = true;
+                        } else if (material.albedoColor != null) {
+                            defines.push("#define ALBEDOCOLOR");
+                        }
+                    } else {
+                        // SpecularGlossiness Model
+                        if (material.reflectivityTexture != null) {
+                            defines.push("#define SPECULARGLOSSINESSTEXTURE");
+                            defines.push(`#define REFLECTIVITY_UV${material.reflectivityTexture.coordinatesIndex + 1}`);
+                            needUv = true;
+                        } else if (material.reflectivityColor != null) {
+                            defines.push("#define REFLECTIVITYCOLOR");
+                        }
+                        if (material.microSurface != null) {
+                            defines.push("#define GLOSSINESSS");
+                        }
+                    }
+                } else if (material instanceof StandardMaterial) {
+                    // if StandardMaterial:
+                    if (material.specularTexture != null) {
+                        defines.push("#define REFLECTIVITYTEXTURE");
+                        defines.push(`#define REFLECTIVITY_UV${material.specularTexture.coordinatesIndex + 1}`);
+                        needUv = true;
+                    }
+                    if (material.specularColor != null) {
+                        defines.push("#define REFLECTIVITYCOLOR");
+                    }
                 }
             }
 
@@ -483,13 +575,18 @@ export class GeometryBufferRenderer {
                             "mPreviousBones",
                             "bumpMatrix",
                             "reflectivityMatrix",
+                            "albedoMatrix",
+                            "reflectivityColor",
+                            "albedoColor",
+                            "metallic",
+                            "glossiness",
                             "vTangentSpaceParams",
                             "vBumpInfos",
                             "morphTargetInfluences",
                             "morphTargetTextureInfo",
                             "morphTargetTextureIndices",
                         ],
-                        samplers: ["diffuseSampler", "bumpSampler", "reflectivitySampler", "morphTargets"],
+                        samplers: ["diffuseSampler", "bumpSampler", "reflectivitySampler", "albedoSampler", "morphTargets"],
                         defines: join,
                         onCompiled: null,
                         fallbacks: null,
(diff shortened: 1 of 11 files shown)


packages/dev/core/src/Shaders/pbr.fragment.fx
```diff
@@ -678,6 +678,26 @@ void main(void) {
             gl_FragData[PREPASS_REFLECTIVITY_INDEX] = vec4(0.0, 0.0, 0.0, writeGeometryInfo);
         #endif
     #endif
+        #ifdef PREPASS_SPECULARGLOSSINESS
```
sebavan, line 681: I am really not sure about the maths in all this part.

I guess you should rely on roughness and specularEnvironmentR0 which are computed in both the Metal and spec gloss case with the actual value used in the shader.

sebavan (changes requested):

packages/dev/core/src/Shaders/default.fragment.fx
```diff
@@ -457,6 +457,14 @@ color.rgb = max(color.rgb, 0.);
             gl_FragData[PREPASS_REFLECTIVITY_INDEX] = vec4(0.0, 0.0, 0.0, writeGeometryInfo);
         #endif
     #endif
+	#ifdef PREPASS_SPECULARGLOSSINESS
+		// specularity = F0 value
+		#if defined(SPECULAR)
+			gl_FragData[PREPASS_SPECULARGLOSSINESS_INDEX] = vec4(specularMapColor.rgb * writeGeometryInfo, specularMapColor.a * writeGeometryInfo); // no specularity if no visibility
```
CraigFeldspar, line 463: It should probably be `specularColor`, which works whether the material has a spec texture or just a spec color. Reflectivity has the wrong variable too, so you can also fix it just above.

And little nit : you can just write `gl_FragData[PREPASS_SPECULARGLOSSINESS_INDEX] = vec4(specularColor, 1.0) * writegeometryInfo`

CraigFeldspar (commented):
This new SPECULARGLOSSINESS texture seems to be an improvement of our existing reflectivity texture, as it carries the same info in the default shader, and takes both workflows into account inside the PBR shader.

Shouldn't it then be more interesting to replace the content of the REFLECTIVITY texture instead of creating a new one ?
WDYT @sebavan ?

packages/dev/core/src/Shaders/pbr.fragment.fx
Mannns: Hello,

Thank you for your review.

- As regards the specularity part:

The metalness acts as a mask.
When the metalness is high, the corresponding specularity is equal to 
the albedo. When the metalness is low, the corresponding specularity is 
(0.04, 0.04, 0.04), default value for dielectric materials.

In my calculus I assumed that the reflectance value is equal to the 
metalness definded in the ORM texture time the metallic attribute of PBR 
Metallic-Roughness model. But I may be wrong for this part.

At the end I have the same result than using specularEnvironmentR0 for the metallic workflow so I guess my calculus is good. 
However, if the specularEnvironmentR0 is equal to what I was trying to compute it's perfect.

- Regarding the roughness:

It seems that microSurface is also computed for both the metallic-roughness and spec-glossiness models. I guess it’s better to keep microSurface instead of 1.0 – roughness to avoid additional computation.

- Thus, if you agree I will change all this part by: 
gl_FragData[PREPASS_SPECULARGLOSSINESS_INDEX] = vec4(specularEnvironmentR0, microSurface) * writeGeometryInfo;

packages/dev/core/src/Shaders/default.fragment.fx
Mannns: Hi,  

Thanks for your review!
I changed the PREPASS_REFLECTIVITY and PREPASS_SPECULARGLOSSINESS to use specularColor as you suggested.

packages/dev/core/src/Shaders/pbr.fragment.fx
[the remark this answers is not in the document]
sebavan: Almost looks like a bug fix to me :-)

packages/dev/core/src/Shaders/default.fragment.fx
sebavan: @CraigFeldspar, would love your feedback on this one ?

CraigFeldspar: Is it failing on the PG because the mirror's spec is 1,1,1 ?

Mannns: > Is it failing on the PG because the mirror's spec is 1,1,1 ?
It seems to. In this environment, when I explicitly define its specularColor to 0, 0, 0 the rendering is goo
PGSpecLow
PGBug
d

packages/dev/core/src/Shaders/pbr.fragment.fx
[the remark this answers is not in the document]
Mannns: Yes I can merge both PrePass into PREPASS_REFLECTIVITY

packages/dev/core/src/Shaders/default.fragment.fx
Mannns: When I remove the SSR post process the problem disappears too. I guess a high reflectivity makes the SRR artifacts more visible...

packages/dev/core/src/Shaders/pbr.fragment.fx
Mannns: Hello,

Don't know why, the specularEnvironnementR0 is really darker than the baseReflectivity (for PBR spec-glossiness model)
image
image

packages/dev/core/src/Shaders/default.fragment.fx
CraigFeldspar: For me it makes sense that a mirror texture + a SSR post process have some artifacts. Can you fix the test so the mirror spec value is 0 ?

CraigFeldspar: to do that : change the PG url in the json file containing all tests, run the validation tests, and right click on the new image to save and replace the old validation image

packages/dev/core/src/Shaders/pbr.fragment.fx
sebavan: This is because in PBR, the color are in linear space and not gamma. @CraigFeldspar do you know the expected space for post processes using it ??? We should put them all consistent across materials so either calling toLinear in Standard and such or toGamma in PBR to ensure we have the expected output.

CraigFeldspar: AFAIK, SSR is the only one using reflectivity - and it should use linear space. Calling `toLinear` in standard is the way to go.
Also, not altering the PBR + SSR output seems the best call to me

CraigFeldspar (approved):
LGTM, nice fixes on SSR, thanks for that

packages/dev/core/src/Shaders/pbr.fragment.fx
Mannns: > This is because in PBR, the color are in linear space and not gamma. @CraigFeldspar do you know the expected space for post processes using it ??? We should put them all consistent across materials so either calling toLinear in Standard and such or toGamma in PBR to ensure we have the expected output.

Both images come from PBR. But the first one calls baseReflectivity and the other specularEnvironmentR0.
So, is baseReflectivity in gamma space and specularEnvironmentR0 in linear space ?

Mannns: > AFAIK, SSR is the only one using reflectivity - and it should use linear space. Calling `toLinear` in standard is the way to go. Also, not altering the PBR + SSR output seems the best call to me

Not sure I did it well. To me specularity is currently not consistent across PBR spec-glossiness and PBR metal-roughness since I don't use specularEnvironmentR0 for both cases.

If I call toLinearSpace in default it will affect standard + SSR (reflectivity will be darker)
If I call specularEnvironmentR0 for both PBR spec-gloss and PBR metal-roughness it will alter PBR + SSR too

Mannns: It also has to be consistent with geometryBuffer data I guess ?

Mannns: As far as I understand, reflectivity was returned in gamma space everywhere (Standard, PBR spec-Glossy for both prePass and geometryBuffer).
It seems more logical to only apply toGammaSpace(specularEnvironmentR0) in order to return reflectivity of Metal-roughness material in the same space than other materials (rather than changing the space of reflectivity for Standard and PBR spec-Glossy and having to deal with regression)

sebavan: > As far as I understand, reflectivity was returned in gamma space everywhere (Standard, PBR spec-Glossy for both prePass and geometryBuffer).
> It seems more logical to only apply toGammaSpace(specularEnvironmentR0) in order to return reflectivity of Metal-roughness material in the same space than other materials (rather than changing the space of reflectivity for Standard and PBR spec-Glossy and having to deal with regression)

This makes perfect sense to me ? I guess it is the latest required change with the Geometry Buffer one that needs to be in sync.

@Mannns you rock !!!

Mannns: Hi @sebavan 

I just added changes to extend reflectivity pass of the Geometry Buffer to all material types.
Since I missed a lot of information (and in particular specularEnvironmentR0) in the G-Buffer, I computed specularity-glossiness with the formula you didn’t approve at the beginning…  The reflectivity seems to match with the one in the PrePass, but if you know an other way to compute it please tell me. 

Plus, my code in the G-Buffer doesn’t take transparency into account, but it was already not taken into account before.

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -704,14 +801,90 @@ export class GeometryBufferRenderer {
...
-                    // Roughness
+                    // Reflectivity
                     if (this._enableReflectivity) {
-                        if (material.specularTexture) {
-                            effect.setMatrix("reflectivityMatrix", material.specularTexture.getTextureMatrix());
-                            effect.setTexture("reflectivitySampler", material.specularTexture);
-                        } else if (material.reflectivityTexture) {
-                            effect.setMatrix("reflectivityMatrix", material.reflectivityTexture.getTextureMatrix());
-                            effect.setTexture("reflectivitySampler", material.reflectivityTexture);
+                        // if (material.specularTexture) {
```
sebavan, line 806: Remove comment

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -704,14 +801,90 @@ export class GeometryBufferRenderer {
...
+                        //     effect.setMatrix("reflectivityMatrix", material.specularTexture.getTextureMatrix());
+                        //     effect.setTexture("reflectivitySampler", material.specularTexture);
+                        // } else if (material.reflectivityTexture) {
+                        //     effect.setMatrix("reflectivityMatrix", material.reflectivityTexture.getTextureMatrix());
+                        //     effect.setTexture("reflectivitySampler", material.reflectivityTexture);
+                        // }
+                        // for PBR materials: cf. https://doc.babylonjs.com/divingDeeper/materials/using/masterPBR
+                        if (material instanceof PBRMetallicRoughnessMaterial) {
```
sebavan, line 814: Replace instanceof by getClassName() for speed purpose

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -704,14 +801,90 @@ export class GeometryBufferRenderer {
...
+                            if (material.roughness != null) {
+                                effect.setFloat("glossiness", 1.0 - material.roughness);
+                            }
+                            if (material.baseTexture != null) {
+                                effect.setTexture("albedoSampler", material.baseTexture);
+                                effect.setMatrix("albedoMatrix", material.baseTexture.getTextureMatrix());
+                            } else if (material.baseColor != null) {
+                                effect.setColor3("albedoColor", material.baseColor);
+                            }
+                        } else if (material instanceof PBRSpecularGlossinessMaterial) {
```
sebavan, line 832: Replace instanceof by getClassName() for speed purpose

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -704,14 +801,90 @@ export class GeometryBufferRenderer {
...
+                                effect.setMatrix("reflectivityMatrix", material.specularGlossinessTexture.getTextureMatrix());
+                            } else {
+                                if (material.specularColor != null) {
+                                    effect.setColor3("reflectivityColor", material.specularColor);
+                                }
+                            }
+                            if (material.glossiness != null) {
+                                effect.setFloat("glossiness", material.glossiness);
+                            }
+                        } else if (material instanceof PBRMaterial) {
```
sebavan, line 845: Replace instanceof by getClassName() for speed purpose

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -360,14 +364,102 @@ export class GeometryBufferRenderer {
...
-                    defines.push("#define HAS_SPECULAR");
-                    defines.push(`#define REFLECTIVITY_UV${material.specularTexture.coordinatesIndex + 1}`);
-                    needUv = true;
-                } else if (material.reflectivityTexture) {
-                    defines.push("#define HAS_REFLECTIVITY");
-                    defines.push(`#define REFLECTIVITY_UV${material.reflectivityTexture.coordinatesIndex + 1}`);
-                    needUv = true;
+                let metallicWorkflow = false;
+                // for PBR materials: cf. https://doc.babylonjs.com/divingDeeper/materials/using/masterPBR
+                if (material instanceof PBRMetallicRoughnessMaterial) {
```
sebavan, line 369: Replace instanceof by getClassName() for speed purpose

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -360,14 +364,102 @@ export class GeometryBufferRenderer {
...
+                    if (metallicWorkflow) {
+                        if (material.baseTexture != null) {
+                            defines.push("#define ALBEDOTEXTURE");
+                            defines.push(`#define ALBEDO_UV${material.baseTexture.coordinatesIndex + 1}`);
+                            needUv = true;
+                        } else if (material.baseColor != null) {
+                            defines.push("#define ALBEDOCOLOR");
+                        }
+                    }
+                } else if (material instanceof PBRSpecularGlossinessMaterial) {
```
sebavan, line 397: Replace instanceof by getClassName() for speed purpose

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -360,14 +364,102 @@ export class GeometryBufferRenderer {
...
+                        needUv = true;
+                    } else {
+                        if (material.specularColor != null) {
+                            defines.push("#define REFLECTIVITYCOLOR");
+                        }
+                    }
+                    if (material.glossiness != null) {
+                        defines.push("#define GLOSSINESSS");
+                    }
+                } else if (material instanceof PBRMaterial) {
```
sebavan, line 411: Replace instanceof by getClassName() for speed purpose

packages/dev/core/src/Rendering/geometryBufferRenderer.ts
```diff
@@ -360,14 +364,102 @@ export class GeometryBufferRenderer {
...
+                            defines.push(`#define REFLECTIVITY_UV${material.reflectivityTexture.coordinatesIndex + 1}`);
+                            needUv = true;
+                        } else if (material.reflectivityColor != null) {
+                            defines.push("#define REFLECTIVITYCOLOR");
+                        }
+                        if (material.microSurface != null) {
+                            defines.push("#define GLOSSINESSS");
+                        }
+                    }
+                } else if (material instanceof StandardMaterial) {
```
sebavan, line 453: Replace instanceof by getClassName() for speed purpose

Mannns: Thank you very much for all your feedback!

sebavan (approved):
LGTM

sebavan:
@Popov72 and @CraigFeldspar can you have a last eye ball at this PR ? thanks a ton

CraigFeldspar (approved):
LGTM

sebavan:
@Popov72 do you see any issues before I merge ?

Popov72 (commented):
I trust you, @sebavan and @CraigFeldspar about the computations!

packages/dev/core/src/Shaders/pbr.fragment.fx
```diff
@@ -672,11 +672,7 @@ void main(void) {
     #endif
 
     #ifdef PREPASS_REFLECTIVITY
-        #if defined(REFLECTIVITY)
-            gl_FragData[PREPASS_REFLECTIVITY_INDEX] = vec4(baseReflectivity.rgb, baseReflectivity.a * writeGeometryInfo);
-        #else
-            gl_FragData[PREPASS_REFLECTIVITY_INDEX] = vec4(0.0, 0.0, 0.0, writeGeometryInfo);
-        #endif
+        gl_FragData[PREPASS_REFLECTIVITY_INDEX] = vec4(toGammaSpace(specularEnvironmentR0), microSurface) * writeGeometryInfo;
```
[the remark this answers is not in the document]
Mannns: Hello,
To me it doesn't break existing code since reflectivity was already returned in gamma space.
I also find strange that prePass returned gamma values, but if we change to linear this will affect SSR rendering..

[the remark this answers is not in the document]
sebavan: Agree mostly for back compat despite prefering a linear way as well.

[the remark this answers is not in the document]
Mannns: What do you think about:
- returning linear reflectivity in both prePass and geomBuffer
- applying toGammaSpace in the SSR postProcess for back compatibility ?

Popov72:
All good for me, thanks!

PR #12524, opened 2022-05-13
InputManager: Fix Picking on PointerUp and add bool to skip pointerup picking

PolygonalSun:
Recently on the forums ([forum post](https://forum.babylonjs.com/t/no-api-to-disable-mesh-picking-for-some-buttons-and-keep-onpointerobservable/29887)), there was a user who found an issue with the pointerUpPredicate not being used by the InputManager properly.  When a dummy predicate was used (`() => false`), there was still a pickedmesh/point in the pickInfo object.  This PR does a couple of things:
- Fix for the picking logic to use the `pointerUpPredicate` and actually call `scene.pick`
- Add a boolean to scene that allows the user to skip picking on pointerup (same as move and down)

### DIFF
--- a/packages/dev/core/src/Inputs/scene.inputManager.ts
+++ b/packages/dev/core/src/Inputs/scene.inputManager.ts
@@ -824,7 +824,6 @@ export class InputManager {
                     this._initActionManager(null, clickInfo);
                 }
 
-                //let pickResult;
                 if (scene.skipPointerUpPicking) {
                     pickResult = new PickingInfo();
                 } else {
--- a/packages/dev/core/src/Inputs/scene.inputManager.ts
+++ b/packages/dev/core/src/Inputs/scene.inputManager.ts
@@ -824,7 +824,7 @@ export class InputManager {
                     this._initActionManager(null, clickInfo);
                 }
 
-                if (scene.skipPointerUpPicking) {
+                if (!pickResult && scene.skipPointerUpPicking) {
                     pickResult = this._currentPickResult;
                 } else {
                     pickResult = scene.pick(this._unTranslatedPointerX, this._unTranslatedPointerY, scene.pointerUpPredicate, false, scene.cameraToUseForPointers);
--- a/packages/dev/core/src/Inputs/scene.inputManager.ts
+++ b/packages/dev/core/src/Inputs/scene.inputManager.ts
@@ -826,6 +826,9 @@ export class InputManager {
                 if (!this._meshPickProceed && ((AbstractActionManager && AbstractActionManager.HasTriggers) || scene.onPointerObservable.hasObservers())) {
                     this._initActionManager(null, clickInfo);
                 }
+                if (!pickResult) {
+                    pickResult = this._currentPickResult;
+                }
 
                 this._processPointerUp(pickResult, evt, clickInfo);
 
--- a/packages/dev/core/src/Inputs/scene.inputManager.ts
+++ b/packages/dev/core/src/Inputs/scene.inputManager.ts
@@ -824,9 +824,11 @@ export class InputManager {
                     this._initActionManager(null, clickInfo);
                 }
 
-                if (!pickResult && scene.skipPointerUpPicking) {
+                if (!pickResult) {
                     pickResult = this._currentPickResult;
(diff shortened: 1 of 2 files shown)


packages/dev/core/src/Inputs/scene.inputManager.ts
```diff
@@ -824,8 +823,12 @@ export class InputManager {
                 if (!this._meshPickProceed && ((AbstractActionManager && AbstractActionManager.HasTriggers) || scene.onPointerObservable.hasObservers())) {
                     this._initActionManager(null, clickInfo);
                 }
-                if (!pickResult) {
-                    pickResult = this._currentPickResult;
+
+                if (scene.skipPointerUpPicking) {
+                    pickResult = new PickingInfo();
```
RaananW, line 828: would it make sense to store a static version of an empty pickingInfo instead of creating an empty one every time? (might not make sense, it's more of a question :-))

RaananW (commented):
Just a small question, otherwise LGTM!

packages/dev/core/src/Inputs/scene.inputManager.ts
PolygonalSun: A static version might make more sense for skipping.  I'm not 100% if anything is changed within the pickResult object as it passes through `_processPointerUp` and `_processPointerDown` (specifically when `step.action` is called) so I'd have to investigate that first before implementing it.

packages/dev/core/src/Inputs/scene.inputManager.ts
```diff
@@ -824,8 +823,12 @@ export class InputManager {
                 if (!this._meshPickProceed && ((AbstractActionManager && AbstractActionManager.HasTriggers) || scene.onPointerObservable.hasObservers())) {
                     this._initActionManager(null, clickInfo);
                 }
-                if (!pickResult) {
-                    pickResult = this._currentPickResult;
+
+                if (scene.skipPointerUpPicking) {
+                    pickResult = new PickingInfo();
+                } else {
+                    pickResult = scene.pick(this._unTranslatedPointerX, this._unTranslatedPointerY, scene.pointerUpPredicate, false, scene.cameraToUseForPointers);
```
bghgary, line 830: After this change, does it mean the code will call scene.pick twice (once for pointer down and once for pointer up) by default?

PolygonalSun: Yes, unless any of the skip bools are set to be true (skipPointerUpPicking is now true by default)

sebavan (approved):

sebavan:
@PolygonalSun could you double check ? I can repro the picking being set on pointer up

PolygonalSun:
> Seems like pick is still occuring Though, I tested (I believe) [this commit](https://github.com/BabylonJS/Babylon.js/pull/12524/commits/ac5565448a6c7ada4543118429a1d6e6d3ae6ec0) and it worked Test pg: https://playground.babylonjs.com/?snapshot=refs/pull/12524/merge#CASE7T#3

Okay, so this pick isn't directly apart of the picking done by up, down, or move but is triggered as part of `_initActionManager` function that would pick a mesh and return an action manager for said mesh (if applicable).  I'm hesitant to change a lot of the logic here so what I'm going to do is to add an if statement to check if all three skip bools are true and if so, it will skip the pick entirely.

sebavan:
This is a different issue than the one you were addressing in here. I would suggest to make it another PR. I do not think the flag combination is the right approach for it. This would probably need a separate way to turn it off ? @deltakosh what do you think ?

PolygonalSun:
Current findings: All of the skips disable picking as they should and the pointerUpPredicate is now being used properly in _initActionManager.

One thing that I've noticed is that the code for `onPrePointerObservable` gets `notifyObservers` called **after** the pick has happened (for `pointerup` only) so changes to `pointerUpPredicate` won't be used if set in an observer for `onPrePointerObservable`.  Since there is a specific logic for when this is used, I'm not sure if I should change it.  What do you think @sebavan?  If we don't need to change it then the PR should have everything it needs.

deltakosh (approved):

PR #12525, opened 2022-05-13
Initial storybook setup

darraghjburke:
Initial PR to create a storybook with a few simple components. Mainly wanting to verify the architecture makes sense. Takes advantage of some nice features of CSS loader, like :local scopes rules and composition (importing classes from other CSS files).

Components right now:
1. Toggle
2. Label
3. Button

NOT IN THIS PR:
- Snapshots
- Jest testing


RaananW (commented):
This is great!!

The only thing the I want to comment on is the location of the components (that are also imported by other components). Storybook should just be where the previews are found. It can technically be outside of the `src` folder, but can stay there as well. I find `Button.tsx` to be misplaced Because the components should be imported by Storybook, but not be in the storybook directory.

RaananW (approved):
This looks great!!!

I am just thinking out loud here - it feels like those components could get some sort of a prefix so their names wouldn't be very abstract like Button or Icon. This can be done when importing as well, but maybe we can prefix them (like BJSButton or BabylonButton or something like that?)

carolhmj (approved):

RaananW:
> > This looks great!!!
> > I am just thinking out loud here - it feels like those components could get some sort of a prefix so their names wouldn't be very abstract like Button or Icon. This can be done when importing as well, but maybe we can prefix them (like BJSButton or BabylonButton or something like that?)
> 
> I spent some time thinking about this...my instinct is that it feels better to write JSX like:
> 
> ```js
> <Header>
> <Button>New</Button>
> <Button>Save</Button>
> </Header>
> ```
> 
> versus:
> 
> ```js
> <BabylonHeader>
> <BabylonButton>New</BabylonButton>
> <BabylonButton>Save</BabylonButton>
> </BabylonHeader>
> ```
> 
> Component libraries tend to name their components without prefixes, for ex MUI and Fluent UI: image image
> 
> However, it really comes down to personal preference. The benefit of a prefix is that the reader can instantly tell the component comes from our library. If we do go that route, though, I'd rather keep it as short as possible...maybe just "BButton" and "BToggle".

Nope, I have no preference. I trust you on this one. If anyone wants to prefix they can `import as`, which is perfectly fine :-)

packages/dev/sharedUiComponents/declaration.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.scss' {
```
RaananW, line 1: This file can go into the src directory, and then there is no need to add it to the "include" list

RaananW (approved):

PR #12526, opened 2022-05-13
Fix a bug in glTF loader when loading skins

bghgary:
See https://forum.babylonjs.com/t/assetcontainer-dispose-throws-typeerror-r-metadata-is-null/30360


deltakosh (approved):

RaananW:
Just wondering - why are we skipping changelog here?

PR #12529, opened 2022-05-16
Fix SubMesh IsGlobal

BlakeOne:
Currently IsGlobal returns true if the subMesh has all the vertices of the mesh but it should check if the subMesh also has all the indices of the mesh IMO.

Because otherwise the mesh's full bounding info is returned by getBoundingInfo when the subMesh has all the vertices **but only some of the indices** of the whole mesh. E.G. the subMeshes created by CreateTiledGround, which shouldn't return the bounding info for the whole mesh but currently are.

Forum discussion w/ test PGs: https://forum.babylonjs.com/t/find-or-define-center-point-of-submesh/30400/3

### DIFF
--- a/packages/dev/core/src/Meshes/subMesh.ts
+++ b/packages/dev/core/src/Meshes/subMesh.ts
@@ -235,7 +235,8 @@ export class SubMesh implements ICullable {
      */
     // eslint-disable-next-line @typescript-eslint/naming-convention
     public get IsGlobal(): boolean {
-        return this.verticesStart === 0 && this.verticesCount === this._mesh.getTotalVertices();
+        return this.verticesStart === 0 && this.verticesCount === this._mesh.getTotalVertices() &&
+            this.indexStart === 0 && this.indexCount === this._mesh.getTotalIndices();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


BlakeOne:
Should it all be on one long line instead? It used to say what the formatting problem/solution was but now I can't find that info anymore, for the last few PRs that had formatting errors...

Duckwhale:
> Should it all be on one long line instead? It used to say what the formatting problem/solution was but now I can't find that info anymore, for the last few PRs that had formatting errors...

You can run ``npm run format:fix`` to let [prettier](https://github.com/prettier/prettier) resolve the formatting issue :)

The error is also displayed in the [logs](https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=20247&view=logs&j=e7b80b24-15e8-561b-2b9b-3069817c0357&t=f86338ee-703d-5071-b860-e70612e5c921&l=27), though it could be difficult to find. Or you can see it by running ``npm run format:check``.

RaananW (approved):

sebavan (approved):

BlakeOne:
> The error is also displayed in the [logs](https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=20247&view=logs&j=e7b80b24-15e8-561b-2b9b-3069817c0357&t=f86338ee-703d-5071-b860-e70612e5c921&l=27), though it could be difficult to find.

I used to be able to find it in the logs, it would say what the particular formatting issue was, now all I can find out is that  there's a formatting error without any needed details, ever since the build/validation system changed recently I think.... But I'll try running the prettier locally to see if it fixes the issue.

PR #12535, opened 2022-05-16
Uv projection

waverider404:
added uv projection function using any chosen reference mesh


packages/dev/core/src/Misc/index.ts
```diff
@@ -7,6 +7,7 @@ export * from "./decorators";
 export * from "./deferred";
 export * from "./environmentTextureTools";
 export * from "./meshExploder";
+export * from "./uvprojector";
```
deltakosh, line 10: should be uvProjector :)

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
+import { VertexBuffer } from "../Buffers";
+import { Ray } from "../Culling/ray";
+import { Vector3 } from "../Maths";
+import { AbstractMesh } from "../Meshes/abstractMesh";
+import { CreateBox } from "../Meshes/Builders/boxBuilder";
+import { Mesh } from "../Meshes/mesh";
+import { Scene } from "../scene";
+import { FloatArray, Nullable } from "../types";
+
+export class Uvprojector {
```
deltakosh, line 10: class name should be UVProjector

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export class Uvprojector {
+    /**
+     * Project uv from a reference mesh
+     * @param target the mesh to project the uv to
+     * @param reference the reference mesh to project the uv from
+     * @param scalefact the bounding box scale for the reference mesh
+     * @param scene
+     */
+     public static ProjectUV(target: Mesh, reference: Nullable<Mesh>, scalefact: number, scene: Scene): void {
```
deltakosh, line 18: I think we can directly export the function (right @RaananW ? or is it ok like that?)

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+                return true;
+            }
+            return false;
+        }
+        
+        //use ray to project the uv data from the reference mesh to the target mesh
+        for(let i = 0; i < indices.length; i++){
+    
+            if(i % 3 == 0){
+                let rayDir = new Vector3(0, 1, 0);
```
deltakosh, line 84: this should be a static one and be reused to save GC

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+    
+                let faceDir = getFaceNormal(vec[indices[i]], vec[indices[i + 1]], vec[indices[i + 2]]);
```
deltakosh, line 86: same ;)

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+export class Uvprojector {
+    /**
+     * Project uv from a reference mesh
+     * @param target the mesh to project the uv to
+     * @param reference the reference mesh to project the uv from
+     * @param scalefact the bounding box scale for the reference mesh
+     * @param scene
+     */
+     public static ProjectUV(target: Mesh, reference: Nullable<Mesh>, scalefact: number, scene: Scene): void {
+        target.bakeCurrentTransformIntoVertices();
```
deltakosh, line 19: This is something that the user may not want. We should get the data as is and then transform the vector3 using the mesh data so we do not change the target mesh

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+            reference.scaling.set(_width, _height, _depth);
+            reference.position.copyFrom(boundingBox.center);
+            reference.bakeCurrentTransformIntoVertices();
+            reference.visibility = 0.3;
+        }
+
+        function getFaceNormal(pt1: any, pt2: any, pt3: any): any{ 
+            let convertToArray = false;
+            if(pt1 instanceof Array || pt1 instanceof Float32Array){
+                pt1 = new Vector3().fromArray(pt1, 0);
```
deltakosh, line 56: we need to save GC and use temp vectors

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+                pt2 = new Vector3().fromArray(pt2, 0);
+                pt3 = new Vector3().fromArray(pt3, 0);
+                convertToArray = true;
+            }    
+            const normal = Vector3.Zero();
```
deltakosh, line 61: same here regarding GC

packages/dev/core/src/Misc/uvprojector.ts
sebavan: and filename uvProjector

packages/dev/core/src/Misc/uvprojector.ts
RaananW: We should slowly stop using static members on classes and move to exporting from the module itself. We are still not doing it though, and it'll be globally done as party of the esm package work. So far now I'm fine with both ways of dealing with that. 
For the Babylon namespace, it's better to have it public static to not pollute the global namespace.

RaananW (commented):
Looks great!

As the others mentioned it will be better to reuse as many vectors as possible. Every call to new (also in methods like subtract and add) generates an object that will be garbage collected on each frame. We avoid that by reusing vectors, usually from the temp vectors array

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+                }
+            }
+        };
+        target.setVerticesData(VertexBuffer.UVKind, uvs);
+    }
+    /**
+     * Get class name
+     * @returns "UvProjector"
+     */
+     public getClassName(): string {
```
Popov72, line 133: I don't think we need this method as the function is a static function.

If we keep it, it should return `UVProject`.

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+                    rayDir = new Vector3(0, 1, 0);
+                }else if(faceDir.z > faceDir.x && faceDir.z > faceDir.y){
+                    rayDir = new Vector3(0, 0, 1);
+                }
+    
+                rayDir = rayDir.normalize();
+                const rayLength = rayDir.length() * Math.abs(_width * _height * _depth);
+    
+                const faceVerts = [vec[indices[i]], vec[indices[i + 1]], vec[indices[i + 2]]];
+                const trigInd = [indices[i], indices[i + 1], indices[i + 2]]
```
Popov72, line 102: Don't create this temporary array and simply use `indices[i + n]` in the loop below, to avoid GC pressure.

packages/dev/core/src/Misc/uvprojector.ts
```diff
@@ -0,0 +1,136 @@
...
+                
+                for(let n = 0; n < faceVerts.length; n++){
+                    
+                    let id = trigInd[n] || 0;
+                    id *= 2;
+    
+                    const origin = new Vector3(faceVerts[n].x, faceVerts[n].y, faceVerts[n].z);
+                    let ray =  new Ray(origin, rayDir, rayLength);
```
Popov72, line 110: You should create the ray a single time outside the loops and only update its `origin` / `direction` properties to avoid GC pressure.

Popov72 (changes requested):

deltakosh:
up? :)

deltakosh:
@waverider404 hey buddy, any update?

waverider404:
@deltakosh 
I wasn't really feeling fine all this time. Just recovering, I will try to get that in by the end of this week

sebavan:
@waverider404 wondering if you had any progress on this one or if we should move it to draft for the time being ?

waverider404:
@sebavan 

Hello, yes can you please move it to draft. I will try to get this of the plate this week. It's been too long

deltakosh:
Hey @waverider404. In order to not add more pressure on you or on our resolution time I will close it for now so you can work on it when you see fit ;)

PR #12537, opened 2022-05-16
[NME] fix wheel behaviour and overscroll

### LINKED ISSUE #12536: [NME] Pinch on trackpad breaks layout
MacBook or any laptop with trackpad...

ezgif-2-d49c4d843b

3GOMESz:
- fix #12536
- disable horizontal scroll (handle return back/next page)
- fix page overscroll

<br/>

B E F O R E
ezgif-2-d49c4d843b

A F T E R
ezgif-2-6eac051240

### DIFF
--- a/packages/tools/nodeEditor/public/index.html
+++ b/packages/tools/nodeEditor/public/index.html
@@ -33,27 +33,6 @@
             }
         </style>
 
-        <script>
-            window.addEventListener(
-            "wheel",
-            function (evt) {
-                if (evt.ctrlKey) {
-                    return evt.preventDefault();
-                }
-
-                if (Math.abs(evt.deltaX) < Math.abs(evt.deltaY)) {
-                    return;
-                }
-
-                const scrollLeftMax = this.scrollWidth - this.offsetWidth;
-                if (this.scrollLeft + evt.deltaX < 0 || this.scrollLeft + evt.deltaX > scrollLeftMax) {
-                    return evt.preventDefault();
-                }
-            },
-            { passive: false }
-        );
-        </script>
-
     </head>
 
     <body>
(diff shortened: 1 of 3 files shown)


sebavan:
ohhhhh thanks a lot for the PR looks good to me but will add ppl who know way more than myself about the nme editor :-)

3GOMESz:
Apparently it's not working at snapshot, I recommend testing locally

RaananW (changes requested):
Thanks! Certainly a needed change.

I'm not sure I'm happy about a global script tag in the html page. This can be done in one of the components as part of the react application. 
TBH I'm not too sure how many of those prevent default calls are needed, but I'm sure you have tested that. I haven't tested the pr yet, this comment is only about the code location.

carolhmj (approved):
Tested and seems to be working nicely, just +1 on Raanan that it would be better to have the event listener in a component instead of in a global script. graphEditor.tsx would be a good place.

3GOMESz:
The idea of the script on index.html was meant to overwrite the browser's default behavior as soon as the page is loaded, before the bundle is even loaded or any components are mounted. If the user zooms in before the page loads the whole JS bundle he will get stuck, the initial idea was to avoid it

3GOMESz:
What is the best choice? Move the overwride to the `graphEditor.tsx` by adding and removing an eventListener in the respective `componentDidMount` and `componentWillUnmount` or leave it in the index for a behavior applied in the browser before anything else?

RaananW:
I would still prefer keeping implementation outside of index.html. There are many different reasons for that. The main one is the separation done between the static resources and the compiled resources.

RaananW (approved):

PR #12551, opened 2022-05-20
Add warning to Bounding Box Gizmo

carolhmj:
Related issue: https://forum.babylonjs.com/t/is-that-the-attachednode-setter-wrong/30473/6

### DIFF
--- a/packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
+++ b/packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
@@ -534,6 +534,11 @@ export class BoundingBoxGizmo extends Gizmo {
                 this._updateDummy();
             });
         }
+        // If the attachedMesh property is not set, then the attachedNode constructor was used, which
+        // the BoundingBoxGizmo doesn't support
+        if (!this.attachedMesh) {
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
```diff
@@ -534,6 +534,11 @@ export class BoundingBoxGizmo extends Gizmo {
                 this._updateDummy();
             });
         }
+        // If the attachedMesh property is not set, then the attachedNode constructor was used, which
+        // the BoundingBoxGizmo doesn't support
+        if (!this.attachedMesh) {
+            Logger.Warn("Using the attachedNode attribute in BoundingBoxGizmo may lead to unexpected behavior.");
```
sebavan, line 540: We should probably highlight that it is not supported instead ?

sebavan (approved):

packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
deltakosh: agreed

carolhmj: Hi, I updated the warning :D

PR #12553, opened 2022-05-20
Add sRGB option to 2D RawTexture

kircher1:


sebavan (approved):
This one feels like plumbing :-) thanks a lot @kircher1

PR #12554, opened 2022-05-20
fix default material to use vertex alpha for thin instances

BlakeOne:
Fix default material to use vertex alpha for thin instances

Forum issue: https://forum.babylonjs.com/t/different-transparency-while-using-thin-instances/30574


BlakeOne:
Not quite right, will resubmit. :)

PR #12560, opened 2022-05-22
Fix warning when cloning with instanced mesh

bghgary:


packages/dev/core/src/Meshes/instancedMesh.ts
```diff
@@ -498,6 +498,7 @@ export class InstancedMesh extends AbstractMesh {
                 "behaviors",
                 "worldMatrixFromCache",
                 "hasThinInstances",
+                "hasBoundingInfo"
```
RaananW, line 501:
```suggestion
                "hasBoundingInfo",
```

RaananW (approved):

PR #12563, opened 2022-05-23
Removed IE support. It was long overdue ;)

deltakosh:


RaananW (approved):
Huzzah!

PolygonalSun (commented):
So if we remove IE11 support, does this mean that we can remove all of the old code that we kept in for compatibility purposes and possibly use newer data structures?

packages/tools/viewer/src/index.ts
```diff
@@ -28,8 +28,6 @@ import "pepjs";
 
 import { initListeners, InitTags } from "./initializer";
 
-// promise polyfill, if needed!
-BABYLON.PromisePolyfill.Apply();
 initListeners();
 
 //deprectaed, here for backwards compatibility
```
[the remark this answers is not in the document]
deltakosh: What do you mean?

deltakosh:
@PolygonalSun: Correct!

RaananW:
This PR just gets better and better :-)

packages/dev/core/src/Rendering/edgesRenderer.ts
```diff
@@ -676,7 +675,7 @@ export class EdgesRenderer implements IEdgesRenderer {
             for (let t = 0; t < mustTesselate.length; ++t) {
                 const triangle = mustTesselate[t];
 
-                this._tessellateTriangle(triangle.edgesPoints, triangle.index, indices, remapVertexIndices);
+                this._tessellateTriangle(triangle.edgesPoints, triangle.index, indices as number[], remapVertexIndices);
```
bghgary, line 678: You don't need this change anymore.

deltakosh: good catch

packages/dev/core/src/Misc/tools.ts
```diff
@@ -167,29 +165,6 @@ export class Tools {
...
     }
 
-    /**
-     * Provides a slice function that will work even on IE
-     * @param data defines the array to slice
-     * @param start defines the start of the data (optional)
-     * @param end defines the end of the data (optional)
-     * @returns the new sliced array
-     */
-    public static Slice<T>(data: T, start?: number, end?: number): T {
```
PolygonalSun, line 177: If we remove these from Tools, this is also a breaking change.  Are we okay with this?

deltakosh: Yes this is ok as there is a native solution for it

bghgary: What if there is a playground that uses these functions that we are removing? Wouldn't the playground be broken?

bghgary: I only found one instance on the Playground that uses Slice or SliceToArray: https://doc.babylonjs.com/playground?q=Tools.Slice&type=code

bghgary: StringTools is more common though :-/

https://doc.babylonjs.com/playground?q=StringTools&type=code

RaananW: Mark it as deprecated, I will move it to the LTS package as part of the esm work

PR #12564, opened 2022-05-23
Fix ThinEngine.updateDynamicIndexBuffer

barroij:
This is to fix a bug we ran into.

To repro :
- at creation time, provide a mesh with indices of type `Array<number>`  => if all the indies are below 65535, the `Array<number>` is normallized into an `UInt16Array` in `ThinEngine._normalizeIndexData`

- call `Mesh.updateIndices` with a `UInt32Array` => instead of converting to `UInt16Array` as if a `Array<number>` had been passed, the `UInt32Array` is forwarded to WebGL, leading to wrong indexing.

### DIFF
--- a/packages/dev/core/src/Engines/Extensions/engine.dynamicBuffer.ts
+++ b/packages/dev/core/src/Engines/Extensions/engine.dynamicBuffer.ts
@@ -32,11 +32,11 @@ ThinEngine.prototype.updateDynamicIndexBuffer = function (this: ThinEngine, inde
 
     if (indexBuffer.is32Bits) {
         // anything else than Uint32Array needs to be converted to Uint32Array
-        arrayBuffer = indices instanceof Uint32Array ? indices : Uint32Array.from(indices)
+        arrayBuffer = indices instanceof Uint32Array ? indices : new Uint32Array(indices)
     }
     else {
         // anything else than Uint16Array needs to be converted to Uint16Array
-        arrayBuffer = indices instanceof Uint16Array ? indices : Uint16Array.from(indices)
+        arrayBuffer = indices instanceof Uint16Array ? indices : new Uint16Array(indices)
     }
 
     this._gl.bufferData(this._gl.ELEMENT_ARRAY_BUFFER, arrayBuffer, this._gl.DYNAMIC_DRAW);
--- a/packages/dev/core/src/Engines/Extensions/engine.dynamicBuffer.ts
+++ b/packages/dev/core/src/Engines/Extensions/engine.dynamicBuffer.ts
@@ -28,18 +28,18 @@ ThinEngine.prototype.updateDynamicIndexBuffer = function (this: ThinEngine, inde
     // Force cache update
     this._currentBoundBuffer[this._gl.ELEMENT_ARRAY_BUFFER] = null;
     this.bindIndexBuffer(indexBuffer);
-    let arrayBuffer;
 
+    let view: ArrayBufferView;
     if (indexBuffer.is32Bits) {
         // anything else than Uint32Array needs to be converted to Uint32Array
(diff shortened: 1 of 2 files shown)


sebavan (approved):

packages/dev/core/src/Engines/Extensions/engine.dynamicBuffer.ts
```diff
@@ -30,10 +30,13 @@ ThinEngine.prototype.updateDynamicIndexBuffer = function (this: ThinEngine, inde
...
-        arrayBuffer = indices;
-    } else {
-        arrayBuffer = indexBuffer.is32Bits ? new Uint32Array(indices) : new Uint16Array(indices);
+    if (indexBuffer.is32Bits) {
+        // anything else than Uint32Array needs to be converted to Uint32Array
+        arrayBuffer = indices instanceof Uint32Array ? indices : Uint32Array.from(indices)
+    }
+    else {
+        // anything else than Uint16Array needs to be converted to Uint16Array
+        arrayBuffer = indices instanceof Uint16Array ? indices : Uint16Array.from(indices)
```
barroij, line 39: And by the way, IMHO, thing are easier to understand if written this way.

Another example is in `WebGPUEngine.updateDynamicIndexBuffer` :
```
let view: ArrayBufferView;
if (indices instanceof Uint16Array) {
    if (indexBuffer.is32Bits) {
        view = Uint32Array.from(indices);
    } else {
        view = indices;
    }
} else if (indices instanceof Uint32Array) {
    if (indexBuffer.is32Bits) {
        view = indices;
    } else {
        view = Uint16Array.from(indices);
    }
} else {
    if (indexBuffer.is32Bits) {
        view = new Uint32Array(indices);
    } else {
        view = new Uint16Array(indices);
    }
}
```

might be clearer written this way as well:
```
    let view: ArrayBufferView;
    if (indexBuffer.is32Bits) {
        // anything else than Uint32Array needs to be converted to Uint32Array
        view = indices instanceof Uint32Array ? indices : Uint32Array.from(indices)
    }
    else {
        // anything else than Uint16Array needs to be converted to Uint16Array
        view = indices instanceof Uint16Array ? indices : Uint16Array.from(indices)
    }
```

barroij:
and here is a very simple playground with the bug : https://playground.babylonjs.com/#VKBJN#2737

barroij:
@sebavan @deltakosh I thought I made a comment on my changes but I cannot see it up there. Did I forget to click the "Comment" button ? :expressionless:

I changes the codes so that the same code is used in `WebGPUEngine.updateDynamicIndexBuffer` and in `ThinEngine.updateDynamicIndexBuffer`.
I could have created a function to share the code but I do not know what's the best place for that function.

Also There might be an issue in `NativeEngine.updateDynamicIndexBuffer` which uses `_normalizeIndexData` to convert the input IndicesArray into `Uint16Array` or `Uint32Array`. This function does not consider the type of  the pre-existing IndexBuffer and then could lead to the same kind of error.

As a side note, concerning this function `_normalizeIndexData`, ot gpes likes this :
```
if (this._caps.uintIndices) {
    if (indices instanceof Uint32Array) {
        return indices;
    } else {
        // number[] or Int32Array, check if 32 bit is necessary
        for (let index = 0; index < indices.length; index++) {
            if (indices[index] >= 65535) {
                return new Uint32Array(indices);
            }
        }

        return new Uint16Array(indices);
    }
}
```

Which does not look like a very good idea too me in case of updatable IndexBuffer, even if it should work in most cases. 
Determining the type on the IndexBuffer base on the content the IndicesArray passed at creation time is a bit dangerous, as there is no guarantee that this initial IndicesArray uses all the possible indices that could be used in the future using `updateIndices`.

deltakosh:
all good for me besides the formatting (just run 'npm run format:fix')

barroij:
formatted !

PR #12567, opened 2022-05-24
Add sRGB handling for ETC texture formats

kircher1:
Because ETC2 is an extension of ETC1, we should also be able to use the ETC2 sRGB format when sRGB is requested for an ETC1 texture.


packages/dev/core/src/Engines/thinEngine.ts
```diff
@@ -4481,6 +4485,22 @@ export class ThinEngine {
...
+                        texture._useSRGBBuffer = false;
+                    }
+                    break;
+                case Constants.TEXTUREFORMAT_COMPRESSED_RGBA8_ETC2_EAC:
+                    if (this._caps.etc2) {
+                        internalFormat = gl.COMPRESSED_SRGB8_ALPHA8_ETC2_EAC;
+                    } else {
+                        texture._useSRGBBuffer = false;
+                    }
+                    break;
```
Popov72, line 4503: You will need to also update WebGPU, in `WebGPUTextureHelper.GetWebGPUTextureFormat`:

https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Engines/WebGPU/webgpuTextureHelper.ts#L707-L723

sebavan (approved):
just a tiny formatting issue and good to go

PR #12574, opened 2022-05-24
Add default formatter as prettier

bghgary:


deltakosh (approved):

sebavan:
Would that not crash if you do not have it installed ? might need to drop a line in the readme in this case ?

sebavan (approved):

bghgary:
> Would that not crash if you do not have it installed ? might need to drop a line in the readme in this case ?

Nope. It just selects None again if Prettier is not installed. Prettier is already in [the recommended list](https://github.com/BabylonJS/Babylon.js/blob/master/.vscode/extensions.json#L5).

PR #12580, opened 2022-05-25
[Build] move babylon dependencies to peer dependencies

RaananW:


RaananW:
Was resolved in a different PR

PR #12586, opened 2022-05-25
DeviceInputSystem: Add check for matchMedia in WebDeviceInputSystem

PolygonalSun:
A user in the forum recently found an issue where the error `matchMedia is not defined` was occurring in an offscreen canvas.  Because the offscreen canvas was in a worker thread, it had no context of the `Window` interface and the `matchMedia` function.  This PR will add a check in the `WebDeviceInputSystem` to verify the existence of this function before attempting to execute it.

Forum Link: https://forum.babylonjs.com/t/offscreen-canvas-matchmedia/30682/10

### DIFF
--- a/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
+++ b/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
@@ -231,7 +231,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
         }
 
         // If the device in use has mouse capabilities, pre-register mouse
-        if (globalThis.matchMedia && matchMedia("(pointer:fine)").matches) {
+        if (typeof matchMedia === "function" && matchMedia("(pointer:fine)").matches) {
             // This will provide a dummy value for the cursor position and is expected to be overridden when the first mouse event happens.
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -231,7 +231,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
         }
 
         // If the device in use has mouse capabilities, pre-register mouse
-        if (matchMedia("(pointer:fine)").matches) {
+        if (globalThis.matchMedia && matchMedia("(pointer:fine)").matches) {
```
bghgary, line 234: `globalThis` is not supported in IE. Either we wait until David's PR is merged or use something different.

PolygonalSun: I just checked `typeof matchMedia === "function"` and it does work in a worker thread.  When I initially tried just using just `matchMedia`, an Uncaught ReferenceError was thrown, which is why I used `globalThis`.  I have no problem just changing this to use typeof since it works in the same way.

deltakosh (approved):

PR #12589, opened 2022-05-26
WebGPU: fix typescript

Popov72:
See https://forum.babylonjs.com/t/gpuoutofmemoryerror-and-gpuvalidationerror-error-with-latest-5-8-0-release/30755


sebavan (approved):

packages/dev/core/src/LibDeclarations/webgpu.d.ts
```diff
@@ -977,13 +977,13 @@ declare class GPUError {
     readonly message: string;
 }
 
-declare class GPUOutOfMemoryError implements GPUError {
+declare class GPUOutOfMemoryError extends GPUError {
     private __brand: void;
```
sebavan, line 981: should nt we remove the _brand override of a private in case of extend ? similar to the class below ?

deltakosh (approved):

PR #12590, opened 2022-05-26
Move frame bar on ACE to the top of the window

carolhmj:
Previous:
image
With the PR:
image

### DIFF
--- a/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/playHeadComponent.tsx
+++ b/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/playHeadComponent.tsx
@@ -20,6 +20,8 @@ export class PlayHeadComponent extends React.Component<IPlayHeadComponentProps, 
     private _onActiveAnimationChangedObserver: Nullable<Observer<IActiveAnimationChangedOptions>>;
     private _viewScale = 1;
     private _offsetX = 0;
+    private _offsetRange = 0;
+    private _isMounted = false;
 
     private _pointerIsDown: boolean;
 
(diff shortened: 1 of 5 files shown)


packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/playHeadComponent.tsx
```diff
@@ -55,25 +72,41 @@ export class PlayHeadComponent extends React.Component<IPlayHeadComponentProps,
...
+            if (!this._isMounted) {
+                return;
+            }
+
             this._viewScale = 1 / scale;
             this.forceUpdate();
 
             this._moveHead(this.props.context.activeFrame);
         });
     }
```
RaananW, line 104: I understand this is not entirely the meaning of this PR, but if you need to check if the component is mounted, it is registered in the wrong place. Instead of tracking whether or not the component is mounted, add those observers in the  componentDidMount function (and removed when it is unmounted).

RaananW (commented):
My only comment is asking about the `isMounted` flag, which is usually not needed. You can register observers once the component mounted, and remove them when unmounted. This way there is no need to check for the component's state.

It would also seem (unrelated to that PR!) that some observers are not cleared (for example onGraphMoved, onGraphScaled and so on) so removing them when the component isn't mounted anymore is anyhow needed.

packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/playHeadComponent.tsx
carolhmj: No problem, I think isMounted was there just to avoid calling the observer when it wasn't mounted, so I'm just adding/removing them now :)

RaananW (approved):

PR #12595, opened 2022-05-27
Fix `AssetManager` `addTextureTask` resolving with `task.texture === undefined` when using `NullEngine`

alvov-evo:
This PR is a follow up on https://forum.babylonjs.com/t/using-assetsmanager-with-nullengine-to-load-textures/8758


packages/dev/core/src/Engines/nullEngine.ts
```diff
@@ -686,7 +686,9 @@ export class NullEngine extends Engine {
         texture.isReady = true;
 
         if (onLoad) {
-            onLoad();
+            setTimeout(() => {
```
RaananW, line 689: i would rather not use setTimeout. I assume this is the main fix for that case, but setTimeout will guaranty that this function will not run inside the render loop.
As the scene is a variable in this function, you could use `scene.onBeforeRenderObservable.addOnce` to add the onLoad callback

RaananW: ThinEngine.QueueNewFrame will still keep the onLoad outside of the render loop (and might use setTimeout if no requestAnimationFrame function doesn't exist in the context running the scene)

I haven't debugged the issue, and you can be sure I believe you it exists :-) I am just not sure in what cases exactly does it appear, and whether or not we can assume that a scene will always be present.

It also seems to me like the function is lacking the onLoadObservable notification. So whereas this solves the issue for this specific topic, I think it might be a little deeper than that. 

BUT!!! This is a fix. @deltakosh - You know the null engine better. If you want to resolve this and merge the fix just approve this and i'll do the rest :-)

deltakosh: I'm fine with the settimeout (until I'm not :))

Let's merge it and see if problems are raised(but I'm pretty confident we should be good)

deltakosh (approved):

packages/dev/core/src/Engines/nullEngine.ts
alvov-evo: @RaananW @deltakosh many thanks 🙇 

Regarding the case where this issue appeared for me - it was when i was writing tests for some function(s) i have that use `AssetsManager` internally to load textures. I'm using `NullEngine` in tests but on texture creation it calls the `onLoad` callback before actually returning the newly created texture (because `NullEngine` doesn't actually load anything), which leads to `AssetManager` task resolving with `undefined`.

The workaround i've been using so far was mocking `AssetManager` methods, but i've also promised to create a PR to fix this two years ago and it kept buzzing me since then, so here we are...

PR #12596, opened 2022-05-27
remove private variables from WebGPU declaration

RaananW:
The reason behind this is the documentation page that fails compilation because of this error:

```
 Class 'GPUOutOfMemoryError' incorrectly implements class 'GPUError'. Did you mean to extend 'GPUError' and inherit its members as a subclass?
  Types have separate declarations of a private property '__brand'.
Error: C:/Users/raweber/Documents/GitHub/Documentation/.temp/docdirectory/doc.d.ts(97057)
 Class 'GPUValidationError' incorrectly implements class 'GPUError'. Did you mean to extend 'GPUError' and inherit its members as a subclass?
  Types have separate declarations of a private property '__brand'.
```

The reason for the error is probably an older version of typedoc (work is being done to update that), which uses an older version of typescript to analyze the file.

As this private variable is not being used anywhere in the code I took delibery of suggesting to remove it, but I will wait for @Popov72 to approve this PR before merging

### DIFF
--- a/packages/dev/core/src/LibDeclarations/webgpu.d.ts
+++ b/packages/dev/core/src/LibDeclarations/webgpu.d.ts
@@ -49,7 +49,6 @@ interface WorkerNavigator {
 }
 
 declare class GPU {
-    private __brand: void;
     requestAdapter(options?: GPURequestAdapterOptions): Promise<GPUAdapter | null>;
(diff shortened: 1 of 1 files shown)


bghgary:
Is there a workaround for this?

PR #12607, opened 2022-05-30
Grid children disappear once linked mesh is outside view

### LINKED ISSUE #12602: Grid children disappear once linked mesh is outside view
Related forum issue: https://forum.babylonjs.com/t/tooltip-losing-content/30775

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com): https://www.babylonjs-playground.com/#XCPP9Y#13148
When the camera is moved so that the sphere is out of view, then moved back and the sphere is hovered, the contents of the tooltip aren't shown. Setting `clipChildren` as false on the parent control makes it not happen anymore.

**Screenshots**
Before moving camera:
image
After moving camera:
image

RaananW:
Fixes #12602
This forces resetting the clipping state of children when their matrix is set as dirty.

### DIFF
--- a/packages/dev/gui/src/2D/controls/container.ts
+++ b/packages/dev/gui/src/2D/controls/container.ts
@@ -142,6 +142,7 @@ export class Container extends Control {
 
     public _flagDescendantsAsMatrixDirty(): void {
         for (const child of this.children) {
+            child._isClipped = false;
             child._markMatrixAsDirty();
         }
     }


packages/dev/gui/src/2D/controls/container.ts
```diff
@@ -142,6 +142,7 @@ export class Container extends Control {
 
     public _flagDescendantsAsMatrixDirty(): void {
         for (const child of this.children) {
+            child._isClipped = false;
             child._markMatrixAsDirty();
```
sebavan, line 146: Shouldn t the marMatrixAsDirty set the flag ??? just wondering :-)

RaananW: yeah, this is a very special case! I was struggeling with this decision.
This special case appears ONLY when a container has a container as a child. Setting isClipped every time the matrix is set dirty will be redundant. This is the least amount of isClipped-resetting.
But I agree with you that it does feel a bit off.

sebavan: maybe let s wait for @deltakosh feedback tomorrow as I remember the caching was pretty heavy to put in place :-)

RaananW: I am anyhow not merging this until approved :-)

PR #12609, opened 2022-05-30
Fix registerInstancedBuffer after Mesh creation

sebavan:
https://forum.babylonjs.com/t/disappearing-mesh-material-on-registerinstancedbuffer-and-instancedbuffers/30868


packages/dev/core/src/Lights/shadowLight.ts
```diff
@@ -387,4 +387,11 @@ export abstract class ShadowLight extends Light implements IShadowLight {
         }
         return this;
     }
+
+    /** @hidden */
+    protected _syncParentEnabledState() {
+        super._syncParentEnabledState();
+        (this.transformedPosition as any) = null;
+        (this.transformedDirection as any) = null;
```
Popov72, line 395: I think you should set these properties only if `this.parent` is `null`, or maybe, alternatively, keep those lines and call `computeTransformedInformation()` to update the `transformedPosition` / `transformedDirection` properties (if necessary)?

PR #12617, opened 2022-05-31
Fix JPEG environment texture

sebavan:


packages/dev/core/src/Materials/Textures/cubeTexture.ts
```diff
@@ -91,8 +91,9 @@ export class CubeTexture extends BaseTexture {
 
     private _noMipmap: boolean;
 
+    /** @hidden */
     @serialize("files")
-    private _files: Nullable<string[]> = null;
+    public _files: Nullable<string[]> = null;
```
apetkov-so, line 96: If it's now public, shouldn't we drop the underscore? Also, why @hidden and public?

sebavan: This is to be used only internally not meant to be back compat outside of the framework

PR #12622, opened 2022-06-01
Disallow users to change the frame of the first key, ensuring that th…

carolhmj:
…ere is always a frame at 0. It was already not possible to drag the first frame outside of 0, but this covers the top bar input.

### DIFF
--- a/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/rangeFrameBarComponent.tsx
+++ b/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/rangeFrameBarComponent.tsx
@@ -128,7 +128,7 @@ export class RangeFrameBarComponent extends React.Component<IRangeFrameBarCompon
     }
 
     private _buildActiveFrame() {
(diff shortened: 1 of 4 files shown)


packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/animations/curveEditor/graph/rangeFrameBarComponent.tsx
```diff
@@ -128,7 +128,7 @@ export class RangeFrameBarComponent extends React.Component<IRangeFrameBarCompon
     }
 
     private _buildActiveFrame() {
-        if (this.props.context.activeFrame !== null && this.props.context.activeFrame !== undefined) {
```
RaananW, line 131: is that a long standing bug? it's completely flipped

carolhmj: It was a bug from my previous ACE PR that I noticed and fixed now

deltakosh (approved):

PR #12628, opened 2022-06-02
Add support for draco compression to Core

pandaGaume:
This PR add support for draco compression to the Core lib.
For the purpose, we divided the DracoCompression class into an Encoder and a Decoder, keeping the old API with a deprecated flag.
We now have a 

- DracoCommons file which encompass all the standard declarations and interfaces
- DracoEncoder for encoding purpose
- DracoDecoder for decoding purpose.
- DracoCompression has been kept for backward compatibility, but greatly reduced as kind of decorator.

**NO** regression has been observed using GLTF Loader with [KHR_draco_mesh_compression](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md)
Test has been Also conducted by compressing and decompressing mesh on the fly using 

```
  // encode the mesh
  var encoder = new BABYLON.DracoEncoder();
  return encoder.encodeMeshAsync(mesh).then((encoded)=> {
          // once done decode the mesh
          var decoder = new BABYLON.DracoDecoder();
          return decoder
              .decodeMeshAsync(encoded.data, encoded.attributes)
              .then((babylonVertexData) => {
                  // and apply the geometry to new mesh
                  const name = mesh.Name + "Dracoized"
                  const babylonMesh = new BABYLON.Mesh(name, scene);
                  babylonVertexData.applyToMesh(babylonMesh)
                  // finally return the mesh
                  return babylonMesh;
              })
              .catch((error) => {
                  throw new Error(`${error.message}`);
              });
```
from [playground ](https://playground.babylonjs.com/#NX5U65#2)
_Note the above playround will not be functional until the PR will be merge._
Test has been conducted changing the Draco wasm configuration to local

```
    let baseEncoderUrl = "http://localhost:1337/";
    
    const codecConfiguration  =  {
        decoder: {
            wasmUrl: baseEncoderUrl + "draco_wasm_wrapper_gltf.js",
            wasmBinaryUrl: baseEncoderUrl + "draco_decoder_gltf.wasm",
            fallbackUrl: baseEncoderUrl + "draco_decoder_gltf.js",
        },
        encoder: {
            wasmUrl: baseEncoderUrl + "draco_encoder_wrapper.js",
            wasmBinaryUrl: baseEncoderUrl + "draco_encoder.wasm",
            fallbackUrl: baseEncoderUrl + "draco_encoder.js",
        },

    };
    BABYLON.DracoEncoder.Configuration = codecConfiguration.encoder;
    BABYLON.DracoDecoder.Configuration = codecConfiguration.decoder;
```


packages/dev/core/src/Meshes/Compression/Draco/dracoCommons.ts
```diff
@@ -0,0 +1,220 @@
...
+    public static WasmBaseUrl: string = "https://preview.babylonjs.com/";
+
+    protected static _isCodecAvailable(codec: IDracoCompressionEngineConfiguration) {
+        return !!((codec.wasmUrl && codec.wasmBinaryUrl && typeof WebAssembly === "object") || codec.fallbackUrl);
+    }
+
+    /**
+     * Default number of workers to create when creating the draco compression object.
+     */
+    public static DefaultNumWorkers = DracoCompressionBase._GetDefaultNumWorkers();
```
bghgary, line 145: This should be a public property on the decoder and encoder. I don't think we want users to touch the base class to change this value.

packages/dev/core/src/Meshes/Compression/Draco/dracoCommons.ts
```diff
@@ -0,0 +1,220 @@
+import type { VertexData, IGetSetVerticesData } from "../../mesh.vertexData";
```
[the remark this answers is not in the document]
bghgary: Or maybe change all the files to:
- `dracoCommon.ts`
- `dracoDecoder.ts`
- `dracoEncoder.ts`
- `dracoCompression.ts` (for backcompat)

packages/dev/core/src/Meshes/Compression/Draco/dracoCompression.encoder.ts
```diff
@@ -0,0 +1,338 @@
...
+    encoderModule: any,
+    attributes: Array<{ key: string; type: number; stride: number; data: Float32Array }>,
+    indices: Nullable<Uint16Array | Uint32Array>,
+    geometryDestination: 0 | 1,
+    options: any,
+    destination?: ArrayBuffer
+): Nullable<{ data: Uint8Array; attributes: { [kind: string]: number } }> {
+    const vertices = attributes.find((a) => a.key == "position")?.data;
+
+    if (vertices?.length) {
```
bghgary, line 35: Should this be a failure? Returning null to the user when there are no vertices is more complicated since the contract now needs to be nullable. Whoever is calling `encodeMeshAsync` on the encoder can easily check to see if the mesh being encoded has vertices.

packages/dev/core/src/Meshes/Compression/dracoCompression.ts
```diff
@@ -1,468 +1,34 @@
...
-        }
-
-        // Use 50% of the available logical processors but capped at 4.
-        return Math.min(Math.floor(navigator.hardwareConcurrency * 0.5), 4);
+    public static set Configuration(value: any) {
+        DracoDecoder.Configuration = value.decoder;
     }
 
-    private static _Default: Nullable<DracoCompression> = null;
+    private static _Default0: Nullable<DracoDecoder> = null;
```
bghgary, line 18: Why `_Default0`? What's the `0` mean?

packages/dev/core/src/Meshes/Compression/dracoCompression.ts
```diff
@@ -1,468 +1,34 @@
...
-        if (this._workerPoolPromise) {
-            return this._workerPoolPromise.then(() => {});
-        }
-
-        if (this._decoderModulePromise) {
-            return this._decoderModulePromise.then(() => {});
-        }
-
-        return Promise.resolve();
-    }
```
bghgary, line 389: This is also a breaking change. Also, I didn't see this method in the encoder/decoder. Assuming we add this to the encoder/decoder, we need to forward this call to the decoder.

deltakosh:
cc @pandaGaume to see if he has a few cycles to finish this one? Thanks a million!

sebavan:
@pandaGaume, do you think you ll be able to finish this one ? or should we flag as draft until someone can pick up where it is ?

bghgary:
@pandaGaume I'm closing this PR for now. We can reopen if it becomes active again.

PR #12633, opened 2022-06-03
Added ```setQuaternion``` to ```Effect```

BarthPaleologue:
I'm back with more Quaternion's stuff.

I added ```setQuaternion``` to the Effect class.

I updated the different pipeline contexts versions to make it possible to set a quaternion as a uniform variable in a shader.
I created a IQuaternionLike interface, though it is the same as IVector4Like I think it makes the code easier to understand and read with this interface.


deltakosh (approved):
Thanks!

Issue #12634, opened 2022-06-04
Mesh’s ActionManager on chrome on mobile devices can cause the canvas to highlight on click/tap.


saaratrix:
In a chromium browser, eg: Chrome or Edge and on a mobile devices the canvas flickers blue because it's highlighted when the canvas' cursor is pointer and you click/tap on the canvas. It can also be observed using chrome's desktop mobile simulation mode.

This happens for example when you have an action manager for OnPointerOverTrigger, OnPointerOutTrigger that changes the `canvas.style.cursor = 'pointer'` when a pointer enters the mesh.
The behaviour can even be observed with just a `<div style="cursor: pointer;"></div>`

And it can be mitigated setting chrome's `-webkit-tap-highlight-color` property.

**Repro**
- Bug repro on [playground](https://playground.babylonjs.com/#J61EMG):
- Expected result: Click on the mesh few times and the canvas will not flicker blue.
- Current result: Click on the mesh few times and the canvas flickers blue.

**Screenshots**
![image](https://user-images.githubusercontent.com/16946048/171990557-8a8da7f6-0ddf-4346-955b-76f1f79c8d13.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chromium based
 - Version: Chrome 102.0.5005.63,  Edge 102.0.1245.30

**Smartphone (please complete the following information):**
 - Device: Sony Xperia 5 iii
 - OS: Android 12
 - Browser: Chromium based 
 - Version: Chrome 102.0.5005.78

**Additional context**
Related forum thread: https://forum.babylonjs.com/t/meshs-actionmanager-on-android-chrome-highlights-selects-canvas-on-click/30328

PR #12635, opened 2022-06-04
Fix where canvas could have a blue highlight effect on tap.

saaratrix:
For #12634.

Added a line to also disable the webkitTapHighlight when touch actions are disabled.


deltakosh (approved):
Thanks a lot

PR #12641, opened 2022-06-11
Added Quaternion support in `ShaderMaterial`

BarthPaleologue:
Last pull request about quaternions (until the next one maybe we never know ^^)

- added `setQuaternion` as a method of `ShaderMaterial`
- added `setQuaternionArray` as a method of `ShaderMaterial`
- added `toArray` as a method of `Quaternion`


packages/dev/core/src/Materials/shaderMaterial.ts
```diff
@@ -1290,6 +1340,10 @@ export class ShaderMaterial extends PushMaterial {
             result.setArray4(key, this._vectors4Arrays[key]);
         }
 
+        for (const key in this._vectors4Arrays) {
```
sebavan, line 1343: shouldn t it be quaternions array ?

sebavan (changes requested):
This looks okay to me but is there no way we could share most of the code with Vector 4 ? it is adding quite some code whereas at the end both structures and usage are identical ?

packages/dev/core/src/Materials/shaderMaterial.ts
BarthPaleologue: I think more code could be shared indeed, but looking at how Color4 is handled (it is also separated), I thought it was better to separate the code. If we share the code between Quaternion and Vector4, we should do it too with Color4 and Vector4. It can be done, I can change Color4 too if we choose to share the code. 
Oh and yes I made a mistake, I will fix it very soon !

sebavan: If it can be done and not too hard, would be great to share with color4 as well :-)

BarthPaleologue: I wonder, maybe we can go further in code sharing and declare that Color3 extends Vector3, Color4 extends Vector4 and Quaternion extends Vector4, we could remove a lot of code by doing something like this I think

BarthPaleologue: I shared the code between Array4, Color4Array and QuaternionArray, I fear this will break some backward compatibility as the serialize() method no longer returns different arrays in the json.

BarthPaleologue: If it does not pose any problem then I can share the code between Color3Array and Array3 too !

BarthPaleologue: no problem :wink:

BarthPaleologue: No bid deal haha, it's not much really it's fun to work on the engine ^^

PR #12643, opened 2022-06-11
Prepare plugin defines after material helper defines

BlakeOne:
For PBRBaseMaterial, this PR moves the call to _callbackPluginEventPrepareDefines to right after the calls to PrepareDefinesForFrameBoundValues and PrepareDefinesForAttributes instead of right before them. This way values set by the plugin for these defines won't be overwritten by these MaterialHelper functions. For StandardMaterial this is already the case (_callbackPluginEventPrepareDefines is called right after them).

Since the anisotropy plugin is relying on the current order of these function calls, PrepareDefinesBeforeAttributes was added to allow defines to be set right before MaterialHelper.PrepareDefinesForAttributes is called.

Forum: https://forum.babylonjs.com/t/pbrbasematerial-overwrites-plugins-defines/31262
Test PG: https://playground.babylonjs.com/#UIXTIF#6

### DIFF
--- a/packages/dev/core/src/Materials/PBR/pbrAnisotropicConfiguration.ts
+++ b/packages/dev/core/src/Materials/PBR/pbrAnisotropicConfiguration.ts
@@ -95,7 +95,7 @@ export class PBRAnisotropicConfiguration extends MaterialPluginBase {
         return true;
     }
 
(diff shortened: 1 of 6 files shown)


sebavan (commented):
LGTM let s see what @Popov72 thinks about it ?

Popov72 (approved):

PR #12646, opened 2022-06-13
Fix SSR and procedural textures clear order

sebavan:
https://forum.babylonjs.com/t/ssr-with-noiseproceduraltexture-caused-scene-not-refreshing/31290


packages/dev/core/src/Meshes/subMesh.ts
```diff
@@ -296,13 +296,14 @@ export class SubMesh implements ICullable {
 
     /**
      * Returns the submesh material
+     * @param getDefaultMaterial Defines whether or not to get the default material is nothing has been defined.
```
Popov72, line 299: nit: "is" => "if"

PR #12652, opened 2022-06-14
Fix texture caching.

carolhmj:
Related forum issue: https://forum.babylonjs.com/t/got-wrong-mapping-in-equirectangularcubetexture/31329
Related playground: https://playground.babylonjs.com/#NIWPUI#1


sebavan (commented):
Should we do it for all cube textures, passing isCube in ???

GG for the fix by the way :-)

carolhmj:
> Should we do it for all cube textures, passing isCube in ???
> 
> GG for the fix by the way :-)

Thanks! I'll add to the other cube textures :D

PR #12654, opened 2022-06-14
**DRAFT** glXF experimental support. Do not merge.

RaananW:
Please do not merge this PR, it is a WIP. The PR was created as a way to test it on the playground


bhouston:
FYI - I wrote an Unreal Engine Blueprints / Unity Visual Scripting like behaviour graph engine: https://github.com/bhouston/behave-graph. We are proposing this as an alternative to the Trigger-Action List method of behaviors.  It is an alternative to Trigger-Action Lists but Trigger-Action Lists can be converted to behavior graphs.  We are both presenting in the same Khronos SIGGRAPH BOF.  We may want to chat about this...

RaananW:
> FYI - I wrote an Unreal Engine Blueprints / Unity Visual Scripting like behaviour graph engine: https://github.com/bhouston/behave-graph. We are proposing this as an alternative to the Trigger-Action List method of behaviors. It is an alternative to Trigger-Action Lists but Trigger-Action Lists can be converted to behavior graphs. We are both presenting in the same Khronos SIGGRAPH BOF. We may want to chat about this...

Hi Ben, great presentation today!
Will be happy to talk about your proposal

PR #12655, opened 2022-06-15
Camera: Modified isMouseEvent bool logic for Safari

PolygonalSun:
There was an issue found in the forum where movement of the UniversalCamera, using a mouse, on Safari wasn't stopping properly.  After investigating the issue, it was found that the mouse events were triggering both mouse and touch movement.  The main issue was that, in FreeCameraTouchInput, the `isMouseEvent` boolean was incorrect being evaluated as false for mouse because the assigning logic was looking for a parameter that isn't available to mouse events generated in Safari.  The fix was to modify the logic to work with Safari-specific mouse events.

Relevant Forum Link: https://forum.babylonjs.com/t/keyboard-control-not-working-for-latest-universalcamera-in-desktop-safari/31314/8

### DIFF
--- a/packages/dev/core/src/Cameras/Inputs/freeCameraTouchInput.ts
+++ b/packages/dev/core/src/Cameras/Inputs/freeCameraTouchInput.ts
@@ -45,6 +45,7 @@ export class FreeCameraTouchInput implements ICameraInput<FreeCamera> {
     private _pointerInput?: (p: PointerInfo, s: EventState) => void;
     private _observer: Nullable<Observer<PointerInfo>>;
     private _onLostFocus: Nullable<(e: FocusEvent) => any>;
+    private _isSafari: boolean;
 
     /**
      * Manage the touch inputs to control the movement of a free camera.
@@ -56,7 +57,9 @@ export class FreeCameraTouchInput implements ICameraInput<FreeCamera> {
          * Define if mouse events can be treated as touch events
          */
         public allowMouse = false
-    ) {}
+    ) {
+        this._isSafari = Tools.IsSafari();
+    }
 
     /**
      * Attach the input controls to a specific dom element to get the input from.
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Cameras/Inputs/freeCameraTouchInput.ts
```diff
@@ -76,7 +79,7 @@ export class FreeCameraTouchInput implements ICameraInput<FreeCamera> {
             this._pointerInput = (p) => {
                 const evt = <IPointerEvent>p.event;
 
-                const isMouseEvent = evt.pointerType === "mouse";
+                const isMouseEvent = evt.pointerType === "mouse" || (this._isSafari && typeof evt.pointerType === "undefined");
```
sebavan, line 82: Would that not break on mobile ??? can you try on ios ?

PolygonalSun: I tested things on my iPad and it's working as expected.  In addition, `Tools.GetPointerPrefix()` evaluates to "pointer" as well as has the correct values for `pointerType` so this shouldn't break iOS.

sebavan: Do you have an ios device to check ? if not I could try locally on one ? better safe than sorry with Safari :-)

sebavan: > Do you have an ios phone device to check ? if not I could try locally on one ? better safe than sorry with Safari :-)

jure:
Good hunting! Reverting the fix from this commit (made 28 days ago) https://github.com/BabylonJS/Babylon.js/commit/6c9bd9cff50b0619dc80b45b662498423cf2d8e6 resolved our issues completely, including the keyboard navigation issue that was resolved in a different way in #12649. So this is definitely the root cause.

The fix in the this PR also resolves both of our issues. If it also resolves the issue from #12541, then it's good to go :).

packages/dev/core/src/Cameras/Inputs/freeCameraTouchInput.ts
PolygonalSun: Yup, everything works on my iPhone.  I even verified the values for Tools.GetPointerPrefix and that pointer events generated on my iPhone have a correct value for pointerType.  We should be good with iOS

sebavan (approved):

PR #12659, opened 2022-06-16
Adds ExtrudeShape* options firstNormal and adjustFrame

Lowclouds:
firstNormal option allows one to pick the first normal of the underlying Path3D path, preventing, in some cases degenerate extrusions.
adjustNormal option causes the underlying Path3D frame of tangents, normals, and binormals to be patched up when the they are either undefined (0,0,0) or reverse direction. This is useful for paths that reverse back on themselves.


Lowclouds:
This playground snippet, #5ICT41#4,  tests these new features - except for instances. I haven't used instances, yet

Lowclouds:
I have a more compact and idiomatic, but functionally identical version

packages/dev/core/src/Meshes/Builders/shapeBuilder.ts
```diff
@@ -230,13 +250,56 @@ function _ExtrudeShapeGeneric(
...
         const tangents = path3D.getTangents();
         const normals = path3D.getNormals();
         const binormals = path3D.getBinormals();
         const distances = path3D.getDistances();
-
+        if (adjustFrame) {
+            /* fix tangents,normals, binormals */
+            for (let i = 0; i < tangents.length; i++) {
+                if (tangents[i].x == 0 && tangents[i].y == 0 && tangents[i].z == 0) {
+                    tangents[i].x = tangents[i - 1].x;
```
RaananW, line 264: if this is a vector3 (which I am pretty sure it is?), why not use `.copyFrom()` instead? it's just cleaner, as copyFrom does exactly what these lines do

packages/dev/core/src/Meshes/Builders/shapeBuilder.ts
```diff
@@ -230,13 +250,56 @@ function _ExtrudeShapeGeneric(
...
+                if (binormals[i].x == 0 && binormals[i].y == 0 && binormals[i].z == 0) {
+                    binormals[i].x = binormals[i - 1].x;
+                    binormals[i].y = binormals[i - 1].y;
+                    binormals[i].z = binormals[i - 1].z;
+                }
+                if (i > 0) {
+                    let v = tangents[i - 1];
+                    if (Vector3.Dot(v, tangents[i]) < 0) {
+                        //                        console.log(`fixing tangents[${i}]: ${tangents[i]}`);
+                        tangents[i].x *= -1;
```
RaananW, line 282: `.scaleInPlace(-1)`

RaananW (commented):
LGTM! Most of my comments are about reusability of the vector class's functions, so keep the code a little cleaner.

packages/dev/core/src/Meshes/Builders/shapeBuilder.ts
Lowclouds: Should I push my changes that fix this?

RaananW: yes please, just push to the same branch and it will be updated here

RaananW (approved):

PR #12662, opened 2022-06-17
Basis transcoder update

vinhui:
This is an update to the basis transcoder. This allows users to also take advantage of the `UASTC` option of the basis format. 

The `basis_transcoder.js` and `basis_transcoder.wasm` are a direct copy from the [Basis Universal repository](https://github.com/BinomialLLC/basis_universal/tree/7a2094b/webgl/transcoder/build).


packages/dev/core/src/Misc/basis.ts
```diff
@@ -327,22 +327,20 @@ function workerFunc(): void {
...
                     importScripts(event.data.url);
                 } catch (e) {
                     postMessage({ action: "error", error: e });
                 }
-                transcoderModulePromise = new Promise<void>((res) => {
-                    Module.onRuntimeInitialized = () => {
-                        Module.initializeBasis();
-                        res();
-                    };
+                transcoderModulePromise = BASIS({
```
vinhui, line 374: I'm aware that typescript would like to know where `BASIS` comes from, but I'm not sure how to do that properly. It gets defined in the imported script.

bghgary: You can replace `Module` with `BASIS` above. That's how you tell typescript this global object exists.

bghgary (changes requested):
Some minor changes required, but looks good otherwise.

vinhui:
> @vinhui Shouldn't you also be updating the `_BasisTextureLoader` in https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Materials/Textures/Loaders/basisTextureLoader.ts#L47 and https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Materials/Textures/Loaders/basisTextureLoader.ts#L89?

Good point, I'll update that.

packages/dev/core/src/Misc/basis.ts
vinhui: So the latest changes should be correct, right?

bghgary: Yes, I believe so.

bghgary: Yes, I believe so.

bghgary:
> Since the basisu library interface has changed, this change would not be backwards compatible because of the usage of the hosted files. How should this be handled? Updating the files at `https://preview.babylonjs.com/basisTranscoder/` would break existing versions while this PR is not compatible with the current version. This could possibly be handled by uploading the new files under a different name and changing that hardcoded url as well.

Good point. I discussed this with the team offline. We think adding a version to the url is the best solution.

Something like:
`https://preview.babylonjs.com/basisTranscoder/1/...`

Thanks for all the updates!

packages/dev/core/src/Misc/basis.ts
```diff
@@ -310,46 +331,56 @@ export const BasisTools = {
...
+                    // Override wasm binary
+                    wasmBinary: event.data.wasmBinary,
                 });
             }
-            transcoderModulePromise.then(() => {
+            transcoderModulePromise.then((m) => {
```
rassie, line 375: The pipeline fails with `Object is possibly 'null'` at this point.

rassie:
@vinhui Sorry for coming up with more work for you, but I suppose you'll need to update `GetInternalFormatFromBasisFormat` at https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Misc/basis.ts#L110 too.

rassie:
> Since the basisu library interface has changed, this change would not be backwards compatible because of the usage of the hosted files. How should this be handled? Updating the files at `https://preview.babylonjs.com/basisTranscoder/` would break existing versions while this PR is not compatible with the current version. This could possibly be handled by uploading the new files under a different name and changing that hardcoded url as well.

If I understand it correctly, one of the things that happened between "old" basis transcoder and the "new" one is the renaming and re-ordering of `transcoder_texture_format` enum (see comment above about `GetInternalFormatFromBasisFormat`). From what I could find, the change happened in `basis_universal` version 1.15 in https://github.com/BinomialLLC/basis_universal/commit/54304fa01443280bbaa9854f16c9f1756b17b40b#diff-f9db3d02491fb0018b8046bb9ff087cfce57e86a0c9ee25d846fca5065efb6ae. So I would assume that any files compressed with `basisu` < 1.15 will be compatible with current Babylon.js version, but anything compressed with `basisu` >= 1.15 will only be compatible with any Babylon.js version released after this PR is merged. Old files might be compatible, but I'd expect a breakage because of that `transcoder_texture_format` enum rename.

Can we alleviate that or do we just add a warning to release notes, something along the lines of "May break unless you re-compress all of your assets"?

rassie:
> > re-ordering of transcoder_texture_format enum
> 
> It looks like from the code, Rich tried to maintain the old enums (the old values are still the same as before). There are new enums though. I think this will only be an issue if new files encoded with new enums are loaded with old loaders, which seems reasonable.

We might be looking at different code :) In BASIS_FORMATS basically all values are different (look at ID 6 or 1 for example) and besides all that, they come straight from Basis source code. The problem is, as you say, first and foremost loading newly encoded files with old loader (seems completely impossible on my machine, format is not detected), but there is probably a problem with old files and new loader, the error message will be different though.

vinhui:
Sorry for the delay. 
@rassie I appreciate you taking a good look :)

vinhui:
> Please make sure to label your PR with "bug", "new feature" or "breaking change" label(s). To prevent this PR from going to the changelog marked it with the "skip changelog" label.

Just to confirm, this is not something I can do, right?

packages/dev/core/src/Materials/Textures/Loaders/basisTextureLoader.ts
```diff
@@ -105,8 +105,9 @@ export class _BasisTextureLoader implements IInternalTextureLoader {
                     LoadTextureFromTranscodeResult(texture, result);
                 });
             })
-            .catch(() => {
+            .catch((err) => {
                 Tools.Warn("Failed to transcode Basis file, transcoding may not be supported on this device");
+                Tools.Warn(err);
```
bghgary, line 110:
```suggestion
                Tools.Warn(`Failed to transcode Basis file: ${err}`);
```

bghgary (approved):
Tiny nit. Looks good otherwise.

bghgary:
> > Since the basisu library interface has changed, this change would not be backwards compatible because of the usage of the hosted files. How should this be handled? Updating the files at `https://preview.babylonjs.com/basisTranscoder/` would break existing versions while this PR is not compatible with the current version. This could possibly be handled by uploading the new files under a different name and changing that hardcoded url as well.
> 
> Good point. I discussed this with the team offline. We think adding a version to the url is the best solution.
> 
> Something like: `https://preview.babylonjs.com/basisTranscoder/1/...` `https://cdn.babylonjs.com/basisTranscoder/1/...` (EDIT: we can/should use the `cdn` path now actually)
> 
> Thanks for all the updates!

I prematurely approved. I don't see this addressed. Marking as draft until it is. :-)

bghgary:
> > Please make sure to label your PR with "bug", "new feature" or "breaking change" label(s). To prevent this PR from going to the changelog marked it with the "skip changelog" label.
> 
> Just to confirm, this is not something I can do, right?

I tagged it as `new feature`.

bghgary:
> > > > Since the basisu library interface has changed, this change would not be backwards compatible because of the usage of the hosted files. How should this be handled? Updating the files at `https://preview.babylonjs.com/basisTranscoder/` would break existing versions while this PR is not compatible with the current version. This could possibly be handled by uploading the new files under a different name and changing that hardcoded url as well.
> > > 
> > > 
> > > Good point. I discussed this with the team offline. We think adding a version to the url is the best solution.
> > > Something like: `https://preview.babylonjs.com/basisTranscoder/1/...` `https://cdn.babylonjs.com/basisTranscoder/1/...` (EDIT: we can/should use the `cdn` path now actually)
> > > Thanks for all the updates!
> > 
> > 
> > I prematurely approved. I don't see this addressed. Marking as draft until it is. :-)
> 
> I wasn't sure if I should update that since it might be easy to forget to upload the files to the cdn. I thought it might be better for the person that will upload the files to also commit the change.

@RaananW can say for sure, but I think the CDN just mirrors what you merge to the `public/basisTranscoder` folder.

vinhui:
FYI, the tests fail because of the CDN. If I change the urls locally to existing files, the failing tests pass.

bghgary:
Hmm, good point. For other configurations, we do this to make sure the configuration points to the built version. Can you try adding Basis to this list? @RaananW would be the best person to confirm this, but he is out right now.

https://github.com/BabylonJS/Babylon.js/blob/master/packages/tools/testTools/src/utils.ts#L65-L76

vinhui:
> Hmm, good point. For other configurations, we do this to make sure the configuration points to the built version. Can you try adding Basis to this list? @RaananW would be the best person to confirm this, but he is out right now.
> 
> https://github.com/BabylonJS/Babylon.js/blob/master/packages/tools/testTools/src/utils.ts#L65-L76
> 
> EDIT: here also https://github.com/BabylonJS/Babylon.js/blob/master/packages/tools/testTools/src/visualizationUtils.ts#L17-L38

I added it, it seems to work locally now.

bghgary:
You don't want to remove the old files since that needs to be there for backwards compatibility.

> packages/tools/babylonServer/public/basisTranscoder/basis_transcoder.js
> packages/tools/babylonServer/public/basisTranscoder/basis_transcoder.wasm

Revert these and I think we are ready to merge.

bghgary:
@vinhui Thanks for doing all this work!

vinhui:
Thank you guys for being so patient and thorough.

And just to double check, the files will automatically get uploaded to the cdn subdomain and the old ones will stay on the preview subdomain?

bghgary:
Both `preview` and `cdn` point to the same place right now. 😉

vinhui:
> @bghgary - this overwritten the old basis files. Where can the old files be found? I thought we are going to add versioning?

Both versions are now in git. The new version is in `public/basisTranscoder/1/` while the old, unchanged ones are in `public/basisTranscoder/`.

RaananW:
> > @bghgary - this overwritten the old basis files. Where can the old files be found? I thought we are going to add versioning?
> 
> Both versions are now in git. The new version is in `public/basisTranscoder/1/` while the old, unchanged ones are in `public/basisTranscoder/`.

Thanks, i reviewed an older PR that didn't have them in the /1/ directory. That's perfectly fine the way it is now :-)

PR #12663, opened 2022-06-17
Bug Fix - Vector3 project on plane

strutcode:
This proposal fixes an apparent incorrect calculation in the `Vector3.projectOnPlaneToRef` method which returns confusing results for many planes, especially those not centered on the world origin.

A couple of unit tests were added to logically demonstrate the failure of the existing code, as well as a new original implementation which seems to work correctly for various cases.

Playground demonstrating the behavior for testing:
https://playground.babylonjs.com/#MHMIPM

### DIFF
--- a/packages/dev/core/src/Maths/math.vector.ts
+++ b/packages/dev/core/src/Maths/math.vector.ts
@@ -1164,7 +1164,7 @@ export class Vector3 {
         // Calculate how close the direction is to the normal of the plane
         const dotProduct = Vector3.Dot(inverseNormal, direction);
 
-        /* 
+        /*
          * Early out in case the direction will never hit the plane.
          *
          * Epsilon is used to avoid issues with rays very near to parallel with the
--- a/packages/dev/core/src/Maths/math.vector.ts
+++ b/packages/dev/core/src/Maths/math.vector.ts
@@ -1150,22 +1150,43 @@ export class Vector3 {
      * @param result defines the Vector3 where to store the result
      */
     public projectOnPlaneToRef(plane: Plane, origin: Vector3, result: Vector3): void {
-        const n = plane.normal;
-        const d = plane.d;
+        // Use the normal scaled to the plane offset as the origin for the formula
+        const planeOrigin = MathTmp.Vector3[0];
+        plane.normal.scaleToRef(-plane.d, planeOrigin)
 
-        const V = MathTmp.Vector3[0];
+        // Since the normal in Babylon should point toward the viewer, invert it for the dot product
+        const inverseNormal = MathTmp.Vector3[1];
+        plane.normal.negateToRef(inverseNormal);
+        
+        // This vector is the direction
+        const direction = this;
 
-        // ray direction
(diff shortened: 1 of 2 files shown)


packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1150,22 +1150,43 @@ export class Vector3 {
...
+
+        // Since the normal in Babylon should point toward the viewer, invert it for the dot product
+        const inverseNormal = MathTmp.Vector3[1];
+        plane.normal.negateToRef(inverseNormal);
+        
+        // This vector is the direction
+        const direction = this;
+
+        // Set a margin to prevent issues with directions very close to perpendicular to the plane
+        const epsilon = 0.000001;
```
RaananW, line 1165: This exists in math.constants (and is already imported to the math module)

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1150,22 +1150,43 @@ export class Vector3 {
...
+        // Use the normal scaled to the plane offset as the origin for the formula
+        const planeOrigin = MathTmp.Vector3[0];
+        plane.normal.scaleToRef(-plane.d, planeOrigin)
+
+        // Since the normal in Babylon should point toward the viewer, invert it for the dot product
+        const inverseNormal = MathTmp.Vector3[1];
+        plane.normal.negateToRef(inverseNormal);
+        
+        // This vector is the direction
+        const direction = this;
```
RaananW, line 1162: why is that needed?

RaananW (commented):
LGTM!

this is your playground with the snapshot - https://playground.babylonjs.com/?snapshot=refs/pull/12663/merge#MHMIPM

Just two questions about the code itself, and you have a formatting issue. you can run `npm run format:fix` if you don't use a formatter directly in your IDE.

packages/dev/core/src/Maths/math.vector.ts
strutcode: Good call! It's probably better to use a consistent fudge value. :)

packages/dev/core/src/Maths/math.vector.ts
strutcode: There's no reason, it's just a habit of mine to name things in a way that makes it as clear as possible what the code is doing. I believe it should be optimized out so it's purely aesthetic.

strutcode:
Strange! I already fixed the formatting issues. I wonder why it's not showing up. I'll figure it out though.

packages/dev/core/src/Maths/math.vector.ts
strutcode: Let me know if I should change this or if it's good as is. :+1:

RaananW (approved):

sebavan:
@strutcode would be great to for you to have a look at https://forum.babylonjs.com/t/project-3d-point-on-plane/9862/18?u=sebavan and maybe create a separate function for your behavior ?

PR #12664, opened 2022-06-18
Added referrer policy support to GUI Image to control xhr request header

BrunevalPE:
This feature will allow user's to control the **header** in the **xhr request to load the image**, `Referrer Policy`.

More info here : 
The Referrer-Policy [HTTP header](https://developer.mozilla.org/en-US/docs/Glossary/HTTP_header) 

Default value is : `strict-origin-when-cross-origin`

### DIFF
--- a/packages/dev/gui/src/2D/controls/image.ts
+++ b/packages/dev/gui/src/2D/controls/image.ts
@@ -293,7 +293,7 @@ export class Image extends Control {
     }
 
     /**
-     * Gets or sets the referrer policy to apply on the img load request, 
+     * Gets or sets the referrer policy to apply on the img load request,
      * you should set this field before set the source field if you want to ensure the header will be present on the xhr loading request
      */
     @serialize()
@@ -541,7 +541,7 @@ export class Image extends Control {
         }
         this._domImage = engine.createCanvasImage();
         this._domImage.referrerPolicy = this._referrerPolicy;
-        
+
         this._domImage.onload = () => {
             this._onImageLoaded();
         };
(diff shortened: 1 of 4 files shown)


packages/dev/gui/src/2D/controls/image.ts
```diff
@@ -291,6 +292,24 @@ export class Image extends Control {
         }
     }
 
+    /**
+     * Gets or sets the referrer policy to apply on the img load request,
+     * you should set this field before set the source field if you want to ensure the header will be present on the xhr loading request
+     */
+    @serialize()
```
sebavan, line 299: Not sure if this should be serialized ???

@RaananW ???

sebavan:
Adding @RaananW as a reviewer who is way more versed that I am on this part :-)

I love the overall idea but wonder about persistence as well as should we do it for the 3d textures and much more ?

packages/dev/gui/src/2D/controls/image.ts
RaananW: That's an interesting one. 
I'll comment about the PR in general, but in this specific instance, if it made sense to merge this PR then it is fine to serialize it.

RaananW (commented):
I think it's more of a question of whether or not this is the right place for this.

We already have `Tools.SetCorsBehavior`, which changes the `crossOrigin` of the specific request/image. We also have the WebRequest modifiers to change any request made ourside of a simple image load. This feels like a 3rd way to solve a similar issue.

I like the idea of adding it, but we need to find a global way to be able to set this. Only in the GUI is probably not suffecient.

BrunevalPE:
Hello ! I'm not sure for the serialization, so I removed it, I will put it back if you think it's relevant.
I don't want to consume memory on Image instances so I changed it to be nullable and not forcing the default state.
Tools seems to be a better place for this code, indeed 👍

RaananW (approved):
LGTM!

@sebavan, any objection merging this?

PR #12668, opened 2022-06-20
Add params scene, mesh to function getAttributes

BlakeOne:
For the class MaterialPluginBase,  add params scene and mesh to the function getAttributes so that they can be used to determine which attributes to use.

Forum: https://forum.babylonjs.com/t/using-custom-attribute-with-material-plugin/31142/7


sebavan (commented):
Just wondering if it should be the mesh or submesh here ?

BlakeOne:
> should be the mesh or submesh here ?

Petty sure it should be the mesh because that's what's passed to the PBR and StandardMaterial functions that trigger the event, not a sub mesh, and also that's what prepareDefines passes.

sebavan (approved):

PR #12670, opened 2022-06-21
[GUI] Lines array is generated later

RaananW:
this._lines is generated when the  control is in view. If you create a control that is not in view it will fail.

This makes sure it doesn't fail even when out of view. 

Test playground - https://www.babylonjs-playground.com/#XCPP9Y#13890


sebavan (approved):

sebavan:
overall It sounds weird we are going in render if we do not have the required infos ???

RaananW:
> overall It sounds weird we are going in render if we do not have the required infos ???

That is the GUI render loop. and I agree. Those objects are only initialized on the first actual render, but the function is running on every render call, even if outside of the frustum.

This requires a slight overhaul of the GUI render loop.

PR #12675, opened 2022-06-22
Add support for more depth stencil formats.

kircher1:
This requires some rework of the logic in ThinEngine._createDepthStencilTexture, but it should be non-breaking. A similar change was made to extend MultiRenderTarget's depth-stencil texture handling too.

I've also added visualization tests to exercise the createDepthStencilTexture code path.


kircher1:
I've been struggling to get the visualization tests working and passing locally (not just the new tests, but also existing tests). And not just for this branch, main too. So, still trying to figure that out.

kircher1:
Need to fix some GL1 related issues.

packages/dev/core/src/Engines/Extensions/engine.renderTarget.ts
```diff
@@ -132,21 +132,61 @@ ThinEngine.prototype._createDepthStencilTexture = function (size: TextureSize, o
...
+            sizedFormat = gl.DEPTH_COMPONENT16;
+        } else if (internalTexture.format === Constants.TEXTUREFORMAT_DEPTH24) {
+            sizedFormat = gl.DEPTH_COMPONENT24;
+        } else if (internalTexture.format === Constants.TEXTUREFORMAT_DEPTH24UNORM_STENCIL8 || internalTexture.format === Constants.TEXTUREFORMAT_DEPTH24_STENCIL8) {
+            sizedFormat = gl.DEPTH24_STENCIL8;
+        } else if (internalTexture.format === Constants.TEXTUREFORMAT_DEPTH32_FLOAT) {
+            sizedFormat = gl.DEPTH_COMPONENT32F;
+        } else if (internalTexture.format === Constants.TEXTUREFORMAT_DEPTH32FLOAT_STENCIL8) {
+            const gl2 = <WebGL2RenderingContext>(this._gl as any);
+            sizedFormat = gl2.DEPTH32F_STENCIL8;
```
Popov72, line 188: I think you should add `DEPTH32F_STENCIL8` in `LibDeclarations/webgl2.d.ts` instead and simply do `sizedFormat = gl.DEPTH32F_STENCIL8`.

packages/tools/tests/test/visualization/config.json
```diff
@@ -1371,6 +1371,19 @@
             "playgroundId": "#BCYE7J#31",
             "referenceImage": "multiCamerasOutputRenderTarget.png"
         },
+        {
+            "title": "Multi cameras and output render target with depth stencil texture (D24_S8)",
+            "renderCount": 2,
+            "playgroundId": "#BCYE7J#33",
```
Popov72, line 1377: This PG already works without the PR: the fact that it is using only a depth and not a depth/stencil texture can't be seen (we can't read the stencil in WebGL2). I don't think it's possible to check that the right format is really used under the hood, so it's probably better to remove this validation test.

packages/tools/tests/test/visualization/config.json
```diff
@@ -1371,6 +1371,19 @@
...
+        {
+            "title": "Multi cameras and output render target with depth stencil texture (D24_S8)",
+            "renderCount": 2,
+            "playgroundId": "#BCYE7J#33",
+            "referenceImage": "multiCamerasOutputRenderTarget.png"
+        },
+        {
+            "title": "Multi cameras and output render target with depth stencil texture (D32F_S8)",
+            "renderCount": 2,
+            "playgroundId": "#BCYE7J#34",
```
Popov72, line 1383: Same comment than above.

Popov72 (commented):
Good job!

You will also have to update WebGPU. It's in `Engines/WebGPU/Extensions/engine.renderTarget.ts`, the `_createDepthStencilTexture` function.

packages/tools/tests/test/visualization/config.json
kircher1: I simplified the case down to just one which exercises a GL error prior to my change that would lead to a blank render target.

https://www.babylonjs-playground.com/#BCYE7J#36

kircher1:
> You will also have to update WebGPU. It's in `Engines/WebGPU/Extensions/engine.renderTarget.ts`, the `_createDepthStencilTexture` function.

@Popov72, from the looks of it, the WebGPU implementation already handles the various depth stencil formats. Correct me if I'm wrong and I will make the necessary updates.

Popov72:
I think line 83 of the file I pointed above should be replaced by:

```typescript
depthTextureFormat: options.depthTextureFormat ?? options.generateStencil ? Constants.TEXTUREFORMAT_DEPTH24_STENCIL8 : Constants.TEXTUREFORMAT_DEPTH32_FLOAT,
```

sebavan (approved):
LGTM as soon as @Popov72 concerns have been addressed

kircher1:
> ```ts
> TEXTUREFORMAT_DEPTH24_STENCIL8 
> ```

@Popov72, does the spread operator for `...options` on line 84 handle overwriting depthTextureFormat?

``` 
 const internalOptions = {
        bilinearFiltering: false,
        comparisonFunction: 0,
        generateStencil: false,
        samples: 1,
        depthTextureFormat: options.generateStencil ? Constants.TEXTUREFORMAT_DEPTH24_STENCIL8 : Constants.TEXTUREFORMAT_DEPTH32_FLOAT,
        ...options,
    };
`

Popov72:
Yes, my bad, everything is ok with no change!

RaananW:
This is good to merge, right?

PR #12678, opened 2022-06-24
Fix when BoundingBoxGizmo issues warning

BlakeOne:
PR to move the warning about setting the attachedNode property of the BoundingBoxGizmo into the setter for attachedNode to make sure that the warning is only logged when the attachedNode property is actually set. Currently there are a few commonly occurring cases when the warning will be issued even when the property isn't being used.

Forum discussion with more info and repros:

### DIFF
--- a/packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
+++ b/packages/dev/core/src/Gizmos/boundingBoxGizmo.ts
@@ -174,6 +174,20 @@ export class BoundingBoxGizmo extends Gizmo {
             }
         });
     }
+
+    /**
+     * Using the attachedNode property in BoundingBoxGizmo is not supported. Please use attachedMesh instead.
+     */
+     public get attachedNode() {
+        return super.attachedNode;
+    }
+
+    public set attachedNode(value) {
+        super.attachedNode = value;
+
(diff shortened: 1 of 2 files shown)


BlakeOne:
Hmm I wonder why the build failed? I'm able to build locally successfully using  "npm run build:dev"

RaananW:
This is the error:

https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=20769&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=b4978d5e-5988-54a5-1480-26654484e930&l=276

Super cannot be used in assessors like that, especially not if it is seeing anything private. The get is not needed, the set can be modified to use a protected variable in the parent

I wonder why it compiles locally. When was the last time you ran npm install?

BlakeOne:
> Super cannot be used in assessors like that

Aah my bad, I was doing it a way that was recommended for ES6 that lets the typing get sorted out automatically. I'll trying running install again too, it's been quite a while... Thanks for the help!

BlakeOne:
Will convert to draft for now, not as straightforward as I thought it would be...

BlakeOne:
Will just change this to a draft for now, it's not as straightforward a solution as I thought...

PR #12682, opened 2022-06-27
Wrong rotation offset camera correction

pandaGaume:
When GLB export, camera rotation was set using WorldTransform matrix `camera.getAbsoluteRotation()` instead as local. Resulting on double every rotation from upper hierarchy.
we update it  using  `camera.rotationQuaternion` which is property of `TargetCamera`
In order to let other user custom Camera to be exported, we do not change the `Camera` parameter type but cast as `any` to allow the property test.
See this forum [question](https://forum.babylonjs.com/t/gltf-export-bug-freecamera-rotation-has-wrong-offset/31278).

### DIFF
--- a/packages/dev/serializers/src/glTF/2.0/glTFExporter.ts
+++ b/packages/dev/serializers/src/glTF/2.0/glTFExporter.ts
@@ -48,7 +48,6 @@ import { _GLTFAnimation } from "./glTFAnimation";
 import { Camera } from "core/Cameras/camera";
 import { EngineStore } from "core/Engines/engineStore";
 import { MultiMaterial } from "core/Materials/multiMaterial";
-import { TargetCamera } from "core/Cameras/targetCamera";
 
 /**
  * Utility interface for storing vertex attribute data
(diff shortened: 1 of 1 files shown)


deltakosh (approved):

packages/dev/serializers/src/glTF/2.0/glTFExporter.ts
```diff
@@ -2216,7 +2218,7 @@ export class _Exporter {
                     }
                     return node;
                 });
-            } else if (babylonNode instanceof Camera) {
+            } else if (babylonNode instanceof TargetCamera) {
```
bghgary, line 2221: Hmm, we don't have any built-in cameras that are not derived from `TargetCamera`, but someone could create a camera that is directly derived from `Camera`. This will no longer work after this change. Are we okay with that?

pandaGaume: I'm ok with that, this was indeed a choice i made to keep strong typing. My first version was with a cast in `any `to test the existence of the `rotationQuaternion `property. 
Another choice was to put the rotation property at `Camera `level (as beeing done with `absoluteRotation `which is only used into XR) but i want to avoid changes at core level.
I can get back to this.
`const rotationQuaternion = (<any>babylonCamera).rotationQuaternion;`
and then rollback the test to 
`if (babylonNode instanceof Camera)`
Let me know.

PR #1043, opened 2016-03-14
Update babylon.skeleton.ts : Add getIndexOnSkeletonByName(string)

dad72:
This allows you to:
```
var hant_left = skeleton.getIndexOnSkeletonByName("hant_left");
sword.attachToBone(skeleton.bones[hant_left], character);
```
It is easier to give a bone name rather index for attachToBone

PR #12686, opened 2022-06-28
Update GLTF Animation serializer to include Camera.

pandaGaume:
`Camera `and `Light` was not exported to GLTF due to single test about `TransformNode`. We introduce a test function, which encompass test for `TransformNode`, `Camera `and `Light`. This test function will evolve to add support for `KHR_animation_pointer`.
See [this](https://forum.babylonjs.com/t/gltf-bug-freecamera-animation-not-being-exported/31259) forum thread.

### DIFF
--- a/packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
+++ b/packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
@@ -90,8 +90,7 @@ export class _GLTFAnimation {
      * @param babylonNode the node to test
      * @returns true if can be animated, false otherwise. False if the parameter is nullm or undefined.
      */
-    public static IsAnimatable(babylonNode: Node): boolean {
-        // TODO - test the KHR_animation_pointer capability.
+    private static _IsAnimatable(babylonNode: Node): boolean {
         return babylonNode && (babylonNode instanceof TransformNode || babylonNode instanceof Camera || babylonNode instanceof Light);
     }
 
@@ -264,7 +263,7 @@ export class _GLTFAnimation {
         animationSampleRate: number
     ) {
         let glTFAnimation: IAnimation;
-        if (_GLTFAnimation.IsAnimatable(babylonNode)) {
+        if (_GLTFAnimation._IsAnimatable(babylonNode)) {
             if (babylonNode.animations) {
                 for (const animation of babylonNode.animations) {
                     const animationInfo = _GLTFAnimation._DeduceAnimationInfo(animation);
--- a/packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
+++ b/packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
@@ -80,17 +80,20 @@ enum _TangentType {
      */
     OUTTANGENT,
 }
+
+type TransformableType = TransformNode | Camera | Light;
+
 /**
  * @hidden
  * Utility class for generating glTF animation data from BabylonJS.
  */
 export class _GLTFAnimation {
     /**
-     * Determine if a node is animatable
+     * Determine if a node is transformable - ie has properties it should be part of animation of transformation.
      * @param babylonNode the node to test
-     * @returns true if can be animated, false otherwise. False if the parameter is nullm or undefined.
+     * @returns true if can be animated, false otherwise. False if the parameter is null or undefined.
      */
-    private static _IsAnimatable(babylonNode: Node): boolean {
+    private static _IsTransformable(babylonNode: Node): babylonNode is TransformableType {
         return babylonNode && (babylonNode instanceof TransformNode || babylonNode instanceof Camera || babylonNode instanceof Light);
     }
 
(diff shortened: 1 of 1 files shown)


packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
```diff
@@ -943,7 +956,7 @@ export class _GLTFAnimation {
     }
 
     private static _GetBasePositionRotationOrScale(
-        babylonTransformNode: TransformNode,
+        babylonTransformNode: any,
```
RaananW, line 959: Why is that any now?

pandaGaume: Because we access specific `Position` and `Rotation`, which are not available at `Node` level. And because it is private i do not see any problem to put `any` instead of `Node`. Maybe i miss some rules :-)

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
```diff
@@ -82,6 +85,16 @@ enum _TangentType {
  * Utility class for generating glTF animation data from BabylonJS.
  */
 export class _GLTFAnimation {
+    /**
+     * Determine if a node is animatable
+     * @param babylonNode the node to test
+     * @returns true if can be animated, false otherwise. False if the parameter is nullm or undefined.
+     */
+    public static IsAnimatable(babylonNode: Node): boolean {
+        // TODO - test the KHR_animation_pointer capability.
```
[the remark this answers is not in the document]
pandaGaume: I will remove it

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
bghgary: It looks like this function is guaranteed to take a TransformNode. https://github.com/BabylonJS/Babylon.js/blob/a18bc8d066805484845d9c8e784d08f88588af3b/packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts#L836
TypeScript automatically narrows the type when using `instanceof`, so you don't actually need to change the types for this function or `_ConvertFactorToVector3OrQuaternion`.

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
```diff
@@ -757,7 +770,7 @@ export class _GLTFAnimation {
 
     private static _ConvertFactorToVector3OrQuaternion(
         factor: number,
-        babylonTransformNode: TransformNode,
+        babylonTransformNode: Node,
```
bghgary, line 772: See other comment.

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
pandaGaume: right but also used line #783 which is the reason of the change. and #1043 for `_ConvertFactorToVector3OrQuaternion`

bghgary: All of these should use TransformNode right? If it's not a TransformNode, this function won't work or am I misunderstanding that?

pandaGaume: It's not so obvious. Camera and TransformNode share similar properties, such rotationQuaternion or Position, and this is what we use to animate both.
This is come from the fact, At the very begining of the process, the serializer use the Camera as a TransformNode to leverage this similarities.
`const nodes: Node[] = [...babylonScene.transformNodes, ...babylonScene.meshes, ...babylonScene.lights, ...babylonScene.cameras];`
So it's not a garanty to get a TransformNode on each of this function.
However, youre right, we might verify if the property is defined prior to use it.

bghgary: Ok, can these functions take a union type like `TransformNode | Camera`? That should avoid using `any`?

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
```diff
@@ -943,37 +965,37 @@ export class _GLTFAnimation {
...
         if (animationChannelTargetPath === AnimationChannelTargetPath.ROTATION) {
             if (useQuaternion) {
-                if (babylonTransformNode.rotationQuaternion) {
-                    basePositionRotationOrScale = babylonTransformNode.rotationQuaternion.asArray();
-                    if (convertToRightHandedSystem) {
-                        _GLTFUtilities._GetRightHandedQuaternionArrayFromRef(basePositionRotationOrScale);
-                        if (!babylonTransformNode.parent) {
-                            basePositionRotationOrScale = Quaternion.FromArray([0, 1, 0, 0]).multiply(Quaternion.FromArray(basePositionRotationOrScale)).asArray();
-                        }
+                const q: Quaternion = (<any>babylonTransformNode).rotationQuaternion;
```
bghgary, line 976: I guess the union type didn't actually help right?

It's not perfect, but I would probably do this to avoid using `any`. One of the problems with using `any` is that you can very easily spell something wrong. Same for the other properties below.

```suggestion
                const q = (babylonTransformNode as TransformNode).rotationQuaternion;
```

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
```diff
@@ -943,37 +965,37 @@ export class _GLTFAnimation {
...
+                    _GLTFUtilities._GetRightHandedQuaternionArrayFromRef(basePositionRotationOrScale);
+                    if (!babylonTransformNode.parent) {
+                        basePositionRotationOrScale = Quaternion.FromArray([0, 1, 0, 0]).multiply(Quaternion.FromArray(basePositionRotationOrScale)).asArray();
                     }
-                } else {
-                    basePositionRotationOrScale = Quaternion.Identity().asArray();
                 }
             } else {
-                basePositionRotationOrScale = babylonTransformNode.rotation.asArray();
+                const r: Vector3 = (<any>babylonTransformNode).rotation;
```
bghgary, line 985:
```suggestion
                const r = (babylonTransformNode as TransformNode).rotation;
```

packages/dev/serializers/src/glTF/2.0/glTFAnimation.ts
pandaGaume: Yes you're right, the problem came from we use other object which are not `TransformNode`, `Camera` and `Light` for instance.
However these objects may have, down the hierarchy, similar properties such `rotation` or `rotationQuaternion` or `position` and Ideally they should share a common interface which encompass these properties.
Type assertion will not change the game, but i'm agree that we may better put something else that `any`. Maybe introduce an interface instead of union type ?

deltakosh:
@bghgary can we merge this one?

bghgary:
Not yet. I've marked it as draft.

pandaGaume:
Just tell me what you want me to change, then we can close it.

bghgary:
I think I would remove the `TransformableType` as it doesn't really help compared to using `Node` directly. And then just cast to `TransformNode` in the `_GetBasePositionRotationOrScale` function to avoid misspelling.

bghgary (approved):

pandaGaume:
@bghgary modifications done.

PR #12689, opened 2022-06-28
[Build] Babylon eslint plugin (step 1 - typedoc)

RaananW:
As the tsdoc eslint task is not very flexible, we have to adjust it to make it work with our repository.

This PR adds 2 new rules:

1) Babylon/syntax which will notify if a comment is not formatted correctly, taking our quircks into account.
2) babylon/available - text-based codedoc inspector for methods
3) babylon/existing - typescript-based codedoc inspector.

The difference between 2 and 3 is that 3 is much (MUCH) more precise and correct, but takes a long time to run, especially in such a large repository. The 2nd step is quick but prone to errors, as it is basically a text-based parser. 

Further eslint rules will be added to this package when needed.


sebavan:
@RaananW  I guess we can close this one ???

RaananW:
Closing and will open a new branch with a fix.

PR #12693, opened 2022-06-29
Added code to track buttons pressed while meta key is active on MacOS

PolygonalSun:
There was an issue found on the forum where `keyup` events for keys weren't firing while the Meta key (Command) was pressed for all browsers on MacOS.  This has led to a scenario where said buttons weren't being properly treated as released.  Upon investigation, it has been determined that this issue occurs at an OS level and therefore, there isn't much we can do about it (https://stackoverflow.com/questions/11818637/why-does-javascript-drop-keyup-events-when-the-metakey-is-pressed-on-mac-browser).  This PR contains a workaround so that if BJS is being run in a browser on MacOS, any buttons pressed while the Meta key is active will be tracked and have `keyup` events fired for them when the Meta key is released.  All other platforms will work as before without this tracking code.

Forum Link: https://forum.babylonjs.com/t/modifier-keyboard-shortcuts-with-devicesourcemanager-copy-to-clipboard/31743/7

### DIFF
--- a/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
+++ b/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
@@ -21,7 +21,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
     private _keyboardActive: boolean = false;
     private _pointerActive: boolean = false;
     private _elementToAttachTo: HTMLElement;
-    private _metaKeys: Set<number>;
+    private _metaKeys: Array<number>;
     private readonly _engine: Engine;
     private readonly _usingSafari: boolean = Tools.IsSafari();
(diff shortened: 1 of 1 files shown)


sebavan (approved):

sebavan:
This is a good candidate for your test suite

barroij:
Hi people,

This PR break the ability to use BabylonJS for testing purpose within a Node environment (using the NullEngine) as creating the scene leads to the error `ReferenceError: navigator is not defined`

it is caused by the line `private readonly _usingMacOS: boolean = /(Mac|iPhone|iPod|iPad)/i.test(navigator.platform);`

It used to work fine in 5.10.0 and it breaks in 5.50.0

Possible workaround for me meanwhile ; `global.navigator = global.navigator || { platform: '' }`

sebavan:
Thanks a ton for the report I ll fix it in a sec

PR #12696, opened 2022-06-29
DeviceInputSystem: Fixed Passive Support Check to prevent Violation Warning

PolygonalSun:
There was a warning found by a user on the forums that, in Verbose mode, a Violation warning was occurring:
image
This warning was happening because the `wheel` event listener didn't have the "passive" option set to true when able.  After closer investigation, it was found that the code used to determine if passive was supported was always returning false, causing this warning to occur.  This PR modifies the logic to fix this.

Forum link: https://forum.babylonjs.com/t/bjs-console-warnings-when-using-cleargizmoonemptypointerevent/31447/8

### DIFF
--- a/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
+++ b/packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
@@ -104,7 +104,7 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
             throw `Unable to find device ${DeviceType[deviceType]}`;
         }
 
-        if (deviceType >= DeviceType.DualShock && deviceType <= DeviceType.DualSense && navigator.getGamepads) {
+        if (deviceType >= DeviceType.DualShock && deviceType <= DeviceType.DualSense) {
             this._updateDevice(deviceType, deviceSlot, inputIndex);
         }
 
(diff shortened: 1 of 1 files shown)


sebavan (approved):

packages/dev/core/src/DeviceInput/InputDevices/webDeviceInputSystem.ts
```diff
@@ -592,13 +592,11 @@ export class WebDeviceInputSystem implements IDeviceInputSystem {
         const noop = function () {};
 
         try {
-            const options: object = {
-                passive: {
-                    get: function () {
-                        passiveSupported = true;
-                    },
+            const options = Object.defineProperty({}, "passive", {
```
sebavan, line 595: Why changing it to this ? just as an FYI for me :-)

PolygonalSun: When I tested the original code, passiveSupported was always coming back as false.  I had found this [documentation](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection) that used Object.defineProperty and when I tested that, it evaluated properly so I subbed it in.

deltakosh (approved):

PR #12702, opened 2022-07-01
Initial Device Input Test Pattern Setup: Unit Tests

PolygonalSun:
This PR contains a set of tests to verify the functionality of the DeviceSource and DeviceSourceManager objects.  These tests will serve as a baseline for other tests coming up in the future.

This also contains a small fix to the DeviceSourceManager when a user tries to query for a device that has no active entries.


packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+            return 0;
+        }
+    }
+    isDeviceAvailable(deviceType: DeviceType): boolean {
+        return deviceType === DeviceType.Mouse || deviceType === DeviceType.Touch;
+    }
+    dispose(): void {
+        // Do nothing
+    }
+}
```
bghgary, line 30: What is the goal of this class in terms of functionality? Right now, it looks like the `pollInput` is being mocked to always be in a state where left mouse button is down? I'm assuming this will evolve as you add more tests?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    };
+
+    public dispose(): void {}
+}
+
+describe("DeviceSource", () => {
+    it("should exist", () => {
+        const deviceInputSystem = new MockDeviceInputSystem(null, null, null);
+        const mouseSource = new DeviceSource(deviceInputSystem, DeviceType.Mouse, 0);
+        expect(mouseSource).not.toBeNull();
```
bghgary, line 54: Is it possible for `mouseSource` to be null? Why test this?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    it("can use onInputChangedObservable", () => {
+        expect.assertions(2);
+        let resultEvent: Nullable<IWheelEvent | IPointerEvent> = null;
+        let onInputChanged = null;
+        // Create DeviceSource and add to its observable
+        const deviceInputSystem = new MockDeviceInputSystem(null, null, onInputChanged);
+        const mouseSource = new DeviceSource(deviceInputSystem, DeviceType.Mouse, 0);
+        onInputChanged = (deviceType: DeviceType, deviceSlot: number, eventData: IUIEvent) => {
+            mouseSource.onInputChangedObservable.notifyObservers(eventData as IPointerEvent);
+        };
```
bghgary, line 74: I'm not sure this is doing. I think it is simulating a mouse event, but I would expect this to be in the mock class with a public method to simulate input things (e.g., clicking a mouse button, etc.)

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        engine._deviceSourceManager = new MockInternalDeviceSourceManager() as InternalDeviceSourceManager;
+    });
+
+    afterEach(function () {
+        engine?.dispose();
+    });
+
+    it("should exist", () => {
+        const deviceSourceManager = new DeviceSourceManager(engine!);
+        expect(deviceSourceManager).not.toBe(null);
```
bghgary, line 101: Ditto. Is it even possible for this to be null? Maybe check for `engine!._deviceSourceManager` instead.

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        expect(engine?._deviceSourceManager?._refCount).toBe(1);
```
bghgary, line 102: I would recommend using `engine!` in most places. It is guaranteed to exist right?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+
+    it("can use getDeviceSource", () => {
+        const deviceSourceManager = new DeviceSourceManager(engine!);
+        const deviceInputSystem = new MockDeviceInputSystem(null, null, null);
+        const touchSource = new DeviceSource(deviceInputSystem, DeviceType.Touch, 0);
+        const touchSource2 = new DeviceSource(deviceInputSystem, DeviceType.Touch, 1);
+        const nullSource = deviceSourceManager.getDeviceSource(DeviceType.Touch, 0);
+        expect(nullSource).toBe(null);
```
bghgary, line 111: nit: Maybe put a newline before this to separate it from the top section

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+
+    it("can use getDeviceSources", () => {
+        const deviceSourceManager = new DeviceSourceManager(engine!);
+        const deviceInputSystem = new MockDeviceInputSystem(null, null, null);
+        const touchSource = new DeviceSource(deviceInputSystem, DeviceType.Touch, 0);
+        const touchSource2 = new DeviceSource(deviceInputSystem, DeviceType.Touch, 1);
+        const emptyArray = deviceSourceManager.getDeviceSources(DeviceType.Touch);
+
+        expect(emptyArray.length).toBe(0);
```
bghgary, line 135: nit: Maybe group these together so that it's easier to read

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        deviceSourceManager._addDevice(touchSource);
+        deviceSourceManager._addDevice(touchSource2);
+
+        const touchArray = deviceSourceManager.getDeviceSources(DeviceType.Touch);
+        expect(touchArray.length).toBe(2);
```
bghgary, line 141: Maybe should also check if the contents are correct

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        // Do nothing
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    public readonly unregisterManager = (manager: IObservableManager): void => {
+        // Do nothing
+    };
+
+    public dispose(): void {}
+}
```
bghgary, line 48: Is this necessary? Can we not use the real `InternalDeviceSourceManager`?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
bghgary: Also, are you unable to use the [mock functionality from Jest](https://jestjs.io/docs/es6-class-mocks)?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        const touchArray = deviceSourceManager.getDeviceSources(DeviceType.Touch);
+        expect(touchArray.length).toBe(2);
+    });
+
+    it("can use onDeviceConnectedObservable", () => {
+        expect.assertions(1);
+        const deviceSourceManager = new DeviceSourceManager(engine!);
+        const deviceInputSystem = new MockDeviceInputSystem(null, null, null);
+        const touchSource = new DeviceSource(deviceInputSystem, DeviceType.Touch, 0);
```
bghgary, line 148: This is a bit confusing. `new DeviceSourceManager` will create an `InternalDeviceSourceManager` which will create a real `DeviceInputSystem` right? Then, this is manually creating a `DeviceSource` with a mock `deviceInputSystem`? Am I reading this right?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+
+    it("can use onDeviceDisconnectedObservable", () => {
+        expect.assertions(2);
+        const deviceSourceManager = new DeviceSourceManager(engine!);
+        const deviceInputSystem = new MockDeviceInputSystem(null, null, null);
+        const touchSource = new DeviceSource(deviceInputSystem, DeviceType.Touch, 0);
+        let observableSource = null;
+
+        deviceSourceManager.onDeviceDisconnectedObservable.add((eventData) => {
```
bghgary, line 167: `eventData` doesn't make sense here (multiple places)

```suggestion
        deviceSourceManager.onDeviceDisconnectedObservable.add((deviceSource) => {
```

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+            observableSource = eventData;
+        });
+
+        deviceSourceManager._addDevice(touchSource);
+        deviceSourceManager._removeDevice(DeviceType.Touch, 0);
+
+        const nullSource = deviceSourceManager.getDeviceSource(DeviceType.Touch, 0);
+
+        expect(observableSource).toEqual(touchSource);
+        expect(nullSource).toBe(null);
```
bghgary, line 177: nit: I recommend putting the variable declaration as close to the usage of it as much as possible.

```suggestion
        expect(observableSource).toEqual(touchSource);

        const nullSource = deviceSourceManager.getDeviceSource(DeviceType.Touch, 0);
        expect(nullSource).toBe(null);
```

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        let observableEvent = null;
+
+        mouseSource.onInputChangedObservable.add((eventData) => {
+            observableEvent = eventData;
+        });
+
+        const mouseEvent: IUIEvent = { pointerId: 0, type: "pointerdown", pointerType: "mouse", clientX: 1, clientY: 1 } as IPointerEvent;
+
+        // Try to call _onInputChanged without any added DeviceSources
+        deviceSourceManager._onInputChanged(DeviceType.Mouse, 0, mouseEvent);
```
bghgary, line 194: This is a strange why to test this. We should hooking in only at the deviceInputSystem level and not anywhere else. Let's chat a bit offline.

bghgary (changes requested):

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: Since I'm going to mock the implementation of the WebDeviceInputSystem, I should be able to use the standard InternalDeviceSourceManager so this will be removed.

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: It could be null but there should no scenario where this occurs so I'm going to remove this test

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I've added a function to create an event and send it through the input system.  It should move all the way throught he onInputChanged pipeline.

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: That makes sense as a change.  I'll make that change

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I'll remove all instances of engine?

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I'll separate and clean up this section2

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I'll clean up this section as well

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I can check that the contents match what I put into the DSM

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: This section will be simplified and cleaned up

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I can make that change

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I'll clean up this section

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: I'll change this as per our offline conversation

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,265 @@
...
+                },
+                changeInput(deviceType: DeviceType, deviceSlot: number, inputIndex: number, currentState: number): void {
+                    const evt = DeviceEventFactory.CreateDeviceEvent(deviceType, deviceSlot, inputIndex, currentState, this as unknown as IDeviceInputSystem);
+                    onInputChanged(deviceType, deviceSlot, evt);
+                },
+                dispose(): void {
+                    // Do Nothing
+                },
+            }
+        });
```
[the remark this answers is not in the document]
PolygonalSun: For this, I'm going to create a class that will have the web implmentation as a static reference and when I start work on the native side, I'll create a native implementation that will be referenced there too.

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,265 @@
...
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                isDeviceAvailable(deviceType: DeviceType): boolean {
+                    return true;
+                },
+                connectDevice(deviceType: DeviceType, deviceSlot: number): void {
+                    onDeviceConnected(deviceType, deviceSlot);
+                },
+                disconnectDevice(deviceType: DeviceType, deviceSlot: number): void {
+                    onDeviceDisconnected(deviceType, deviceSlot);
+                },
```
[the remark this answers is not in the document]
PolygonalSun: I've added the checks to all three of these my next push.  connectDevice and disconnectDevice are pretty much the same as register/unregisterDevice

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        expect(mouseSource.deviceType).toBe(DeviceType.Mouse);
+        expect(mouseSource.deviceSlot).toBe(0);
+        expect(mouseSource.onInputChangedObservable).not.toBeUndefined();
+    });
+    it("can poll with getInput", () => {
+        const mouseSource = new DeviceSource(wdis!, DeviceType.Mouse, 0);
+        const leftClick = mouseSource.getInput(PointerInput.LeftClick);
+        expect(leftClick).toBe(0);
+    });
+});
```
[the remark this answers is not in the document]
PolygonalSun: pollInput (and by extension getInput) should now get input changes made via changeInput

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        engine = new NullEngine();
+    });
+
+    afterEach(function () {
+        engine!.dispose();
+    });
+
+    it("should exist", () => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const deviceSourceManager = new DeviceSourceManager(engine!);
```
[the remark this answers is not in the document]
PolygonalSun: Changed

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -8,134 +8,179 @@ import { NullEngine } from "core/Engines/nullEngine";
...
-        });
-    });
+    const onDeviceConnected = (_deviceType: DeviceType, _deviceSlot: number) => {};
+    const onDeviceDisconnected = (_deviceType: DeviceType, _deviceSlot: number) => {};
+    const onInputChanged = (_deviceType: DeviceType, _deviceSlot: number, _eventData: IUIEvent) => {};
 
     beforeEach(() => {
         engine = new NullEngine();
         wdis = new WebDeviceInputSystem(engine, onDeviceConnected, onDeviceDisconnected, onInputChanged);
     });
```
bghgary, line 126: \
nit:
```suggestion
    beforeEach(() => {
        engine = new NullEngine();
        wdis = new WebDeviceInputSystem(engine, () => {}, () => {}, () => {});
    });
```

bghgary (commented):
Looks good. Minor things now.

bghgary (approved):

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -8,134 +8,179 @@ import { NullEngine } from "core/Engines/nullEngine";
...
 import type { Nullable } from "core/types";
 
-jest.mock("core/DeviceInput/InputDevices/webDeviceInputSystem");
+interface ITestDeviceInputSystem {
+    connectDevice(deviceType: DeviceType, deviceSlot: number, numberOfInputs: number): void;
+    disconnectDevice(deviceType: DeviceType, deviceSlot: number): void;
+    changeInput(deviceType: DeviceType, deviceSlot: number, inputIndex: number, currentState: number): void;
+}
+
+class TestDeviceInputSystem implements ITestDeviceInputSystem{
```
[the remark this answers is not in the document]
PolygonalSun: Fixed

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        this._onDeviceConnected = () => {};
+        this._onDeviceDisconnected = () => {};
+        this._onInputChanged = () => {};
+    }
+}
+
+jest.mock("core/DeviceInput/InputDevices/webDeviceInputSystem", () => {
+    return {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        WebDeviceInputSystem: jest.fn().mockImplementation((
```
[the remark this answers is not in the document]
PolygonalSun: Nope.  TBH, I thought that this way of doing it was a bit weird but I found a blog that used a similar way and it worked.  Quotations work too so I'm adding them.

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
PolygonalSun: Fixed

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
```diff
@@ -8,134 +8,179 @@ import { NullEngine } from "core/Engines/nullEngine";
...
-        const deviceSourceManager = new DeviceSourceManager(engine!);
+        new DeviceSourceManager(engine!);
         expect(engine!._deviceSourceManager).not.toBe(null);
         expect(engine!._deviceSourceManager!._refCount).toBe(1);
     });
 
     it("can use getDeviceSource", () => {
         const deviceSourceManager = new DeviceSourceManager(engine!);
-        const deviceInputSystem = engine!._deviceSourceManager!._deviceInputSystem as any;
+        const deviceInputSystem = engine!._deviceSourceManager!._deviceInputSystem as unknown as ITestDeviceInputSystem;
```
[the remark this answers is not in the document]
PolygonalSun: Added a helper function

packages/dev/core/test/unit/DeviceInput/babylon.deviceInput.test.ts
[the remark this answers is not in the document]
PolygonalSun: I just ran Prettier and it keeps removing them so I guess they're just staying off.

PR #12707, opened 2022-07-06
[DRAFT] Create mrtkv3 button

chinezenwosu-ms:
This is currently a work in progress.

TODOs
- [ ] Fix buggy click and hover animation
- [ ] Fix positioning of button icon
- [ ] Figure out how to handle versioning between this button and the TouchHolographic button. Also, rename button as needed


deltakosh:
cc @RaananW to see if we can merge?

RaananW:
This PR is in draft mode for a reason (see description), so I wouldn't want to merge it just yet

chinezenwosu-ms:
Hi @RaananW I'm trying to figure out the best way to version or name this button. As more buttons or components are created in the future, things might get very complicated, and it would be confusing to users which buttons to use. My team and I might be creating more MRTKV3 components and I would appreciate any help I can get on how subsequent new versions of these components should be represented in code.

RaananW:
> 

That's a good question.
Regarding structure I think it will be great if you create a V3 (or MRTK3) directory and put the component(s) there. However, this will present an issue with our UMD packages if the component has the same name as the V2 component. It would need to be exported as a different component (or in a different namespace, which will be less than ideal).
I'll get back to you on this one very soon.

RaananW:
So - did some thinking (and I am so sorry for the late reply).

Let's move everything MRTKv3-related to a different directory inside the controls' directory called MRTKv3. When exporting from the index.ts file of this directory, you will need to import and export as, i.e.:

```
import { TouchHolographicButton } from './touchHolographicButton';
export { TouchHolographicButton as TouchHolographicButtonV3 };
```

This way we can have 2 different implementations of the touch holographic buttons, and the UMD will have V3 at the end of it. When importing using es6 one could import the holographic button directly from either V3 or the V2 implementation.

I hope it makes sense! let me know if you need further explanation or any help with that.

sebavan:
@chinezenwosu-ms any updates on this PR ?

packages/dev/gui/src/3D/controls/touchHolographicV3Button.ts
```diff
@@ -0,0 +1,1028 @@
...
+import { Vector3 } from "core/Maths/math.vector";
+
+/**
+ * Class used to create the mrtkv3 button
+ */
+export class TouchHolographicV3Button extends TouchButton3D {
+    /**
+     * Base Url for the frontplate model.
+     */
+    public static FRONTPLATE_MODEL_BASE_URL = "https://raw.githubusercontent.com/chinezenwosu-ms/TempAssets/main/GLB/";
```
RaananW, line 40: We can upload the assets to our assets library (https://github.com/BabylonJS/Assets). It will be available on our CDN.

packages/dev/gui/src/3D/controls/touchHolographicV3Button.ts
```diff
@@ -0,0 +1,1028 @@
...
+
+    /**
+     * File name for the backglow model.
+     */
+    public static BACKGLOW_MODEL_FILENAME = "mrtk-fluent-button.glb";
+
+    /**
+     * Base Url for the innerquad model.
+     */
+    public static INNERQUAD_MODEL_BASE_URL = "https://raw.githubusercontent.com/chinezenwosu-ms/TempAssets/main/GLB/";
```
RaananW, line 70: Same as above - let's upload this to the assets lib.

packages/dev/gui/src/3D/controls/touchHolographicV3Button.ts
```diff
@@ -0,0 +1,1028 @@
...
+        if (this._tooltipTextBlock) {
+            this._tooltipTextBlock.text = text;
+        }
+    }
+
+    public get tooltipText() {
+        if (this._tooltipTextBlock) {
+            return this._tooltipTextBlock.text;
+        }
+        return null;
```
RaananW, line 250:
```suggestion
        return this._tooltipTextBlock?.text;
```

packages/dev/gui/src/3D/controls/touchHolographicV3Button.ts
```diff
@@ -0,0 +1,1028 @@
...
+     */
+    public get shareMaterials(): boolean {
+        return this._shareMaterials;
+    }
+
+    /**
+     * Sets whether the backplate is visible or hidden. Hiding the backplate is not recommended without some sort of replacement
+     */
+    public set isBackplateVisible(isVisible: boolean) {
+        if (this.mesh && !!this._backMaterial) {
```
RaananW, line 347: nit - is the `!!` needed?

packages/dev/gui/src/3D/controls/touchHolographicV3Button.ts
```diff
@@ -0,0 +1,1028 @@
...
+        return tn;
+    }
+
+    private _createBackPlate(scene: Scene) {
+        const backPlateMesh = CreateBox(`${this.name}_backPlate`, {}, scene);
+        backPlateMesh.isPickable = false;
+        backPlateMesh.visibility = 0;
+        backPlateMesh.scaling.z = 0.2;
+
+        SceneLoader.ImportMeshAsync(undefined, TouchHolographicV3Button.BACKPLATE_MODEL_BASE_URL, TouchHolographicV3Button.BACKPLATE_MODEL_FILENAME, scene).then((result) => {
```
RaananW, line 593: Does it make sense to track if the mesh initialization is done? i.e. when the button is actually ready to be rendered?

RaananW: Or maybe use async/await and only return when the mesh is ready? (if this part doesn't make sense i am fine with keeping the state as it is :-))

RaananW (commented):
Added a few comments for the code, will fully review when the move is done.

tayomadein:
Hello @RaananW @sebavan, I am picking up the work on this PR and I will work through your comments and share.

tayomadein:
Hello @RaananW @sebavan I was unable to continue working on the same branch + PR, so I have moved the work here https://github.com/BabylonJS/Babylon.js/pull/12989 and implemented some of your feedback.

I moved the button control to the MRTKv3 folder, updated the nits and created a PR for the assets. We can close this PR once the comments and have resolved and continue working on the new PR, let me know what you think about this.

packages/dev/gui/src/3D/controls/touchHolographicV3Button.ts
tayomadein: Is this comment related to only the backplate mesh or all the meshes used to create the button?

sebavan:
closing to only keep https://github.com/BabylonJS/Babylon.js/pull/12989 open

PR #12726, opened 2022-07-08
Create screenshot: Fix resources not recreated with the right sizes

Popov72:
See https://forum.babylonjs.com/t/babylonjs-5-0-5-5-5-ios-15-4-safari-15-4-issues/30195/27


deltakosh (approved):

Popov72:
It is the basis loader test which is failing, so likely not coming from my PR.

deltakosh:
cc @RaananW

bghgary:
We just merged https://github.com/BabylonJS/Babylon.js/pull/12662 which is likely the reason. But the CI was fine for that PR?

bghgary:
Oh, I know why. You have to make sure to merge latest after that change goes in because the config needs to point to the correct Basis files.

bghgary:
Actually, never mind. It should still work since we kept the old Basis files at the same location. The visualization test for the basis loader still works on the playground: https://playground.babylonjs.com/#4RN0VF#102.

PR #12728, opened 2022-07-11
SSR improvements

Mannns:
As I explained in this thread : https://forum.babylonjs.com/t/ssr-when-i-rotated-a-mesh-the-reflection-shows-wrong-and-connected-to-the-mesh/30915/17, I worked on a new version of SSR.

I changed the main algorithm (intersection research) : I'm using 2D ray marching.
I also changed some things:
 -   I corrected the final shading (little problem around the Fresnel formula that impacts the rendering of reflection)
 -   I use linear space reflectivity
 -   I let the opportunity to the user to apply a cubeTexture (Skybox or Probe) as a fallback
 -   I use reflectivity as a vec3 to allow reflection rendering with metallic-roughness workflow

I also defined a "backwardCompatibility" flag to avoid regression. This flag allows the user to use the first version of the algorithm instead of mine.


deltakosh:
This is massive! thanks a ton

Would you be ok to also update the doc page?

deltakosh:
cc @julien-moreau also for review

Mannns:
> This is massive! thanks a ton
> 
> Would you be ok to also update the doc page?

Yes I will do it

julien-moreau:
That's awesome @Mannns !!!!
LGTM I'm just getting the branch to test it :)

packages/dev/core/src/PostProcesses/screenSpaceReflectionPostProcess.ts
```diff
@@ -11,40 +11,153 @@ import { ScreenSpaceReflectionsConfiguration } from "../Rendering/screenSpaceRef
...
-     * Gets or sets the current reflection strength. 1.0 is an ideal value but can be increased/decreased for particular results.
+     * Gets or sets the step size used to iterate until the effect finds the color of the reflection's pixel. Typically in interval [0.1, 1.0]. Default value is 1.0.
+     * For backward compatibility only.
      */
     @serialize()
-    public strength: number = 1;
+    public step: number = 1.0;
+
+    private _enableSmoothReflections: boolean = false;
+    private _reflectionSamples: number = 64;
```
julien-moreau, line 49: "reflectionSamples" is missing getter and setter. We need it for backward compatibility

packages/dev/core/src/PostProcesses/screenSpaceReflectionPostProcess.ts
```diff
@@ -11,40 +11,153 @@ import { ScreenSpaceReflectionsConfiguration } from "../Rendering/screenSpaceRef
...
     /**
-     * Gets or sets the current reflection strength. 1.0 is an ideal value but can be increased/decreased for particular results.
+     * Gets or sets the step size used to iterate until the effect finds the color of the reflection's pixel. Typically in interval [0.1, 1.0]. Default value is 1.0.
+     * For backward compatibility only.
      */
     @serialize()
-    public strength: number = 1;
+    public step: number = 1.0;
+
+    private _enableSmoothReflections: boolean = false;
```
julien-moreau, line 48: "enableSmoothReflections" is missing getter and setter. We need it for backward compatibility :(

julien-moreau (changes requested):

packages/dev/core/src/PostProcesses/screenSpaceReflectionPostProcess.ts
Mannns: getters et setters are back! Thanks for your feedback!

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+    vec3 reflected = normalize(reflect(unitPosition, unitNormal)); // incident direction = unit position in camera space
+
+    // *************** Compute reflection info  ***************
+    ReflectionInfo info;
+
+    // hash(position) represents a random vector3, jitt represents a bias to simulate roughness (light deviation)
+    vec3 jitt = mix(vec3(0.0), hash(texture2D(positionSampler, vUV).xyz), roughness) * roughnessFactor;
+
+    #ifdef RIGHT_HANDED_SCENE
+        if (position.z < -distanceFade || distanceFade == 0.0){ // no need to compute reflection, the point we are evaluating is further than the distanceFade
```
Popov72, line 456: The left handed scene case is exactly the same save for the check, so you should do something like this instead:

```glsl
#ifdef RIGHT_HANDED_SCENE
    if (position.z < -distanceFade || distanceFade == 0.0) {
#else
    if (position.z > distanceFade || distanceFade == 0.0) {
#endif
[common code]
```

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+                info.coords = vUV;
+                info.visibility = 0.0;
+                info.miss = true;
+                info.visibilityBackup = 1.0;
+            #else
+                gl_FragColor = texture2D(textureSampler, vUV);
+                return;
+            #endif
+        } else {
+            vec2 texSize = gl_FragCoord.xy / vUV;
```
Popov72, line 467: This computation will fail when one of `vUV.x` or `vUV.y` is 0, and I think you won't always have the same result from pixel to pixel because of precision problems.

You should either pass the width/height as uniforms or use `textureSize(textureSampler)` instead.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+
+        }
+    #endif
+
+    // *************** Get reflection color ***************
+
+    vec3 reflectedColor;
+    
+    #if defined(BACKUP_TEXTURE_SKYBOX) || defined(BACKUP_TEXTURE_PROBE)
+        if (backupOnlyWhenTooSpecular && dot(spec, vec3(1.0))/3.0 > 0.9) {
```
Popov72, line 493: Do we really need this?

That's a new constant that the user will probably want to tweak, so more parameters...

Also, I think the parameter is not very clear (for me at least), something like `forceBackupOnlyWhenTooSpecular` would be clearer.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+    //  *********************** Shading *******************************
+
+    // Fresnel
+    // "The specular map contains F0 for dielectrics and the reflectance value for raw metal"
+    vec3 F0 = spec;
+  
+    vec3 reflectionCoeff = fresnelSchlick(max(dot(unitNormal, -unitPosition), 0.0), F0) // https://lettier.github.io/3d-game-shaders-for-beginners/fresnel-factor.html
+                            * (info.visibility + info.visibilityBackup); 
+    reflectionCoeff = clamp(vec3(pow(reflectionCoeff.x * strength, falloffExponent - 2.0), pow(reflectionCoeff.y * strength, falloffExponent - 2.0), pow(reflectionCoeff.z * strength, falloffExponent - 2.0)), 0.0, 1.0); // -2 to compensate for default value (3.0)
+
```
Popov72, line 534: Why not `pow(reflectionCoeff * strength, fallOffExponent - 2.0)`?

Also why compensating for the `3` value in the shader? It's not a problem to have a default value of 3 as long as it is documented, and the user can change to smaller or any value they want during the setup. So, `pow(reflectionCoeff * strength, fallOffExponent)` would be better and more logical to me.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+{
+    #ifdef SSR_SUPPORTED
+
+    // *************** Get data from samplers ***************
+
+    vec4 original = texture2D(textureSampler, vUV);
+    vec3 spec = toLinearSpace(texture2D(specularSampler, vUV).rgb);
+
+    if (dot(spec, vec3(1.0)) <= 0.0){
+        gl_FragColor = texture2D(textureSampler, vUV); // no reflectivity, no need to compute reflection
```
Popov72, line 435: You could simply use `original` here.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+
+    #ifdef RIGHT_HANDED_SCENE
+        if (position.z < -distanceFade || distanceFade == 0.0){ // no need to compute reflection, the point we are evaluating is further than the distanceFade
+            #if defined(BACKUP_TEXTURE_SKYBOX) || defined(BACKUP_TEXTURE_PROBE)
+                info.coords = vUV;
+                info.visibility = 0.0;
+                info.miss = true;
+                info.visibilityBackup = 1.0;
+            #else
+                gl_FragColor = texture2D(textureSampler, vUV);
```
Popov72, line 463: You can simply use `original` here instead of the lookup.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+        if (!info.miss){
+            reflectedColor += texture2D(textureSampler, info.coords).xyz * info.visibility;
+        }
+    #else 
+        if (backupOnlyWhenTooSpecular && dot(spec, vec3(1.0))/3.0 > 0.9) {
+            info.visibility = 0.0;
+            info.visibilityBackup = 0.0;
+        }
+        if (info.miss){
+            gl_FragColor = texture2D(textureSampler, vUV);
```
Popov72, line 518: You can simply use `original` here.

Popov72 (commented):
Very good job!

My main complaint is with the usage of arbitrary constants. We should try to limit at the minimum those constants in the code and clearly document what they are/why we use them when we do use them. And as much as possible, the user should be able to set their values to avoid tying our implementation to some specific cases.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
Mannns: To me, choosing 3 as default value doesn't make sense. However, it was the default value of the first version of SSR and I choose not to change this default value to avoid regression..
I guess it was used to compensate for the mistake in FresnelSchlick call (in the first version this value is always 1.0)

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
Mannns: It may be needed when the user defines mirror like materials. SSR and probe don't match well, and it's prettier to use only one method rather than a mix of methods in this case.

What do you think about letting the user choosing the value, instead of using a boolean ?

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+                * clamp(dot(normalize( texture2D(normalSampler, uv).xyz), normalize( texture2D(normalSampler, (currFrag + increment.yx * resolution)/texSize).xyz)), 0.0, 1.0)
+                * clamp(dot(normalize( texture2D(normalSampler, uv).xyz), normalize( texture2D(normalSampler, (currFrag - increment.yx * resolution)/texSize).xyz)), 0.0, 1.0)
+              
+                * (1.0 - max (dot(-normalize(hitCoordVS), normalize(dirVS)), 0.0)) // to fade out the reflection as the reflected direction point to the camera's position (hit behind the camera)
+                * (1.0 - currSearch) // the reflection should be sharper when near from the starting point
+                * (1.0 - clamp (abs(hitCoordVS.z / distanceFade), 0.0, 1.0)) // to fade out the reflection near the distanceFade
+                * (1.0 - clamp (deltaDepth/maxTol, 0.0, 1.0)) // since the hit point is not always precisely found, we fade out the reflected color if we aren't precise enough 
+                * clamp(1.0 - (dCoordScreen.x + dCoordScreen.y), 0.0, 1.0); // to fade out the reflection near the edge of the screen
```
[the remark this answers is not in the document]
Mannns: I added this factors to deal with this artifacts : https://stackoverflow.com/questions/65760242/fix-artifacts-in-screen-space-reflections
image

And, I think the result is prettier with this factors (and using the visibility * 2.0 to compensate)
image
image

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
Popov72: Yes, let's allow the user to change the 0.9 value instead of having `backupOnlyWhenTooSpecular`.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
Popov72: I'm not sure to understand, in the existing version the line is `float reflectionMultiplier = clamp(pow(spec * strength, reflectionSpecularFalloffExponent) * screenEdgefactor * reflected.z, 0.0, 0.9);` so we are using `reflectionSpecularFalloffExponent` directly?

Regarding the backward compatibility, I wonder if it would not be easier to simply separate the implementations and have a `ScreenSpaceReflection2` class, the same way we have SSAO / SSAO2... That way we don't change anything of the existing typescript / shader code and we create new files instead.

What do you think @deltakosh / @sebavan?

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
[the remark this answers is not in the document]
Popov72: I wonder in what measure these artifacts come from how the ray marching is done compared to using a robust DDA algorithm like the one described [here](https://jcgt.org/published/0003/04/04/paper.pdf) and used [here](https://github.com/kode80/kode80SSR) and [there](http://casual-effects.blogspot.com/2014/08/screen-space-ray-tracing.html)...

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+    if (hit0 == 0.0){
+        #if defined(BACKUP_TEXTURE_SKYBOX) || defined(BACKUP_TEXTURE_PROBE)
+            info.visibilityBackup = 1.0;
+        #endif
+
+    } else {
+        info.miss = false;
+
+        if (length(rayDepth - hitCoordVS) < thickness * 0.05 //){
+                   || dot(normalize(dirVS), normalize(texture2D(normalSampler, uv).xyz)) > -0.085 ){
```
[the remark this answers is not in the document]
Mannns: I will remove both tests, they seem useless since I added other visibility factors

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
Mannns: To me, reflectionSpecularFalloffExponent and strength are used for artistic purpose only. Thus, default value shouldn't affect the reflection coefficient based on Fresnel formula.

The Fresnel formula already takes into account the specularity (through F0). 
The existing version miss a minus before the unitPosition when calling fresnelSchlick function. So, this function always returns 1. This is why, I guess, the reflection multiplier is based on spec and uses 3.0 as reflectionSpecularFalloffExponent default value. 


Regarding backwardCompatibility, I defined the backwardCompatibility flag as mentioned in this thread : https://forum.babylonjs.com/t/ssr-when-i-rotated-a-mesh-the-reflection-shows-wrong-and-connected-to-the-mesh/30915/19. But I can create a new class with new files instead.

Popov72: Yes, the missing minus sign is a bug and the default of 3 value is not very sound.

That's why I think a separate implementation would be better. That would also be better for tree shaking.

Let's see what @deltakosh think about this (I think @sebavan is in holidays).

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+                info.visibilityBackup = 1.0;
+            #endif
+            return info;
+        }
+    #endif
+
+    // Calculate the start and end point of the reflection ray in screen space.
+    vec4 startSS = projection * startVS; // Project to screen space.
+    startSS.xyz /= startSS.w; // Perform the perspective divide.
+    startSS.xy = startSS.xy * 0.5 + vec2(0.5); // Convert from clip space to texture space.
```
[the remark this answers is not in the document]
Mannns: Is the WEBGPU define already set somewhere ? or do I have to set it in the SSR class ?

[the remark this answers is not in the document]
Popov72: It is set automatically by the engine when in WebGPU mode, you only have to check for it as in the code above.

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,338 @@ void main()
...
+
+    vec4 original = texture2D(textureSampler, vUV);
+    vec3 spec = toLinearSpace(texture2D(reflectivitySampler, vUV).rgb);
+
+    if (dot(spec, vec3(1.0)) <= 0.0){
+        gl_FragColor = original; // no reflectivity, no need to compute reflection
+        return;
+    }
+
+    float roughness = 1.0 - texture2D(reflectivitySampler, vUV).a;
```
Popov72, line 435: You should set `vec4 spec = toLinearSpace(texture2D(reflectivitySampler, vUV));` and do `roughness = 1.0 - spec.a` to save a texture lookup (`toLinearSpace` won't change the `alpha` component).

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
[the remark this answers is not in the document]
Popov72: Can you share the PG links you used for these screenshots?

I'm doing some testings on my end, hopefully we will be able to come up with something more robust by combining our findings.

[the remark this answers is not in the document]
Mannns: Hello,
I'm in holyday for a week.
I don't have the playgrounds right now but they looked like that:
https://playground.babylonjs.com/#KA93U#775 
https://playground.babylonjs.com/#PIZ1GK#571

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
Mannns: Okay !
I'll create a new class and separate both versions.

sebavan:
This is an amazing addition !!! @Mannns I ll do a final review once all @Popov72 comments have been resolved, I Love the idea of a SSR2 implementation not breaking back compat. Enjoy your vacations !!!

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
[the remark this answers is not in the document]
Popov72: Here's what I could come up with using the code [there](https://jcgt.org/published/0003/04/04/paper.pdf):

image

As you can see, there are a lot less artefacts (only at the base of the ball reflections) and they could be hidden with some blur. I'm going to play a bit more with the code and will make it available in case you are interested in integrating some parts.

image

Popov72:
Here's what I could come up with using the code [there](https://jcgt.org/published/0003/04/04/paper.pdf):

image

As you can see, there are a lot less artefacts (only at the base of the ball reflections) and they could be hidden with some blur. I'm going to play a bit more with the code and will make it available in case you are interested in integrating some parts.

image

packages/dev/core/src/Shaders/screenSpaceReflection.fragment.fx
```diff
@@ -174,3 +196,355 @@ void main()
...
+void main(void)
+{
+    #ifdef SSR_SUPPORTED
+
+    // *************** Get data from samplers ***************
+
+    vec4 original = texture2D(textureSampler, vUV);
+    vec3 spec = toLinearSpace(texture2D(specularSampler, vUV).rgb);
+
+    if (dot(spec, vec3(1.0)) <= 0.0){
```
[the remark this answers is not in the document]
Mannns: I tried to check for 0.12 and I have the same problem for standard and pbr materials: we miss some reflections / we loose reflection continuity and get ugly artifact
bug
s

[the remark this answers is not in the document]
Popov72: For standard materials you should not have the problem because the reflectivity values should be remapped from 0..1 to 0.04..1 before being saved in the reflectivity gbuffer (something like `new = old*0.96+0.04`, assuming the value is in linear space).

Regarding PBR, yes, testing for 0.04*3 is probably not the way to go because even a `metallic=0` material (dieletric) can have reflections depending on the roughness value:

**Roughness=0**:
image

**Roughness=1**:
image

The reflectivity of the material in both screenshots is the same, (0.04, 0.04, 0.04).

What bothers me by hard coding 0 is that we will have reflections computed for all texels when using PBR materials (because the lowest possible value is 0.04*3 which is > 0) even in cases where the reflection contributes nothing to the end result...

At least with the standard material we can decide where the reflections should occur thanks to the specular texture ((0,0,0) in the texture means no reflection and we can skip the computation). I wonder if we should not have the same possibility for PBR materials and have a texture that let the user choose what is reflective or not (wrt SSR)...

Let's keep the check as it is currently but we may need to revisit this question at some time.

Pinging @sebavan / @deltakosh so they are aware of the problem.

packages/dev/core/src/Shaders/screenSpaceReflection2.fragment.fx
```diff
@@ -0,0 +1,386 @@
...
+        info.visibilityBackup = 1.0;
+    #endif
+    return info;
+    }
+
+
+    // Calculate the start and end point of the reflection ray in screen space.
+    vec4 startSS = projection * startVS; // Project to screen space.
+    startSS.xyz /= startSS.w; // Perform the perspective divide.
+    #ifndef WEBGPU
```
Popov72, line 96: My mistake, only the Z coordinate is in the 0..1 range in WebGPU, so you can remove the `#ifndef` as x/y also need the remapping in WebGPU.

packages/dev/core/src/Rendering/screenSpaceReflectionsConfiguration.ts
```diff
@@ -19,5 +19,5 @@ export class ScreenSpaceReflectionsConfiguration implements PrePassEffectConfigu
     /**
      * Textures that should be present in the MRT for this effect to work
      */
-    public readonly texturesRequired: number[] = [Constants.PREPASS_NORMAL_TEXTURE_TYPE, Constants.PREPASS_REFLECTIVITY_TEXTURE_TYPE, Constants.PREPASS_POSITION_TEXTURE_TYPE];
+    public readonly texturesRequired: number[] = [Constants.PREPASS_NORMAL_TEXTURE_TYPE, Constants.PREPASS_POSITION_TEXTURE_TYPE, Constants.PREPASS_REFLECTIVITY_TEXTURE_TYPE];
```
Popov72, line 22: I think you want to revert the changes from this file?

Popov72:
Another follow up:

|Without the backface depth buffer|With the backface depth buffer|
|---------|----------|
|image|image|

Look at the reflection of the balls on the floor and the reflection of the red ball in the blue ball. These screenshots have been taken with `step=30`.

The backface depth buffer is used to compute a per-pixel geometry thickness instead of relying on a constant value (see https://github.com/kode80/kode80SSR).

Another one (`step=10`):

Without the backface depth buffer:
image

With the backface depth buffer:
image

Mannns:
> Another follow up:
> Without the backface depth buffer 	With the backface depth buffer
> image 	image
> 
> Look at the reflection of the balls on the floor and the reflection of the red ball in the blue ball. These screenshots have been taken with `step=30`.
> 
> The backface depth buffer is used to compute a per-pixel geometry thickness instead of relying on a constant value (see https://github.com/kode80/kode80SSR).
> 
> Another one (`step=10`):
> 
> Without the backface depth buffer: image
> 
> With the backface depth buffer: image
> 
> Source code, for reference: [master...Popov72:Babylon.js:ssr-improve](https://github.com/BabylonJS/Babylon.js/compare/master...Popov72:Babylon.js:ssr-improve

Love the rendering !
I will have a look ASAP
Just wonder how it behaves with a scene with walls or high mesh density.. ?

Popov72:
> Just wonder how it behaves with a scene with walls or high mesh density.. ?

It has all the problems of SSR, so depending on the scene it can fail miserably or be quite good.

For eg, here are 3 screenshots of the HillValley scene where I have simply enabled the new SSR (no tweaking of the meshes/materials):

image

image

image

I'm going to try to add a blur pass, as implemented [here](https://github.com/kode80/kode80SSR).

Popov72:
Blurring added.

No blur:
image

Some blur (`roughness=0.1`):
image

Lot of blur (`roughness=1`):
image

Mannns:
> Blurring added.

What you did is awesome !!

Popov72:
The PG I'm using for my testings: https://playground.babylonjs.com/#PIZ1GK#639

Note that it works only with the geometry renderer at the time and not with the prepass renderer yet.

Mannns:
> The PG I'm using for my testings: https://playground.babylonjs.com/#PIZ1GK#639
> 
> Note that it works only with the geometry renderer at the time and not with the prepass renderer yet.

May I help you ?
I added the depth in screenSpaceReflectionConfiguration.
Then I observed that the textureSampler is weird when enabling prepass:
image
(looks like color * reflectivity)

And it looks like this with Gbuffer :
image
(I don't really understand why)

Popov72:
Yes, the prepass renderer configuration was missing.

When fixed, the problem with the display was because the MRT is not created right from the start by the prepass renderer and the call to `ScreenSpaceReflection3PostProcess._getTextureSize()` was failing.

I have fixed the problem: http://localhost:1338/#PIZ1GK#642

However, we can now see that there are some default roughness on the road, something we don't see with the geometry renderer => the prepass renderer is right on this one, it's the geometry renderer that does not write the correct roughness in the gbuffer texture when using the standard material.

Maybe you can a look and see if you can fix the problem?

There's also another problem when enabling the blur pass (`ssr.blurDispersionStrength > 0`) with the prepass renderer, it seems we can't retrieve the input texture of a post process with `postProcess.inputTexture`, it returns `undefined`...

sebavan:
I converted as a draft as I enjoy a lot the progress but want to prevent any early merge during the finalization. You guyz are awesome !!!

Mannns:
> However, we can now see that there are some default roughness on the road, something we don't see with the geometry renderer => the prepass renderer is right on this one, it's the geometry renderer that does not write the correct roughness in the gbuffer texture when using the standard material.
> 
> Maybe you can have a look and see if you can fix the problem?

I just fixed it and push it on my master branch.
Since I had a merge conflict, I tried to catch up to the latest version of Babylon and now my pull request looks messy :(

So, in geometry.fragment.fx
it's: 
reflectivity = texture2D(reflectivitySampler, vReflectivityUV); 

instead of:
reflectivity.rbg = texture2D(reflectivitySampler, vReflectivityUV).rbg

Popov72:
Good catch, I have also fixed it in my sources and a couple of other ones related to linear/gamma space for the reflectivity.

I still have some problems when comparing the geometry renderer / prepass renderer:
* the prepass renderer does not handle transparent materials correctly when writing to the normal/reflectivity render target textures. Blending is enabled whereas it should not, but I don't know if it's possible to enable blending on some RT and not on some others, as we still need blending for the color RT... I tried to look for this issue in Google but could not find anything relevant to WebGL2 (in OpenGL you have `glDisablei` to disable blending on a specifc RT)... @sebavan Do you know if that is possible? I *think* I saw someone speaking of an extension for WebGL2 that would let use different alpha blending parameters on a per render target basis (so it would probably also allow to enable/disable alpha blending on a per render target basis) but could not google it...
* I have sometimes some differences in the reflections and it seems it's the geometry renderer which is at fault, I need to dig more...

Popov72:
> I have sometimes some differences in the reflections and it seems it's the geometry renderer which is at fault, I need to dig more...

It is also linked to the blending mode in the MRT. It appears it is sometimes better with the blending enabled when writing the depth/normal in the prepass renderer mode and sometimes it is worse, depending on the camera view...

yuqianma:
Great result !
Any plan to merge it?

Popov72:
It still needs some more work (PBR is not ok), I should be able to work on this again in the next weeks.

yuqianma:
Hummmmm...
No rush but... really want to use this cool effect.
I wish I could help but it's more than I can handle. 🍻

Popov72:
We need to finish a few things first, but this one is definitely still on the radar!

RaananW:
Please close this PR and open a new one from a cleaned clone of the repo.

Thanks! :-)

Popov72:
I will reopen the PR.

PR #12750, opened 2022-07-15
Add engine option for using exact sRGB conversions in the shader.

kircher1:
This provides an option to use the more expensive but accurate conversion for transforming colors in the shader to and from linear and sRGB color spaces. The conversion constants and formulas are described in the Wikipedia article for sRGB: https://en.wikipedia.org/wiki/SRGB

I opted to make this option a property passed into the engine, as it feels somewhat similar functionally to the engine option for high precision floats i.e., it affects many things (in this case, many shaders). IMO it would be odd to use the approximation in one shader and the exact formula in another. I think most of the time, you'd either want it always or never.


kircher1:
Not in this change, but in the future, I think the engine option should be also affect the CPU-side math implementations of toLinear/toGamma on the Color classes.

packages/dev/core/src/Shaders/ShadersInclude/helperFunctions.fx
```diff
@@ -45,34 +45,94 @@ mat3 inverseMat3(mat3 inMatrix) {
...
 }
 
+#if USEEXACTSRGBCONVERSIONS
+vec3 toLinearSpaceExact(vec3 color)
+{
+    vec3 nearZeroSection = 0.0773993808 * color;
+    vec3 remainingSection = pow(0.947867299 * (color + vec3(0.055)), vec3(2.4));
+    #if defined(WEBGL2) || defined(WEBGPU)
+        return mix(remainingSection, nearZeroSection, lessThanEqual(color, vec3(0.01)));
+    #else
```
kircher1, line 55: Note the difference in these cases is due to the fact glsl 1.0 doesn't support a bvec as the interpolation factor (but glsl 3.0 does). I'm assuming that means WebGPU can handle this too.

deltakosh (approved):

packages/dev/core/src/Materials/Background/backgroundMaterial.ts
```diff
@@ -137,6 +137,7 @@ class BackgroundMaterialDefines extends MaterialDefines implements IImageProcess
     public SKIPFINALCOLORCLAMP = false;
     public EXPOSURE = false;
     public MULTIVIEW = false;
+    public USEEXACTSRGBCONVERSIONS = false;
```
Popov72, line 140: To simplify things and not having to declare this variable in all materials (background, PBR, standard, etc), you should instead modify the `ThinEngine._getGlobalDefines` function.

It will simplify things a great deal because the define will be injected automatically for all shaders and you won't need to change any .ts files (except `thinEngine.ts`): only shader code has to be updated to take care of the define.

packages/dev/core/src/Shaders/ShadersInclude/helperFunctions.fx
```diff
@@ -45,34 +45,94 @@ mat3 inverseMat3(mat3 inMatrix) {
               b21, (-a21 * a00 + a01 * a20), (a11 * a00 - a01 * a10)) / det;
 }
 
+#if USEEXACTSRGBCONVERSIONS
+vec3 toLinearSpaceExact(vec3 color)
```
Popov72, line 49: Maybe you can add a comment pointing to the reference for these computation (https://en.wikipedia.org/wiki/SRGB)?

Popov72 (commented):
All good for me but the implementation can be simplified a lot, see my comment.

packages/dev/core/src/Shaders/ShadersInclude/helperFunctions.fx
Popov72: Yes, this will be correctly converted to WGSL by the GLSL=>WGSL converter.

packages/dev/core/src/Materials/Background/backgroundMaterial.ts
kircher1: Indeed, that is so much simpler! I had no idea about the global defines. Thank you for pointing that out.

Popov72 (approved):

sebavan (approved):
Love it !!!

PR #12754, opened 2022-07-15
fixed scene tree not sticking

amritanarasimhan:
When clicking arrow button in scene tree, children were not appearing unless the mouse was moved

### DIFF
--- a/packages/tools/guiEditor/src/components/sceneExplorer/treeItemSelectableComponent.tsx
+++ b/packages/tools/guiEditor/src/components/sceneExplorer/treeItemSelectableComponent.tsx
@@ -25,7 +25,9 @@ export interface ITreeItemSelectableComponentProps {
     filter: Nullable<string>;
 }
 
+
 export interface ITreeItemSelectableComponentState {
+    isExpanded: boolean;
     dragOver: boolean;
     isSelected: boolean;
     isHovered: boolean;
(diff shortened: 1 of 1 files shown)


packages/tools/guiEditor/src/components/sceneExplorer/treeItemSelectableComponent.tsx
```diff
@@ -140,16 +148,10 @@ export class TreeItemSelectableComponent extends React.Component<ITreeItemSelect
...
 
-        if (!entity.reservedDataStore) {
-            entity.reservedDataStore = {
-                isExpanded: true,
-                setExpandedState: (expanded: boolean) => (entity.reservedDataStore.isExpanded = expanded),
-            };
-        }
-        const isExpanded = entity.reservedDataStore.isExpanded || Tools.LookForItems(this.props.entity, this.props.selectedEntities);
-        entity.reservedDataStore.isExpanded = isExpanded;
+        const isExpanded = this.state.isExpanded || Tools.LookForItems(this.props.entity, this.props.selectedEntities);
```
RaananW, line 151: I wonder what is the need for the code `Tools.LookForItems(this.props.entity, this.props.selectedEntities)`. I saw that it is part of the old code as well, but it seems like `isExpanded`  was used later in the code as well, but you are using `this.state.isExpanded`, which might make this definition redundant?

RaananW (commented):
Just a small question about the code changes :-)

We can discuss that in a call if there is something wrong with my explanation.

packages/tools/guiEditor/src/components/sceneExplorer/treeItemSelectableComponent.tsx
amritanarasimhan: Thanks for the catch! I meant to use isExpanded to keep it close to the original code, but changed it to this.state.isExpanded while debugging and forgot to change it back.

carolhmj (changes requested):
I think this change had an unintended side effect: in the master branch, if you select an item on the canvas that is not expanded on the scene tree, the scene tree automatically expands to show it:
ezgif-1-1bb5c33d2a

While on this branch, this behavior doesn't happen anymore:
ezgif-1-6c66610cd4

(The GUI I used is here: http://localhost:1338/#C9NT11#1)

carolhmj (approved):

PR #12755, opened 2022-07-15
toolbar can expand now

amritanarasimhan:
No previous functionality for expanding toolbar on smaller screens (scroll only)


packages/tools/guiEditor/src/workbenchEditor.tsx
```diff
@@ -264,39 +333,108 @@ export class WorkbenchEditor extends React.Component<IGraphEditorProps, IGraphEd
...
+                                        </div>
+                                    );
+                                })}
+                                {}
+                            </div>
+                        }
+                    </div>
+                </>
+            );
+        }
```
RaananW, line 438: It seems like this code repeats itself quite a lot. You can set the different properties before returning the JSX, and  then return a single implementation of the same JSX. for example, something like this:

```javascript
// ...
const controlTypeMap = this.state.toolbarExpand ? ControlTypes : [...ControlTypesFHalf, ...ControlTypesSHalf];
const idForElement = this.state.toolbarExpand ? "toolbarGrab" : "toolbarGrabExpanded";
``` 

and then use those variables to render the toolbar.

packages/tools/guiEditor/src/workbenchEditor.tsx
```diff
@@ -191,69 +195,134 @@ export class WorkbenchEditor extends React.Component<IGraphEditorProps, IGraphEd
...
+        this.setState({ toolbarExpand: !this.state.toolbarExpand });
+    }
 
     render() {
-        return (
-            <Portal globalState={this.props.globalState}>
-                <div id="ge-header">
-                    <div className="command-bar">
-                        <CommandBarComponent globalState={this.props.globalState} />
+        if (this.state.toolbarExpand) {
```
RaananW, line 203: is there a way to reuse the JSX here? like in the example I provided under (in `createToolbar`) it feels like you can reuse the code quite a lot here as well.

packages/tools/guiEditor/src/controlTypes.ts
```diff
@@ -91,3 +91,80 @@ export const ControlTypes = [
...
+    },
+    {
+        className: "VirtualKeyboard",
+        icon: keyboardIcon,
+    },
+    {
+        className: "ColorPicker",
+        icon: colorPickerIcon,
+    },
+];
```
RaananW, line 170: these parameters can be added to ControTypes instead of creating new arrays with almost the same elements. You could add new keys to ControlTypes (for example `halfIcon:...` that will hold the icon for the extra parameters that were added).

packages/tools/guiEditor/src/controlTypes.ts
```diff
@@ -91,3 +91,80 @@ export const ControlTypes = [
...
+    {
+        className: "InputText",
+        icon: inputFieldIcon,
+    },
+    {
+        className: "InputPassword",
+        icon: passwordFieldIcon,
+    },
+];
+export const ControlTypesSHalf = [
```
RaananW, line 133: What is the difference between SHalf and FHalf? Is the difference needed? can both be in a single array?

packages/tools/guiEditor/src/main.scss
```diff
@@ -83,7 +88,55 @@
...
+
+        &:hover {
+            background: var(--selectionGrey);
+            .sub-items {
+                display: block;
+            }
+        }
+    }
+    .toolbar-icon {
+        width: 20px;
```
RaananW, line 139: with the right classes' definition, you can keep the id to be the same, and just add the difference between the two.
I mean that `toolbarGrabExpanded` is not needed. you can keep `toolbarGrab`, and add the class `expanded` or `toolbar-expanded` with the definition of the changes only.

RaananW (changes requested):
Most (if not all) of my comments are about code reusability. There is a lot of code that is repeating and can be adjusted and removed.

RaananW (approved):

PR #12757, opened 2022-07-16
add samplingmode option to effectlayer

Hypnosss:
Related: https://forum.babylonjs.com/t/how-to-change-sampling-mode-when-generating-down-scaled-glowlayers-blurtexture/32096/7
Not really familiar with ts so please for give me if it not a gental code.

### DIFF
--- a/packages/dev/core/src/Layers/glowLayer.ts
+++ b/packages/dev/core/src/Layers/glowLayer.ts
@@ -92,6 +92,11 @@ export interface IGlowLayerOptions {
      * Defines the blend mode used by the merge
      */
     alphaBlendingMode?: number;
+
+    /**
+     * Defines the sampling mode used by the blur textures.
+     */
+    samplingMode?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


deltakosh (approved):

deltakosh:
Just an issue with formatting and we are good to go :) (you can run npm run format:fix to auto fix it :))

btw @RaananW : could we simply run it on each PR? Is that a bad idea?

Popov72 (approved):

sebavan (approved):

RaananW:
> Just an issue with formatting and we are good to go :) (you can run npm run format:fix to auto fix it :))
> 
> btw @RaananW : could we simply run it on each PR? Is that a bad idea?

There are a few catches when discussing commiting to someone else's  PR. I would rather not do that.
I think that, in general, it will be better to only do it on master after merging a pull request. So allow merging unformatted, and format after merge. This has its downsides as well.
In any case - we can gladly discuss this :-)

sebavan (commented):
Are we sure we want to change the sampling mode for blur ??? it seems it won t achieve anything useful ? I am wondering if this is indeed necessary ?

Hypnosss:
> @Hypnosss what do you think ? as reading the forum thread it looks like this is not necessary as well ?

It is useful when anyone wants to do the same thing. However, I'm not sure if it is a request that common enough to put into the source code, and if it is a good way to archive the request.

sebavan:
so let me close it for now, and we ll reopen if anyone asks for it a second time :-)

PR #12762, opened 2022-07-18
ArcRotateCamera: Modify rotation logic to use invertRotation flag

PolygonalSun:
The `invertRotation` flag is primarily used with the FreeCamera and any Cameras that inherit from it.  Since the ArcRotateCamera uses different rotation logic, it never actually makes use of the `invertRotation` flag or its numerical counterpart, the `invertRotationSpeed` property.  This PR allow the user to use this flag with the ArcRotateCamera and specify the inverted rotational speed.  The discrepancy was initially brought up in the forum thread: https://forum.babylonjs.com/t/inverting-arcrotatecamera-controls-for-photodome/32147/6.

### DIFF
--- a/packages/dev/core/src/Cameras/arcRotateCamera.ts
+++ b/packages/dev/core/src/Cameras/arcRotateCamera.ts
@@ -921,9 +921,9 @@ export class ArcRotateCamera extends TargetCamera {
             if (this.parent && this.parent._getWorldMatrixDeterminant() < 0) {
                 inertialAlphaOffset *= -1;
             }
-            this.alpha += inertialAlphaOffset;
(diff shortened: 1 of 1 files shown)


RaananW (approved):
LGTM!

There is also the `inverseRotationSpeed` parameter, but I am not sure how to incorporate it in an ArcRotateCamera.

sebavan:
> There is also the `inverseRotationSpeed` parameter, but I am not sure how to incorporate it in an ArcRotateCamera.

We discussed this one with @PolygonalSun yesterday and think it might be weird as we already have the sensitvity thingies and relying on both was strange ?

PR #12767, opened 2022-07-19
KHR animation pointer

### LINKED ISSUE #12544: Implement KHR_animation_pointer
Specification PR: https://github.com/KhronosGroup/glTF/pull/2147
Sample Models PR: https://github.com/KhronosGroup/glTF-Sample-Models/pull/353

## For reference

Original Microsoft Proposal: https://github.com/najadojo/glTF/tree/EXT_property_animation/extensions/2.0/Vendor/EXT_property_animation

Implementation of original Microsoft Proposal: https://github.com/najadojo/Babylon.js/blob/EXT_property_animation/loaders/src/glTF/2.0/Extensions/EXT_property_animation.ts

pandaGaume:
This is the first version of the `KHR_animation_pointer` for the GLTF loader.
The basic idea behind the code is to turn every kind of animation into a pointer, in order to keep only one code path for all. This has been tested with success with all the samples provided as part of the Khronos proposal.
Even if functional, the code is still not very clean and many ehancement must be done in order to reach a certain level of quality. However, calendar issue drive us to put this onto PR rapidely.
[AnimateAllTheThings.zip](https://github.com/BabylonJS/Babylon.js/files/9144599/AnimateAllTheThings.zip)
This can be tested with the file above, which encompass all the cases supported by the `KHR_animation_pointer`.
The main changes are located into 3 files.
`GLTFLoader` has been simplified by moving all the animation related code to the `KHR_animation_pointer.ts` where the extension lies. `KHR_animation_pointer.map.ts` contains the pointer tree mapping and miscellaneous function to serve the pointer parsing and logic.
All the other change are to support indirection between GLTF entities and Babylon ones. This is subject to discussion, because not very "clean".
May close #12544

### DIFF
--- a/packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.map.ts
+++ b/packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.map.ts
@@ -10,8 +10,7 @@ import type { Mesh } from "core/Meshes/mesh";
 import type { Nullable } from "core/types";
 
 export type GetGltfNodeTargetFn = (source: IGLTF, indices: string) => any;
-export type GetValueFn = (target: any, source: Float32Array, offset: number, scale?: number) => any;
-export type GetStrideFn = (target: any) => number;
+type GetValueFn = (target: any, source: Float32Array, offset: number, scale?: number) => any;
 
 export interface IAnimationPointerPropertyInfos {
     type: number;
@@ -21,7 +20,7 @@ export interface IAnimationPointerPropertyInfos {
     buildAnimations(target: any, fps: number, keys: any[], group: AnimationGroup, animationTargetOverride: Nullable<IAnimatable>, params?: any): void;
 }
 
-const _parseIntIndex = (str: string) => {
+const parseIntIndex = (str: string) => {
     const targetIndex = parseInt(str);
     if (isNaN(targetIndex)) {
         return -1;
@@ -29,9 +28,9 @@ const _parseIntIndex = (str: string) => {
     return targetIndex;
 };
 
-const _getGltfNode: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
+const getGltfNode: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     if (gltf.nodes) {
-        const i = _parseIntIndex(index);
+        const i = parseIntIndex(index);
         if (i != -1) {
             return gltf.nodes[i];
         }
@@ -39,9 +38,9 @@ const _getGltfNode: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     return null;
 };
 
-const _getGltfMaterial: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
+const getGltfMaterial: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     if (gltf.materials) {
-        const i = _parseIntIndex(index);
+        const i = parseIntIndex(index);
         if (i != -1) {
             return gltf.materials[i];
         }
@@ -49,9 +48,9 @@ const _getGltfMaterial: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     return null;
 };
 
-const _getGltfCamera: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
+const getGltfCamera: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     if (gltf.cameras) {
-        const i = _parseIntIndex(index);
+        const i = parseIntIndex(index);
         if (i != -1) {
             return gltf.cameras[i];
         }
@@ -59,50 +58,50 @@ const _getGltfCamera: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     return null;
 };
 
-const _getGltfExtension: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
+const getGltfExtension: GetGltfNodeTargetFn = (gltf: IGLTF, index: string) => {
     if (gltf.extensions && index) {
         return gltf.extensions[index];
     }
     return null;
 };
 
-const _getMatrix: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getMatrix: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? Matrix.FromArray(source, offset).scale(scale) : Matrix.FromArray(source, offset);
 };
 
-const _getVector3: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getVector3: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? Vector3.FromArray(source, offset).scaleInPlace(scale) : Vector3.FromArray(source, offset);
 };
 
-const _getQuaternion: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getQuaternion: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? Quaternion.FromArray(source, offset).scaleInPlace(scale) : Quaternion.FromArray(source, offset);
 };
 
-const _getColor3: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getColor3: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? Color3.FromArray(source, offset).scale(scale) : Color3.FromArray(source, offset);
 };
 
-const _getAlpha: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getAlpha: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? source[offset + 3] * scale : source[offset + 3];
 };
 
-const _getFloat: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getFloat: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? source[offset] * scale : source[offset];
 };
 
-const _getMinusFloat: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getMinusFloat: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return -(scale ? source[offset] * scale : source[offset]);
 };
 
-const _getNextFloat: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getNextFloat: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return scale ? source[offset + 1] * scale : source[offset + 1];
 };
 
-const _getFloatBy2: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
+const getFloatBy2: GetValueFn = (_target: any, source: Float32Array, offset: number, scale?: number) => {
     return (scale ? source[offset] * scale : source[offset]) * 2;
 };
 
-const _getWeights: GetValueFn = (target: any, source: Float32Array, offset: number, scale?: number) => {
+const getWeights: GetValueFn = (target: any, source: Float32Array, offset: number, scale?: number) => {
     if (target._numMorphTargets) {
         const value = new Array<number>(target._numMorphTargets!);
         for (let i = 0; i < target._numMorphTargets!; i++) {
@@ -146,7 +145,7 @@ abstract class AbstractAnimationPointerPropertyInfos implements IAnimationPointe
 }
 
 class TransformNodeAnimationPointerPropertyInfos extends AbstractAnimationPointerPropertyInfos {
-    public constructor(type: number, name: string, get: GetValueFn = _getVector3) {
+    public constructor(type: number, name: string, get: GetValueFn = getVector3) {
         super(type, name, get);
     }
     public isValid(target: any): boolean {
@@ -159,7 +158,7 @@ class TransformNodeAnimationPointerPropertyInfos extends AbstractAnimationPointe
 }
 
 class CameraAnimationPointerPropertyInfos extends AbstractAnimationPointerPropertyInfos {
-    public constructor(type: number, name: string, get: GetValueFn = _getFloat) {
+    public constructor(type: number, name: string, get: GetValueFn = getFloat) {
         super(type, name, get);
     }
 
@@ -173,7 +172,7 @@ class CameraAnimationPointerPropertyInfos extends AbstractAnimationPointerProper
 }
 
 class MaterialAnimationPointerPropertyInfos extends AbstractAnimationPointerPropertyInfos {
-    public constructor(type: number, name: string, get: GetValueFn = _getFloat, public fillMode: any = Material.TriangleFillMode) {
+    public constructor(type: number, name: string, get: GetValueFn = getFloat, public fillMode: any = Material.TriangleFillMode) {
         super(type, name, get);
     }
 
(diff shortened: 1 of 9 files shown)


packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.map.ts
```diff
@@ -0,0 +1,510 @@
...
+    buildAnimations(target: any, fps: number, keys: any[], group: AnimationGroup, animationTargetOverride: Nullable<IAnimatable>, params?: any): void;
+}
+
+const _parseIntIndex = (str: string) => {
+    const targetIndex = parseInt(str);
+    if (isNaN(targetIndex)) {
+        return -1;
+    }
+    return targetIndex;
+};
```
bghgary, line 30: No need for `_` for non-exported local functions and you can use function syntax. Ditto elsewhere.
``````suggestion
function parseIntIndex(str: string): number {
    const targetIndex = parseInt(str);
    if (isNaN(targetIndex)) {
        return -1;
    }
    return targetIndex;
}
```

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.map.ts
```diff
@@ -0,0 +1,510 @@
...
+import { Color3 } from "core/Maths/math.color";
+import type { IGLTF } from "../glTFLoaderInterfaces";
+import { Material } from "core/Materials/material";
+import type { IAnimatable } from "core/Animations/animatable.interface";
+import type { Mesh } from "core/Meshes/mesh";
+import type { Nullable } from "core/types";
+
+export type GetGltfNodeTargetFn = (source: IGLTF, indices: string) => any;
+export type GetValueFn = (target: any, source: Float32Array, offset: number, scale?: number) => any;
+export type GetStrideFn = (target: any) => number;
```
bghgary, line 14: Is this used?

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.map.ts
```diff
@@ -0,0 +1,510 @@
...
+import { Quaternion, Vector3, Matrix } from "core/Maths/math.vector";
+import { Color3 } from "core/Maths/math.color";
+import type { IGLTF } from "../glTFLoaderInterfaces";
+import { Material } from "core/Materials/material";
+import type { IAnimatable } from "core/Animations/animatable.interface";
+import type { Mesh } from "core/Meshes/mesh";
+import type { Nullable } from "core/types";
+
+export type GetGltfNodeTargetFn = (source: IGLTF, indices: string) => any;
+export type GetValueFn = (target: any, source: Float32Array, offset: number, scale?: number) => any;
```
bghgary, line 13: I'm not sure why these are exported. They are not used outside of this file, right?

pandaGaume: The first one is used as delegate type into the pointer parsing to access the object. i remove the second.

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.map.ts
pandaGaume: GetStrideFn  is not used anymore, removed

packages/dev/loaders/src/glTF/2.0/Extensions/index.ts
```diff
@@ -20,6 +20,8 @@ export * from "./KHR_mesh_quantization";
 export * from "./KHR_texture_basisu";
 export * from "./KHR_texture_transform";
 export * from "./KHR_xmp_json_ld";
+export * from "./KHR_animation_pointer";
+export * from "./KHR_animation_pointer.map";
```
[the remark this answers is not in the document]
pandaGaume: Yes, and no. if anyone need to add an extra extension (not a official one) and support the animation pointer, we need a way to add this new extension to the path without touch of the core one. I'm agree this is overkill actually and we might decide of a common interface and types to do this.
I remove it for now.

packages/dev/loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1501,238 +1500,35 @@ export class GLTFLoader implements IGLTFLoader {
...
+            };
+            channel.target.path = AnimationChannelTargetPath.POINTER;
+            delete channel.target.node;
+
+            // ensure to declare extension used.
+            this._gltf.extensionsUsed = this._gltf.extensionsUsed || [];
+            if (this._gltf.extensionsUsed.indexOf(GLTFLoader._KHRAnimationPointerName) === -1) {
+                this._gltf.extensionsUsed.push(GLTFLoader._KHRAnimationPointerName);
+            }
         }
```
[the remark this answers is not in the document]
pandaGaume: Yes i had the same question :-) I will try to find a clever way to do it without changing the source itself.

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+                return 3;
+            case AccessorType.VEC4:
+            case AccessorType.MAT2:
+                return 4;
+            case AccessorType.MAT3:
+                return 9;
+            case AccessorType.MAT4:
+                return 16;
+        }
+    }
```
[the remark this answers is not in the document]
pandaGaume: Grab & copy the tools from the Serializer then

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+        animationTargetOverride: Nullable<IAnimatable> = null
+    ): Promise<void> {
+        if (channel.target.path != AnimationChannelTargetPath.POINTER) {
+            throw new Error(`${context}/target/path: Invalid value (${channel.target.path})`);
+        }
+
+        if (channel.target.node != undefined) {
+            // According to KHR_animation_pointer specification
+            // If this extension is used, the animation.channel.target.node must not be set.
+            // Because the node isn’t defined, the channel is ignored and not animated due to the specification.
```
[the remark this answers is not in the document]
pandaGaume: It's specific to this extension when present, the node property must NOT be set. I do not see where it should be in the core spec. Let me know if i'm wrong.

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * [Specification PR](https://github.com/KhronosGroup/glTF/pull/2147)
+ */
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export class KHR_animation_pointer implements IGLTFLoaderExtension {
+    /**
+     * used to gently ignore invalid pointer. If false, invalid pointer will throw exception.
+     */
+    public static IgnoreInvalidPointer: boolean = true;
```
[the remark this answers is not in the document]
pandaGaume: I will chnge it at local

packages/dev/loaders/src/glTF/2.0/glTFLoader.ts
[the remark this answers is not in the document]
pandaGaume: The easiest way seems  to do a snapshot of the animation channel  like a 
`save()` & `restore()` pattern.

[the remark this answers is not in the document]
bghgary: It's okay for now. We can do something different if there is an issue in the future.

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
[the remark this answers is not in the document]
bghgary: The core spec:
https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#animations
> When node isn’t defined, channel SHOULD be ignored.

I think you are trying to say that if the node is defined, then KHR_animation_pointer shouldn't do anything with this channel? The last sentence is maybe backwards? Do you mean `Because the node is defined`?

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+
+                    // we extract the corresponding values from the readed value.
+                    // the reason for that is one GLTF value may be dispatched to several Babylon properties
+                    // one of example is baseColorFactor which is a Color4 under GLTF and dispatched to
+                    // - albedoColor as Color3(Color4.r,Color4.g,Color4.b)
+                    // - alpha as Color4.a
+                    for (const propertyInfo of animationTarget.properties) {
+                        // Ignore animations that have no animation valid targets.
+                        if (!propertyInfo.isValid(animationTarget.target)) {
+                            return Promise.resolve();
```
bghgary, line 159: nit: since you are already in a continuation, you just need to return.

```suggestion
                            return;
```

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+                            }
+                        }
+
+                        // each properties has its own build animation process.
+                        // these logics are located into KHR_animation_pointer.map.ts
+                        propertyInfo.buildAnimations(animationTarget.target, fps, keys, babylonAnimationGroup, animationTargetOverride, animationTarget.params);
+                    }
+                }
+            }
+            return Promise.resolve();
```
bghgary, line 212: ditto:
```suggestion
```

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+            throw new Error(`${context}/target/extensions.KHR_animation_pointer.pointer MUST be set.`);
+        }
+
+        const sampler = ArrayItem.Get(`${context}/sampler`, animation.samplers, channel.sampler);
+
+        return this._loadAnimationSamplerAsync(`${context}/samplers/${channel.sampler}`, sampler).then((data) => {
+            // this is where we process the pointer.
+            const animationTarget = this._parseAnimationPointer(`${context}/extensions/${this.name}/pointer`, pointer);
+
+            if (animationTarget) {
```
bghgary, line 141: I would recommend early out here to avoid indenting as much.

```suggestion
            if (!animationTarget) {
                return;
            }
```

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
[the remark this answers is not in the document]
pandaGaume: Yes exactly, i already change the comment to `is defined`

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -0,0 +1,343 @@
...
+                // build the keys
+                // build the animations into the group
+                const babylonAnimationGroup = animation._babylonAnimationGroup;
+                if (babylonAnimationGroup) {
```
bghgary, line 145: ditto:
```suggestion
                if (!babylonAnimationGroup) {
                    return;
                }
```

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
pandaGaume: wasn't know

packages/dev/loaders/src/glTF/2.0/glTFUtilities.ts
```diff
@@ -0,0 +1,29 @@
...
+            case AccessorType.VEC4:
+            case AccessorType.MAT2:
+                return 4;
+            case AccessorType.MAT3:
+                return 9;
+            case AccessorType.MAT4:
+                return 16;
+        }
+    }
+}
```
bghgary, line 29: I'm still not sure why this is in a class. Are you trying to namespace it? I would recommend:
```suggestion
/**
 * Used internally to determine how much data to be gather from input buffer at each key frame.
 * @param accessorType the accessor type
 * @returns the number of item to be gather at each keyframe
 */
export function getDataAccessorElementCount(accessorType: AccessorType) {
    switch (accessorType) {
        case AccessorType.SCALAR:
            return 1;
        case AccessorType.VEC2:
            return 2;
        case AccessorType.VEC3:
            return 3;
        case AccessorType.VEC4:
        case AccessorType.MAT2:
            return 4;
        case AccessorType.MAT3:
            return 9;
        case AccessorType.MAT4:
            return 16;
    }
}
```

packages/dev/loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1509,9 +1509,8 @@ export class GLTFLoader implements IGLTFLoader {
             channel.target.extensions = channel.target.extensions || {};
             channel.target.extensions.KHR_animation_pointer = {
                 pointer: `/nodes/${channel.target.node}/${channel.target.path}`,
+                lazzy: true,
```
bghgary, line 1512: Left over from testing?

packages/dev/loaders/src/glTF/2.0/glTFLoader.ts
```diff
@@ -1509,9 +1509,8 @@ export class GLTFLoader implements IGLTFLoader {
             channel.target.extensions = channel.target.extensions || {};
             channel.target.extensions.KHR_animation_pointer = {
                 pointer: `/nodes/${channel.target.node}/${channel.target.path}`,
+                lazzy: true,
             };
-            channel.target.path = AnimationChannelTargetPath.POINTER;
-            delete channel.target.node;
```
bghgary: is this intentional?

packages/dev/loaders/src/glTF/2.0/glTFLoader.ts
bghgary, line 1512: This isn't much better than the original way. Let's just go with the original way for now. We can deal with this later if there is an issue.

packages/dev/loaders/src/glTF/2.0/glTFUtilities.ts
pandaGaume: I mimic what is available into the GLTFSerializer

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
```diff
@@ -5,10 +5,11 @@ import { AnimationGroup } from "core/Animations/animationGroup";
 import type { IAnimatable } from "core/Animations/animatable.interface";
 import type { IAnimation, IAnimationChannel, _IAnimationSamplerData, IAnimationSampler } from "../glTFLoaderInterfaces";
 
-import { AccessorType, AnimationChannelTargetPath, AnimationSamplerInterpolation } from "babylonjs-gltf2interface";
+import { AnimationChannelTargetPath, AnimationSamplerInterpolation } from "babylonjs-gltf2interface";
 import { AnimationKeyInterpolation } from "core/Animations/animationKey";
 import { CoreAnimationPointerMap } from "./KHR_animation_pointer.map";
 import type { GetGltfNodeTargetFn, IAnimationPointerPropertyInfos } from "./KHR_animation_pointer.map";
```
bghgary, line 11: I don't see `GetGltfNodeTargetFn` being used anywhere.

packages/dev/loaders/src/glTF/2.0/glTFUtilities.ts
bghgary: I don't know if the GLTFSerializer is the best place to look for good practices :-)

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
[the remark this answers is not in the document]
bghgary: Ahh, yes, I didn't see that yet.

packages/dev/loaders/src/glTF/2.0/Extensions/KHR_animation_pointer.ts
pandaGaume: Into `_parseAnimationPointer`

bghgary:
One last comment to resolve, then it looks good to me. We should also add tests and add this extension to the inspector. I can take care of this later.

bghgary (approved):

PR #12786, opened 2022-07-22
fix splice hook

sebavan:
https://forum.babylonjs.com/t/how-to-remove-node-in-rendertargettexture-renderlist/32375/5


RaananW (approved):
nitpicky - it is better to use `oldSplice.call`  in that case. But the result is the same :-)

PR #12787, opened 2022-07-22
isPointerBlocker fixes

carolhmj:
Related issues:
https://forum.babylonjs.com/t/ispointerblocker-improvement/31152
https://forum.babylonjs.com/t/on-inputtext-click-the-pointer-should-not-go-through-it/32336

* Don't pass pointer events up the hierarchy if a control has isPointerBlocker set
* Fix inputText always passing pointer events to the scene even when isPointerBlocker is set to true
* Add some documentation to clarify which controls set isPointerBlocker to true by default


sebavan (approved):
LGTM but let s validate with @RaananW as well ?

RaananW (approved):

PR #12789, opened 2022-07-23
Fix distorted normals in large-radius flat IcoSpheres.

jemc:
Prior to this commit, a subdivided IcoSphere with a large radius would have distorted normals
for the vertices that were in the original set prior to subdivision.

Here is an example of a flat IcoSphere with `radius: 100`:

image

After some troubleshooting, I discovered that this problem was due to scaling at an inappropriate
point in the process of generating the mesh. After removing this inappropriate scaling,
the normals of the vertices now look as expected without distortion.

image

Non-flat icospheres are unaffected - they still look appropriately round, as they did before.

image


jemc:
(Note that I cannot apply labels to my own PR, so someone else will need to apply the "bug" label)

RaananW (approved):
LGTM, waiting for @sebavan to be sure

sebavan:
I love the fix but miss some backgrounds on the maths there, wondering if @BabylonJSGuide could have a look ?

deltakosh:
All good for me

Issue #12795, opened 2022-07-26
Change private members in gizmo to protected ones


braineo:
**Is your feature request related to a problem? Please describe.**

I am trying to customize `AxisDragGizmo` and replace the ones in `PositionGizmo` in TypeScript.

Since `AxisDragGizmo` has a lot `private` not `protected` variables, I needed to copy most of the code of `AxisDragGizmo` instead of extending it.

But `PositionGizmo` only accept `xGizmo` to be instance of `AxisDragGizmo`. Maybe Interface can be defined for each gizmo to make customization easy?

```ts
export class PositionGizmo extends Gizmo {
    /**
     * Internal gizmo used for interactions on the x axis
     */
    public xGizmo: AxisDragGizmo;
```

**Describe the solution you'd like**

It would be nice either having members being protected ones so that developers can subclass gizmo, or having Interfaces for gizmos

**Discussion**
[discussion](https://forum.babylonjs.com/t/expose-interface-for-gizmo/32497)

sebavan:
I would definitely prefer an interface here. Making everything protected won t solve the issue. 

An interface would help keeping a common contract in a cleaner way.

@CedricGuillemet, @RaananW  ?

braineo:
@sebavan probably both are needed

for example in `PositionGizmo`, if xGizmo can accept `IAxisDragGizmo`, that's great. Any class implemented `IAxisDragGizmo` can be assigned.

```
export class PositionGizmo extends Gizmo {
    public xGizmo: IAxisDragGizmo;
```

But it does not change the fact extending `AxisDragGizmo` is hard since there are many private members, including mesh and material. In order to make it look different, developer needs to reimplement it and cannot reuse the existing code.

RaananW:
I agree with both of you - an interface would be a wonderful solution, but as @braineo says - extending the existing class will be hard because of the private members. I am not sure what functionality you are trying to extend/change, but I assume you need access to those:

```javascript
    private _gizmoMesh: Mesh;
    private _coloredMaterial: StandardMaterial;
    private _hoverMaterial: StandardMaterial;
    private _disableMaterial: StandardMaterial;
    private _dragging: boolean = false;
```

I am not opposing to making them protected. as long as they stay within the scope of the class it is fine.

From the way the gizmo is used in PositionGizmo and the way it is implemented (and without diving too much into the code TBH), it feels like an interface to the Gizmo class is missing.

braineo:
[...]

PR #12796, opened 2022-07-26
Change private members in gizmo to protected

braineo:
close https://github.com/BabylonJS/Babylon.js/issues/12795

I changed most of member from private to protected, even for those with public setters and getters, as setters and getters can be overwritten.

Except for

1. caching variables like `_tmpQuaternion`
1. static method of mesh creation

One thing I'm not sure about is the shader code in`planeRotationGizmo.ts`. If I change from `private` to `protected`, eslint complains about naming convention, but I wonder whether it is worth to rename.

```ts
private static _RotationGizmoFragmentShader
```

I'll leave this as a draft for discussion for now, and meanwhile continue to test the extendability on this PR.

### DIFF
--- a/packages/dev/core/src/Gizmos/axisDragGizmo.ts
+++ b/packages/dev/core/src/Gizmos/axisDragGizmo.ts
@@ -23,7 +23,7 @@ export class AxisDragGizmo extends Gizmo {
      * Drag behavior responsible for the gizmos dragging interactions
      */
     public dragBehavior: PointerDragBehavior;
-    private _pointerObserver: Nullable<Observer<PointerInfo>> = null;
+    protected _pointerObserver: Nullable<Observer<PointerInfo>> = null;
     /**
      * Drag distance in babylon units that the gizmo will snap to when dragged (Default: 0)
      */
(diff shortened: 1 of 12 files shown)


CedricGuillemet (approved):
LGTM

RaananW (commented):
I have commented on the issue create and have one comment regarding this specific PR - 
is there a reason to make the private static function protected?

braineo:
> I have commented on the issue create and have one comment regarding this specific PR - is there a reason to make the private static function protected?

Not really functions, but static shader code. these 2 lines are in the whole 200+ line constructor function. maybe it is useful to overwrite the shader for convenience?

```
        Effect.ShadersStore["rotationGizmoVertexShader"] = PlaneRotationGizmo._RotationGizmoVertexShader;
        Effect.ShadersStore["rotationGizmoFragmentShader"] = PlaneRotationGizmo._RotationGizmoFragmentShader;
```

deltakosh:
Hey @braineo, I think we are good to go

Wanna remove the draft? (I can do it if you prefer :))

braineo:
> Hey @braineo, I think we are good to go
> 
> Wanna remove the draft? (I can do it if you prefer :))


Oh there are other gizmo interfaces not added, I’m bit busy this week, gonna add all when I find some time.

sebavan:
@braineo, no problem I guess another PR would do for the other ones

PR #12799, opened 2022-07-26
Added `Color3.FromHSV`

BarthPaleologue:
A simple static method to create a new Color3 from HSV color space.


carolhmj (commented):
The method seems ok but there are a lot of commits in the history? @RaananW should we do a merge commit or is it better to clean the history?

sebavan (approved):

PR #12802, opened 2022-07-27
[GC] Switch `= []` to `.length = 0` where possible

RaananW:
Quickly scraped all references to ` this.[anything] = [];` and checked if they are already created previously. In this case, setting to `length = 0` will be enough and will prevent the array from being garbage-collected.


runtothedoor:
👏 👏 👏

runtothedoor (approved):
All great thinking that this is never done with consts

sebavan (approved):

PR #12811, opened 2022-07-29
Safari/Selenium Tests

PolygonalSun:
This PR has the tentative changes to test the input system in Safari


packages/tools/testTools/src/seleniumTestUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+                const evaluateInitEngineForVisualization = ${evaluateInitEngineForVisualization.toString()}
+                const evaluatePrepareScene = ${evaluatePrepareScene.toString()}
+                const evaluateRenderSceneForVisualization = ${evaluateRenderSceneForVisualization.toString()}
+
+                globalThis.testSuccessful = false;
+
+                await evaluateInitEngineForVisualization("webgl1", false, false, "${globalConfig.baseUrl}");
+                await evaluatePrepareScene({playgroundId: "${pgId}"}, _globalConfig);
+                evaluateRenderSceneForVisualization(${framesToRender});
+            `
```
bghgary, line 30: I would recommend putting this code into a local function and executing it so that there is syntax checking.

packages/tools/testTools/src/seleniumTestUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+};
+
+// Given a test page with snippet information, check if globalThis.testSuccessful is true
+export const CheckTestSuccessStatus = async (driver: ThenableWebDriver): Promise<boolean> => {
+    return (await driver.executeScript(`
+            const evaluateDisposeSceneForVisualization = ${evaluateDisposeSceneForVisualization.toString()}
+            await evaluateDisposeSceneForVisualization({ forceUseReverseDepthBuffer: false, forceUseNonCompatibilityMode: false })
+
+            return globalThis.testSuccessful;
```
bghgary, line 40: ditto

packages/tools/tests/test/selenium/safari.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    // Check if allowMouse logic for camera touch input is validating correctly
+    // PG: https://playground.babylonjs.com/#ITQ2NZ#9
+    it("check isMouseEvent", async () => {
+        await LoadPlayground(driver, "#ITQ2NZ#9", getGlobalConfig(), 1000);
+        const el = await driver.findElement(By.id("babylon-canvas"));
+
+        // With allowMouse = true, touch controls should move camera forward if isMouseEvent is true
+        await driver.actions({ bridge: true }).move({ origin: el }).press().move({ x: 0, y: -200, origin: el }).release().perform();
```
bghgary, line 24: Did you check if `{ bridge: true }` is actually necessary like we discussed offline?

bghgary (commented):
LGTM

packages/tools/tests/test/selenium/safari.test.ts
PolygonalSun: So I looked into the bridge bool and from what I've got.  With the web-driver, if `bridge = true`, it will attempt to use the W3C protocol for handling the input automation.  This includes taking all of the input function calls and putting them into a single call, reducing latency.  If this gets rejected, it would default to the legacy actions and call each action using a legacy protocol.  Furthermore, the W3C protocol works from the middle of a WebElement, as opposed to the top-left corner like the legacy one.  Setting the bridge bool to true also adds a step to legacy calls to work from the middle, if it ends up using it.  TBH, it isn't necessary but it's convenient just for that last point.

Ref: https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/lib/input_exports_Actions.html

PolygonalSun (commented):
Changed script snippets to functions for syntax checking

sebavan:
LGTM as well, will let @RaananW review more as well :-)

RaananW (commented):
LGTM. I only have one comment about the name of the task and project. `test:selenium` feels a little weird, because you are mentioning the test environment, but not the test type. Maybe `test:interactions` ?

sebavan:
I can do the rename in the PR if you want @RaananW so I can merge it in as @PolygonalSun is OOF ?

RaananW:
Sounds good, Let's change the name and merge

bghgary:
Just add a comment with a suggestion and commit it that way? :-)

sebavan:
It actually also requires to change the folder structure for the folder name and such :-) I ll see if I can do it tomorrow or I leave it to @PolygonalSun

RaananW:
If we want to merge this - it is fine. We can do the renaming afterwards. @sebavan i doubt you have access to @PolygonalSun remote branch to make the changes, so this will probably be the best way.

sebavan:
replaced by #12838, as @PolygonalSun is OOF for the rename

bghgary:
> If we want to merge this - it is fine. We can do the renaming afterwards. @sebavan i doubt you have access to @PolygonalSun remote branch to make the changes, so this will probably be the best way.

All maintainers should have access to submitter's branch unless they explicitly uncheck the checkbox that allows it.

image

PR #12815, opened 2022-08-01
Inspector step modifier keys

vinhui:
In one of our projects I had to play around with the positions of objects a bunch to get the perfect position. It annoyed me that the `step` of the position fields was a bit low (pressing the up/down arrow keys only moved it slightly).

This PR solves that issue by using modifier keys. You can now hold `shift` to increase the step x10. Additionally, if you hold `ctrl` (`meta` on mac) as well, it will increase x100.

I implemented this feature for both the `NumericInputComponent` and `FloatLineComponent` so it should work for all number input fields in the inspector.


RaananW:
Wow! thanks a lot :-)

@PatrickRyanMS - would be great to hear your input here :-)

packages/dev/sharedUiComponents/src/lines/floatLineComponent.tsx
```diff
@@ -155,26 +155,31 @@ export class FloatLineComponent extends React.Component<IFloatLineComponentProps
     }
 
     incrementValue(amount: number) {
-        if (this.props.step) {
```
deltakosh, line 158: wait this is a breaking change

deltakosh: I mean this is called by the InputArrowsComponent will then ignore the steps if we remove this line

deltakosh (changes requested):
Need to fix a tiny breaking change and good to go

RaananW:
@deltakosh - I believe @PatrickRyanMS has a solution for that, similar to the one offered in the GUI editor. I would rather wait for his input and see if it is worth it to merge this or wait till we fix it globally throughout the framework's tools.

deltakosh:
Sounds good!

packages/dev/sharedUiComponents/src/lines/floatLineComponent.tsx
vinhui: Good point, I missed that that was called somewhere else as well.
As a side note, where in the inspector is the `InputArrowsComponent` used? I couldn't find a usage to test if everything worked.

vinhui:
I'll wait with fixing it in that case

PatrickRyanMS:
We will bring the GUI editor system for incrementing/decrementing values with click + drag but having modifier + arrow keys as a separate form of value control should not be discounted. Click + drag is a nice way to interactively change a value, but it's not precise. Having modifier + arrow keys for more precise control of values also seems useful so long as there is no conflict between the systems. Right now the GUI editor supports both click + drag and arrow key modification of a value field. The biggest difference is that the GUI editor uses only two decimal places in its values and the arrow keys move that value by 0.1 in step. So an arrow key modifier for stepping * 100 might be overkill there. 

I don't see a real need to track percentage values to four decimal places in the GUI editor, so I don't know that we can standardize the input values from GUI editor to the other editors (ACE, NME, Inspector, etc) so we may need to slightly modify the multiplier per tool if we can. We can still keep the mental model of one modifier gives a medium step and a second modifier gives a large step, but will likely need to customize those steps per editor.

sebavan:
@vinhui so could you push the tiny fix so that I would merge the PR in ? 

We had a chat with @PatrickRyanMS and we will revisit when working on the GUI and ensure we keep the functionnality :-)

vinhui:
Pushed a fix. Although I'm still not entirely sure where the arrow component is used, the fix is so simple it couldn't possibly go wrong (still wouldn't mind a second opinion though).

As mentioned in the OP, I mainly implemented this because I wanted a higher step value for that project. This was the smallest change I could think of to achieve that without impacting anyone else's workflow. There are probably other solutions to this which wont be as hidden as this is (although this is not uncommon functionality in other applications). 

I wouldn't mind babylon storing (i.e. localstorage) some settings like a step modifier in some preferences. Could maybe also remember last selected object, last selected transform tool etc.

vinhui:
Side note, could someone also label this PR so that it will show up in the changelog?

sebavan (approved):

sebavan (approved):
LGTM, @deltakosh  ???

deltakosh (approved):

PR #12821, opened 2022-08-02
DeviceInputSystem: Update Device Input folder structure

PolygonalSun:
The purpose of this PR is to consolidate all of the DeviceInputSystem specific files into the DeviceInput folder.  All of the import references should also be updated.


deltakosh:
Is the breaking change only about the folder structure? And why do we need that folder structure change?

PolygonalSun:
Closing to address with team on Monday

sebavan:
is inputInterfaces not exposed or mainly used publicly ? 

If not LTGM.

PolygonalSun:
> is inputInterfaces not exposed or mainly used publicly ?
> 
> If not LTGM.

inputInterfaces is only used internally for the DeviceInputSystem implementations so we should be good.

sebavan (approved):

PR #12829, opened 2022-08-03
setSnapshotRenderingMode should have no effect on WebGL engine

carolhmj:
Playground example: https://playground.babylonjs.com/?#SYQW69#951

### DIFF
--- a/packages/dev/core/src/Engines/thinEngine.ts
+++ b/packages/dev/core/src/Engines/thinEngine.ts
@@ -700,7 +700,7 @@ export class ThinEngine {
     }
 
     public set snapshotRenderingMode(mode: number) {
-        this._snapshotRenderingMode = mode;
+        // Noop on webgl, as snapshot rendering is only for WebGPU
     }
 
     protected _useExactSrgbConversions = false;


Popov72:
The problem is not in Babylon but in the PG, line 76. It should be:

```javascript
if (engine.snapshotRendering && engine.snapshotRenderingMode === BABYLON.Constants.SNAPSHOTRENDERING_FAST) {
```

Your fix would not fix the PG if `SNAPSHOTRENDERING_STANDARD` was tested instead of `SNAPSHOTRENDERING_FAST`.

sebavan:
@Popov72 amazing catch !!! :-) should we still do the noop as well then ?

Popov72:
No, I think we should still udpate the property as not doing anything is not a no-op: the value is still `SNAPSHOTRENDERING_STANDARD` which is a valid mode (we don't have a `SNAPSHOTRENDERING_NOSNAPSHOT` value).

carolhmj:
Shall we close the PR then?

sebavan:
I think you could force it to always be standard independently of the mode if I understand @Popov72 correctly ?

Popov72:
Yes you could do that but I'm not sure it's better than doing nothing :)

sebavan:
ok so let s close the PR then @Popov72 ???

PR #12840, opened 2022-08-04
Nme drop wire

deltakosh:


sebavan (approved):
No idea how it works but it looks coool ;-)

PR #12841, opened 2022-08-04
Fix #12839

### LINKED ISSUE #12839: Add the possiblity to create custom top level nodes in the scene tree.
Currently, we can customize the inspector with plenty of flexibility and add some custom properties and such on existing types like meshes and the like of it.

image

The overall idea would be to allow the creation of a node at the same level than Textures and Materials on which we could display our custom properties.

It would prevent the need of creating for instance an empty mesh to add custom properties on it.

deltakosh:


deltakosh:
Once merged, this PG should work: https://playground.babylonjs.com/#XK769K

packages/dev/inspector/src/components/sceneExplorer/sceneExplorer.scss
```diff
@@ -436,13 +436,17 @@
             grid-template-columns: 1fr 20px 20px auto 5px;
             align-items: center;
 
-            .enableLight {
+            .visibi {
```
sebavan, line 439: Not sure it should have changed but might be fully wrong ??? all the rest is GG :-)

sebavan (approved):

PR #12850, opened 2022-08-08
Created an offset when pasting controls

amritanarasimhan:
When controls were pasted, they were visually directly on top of each other. Now, there is a visible change on screen when controls are pasted due to the offset


packages/tools/guiEditor/src/workbenchEditor.tsx
```diff
@@ -26,6 +26,8 @@ interface IGraphEditorState {
     showPreviewPopUp: boolean;
 }
 
+const PASTE_OFFSET = 10; //number of pixels a pasted control is offsetted from the copied control
```
sebavan, line 29: Would be great if we could share the value with the other one ?

sebavan (commented):
PasteOffset is not initialized and will be 0, is it still useful and missing a value ?

sebavan (approved):
Let s also add @carolhmj to review.

packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -45,6 +45,7 @@ const ARROW_KEY_MOVEMENT_LARGE = 5; // px
 const MAX_POINTER_TRAVEL_DISTANCE = 5; //px^2. determines how far the pointer can move to be treated as a drag vs. a click
 
 export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps> {
+    private _pasteOffset = 10;
```
carolhmj, line 48: Is there any situation where we would want to change the paste offset? If not, it would be better to have it as a module const instead.

carolhmj (changes requested):
While testing, I noticed the offset is being applied even to controls that are not pasted, is this desired behavior?
PASTE-GE

packages/tools/guiEditor/src/diagram/workbench.tsx
amritanarasimhan: Yes! The offset being applied is intentional. This is so that if a control is selected multiple times, it is easier to visually distinguish between the controls on screen. I did also send it to Patrick to confirm though, so if he wants to change the design I can do that

packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -61,9 +62,26 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
     private _pointerTravelDistance = 0;
     private _processSelectionOnUp = false;
     private _visibleRegionContainer: Container;
+    private _currLeft: number = 0;
+    private _currTop: number = 0;
     public get visibleRegionContainer() {
         return this._visibleRegionContainer;
     }
+    public get controlOffset() {
```
sebavan, line 70: You should remove this one :-) and instead export the const CONTROL_OFFSET.

On the other file, you could simply import the constant

amritanarasimhan: Oh that makes sense! But after talking to Patrick I think I can just completely remove the code from the other file :)

sebavan, line 70: I guess you can remove all the public getter you added as they are not used anymore :-) and this will be good to go

sebavan (changes requested):
Only the getter to be removed and good to go :-)

packages/tools/guiEditor/src/diagram/workbench.tsx
amritanarasimhan: Sorry for the delay! Pushed those changes before editing all my code because I needed to move into another branch for some issues yesterday, but should be all good to go now :)

RaananW (approved):
Don't forget to resolve conflicts :-)

sebavan:
and to remove the added getter as they are not required :-)

```
 public get currTop() {
        return this._currTop;
    }
    public get currLeft() {
        return this._currLeft;
    }
    public set currTop(val: number) {
        this._currTop = val;
    }
    public set currLeft(val: number) {
        this._currLeft = val;
    }
```

sebavan:
You now just need to merge with master :-)

PR #12852, opened 2022-08-09
Add IGizmo-extending interfaces for standard Gizmo variants

Hsifnus:
This pull request is a follow-up to #12795 and #12796 and adds interfaces representing what public properties should be implemented by custom gizmos seeking to fulfill the role of one of the standard `Gizmo` variants (`AxisDragGizmo`, `PositionGizmo`, etc).

Additionally, `Gizmo`-type properties within relevant classes now use the corresponding gizmo interface instead for easier integration of custom gizmos into `PositionGizmo`/`RotationGizmo`/`ScaleGizmo` as well as the standard `GizmoManager`.


sebavan (approved):
LGTM

PR #12856, opened 2022-08-09
Fix scene serialization and loading issues.

carolhmj:
Related forum issue: https://forum.babylonjs.com/t/importing-mesh-loses-hierarchy/32779

When we parse a mesh, the parentId from the JSON probably won't be the same as the uniqueId of the recreated parent, so keep a mapping to properly find this recreated parent.


packages/dev/core/src/Loading/Plugins/babylonFileLoader.ts
```diff
@@ -803,7 +808,7 @@ SceneLoader.RegisterPlugin({
                 for (let index = 0, cache = scene.meshes.length; index < cache; index++) {
                     currentMesh = scene.meshes[index];
                     if (currentMesh._waitingParentId) {
-                        const parent = scene.getLastEntryById(currentMesh._waitingParentId);
```
deltakosh, line 806: can we get rid of _waintingParentId then?

carolhmj: I don't think we can, because the nodes still need a way to store the id that was on the JSON while we're still loading all the nodes

deltakosh (approved):

sebavan (changes requested):
Something sounds weird, I am not sure why the test should change ?

sebavan (approved):

PR #12858, opened 2022-08-09
Update ComputePressureObserver

stefansundin:
Hello.

I have been working on updating our electron app to the latest electron version (20.0.1, previously we used 17.3.1) and we had a breakage in our use of babylon.js when the scene is disposed:

> TypeError: Failed to execute 'unobserve' on 'ComputePressureObserver': 1 argument required, but only 0 present.

Note that not all electron applications will experience this problem. ComputePressureObserver is only exposed if you use `experimentalFeatures: true`.

It appears that electron 19 and up is using a Chromium version with an updated ComputePressureObserver API. Here is the relevant change in [the working draft](https://wicg.github.io/compute-pressure/):
- https://github.com/WICG/compute-pressure/issues/21
- https://github.com/WICG/compute-pressure/pull/25

The error is ignored in `observe` but not in `unobserve`. My first commit here catches the error in `unobserve`. I first tried to use `catch()` like in `observe` but it didn't work so I had to use `try { ... } catch {}`. Even though it isn't documented in the spec, `observe` is returning a promise but `unobserve` is not.

The second commit updates the use of the observer to work with electron >= 19. This disables the observer for electron < 19 (`ComputePressureObserver.supportedSources` is not defined), but I don't think many will notice since the only thing this is used for is the CPU utilization metric in the Realtime Performance Viewer.

Screen Shot 2022-08-09 at 14 41 31

The spec appears to have more updates than what electron supports. Doesn't look like the `samplerate` property works.

I made this simple Electron Fiddle that can be used to test with: https://gist.github.com/stefansundin/4cee8e88133efb196f54f2d2a9f05155. You need to run `npm run start` to run the babylon.js server as well.

Please let me know if you want me to change anything! Thanks!

### DIFF
--- a/packages/dev/core/src/Misc/computePressure.ts
+++ b/packages/dev/core/src/Misc/computePressure.ts
@@ -25,6 +25,7 @@ export class ComputePressureObserverWrapper {
 
     /**
      * Method that must be called to begin observing changes, and triggering callbacks.
+     * @param source defines the source to observe
      */
     observe(source: IComputePressureSource): void {
         this._observer?.observe &&
@@ -35,6 +36,7 @@ export class ComputePressureObserverWrapper {
 
     /**
      * Method that must be called to stop observing changes and triggering callbacks (cleanup function).
+     * @param source defines the source to unobserve
      */
     unobserve(source: IComputePressureSource): void {
         try {
--- a/packages/dev/core/src/Misc/computePressure.ts
+++ b/packages/dev/core/src/Misc/computePressure.ts
@@ -20,15 +20,15 @@ export class ComputePressureObserverWrapper {
      * Returns true if ComputePressureObserver is available for use, false otherwise.
      */
     public static get IsAvailable() {
-        return IsWindowObjectExist() && "ComputePressureObserver" in window;
+        return IsWindowObjectExist() && "ComputePressureObserver" in window && (<any>window).ComputePressureObserver?.supportedSources?.includes("cpu");
     }
 
     /**
      * Method that must be called to begin observing changes, and triggering callbacks.
      */
-    observe(): void {
+    observe(source: IComputePressureSource): void {
         this._observer?.observe &&
-            this._observer?.observe().catch(() => {
+            this._observer?.observe(source).catch(() => {
                 // Ignore error
             });
     }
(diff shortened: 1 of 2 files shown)


carolhmj:
LGTM but I'll tag @deltakosh to give a look too

deltakosh:
This is a great one! Thanks for updating our support ;)

Please fix the format issue and we can merge

deltakosh (approved):

sebavan (approved):
GG !!!

stefansundin:
Thanks for the reviews!

I ran `npm run format` and `npm run lint` but I got these errors:

```
$ npm run format

> @babylonjs/root@1.0.0 format
> npm run format:check && npm run format:fix


> @babylonjs/root@1.0.0 format:check
> prettier --check ./packages/**/src/**/*.{ts,tsx,js,json,scss,css}

Checking formatting...
[error] No files matching the pattern were found: "./packages/**/src/**/*.json".
[error] No files matching the pattern were found: "./packages/**/src/**/*.css".
All matched files use Prettier code style!

$ npm run lint

> @babylonjs/root@1.0.0 lint
> npm run lint:check && npm run lint:fix


> @babylonjs/root@1.0.0 lint:check
> eslint ./packages/**/src/**/*.{ts,tsx,js,json}


Oops! Something went wrong! :(

ESLint: 8.11.0

No files matching the pattern "./packages/**/src/**/*.json" were found.
Please check for typing mistakes in the pattern.
```

The solution was to run the commands with updated patterns, i.e.:

```
$ prettier --check ./packages/**/src/**/*.{ts,tsx,js,scss}
$ eslint ./packages/**/src/**/*.{ts,tsx,js}
```

So I guess for some reason my setup doesn't have the missing files so the whole command failed instead of running it on the files that actually do exist. A potential solution to this might be to add a few dummy files that will always exist so that the patterns are always satisfied.

Just thought I should raise this potential issue that might bite babylon.js newcomers! :)

packages/dev/core/src/Misc/computePressure.ts
```diff
@@ -20,24 +20,30 @@ export class ComputePressureObserverWrapper {
      * Returns true if ComputePressureObserver is available for use, false otherwise.
      */
     public static get IsAvailable() {
```
stefansundin, line 22: I also considered adding a `source` argument to `IsAvailable` but that isn't possible since it is a getter. It would then check for that source with `includes(source)` instead of `includes("cpu")`. Let me know if you prefer that and I can make it a regular function.

deltakosh: We cannot change that getter because of backward compat but we can do this:
- isAvailable tests on cpu
- we add a new function to test for a specific source

deltakosh:
To fix formating: npm run format:fix :)

packages/dev/core/src/Misc/computePressure.ts
stefansundin: Ok, makes sense. Maybe it isn't necessary for this change.

PR #12867, opened 2022-08-12
Add boolean invertDirection to allow inverting the scrollbar navigati…

carolhmj:
…on direction.

Forum issue: https://forum.babylonjs.com/t/scrollviewer-horizontal-direction-is-in-reverse/32995/


RaananW (commented):
I think `invertDirection` might be confusing. Maybe `invertScrollDirection`? invertDirection might be considered inverting the direction of the component itself

carolhmj:
> I think `invertDirection` might be confusing. Maybe `invertScrollDirection`? invertDirection might be considered inverting the direction of the component itself

Makes sense, changed the name!

sebavan (approved):

PR #12868, opened 2022-08-12
GUI Editor launches from PG with correct size

amritanarasimhan:
Launching the GUI Editor from the PG would always have the GUI texture be at the default size (1024x1024) instead of whatever size it was orignally defined as

### DIFF
--- a/packages/dev/inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
+++ b/packages/dev/inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
@@ -45,8 +45,9 @@ export function SetGUIEditorURL(guiEditorURL: string) {
  * if you are in an ES6 environment, you must first call InjectGUIEditor to provide the gui-editor package
  * If you are in a UMD environment, it will load the package from a URL
  * @param adt
+ * @param fromPG defines whether editor is being opened from the Playground
  */
-export async function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture) {
+export async function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture, fromPG?: boolean) {
     guiEditorContainer = guiEditorContainer || _getGlobalGUIEditor();
     if (!guiEditorContainer) {
         if (typeof BABYLON !== "undefined") {
@@ -65,5 +66,5 @@ export async function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture) {
             throw `Tried to call EditAdvancedDynamicTexture without first injecting the GUI editor. You need to call InjectGUIEditor() with a reference to @babylonjs/gui-editor. It can be imported at runtime using await import("@babylonjs/gui-editor").`;
         }
     }
-    guiEditorContainer.GUIEditor.Show({ liveGuiTexture: adt });
+    guiEditorContainer.GUIEditor.Show({ liveGuiTexture: adt }, fromPG);
 }
--- a/packages/dev/inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
+++ b/packages/dev/inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
@@ -45,9 +45,9 @@ export function SetGUIEditorURL(guiEditorURL: string) {
  * if you are in an ES6 environment, you must first call InjectGUIEditor to provide the gui-editor package
  * If you are in a UMD environment, it will load the package from a URL
  * @param adt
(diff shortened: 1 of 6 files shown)


packages/dev/inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
```diff
@@ -45,8 +45,9 @@ export function SetGUIEditorURL(guiEditorURL: string) {
  * if you are in an ES6 environment, you must first call InjectGUIEditor to provide the gui-editor package
  * If you are in a UMD environment, it will load the package from a URL
  * @param adt
+ * @param fromPG defines whether editor is being opened from the Playground
  */
-export async function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture) {
+export async function EditAdvancedDynamicTexture(adt: AdvancedDynamicTexture, fromPG?: boolean) {
```
RaananW, line 50: maybe instead of `fromPG` we can use `embed` or something like that?

packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -829,7 +830,11 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
         const light = new HemisphericLight("light1", Axis.Y, this._scene);
         light.intensity = 0.9;
 
-        this._guiSize = this._defaultGUISize;
+        if (fromPG) {
+            this.props.globalState.fromPG = true;
+            this._guiSize.width = Number(window.localStorage.getItem("width"));
+            this._guiSize.height = Number(window.localStorage.getItem("height"));
```
RaananW, line 836: where were those defined? and what happens is they are not defined?

RaananW (commented):
Just a few questions :-)

packages/tools/guiEditor/src/components/commandBarComponent.tsx
```diff
@@ -96,6 +99,11 @@ export class CommandBarComponent extends React.Component<ICommandBarComponentPro
         const copyyIcon = this.props.globalState.selectedControls.length === 0 ? copyIconDisabled : copyIcon;
         const deleteeIcon = this.props.globalState.selectedControls.length === 0 ? deleteIconDisabled : deleteIcon;
         const pasteeIcon = isPasteDisabled ? pasteIconDisabled : pasteIcon;
+
+        const responsiveSelected = this.props.globalState.fromPG ? window.localStorage.getItem("responsive") === "true" : DataStorage.ReadBoolean("Responsive", true);
+        const responsiveUI = this.props.globalState.fromPG ? window.localStorage.getItem("responsiveUI") === "true" : DataStorage.ReadBoolean("Responsive", true);
+        const unresponsiveUI = this.props.globalState.fromPG ? window.localStorage.getItem("responsiveUI") === "false" : !DataStorage.ReadBoolean("Responsive", true);
+
```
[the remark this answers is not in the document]
amritanarasimhan: It is! I guess I just did it this way because breaking it down to this level was what made the most sense to me when thinking about the logic, but I can also change it since it is a bit repetitive

packages/tools/guiEditor/src/diagram/workbench.tsx
amritanarasimhan: guiSize is initially defined on line 89! It's set to the default size of 1024x1024

packages/dev/inspector/src/components/sceneExplorer/entities/gui/guiTools.ts
amritanarasimhan: Is that change due to just having better naming conventions, or because I've also defined fromPG in globalState? I'm just wondering if I should change fromPG in globalState to embed as well?

packages/tools/guiEditor/src/diagram/workbench.tsx
RaananW: Is it possible to use the data storage instead of directly the localStorage? or does it make no sense here?

amritanarasimhan: Oh I see what you're saying! I think I can directly just use dataStorage. I'll change that now

amritanarasimhan: Actually, I just tried using DataStorage and it doesn't seem to be saving the data on reload. I'm not sure why because I see the DataStorage implementation does use localStorage, so I'll try to debug this a bit

RaananW (approved):

PR #12871, opened 2022-08-13
Adds a FileButton to the InspectableTypes

kircher1:
This playground would demo how to use it: https://www.babylonjs-playground.com/#XK769K#5

### DIFF
--- a/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/customPropertyGridComponent.tsx
+++ b/packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/customPropertyGridComponent.tsx
@@ -15,6 +15,7 @@ import type { LockObject } from "shared-ui-components/tabs/propertyGrids/lockObj
 import { ButtonLineComponent } from "shared-ui-components/lines/buttonLineComponent";
 import { OptionsLineComponent } from "shared-ui-components/lines/optionsLineComponent";
 import { TextLineComponent } from "shared-ui-components/lines/textLineComponent";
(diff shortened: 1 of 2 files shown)


packages/dev/inspector/src/components/actionTabs/tabs/propertyGrids/customPropertyGridComponent.tsx
```diff
@@ -131,6 +132,20 @@ export class CustomPropertyGridComponent extends React.Component<ICustomProperty
...
                 return <TextLineComponent key={inspectable.label} label={inspectable.label} value={" "} />;
+            case InspectableType.FileButton:
+                return (
+                    <FileButtonLineComponent
+                        key={inspectable.label}
+                        label={inspectable.label}
+                        onClick={
+                            inspectable.callback ||
+                            function () {
+                                console.warn("no call back function added");
```
RaananW, line 143:
```suggestion
                                Tools.Warn("no callback function added");
```
We do our best not to use console (and especially console.warn) directly. We have Warn in Tools, making it possible to abstract Logging in different scenarios.

RaananW (changes requested):
Try to avoid using `console` directly :-)

Also - what is the main reason to have something like that? What's the use-case?

RaananW:
@sebavan - Tools is not imported, it was just a suggestion :-)
But I think it is more important to understand the need for that before

sebavan:
The need is to have a custom tool where you can load file like a custom texture loader for instance

sebavan (approved):

PR #12872, opened 2022-08-13
comment corrections

BabylonJSGuide:
Comment changes only no code changes

floored should be fractional

One = (1.0, 1.0, 1.0) is not a unit vector as it has magnitude sqrt(3) not 1

### DIFF
--- a/packages/dev/core/src/Maths/math.vector.ts
+++ b/packages/dev/core/src/Maths/math.vector.ts
@@ -1399,7 +1399,7 @@ export class Vector3 {
     }
 
     /**
-     * Gets a new Vector3 from current Vector3 floored values
+     * Gets a new Vector3 from current Vector3 fractional values
      * @returns a new Vector3
      */
     public fract(): Vector3 {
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -3070,7 +3072,7 @@ export class Vector4 {
     }
 
     /**
-     * Gets a new Vector4 from current Vector3 floored values
+     * Gets a new Vector4 from current Vector3 fractional values
```
kircher1, line 3075: Gets a new Vector4 from current ~Vector3~ _**Vector4**_ fractional values

BabylonJSGuide: Well spotted. Will correct tomorrow.

BabylonJSGuide: Done

RaananW (approved):

PR #12883, opened 2022-08-18
Testing: Update Safari Interactions test to point to BrowserStack

PolygonalSun:
There are slight modifications that have been made to the safari tests so that they use BrowserStack.  This change should allow these tests to be used in the CI.


PolygonalSun (commented):
.

RaananW (commented):
How would those test run locally (on the local version)? Can browserstack be set only when running in the CI? (or when setting a specific flag?)

RaananW:
@PolygonalSun - any update on this one?

RaananW (approved):

PR #12884, opened 2022-08-18
Enable use of Tools.LoadScript in a WebWorker

### LINKED ISSUE #12882: Tools.LoadScript shouldn't return without calling onError / Infinite loading on ImportMeshAsync
This line in LoadScript returns without calling either onSuccess or onError if no window object is found, this leads to the parent promise never resolving and being stuck forever in a loading state. Please call onError here before returning

https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Misc/tools.ts#L420

Steps to repro: in WebWorker (i.e. location with no window object) try calling `SceneLoader.ImportMeshAsync` with a model that uses EXT_meshopt_compression

The ImportMeshAsync promise will never resolve

wmurphyrd:
Fixes #12882 

* Checks if `LoadScript` is called from a `WebWorker` context and, if so, uses `importScripts` instead of `HTMLScriptElement` to load the script.
* Ensures `LoadScript` always resolves with either `onSuccess` or `onError`

Use case: using `ImportMeshAsync` in a `WebWorker` for meshes requiring `MeshoptCompression`

RE the typings: there's apparently no clean solution as [TypeScript is extremely not a fan of code designed to run in both DOM and WebWorker](https://github.com/Microsoft/TypeScript/issues/20595), these hideous double type assertions seemed less bad than falsely declaring importScripts as available on the Window interface

### DIFF
--- a/packages/dev/core/src/LibDeclarations/browser.d.ts
+++ b/packages/dev/core/src/LibDeclarations/browser.d.ts
@@ -22,6 +22,12 @@ interface Window {
     setImmediate(handler: (...args: any[]) => void): number;
 }
 
+interface WorkerGlobalScope {
+    importScripts: (...args: string[]) => void;
+}
+
+type WorkerSelf = WindowOrWorkerGlobalScope & WorkerGlobalScope;
+
 interface HTMLCanvasElement {
     requestPointerLock(): void;
     msRequestPointerLock?(): void;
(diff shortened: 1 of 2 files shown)


sebavan (approved):
GG !!!! thanks a lot

sebavan:
Just a tiny format issue and good to go :-)

deltakosh (approved):
you rock! thanks a lot

PR #12888, opened 2022-08-20
Create a separate callback for the FileButton

kircher1:
The previous PR I recently submitted to add a FileButton can produce an error in the Typescript compiler since the `callback` function of the inspectable does not expect a parameter (so putting in a File parameter causes a compiler error).

This can be worked around by using a callback that takes a file param and then recasting it into a callback that takes no parameter i.e.,

```
        {
            label: "File Button demo",
            propertyName: "unused",
            type: InspectableType.FileButton,
            callback: ((file: File) => {
                // ...
            }) as () => void,
            accept: ".jpg, .jpeg",
        },
```

But that's not so great.

Here's a better solution. Add an explicit fileCallback that takes the file as a parameter. This is the callback that will be used with the FileButton.

### DIFF
--- a/packages/dev/core/src/Misc/iInspectable.ts
+++ b/packages/dev/core/src/Misc/iInspectable.ts
@@ -94,10 +94,13 @@ export interface IInspectable {
      */
     step?: number;
     /**
-     * Gets the callback function when using "Button" or "FileButton" mode.
-     * In "FileButton" mode, the file is passed as the parameter.
+     * Gets the callback function when using "Button" mode
      */
     callback?: () => void;
+    /**
+     * Gets the callback function when using "FileButton" mode
+     */
+    fileCallback?: (file: File) => void;
     /**
      * Gets the list of options when using "Option" mode
      */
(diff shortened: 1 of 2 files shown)


kircher1:
Here's the updated playground https://playground.babylonjs.com/?snapshot=refs/pull/12888/merge#XK769K#6

```
{
    label: "Open a file",
    type: BABYLON.InspectableType.FileButton,
    fileCallback: (x) => console.log("This is the file: " + x),
    accept: ".jpg, .png"
}
```

deltakosh:
What about having an optional parameter on the callback instead?

kircher1:
I did try, but it doesn't play nice with the `OptionsLineComponent` as the now `callback(file: File?)` would get assigned to `onSelect` which expects a string|number parameter (not a File).

So typescript compilation fails like this: `type '((file?: File | undefined) => void) | ((value: string | number) => void)' is not assignable to type '((value: string | number) => void) | undefined'.`

deltakosh:
I see! makes sense

So LGTM

deltakosh (approved):

PR #12891, opened 2022-08-22
Use LoadFile when loading images so that WebRequest custom headers ar…

carolhmj:
…e properly passed.

Related forum issue: https://forum.babylonjs.com/t/why-does-textureassettask-ignore-webrequest/33197
Related Playground: https://playground.babylonjs.com/#ZJYNY#1297

vvvvvvvv


RaananW (commented):
This will work (as expected!), but I find a different variant better - load the array buffer and set a blob URL. This will be memory efficient.

carolhmj:
Thanks, I updated it!

RaananW (approved):

PR #12897, opened 2022-08-24
[Build] Remove tslib reference in es6 packages (First step towards esm)

### LINKED ISSUE #12784: [Build] Remove tslib reference in es6 packages (First step towards esm)
This is partly related to esm, but can be done externally and unrelated to the esm work.

In our es6 packages we still require tslib. tslib is used as a polyfill for missing functionality for the target. The issue is that we are using decorators, which is not available in any javascript version. It is currently in a stage-2 proposal (https://github.com/tc39/proposals#stage-2), after not even being requested for a long time. It also requires `typescript`  to use the `experimentalDeocators` flag to have the feature enabled.

If we remove the use for decorators (or polyfill them ourselves similar to the way typescript does it), we can eliminate the need for a tslib reference which in turn will allow `@babylonjs/core` to be used directly in browser.

RaananW:
Fixes #12784

This PR removes the tslib dependency (while still providing it locally in its es6 version). This will allow using the es6 packages in the browser directly and will minimally help tree shaking, as not all of tslib is even needed (ATM only decorators).

Another change - the output for the esm packages is no longer es5! it is now es2021.

### DIFF
--- a/packages/dev/core/src/Engines/thinEngine.ts
+++ b/packages/dev/core/src/Engines/thinEngine.ts
@@ -9,6 +9,7 @@ import type { ShaderProcessingContext } from "./Processors/shaderProcessingOptio
 import type { UniformBuffer } from "../Materials/uniformBuffer";
 import type { Nullable, DataArray, IndicesArray } from "../types";
 import type { EngineCapabilities } from "./engineCapabilities";
+import type { Observer } from "../Misc/observable";
 import { Observable } from "../Misc/observable";
 import { DepthCullingState } from "../States/depthCullingState";
 import { StencilState } from "../States/stencilState";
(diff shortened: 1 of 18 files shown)


packages/dev/core/src/Engines/thinEngine.ts
```diff
@@ -1341,6 +1340,7 @@ export class ThinEngine {
                 this._gl.MAX = blendMinMaxExtension.MAX_EXT;
                 this._gl.MIN = blendMinMaxExtension.MIN_EXT;
             }
+            console.log("test");
```
RaananW, line 1343:
```suggestion
```

RaananW:
@sebavan - no package was released yet that supports this. this is just a TODO list for both of us to quickly discuss side effects of this change

PR #12902, opened 2022-08-25
Fix picking on ADV to account for texture coordinates

carolhmj:
This was the underlying issue for this report: https://forum.babylonjs.com/t/why-elements-holographicslate-are-not-always-clickable/33299, but was also affecting any texture that had transformations (offset, rotation and scale).

Example playground: https://playground.babylonjs.com/#H8NS43


RaananW (approved):
😲

PR #12904, opened 2022-08-26
InstancedMesh should be cloned with the new source mesh

RaananW:
Forum thread:
https://forum.babylonjs.com/t/instancedmeshes-cloned-from-assetcontainer-set-new-source/33301

Quick explanation - when using the assets container to instantiate meshes to the scene and cloning a new InstancedMesh, its sourceMesh is the one that is in the assets container and not the clone that was added to the scene.

This PR does the following:

* Adding the option to set the new source mesh to the instanced mesh cloning function
* when cloning a mesh that has an instanced mesh as a child the source mesh will be updated
* when using the assets container the new instanced mesh will have the cloned mesh as a source mesh.

### DIFF
--- a/packages/dev/core/src/Meshes/instancedMesh.ts
+++ b/packages/dev/core/src/Meshes/instancedMesh.ts
@@ -453,17 +453,16 @@ export class InstancedMesh extends AbstractMesh {
 
     /**
      * Creates a new InstancedMesh from the current mesh.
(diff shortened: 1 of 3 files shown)


RaananW:
PR to test:

https://playground.babylonjs.com/?snapshot=refs/pull/12904/merge#5NFRVE#131

deltakosh (approved):

PR #12905, opened 2022-08-26
Fix scene not clearing in some Default Rendering Pipeline with multicamera cases

carolhmj:
Examples of where this would happen:

- Have a DRP with a camera different than the scene's active camera (rendering to a RTT for instance): https://playground.babylonjs.com/#QW6XJE#4
- Set multiple active cameras AFTER the DRP has been created: https://playground.babylonjs.com/#QW6XJE#13

### DIFF
--- a/packages/dev/core/src/Materials/Textures/renderTargetTexture.ts
+++ b/packages/dev/core/src/Materials/Textures/renderTargetTexture.ts
@@ -21,6 +21,8 @@ import type { IRenderTargetTexture, RenderTargetWrapper } from "../../Engines/re
 import "../../Engines/Extensions/engine.renderTarget";
 import "../../Engines/Extensions/engine.renderTargetCube";
 import { Engine } from "../../Engines/engine";
+import { ArrayTools } from "core/Misc/arrayTools";
+import type { INotifyArrayChangeType } from "core/Misc/arrayTools";
 
 declare type Material = import("../material").Material;
 
@@ -58,12 +60,25 @@ export class RenderTargetTexture extends Texture implements IRenderTargetTexture
         return this._renderList;
     }
 
-    public set renderList(value: Nullable<Array<AbstractMesh>>) {
-        this._renderList = value;
+    private _renderListHasChangedObservable = new Observable<INotifyArrayChangeType<AbstractMesh>>();
+    private _renderListHasChangedObserver: Nullable<Observer<INotifyArrayChangeType<AbstractMesh>>> = null;
 
-        if (this._renderList) {
-            this._hookArray(this._renderList);
-        }
+    private _startObservingRenderListEvents() {
+        this._renderListHasChangedObserver = this._renderListHasChangedObservable.add(({ target, previousLength }) => {
+            if (target && previousLength && ((target.length > 0 && previousLength === 0) || (target.length === 0 && previousLength > 0))) {
+                this.getScene()?.meshes.forEach((mesh) => {
+                    mesh._markSubMeshesAsLightDirty();
+                });
+            }
+        });
+    }
+
+    private _stopObservingRenderListEvents() {
+        this._renderListHasChangedObservable.remove(this._renderListHasChangedObserver);
+    }
+
+    public set renderList(value: Nullable<Array<AbstractMesh>>) {
+        this._renderList = ArrayTools.MakeObservableArray(this._renderListHasChangedObservable, value);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


packages/dev/core/src/PostProcesses/RenderPipeline/Pipelines/defaultRenderingPipeline.ts
```diff
@@ -787,6 +804,14 @@ export class DefaultRenderingPipeline extends PostProcessRenderPipeline implemen
...
             this._resizeObserver = null;
         }
+        if (this._activeCameraChangedObserver) {
+            this._scene.onActiveCameraChanged.remove(this._activeCameraChangedObserver);
+            this._activeCameraChangedObserver = null;
+        }
+        if (this._activeCamerasChangedObserver) {
+            this._scene.onActiveCamerasChanged.remove(this._activeCamerasChangedObserver);
+            this._activeCamerasChangedObserver = null;
+        }
```
RaananW, line 814: it's enough to remove the observer, even if it is null. the `remove` function takes null as well.

RaananW (commented):
Nice :-)

I see Proxies are helpful in that case :-)

If we will use a Proxy somewhere else as well it might be helpful to move the makeArrayAProxy functionality to tools.

Is there anything we need to do when disposing the proxied array?

deltakosh:
Really appreciate the use of Reflect and Proxy. I agree with @RaananW let's make that part of tools just in case

packages/dev/core/src/scene.ts
```diff
@@ -1012,8 +1017,39 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         return this._lightsEnabled;
     }
 
+    private _activeCameras: Nullable<Camera[]> = new Array<Camera>();
```
sebavan, line 1020: the default one should be observable :-)

packages/dev/core/src/scene.ts
```diff
@@ -1012,8 +1017,39 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
     }
 
+    private _activeCameras: Nullable<Camera[]> = new Array<Camera>();
     /** All of the active cameras added to this scene. */
-    public activeCameras: Nullable<Camera[]> = new Array<Camera>();
+    public get activeCameras(): Nullable<Camera[]> {
+        return this._activeCameras;
+    }
+
+    private _proxySet(target: Camera[], property: string | symbol, c: Camera) {
```
sebavan, line 1026: _proxySet/push/makeObs... would be better in their own class or with a specific name for cameras, also I wonder if it could not be fully shared with the renderTarget implementations of _hookArray ?

RaananW: the proxy will be abstracted and moved to tools to share the implementation

carolhmj:
@sebavan sharing the implementation with the render target texture now :D

sebavan (approved):
GG !!!!

RaananW (approved):

packages/dev/core/src/Misc/arrayTools.ts
```diff
@@ -56,43 +55,51 @@ export class ArrayTools {
...
 
     private static _ProxySet<T>(observable: Observable<INotifyArrayChangeType<T>>, target: Array<T>, property: string | symbol, value: T): boolean {
-        observable.notifyObservers({ target, values: [value], operation: "set" });
+        observable.notifyObservers({ target, previousLength: target.length });
         return Reflect.set(target, property, value);
     }
 
     private static _ProxyPushOrUnshift<T>(operation: "push" | "unshift", observable: Observable<INotifyArrayChangeType<T>>, target: Array<T>, ...values: Array<T>): number {
-        observable.notifyObservers({ target, values, operation });
+        observable.notifyObservers({ target, previousLength: target.length });
```
Popov72, line 63: Note that for the `renderList` array of the `RenderTargetTexture` class it is a change in behaviour: the **push** operation was done before the additional code was executed. Here, the additional code will be execute first (through the `notifyObservers` call) and the **push** operation will be done afterwards.

I don't know if it will break something or not, but it's for sure a different behaviour. To keep the existing behaviour it would be safer to notify the observers after the operation has been done.

Also, I think it should be wise to always notify the observers at the same point (either before or after the operation on the array has been performed) for the sake of clarity. Currently:
* set, push/unshift, delete: observers notified before the operation
* pop/shift, splice: observers notified after the operation

Or maybe it is on purpose? In that case, adding some doc on `MakeObservableArray` explaining when the observers are notified in each case would probably be a good idea.

packages/dev/core/src/Materials/Textures/renderTargetTexture.ts
```diff
@@ -58,12 +60,25 @@ export class RenderTargetTexture extends Texture implements IRenderTargetTexture
...
-        this._renderList = value;
+    private _renderListHasChangedObservable = new Observable<INotifyArrayChangeType<AbstractMesh>>();
+    private _renderListHasChangedObserver: Nullable<Observer<INotifyArrayChangeType<AbstractMesh>>> = null;
 
-        if (this._renderList) {
-            this._hookArray(this._renderList);
-        }
+    private _startObservingRenderListEvents() {
+        this._renderListHasChangedObserver = this._renderListHasChangedObservable.add(({ target, previousLength }) => {
+            if (target && previousLength && ((target.length > 0 && previousLength === 0) || (target.length === 0 && previousLength > 0))) {
```
Popov72, line 68: I think you want to check `previousLength !== undefined` and not simply `previousLength` in `target && previousLength`

packages/dev/core/src/Misc/arrayTools.ts
```diff
@@ -56,43 +55,51 @@ export class ArrayTools {
...
     private static _ProxyDelete<T>(observable: Observable<INotifyArrayChangeType<T>>, target: Array<T>, property: string | symbol) {
-        observable.notifyObservers({ target, values: [Reflect.get(target, property)], operation: "delete" });
+        observable.notifyObservers({ target, previousLength: target.length });
         return Reflect.deleteProperty(target, property);
     }
 
     private static _ProxyPopOrShift<T>(operation: "pop" | "shift", observable: Observable<INotifyArrayChangeType<T>>, target: Array<T>) {
         const value = operation === "pop" ? Array.prototype.pop.apply(target) : Array.prototype.shift.apply(target);
-        observable.notifyObservers({ target, operation, values: [value] });
+        observable.notifyObservers({ target, previousLength: target.length });
```
Popov72, line 74: You should get the previous length before doing the operation and pass it to the observers, as `previousLength: target.length` will in reality pass the current length.

packages/dev/core/src/Misc/arrayTools.ts
```diff
@@ -56,43 +55,51 @@ export class ArrayTools {
...
         return value;
     }
 
     private static _ProxySplice<T>(observable: Observable<INotifyArrayChangeType<T>>, target: Array<T>, start: number, deleteNumber: number, ...added: Array<T>) {
         const values = Array.prototype.splice.apply(target, [start, deleteNumber, added]);
-        observable.notifyObservers({ target, values, operation: "splice" });
+        observable.notifyObservers({ target, previousLength: target.length });
```
Popov72, line 80: Same comment than above.

PR #12907, opened 2022-08-28
Fixing oit + derivatives branching

CraigFeldspar:
https://forum.babylonjs.com/t/order-independent-transparency-bugs-with-bump-texture-and-rendering-pipeline/29173/13

### DIFF
--- a/packages/dev/core/src/Shaders/default.fragment.fx
+++ b/packages/dev/core/src/Shaders/default.fragment.fx
@@ -99,8 +99,6 @@ void main(void) {
 
 #define CUSTOM_FRAGMENT_MAIN_BEGIN
 
-#include<oitFragment>
-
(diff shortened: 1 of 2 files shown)


Popov72 (approved):
I don't really understand why moving the code to the end fixes the problem.

In my understanding, if there's a derivative function used after a branching/return code, the GPU will still execute the code after the branch/return if at least one of the thread of the quad doesn't take the branch/return path.

But if moving the code fixes the problem it's fine to me (I probably got the working of the GPU wrong)!

[EDIT]
As expected I'm wrong; http://www.aclockworkberry.com/shader-derivative-functions/

According to this article, the compiler could flatten the code in such cases, but the GPU won't do anything special to handle this case.
[/EDIT]

sebavan (approved):

PR #12911, opened 2022-08-29
Add "metadata" support to "ReflectionProbe" class

julien-moreau:


deltakosh (approved):

packages/dev/core/src/Probes/reflectionProbe.ts
```diff
@@ -74,6 +74,11 @@ export class ReflectionProbe {
     @serializeAsVector3()
     public position = Vector3.Zero();
 
+    /**
+     * Gets or sets an object used to store user defined information for the sound.
```
Popov72, line 78: nit: sound => probe

julien-moreau: OMG nice catch

PR #12912, opened 2022-08-29
Add type on blob creation

carolhmj:
This fixes loading of SVG Textures

### DIFF
--- a/packages/dev/core/src/Misc/fileTools.ts
+++ b/packages/dev/core/src/Misc/fileTools.ts
@@ -247,8 +247,9 @@ export const LoadImage = (
     const noOfflineSupport = () => {
         LoadFile(
             url,
-            (data) => {
-                const blob = new Blob([data]);
+            (data, _, contentType) => {
+                const type = !mimeType && contentType ? contentType : mimeType;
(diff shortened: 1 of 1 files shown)


RaananW (approved):
👍

packages/dev/core/src/Misc/fileTools.ts
```diff
@@ -357,7 +358,7 @@ export const ReadFile = (
 // eslint-disable-next-line @typescript-eslint/naming-convention
 export const LoadFile = (
     fileOrUrl: File | string,
-    onSuccess: (data: string | ArrayBuffer, responseURL?: string) => void,
+    onSuccess: (data: string | ArrayBuffer, responseURL?: string, contentType?: Nullable<string> | undefined) => void,
```
RaananW, line 361:
```suggestion
    onSuccess: (data: string | ArrayBuffer, responseURL?: string, contentType?: string) => void,
```

I guess this would be enough, no?

carolhmj: The Nullable is required but you're right there's no need for the undefined 😅

PR #12913, opened 2022-08-29
Generate sampler name in buildBlock

carolhmj:
Related issue: https://forum.babylonjs.com/t/nme-bug-about-screen-depth/33404

### DIFF
--- a/packages/dev/core/src/Materials/Node/Blocks/Dual/sceneDepthBlock.ts
+++ b/packages/dev/core/src/Materials/Node/Blocks/Dual/sceneDepthBlock.ts
@@ -12,8 +12,6 @@ import type { Effect } from "../../../effect";
 
 declare type NodeMaterial = import("../../nodeMaterial").NodeMaterial;
 
-let ID = 0;
-
 /**
  * Block used to retrieve the depth (zbuffer) of the scene
  * @since 5.0.0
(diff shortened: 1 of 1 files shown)


RaananW (approved):

packages/dev/core/src/Materials/Node/Blocks/Dual/sceneDepthBlock.ts
```diff
@@ -184,6 +186,7 @@ export class SceneDepthBlock extends NodeMaterialBlock {
     protected _buildBlock(state: NodeMaterialBuildState) {
         super._buildBlock(state);
 
+        this._samplerName = "sceneDepthSampler" + ID++;
```
Popov72, line 189: That would be better to use something like:

```
this._samplerName = state._getFreeVariableName(this.name + "Sampler");
```
It is what `TextureBlock` is doing.

Popov72 (approved):

packages/dev/core/src/Materials/Node/Blocks/Dual/sceneDepthBlock.ts
carolhmj: Thanks for the feedback, updated it! :D

PR #12914, opened 2022-08-30
Clamp negative rgb values to zero to avoid parse errors in native

docEdub:
When setting a canvas's fill style to `"rgb(x, y, z)"`, browsers clamp negative x/y/z values to zero but native throws parse errors like "Exception in HostFunction: Unable to parse color: rgb(-51, 13, -128)".

This change fixes the issue by clamping negative rgb values to zero before using them in the style's rgb function.


deltakosh (approved):

bghgary (commented):
We should fix this on the native side. It's okay to also make this change, but the real fix should be in the native side.

PR #12920, opened 2022-08-30
Font Family Drowpdown

amritanarasimhan:
Created a dropdown menu for font family options, with the ability to add custom fonts beyond the ones already provided


packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +83,103 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
         const store = this.props.extractValue ? this.props.extractValue(this.props.target) : this.props.target[this.props.propertyName];
 
         if (!this.props.noDirectUpdate) {
             this.props.target[this.props.propertyName] = this._remapValueOut(value as number);
         }
-        this.setState({ value: value });
-
-        if (this.props.onSelect) {
-            this.props.onSelect(value);
+        if (value != 0) {
```
carolhmj, line 108: Some things have caught my attention here. First, comparing to 0 using == (doesn't take types into account) instead of === (takes types into account) has some consequences:
```javascript
> false == 0
< true
> false === 0
< false
> '' == 0
< true
> '' === 0
< false
```
Is this intentional?

Second, I'm wondering about the need for this if. It means that we will never set the value 0 on the state?

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +83,103 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
+                    <div className="label" title={this.props.label}>
+                        {this.props.label}
+                    </div>
+                    <div className="options">
+                        <input
+                            type="text"
+                            placeholder="Enter a custom font here"
+                            id="customFont"
+                            onKeyDown={(event) => {
+                                event.keyCode === 13 && this.props.addVal != undefined
```
carolhmj, line 142: Keycode is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode, it's recommended to use code instead: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

But imho I prefer key https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key as its easier to read (it's easier to check key === "A" than remember the key code for A is 65 )

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +83,103 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
+                                    ? (this.props.addVal(
+                                          { label: (document.getElementById("customFont") as HTMLInputElement).value, value: this.props.options.length + 1 },
```
carolhmj, line 144: Nit: Instead of using document.getElementById to get the input, you can just do event.target.value

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +83,103 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
+            );
+        } else {
+            return (
+                <div className={"listLine" + (this.props.className ? " " + this.props.className : "")}>
+                    {this.props.icon && <img src={this.props.icon} title={this.props.iconLabel} alt={this.props.iconLabel} color="black" className="icon" />}
+                    <div className="label" title={this.props.label}>
+                        {this.props.label}
+                    </div>
+                    <div className="options">
+                        <select onChange={(evt) => this.updateValue(evt.target.value)} value={this.state.value === -1 || null || undefined ? 1 : this.state.value}>
```
carolhmj, line 171: `this.state.value === -1 || null || undefined` will always return the result of this.state.value === -1 , do you mean to check if this.state.value is null or undefined?

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +83,103 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
+                            {this.props.options.map((option, i) => {
+                                return (
+                                    <option selected={option.selected} key={option.label + i} value={option.value} title={option.label}>
```
carolhmj, line 174: Are the option.labels unique? If so, there is no need to add the index on the key, and it might even cause problems: https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/

packages/tools/guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -563,7 +648,27 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
+                                propertyName="fontFamily"
+                                options={fonts}
+                                addVal={this.addVal}
+                                valueProp={this.state.value}
+                                onSelect={(newValue) => {
+                                    const fontFamily = this.state.fontFamilyOptions.filter(({ value }) => value === newValue).map(({ label }) => label);
+                                    proxy.fontFamily = fontFamily[0];
+                                }}
+                                extractValue={() => {
+                                    switch (proxy.fontFamily) {
```
carolhmj, line 664: An if-else instead of a switch would be more readable

carolhmj (changes requested):
Awesome job! Just have some considerations to make.

carolhmj:
Doing some testing, I caught this case where if we change a font style and go back to normal style, the style dropdown disappears:
font-dropdown

carolhmj:
Also, if I click on the custom font, click on the textbox and click away without typing anything, it fires the "font doesn't exist in browser" warning. It would be better for the user if this warning didn't fire in case there isn't anything on the input:
image

carolhmj:
When clicking away from the font custom input if the input is invalid, the text block "jumps"
https://user-images.githubusercontent.com/6002144/187727609-6b79a01f-9aa9-467a-8c92-d3922c7fe7ee.mp4

carolhmj:
If I set a custom font to an element and then select another element, the custom font is still selected in the dropdown:
font-dropdown-2

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: I did mean to take types into account. Would that be !== instead of != then?

And yes, we do not want the value of state to ever be 0. What the value of state is used for is to change the font to whatever font is correlated with that value. However, in my list, I have made the 'Custom Font' option correlated with the value 0. That means that without this check, when I click on Custom Font, the font of the selected text changes from, what it originally was to a default font (since Custom Font is not the name of a valid font). In order to prevent that from happening, if that value is 0, I simply do not change the value of state.

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: I did try that originally, but I got the error "Property 'value' does not exist on type 'EventTarget'"

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: Yes, I did! Thanks for catching that

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: option.labels are unique, but this isn't code that I wrote and I wanted to keep it as close as possible to the orignal. Would you recommend removing the index and keeping it as key = {option.label}?

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
Link2Twenty: (event.target as HTMLInputElement).value

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
carolhmj: Hmmm let's keep as it is then so we don't change too much at a time

carolhmj:
When opening this layout: #BC81Y2#3 and clicking on the "Sales" text label, I got a crash:
image

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
carolhmj: > I did mean to take types into account. Would that be !== instead of != then?

Yes! 

Understand your reasoning now, thanks for explaining! 👍

carolhmj (changes requested):
Getting closer! Just the remaining conversations and issues to solve 😄

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +88,117 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
-        this.setState({ value: value });
 
-        if (this.props.onSelect) {
-            this.props.onSelect(value);
+        //selecting a regular option from font dropdown
+        if (value != 0 && this.props.fromFontDropdown) {
+            this.setState({ value: value });
+            if (this.props.onSelect) {
+                this.props.onSelect(value);
+            }
```
RaananW, line 118: Looks like this code is being called in any case. Can we extract it to after the if/else if/else loops? you can also do something like this:

`this.prop.onSelect && this.prop.onSelect(value)`

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +88,117 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
+
     render() {
-        return (
-            <div className={"listLine" + (this.props.className ? " " + this.props.className : "")}>
-                {this.props.icon && <img src={this.props.icon} title={this.props.iconLabel} alt={this.props.iconLabel} color="black" className="icon" />}
-                <div className="label" title={this.props.label}>
-                    {this.props.label}
+        if (this.state.addCustom) {
+            return (
+                <div className={"listLine" + (this.props.className ? " " + this.props.className : "")}>
```
RaananW, line 145:
```suggestion
                <div className={`listLine ${this.props.className ?? ""}`}>
```

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +88,117 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
+                    </div>
+                    <div className="options">
+                        <input
+                            type="text"
+                            placeholder="Enter a custom font here"
+                            id="customFont"
+                            onKeyDown={(event) => {
+                                event.key === "Enter" && this.props.addVal != undefined
+                                    ? (this.props.addVal(
+                                          { label: (document.getElementById("customFont") as HTMLInputElement).value, value: this.props.options.length + 1 },
```
RaananW, line 158: can this be fetched using a (react) reference? Otherwise this will require the application using it to have this specific id instead of passing a reference to the right html element.

Also - the id limits the data to be used (i.e. customFont is confusing. This can be used for other types of values, but this suggests it's only for custom fonts).

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -80,44 +88,117 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
-                        })}
-                    </select>
+            );
+        } else {
+            return (
+                <div className={"listLine" + (this.props.className ? " " + this.props.className : "")}>
+                    {this.props.icon && <img src={this.props.icon} title={this.props.iconLabel} alt={this.props.iconLabel} color="black" className="icon" />}
+                    <div className="label" title={this.props.label}>
+                        {this.props.label}
+                    </div>
```
RaananW, line 185: Those lines are repeated in both if and else. Can you extract them out of those loops?

packages/tools/guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
```diff
@@ -161,6 +198,54 @@ export class CommonControlPropertyGridComponent extends React.Component<ICommonC
...
+    public addVal = (newVal: { label: string; value: number }, prevVal: number) => {
+        if (newVal.label === "") {
+            return;
+        }
+        (async () => {
+            await document.fonts.ready;
+            let displayVal = false;
+            const fonts = JSON.parse(String(window.sessionStorage.getItem("fonts")));
+            if (!(fonts.find((element: IInspectableOptions) => element.label.toLowerCase() === newVal.label.toLowerCase()) === undefined)) {
+                alert("This font is already available");
```
RaananW, line 210: @PatrickRyanMS - are we fine with alerts? or do we have a different system for messages (like a toast at the bottom?)

RaananW (changes requested):
Great work :-) And a very important PR.

Just have a few suggestions and changes. I will also wait for @carolhmj to approve it as well.

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: This.props.onSelect(value) is not called in the case that this.props.fromFontDropdown is true (this.state.value and value are different). I tried to combine the if and the else here since that code is called in both those cases, but when I do that the dropdown stops functioning properly (I am unable to click on any of the options)

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: I fixed this! I think Carol had a similar change so I don't use id anymore

amritanarasimhan: I just haven't committed yet

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: What would the syntax for that be? If I extract that part it's missing a closing div tag, but if I add that in then I don't think I'd be able to use that within the if else statements right? Since technically there is no closing div tag for the first div because there's more code added within it?

packages/tools/guiEditor/src/components/propertyTab/propertyGrids/gui/commonControlPropertyGridComponent.tsx
amritanarasimhan: I asked Patrick about it and I believe he said that alerts were the best option for this!

RaananW:
There is a formatting issue, and one more when the editor is in pop up mode (in the playground).

The alerts are being triggered on the original window host and not on the popup window. You need to use `this.props.globalState.hostWindow.alert` instead of only `alert`

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
```diff
@@ -44,20 +48,32 @@ export class OptionsLineComponent extends React.Component<IOptionsLineComponentP
...
             nextState.value = newValue;
+
+            return true;
+        }
+
+        if (this.props.options !== nextProps.options) {
+            return true;
+        }
+
+        if (this.props.options !== nextProps.options) {
```
carolhmj, line 73: Isn't this a repeat of the above condition?

carolhmj (changes requested):
Just a small question :D

packages/dev/sharedUiComponents/src/lines/optionsLineComponent.tsx
amritanarasimhan: It is! That must have happened while solving the merge conflicts. I've removed it!

carolhmj (approved):
Great job! Thank you so much for your work! 🥳

RaananW (approved):

PR #12930, opened 2022-08-31
Color balancing and dithering for image processing

kircher1:
Two new effects for image process:
1) A simple RGB color balance which can be used for white balancing or otherwise adjusting the color cast of the post process. It attempts to match the luminance of the color prior to balancing by rescaling the resulting color. That works mostly except for at the extremes.
2) A dither that can be applied for de-banding. This is a cheaper alternative to the more artistic grain, which requires a separate render target pass currently.

Corresponding entries have been added into the tsx UI files, and I've been testing this via the playground -> enable the inspector -> add the default rendering pipeline.


sebavan (approved):
Thanks a lot !!!

kircher1:
One thought I had, the color balance color is actually inverted internally before uploading to the shader... this may be unclear to the end user. I think a better approach would be to not invert the color and let the client do that if needed.

sebavan:
I would tend to agree, would you mind doing the change ?

kircher1:
Yeah, I'll take a look. Thanks for the fast response!

kircher1:
It's applying scale directly now, but I wonder if balance is a good term anymore... Let me take the PR back for now and do some more research into the color math behind this. (The dither is working great though so I can spin that off into a different PR.)

PR #12932, opened 2022-09-01
Add dithering effect to image processing.

kircher1:
This can be used to reduce banding in final output. It's good to have it as part of the image processing functions, as the more artistic alternative to this would be the grain post process which requires a separate pass and thus has more overhead. If all you're looking for is to reduce banding, this will be a simple/efficient option.

I've been testing this change via the inspector -> add a default rendering pipeline -> enable dithering.


carolhmj:
Hey there! Great contribution! I tried following your steps, but enabling dither doesn't seem to be having any effect on my test scene: https://playground.babylonjs.com/#IQPBS4#26. When looking at the shader, I see the dither function was defined, but it doesn't seem to be called:
image

sebavan (approved):

sebavan:
Looks working for me :-) 
image

PR #12936, opened 2022-09-01
Reducing amount of code in Observable and Logger

RaananW:


deltakosh (approved):

sebavan (approved):

packages/dev/core/src/Misc/logger.ts
```diff
@@ -55,23 +63,14 @@ export class Logger {
...
-                    Logger.Log(Logger.MessageLimitReached.replace(/%LIMIT%/g, "" + entry.limit).replace(/%TYPE%/g, "log"));
-                    break;
-                case 1:
-                    Logger.Warn(Logger.MessageLimitReached.replace(/%LIMIT%/g, "" + entry.limit).replace(/%TYPE%/g, "warning"));
-                    break;
-                case 2:
-                    Logger.Error(Logger.MessageLimitReached.replace(/%LIMIT%/g, "" + entry.limit).replace(/%TYPE%/g, "error"));
-                    break;
-            }
+            Logger.Log(Logger.MessageLimitReached.replace(/%LIMIT%/g, "" + entry.limit).replace(/%TYPE%/g, type.name ?? ""));
```
Popov72, line 73: Shouldn't you use `Logger.Log` / `Logger.Warn` / `Logger.Error` depending on the level?

RaananW: i'll fix that, great catch!

PR #12937, opened 2022-09-01
Fix loading of base64 svgs

carolhmj:


packages/dev/core/src/Misc/fileTools.ts
```diff
@@ -671,6 +673,16 @@ export const IsBase64DataUrl = (uri: string): boolean => {
     return Base64DataUrlRegEx.test(uri);
 };
 
+export const TestBase64DataUrl = (uri: string): { match: boolean; type: string } => {
+    const results = Base64DataUrlRegEx.exec(uri);
```
sebavan, line 677: Can you replace with IsBase64DataUrl ?

sebavan: should it also check for length ?

sebavan (changes requested):

packages/dev/core/src/Misc/fileTools.ts
carolhmj: I created a new method because besides testing for the regex I need the value of it, which exec returns

RaananW (approved):

sebavan (approved):

PR #12942, opened 2022-09-02
Add polar and spherical coordinate system support

dr-vortex:
Also fixes a misspelling with the `Vector3.GetAngleBetweenVectors` jsdoc parameters


packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1661,11 +1700,39 @@ export class Vector3 {
...
+		return new Vector3(x, y, z);
+	}
+
+	/**
+	* Get rotation between two vectors
+	* @param origin defines the starting point
+	* @param target defines the ending point
+	* @returns the rotation between the vectors
+	*/
+	public static GetRotationBetweenVectors(origin: DeepImmutable<Vector3>, target: DeepImmutable<Vector3>): Vector3{
```
sebavan, line 1723: what is the difference with getAngleBetweenVectors ? it sounds confusing to have both ?

dr-vortex: getAngleBetweenVectors returns only 1 angle, relative to a plane/normal. GetRotationBetweenVectors on the other hand returns both angles, without a normal.

For example, setting the rotation of a Node to the values returned by giving GetRotationBetweenVectors the node's position and a target point would have it facing the target point.

So, getAngleBetweenVectors will only return the angle relative to *one* given axis, while GetRotationBetweenVectors returns the angle**s** relative to all 3 axes.

BabylonJSGuide (changes requested):
Returning polar coordinates as a Vector2 is confusing. If Polar was to be implemented it should have its own class with properties Polar.radius, Polar.angle and methods Polar.toVector2, Polar.fromVector2. However since polar coordinates can be obtained simply from a Vector 2, V,  using

radius = V.length();
angle = Math.sign(V.x) * Math.acos(V.x / radius)  for  -&pi; &le; angle &le; &pi; and add &pi; if you want 0 &le; angle &le; 2&pi;

the question is whether it is worth doing.

BabylonJSGuide (changes requested):
Again returning polar (actually spherical)  coordinates as a Vector3 is confusing and would require a new class Spherical with properties radius, theta, phi  where theta is the inclination and phi the azimuth together with appropriate methods.

Obtaining spherical from a Vector3 V could be achieved by

radius = V.length();
theta = Math.acos(V.z / radius)  0 &le; theta &le; &pi;

getPhi would need a function
```javascript
const getPhi = (V) => {
    if (V.x === 0) {
        if (V.y > 0) {
            return Math.PI / 2
        }
        else if (V.y < 0) {
            return -Math.PI / 2;
        }
        else return NaN;
    }
    if (V.x > 0 ) {
        return Math.atan(V.y / V.x)
    }
    if (V.y >= 0) {
            return Math.atan(V.y / V.x) + Math.PI;
    }
    else {
        return Math.atan(V.y / V.x) - Math.PI
    }
}
```
Whether implementing  or not there is no need need for the extra and confusing getRotationBetweenVectors

dr-vortex:
@BabylonJSGuide,

I've updated the code to have separate classes for both coordinate systems

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -482,6 +490,28 @@ export class Vector2 {
...
+	 * @param vector1 defines 2nd vector
+	 * @returns the angle between vector0 and vector1
+	 */
+	public static GetAngleBetweenVectors(vector0: Vector2, vector1: Vector2): Number{
+		let diff = vector1.subtract(vector0);
+		return Math.atan(diff.y / diff.x);
+	}
+
+	/**
+	 * Converts a polar coordinate Vector2 to its rectangular coordinates
```
BabylonJSGuide, line 505: Converts a polar coordinate to its rectangular coordinates

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -482,6 +490,28 @@ export class Vector2 {
...
+	 * @param polar the polar coordinates
+	 * @returns the rectangular coordinates
+	 */
+	public static fromPolar(polar: Polar): Vector2{
```
BabylonJSGuide, line 509: Static so needs Pascal Case

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -793,6 +823,42 @@ export class Vector2 {
...
+/**
+ * Class used to store (r, theta) vector representation
+ */
+export class Polar {
+	public length: number;
+	public theta: number;
+
+	/**
+	 * Creates a new Polar object
+	 * @param length the length of the vector
```
BabylonJSGuide, line 835: Polar terminology is to use radius for the length

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -98,6 +98,14 @@ export class Vector2 {
         return result;
     }
 
+	/**
+	 * Gets the polar coordinates of the current Vector2
+	 * @returns a new Polar with the polar coordinates
+	 */
+	public toPolar(): Polar{
```
Popov72, line 105: A `toPolarToRef(result: Polar)` would also be nice to have, to avoid creating a new `Polar` object each time.

packages/dev/core/src/Maths/math.vector.ts
Popov72, line 509: A `fromPolarToRef(result: Vector2)` would also be nice to have, to avoid creating a new `Vector2` object each time.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -943,6 +1009,15 @@ export class Vector3 {
         return Quaternion.RotationYawPitchRoll(this._y, this._x, this._z);
     }
 
+	/**
+	 * Converts the current Vector3 to its polar coordinates
+	 * @returns the current polar coordinates (r, theta, phi)
+	 */
+	public toPolar(): Vector3{
```
Popov72, line 1016: Same comment than for `Vector2` => add a `toPolarToRef(result: Vector3)` function.

More generally, for all new functions added, you should also add a `xxxToRef` function that takes as the last parameter the result object (see also `getRotationToVector` / `toSpherical` / `fromSpherical` / `RotationBetweenVectors`).

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -793,6 +823,42 @@ export class Vector2 {
...
+	public toVector2(): Vector2{
+		return Vector2.fromPolar(this);
+	}
+
+	/**
+	 * Converts a given Vector2 to its polar coordinates
+	 * @param v the Vector2 to convert
+	 * @returns a Polar
+	 */
+	public static fromVector2(v: Vector2): Polar{
```
BabylonJSGuide, line 856: Static so Pascal Case need.

Does not give correct values for theta in all cases - see https://playground.babylonjs.com/#ZUFVG7

For correct values of theta see - https://playground.babylonjs.com/#ZUFVG7#1
no need for GetAngleBetweenVectors

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -943,6 +1009,15 @@ export class Vector3 {
         return Quaternion.RotationYawPitchRoll(this._y, this._x, this._z);
     }
 
+	/**
+	 * Converts the current Vector3 to its polar coordinates
```
BabylonJSGuide, line 1013: spherical not polar

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide, line 1016: Needs checking that correct values are given for theta and phi

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -2735,6 +2839,53 @@ export class Vector3 {
...
+	/**
+	 * Gets the rotation on all axes between two vectors
+	 * @param vector0 defines the first vector
+	 * @param vector1 defines the second vector
+	 * @param offset defined the amount to offset the output rotation
+	 * @returns a new Vector3 in the form (pitch, yaw, roll)
+	 * Output roll without offset will always be 0
+	 */
+	public static RotationBetweenVectors(vector0: DeepImmutable<Vector3>, vector1: DeepImmutable<Vector3>, offset: DeepImmutable<Vector3>): Vector3{
+		let diff: Vector3 = vector1.subtract(vector0),
```
Popov72, line 2852: Avoid creating a new vector, use a temporary `Vector3` from `TmpVectors` instead.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -482,6 +490,28 @@ export class Vector2 {
...
     }
 
+	/**
+	 * Gets the angle between 2 vectors
+	 * @param vector0 defines 1st vector
+	 * @param vector1 defines 2nd vector
+	 * @returns the angle between vector0 and vector1
+	 */
+	public static GetAngleBetweenVectors(vector0: Vector2, vector1: Vector2): Number{
+		let diff = vector1.subtract(vector0);
```
Popov72, line 500: Avoid creating a new vector, use a temporary `Vector2` from `TmpVectors` instead.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -943,6 +1009,15 @@ export class Vector3 {
         return Quaternion.RotationYawPitchRoll(this._y, this._x, this._z);
     }
 
+	/**
+	 * Converts the current Vector3 to its polar coordinates
+	 * @returns the current polar coordinates (r, theta, phi)
+	 */
+	public toPolar(): Vector3{
+		let angles = Vector3.GetRotationBetweenVectors(Vector3.Zero(), this);
```
BabylonJSGuide, line 1017: GetRotationBetweenVectors does not exist seems to be now called RotationBetweenVectors and as a method should be camel case

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1641,6 +1716,23 @@ export class Vector3 {
         return this;
     }
 
+	/**
+	 * Gets the rotation on all axes from the current vector to a vector
+	 * @param vector the target vector
+	 * @returns the rotation in the form (pitch, yaw, roll)
+	 */
+	public getRotationToVector(vector: DeepImmutable<Vector3>): Vector3 {
+		return Vector3.RotationBetweenVectors(this, vector);
```
BabylonJSGuide, line 1725: here RotationBetweenVectors is being used as a Vector3 method with two parameters

Line 2851 it is coded as a static with three parameters

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -2735,6 +2839,53 @@ export class Vector3 {
...
+ 
+	/**
+	 * Gets the rotation on all axes between two vectors
+	 * @param vector0 defines the first vector
+	 * @param vector1 defines the second vector
+	 * @param offset defined the amount to offset the output rotation
+	 * @returns a new Vector3 in the form (pitch, yaw, roll)
+	 * Output roll without offset will always be 0
+	 */
+	public static RotationBetweenVectors(vector0: DeepImmutable<Vector3>, vector1: DeepImmutable<Vector3>, offset: DeepImmutable<Vector3>): Vector3{
```
BabylonJSGuide, line 2851: RotationBetweenVectors here is a static but called on line 1725 as a Vector3 method.

BabylonJSGuide:
As can been seen the use of 'GetAngleBetweenVectors' and 'GetRotationBetweenVectors' or is it 'RotationBetweenVectors' or are they the same or different, whether static or a method , over complicates the coding and are unnecessary for Polar or Spherical coordinates. 

toSpherical could be achieved by including a, probably smarter version of the getPhi, function above simplifying the code.

dr-vortex:
@BabylonJSGuide and @Popov72,

Thank you both for all the feedback. I've added toRef functions for all the related functions.

As for the (so far) controversial `RotationBetweenVectors`, the function is very useful for determining the amount of rotation on all axes between to Vector3s. `Vector3.GetAngleBetweenVectors` doesn't do this, it instead finds only 1 angle relative to a plane, which complicates things on the user's end. RotationBetweenVectors is designed as an easy way for developers to get the full rotation without doing 3D geometry.

As for the value of phi, `RotationBetweenVectors` returns the correct values in all cases I've tried. The function does not convert co-terminal angles back to their 0 < θ < 2π form and treats 0/0 as 0 not NaN, which is intended.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -2735,6 +2919,92 @@ export class Vector3 {
...
+	/**
+	 * Gets the rotation on all axes between two vectors and assigns it to the result
+	 * @param vector0 defines the first vector
+	 * @param vector1 defines the second vector
+	 * @param offset defines the amount to offset the output rotation
+	 * @param result the vector to apply the result
+	 * @returns the updated result in the form (pitch, yaw, roll)
+	 * Output roll without offset will always be 0
+	 */
+	public static RotationBetweenVectorsToRef(vector0: DeepImmutable<Vector3>, vector1: DeepImmutable<Vector3>, offset: DeepImmutable<Vector3>, result: DeepImmutable<Vector3>): Vector3{
```
BabylonJSGuide, line 2932: offset is never used, output roll is always 0

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -2735,6 +2919,92 @@ export class Vector3 {
...
+	/**
+	 * Gets the rotation on all axes between two vectors
+	 * @param vector0 defines the first vector
+	 * @param vector1 defines the second vector
+	 * @param offset defines the amount to offset the output rotation
+	 * @returns a new Vector3 in the form (pitch, yaw, roll)
+	 * Output roll without offset will always be 0
+	 */
+	public static RotationBetweenVectors(vector0: DeepImmutable<Vector3>, vector1: DeepImmutable<Vector3>, offset: DeepImmutable<Vector3>): Vector3{
+		let rotation = new Vector3();
```
BabylonJSGuide, line 2950: another place for TmpVectors.Vector3[0];

BabylonJSGuide:
RotationBetweenVectorsToRef

Beginning to understand what this is for. Would you confirm that its use is given an object moving in one vector direction to find the rotation in terms of pitch, yaw, roll needed to change the objects direction to move along a second vector direction?

Not sure what offset is for and as it is not used within the code how it affects the roll.

I would use the following description and name it 'PitchYawRollForDirectionChange' rather than  'RotationBetweenVectors'. Will be interested to see what @sebavan, @Popov72  and others think

```
/**
	 * Gets the pitch, yaw, roll needed to change direction from one vector to another
	 * @param vector0 defines the first direction
	 * @param vector1 defines the second direction
	 * @param offset defines the amount to offset the output rotation ???????????????
	 * @param result the rotation (pitch, yaw, roll) to apply
	 * @returns the updated result in the form (pitch, yaw, roll)
	 * Output roll without offset will always be 0 ???????? and with offset ?????????
	 */
```

BabylonJSGuide:
I tried to understand the RotationBetweenVectors by creating a PG https://playground.babylonjs.com/#ZUFVG7#4.

My expectation was using the y axis and V as the vectors the asymetric cone would rotate to the V which it does not, https://playground.babylonjs.com/#ZUFVG7#5

It could well be that my usage and expectations are wrong. Would you use these PGs as a basic to demonstrate your expectations and usage?

Based on my expectations of obtaining the rotation that changes an objects direction from one to another I came up with
https://playground.babylonjs.com/#QYA3JC but again my expectations of your function may be wrong.

BabylonJSGuide:
For spherical coordinates we have both been careless and used formulas for theta and phi based on z being the vertical hence swapping around inclination and  azimuth and not being clear about their use.

To be consise
Using the standard Babylon.js axes with y vertical and origin at O, let OP be a position vector ( x, y z). Let Q be the projection of P onto the xy plane
Let theta be the inclination of OP from the positive y axis with 0 &le; theta &le; &pi;
Let phi be the azimuth, the angle OQ makes from the positive x axis in a counter clockwise direction with -&pi; &le; phi &le; &pi;

Then we can simply use the function (made Typescript and public, inlude ToRef, etc.)

Looks like I got sidetracked as well.

```javascript
SphericalFromVector3 = (v) => {
        const radius = v.length();
        const theta = Math.acos(v.y / radius);
        const phi = Math.atan2(v.z,  v.x); //use of atan2 give range -&pi; to &pi;
        return {radius: radius, theta: theta, phi: phi}
}
```

as in https://playground.babylonjs.com/#QYA3JC#1

sebavan (changes requested):
I like the idea to have polar support :-)

Can we move the polar class in math.polar.ts ? to not impact size for people not using it ?

I will flag as draft for now until most of the comments are resolved and I ll do a final pass of review after that.

dr-vortex:
@sebavan Would you also want Spherical moved to math.polar.ts?

@BabylonJSGuide 
> Would you confirm that its use is given an object moving in one vector direction to find the rotation in terms of pitch, yaw, roll needed to change the objects direction to move along a second vector direction?

Yes, that is what the function does, and that is a great explanation.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -98,6 +99,24 @@ export class Vector2 {
         return result;
     }
 
+	/**
```
sebavan, line 102: Can we remove those two functions ?

The goal is to only import math.polar as a type to prevent the addition of polar in math.vector when unsused ?

sebavan: ppl could use the Polar functions for this.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1641,6 +1708,42 @@ export class Vector3 {
...
+	public getRotationToVector(vector: DeepImmutable<Vector3>): Vector3 {
+		return Vector3.RotationBetweenVectors(this, vector, Vector3.ZeroReadOnly);
+	}
+
+	/**
+	 * Gets the spherical coordinates of the current vector
+	 * @param ref the reference Spherical to assign the result
+	 * @returns the spherical coordinates for the current vector
+	 */
+	public toSphericalToRef(ref: Spherical): Spherical {
```
sebavan, line 1735: Same comment as earlier, let s keep those on the spherical side to prevent the extra import

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -482,6 +501,54 @@ export class Vector2 {
...
+		return result;
+	}
+
+	/**
+	 * Converts a polar coordinate to its rectangular coordinates
+	 * @param polar the polar coordinates
+	 * @param ref the reference to assign the result
+	 * @returns the reference Vector2
+	 */
+	 public static FromPolarToRef(polar: Polar, ref: Vector2): Vector2{
```
sebavan, line 536: Could this be add to the polar code instead ?

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1661,11 +1764,34 @@ export class Vector3 {
         return s;
     }
 
+	/**
+	 * Gets the (x, y, z) representation for a set of spherical coordinates
+	 * @param spherical defines the spherical coordinates
+	 * @param ref defines the target to assign the coordinates to
+	 * @returns the updated Vector3
+	 */
+	 public static FromSphericalToRef(spherical: Spherical, ref: DeepImmutable<Vector3>): Vector3{
```
sebavan, line 1773: Could this be add to the spherical code instead ?

BabylonJSGuide:
> I've tested this in the below PG and it does not work correctly:
> https://playground.babylonjs.com/#2I8Y3S#12

I never said it worked when used as part of the 'rotation' methods to change direction of an object from one direction to another. 

It is cleaner and more understandable code for working out spherical coordinates only.

The 'rotation change direction' functions would then stand alone and not be part of polar or spherical. This would also make the math.polar.ts code simpler.

The basis of any 'rotation change direction' code would be from https://playground.babylonjs.com/#QYA3JC and part of math.vector.ts

BabylonJSGuide:
> > Would you confirm that its use is given an object moving in one vector direction to find the rotation in terms of pitch, yaw, roll needed to change the objects direction to move along a second vector direction?
> 
> Yes, that is what the function does, and that is a great explanation.

In which case in terms of this PG https://playground.babylonjs.com/#ZUFVG7#5 how do you use it so it aligns correctly?

Also please explain the intention of the offset vector

packages/dev/core/src/Maths/math.polar.ts
```diff
@@ -0,0 +1,126 @@
...
+	 * @returns a Polar
+	 */
+	public static FromVector2(v: Vector2): Polar{
+		let polar = new Polar();
+		Polar.FromVector2ToRef(v, polar);
+		return polar;
+	}
+}
+
+export class Spherical {
```
BabylonJSGuide, line 69: Standard notation for spherical is radius not length

dr-vortex:
@BabylonJSGuide 
Offset was unneeded - it would just be added to the output rotation

As for the PG / code you showed above, I apologize for the misunderstanding. I had assumed it was for PitchYawRollForDirectionChange

I fixed an issue where it would try to compute phi wrong by using itself in the equation instead of theta, since I hadn't corrected that when renaming some variables.
For clarification (as of the last commit):
phi is the vertical angle, rotation x , and pitch
theta is the horizontal angle, rotation y, and yaw

The reason for ordering it (pitch, yaw, roll) is that Babylon.js uses that format in rotation Vector3s. (that is where the rotation x/y comes from in the above).

I'll look into using `Math.atan2` for spherical conversions, though I don't see a reason to change to using it over the current method.

dr-vortex:
On the topic of https://playground.babylonjs.com/#ZUFVG7#5:

Given a Mesh, M, and a target Vector3, V:
`M.rotation = PitchYawRollForDirectionChange(M.position, V.position);`
should have M facing V

dr-vortex:
Looking at https://playground.babylonjs.com/#ZUFVG7#5, I found the phi equation was incorrect. I've updated the equations, please see this PG for the fixed PitchYawRollForDirectionChange: https://playground.babylonjs.com/#ZUFVG7#6

Also note that in the PG the cone is rolled by -pi/2 so that the cone tip is facing the target instead of the box (since the cone faces upwards by default). 

I noticed that the rotation fetched in both PGs gets the rotation from yAxis to the target instead of the mesh's position to the target, which would not give the intended result.

packages/dev/core/src/Maths/math.polar.ts
```diff
@@ -0,0 +1,125 @@
...
```
sebavan, line 69: Missing the typedoc comment :-)

sebavan (commented):
Looks all good to me in organization GG !!!!

I let the maths part to you and @BabylonJSGuide :-) ping me once ready and I can merge it :-)

BabylonJSGuide:
Unfortunately I won't be able to give a thorough check until tomorrow or the weekend.

BabylonJSGuide:
> I'll look into using `Math.atan2` for spherical conversions, though I don't see a reason to change to using it over the current method.

Math.atan gives results from -&pi;/2 to &pi;/2 so get same results for y/x and -y/-x

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan

Mat.atan2 gives results from -&pi; to &pi; and because y  and x are passed separately y, x and -y, -x do not give same result.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2.

There are some other issues which I do not have time to investigate until tomorrow so I would  wait to do any more commits until my next review.

dr-vortex:
Thanks! I will update the atan to be atan2 and commit that but wait until tomorrow to do any other changes.

packages/dev/core/src/Maths/math.polar.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Class used for (radius, theta, phi) vector representation.
+ */
+export class Spherical {
+	public radius: number;
+	public theta: number;
+	public phi: number;
+
+	constructor(radius, theta, phi){
```
BabylonJSGuide, line 78: Physics convention theta is vertical angle and phi is horizontal angle.
Maths convention theta is horizontal angle and phi is vertical angle

From your formulas you have chosen the physics convention, however both conventions have the vertical angle measured from the positive Y axis to the vector not from the positive y axis so a change will be needed in the FromVector formula.

packages/dev/core/src/Maths/math.polar.ts
```diff
@@ -0,0 +1,129 @@
...
+		this.phi = phi;
+	}
+
+	/**
+	 * Assigns the rectangular coordinates of the current Spherical to a Vector3
+	 * @param ref the Vector3 to update
+	 * @returns the updated Vector3
+	 */
+	public toVector3ToRef(ref: DeepImmutable<Vector3>): Vector3{
+		let x = this.radius * Math.sin(this.theta) * Math.cos(this.phi);
```
BabylonJSGuide, line 90: This follows the physics convention and no changes needed.

packages/dev/core/src/Maths/math.polar.ts
```diff
@@ -0,0 +1,129 @@
...
+
+	/**
+	 * Assigns the spherical coordinates from a Vector3
+	 * @param vector the vector to convert
+	 * @param ref the Spherical to update
+	 * @returns the updated ref
+	 */
+	public static FromVector3ToRef(vector: DeepImmutable<Vector3>, ref: Spherical): Spherical{
+		ref.radius = vector.length();
+		ref.theta = Math.atan2(diff.z, diff.x);
```
BabylonJSGuide, line 114: The formula here use the maths convention and need to be consistent with the physics convention. Also the use of sine gives the angle from the positive X axis not the conventional (both maths and physics) from the positive Y axis

formula should be
```
ref.theta = Math.acos(vector.y / r);
ref.phi = Math.atan2(vector.z, vector.x)
```

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -482,6 +483,32 @@ export class Vector2 {
...
     }
 
+	/**
+	 * Gets the angle between 2 vectors
+	 * @param vector0 defines 1st vector
+	 * @param vector1 defines 2nd vector
+	 * @param ref the reference to assign the result
+	 * @returns the updated reference
+	 */
+	 public static GetAngleBetweenVectorsToRef(vector0: Vector2, vector1: Vector2, ref: number): number {
```
BabylonJSGuide, line 493: The calculation for GetAngleBetweenVectorsToRef is not correct. I do not believe this nor GetAngleBetweenVectors for Vector2s is necessary

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -2735,6 +2781,39 @@ export class Vector3 {
...
+	/**
+	 * Gets the pitch, yaw, roll needed to change direction from one vector to another
+	 * @param vector0 defines the first vector
+	 * @param vector1 defines the second vector
+	 * @param offset defines the amount to offset the output rotation
+	 * @param result the rotation (pitch, yaw, roll) to apply
+	 * @returns the updated result in the form (pitch, yaw, roll)
+	 * Output roll will always be 0 + offset.z
+	 */
+	public static PitchYawRollForDirectionChangeToRef(vector0: DeepImmutable<Vector3>, vector1: DeepImmutable<Vector3>, result: DeepImmutable<Vector3>): Vector3{
```
BabylonJSGuide, line 2768: This does not work as specified.

In this PG https://playground.babylonjs.com/#ZUFVG7#6
even though the asym cone points along the given vector V there are two issues 

The current vector along which the asym cone points is the Y axis however this is not passed to the function it is the mesh position (a zero vector) noa direction that is passed and also without a correction of a rotation of -PI / 2 the cone points in the wrong direction. 

My recommendation is to use the following functions. Note as pointed out in https://forum.babylonjs.com/t/rotating-vector-to-align-with-other-vector/33776/9?u=johnk a little more work is needed.

dr-vortex: The intent of the function is not to take into account the current rotation of a mesh, only its position. 

The reason -pi/2 is added is because while the cone points upward by default, the actual rotation is in the direction of the box - not the cone's tip. Hence, when properly aligned to the target, the cone's tip would not face it, the box would.

In https://playground.babylonjs.com/#QYA3JC you use the following functions:
```js
const QuaternionFromTo = (fromVec, toVec) => {
    const normal = BABYLON.Vector3.Cross(fromVec, toVec);
    const angle = BABYLON.Vector3.GetAngleBetweenVectors(fromVec, toVec, normal);
    return BABYLON.Quaternion.RotationAxis(normal, angle);
}

const RotationFromTo = (fromVec, toVec) => {
    return QuaternionFromTo(fromVec, toVec).toEulerAngles();
}
```
Which seems to get the rotation from a rotation Vector3 to a position Vector3 - could you clarify the purpose of them?

dr-vortex: Also, I've attempted to use this with the below PG, but it does not appear to work properly (the box should face the same direction it is moving)
https://playground.babylonjs.com/#2I8Y3S#14

packages/dev/core/src/Maths/math.polar.ts
BabylonJSGuide: Once more the spherical function has nothing to do with the rotationBetween type functions. Using these functions rather than trying out spherical.FromVector3ToRef resolves nothing.

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: > Would you confirm that its use is given an object moving in one vector direction to find the rotation in terms of pitch, yaw, roll needed to change the objects direction to move along a second vector direction?

2 days ago

Yes, that is what the function does, and that is a great explanation.

Today
The intent of the function is not to take into account the current rotation of a mesh, only its position.

@sebavan I cannot approve this PR. Ad I understood the intention there are too many issues to be resolved.

@dr-vortex and I seem to be working across each other. It is not enough to make a comment and say the conversation is resolved.

packages/dev/core/src/Maths/math.polar.ts
dr-vortex: Apologies - I did not think about the formula being only for Spherical.FromVector3ToRef.

packages/dev/core/src/Maths/math.vector.ts
dr-vortex: For clarification:

> The current vector along which the asym cone points is the Y axis however this is not passed to the function it is the mesh position (a zero vector) not direction that is passed and also without a correction of a rotation of -PI / 2 the cone points in the wrong direction.

Why would the current direction of the cone be passed?

dr-vortex: Also, for my earlier comment
> Would you confirm that its use is given an object moving in one vector direction to find the rotation in terms of pitch, yaw, roll needed to change the objects direction to move along a second vector direction?
> Yes, that is what the function does, and that is a great explanation.

I clearly misread the question.

I'm sorry for all the confusion caused, I originally and continue to intend for the function to be based on an origin position rather than an origin direction.

BabylonJSGuide: > Also, I've attempted to use this with the below PG, but it does not appear to work properly (the box should face the same direction it is moving)
> https://playground.babylonjs.com/#2I8Y3S#14

It works here https://playground.babylonjs.com/#1PX9G0#61

BabylonJSGuide:
image   
image

@dr-vortex I obviously do not understand the intention of your function. In the above images the sphere represents the origin, the green line lies along the position vector of the mesh, the diagonal line represents the second parameter vector. The mesh position is the same for both images. After 

```
PitchYawRollForDirectionChange(mesh.position, vector1)
```

is used where will the mesh lie and where will the axis through the point of the cone and the center of its base be in (1) the upper image and (2) the lower image?

packages/dev/core/src/Maths/math.vector.ts
dr-vortex: Interesting...
https://playground.babylonjs.com/#1PX9G0#63

I would expect the smaller box with a position relative to the larger box of Vector3.Forward() to be in the same direction of travel, though it is off to the side. Maybe this is just an oversight?

dr-vortex:
In both images, the cone would face the target vector and the position would remain unchanged. Here is a code explanation that is more "descriptive"
```js
//mesh is the cone,  target is the Vector3

mesh.rotation = PitchYawRollForDirectionChange(mesh.position, target);


let distance = Vector3.Distance(mesh.position, target);

let targetedPosition = mesh.getDirection(Vector3.Forward().scale(distance));

//targetedPosition == target
```

Something that is important to note with the cone is that the tip of the cone should not be facing the target because that is the cone's "up" direction and the function is computing the rotation so that the "forward" direction aligns with the target.

BabylonJSGuide:
> the cone would face the target vector

Do not understand what this means. What part of the cone would face the target? The idea of using a cone is that the point of the cone is pointing to the target. Are you saying that the target can never be on the y axis?

Imagine the cone as some sort of spaceship with the tip pointing forward. So in the lower image the fin is vertical and the spaceship is moving left, the diagonal white line gives the direction for the second vector. Is this a possible scenario?

Using the tip of the cone as that which must face the target please describe or produce an image of a possible scenario.

BabylonJSGuide:
@dr-vortex  I will not be able to add an accepting review until I understand the purpose of your function 'PitchYawRollForDirectionChangeToRef' and that it works as you expect. In the following PG I have set up the asym cone so that the tip of the cone is in the forward position and the fin is up and by baking the transformation it currently has rotation zero. 

Using the tip and fin in your description please describe the purpose of your function and demonstrate how it works and that it does work as expected by adding it to the PG and applying it to the asym cone.

dr-vortex:
Experimenting with the function, the theta function `atan2(z, x)` should be `atan2(x, z)`, and we can use atan2 instead of asin for phi. I've made [this](https://playground.babylonjs.com/#QYA3JC#5) PG that demonstrates the function with an animation.

dr-vortex:
No, the function should work without a plane. The animation just sets the rotation at every increment of the frame rate (-some) to the rotation between the last node and the current node in the path. I believe the issue with the cone not facing in the direction of travel is an oversight of the animation, not the function.

packages/dev/core/src/Maths/math.polar.ts
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	/**
+	 * Assigns the rectangular coordinates of the current Spherical to a Vector3
+	 * @param ref the Vector3 to update
+	 * @returns the updated Vector3
+	 */
+	public toVector3ToRef(ref: DeepImmutable<Vector3>): Vector3{
+		let x = this.radius * Math.sin(this.theta) * Math.cos(this.phi);
+		let y = this.radius  * Math.sin(this.theta) * Math.sin(this.phi);
```
BabylonJSGuide, line 96: Because y is vertical not z y and z need swapping over

ie
```
let z = this.radius * Math.sin(this.theta) * Math.sin(this.phi);
let y = this.radius * Math.cos(this.theta);
```

This is my mistake.

BabylonJSGuide:
The Polar class is working https://playground.babylonjs.com/#B1WXN8
With the change required above Spherical is working https://playground.babylonjs.com/#ML0BEE#1

@dr-vortex my suggestion is that for the moment any changes to Vector2 and Vector3 in the math.vector.ts are removed from the PR which I could then approve. Once you know for sure your functions and animations are working as you require then submit a PR with with the working functions.

@sebavan will know more but in terms of the build failures then you need to look at https://doc.babylonjs.com/divingDeeper/developWithBjs/howToStart and check your local build. I think the common failures are not putting and entry into 'whats new', not commenting on parameters and not having the required format such as extra space, extra lines etc, you can check this with (I think)

```
npm run format:fix
```
as per doc page above.

sebavan:
Thanks both for all the love you provide for this feature !!!

Yup, about the build, the format fix will help a lot, and the rest of the issues are:

image

dr-vortex:
@sebavan 
What settings do you use for prettier? I'd like to format it correctly but don't know what settings to use.

sebavan:
they are enclosed in the project but the simpler is to run 

npm run format:fix

sebavan:
@RaananW can you think of smthg else to help @dr-vortex with the formatting ?

dr-vortex:
Hm...

I cloned my fork, then did `npm run format:fix` which pointed out that some lets should be consts, I fixed that, then ran it again along with format document (Ctrl + Shift + I in VS Code). No changes or errors the second time.

sebavan (approved):

dr-vortex:
@sebavan and @BabylonJSGuide,

Ready to merge, finally! Thank you both for all the help and feedback! I will start a draft PR and work on some operator functions (add, subtract, etc.) for Polar and Spherical.

PR #12943, opened 2022-09-02
Added AudioBuffer as parameter for Sound

sorskoot:
I ran into a situation where I wanted to use the sound engine in Babylon to play my pre-generated Web Audio Buffer. I managed to get it to work by looking at the code, but I had to access some private properties. I noticed that it would be pretty small refactoring to implement it correctly since the code was already there. This PR enable the passing of an AudioBuffer to the constructor of BABYLON.Sound.

This (very simplified) example shows how it is used:

```

    const camera = new BABYLON.ArcRotateCamera("Camera", -Math.PI / 2, Math.PI / 4, 15, BABYLON.Vector3.Zero());

    // Create buffer
    let sineBuffer = BABYLON.Engine.audioEngine.audioContext.createBuffer(1, 12e3, 48e3);
    const hitBufferData = sineBuffer.getChannelData(0);
    for (let i = 12e3; i--;) hitBufferData[i] = Math.cos(i * .04); 

    // Create and use the sound
    this.sineSound = new BABYLON.Sound("sine", sineBuffer); 
    this.sineSound.play();

    return scene;
};
```

### DIFF
--- a/packages/dev/core/src/Audio/sound.ts
+++ b/packages/dev/core/src/Audio/sound.ts
@@ -158,7 +158,7 @@ export class Sound {
     private _registerFunc: Nullable<(connectedMesh: TransformNode) => void>;
     private _isOutputConnected = false;
     private _htmlAudioElement: HTMLAudioElement;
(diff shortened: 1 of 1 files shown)


sebavan (approved):
I love it, thanks

PR #12945, opened 2022-09-03
Optimize animation interpolate function

bghgary:
Not 100% sure this actually helps since we are seeing weird results in the profile, but it is certainly easier to read.


deltakosh (approved):
Not sure what to review as the diff is all over the place but I spotted nothing wrong

bghgary:
> Not sure what to review as the diff is all over the place but I spotted nothing wrong

If you hide whitespace changes, it will be much easier to read :-)

sebavan (approved):

PR #12946, opened 2022-09-05
Clear some values used during loading at the end of the load

carolhmj:
Related issue: https://forum.babylonjs.com/t/imported-meshes-appear-broken-if-they-have-the-same-geometry-uniqueid/33567/3


sebavan (commented):
LGTM

PR #12948, opened 2022-09-05
Use CSS Modules in Graph System

carolhmj:
This is part of the work to make the graph system more reusable.


packages/dev/buildTools/src/webpackTools.ts
```diff
@@ -119,7 +120,9 @@ export const getRules = (
                         loader: "css-loader",
                         options: {
                             sourceMap: true,
-                            modules: true,
+                            modules: {
+                                localIdentName: options.mode === "development" ? "[path][name]__[local]" : "[hash:base64]",
```
sebavan, line 124: This considers prod being the default (if nothing is passed in mode) whereas the mode a bit later in the files default to dev :-)

sebavan (changes requested):
Looks good to me aside of the tiny issue :-) but I ll let @RaananW do the final review

packages/dev/buildTools/src/webpackTools.ts
carolhmj: Goood catch Seb!

RaananW (commented):
Thank you so much for that :-)

I just want to be sure that it doesn't influence the existing tools before merging. Does it change anything in the node editor/gui editor? As it was previously globally applied, it is possible that those classes that were removed are used in one of the tools.

Otherwise - approved, great work

sebavan (approved):
Looks all good to me, @RaananW can we merge it ?

RaananW:
I'll wait with the merge until azure is stable, but yes, we can merge

PR #12952, opened 2022-09-06
Minimize code in effect and pipeline context

RaananW:
The idea here is to keep the typescript declarations of those functions (i.e. - they exist still), but use a proxy to catch missing functions and call the right function when the missing functions are called. 

This will minimize the initial needed code when importing those classes.

Regarding reusability - there might be 2-3 lines of code that are reusable, but the overhead of moving it out of the constructor context will cost more than the code written.


carolhmj:
Nice!!!

sebavan (commented):
I am wondering why using a proxy instead of assigning the functions here ?

Mainly cause I am afraid of the extra getter cost which reduced perf in the past in some situation ?

Popov72:
As @sebavan I'm a bit worried of the additional cost of the indirections (there are more function calls / conditional code and additional string testing for each function call).

Those functions are potentially called a lot of time by `bindForSubMesh` as they setup the shader parameters (`bindForSubMesh` is one of the 2 or 3 functions that take most of the time in a typical frame).

Granted, the performance penalty is only for subsystems that don't support UBO (so WebGL1), as when UBO is supported we don't resort to those functions. So maybe it's acceptable and the cleaner code makes for the perf penalty (if penalty there is, I didn't perform any test)...

deltakosh (changes requested):
I really think this could have a measurable performance impact on heavy scenes with a lot of materials.

sebavan:
I guess a simple for loop to bind the function in the constructor would achieve the same without proxy or indirections

RaananW:
> I guess a simple for loop to bind the function in the constructor would achieve the same without proxy or indirections

Let's discuss this. I am running perf tests now and will have answers for you soon

RaananW:
This is a suggestion without a proxy. Tried to minimize the code, but we need to check how much it is actually saving.

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
```diff
@@ -23,6 +40,42 @@ export class WebGLPipelineContext implements IPipelineContext {
     public programLinkError: Nullable<string> = null;
     public programValidationError: Nullable<string> = null;
 
+    constructor() {
+        const proxyFunction = (functionName: Partial<keyof this>, uniformName: string, ...payload: any[]) => {
```
sebavan, line 44: I wonder if perf wise proxyFunction shouldnt be a function taking the name in parameter and return the right construct so that run time cost would be left fully unchanged ???

deltakosh (approved):

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
```diff
@@ -23,6 +40,49 @@ export class WebGLPipelineContext implements IPipelineContext {
...
 
+    constructor() {
+        const args: any[] = [];
+        const proxyFunction: (functionName: string) => ((/*uniformName: string, ...payload: any[]*/) => void) | undefined = (functionName: string) => {
+            const cacheFunction = cacheToSetProxyReference[functionName as string];
+            if (cacheFunction) {
+                const cacheFunc = this[`_cache${cacheFunction}` as Partial<keyof WebGLPipelineContext>];
+                return function (this: WebGLPipelineContext /*uniformName: string, ...payload: any[]*/) {
+                    const func = this.engine[functionName as keyof ThinEngine];
+                    Array.prototype.push.apply(args, arguments);
```
sebavan, line 51: args needs to be reset if using push ?

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
```diff
@@ -23,6 +40,49 @@ export class WebGLPipelineContext implements IPipelineContext {
...
     public programValidationError: Nullable<string> = null;
 
+    constructor() {
+        const args: any[] = [];
+        const proxyFunction: (functionName: string) => ((/*uniformName: string, ...payload: any[]*/) => void) | undefined = (functionName: string) => {
+            const cacheFunction = cacheToSetProxyReference[functionName as string];
+            if (cacheFunction) {
+                const cacheFunc = this[`_cache${cacheFunction}` as Partial<keyof WebGLPipelineContext>];
+                return function (this: WebGLPipelineContext /*uniformName: string, ...payload: any[]*/) {
+                    const func = this.engine[functionName as keyof ThinEngine];
```
sebavan, line 50: this could be moved before the return

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
```diff
@@ -110,32 +170,45 @@ export class WebGLPipelineContext implements IPipelineContext {
...
+     * @param _uniformName
+     * @param _x
+     * @param _y
+     * @param _z
+     * @param _w
      * @hidden
      */
-    public _cacheFloat2(uniformName: string, x: number, y: number): boolean {
-        let cache = this._valueCache[uniformName];
+    public _cacheFloatN(_uniformName: string, _x: number, _y: number, _z?: number, _w?: number): boolean {
```
sebavan, line 180: should we make _y optional and use it for setFloat and setInt as well ?

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
```diff
@@ -110,32 +170,45 @@ export class WebGLPipelineContext implements IPipelineContext {
...
-        if (cache[0] !== x) {
-            cache[0] = x;
-            changed = true;
-        }
-        if (cache[1] !== y) {
-            cache[1] = y;
-            changed = true;
-        }
-
+        cache.forEach(function (num, idx) {
```
sebavan, line 193: is this not GCing the function on every call ?

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
RaananW: No, it can't,m because engine is assigned after the function is defined.

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
RaananW: yes, i am doing that now

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
RaananW: There is no cache for setInt and setFloat

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
RaananW: good catch, changing to for

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
sebavan: would be great to add no ?

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
sebavan: so should we move this in engine assignment ? or is it overkill ?

RaananW: a bit of overkill for one line that will be ca. 10 characters. We will need to add a setter and getter for engine, add the rest to be members of the class, this will add more code than it saves.

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
RaananW: Added :-)

sebavan (approved):

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
```diff
@@ -110,32 +174,45 @@ export class WebGLPipelineContext implements IPipelineContext {
...
      */
-    public _cacheFloat2(uniformName: string, x: number, y: number): boolean {
-        let cache = this._valueCache[uniformName];
+    public _cacheFloatN(_uniformName: string, _x: number, _y?: number, _z?: number, _w?: number): boolean {
+        /**
+         * arguments will be used to abstract the cache function.
+         * arguments[0] is the uniform name. the rest are numbers.
+         */
+        let cache: number[] = this._valueCache[arguments[0]];
         if (!cache || cache.length !== 2) {
```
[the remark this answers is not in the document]
sebavan: amazing catch

sebavan (changes requested):
preventing the PR to be merged in following @Popov72 comments

packages/dev/core/src/Engines/WebGL/webGLPipelineContext.ts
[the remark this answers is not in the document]
RaananW: should be arguments.length - 1. changed now, thanks!

sebavan (approved):

PR #12954, opened 2022-09-06
Performance mode

deltakosh:
The goal is to protect backward compatibility but also offer an option for users to say they are ok to not be backward compatible (OR less easy to use) to favor perf by default

### DIFF
--- a/packages/dev/core/src/Meshes/abstractMesh.ts
+++ b/packages/dev/core/src/Meshes/abstractMesh.ts
@@ -2,6 +2,7 @@ import { Observable } from "../Misc/observable";
 import type { Nullable, FloatArray, IndicesArray, DeepImmutable } from "../types";
 import type { Camera } from "../Cameras/camera";
 import type { Scene, IDisposable } from "../scene";
+import { ScenePerformancePriority } from "../scene";
 import type { Vector2 } from "../Maths/math.vector";
 import { Quaternion, Matrix, Vector3, TmpVectors } from "../Maths/math.vector";
 import { Engine } from "../Engines/engine";
@@ -822,13 +823,20 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
     constructor(name: string, scene: Nullable<Scene> = null) {
         super(name, scene, false);
 
-        this.getScene().addMesh(this);
+        scene = this.getScene();
+
+        scene.addMesh(this);
 
         this._resyncLightSources();
 
         // Mesh Uniform Buffer.
         this._uniformBuffer = new UniformBuffer(this.getScene().getEngine(), undefined, undefined, name, !this.getScene().getEngine().isWebGPU);
         this._buildUniformLayout();
+
+        if (scene.performancePriority !== ScenePerformancePriority.BackwardCompatible) {
+            this.alwaysSelectAsActiveMesh = true;
+            this.isPickable = false;
+        }
     }
 
     protected _buildUniformLayout(): void {
--- a/packages/dev/core/src/scene.ts
+++ b/packages/dev/core/src/scene.ts
@@ -137,7 +137,7 @@ export enum ScenePerformancePriority {
     BackwardCompatible,
     /** Some performance options will be turned on trying to strike a balance between perf and ease of use */
     Intermediate,
-    /** Performance will be to priority */
+    /** Performance will be top priority */
     Aggressive,
 }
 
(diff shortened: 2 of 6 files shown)


sebavan:
Love it :-)

deltakosh:
One thing I will do after merging this (when we are all ok with it) will be to document the consequences. What you will lose /have to pay to get to the best performance

packages/dev/core/src/scene.ts
```diff
@@ -129,6 +129,18 @@ export interface SceneOptions {
...
 }
 
+/**
+ * Define how the scene should favor performance over ease of use
+ */
+export enum ScenePerformancePriority {
+    /** Default mode. No change. Performance will be treated as less important than backward compatibility */
+    BackwardCompatible,
+    /** Some performance options will be turned on trying to strike a balance between perf and ease of use */
+    Intermediate,
```
bghgary, line 139: Is this used?

bghgary (commented):
I like it. Do we want to maintain backwards compatibility when people set to `Aggressive`? I.e., if later, we decided that more things need to change for `Aggressive` mode, can we update it?

packages/dev/core/src/scene.ts
deltakosh: Not yet, I'm adding some for it now :)

packages/dev/core/src/scene.ts
```diff
@@ -256,6 +268,29 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
+    public set performancePriority(value) {
+        if (value === this._performancePriority) {
+            return;
+        }
+
+        this._performancePriority = value;
+
+        switch (value) {
+            case ScenePerformancePriority.Aggressive:
+                this.skipPointerMovePicking = true;
```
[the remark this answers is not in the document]
deltakosh: Lovely!

deltakosh:
> I like it. Do we want to maintain backwards compatibility when people set to `Aggressive`? I.e., if later, we decided that more things need to change for `Aggressive` mode, can we update it?

Nope, I will quote that in the doc. Aggressive is not backward compatible

packages/dev/core/src/Meshes/abstractMesh.ts
```diff
@@ -822,13 +823,20 @@ export class AbstractMesh extends TransformNode implements IDisposable, ICullabl
...
+        scene.addMesh(this);
 
         this._resyncLightSources();
 
         // Mesh Uniform Buffer.
         this._uniformBuffer = new UniformBuffer(this.getScene().getEngine(), undefined, undefined, name, !this.getScene().getEngine().isWebGPU);
         this._buildUniformLayout();
+
+        if (scene.performancePriority !== ScenePerformancePriority.BackwardCompatible) {
+            this.alwaysSelectAsActiveMesh = true;
```
Popov72, line 837: Do we really want to do that?

I think there could be cases where setting `alwaysSelectAsActiveMesh=true` will actually make perf worse...

Maybe do it only in the **Agressive** mode?

[EDIT] Doing it only in the **Agressive** mode would be moot as there's already `this._skipFrustumClipping=true` in this mode...

packages/dev/core/src/scene.ts
```diff
@@ -256,6 +268,34 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
...
+                this.autoClear = false;
+                break;
+            case ScenePerformancePriority.Aggressive:
+                this.autoClear = false;
+                this.skipPointerMovePicking = true;
+                this.skipFrustumClipping = true;
+                break;
+        }
+    }
+
```
Popov72, line 298: Maybe rewrite it like this to avoid duplicated code:
```typescript
switch (value) {
    case ScenePerformancePriority.Aggressive:
        this.skipFrustumClipping = true;
        // note: intentional fall-through
    case ScenePerformancePriority.Intermediate:
        this.skipPointerMovePicking = true;
        this.autoClear = false;
        break;
}
```

packages/dev/core/src/scene.ts
```diff
@@ -129,6 +129,18 @@ export interface SceneOptions {
...
 
+/**
+ * Define how the scene should favor performance over ease of use
+ */
+export enum ScenePerformancePriority {
+    /** Default mode. No change. Performance will be treated as less important than backward compatibility */
+    BackwardCompatible,
+    /** Some performance options will be turned on trying to strike a balance between perf and ease of use */
+    Intermediate,
+    /** Performance will be to priority */
```
Popov72, line 140: nit: I guess you meant "top" instead of "to"?

Popov72 (approved):
Nice!

packages/dev/core/src/Meshes/abstractMesh.ts
deltakosh: We are massively CPU bound so I truly believe this is required

packages/dev/core/src/scene.ts
deltakosh: good one

sebavan (approved):

PR #12959, opened 2022-09-07
Stop vis test in puppeteer if debug enabled

RaananW:


sebavan (approved):
Should it not be done for webgpu as well ?

RaananW:
> Should it not be done for webgpu as well ?

Gosh, I forgot to push the change. Give me a sec.

PR #12960, opened 2022-09-07
Introduce autoFixFaceOrientation feature to solidParticleSystem

Michalzr:
**What:**
Introduce an option to "autofix face orientation" in solid particle system

**How**
When there is a transformation which changes the orientation (for example scaling `(-1, 1, 1)`), the solid particle system will automatically flip all the faces of the particle affected by this transformation.

**Why**
I needed this functionality in my application, so I decided to make a PR and hope that it will get into the babylon itself :D

**Example**
https://playground.babylonjs.com/#GLZ1PX#714
in this example, the cone on the right has flipped faces. Using the new flag will fix this issue.


sebavan (approved):
Looks good to me :-) just need to fix the formatting and it should be good to go.

@RaananW any thoughts?

Michalzr:
> Looks good to me :-) just need to fix the formatting and it should be good to go.
> 
> @RaananW any thoughts?

Thanks :)
I fixed the formatting, added a jsdoc for the new parameter and fixed a bug.

When the pipeline passes I will mark this PR ready for review

RaananW (approved):
LGTM

PR #12962, opened 2022-09-07
Issue158

amritanarasimhan:
Dragging the mouse up and down zooms in and out. However, I'm having a problem where if you mouse down on a control and drag, that control is not recognized as selected so dragging the control zooms in and out as well. I was wondering if there was anything to identify when control has been selected other than selectedcontrols?

### DIFF
--- a/packages/tools/guiEditor/src/diagram/workbench.tsx
+++ b/packages/tools/guiEditor/src/diagram/workbench.tsx
@@ -1033,10 +1033,7 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
 
     zoomDrag(event: React.MouseEvent) {
         let delta = 0;
-        if (event.movementY < 0) {
-            delta = -event.movementY;
-        }
-        if (event.movementY > 0) {
+        if (event.movementY < 0 || event.movementY > 0) {
             delta = -event.movementY;
         }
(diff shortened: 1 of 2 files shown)


packages/tools/guiEditor/src/diagram/workbench.tsx
```diff
@@ -1025,6 +1027,19 @@ export class WorkbenchComponent extends React.Component<IWorkbenchComponentProps
...
+    }
+
+    zoomDrag(event: React.MouseEvent) {
+        let delta = 0;
+        if (event.movementY < 0) {
+            delta = -event.movementY;
+        }
+        if (event.movementY > 0) {
+            delta = -event.movementY;
+        }
```
RaananW, line 1041: I might be missing something, but both of those seem to do the same?

RaananW (commented):
LGTM! just one question

packages/tools/guiEditor/src/diagram/workbench.tsx
amritanarasimhan: You're right! Not sure why I have those as two different conditionals.

RaananW (approved):
Discussed offline, one small change is needed, otherwise good to go!

PR #12968, opened 2022-09-09
Some code reduction in thinEngine

RaananW:
Reuse the same code structure as much as possible.
I don't  use bind because of classes extending ThinEngine (they will need to be changes as well)

### DIFF
--- a/packages/dev/core/src/Engines/thinEngine.ts
+++ b/packages/dev/core/src/Engines/thinEngine.ts
@@ -3208,19 +3208,18 @@ export class ThinEngine {
     // type def for typescript only
     private _setUsingContext(functionName: string, arrayLength: number, uniform: Nullable<WebGLUniformLocation>, ...values: any[]): boolean;
     private _setUsingContext(): boolean {
-        this._args.length = 0;
-        Array.prototype.push.apply(this._args, arguments);
-        if (!this._args[2] || (this._args[1] && (this._args[1] === 1 ? !this._args[3].length : this._args[3] % this._args[1]))) {
+        const a = arguments;
+        for (let i = 2; i < a.length; ++i) {
+            this._args[i - 2] = a[i];
+        }
+        if (!a[2] || (a[1] && (a[1] === 1 ? !a[3].length : a[3] % a[1]))) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Engines/thinEngine.ts
```diff
@@ -3203,20 +3203,34 @@ export class ThinEngine {
         }
     }
 
+    private _args: any[] = [];
+
+    // type def for typescript only
+    private _setUsingContext(functionName: string, arrayLength: number, uniform: Nullable<WebGLUniformLocation>, ...values: any[]): boolean;
+    private _setUsingContext(): boolean {
+        const a = arguments;
```
deltakosh, line 3211: This will be slower right? there is for loop and several tests

We cannot do that so deep in the engine

RaananW: The only thing added was the for, but it can be done using a push instead. the tests all existed before.
This is just a suggestion; I don't mind closing this.

deltakosh: It depends. For setInt for instance it is a loss

I would not recommend touching these functions honestly if you don't mind

sebavan: The code for setInt for instance would be:

```
_setUsingContext() {
        const a = arguments;
        for (let i = 2; i < a.length; ++i) {
            this._args[i - 2] = a[i];
        }
        if (!a[2] || (a[1] && (a[1] === 1 ? !a[3].length : a[3] % a[1]))) {
            return false;
        }
        // Matrix functions need false as 2nd variable
        if (a[0].includes("Matrix")) {
            this._args.unshift(this._args.shift(), false);
        }
        this._gl[`uniform${a[0]}`].apply(this._gl, this._args);
        return true;
    }

_setUsingContext();
```

instead of 

```
if (!uniform) {
  return false;
}

this._gl.uniform1i(uniform, value);
```

I am not sure as well the cost is worth the size reduction here as it looks like the function would lose about 10% perf on a quick benchmark

RaananW: It's already closed

PR #12976, opened 2022-09-13
Add createRadialGradient to ICanvas

yuripourre:
Description borrowed from: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createRadialGradient

### DIFF
--- a/packages/dev/core/src/Engines/ICanvas.ts
+++ b/packages/dev/core/src/Engines/ICanvas.ts
@@ -394,6 +394,18 @@ export interface ICanvasRenderingContext {
     createLinearGradient(x0: number, y0: number, x1: number, y1: number): ICanvasGradient;
 
     /**
+     * Creates a linear gradient along the line given by the coordinates represented by the parameters.
+     * @param x0 The x-axis coordinate of the start circle.
+     * @param y0 The y-axis coordinate of the start circle.
+     * @param r0 The radius of the start circle. Must be non-negative and finite.
+     * @param x1 The x-axis coordinate of the end point.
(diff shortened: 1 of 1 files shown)


RaananW (approved):
LGTM, will let @sebavan merge.

sebavan:
I guess this should only be done if we intent to support it as well on Native otherwise ppl should cast in the web to HTMLElement Canvas.

@bghgary, @CedricGuillemet is it something we can easily add in Native ?

sebavan (changes requested):
Blocking the merge until we get some info from the native team.

CedricGuillemet:
It's partially supported : https://github.com/memononen/nanovg/blob/397f3300bc14bdb856ec429c71c345ab0340b0a6/src/nanovg.h#L414

with native x0,y0 and x1,y1 must be same value.

sebavan:
> It's partially supported : https://github.com/memononen/nanovg/blob/397f3300bc14bdb856ec429c71c345ab0340b0a6/src/nanovg.h#L414
> 
> with native x0,y0 and x1,y1 must be same value.

So it does make sense to merge this one ?

sebavan:
Mainly wondering as it could create discrepancies between the two ?

CedricGuillemet:
> Mainly wondering as it could create discrepancies between the two ?

There are already discrepancies between native and Web. Like drawImage is partially supported and some modes will render nothing or incorrectly. We don't have a lot of traction for Native GUI atm. So, if this PR is mandatory for future dev, merge it.

PR #12843, opened 2022-08-05
Avoid hooking rtt renderList multiple times

fchoisy:
See https://forum.babylonjs.com/t/rendertargettexture-renderlist-crash/31741

PR #12982, opened 2022-09-13
Fix Array Observers

sebavan:
Ref: https://github.com/BabylonJS/Babylon.js/pull/12905


sebavan:
Still need to address https://github.com/BabylonJS/Babylon.js/pull/12843 back

RaananW (approved):

PR #12984, opened 2022-09-14
Fix billboard translation with parenting

CedricGuillemet:
Follow up with https://forum.babylonjs.com/t/billboardmode-rotation-bug/33908/4

billboard translation was not transformed by parent transform.

### DIFF
--- a/packages/dev/core/src/Meshes/transformNode.ts
+++ b/packages/dev/core/src/Meshes/transformNode.ts
@@ -1111,7 +1111,8 @@ export class TransformNode extends Node {
                 TmpVectors.Matrix[7].setTranslation(translation);
 
                 // set localMatrix translation to be transformed against parent's world matrix.
-                this._localMatrix.setTranslation(Vector3.TransformCoordinates(this._position, parent.getWorldMatrix()));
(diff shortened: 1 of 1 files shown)


packages/dev/core/src/Meshes/transformNode.ts
```diff
@@ -1110,6 +1110,9 @@ export class TransformNode extends Node {
                 Matrix.ScalingToRef(scale.x, scale.y, scale.z, TmpVectors.Matrix[7]);
                 TmpVectors.Matrix[7].setTranslation(translation);
 
+                // set localMatrix translation to be transformed against parent's world matrix.
+                this._localMatrix.setTranslation(Vector3.TransformCoordinates(this._position, parent.getWorldMatrix()));
```
RaananW, line 1114: We should use TransformCoordinatesToRef into a temp vector to avoid GC here.

RaananW (commented):
Just a small comment, otherwise LGTM!

RaananW:
Oh, and the billboard playground fails

sebavan (approved):

PR #12992, opened 2022-09-18
test(particles.cloudPoint): add tests for intersectsMesh function

Dok11:
Hi!
I add some tests to Particles/cloudPoint, the function is intersectsMesh.

Here I found some points.
1. The property of PointsCloudSystem `mesh` has incorrect type. In fact this property may have `undefined` as value, because it's not defined in the constructor. So I add optional mark `?` to type and required checks in the class.
2. And after this I found what the `intersectsMesh` function did not check the mesh property existing in the point cloud system. So I add a new Error here, that notice about required mesh to next calculations.
3. The line `isSphere = isSphere ? isSphere : false;` look like redundant, because TypeScript promise that it have boolean type.
4. Next I move the getBoundingInfo call into one line `const bbox = target.getBoundingInfo().boundingBox;`. It's look for me as potential performance boost, maybe.
5. And at the end I remove `else` branch because it rendundant because the code above have return statements.

And I limited test cases from 729 possible combinations to 126, they seems like corner cases for me. I dont know which cases we can reduce to save the working guarantee.

### DIFF
--- a/packages/dev/core/src/Particles/cloudPoint.ts
+++ b/packages/dev/core/src/Particles/cloudPoint.ts
@@ -149,29 +149,29 @@ export class CloudPoint {
         if (!target.hasBoundingInfo) {
             return false;
         }
-        isSphere = isSphere ? isSphere : false;
+
+        if (!this._pcs.mesh) {
+            throw new Error("Point Cloud System doesnt contain the Mesh");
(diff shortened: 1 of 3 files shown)


RaananW (approved):
LGTM!

PR #12999, opened 2022-09-20
Move from @hidden to @internal

RaananW:
Setting as Draft for now to check side effects
There shouldn't be any, just need to be 100% sure :-)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -96,6 +96,10 @@
         "react": {
             "pragma": "h",
             "createClass": ""
+        },
+        "jsdoc": {
+            "ignorePrivate": true,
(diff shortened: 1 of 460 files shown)


sebavan:
Is it not messing the playground checks for typings ? I remember we had special rules there but can not remember if there were for @hidden or only removing _ functions.

RaananW:
> Is it not messing the playground checks for typings ? I remember we had special rules there but can not remember if there were for @hidden or only removing _ functions.

I could not find any reference of hidden in the playground, and have checked the playground - it doesn't show any hidden/internal options.

sebavan (approved):

PR #13004, opened 2022-09-21
Add rotation from one vector3 to another

BabylonJSGuide:
First commit wherever a result was returned this is now reflected in comments. Only comments changed, if current vector was returned then this was kept and and comment not changed. Ref https://forum.babylonjs.com/t/comments-refer-to-wrong-vectors-returned-by-method-using-toref/34093/3?u=johnk

An alternative way of managing an animation along a curve by calculating the rotation needed from a set rotation to a tangent. Example PGs to follow.


packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -4113,6 +4173,34 @@ export class Quaternion {
...
+    /**
+     * Creates the rotation quaternion needed to rotate from one Vector3 onto another Vector3 and stores in a result Quaternion
+     * @param fromVector the starting vector
+     * @param toVector the ending vector
+     * @param result the rotation quaternion needed
+     * @returns the result
+     */
+    public static RotationQuaternionFromOntoToRef(fromVector: DeepImmutable<Vector3>, toVector: DeepImmutable<Vector3>, result: Quaternion) {
+        const normal: Vector3 = Vector3.Cross(fromVector, toVector);
+        if (normal.equals(Vector3.Zero())) {
```
RaananW, line 4196: Better to use `Vector3.ZeroReadOnly` to avoid allocating a new vector

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -4113,6 +4173,34 @@ export class Quaternion {
...
+            fromVector.getNormalToRef(normal);
+        }
+        const angle = Vector3.GetAngleBetweenVectors(fromVector, toVector, normal);
+        result = Quaternion.RotationAxis(normal, angle);
```
RaananW, line 4200: wouldn't that create a new quaternion? Is there RotationAxisToRef?

RaananW (commented):
Just two comments! thanks :-)

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1107,18 +1107,54 @@ export class Vector3 {
...
+        const ref: Vector3 = MathTmp.Vector3[0];
+        return this.getNormalToRef(ref);
+    }
+
+    /**
+     * Creates a vector normal (perpendicular) to the current Vector3 and stores the result in the given vector
+     * @param result defines the Vector3 object where to store the resultant normal
+     * returns the result
+     */
+    public getNormalToRef(result: DeepImmutable<Vector3>): Vector3 {
```
sebavan, line 1130: is there not an infinity of possibility ???

sebavan (changes requested):

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: Makes sense. Done. 
Also made more use of ToRef methods

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: Yes there is and have changed it. Done

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: There is any of which will allow a rotation around it as axis of the cuurent vector about Math.PI radians which is what I need. This function produces one such normal.

Its use is in `RotationQuaternionFromOntoToRef` about line 4198.

Perhaps as it is only designed for use within this method that if you think it better I could remove the `getNormal` method all together and make the `getNormalToRef` method a `@hidden` one and rename it as `_getNormalToRef` to show its not for the user.

Also in this case would there be any benefit to changing to to a static method?

RaananW (commented):
changes look good to me. i'll let @sebavan give the final ok

packages/dev/core/src/Maths/math.vector.ts
sebavan: Yup I would rather keep it private as it is not clear what other use it would have.

There is no difference with static.

dr-vortex:
@BabylonJSGuide I tried to make a PG with this but it does not work: https://playground.babylonjs.com/#QYA3JC#22

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: Will make changes tomorrow. Let you know when done.

BabylonJSGuide:
@dr-vortex Will produce a working PG tomorrow. Will also study your PR to see if it has any advantages.

dr-vortex:
Thanks!

Might I also recommend we move the rotation between vectors related stuff to #13012?

sebavan (changes requested):
Flagging as Request changes to prevent accidental merge waiting on https://github.com/BabylonJS/Babylon.js/pull/13012 comments resolution.

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: Done

BabylonJSGuide:
> Thanks!
> 
> Might I also recommend we move the rotation between vectors related stuff to #13012?

No for the following reasons.

1. Each need to stand seperately on their own merits.
2. At the moment I do not understand the way your function is used and would need to see if they were sufficiently differences
3.  As I stated earlier I will study your method and get to understand it and see if it has any advantages/differences

BabylonJSGuide:
Finally undertsand @dr-vortex 's method. As a mathematician when I talk about vectors I automatically think direction, whilst for his finction he his thinking position. My function finds the rotation between to direction vectors, whilst his find the rotation needed for the mesh to travel along a straight line joining two position vectors or points. Hence the confusion.

dr-vortex examples that  I now follow https://playground.babylonjs.com/#QYA3JC#28, https://playground.babylonjs.com/#QYA3JC#29

@RaananW and @sebavan I will leave it up to you whether you decide whether either, both or neither are sufficiently different to what is already possible to accept the PRs.

I have made some suggests on #13012

BabylonJSGuide:
First PG test for my method https://playground.babylonjs.com/?snapshot=refs/pull/13004/merge#S7HDGQ
2nd test for my method https://playground.babylonjs.com/?snapshot=refs/pull/13004/merge#S7HDGQ#1

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1661,6 +1689,30 @@ export class Vector3 {
         return s;
     }
 
+    /**
+     * Get rotation needed to to rotate from one Vector3 onto another Vector3
+     * @param fromVector the starting vector
+     * @param toVector the ending vector
+     * @returns the rotation needed
+     */
+    public static RotationFromOnto(fromVector: DeepImmutable<Vector3>, toVector: DeepImmutable<Vector3>) {
```
sebavan, line 1698: We should remove the functions here. It is adding extra code and I would prefer to be mindful with package size. Let s keep it only in the quaternion class ?

sebavan (changes requested):
Only one tiny comment and good to go !!! thanks

packages/dev/core/src/Maths/math.vector.ts
BabylonJSGuide: Reasonable, do it tomorrow.

sebavan: no rush it won t be merged before Monday anyway ;-)

BabylonJSGuide: Done

BabylonJSGuide:
1st test rotationQuaternion method https://playground.babylonjs.com/?snapshot=refs/pull/13004/merge#S7HDGQ#6
2nd test rotationQuaternion method https://playground.babylonjs.com/?snapshot=refs/pull/13004/merge#S7HDGQ#7

PR #13006, opened 2022-09-21
Add smoothing for freeCameraDeviceOrientationInputs

ilrico:
Add a simple low-pass filter on alpha/beta/gamma to limit objects "shaking"

### DIFF
--- a/packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
+++ b/packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
@@ -16,6 +16,7 @@ declare module "../../Cameras/freeCameraInputsManager" {
         _deviceOrientationInput: Nullable<FreeCameraDeviceOrientationInput>;
         /**
          * Add orientation input support to the input manager.
+         * @param {number} [smoothFactor] deviceOrientation smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
          * @returns the current input manager
          */
         addDeviceOrientation(smoothFactor?: number): FreeCameraInputsManager;
@@ -24,6 +25,7 @@ declare module "../../Cameras/freeCameraInputsManager" {
 
 /**
  * Add orientation input support to the input manager.
+ * @param {number} [smoothFactor] deviceOrientation smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
  * @returns the current input manager
  */
 FreeCameraInputsManager.prototype.addDeviceOrientation = function (smoothFactor?: number): FreeCameraInputsManager {
--- a/packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
+++ b/packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
@@ -18,7 +18,7 @@ declare module "../../Cameras/freeCameraInputsManager" {
          * Add orientation input support to the input manager.
          * @returns the current input manager
          */
-        addDeviceOrientation(): FreeCameraInputsManager;
+        addDeviceOrientation(smoothFactor?: number): FreeCameraInputsManager;
     }
 }
 
@@ -26,9 +26,12 @@ declare module "../../Cameras/freeCameraInputsManager" {
  * Add orientation input support to the input manager.
  * @returns the current input manager
  */
-FreeCameraInputsManager.prototype.addDeviceOrientation = function (): FreeCameraInputsManager {
+FreeCameraInputsManager.prototype.addDeviceOrientation = function (smoothFactor?: number): FreeCameraInputsManager {
     if (!this._deviceOrientationInput) {
         this._deviceOrientationInput = new FreeCameraDeviceOrientationInput();
+        if(smoothFactor) {
+            this._deviceOrientationInput.setSmoothFactor(smoothFactor)
+        }
         this.add(this._deviceOrientationInput);
     }
 
@@ -47,6 +50,8 @@ export class FreeCameraDeviceOrientationInput implements ICameraInput<FreeCamera
 
     private _constantTranform: Quaternion;
     private _screenQuaternion: Quaternion = new Quaternion();
+
+    private _smoothFactor: number = 0;
 
     private _alpha: number = 0;
     private _beta: number = 0;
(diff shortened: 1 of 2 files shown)


packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -172,9 +177,9 @@ export class FreeCameraDeviceOrientationInput implements ICameraInput<FreeCamera
     };
 
     private _deviceOrientation = (evt: DeviceOrientationEvent) => {
-        this._alpha = evt.alpha !== null ? evt.alpha : 0;
-        this._beta = evt.beta !== null ? evt.beta : 0;
-        this._gamma = evt.gamma !== null ? evt.gamma : 0;
+        this._alpha = evt.alpha !== null ? Tools.SmoothAngleChange(this._alpha, evt.alpha, this._smoothFactor) : 0;
+        this._beta = evt.beta !== null ? Tools.SmoothAngleChange(this._beta, evt.beta, this._smoothFactor) : 0;
+        this._gamma = evt.gamma !== null ? Tools.SmoothAngleChange(this._gamma, evt.gamma, this._smoothFactor) : 0;
```
RaananW, line 182: SmoothAngleChange could be very expensive. It seems like there is no reason to call it if smoothFactor is 0 (which is the default). We can save those calles and only run it is the factor is > 0

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -222,6 +227,14 @@ export class FreeCameraDeviceOrientationInput implements ICameraInput<FreeCamera
...
         return "deviceOrientation";
     }
+
+    /**
+     * Set the smoothFactor for devive orientation sensors, to limit objects shaking
+     * From AR.js documentation, best value should be 0.9
+     */
+     public setSmoothFactor(smoothFactor=0.9): void {
+        this._smoothFactor = smoothFactor
+    }
```
RaananW, line 237: any reason why we don't want to provide a getter (or avoid getting/setting altogether)? A public member will be fine. Otherwise - there i no reason to provide a default value in a setter.

RaananW (commented):
A few comments. thanks! :-)

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
ilrico: 100% agree, first implementation did branch those 2 cases. I reverse to it.

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
ilrico: every property in the class was private, so I assumed it was a design choice, but I also think we are adults (my python background ;-) ). so let's make it public indeed.

RaananW: "adults"

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -16,19 +16,23 @@ declare module "../../Cameras/freeCameraInputsManager" {
...
 /**
  * Add orientation input support to the input manager.
  * @returns the current input manager
  */
-FreeCameraInputsManager.prototype.addDeviceOrientation = function (): FreeCameraInputsManager {
+FreeCameraInputsManager.prototype.addDeviceOrientation = function (smoothFactor?: number): FreeCameraInputsManager {
     if (!this._deviceOrientationInput) {
         this._deviceOrientationInput = new FreeCameraDeviceOrientationInput();
+        if(smoothFactor) {
+            this._deviceOrientationInput.smoothFactor = smoothFactor
```
PolygonalSun, line 34: nit: semi-colon

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -52,6 +56,8 @@ export class FreeCameraDeviceOrientationInput implements ICameraInput<FreeCamera
     private _beta: number = 0;
     private _gamma: number = 0;
 
+    public smoothFactor: number = 0;
```
PolygonalSun, line 59: Need comment for public variables

PolygonalSun (commented):
Looks okay but I'd recommend running `npm run format:fix` since it looks like the formatting/linting task is failing.

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
ilrico: comment added for addDeviceOrientation() param, but I feel public variable comment has already been added

RaananW (approved):

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -52,6 +57,9 @@ export class FreeCameraDeviceOrientationInput implements ICameraInput<FreeCamera
     private _beta: number = 0;
     private _gamma: number = 0;
 
+    // alpha+beta+gamma smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
```
RaananW, line 60:
```suggestion
    /** alpha+beta+gamma smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing */
```

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -16,19 +16,24 @@ declare module "../../Cameras/freeCameraInputsManager" {
...
          * @returns the current input manager
          */
-        addDeviceOrientation(): FreeCameraInputsManager;
+        addDeviceOrientation(smoothFactor?: number): FreeCameraInputsManager;
     }
 }
 
 /**
  * Add orientation input support to the input manager.
+ * @param {number} [smoothFactor] deviceOrientation smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
```
RaananW, line 28:
```suggestion
 * @param smoothFactor deviceOrientation smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
```

packages/dev/core/src/Cameras/Inputs/freeCameraDeviceOrientationInput.ts
```diff
@@ -16,19 +16,24 @@ declare module "../../Cameras/freeCameraInputsManager" {
         _deviceOrientationInput: Nullable<FreeCameraDeviceOrientationInput>;
         /**
          * Add orientation input support to the input manager.
+         * @param {number} [smoothFactor] deviceOrientation smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
```
RaananW, line 19:
```suggestion
         * @param smoothFactor deviceOrientation smoothing. 0: no smoothing, 1: new data ignored, 0.9 recommended for smoothing
```

PolygonalSun (approved):

sebavan (approved):

ilrico:
I assume tag <new feature> should be added for changelog but could find out how to edit tags. I guess I don't have permission.

PR #13011, opened 2022-09-22
Tiny perf fix

deltakosh:


sebavan (approved):

packages/dev/core/src/scene.ts
```diff
@@ -4003,12 +3993,15 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
             }
         }
 
+        const forcePush =
```
Popov72, line 3996: Is removing `mesh.subMeshes.length === 1` in the conditions to force push intended?

deltakosh: yes as this is not a submesh dependency. I wanted to avoid doing all the unnecessary tests per submesh (they can be done per mesh)

PR #13012, opened 2022-09-22
Add function to get angles between two vectors (Vector3.GetAnglesBetweenVectorsForDirectionChange)

dr-vortex:
Also add prettier settings to workspace settings.

See [this](https://github.com/BabylonJS/Babylon.js/pull/12942) PR for more details.

Demo PGs: 
https://playground.babylonjs.com/#QYA3JC#25
https://playground.babylonjs.com/#QYA3JC#26
(Press space to start)

@BabylonJSGuide 
I have the opportunity to talk with a physicist almost every day of the week, so I asked him about the problem and worked out the same equations used. After researching how `Math.atan2` works though, and found an issue with the equations: for the phi calculation, the `diff.y` should have been the second parameter of `Math.atan2`, not the first.


sebavan (changes requested):
Is it not similar to https://github.com/BabylonJS/Babylon.js/pull/13004

cc @RaananW for prettier

dr-vortex:
> Is it not similar to #13004 

@sebavan,
How is this related to #13004?

sebavan:
> > Is it not similar to #13004
> 
> @sebavan, How is this related to #13004?

RotationQuaternionFromOntoToRef seems to be doing the exact same finding the required rotation bewtween two vectors.

.vscode/settings.json
```diff
@@ -15,5 +15,10 @@
     },
     "[typescript]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"
-    },
-}
\ No newline at end of file
+	},
```
RaananW, line 18: why do we need that? the prettier plugin should take the configuration  in the repo

dr-vortex: Using the prettier extension on VSCode does not use the prettier config file.

dr-vortex:
> RotationQuaternionFromOntoToRef seems to be doing the exact same finding the required rotation bewtween two vectors.

https://playground.babylonjs.com/#QYA3JC#24 It does not seem to work.

sebavan:
> > RotationQuaternionFromOntoToRef seems to be doing the exact same finding the required rotation bewtween two vectors.
> 
> https://playground.babylonjs.com/#QYA3JC#24 It does not seem to work.

Let s figure it out and only keep one implementation. We need to be careful and prevent duplicated features.

dr-vortex:
Agreed, could we maybe move it to this PR?

I'd like to know what @BabylonJSGuide thinks of this PR since we worked on it in #12942.

sebavan:
Not sure it should be moved in this PR, lets compare the functionalities and keep the one closest from the goal. You can check with @BabylonJSGuide.

.vscode/settings.json
RaananW: i'll check that gladly, but it doesn't belong in this PR.

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1716,6 +1716,35 @@ export class Vector3 {
...
+        return ref;
+    }
+
+    /**
+     * Gets the angles between two vectors needed to change direction
+     * @param start the starting vector
+     * @param target the target vector
+     * @returns the angles between the two vectors in the form (pitch, yaw, roll)
+     */
+    public static GetAnglesBetweenVectorsForDirectionChange(start: Vector3, target: Vector3): Vector3 {
```
BabylonJSGuide, line 1740: Better to call ToRef version

BabylonJSGuide:
@dr-vortex now understand the difference - yours deals with positions mine with directions.

Suggestions 

1. Be clear about which axis of the mesh is following the curve

1. A better name would be pitchYawRollToMoveBetweenPoints(point0, point1)

2. Change comment from

* Gets the angles between two vectors needed to change direction

to something like

* gets the pitch, yaw, roll that enables a mesh/object(?) to move along a straight line between two points (vector3)

4. Possibly add an optional roll parameter, ie pitchYawRollToMoveBetweenPoints(point0, point1, roll?) which would default to zero but could be updated by the user to affect the roll as it follows the curve.

Need to wait to see if @RaananW and @sebavan think there is sufficient differences between both and/or between existing methods for none, one or both to be approved.

dr-vortex:
Yes, I've been thinking about it with positions. For your 4th point, I had originally included an offset in #12942 but removed it. I agree we should wait to see if they should be implemented separately.

.vscode/settings.json
dr-vortex: Sounds good. I'll open another PR for it.

sebavan: You should revert the change on this one :-)

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1716,6 +1716,32 @@ export class Vector3 {
...
     }
 
+    /**
+     * Gets the angles between two points needed to face the target from the start
+     * @param start the starting point
+     * @param target the target point
+     * @param ref the vector3 to store the result
+     * @returns the updated ref with the angles between the two points in the form (pitch, yaw, roll)
+     */
+    public static GetAnglesBetweenPointsForDirectionChangeToRef(start: Vector3, target: Vector3, ref: Vector3): Vector3 {
```
sebavan, line 1726: Please rename to something closer to what @BabylonJSGuide proposed.

sebavan (changes requested):
Ok both PR fills different needs and are valid. Let s just rename and document more in this one what the fonction is for and good to go.

Would be great to attach unit tests as well if you can to ensure we do not regress later on ?

sebavan:
Let s validate the naming and doc comment with @BabylonJSGuide and ready to ship :-)

dr-vortex:
Sounds good!

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1716,6 +1716,32 @@ export class Vector3 {
         return Scalar.NormalizeRadians(angle);
     }
 
+    /**
+     * Gets the angles between two points needed to face the target from the start
```
BabylonJSGuide, line 1720: Since the method return a Vector3 this cannot be desribed as angles. Change comments to

```
 /**
     * Gets the rotation that aligns the roll axis (Y) to the line joining two points and stores it in the ref Vector3
     * @param start the starting point
     * @param target the target point
     * @param ref the vector3 to store the result
     * @returns the rotation in the form (pitch, yaw, 0)
     */
```

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1716,6 +1716,32 @@ export class Vector3 {
...
+    public static PitchYawRollToMoveBetweenPointsToRef(start: Vector3, target: Vector3, ref: Vector3): Vector3 {
+        const diff = TmpVectors.Vector3[0];
+        target.subtractToRef(start, diff);
+        ref.y = Math.atan2(diff.x, diff.z) || 0;
+        ref.x = Math.atan2(Math.sqrt(diff.x ** 2 + diff.z ** 2), diff.y) || 0;
+        return ref;
+    }
+
+    /**
+     * Gets the angles between two points needed to face the target from the start
```
BabylonJSGuide, line 1735: Since the method return a Vector3 this cannot be desribed as angles. Change comments to

```
 /**
     * Gets the rotation that aligns the roll axis (Y) to the line joining the start point to the target point
     * @param start the starting point
     * @param target the target point
     * @returns the rotation in the form (pitch, yaw, 0)
     */
```

packages/dev/core/src/Maths/math.vector.ts
```diff
@@ -1716,6 +1716,32 @@ export class Vector3 {
...
+     * @param start the starting vector
+     * @param target the target vector
+     * @returns the angles between the two points in the form (pitch, yaw, roll)
+     */
+    public static PitchYawRollToMoveBetweenPoints(start: Vector3, target: Vector3): Vector3 {
+        const ref = Vector3.Zero();
```
BabylonJSGuide, line 1741: use TmpVectors.Vector3[0]

BabylonJSGuide:
Before approval at least the following PG should be tested

https://playground.babylonjs.com/?snapshot=refs/pull/13012/merge#S7HDGQ#5

sebavan:
@dr-vortex I ll merge as soon as @BabylonJSGuide comments have been addressed.

Thanks a lot you both !!!

sebavan (approved):

RaananW (approved):

PR #13014, opened 2022-09-22
More features for the aggressive performance mode

deltakosh:
Some tiny perf changes + a major one: Now in aggressive mode, the renderingGroups are not reset between frames.

### DIFF
--- a/packages/dev/core/src/Rendering/renderingManager.ts
+++ b/packages/dev/core/src/Rendering/renderingManager.ts
@@ -86,6 +86,12 @@ export class RenderingManager {
     private _renderingGroupInfo: Nullable<RenderingGroupInfo> = new RenderingGroupInfo();
 
     /**
+     * Gets or sets a boolean indicating that the manager will not reset between frames.
+     * This means that if a mesh becomes invisible or transparent it will not be visible until this boolean is set to false again.s
+     */
+    public maintainStateBetweenFrames = false;
+
+    /**
      * Instantiates a new rendering group for a particular scene
      * @param scene Defines the scene the groups belongs to
      */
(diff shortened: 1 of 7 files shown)


packages/dev/core/src/scene.ts
```diff
@@ -283,10 +283,21 @@ export class Scene extends AbstractScene implements IAnimatable, IClipPlanesHold
         this._performancePriority = value;
 
         switch (value) {
+            case ScenePerformancePriority.BackwardCompatible:
+                this.skipFrustumClipping = false;
+                this._renderingManager.maintainStateBetweenFrames = false;
+                this.skipPointerMovePicking = false;
+                this.autoClear = true;
+                break;
+            case ScenePerformancePriority.Intermediate:
```
deltakosh, line 292: this change was to allow users to revert back to a different stage

packages/dev/core/src/Rendering/renderingManager.ts
```diff
@@ -85,6 +85,12 @@ export class RenderingManager {
     private _customTransparentSortCompareFn: { [id: number]: Nullable<(a: SubMesh, b: SubMesh) => number> } = {};
     private _renderingGroupInfo: Nullable<RenderingGroupInfo> = new RenderingGroupInfo();
 
+    /**
+     * Gets or sets a boolean indicating that the manager will not reset between frames.
+     * This means that if a mesh becomes invisible or transparent it will not be visible until this boolean is set to false again.s
+     */
+    public maintainStateBetweenFrames = false;
```
Popov72, line 92: I think you will need a setter that will reset all the `_wasDispatched` properties to `false` when `maintainStateBetweenFrames` is switched from `true` to `false`?

deltakosh: yeah this is a good catch! PR is coming!

